PR #1, opened 2026-07-19
Initialize repository with harness code and assets

hiyouga:
## Summary

Initial import — populates the repository with all source code, config, and README assets.

**Included (544 files):**
- `packages/` — `cli`, `core`, `server`, `web`, `docs`, `landing`, `skills`
- `scripts/`, `install.sh`
- `README.md`, `README.zh.md`, `LICENSE`
- `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `tsconfig.base.json`, `tsconfig.json`
- `.github/` workflows, `.gitignore`, `.prettierignore`, `.prettierrc.json`, `.env.example`

> ⚠️ Heads-up: `.github/workflows/` (CI, release, pages) may reference external repos or secrets — review before enabling Actions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018ihk8iQuo3kv2aPjAYEPuR

PR #7, opened 2026-07-20
Changelog, dev startup, README, AgentHub 0.4.0, Qwen Token Plan

hiyouga:
## Summary

Four commits, six related changes:

### 1. Per-version changelog directory

`changelog/` records every update in English, one folder per release version: each change adds `<version>/YYYY-MM-DD-<semantic-id>.md` under the next unreleased version (`0.0.2/`), and each version folder's `README.md` indexes its entries with one-sentence summaries (the top-level `changelog/README.md` documents only the layout, mirroring agenthub PR #162).

### 2. Conflict-free dev prestep (`scripts/dev-prebuild.mjs`) with auto `pnpm install`

The skills+core prebuild behind `dev:server`/`dev:web` is serialized by an on-disk lock with a 5s success stamp: concurrent starts build exactly once instead of clobbering each other's `dist/` (tsup `clean: true`), while edit-then-restart still rebuilds. The prestep also keeps `pnpm install` current automatically — a lockfile-hash stamp means fresh clones and pulled lockfile changes install before starting while an up-to-date tree pays nothing; the lock and stamps live in the OS temp dir keyed by the repo path (works before node_modules exists), and `dev:docs`/`dev:landing` run the install check only.

### 3. Combined and missing dev commands

`pnpm dev` (server+web via concurrently, prefixed logs), `pnpm dev:landing` (port 7366). `concurrently` added as a root devDependency; READMEs updated.

### 4. README restructured around the product story

Agents-build-agents pitch + community links; three showcases — benchmark charts (`assets/readme/benchmark-*.svg` from the landing data), a one-sentence RAG demo with real product screenshots (new `capture-readme-demo.mjs`), self-evolution with a video placeholder; supported-models table; Web-first installation; roadmap; new `CONTRIBUTING.md`; citation. `README.zh.md` mirrors.

### 5. AgentHub 0.4.0 + opaque fidelity payload

`@prismshadow/agenthub` 0.3.3 → 0.4.0 (agenthub PR #159). OmniMessage complete payloads replace item-level `signature`/`phase` with one opaque `fidelity` object carried verbatim through Trace/replay/resume; the streaming translator mirrors AgentHub baseClient aggregation (a run of equal fidelity is one thinking block — carries the reasoning-field replay fix through PenguinHarness; text splits on `fidelity.phase`, closes on `fidelity.signature`). Malformed classification adapted to 0.4's `ToolCallArgumentParseError`/`EmptyResponseError` so truncated streams still reconnect and retry (was a silent regression, caught by the malformed e2e). The `agenthub-dev` skill joins the built-in library (Penguin Development group, completed to the library contract). Docs (omni-message / interfaces / sessions-and-traces, en+zh) updated. Pre-release policy: old traces with `signature`/`phase` are not migrated (fields ignored on replay).

### 6. Qwen Token Plan provider group

New subscription-gateway group `qwen-token-plan` after SiliconFlow: preset base URL `https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1`, key/model-id links to platform.qianwenai.com, `OPENAI_*` env fallback. Five models (`qwen3.8-max-preview`, `qwen3.7-max`, `qwen3.7-plus`, `glm-5.2`, `deepseek-v4-pro`) with `client_type: openai` + inlined endpoint; vision flags verified against the plan's supported-model table (3.8-max-preview and 3.7-plus see images). Pricing and context windows from each model's official page (CNY list prices; the preview model has only a quota promo and stays unpriced). Catalog uniqueness corrected from bare model id to the `(provider, model_id)` pair (gateways resell vendor models under upstream ids). Provider logo is the official Qwen emblem (icon only) flattened to currentColor; models docs (en/zh) updated. A second group, **Qwen Pay-As-You-Go** (DashScope's OpenAI-compatible endpoint), sits below Token Plan with qwen3.7-max, qwen3.7-plus, and the vendor-prefixed resales kimi/kimi-k3 and ZHIPU/GLM-5.2, priced from each model's official page; it shares the Qwen emblem and `modelHomepageUrl` URL-encodes the slash-prefixed ids. A **Fireworks AI** gateway group (below OpenRouter, OpenAI protocol, `https://api.fireworks.ai/inference/v1`) adds five presets — GLM-5.2, Kimi K2.7 Code, DeepSeek V4 Pro, MiniMax M3, DeepSeek V4 Flash — priced from each Fireworks page, with API ids in the full `accounts/fireworks/models/<slug>` form and per-model homepage mapping. Catalog entries within each provider are hand-reordered to dictionary order with newer versions of a series first (no runtime sorting).

### 7. Models page "sync presets" button

An owner-only button next to the search box merges the built-in catalog into the Project's model table (`catalog-sync.ts`, pure + unit-tested): union keyed by `(provider, model_id)` — missing catalog entries are added (gateway base URLs preset), intersecting entries are reset to the catalog's fields (the catalog wins on any difference, including removing pricing the catalog doesn't carry), locally added models are kept verbatim. **Credentials are structurally untouched** — merged rows submit no `apiKey`, so the PUT full-table replace keeps stored keys. Toasts report "N added, M updated" / "already up to date" (no PUT when clean).

### 8. OpenRouter catalog expansion

The OpenRouter group grows from 4 to 18 entries — Claude Fable 5 / Opus 4.8 / Opus 4.7 / Sonnet 5, GPT-5.6 Sol / Terra / GPT-5.5, Kimi K3, GLM-5.2, DeepSeek V4 Pro / Flash, Grok 4.5, Gemini 3.5 Flash, and a free Nemotron 3 Ultra tier — with pricing, context windows, and vision flags taken from each model's OpenRouter page. Pages list no cache pricing, so cache_read carries the standard input price; the `:free` tier stores a genuine $0 price (costs compute to 0) with its own case in the pricing invariant.

### 9. Model-test probe: thinking-only responses count as reachable

Testing a reasoning-heavy model (e.g. `qwen3.8-max-preview`) failed with `OpenaiClient returned no content other than thinking (finish_reason="length")` — the probe's deliberate `maxTokens: 16` cap gets burned entirely on thinking when the upstream ignores the disabled thinking level, and AgentHub 0.4's `EmptyResponseError` surfaced as a test failure. `testModel` now tracks streamed model content and passes a `malformed` ending that arrived after genuine content (endpoint, key, and model id all demonstrably work); timeouts, auth failures, and empty malformed endings still fail. Pure `isProbeContent`/`probeVerdict` helpers with unit tests covering the exact reported case.

### 10. Skill-library update reminder + model-page refinements

- **Skill library**: the installed-skills snapshot tracks each copy's version; cards show an accent one-click "update all outdated Agents" button when any install is older than the library, and the manage dialog marks outdated rows with a per-Agent Update button. Group borders removed; card icons use the theme-accent background.
- **Model page**: changing the default model resets stored drafts to follow the new default; the chat model dropdown mirrors the library page's order (`orderModelsLikeLibrary`); GPT models are uniformly vision-capable (gpt-5.6-sol/terra flipped); model cards link to each model's homepage (`modelHomepageUrl`: per-model gateway URLs, vendor docs fallback).

### 11. Group speed test + vendor-page data

Each model group header gains an owner-only speed test: a confirmation dialog warns about quota, then the group's models are probed **one at a time** (TTFT + TPS via the test endpoint's new `speed` flag, output cap 64), with tone-colored badges (green/yellow/red, clock = TTFT ms, zap = tok/s) landing on each card as results arrive. The model-homepage link moves from the card corner into the config dialog. Catalog data: the official Fireworks burst logo, SiliconFlow gains Kimi-K2.7-Code and DeepSeek-V4-Flash, and Z.AI / Moonshot models link to per-model docs pages.

## Verification

- `pnpm format:check`, `pnpm typecheck` (7 packages), `pnpm build` — clean
- `pnpm test` — 1,063 unit tests green across 7 packages
- Playwright e2e — 14/14 (including the malformed-retry regression the fidelity upgrade would otherwise have shipped)
- Benchmark SVGs, RAG screenshots, and the new provider logo rendered and visually checked in light + dark

### 12. Landing trace screenshots: same opened timeline in en and zh

The English trace shots shipped the empty "select a Session" state while the Chinese ones showed a full trace. The capture script's `/traces?sessionId=` deep link is ignored without `agentId` (the product's own links always carry both), and the fallback title click silently failed for English. The script now uses the canonical `agentId+sessionId` deep link and waits for the timeline's `exec_command` lanes before shooting — an empty capture fails loudly instead of shipping. The scripted English title also shrinks to 27 chars to stay under core's 30-char clip (it rendered as "…Agent ap"). All four `traces-*.webp` now show the same opened trace: overall stats plus an execution timeline with two tool calls.

### 13. One story across landing, README, and blog + penguin.ooo/install.sh

`curl -fsSL https://penguin.ooo/install.sh | sh` now works: the site ships a thin `install.sh` forwarding to the latest GitHub release installer, and every surface (hero, READMEs, docs, blog) switches to it. The landing hero retells the README's 1×/100× story with a "zero-code Harness CLI and Web UI" subtitle; a single numbered **Why** section (1 benchmark suites, 2 one-sentence build demo, 3 self-evolution + video-soon pill) replaces Pillars/Showcase/Benchmark/SelfImprove, with nav anchors updated and `#benchmark` preserved. READMEs restructure to match: shields.io badges, no top screenshot, numbered emoji reasons, one-line any-OpenAI-protocol models note, requirements table, emoji installs, desktop + Windows roadmap items, linked credits (LlamaFactory / PrismShadow / Fable 5). The intro blog post leads with the same story.

### 14. Announcement bar, AMD Fireworks-credits post, GDPevo launch story

A rotating announcement bar above the nav (auto-rotate, hover pause, chevron switching, bilingual) carries two entries: Kimi K3 + Qwen 3.8 Max availability → models docs, and the $50 Fireworks credits campaign → the new `fireworks-credits-amd` post. That post announces the AMD AI Developer Program partnership with the full application walkthrough (screenshots adapted from WhatGhost's guides, credited) and a three-step PenguinHarness setup. The launch post is rewritten to open with the GDPevo origin story (self-evolution validated in [GDPevo](https://prism-shadow.github.io/GDPevo/), brought to everyone as the founding motivation), then mirrors the README — numbered reasons with self-hosted `/blog-assets/` images, security contract, models table, usage, roadmap, and a closing community call-to-action.

### 15. Landing back to the classic structure; tabbed features/cases; docs nav parity

The section-14-era overhaul is walked back: rotating hero headline, three pillars, self-improvement loop (+ demo-video pill), Quickstart, standalone Benchmark, CONTRACT.md, and Security all return. The 1×/100× LangChain story becomes its own two-card comparison section instead of the hero. The nine features turn into switchable tabs whose active panel embeds the real chat / Trace / evaluation screenshots (Trace + benchmark shots are back on the page); use cases become a tabbed Cases section (RAG only for now); a community links section closes the page. The docs nav now mirrors the landing nav exactly for seamless cross-site navigation, and the launch blog post returns to its original skeleton with the GDPevo story, images, models, roadmap, and community sections kept as additions.

### 16. Landing motion polish + nav highlight fix

The announcement bar turns light brand-blue, links both entries to blog posts, slides horizontally between them (auto-advance, hover pause) with dot indicators, and loses every arrow. The LangChain comparison gains a looping construction animation (gray blocks land one at a time and never finish; the brand-blue skyline rises in ~1.6s and holds). The self-improvement trend charts move to one rAF clock with a value label riding each curve's head. Features split by capture: only the three screenshot-backed views form tabs, the rest stay as the classic grid closed by "and more…". Blog usage guidance is Web-only (no CLI/model-config commands). The nav hover pill — which used to sweep in from the nav's left edge on first hover — now appears and fades in place, in both the landing and docs navs.

### 17. Merge main: restore the persistent nav active highlight

The branch predated main's #6 (nav state + anchor scrolling) and was missing the black current-item chip — and would have reverted that fix on merge. `origin/main` is now merged in: `getActiveNavItem` tracking with hash-synced `/#id` links and `aria-current` chips, `.section-anchor` sticky-header scroll spacing, router-link anchors in footer/hero/CTA, and the docs brand link returning to the landing site — reconciled with the announcement bar, the landing-parity docs nav (its Docs link now carries the current-page chip), and the in-place hover pill.

https://claude.ai/code/session_01GQpKq27ARkzj3JMMxkPEKg


hiyouga:
## Code review

Reviewed at `d747265`. The diff was split into 11 subsystem slices reviewed in parallel, and every candidate finding went through an adversarial verifier instructed to refute it: 60 candidates → 30 confirmed / 30 refuted, deduping to **23 distinct issues**.

### Verification (independently reproduced)

| Check | Result |
| --- | --- |
| `pnpm typecheck` | exit 0, 7 packages |
| `pnpm format:check` | clean |
| `pnpm test` | exit 0 — **1,101 passing**, 7 packages |

One caveat: [`packages/core/test/llm.e2e.test.ts`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/core/test/llm.e2e.test.ts) is **skipped** (live-API gated). It is the only test that would exercise the new fidelity translator against a real endpoint, so the AgentHub 0.4.0 migration — the riskiest change here — has no real-provider coverage in a normal run.

---

## Blocking

### 1. `inferProviderForUpstream` still assumes globally-unique model ids → wrong provider, wrong endpoint, misrouted key

[`packages/core/src/state/model-catalog.ts#L778`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/core/src/state/model-catalog.ts#L778)

The headline invariant change is bare `model_id` → `(provider, model_id)`, but this function was left as `MODEL_CATALOG.find((m) => m.modelId === upstreamId)?.provider ?? "custom"`, and its doc comment at L775 still asserts ids are globally unique. The new catalog breaks that — four ids are now duplicated:

| id | providers | inferred |
| --- | --- | --- |
| `glm-5.2` | qwen-token-plan, zhipu | qwen-token-plan |
| `qwen3.7-max` | qwen-token-plan, qwen-pay-as-you-go | qwen-token-plan |
| `qwen3.7-plus` | qwen-token-plan, qwen-pay-as-you-go | qwen-token-plan |
| `deepseek-v4-pro` | deepseek, qwen-token-plan | deepseek |

First literal position wins, silently. `glm-5.2` is a **behavior regression** — on `main` it resolved to `zhipu`. This function is the CLI's default for `--provider` ([`config.ts#L140`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/cli/src/commands/config.ts#L140)), so:

```sh
penguin config model add --model-id glm-5.2 --api-key <Z.AI key> --set-default
```

writes a Z.AI credential onto the Qwen Token Plan entry and ships it to `token-plan.cn-beijing.maas.aliyuncs.com` on the first request.

The existing test only asserts `deepseek-v4-pro → deepseek`, which passes purely by literal ordering.

**Fix:** return ambiguous/`undefined` on multi-match and make the CLI require `--provider` in that case; update the L775 comment; add tests for `glm-5.2` and `qwen3.7-max`.

### 2. Escape in the slash menu wipes the entire composer

[`packages/web/src/features/chat/chat-input.tsx#L846`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/web/src/features/chat/chat-input.tsx#L846)

The Escape branch (`setText(""); onTextChange?.("")`) dates from when the menu could only open on a draft that *started* with `/`, where clearing the body was equivalent to removing the command. This PR made slash matching positional (`matchSlash(text, caret)` at L698), so the menu now opens on a `/token` anywhere in the draft — but Escape still nukes the whole body. The textarea is controlled and the value is replaced programmatically, so **Ctrl+Z does not recover it**.

Repro: type `Refactor the streaming parser and drop the legacy path /com`, watch the menu open on `/com`, press Escape → the entire draft is gone.

The `@`-mention branch immediately below (L869) deliberately does *not* clear, and its comment explaining why ("the `@` prefix is part of the text body") is now equally true of `/`.

**Fix:** mirror the mention branch — track a `slashDismissed` offset and close the menu without touching the text. At minimum use `removeSlashToken(text, slashTok)`.

---

## Should fix

### 3. `penguin.ooo/install.sh` silently discards all flags

[`packages/landing/public/install.sh#L11`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/landing/public/install.sh#L11)

The forwarder is `curl -fsSL "<latest>/install.sh" | sh` with no `-s -- "$@"`. The real installer parses `--universal` and `--version <v>` ([`install.sh#L28-L43`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/install.sh#L28-L43)), and this PR simultaneously switched *every* documented install command to the forwarder URL (README L105, `installation.en/zh.md` L16, quickstart, blog, `links.ts` L19) while `installation.en.md` L37 still documents `... | sh -s -- --universal`.

Two consequences:

- Unsupported-arch users are told to "re-run with `--universal`" through a channel that can never deliver it — an unbreakable loop with no diagnostic that the flag was dropped.
- `--version vX.Y.Z` pinning silently downgrades to *latest*, which is supply-chain-relevant.

**Fix:** `tmp=$(mktemp); curl -fsSL "$URL" -o "$tmp" || { rm -f "$tmp"; exit 1; }; sh "$tmp" "$@"; rc=$?; rm -f "$tmp"; exit $rc`. This also removes the current double-`curl | sh` partial-execution exposure — notable because [`install.sh#L119`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/install.sh#L119) does `rm -rf "$INSTALL_DIR/bin" ...` before the mv loop, so a connection cut there wipes an existing install with nothing to restore it.

### 4. Speed-mode TPS measures network latency, not token rate

[`packages/server/src/services/project-config-service.ts#L247`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/server/src/services/project-config-service.ts#L247)

`maxTokens` rises to 64 for speed mode, but both modes share one prompt (L256) that demands *a single word*: `ping - reply with the single word "pong" and nothing else.` So:

- Compliant models emit 1–3 tokens, the 64-token cap never binds, and `windowMs` is dominated by the final usage chunk's round trip. An identical model reports green 66.7 tok/s or yellow 33.3 tok/s depending on 30ms of network jitter.
- Reasoning-heavy models *do* bind the cap, but then end `malformed` with no `token_usage`, so `tps` is omitted entirely.

Neither class produces a usable number.

**Fix:** give speed mode its own prompt that generates real output near the cap (keeping the `<think></think>` hint), and skip `tps` unless `outputTokens` clears a floor (~16) and `windowMs` clears ~100ms.

### 5. `removeSlashToken` rewrites whitespace across the whole draft

[`packages/web/src/features/chat/slash-token.ts#L35`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/web/src/features/chat/slash-token.ts#L35)

The doc comment says it "collapses a doubled space at the seam", but `.replace(/  +/g, " ")` is global and unanchored. Verified by execution:

```js
removeSlashToken("a  b /compact", { start: 5, end: 13 }) // → "a b "
```

The surviving double space is six characters from the removal point and is still destroyed. Any pasted code, YAML, aligned table, or markdown indentation in the draft is silently reflowed the moment a slash command runs. The two new tests only use single-space inputs.

**Fix:** collapse only at the join point, and add a test with an indented body.

### 6. Raw NUL byte makes `catalog-sync.ts` binary to git

`packages/web/src/features/models/catalog-sync.ts` L51

The composite key is written with a literal `0x00` byte at file offset 2287 rather than the `\0` escape. Runtime behavior is fine and collision-free — but git classifies the file as binary, so this 74-line module **that mutates the user's saved model table** shows zero reviewable lines in `gh pr diff`, `git log -p` prints nothing, `git grep` reports "Binary file matches", and a future merge conflict cannot be resolved textually.

Two other places in this repo build the same key with the escape and stay plain text: `models-page.tsx` L134 and `project-config-service.ts` L88.

**Fix:** one character — use `\0`. No behavior change, and the module becomes reviewable.

### 7. Announcement bar: no focus pause, focused link left inside `aria-hidden`

[`packages/landing/src/components/announcement-bar.tsx#L71`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/landing/src/components/announcement-bar.tsx#L71)

Only `onMouseEnter`/`onMouseLeave` pause rotation (L52-53). Slides carry `aria-hidden={i !== pos}` (L71) and `tabIndex={i === pos ? 0 : -1}` (L76); flipping `tabIndex` does not blur an already-focused element, so after the timer fires the focused link sits inside an `aria-hidden="true"` subtree — an explicit ARIA violation — and has been translated out of the `overflow-hidden` viewport, so the focus ring is invisible. It is also a WCAG 2.2.2 (Pause, Stop, Hide) failure: a keyboard-only user cannot stop the auto-update.

**Fix:** `onFocusCapture` / `onBlurCapture` on the wrapper that already carries the hover handlers.

Related, same file L62: under `prefers-reduced-motion` the carousel advances to the clone slide and then stalls permanently.

### 8. The new fidelity rules in the streaming translator have no test

[`packages/core/src/llm/generative-model.ts#L345-L355`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/core/src/llm/generative-model.ts#L345-L355)

Two brand-new rules ship here: a `fidelity.signature` **closes** the open text segment, and fidelity keys **merge** across deltas (`{ ...this.textFidelity, ...item.fidelity }`). Neither is exercised — confirmed by mutation: deleting the `signatureClosed &&` branch *and* downgrading the merge to plain assignment leaves all 368 core tests green.

What that hides:

- Stream `{text:"part one", fidelity:{signature:"sigA"}}` then `{text:"part two"}`. If the close-on-signature branch regresses, the two merge into one message carrying `sigA`, and on resume the provider is handed a signature covering text it never signed → rejected turn.
- With the merge downgraded, `{text:"",fidelity:{phase:"answer"}}` → `{text:"final"}` → `{text:"",fidelity:{signature:"s"}}` yields `{signature:"s"}` instead of `{phase:"answer",signature:"s"}` — GPT-5 phase segmentation silently lost on replay.

Given the e2e test is skipped, two unit tests here are cheap insurance.

---

## Minor

- [`model-catalog.ts#L515`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/core/src/state/model-catalog.ts#L515) — new Chinese `"8折/5折"` in a source comment; CLAUDE.md and the new `CONTRIBUTING.md` both make English mandatory here.
- [`penguin-cli/SKILL.md#L7`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/skills/skills/penguin-cli/SKILL.md#L7) — version bumped 1 → 2 but `updated` left at `2026-07-17`, so the card reads "v2 · updated 3 days ago".
- [`models-page.tsx#L333`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/web/src/features/models/models-page.tsx#L333) — speed-test badges are never cleared on Project switch, so Project A's timings render on Project B's identical `(provider, model_id)` cards.
- [`work-group.tsx#L112`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/web/src/features/chat/work-group.tsx#L112) — group duration jumps *backward* when the group settles (live wall clock → settled span).
- [`md.tsx#L54`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/web/src/features/chat/md.tsx#L54) — the inline `components.pre` arrow gives every code block a fresh component identity per render; during streaming (~8 commits/s) that remounts already-closed code blocks, dropping selection and Copy state.
- [`penguin-sdk/SKILL.md#L141`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/skills/skills/penguin-sdk/SKILL.md#L141) — the RAG recipe's embedded agent installs the *entire* skill library, so every `POST /api/ask` carries 11 unrelated skill descriptions in its system prompt.
- [`packages/skills/README.md#L15`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/skills/README.md#L15) — documents a "Web Research" group that does not exist, plus six retired group names.
- [`README.md#L79`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/README.md#L79) — the model table claims Moonshot AI serves Kimi K3; the catalog ships only K2.6 / K2.5 under Moonshot.
- [`session-manager.test.ts#L563`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/server/test/session-manager.test.ts#L563) — the "sub-session text doesn't count" test does not actually assert it; removing the origin guard at `session-manager.ts` L733 leaves it green.
- [`sessions.ts#L368`](https://github.com/Prism-Shadow/penguin-harness/blob/d747265c00c96080e8254d94d52081ac918dc5ae/packages/server/src/http/routes/sessions.ts#L368) — the new `preview=1` CSP-sandbox response is undocumented in `server-api.*.md`, the changelog and the specs, despite being part of the untrusted-workspace-file threat model.

---

## Checked and cleared

Worth stating explicitly — these were the highest-risk hypotheses and all were **refuted** on inspection:

- **"Sync presets" cannot lose or misroute credentials.** The full round trip (`presetToRow → rowToEntry → updateModels → getModels → toRow`) is idempotent, merged rows emit no `apiKey`/`clearApiKey`, and the PUT full-table replace preserves stored keys via the `{ ...prev }` base. The only key-mutating path anyone could construct requires a *prior failed save* leaving a pending clear in state. Plaintext never leaves `getModels` unmasked.
- **`md.tsx` is XSS-safe** — react-markdown v10, no `rehype-raw`, default URL sanitization, Shiki-escaped `dangerouslySetInnerHTML`.
- **The 0.4.0 error reclassification is correct.** Both `ToolCallArgumentParseError` and `EmptyResponseError` are thrown inside `streamingResponse`, before `streamingResponseStateful` commits the turn to history — so `malformed` (bounded by `maxReconnects`) is the right outcome and nothing previously retryable now falls through to `failed`. No `payload.signature` / `payload.phase` readers survive anywhere in `packages/*/src`.
- **`dev-prebuild.mjs`'s lock is sound** — the stale-steal window and the shared-temp-dir symlink concern were both examined and did not hold up.

---

## Process notes

- **Design sync cannot be verified from this PR.** `design/` is a gitignored symlink to a sibling repo, which currently has `01-PRINCIPLES` / `05-ARCHITECTURE` / `06-PROTOTYPE` / `08-CHANGELOG` modified **and uncommitted**. CLAUDE.md makes the specs ground truth and requires them to move in lockstep, but a reviewer of #7 has no way to see or approve that half. Even against the local working copy, four gaps remain: the skill grouping and two-button card (`05-ARCHITECTURE.md` L271), the `{{PROVIDER}}` / `{{MODEL_ID}}` placeholders (L374), and the Models page's sync-presets + speed-test features (`06-PROTOTYPE.md` L101). Worth committing and pushing the design repo before merging.
- **The PR description has drifted.** It documents 17 sections but stops short of the branch head — skill markers in titles, Provider/Model ID in the prompt, skill root/fast-stop, admin seed rotation and the one-shot RAG flow are not in it, and the "Verification" block is stranded between sections 11 and 12. The changelog entries *do* cover them, so it is only the description that is stale.
- **Size.** 51 commits across ~17 unrelated areas is why the two blocking bugs are the kind that hide: both live at the *seam between* areas — the catalog invariant vs. CLI provider inference, and positional slash vs. an Escape handler written for the old model. Splitting model-catalog / AgentHub-0.4 / landing-site into separate PRs would have surfaced them.

---

**Recommendation:** fix #1 and #2 before merge — #1 sends a user's API key to an unrelated vendor's endpoint, #2 destroys unrecoverable user input on a single keystroke. Everything else can follow up.

rank-Yu (approved):

zzzbitz (approved):

PR #8, opened 2026-07-20
[fix(web)] localize copied task-stats line instead of hardcoded Chinese

zhangnju:
## Summary
  - `formatTaskStats` hardcoded its entire output in Chinese (`统计信息`, `输入 tokens`, `已缓存`, `输出 tokens`), and its only caller in
  `task-stats-line.tsx` passed no label — so after the repo switched to English as the working language (#5), English users still got Chinese text whenever they copied a reply's stats line.
  - The function now takes a `labels` object supplied by the active locale dictionary, keeping it a pure, locale-agnostic formatter.
  - Adds a `statCached` string to both the `zh` and `en` dictionaries; the existing `statsLabel` / `statInput` / `statOutput` keys are reused.

## Test plan
  - [x] `pnpm --filter @prismshadow/penguin-web typecheck`
  - [x] `pnpm --filter @prismshadow/penguin-web test` (288 passed, includes
  updated `task-stats.test.ts` with en + zh label cases)

### DIFF
--- a/packages/web/src/lib/omni/task-stats.ts
+++ b/packages/web/src/lib/omni/task-stats.ts
@@ -297,18 +297,23 @@ export interface TaskStatsLabels {
   cached: string;
   /** Output-tokens label, e.g. "Output tokens" / "输出 tokens". */
   output: string;
+  /** Opening wrapper before the cached amount — keeps per-locale typography (e.g. " (" / "（"). */
+  parenOpen: string;
+  /** Closing wrapper after the cached amount (e.g. ")" / "）"). */
+  parenClose: string;
 }
 
 /**
  * Stats row text (copy fallback / used when there's no body text): same basis as the stats row —
  * this round's usage "input tokens (including cached amount) · output tokens · output TPS", e.g.
- * `[Stats] Input tokens 4k (cached 3k) · Output tokens 1.2k · 42.5 tok/s`. Labels come from the
- * caller's active locale dictionary so the copied text matches the UI language.
+ * `[Stats] Input tokens 4k (cached 3k) · Output tokens 1.2k · 42.5 tok/s`. Labels (including the
+ * parenthesis wrappers) come from the caller's active locale dictionary, so the copied text keeps
+ * per-locale typography — ASCII `(…)` in English, fullwidth `（…）` in Chinese.
  */
 export function formatTaskStats(s: TaskStats, labels: TaskStatsLabels): string {
   const b = s.tokensByBucket;
   return (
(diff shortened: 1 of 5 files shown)


packages/web/src/lib/omni/task-stats.ts
```diff
@@ -287,17 +287,29 @@ export function endTask(t: TaskStatsTracker, elapsedMs: number): TaskStats | nul
...
+ * `[Stats] Input tokens 4k (cached 3k) · Output tokens 1.2k · 42.5 tok/s`. Labels come from the
+ * caller's active locale dictionary so the copied text matches the UI language.
  */
-export function formatTaskStats(s: TaskStats, label = "统计信息"): string {
+export function formatTaskStats(s: TaskStats, labels: TaskStatsLabels): string {
   const b = s.tokensByBucket;
   return (
-    `[${label}] 输入 tokens ${humanizeTokens(b.cacheRead + b.cacheWrite)}（已缓存 ${humanizeTokens(b.cacheRead)}）` +
-    ` · 输出 tokens ${humanizeTokens(b.output)}` +
+    `[${labels.stats}] ${labels.input} ${humanizeTokens(b.cacheRead + b.cacheWrite)} (${labels.cached} ${humanizeTokens(b.cacheRead)})` +
```
Youhai020616: Nit: unifying on ASCII ` (…)` means the zh copied line loses its fullwidth parentheses — `输入 tokens 4k（已缓存 3k）` becomes `输入 tokens 4k (已缓存 3k)`. A reasonable trade for a single template; if you'd rather keep zh typography, the parenthesized segment could also come from the dictionary (e.g. a wrap template per locale). Fine to land as-is.

Youhai020616 (approved):
Verified: `formatTaskStats` has exactly one non-test caller (the copy path in `task-stats-line.tsx`), and `Strings = typeof zh` plus the `en: Strings` annotation force both dictionaries to carry the new `statCached` key — so the required-`labels` signature is fully typecheck-enforced and can't miss a caller or a locale. Reusing the existing `statsLabel`/`statInput`/`statOutput` keys and keeping the formatter pure while the view layer supplies words is the right structure. One typography nit inline.

packages/web/src/lib/omni/task-stats.ts
zhangnju: Good catch — addressed in fa31300. The parenthesis wrappers now come from the locale dictionary (`statParenOpen`/`statParenClose`): Chinese keeps fullwidth `（…）` and English keeps ASCII ` (…)`, while `formatTaskStats` stays a single pure template. Updated the zh test to assert the fullwidth output. Thanks for the review!

PR #9, opened 2026-07-20
test(core): cover CappedTextBuffer and ToolCallIdAllocator

zhangnju:
## Summary
  - Adds unit tests for two pure, previously untested core modules.
  - `CappedTextBuffer`: capacity drop keeps the tail, the omitted tally accumulates across appends, `drain()` prefixes the drop marker and clears the buffer, and `isEmpty` tracks both text and pending-dropped state (incl. a single append larger than the cap and a zero-cap buffer).
  - `ToolCallIdAllocator` / `stripToolCallIdSuffix`: unique ids pass through unchanged, collisions get a `#n` suffix from 2, `markUsed` seeds the taken set, and suffix stripping is shape-based and idempotent.

 ## Test plan
  - [x] `pnpm --filter @prismshadow/penguin-core typecheck`
  - [x] `pnpm --filter @prismshadow/penguin-core test` (382 passed / 2 skipped;
  18 new tests across the two files)

### DIFF
--- a/packages/core/test/capped-buffer.test.ts
+++ b/packages/core/test/capped-buffer.test.ts
@@ -1,0 +1,62 @@
+/**
+ * CappedTextBuffer unit tests: capacity drop keeps the tail, the omitted tally accumulates,
+ * drain() prefixes the drop marker and clears the buffer, and isEmpty tracks both text and tally.
+ */
+import { describe, expect, it } from "vitest";
+import { CappedTextBuffer } from "../src/environment/tools/background/index.js";
+
+describe("CappedTextBuffer", () => {
+  it("starts empty and drains to an empty string", () => {
+    const buf = new CappedTextBuffer(10, "earlier output");
+    expect(buf.isEmpty).toBe(true);
+    expect(buf.drain()).toBe("");
+  });
+
+  it("keeps content under capacity verbatim", () => {
+    const buf = new CappedTextBuffer(10, "earlier output");
+    buf.append("hello");
+    expect(buf.isEmpty).toBe(false);
+    expect(buf.drain()).toBe("hello");
+    expect(buf.isEmpty).toBe(true);
+  });
+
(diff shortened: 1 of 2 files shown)


packages/core/test/capped-buffer.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    buf.append("xyz"); // drop 1, keep "yz"
+    buf.drain();
+    expect(buf.isEmpty).toBe(true);
+    expect(buf.drain()).toBe("");
+  });
+
+  it("uses the drop label in the marker", () => {
+    const buf = new CappedTextBuffer(1, "earlier subagent output");
+    buf.append("ab"); // drop 1
+    expect(buf.drain()).toBe("[... 1 chars of earlier subagent output dropped ...]\nb");
```
Youhai020616, line 53: Observation, not a request: this locks in the `1 chars` grammar from the source template. Documenting current behavior is the right call for a test-only PR — just noting that a future pluralization fix in `capped-buffer.ts` will need to update this expectation too.

Youhai020616 (approved):
Traced all 18 assertions against the `capped-buffer.ts` / `tool-call-ids.ts` implementations — every expectation matches, including the edge cases (cap 0, single over-cap append, tally accumulation across appends, `markUsed` seeding, shape-based suffix strip). Import style, header comments, and file naming follow the existing core test conventions. Nice pick of targets: both modules are pure, shared by the command/subagent sessions, and previously uncovered. One observation inline.

packages/core/test/capped-buffer.test.ts
zhangnju: Agreed — this PR intentionally documents current behavior, so the test pins the `N chars` template as-is (including `1 chars`). Good flag: when `capped-buffer.ts` gets a pluralization fix, this expectation should move with it in that same source PR. Kept out of scope here to keep this one test-only. Thanks for the thorough trace of all 18 assertions!

PR #10, opened 2026-07-20
fix(server): validate positiveIntParam and optionalDateParam inputs

zhangnju:
## Summary
  - `positiveIntParam` used `Number.parseInt(..., 10)`, which returns `12` for
  `"12abc"` — a malformed path parameter (e.g. a Trace file index) silently passed validation. It now requires a digits-only string (`^\d+$`) before parsing.
  - `optionalDateParam`'s `YYYY-MM-DD` regex accepted impossible dates like `2026-13-40` or `2026-02-30`, letting malformed values flow into the usage query. It now round-trips the parsed components through UTC to confirm a real calendar day before accepting.

 ## Test plan
  - [x] `pnpm --filter @prismshadow/penguin-server typecheck`
  - [x] `pnpm --filter @prismshadow/penguin-server test` (270 passed, incl. new
  `validate.test.ts`: trailing garbage / signs / zero for ints; leap day,
  malformed shapes, and impossible dates for dates)

### DIFF
--- a/packages/server/src/http/validate.ts
+++ b/packages/server/src/http/validate.ts
@@ -166,8 +166,20 @@ export function optionalNumber(
 /** Validate a yyyy-mm-dd query parameter (defaults to undefined). */
 export function optionalDateParam(value: string | undefined, label: string): string | undefined {
   if (value === undefined || value === "") return undefined;
-  if (!/^\d{4}-\d{2}-\d{2}$/.test(value)) {
-    throw badRequest(`${label} must be in YYYY-MM-DD format.`);
+  const m = /^(\d{4})-(\d{2})-(\d{2})$/.exec(value);
+  // The shape check alone accepts impossible dates (2026-13-40, 2026-02-30); verify it's a real
+  // calendar day by round-tripping through UTC (which never rolls over into an adjacent month).
+  if (m) {
+    const [, y, mo, d] = m;
+    const dt = new Date(`${value}T00:00:00Z`);
+    if (
+      !Number.isNaN(dt.getTime()) &&
+      dt.getUTCFullYear() === Number(y) &&
+      dt.getUTCMonth() + 1 === Number(mo) &&
+      dt.getUTCDate() === Number(d)
+    ) {
+      return value;
+    }
   }
-  return value;
+  throw badRequest(`${label} must be a valid date in YYYY-MM-DD format.`);
 }
--- a/packages/server/src/http/validate.ts
+++ b/packages/server/src/http/validate.ts
@@ -42,7 +42,10 @@ export function requireValidId(c: Context, name: string): string {
 
 /** Parse a positive-integer path parameter (e.g. Trace file index). */
 export function positiveIntParam(c: Context, name: string): number {
-  const v = Number.parseInt(pathParam(c, name), 10);
+  // Match digits only: Number.parseInt would accept trailing garbage ("12abc" -> 12).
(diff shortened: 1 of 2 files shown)


packages/server/test/validate.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  it("parses a plain positive integer", () => {
+    expect(positiveIntParam(ctxWithParam("idx", "12"), "idx")).toBe(12);
+  });
+
+  it("rejects trailing garbage (parseInt would accept it)", () => {
+    expect(() => positiveIntParam(ctxWithParam("idx", "12abc"), "idx")).toThrow(HttpError);
+  });
+
+  it("rejects a leading sign, whitespace, and non-digits", () => {
+    for (const bad of ["+1", " 1", "1 ", "1.5", "0x10", ""]) {
```
Youhai020616: Heads-up: the `""` case doesn't exercise the new digits-only guard — `pathParam` throws a 404 `HttpError` for empty values before the validation runs, so this assertion passes via a different path than the rest of the loop. Consider dropping `""` here or asserting the 404 separately so the test measures what its name says.

packages/server/src/http/validate.ts
```diff
@@ -42,7 +42,10 @@ export function requireValidId(c: Context, name: string): string {
 
 /** Parse a positive-integer path parameter (e.g. Trace file index). */
 export function positiveIntParam(c: Context, name: string): number {
-  const v = Number.parseInt(pathParam(c, name), 10);
+  // Match digits only: Number.parseInt would accept trailing garbage ("12abc" -> 12).
+  const raw = pathParam(c, name);
+  if (!/^\d+$/.test(raw)) throw badRequest(`${name} must be a positive integer.`);
+  const v = Number.parseInt(raw, 10);
   if (!Number.isInteger(v) || v < 1) throw badRequest(`${name} must be a positive integer.`);
```
Youhai020616: With the `^\d+$` guard above, `!Number.isInteger(v)` can no longer fire (digits-only input always parses to an integer). Optional hardening while you're here: `Number.isSafeInteger(v)` would also reject overlong indices like `"99999999999999999999"` (currently accepted as 1e20).

Youhai020616 (approved):
Reviewed against all four `positiveIntParam` call sites (trace file index, session routes) and the `optionalDateParam` caller in `usage.ts`; also confirmed nothing depends on the old error message. Both fixes are real and correctly implemented — the digits-only guard closes the `parseInt("12abc") -> 12` hole, and the UTC round-trip (plus component comparison) rejects impossible calendar dates regardless of engine rollover behavior. Two non-blocking notes inline.

packages/server/test/validate.test.ts
zhangnju: Fixed in 831bfbe. Dropped `""` from the digits-only loop and added a dedicated test asserting that an empty/missing path param is rejected upstream by `pathParam` (404), so each test now measures the path its name claims.

packages/server/src/http/validate.ts
zhangnju: Good call — done in 831bfbe. Switched to `Number.isSafeInteger`, which also rejects overlong indices like `"99999999999999999999"` (previously parsed to 1e20 and accepted). Added a test for that case. You are right that with the `^\d+$` guard the old `isInteger` branch was effectively dead; `isSafeInteger` now gives it a real job. Thanks!

PR #22, opened 2026-07-22
feat(web): chat model picker lists key-configured models first, with a show-all expander

hiyouga:
## What

The draft chat's model-selection dropdown previously listed every model in the library, configured or not, so the models actually usable were buried among key-less entries. It now defaults to only the models with an API key configured, with a muted row at the very bottom — "Show models without a key (N)" — that expands the remaining ones in place (the menu stays open, the selection is untouched). Each open of the dropdown starts collapsed again. Key-less rows that do appear (after expanding, or via the exceptions below) carry a subtle "no key" tag on the right, styled like the existing "default" tag.

## Configured-key semantics

"Configured" means a **stored (masked) key**: `credential.apiKeyMasked` on `ModelInfo`. This matches the two existing consumers of that judgment — the chat credential guide (`chat-page.tsx`) and the Models page key status (`S.models.keyConfigured` / `S.models.noKey`). `envKey` deliberately does **not** count: it is merely the NAME of a fallback environment variable, and its presence says nothing about whether that variable is actually set.

## Edge cases

- The currently selected model and the project default model are always visible even without a key, so the active choice can never become invisible.
- If NO model has a configured key, the filter degrades to show-all (everything is listed) instead of presenting an empty dropdown.
- The search query filters whatever is visible; when a query only matches hidden key-less models, the collapsed list shows "no matching models" plus the expander row, so the match is one click away.

## Implementation

Pure logic lives in `packages/web/src/features/models/model-grouping.ts` (`hasConfiguredKey`, `visibleChatModels`) and is unit-tested in `packages/web/test/model-grouping.test.ts` (configured filtering, library ordering, selected/default exceptions, zero-configured fallback, query interaction, showAll). `ModelSelect` in `chat-input.tsx` gains local `showAll` state and the bottom expander row; the dropdown is otherwise unchanged. The new string is added to both dictionaries (`strings.ts` zh, `strings-en.ts` en).

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (core / server / cli / web)
- `pnpm test` — pass (web: 318 tests incl. 16 in model-grouping.test.ts; cli: 119)

### DIFF
--- a/packages/web/src/features/chat/chat-input.tsx
+++ b/packages/web/src/features/chat/chat-input.tsx
@@ -208,6 +208,13 @@ function modelLabel(m: ModelInfo): string {
 }
 
 /**
+ * "No key" marker for the model dropdown's key-less rows: a key struck through by a prohibition
+ * slash (24x24 line art, grayscale via currentColor, matching the approval-mode icon style).
+ */
+const NO_KEY_ICON =
+  "M21 2l-2 2m-7.61 7.61a5.5 5.5 0 1 1-7.778 7.778 5.5 5.5 0 0 1 7.777-7.777zm0 0L15.5 7.5m0 0l3 3L22 7l-3-3m-3.5 3.5L19 4M2 2l20 20";
+
+/**
  * Model selector (draft state only; docked to the left of the send button): both the button and
  * candidate items show the provider logo. The menu opens **downward** — the draft card is
  * vertically centered with room below; a top quick-search box (reusing the model page's rule:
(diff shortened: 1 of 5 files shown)


hiyouga:
未配置key用图标展示，一个钥匙上面禁止

hiyouga:
已改（5375689）：下拉里未配置 key 的模型行不再显示文字标签，改为图标——一把钥匙加一道禁止斜线（沿用 GlyphIcon 的 24x24 线稿风格，灰度随行文字颜色），"未配置 key" 文案保留在图标的 title / aria-label 里供悬停与读屏。底部"显示未配置 key 的模型（N 个）"展开行仍为文字。

范围说明：本次只改了此 PR 新增的对话下拉标记；模型仓库页卡片上的 key 状态仍是原有文字展示，如需一并换成该图标可以再提个后续改动。

rank-Yu (approved):

PR #24, opened 2026-07-22
feat(core): reserved-port and API-key guardrails in the default system prompt

hiyouga:
## Server port: one constant, and a prompt rule that keeps agents off our ports

Agents kept killing PenguinHarness's own services when they wanted a port — the default server port was hardcoded in two places, and the default prompt had just a generic "never kill a process you did not start" rule.

`packages/core/src/internal/ports.ts` now carries `DEFAULT_SERVER_PORT = 7364`, the single source of truth for the main server / Web UI default (`penguin server` / `penguin web`). Per review it lives under `internal/`; the core barrel keeps one documented named re-export (`export { DEFAULT_SERVER_PORT } from "./internal/ports.js";`) because the CLI's `DEFAULT_PORT` (`packages/cli/src/commands/serve.ts`) and the server's `PORT` fallback (`packages/server/src/config.ts`) both derive from it via the barrel — so the numbers cannot drift. It is a fallback only: the `--port` flag and the `PORT` env var still override it at runtime. The web / landing / docs vite dev ports (7365–7367) stay in their own configs: vite configs cannot import core TS, and per review the central `RESERVED_PORTS` list was dropped — it lost its only consumer once the prompt stopped enumerating ports.

Per review, the prompt does not list port numbers (ports can be overridden at runtime, so a baked-in list could go stale); the `# Constraints` bullet covers both killing listeners and squatting the ports:

> - Never kill a process you did not start yourself unless the user explicitly asks you to, including PenguinHarness's own services. Never take a PenguinHarness service port for your own servers; when a port you want is busy, pick another free port instead of killing the listener.

## API auth/key failures: retry once, then stop and ask

When a tool call hits an authentication/authorization or API-key error, retrying past one attempt is pointless — the fix is a new key, which the agent cannot supply itself, and vault updates only reach the shell environment of the next conversation. New `# Constraints` bullet:

> - On an API authentication/authorization or API-key error (401/403, invalid or missing key), retry at most once.

and new `# Stop rules` bullet (per review: no CLI commands — the user updates the key in the agent's vault / model settings, and the secret value must never enter the chat):

> - If an API auth/key error persists after that one retry, stop calling tools and ask the user to update the key in the agent's vault or the model settings outside the chat — the secret value must never be pasted into the conversation. Updated secrets only take effect in the next conversation, so further retries cannot succeed.

This is prompt-level only: it governs the agent's own tool calls (e.g. `curl` against some third-party API mid-task). The engine-side classification in `generative-model.ts` already treats 401/403 on LLM requests as non-retryable and is unchanged.

## Seed-only caveat

`DEFAULT_SYSTEM_PROMPT` seeds each agent's editable `system_config.yaml` at initialization. `loadOrInitAgentState` only writes defaults when `system_config.yaml` does not exist, and `provisionProjectAgents` never overwrites an existing agent — there is no re-seed on version bumps. Existing agents therefore keep their current prompt; the new rules reach newly created agents. No migration is attempted here.

## Verification

`pnpm -r build`, `pnpm typecheck`, and `pnpm format:check` clean. `pnpm test` — 1129 passing (+2 skipped) across 7 packages (core 373, server 285, web 312, cli 120, skills 16, landing 12, docs 11), including a core test asserting the default prompt carries both guardrails (and no port numbers or CLI commands) and a CLI test pinning `DEFAULT_PORT === DEFAULT_SERVER_PORT`.

### DIFF
--- a/packages/core/src/ports.ts
+++ b/packages/core/src/ports.ts
@@ -1,0 +1,21 @@
+/**
+ * Reserved PenguinHarness service ports — the single source of truth for the ports our
+ * own services listen on.
+ *
+ * The CLI (`penguin server` / `penguin web`) and the server derive their default port
+ * from here. The vite dev configs (packages/web|landing|docs) cannot import core TS, so
+ * they hardcode the same numbers with a comment pointing back at this file. The full
+ * list is interpolated into the default system prompt so agents never kill processes
+ * listening on these ports.
+ */
+
+/** Default main server / Web UI port (`penguin server` / `penguin web`; deliberately avoids common defaults like 3000/8080). */
+export const DEFAULT_SERVER_PORT = 7364;
+
+/** Ports reserved for PenguinHarness's own services; never kill their listeners to free a port. */
+export const RESERVED_PORTS: readonly number[] = [
+  DEFAULT_SERVER_PORT, // main server / Web UI
+  7365, // web dev server (packages/web/vite.config.ts; proxies /api to the main server)
+  7366, // landing dev server (packages/landing/vite.config.ts)
+  7367, // docs dev server (packages/docs/vite.config.ts)
+];
--- a/packages/core/src/ports.ts
+++ b/packages/core/src/ports.ts
@@ -1,21 +1,9 @@
 /**
- * Reserved PenguinHarness service ports — the single source of truth for the default
- * ports our own services listen on.
- *
- * The CLI (`penguin server` / `penguin web`) and the server derive their default port
- * from here; these are fallbacks only — the `--port` flag and the PORT environment
- * variable still override them at runtime. The vite dev configs (packages/web|landing|docs)
- * cannot import core TS, so they hardcode the same numbers with a comment pointing back
- * at this file.
+ * Default PenguinHarness server port — the single source of truth shared by the CLI
+ * (`penguin server` / `penguin web`) and the server, which previously each hardcoded
+ * the number. It is a fallback only: the `--port` flag and the PORT environment
+ * variable override it at runtime.
  */
 
-/** Default main server / Web UI port (`penguin server` / `penguin web`; deliberately avoids common defaults like 3000/8080). */
+/** Default main server / Web UI port; deliberately avoids common defaults like 3000/8080. */
 export const DEFAULT_SERVER_PORT = 7364;
-
-/** Default ports of PenguinHarness's own services; never kill their listeners or reuse the ports for other servers. */
-export const RESERVED_PORTS: readonly number[] = [
-  DEFAULT_SERVER_PORT, // main server / Web UI
-  7365, // web dev server (packages/web/vite.config.ts; proxies /api to the main server)
-  7366, // landing dev server (packages/landing/vite.config.ts)
-  7367, // docs dev server (packages/docs/vite.config.ts)
-];
--- a/packages/core/src/ports.ts
+++ b/packages/core/src/ports.ts
@@ -1,9 +1,0 @@
-/**
- * Default PenguinHarness server port — the single source of truth shared by the CLI
- * (`penguin server` / `penguin web`) and the server, which previously each hardcoded
- * the number. It is a fallback only: the `--port` flag and the PORT environment
- * variable override it at runtime.
- */
-
-/** Default main server / Web UI port; deliberately avoids common defaults like 3000/8080. */
-export const DEFAULT_SERVER_PORT = 7364;
--- a/packages/core/src/ports.ts
+++ b/packages/core/src/ports.ts
@@ -1,18 +1,18 @@
 /**
- * Reserved PenguinHarness service ports — the single source of truth for the ports our
- * own services listen on.
+ * Reserved PenguinHarness service ports — the single source of truth for the default
+ * ports our own services listen on.
  *
  * The CLI (`penguin server` / `penguin web`) and the server derive their default port
- * from here. The vite dev configs (packages/web|landing|docs) cannot import core TS, so
- * they hardcode the same numbers with a comment pointing back at this file. The full
- * list is interpolated into the default system prompt so agents never kill processes
- * listening on these ports.
+ * from here; these are fallbacks only — the `--port` flag and the PORT environment
+ * variable still override them at runtime. The vite dev configs (packages/web|landing|docs)
+ * cannot import core TS, so they hardcode the same numbers with a comment pointing back
+ * at this file.
  */
 
 /** Default main server / Web UI port (`penguin server` / `penguin web`; deliberately avoids common defaults like 3000/8080). */
 export const DEFAULT_SERVER_PORT = 7364;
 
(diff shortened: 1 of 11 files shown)


packages/core/src/ports.ts
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * The CLI (`penguin server` / `penguin web`) and the server derive their default port
+ * from here. The vite dev configs (packages/web|landing|docs) cannot import core TS, so
+ * they hardcode the same numbers with a comment pointing back at this file. The full
+ * list is interpolated into the default system prompt so agents never kill processes
+ * listening on these ports.
+ */
+
+/** Default main server / Web UI port (`penguin server` / `penguin web`; deliberately avoids common defaults like 3000/8080). */
+export const DEFAULT_SERVER_PORT = 7364;
```
hiyouga: 可以支持环境变量吗？

packages/core/src/state/default-config.ts
```diff
@@ -92,13 +93,15 @@ Communicate with the user precisely and concisely, yet with warmth. Do not repea
...
+- Never kill a process you did not start yourself unless the user explicitly asks you to, and never kill a process listening on the reserved PenguinHarness service ports (${RESERVED_PORTS.join(", ")}). When a port you want is busy, pick another free port; never free it by killing the listener.
 - If a tool call fails, read the error, adjust, and retry; never repeat the same failing input.
+- On an API authentication/authorization or API-key error (401/403, invalid or missing key), retry at most once.
 
 # Stop rules
 - Stop and give the final answer once the success criteria are met.
 - If the request is ambiguous, stop and ask the user for clarification instead of guessing their intent.
 - If you hit an error you cannot resolve, stop and report the blocker to the user.
+- If an API auth/key error persists after that one retry, stop calling tools and ask the user to supply or update the key (per-agent secrets: \`penguin config vault set\`; model credentials: \`penguin config model add\`). Updated secrets only take effect in the next conversation, so further retries cannot succeed.
```
hiyouga: 这个是让用户去Agent设置的密钥管理柜里面配置，不要提示Agent自己去用命令设置，去掉这些CLI指令吧

packages/core/src/state/default-config.ts
```diff
@@ -92,13 +93,15 @@ Communicate with the user precisely and concisely, yet with warmth. Do not repea
 # Constraints
 - Make the smallest change that satisfies the request; do not modify unrelated files.
 - Destructive operations are forbidden.
-- Never kill a process you did not start yourself (e.g. to free a busy port) unless the user explicitly asks you to.
+- Never kill a process you did not start yourself unless the user explicitly asks you to, and never kill a process listening on the reserved PenguinHarness service ports (${RESERVED_PORTS.join(", ")}). When a port you want is busy, pick another free port; never free it by killing the listener.
```
hiyouga: 确实 多建议Agent换别的端口开。

hiyouga: 这里不要列出来端口吧，不要硬编码了。也很乱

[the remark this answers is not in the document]
hiyouga: Fixed in d2e178f — the bullet now also forbids binding these ports, not just killing existing listeners: "Never take a PenguinHarness service port for your own servers; when a port you want is busy, pick another free port instead of killing the listener." The no-kill clause is now scoped to processes the agent did not start plus PenguinHarness's own services, so if the agent does accidentally start something on such a port, cleaning up its own process is still allowed.

packages/core/src/state/default-config.ts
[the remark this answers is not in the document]
hiyouga: Fixed in d2e178f — "supply" is gone and the update is directed outside the conversation: "ask the user to update the key in the agent's vault or the model settings outside the chat — the secret value must never be pasted into the conversation." This keeps it consistent with the existing guarantee that secret values never enter model context.

packages/core/src/ports.ts
hiyouga: 支持的：这个常量只是缺省回退值，运行时 `--port` 参数和 `PORT` 环境变量仍然优先（CLI 的 `resolvePort`：参数 > PORT > 缺省；server 的 `resolveServerConfig` 同样是 `env.PORT || DEFAULT_SERVER_PORT`）。已在 d2e178f 的注释里写明。另外按另一条意见，prompt 里已不再写死端口号，所以用户改了 PORT 之后 prompt 也不会失真。

packages/core/src/state/default-config.ts
hiyouga: 已在 d2e178f 去掉这些 CLI 指令。现在只引导用户在对话外到 Agent 的 vault（密钥管理）或模型设置里更新密钥，并明确密钥值绝不能贴进对话："ask the user to update the key in the agent's vault or the model settings outside the chat — the secret value must never be pasted into the conversation."

packages/core/src/state/default-config.ts
hiyouga: 是的，换端口是第一选择，d2e178f 里保留并加强了这一点：端口被占用时直接换空闲端口，不去查杀占用者；PenguinHarness 的服务端口也不允许 Agent 自己拿来起服务。

hiyouga: 已在 d2e178f 去掉端口列举，prompt 里不再出现具体端口号（端口本来就能被 PORT/--port 覆盖，写死列表还会失真），改为一句通用规则："Never take a PenguinHarness service port for your own servers; when a port you want is busy, pick another free port instead of killing the listener." RESERVED_PORTS 常量保留在 core 里，仅作为 CLI/server 缺省端口的单一来源，不再注入 prompt。

packages/core/src/ports.ts
hiyouga, line 1: 这个文件还有用吗？

hiyouga: 还有用，但已在 e2742e3 精简：prompt 不再注入端口列表后 RESERVED_PORTS 确实没有消费方了，已整个删掉。文件里现在只剩 DEFAULT_SERVER_PORT —— 它仍是 CLI（`penguin server` / `penguin web`）和 server 缺省端口的单一来源（原本两处各自硬编码 7364），运行时依旧可被 `--port` / `PORT` 覆盖。三个 vite dev 端口回归各自配置文件独立维护，注释也已同步更新。

hiyouga: 这个放internal吧

hiyouga: 已在 21b78af 移到 internal：packages/core/src/internal/ports.ts。因为 CLI 和 server 都是经 core barrel 引用这个缺省值，参照 session-title 的先例，barrel 里保留了唯一一条具名再导出 `export { DEFAULT_SERVER_PORT } from "./internal/ports.js";`（附一行注释说明只为 CLI/server 缺省端口而存在），其余不对外。

rank-Yu (approved):

PR #30, opened 2026-07-22
docs(changelog): 0.2.0 entries for the July 22 batch

hiyouga:
## Aggregated 0.2.0 changelog for the July 22 PR batch

A new `changelog/0.2.0/` (unreleased) folder summarizing the current PR batch, grouped by surface per `changelog/README.md` — not one file per PR:

| Entry | Covers PRs |
| --- | --- |
| `2026-07-22-llm-requests.md` | #21 (merged) |
| `2026-07-22-agent-guardrails.md` | #24 |
| `2026-07-22-model-settings.md` | #28 (incl. its in-flight follow-ups) |
| `2026-07-22-web-app.md` | #22, #26, #27, and the upcoming letter-avatar PR |
| `2026-07-22-skills.md` | #19 |
| `2026-07-22-sites-and-blog.md` | #20, #23, #25 |

The root `CHANGELOG.md` is deliberately untouched: per the conventions its release line is added during release preparation, and released folders stay frozen.

Best merged after (or alongside) the feature PRs it describes; entries describe the final intended shape of the in-flight work and can be amended here if review changes any of it.

## Verification

Markdown only — `prettier --check` clean on the new files. No code paths touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #33, opened 2026-07-22
fix(web): render the subagent expansion below the tool call's own content

hiyouga:
## Problem

On the chat page, a `run_subagent` tool card nests the child session's conversation (SubagentCard, expandable via the 子会话 header). It rendered **above** the tool call's own expanded content, so opening the tool card read as: header → child conversation → the tool's arguments/output. The product requirement is for the expanded subagent conversation to sit **below** the tool's own content.

## DOM order, before → after

Inside `ToolCallCard` (`packages/web/src/features/chat/tool-call-card.tsx`):

**Before:** header row → pending-approval block → **SubagentCard** → expanded details (`open &&`: arguments `<pre>`, output `<pre>`, images)

**After:** header row → pending-approval block → expanded details (arguments, output, images) → **SubagentCard**

Pure ordering fix — the JSX block moved verbatim (comment updated to document the position):

- Visibility semantics unchanged: the subagent card still renders regardless of the tool card's collapsed state, so the collapsed-card case (the default) looks exactly as before; only the expanded case changes.
- Expansion state, auto-expand-while-running / auto-collapse-on-finish, origin-chain context, and streaming behavior are untouched (all live inside `SubagentCard`, which only changed position among its siblings).
- The standalone `case "subagent"` path in `message-item.tsx` (orphan child sessions without a tool card) has no tool content to order against and is unaffected.

## Test

`subagent.spec.mjs` now also expands the `run_subagent` tool card after opening the child session and asserts via `compareDocumentPosition` that the 子会话 card **follows** the arguments `<pre>` in document order (this assertion fails on the old order), then collapses the card again before the reload section.

## Verification

- `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck`, `pnpm format:check`: all pass (node v24.18.0).
- `pnpm test`: all pass (web 351, core 396+2 skipped, server 297, cli 121, docs 11, skills 16, landing 26).
- E2E (`SKIP_BUILD=1 pnpm --filter @prismshadow/penguin-web test:e2e`): **14/14 passed (38.7s)**, including the updated `subagent.spec.mjs` with the new DOM-order assertion.

### DIFF
--- a/packages/web/e2e/subagent.spec.mjs
+++ b/packages/web/e2e/subagent.spec.mjs
@@ -99,8 +99,6 @@ test("subagent card survives a reload; child session title is generated from its
     await subagentBtn.elementHandle(),
   );
   expect(cardFollowsArgs, "subagent expansion renders below the tool arguments/output").toBe(true);
-  // Collapse the tool card again so the reload section below starts from the default state.
-  await toolRow.click();
 
   // --- Must still hold after reload (the fix's goal: parent Trace's child session pointer + server-side expansion) ---
   await page.reload();
--- a/packages/web/e2e/subagent.spec.mjs
+++ b/packages/web/e2e/subagent.spec.mjs
@@ -84,6 +84,24 @@ test("subagent card survives a reload; child session title is generated from its
   // Under live streaming, the child session card is nested inside the run_subagent tool card.
   await openSubagent();
 
+  // --- DOM order: the subagent expansion renders BELOW the tool's own content ---
+  // Expand the run_subagent tool card itself so its arguments block is in the DOM, then assert
+  // the child-session card comes after it in document order (it used to render between the tool
+  // header and the expanded arguments/output).
+  const toolRow = page.locator("button[aria-expanded]").filter({ hasText: "run_subagent" }).first();
+  await toolRow.click();
+  await expect(toolRow).toHaveAttribute("aria-expanded", "true");
+  const argsPre = page.locator("pre", { hasText: '"prompt"' }).first();
+  await expect(argsPre).toBeVisible();
+  const subagentBtn = page.getByRole("button", { name: /子会话/ }).first();
+  const cardFollowsArgs = await argsPre.evaluate(
+    (pre, card) => Boolean(pre.compareDocumentPosition(card) & Node.DOCUMENT_POSITION_FOLLOWING),
+    await subagentBtn.elementHandle(),
+  );
+  expect(cardFollowsArgs, "subagent expansion renders below the tool arguments/output").toBe(true);
+  // Collapse the tool card again so the reload section below starts from the default state.
+  await toolRow.click();
+
   // --- Must still hold after reload (the fix's goal: parent Trace's child session pointer + server-side expansion) ---
   await page.reload();
   await openSubagent();
--- a/packages/web/src/features/chat/tool-call-card.tsx
+++ b/packages/web/src/features/chat/tool-call-card.tsx
@@ -194,18 +194,6 @@ export function ToolCallCard({ item, ctx }: { item: ToolCallItem; ctx: StreamRen
         </div>
       )}
 
-      {/* Subagent card: always visible (not hidden after completion, only collapsed by default), unaffected by the tool card's collapsed state. */}
-      {item.subagent && (
-        <div className="px-3 pb-2">
-          <SubagentCard
-            sessionId={item.subagentSessionId ?? ""}
-            model={item.subagent}
-            running={!item.outputComplete}
-            ctx={ctx}
-          />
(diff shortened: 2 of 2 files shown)


rank-Yu (approved):

hiyouga (commented):
**Verdict: approve-with-nits** — the JSX move is correct and the new e2e assertion genuinely fails on the pre-fix order; the only real gaps are a missing changelog entry and two doc/polish nits.

### Findings

**[medium] `changelog/0.2.0/2026-07-22-web-app.md` — no changelog entry, and PR #30's Web App entry does not cover this.**
That entry's four `##` sections are "Sidebar groups by Workspace", "Model picker lists configured models first", "Letter avatars for custom model groups and Agents", "Chat message rendering" — none mentions the subagent card or its position. Merging leaves 0.2.0 with no record of the change.
*Fix:* once #30 lands (it creates `changelog/0.2.0/`), extend the existing Web App entry's "Chat message rendering" section with one sentence and refresh the matching README line. No new file — the surface entry already exists.

**[low] `design/specs/06-PROTOTYPE.md:51` — the spec sentence governing this exact behavior is not updated with the ordering rule the PR calls a product requirement.**
Line 51 pins the nesting and the default expand/collapse policy but says nothing about position relative to the tool's own arguments/output — so nothing in the specs contradicted the old order either. The requirement now exists only in a code comment, and the next reader has no way to know the order is load-bearing.
*Fix:* append a clause to `06-PROTOTYPE.md:51`, e.g. `；该嵌套块渲染在工具自身 arguments / 输出之下。` (`design/` is a symlink to `penguin-harness-design`, so this lands there, not here.)

**[low] `packages/web/src/features/chat/tool-call-card.tsx:2-4` — the file header contradicts the PR's own new inline comment.**
The header says clicking expands full arguments, output, "and any nested subagent", but `{item.subagent && …}` is a sibling of `{open && …}`, not inside it — the subagent card renders whether or not the card is open. Pre-existing inaccuracy, but this PR rewrites the adjacent comment at line 230 to state the opposite invariant, leaving two comments in one file disagreeing.
*Fix:* reword the header to "clicking expands full arguments and output; a nested subagent renders below them regardless of collapsed state".

**[low] `packages/web/src/features/chat/tool-call-card.tsx:232` — the wrapper `className="px-3 pb-2"` has no top spacing, which was fine under the header button but now abuts a filled block.**
With the card expanded and only arguments present (the common state while a subagent runs), the args `<pre>` carries `bg-gray-50 dark:bg-gray-950/60` and the dashed subagent card starts 0px from it, so it reads as glued to the tinted block's bottom edge.
*Fix:* `px-3 pb-2 pt-2`, or add `border-t` to match the sibling `<pre>` blocks.

### Notes

- The new assertion is load-bearing, verified end to end: pre-fix the subagent `<div>` was child index 2 and the `open &&` details div index 3, so `compareDocumentPosition` returns `DOCUMENT_POSITION_PRECEDING` (2), `& 4` is 0, and the assertion fails. The locators resolve unambiguously against the fixture — the run_subagent arguments block is the only `<pre>` containing `"prompt"`, and there is exactly one `子会话` button.
- No state or layout regression from the move. The card root's four children are a static positional list, so React keeps `SubagentCard`'s fiber identity by index across `open` toggles — its `open`/`userToggled`/`wasRunning` state and the auto-collapse-on-finish effect survive. `aria-expanded` lives only on the header button, untouched. `divide-y` applies to whole tool cards, not card internals.
- Minor: the `await toolRow.click()` at `subagent.spec.mjs:103` that "collapses the tool card again" is dead — the next line is `page.reload()`, which resets all React state anyway. Separately, `subagent-card.tsx:46` has no `aria-expanded` on its toggle (pre-existing, out of scope).

hiyouga:
Addressed the review nits in d081aaa:

1. **Header comment** — reworded as suggested: "clicking expands full arguments and output; a nested subagent renders below them regardless of collapsed state." It no longer contradicts the inline comment.
2. **Spacing** — went with `pt-2` on the moved wrapper (`px-3 pb-2 pt-2`) rather than a `border-t`: the subagent card is an inset dashed-bordered card, not a full-width section like the sibling `<pre>` blocks, so a separator line would still leave the card abutting the line, while symmetric `pt-2`/`pb-2` gives it a clean gap below the tinted args/output blocks when expanded and below the header row when collapsed.
3. **Dead click** — removed the no-op `toolRow` collapse click before `page.reload()` in `subagent.spec.mjs`.

Validation rerun: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (all packages green), e2e 14/14 passed (44.8s), `pnpm format:check` clean. No changelog entry per the coordination note (second wave after #30).

PR #34, opened 2026-07-22
feat(web): drop "none" from the offered thinking levels

hiyouga:
## Summary

Product feedback on the thinking-level pickers shipped in #28: 思考等级选项去掉「无」，好多模型都不支持 — many models cannot disable thinking, so the **user-facing pickers no longer offer the off tier**.

- **Chat draft picker** (`chat-input.tsx` + `thinking-level.ts`): the menu now lists 低/中/高/极高 (Low/Medium/High/Extreme High) from a new `SELECTABLE_THINKING_LEVELS` list; `THINKING_LEVELS` (all five) stays as the display-lookup table.
- **Agent settings** (`S.agent.thinkingLevelOptions` zh+en): the `none` row is removed; the `""` inherit row stays.

## Backward compatibility (deliberate)

`"none"` remains a **valid stored/wire value** — nothing narrows `ThinkingLevelName`, and the server's agent-config validation (`agent-config-service.ts` `THINKING_LEVELS`) still accepts it, so existing `system_config.yaml` files carrying `none` keep loading and saving. The pickers only stop **offering** it:

- the chat trigger and the session read-only tag still display a stored/legacy `none` via the kept label (无/None) in `S.chat.thinkingLevelNames`;
- the agent settings `OptionMenu` appends a display-only `none` row (with a "stored legacy tier" note) **only while the current value is `none`** — the real state shows instead of "—", it is never silently rewritten, and the row disappears once another level is picked;
- internal `"none"` uses are untouched: the connectivity/speed probes and the title/vision bare LLMs keep `thinkingLevel: "none"`.

Docs (`models.{en,zh}.md`) note that the Web pickers offer `low` and above only.

## Verification

Full chain on node v24 — `pnpm -r build` (7/7) → `pnpm typecheck` (7/7) → `pnpm test` → e2e → `pnpm format`/`format:check`, all green:

- Unit: core 396 passed (+2 skipped), server 297, web **352** (thinking-level tests updated: selectable list excludes `none`, all five stored levels stay displayable), cli 121, skills 16, docs 11, landing 26
- **E2E: 14 passed** — the draft spec now additionally asserts the open menu offers 低 and has **no** 无 row, alongside the existing pick-高 → agent-config write-through → session_meta assertions

### DIFF
--- a/packages/docs/content/models.en.md
+++ b/packages/docs/content/models.en.md
@@ -77,7 +77,7 @@ Some models in the preset catalog: deepseek-v4-pro / deepseek-v4-flash, gemini-3
 
 ## Thinking levels
 
-Five levels: `none | low | medium | high | xhigh`, configured per Agent as `model.thinking_level` in `system_config.yaml`, default medium. The chat draft view offers a quick picker next to the model selector: a picked level is written back to the selected Agent's setting immediately (the switched-to level becomes that Agent's new default and applies from the next session; a running session keeps the level it was created with). See [Configuration](/configuration).
+Five levels: `none | low | medium | high | xhigh`, configured per Agent as `model.thinking_level` in `system_config.yaml`, default medium. The Web pickers offer `low` and above only (many models cannot disable thinking; `none` stays a valid stored value and still displays). The chat draft view offers a quick picker next to the model selector: a picked level is written back to the selected Agent's setting immediately (the switched-to level becomes that Agent's new default and applies from the next session; a running session keeps the level it was created with). See [Configuration](/configuration).
 
 ## Models decoupled from Agents
 
--- a/packages/docs/content/models.zh.md
+++ b/packages/docs/content/models.zh.md
@@ -77,7 +77,7 @@ api_key = "sk-..."
 
 ## 思考等级
 
-思考等级共五档：`none | low | medium | high | xhigh`，按 Agent 在 `system_config.yaml` 的 `model.thinking_level` 配置，默认 medium。对话草稿页在模型选择器旁提供快捷拾取器：选定档位立即写回所选 Agent 的该项配置（切换后的档位即成为该 Agent 的新默认，自下一个 Session 生效；进行中的 Session 沿用创建时的档位）。见 [配置参考](/configuration)。
+思考等级共五档：`none | low | medium | high | xhigh`，按 Agent 在 `system_config.yaml` 的 `model.thinking_level` 配置，默认 medium。Web 拾取器只提供 `low` 及以上档位（多数模型不支持关闭思考；`none` 仍是合法的已存值，能正常显示）。对话草稿页在模型选择器旁提供快捷拾取器：选定档位立即写回所选 Agent 的该项配置（切换后的档位即成为该 Agent 的新默认，自下一个 Session 生效；进行中的 Session 沿用创建时的档位）。见 [配置参考](/configuration)。
 
 ## 模型与 Agent 解耦
 
--- a/packages/web/e2e/draft.spec.mjs
+++ b/packages/web/e2e/draft.spec.mjs
@@ -87,14 +87,16 @@ test("draft: pick model/approval -> reload restores them -> send creates the ses
   await expect(page.getByRole("button", { name: "审批模式" })).toContainText("放行只读");
 
   // Conversation-time thinking level (backed by the Agent settings): the picker shows the
-  // seeded default (medium, short name 中); the menu carries a title bar and exactly the five
-  // short-name rows (no descriptions, no default row); picking 高 writes straight through to
-  // the Agent config, so the session created on send runs with it and it becomes the Agent's
-  // new default.
+  // seeded default (medium, short name 中); the menu carries a title bar and the short-name
+  // rows 低/中/高/极高 only — no descriptions, no default row, and no 无 (many models cannot
+  // disable thinking); picking 高 writes straight through to the Agent config, so the session
+  // created on send runs with it and it becomes the Agent's new default.
   const thinkingBtn = page.getByRole("button", { name: "思考等级" });
   await expect(thinkingBtn).toContainText("中");
   await thinkingBtn.click();
   await expect(page.getByText("思考等级", { exact: true })).toBeVisible(); // menu title bar
+  await expect(page.getByRole("button", { name: "低", exact: true })).toBeVisible();
+  await expect(page.getByRole("button", { name: "无", exact: true })).toHaveCount(0);
   await page.getByRole("button", { name: "高", exact: true }).click();
   await expect(thinkingBtn).toContainText("高");
   await expect
--- a/packages/web/src/features/agents/agent-settings-page.tsx
+++ b/packages/web/src/features/agents/agent-settings-page.tsx
@@ -31,6 +31,7 @@ import { Modal } from "../../components/ui/modal";
 import { Skeleton } from "../../components/ui/skeleton";
 import { VaultTab } from "./vault-tab";
 import { SchedulesTab } from "./schedules-tab";
+import { thinkingLevelOptionsFor } from "../chat/thinking-level";
 
 type TabKey = "overview" | "prompt" | "runtime" | "tools" | "vault" | "schedules";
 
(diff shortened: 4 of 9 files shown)


rank-Yu (approved):

hiyouga (commented):
**Verdict: needs-changes** — the backward-compat engineering is correct and I found no functional regression, but the change lands with no changelog entry (and makes the pending 0.2.0 entry factually wrong) and introduces new Chinese code comments against the repo's English rule.

> **Merge conflict:** this branch conflicts with #39 on `packages/web/src/features/agents/agent-settings-page.tsx`. Whichever lands second needs a rebase.

### Findings

**[medium] No changelog entry — and the pending 0.2.0 entry it belongs to now states the opposite of the shipped behavior.**
The branch adds nothing under `changelog/` (only the frozen `0.1.0/` tree). Meanwhile `changelog/0.2.0/2026-07-22-model-settings.md` on #30's branch says the chat draft gains a picker "with the five short levels (None / Low / Medium / High / Extreme High)" — after this merges, 0.2.0 ships describing a None row the picker no longer offers.
*Fix:* extend that entry's "Thinking level moves to the conversation" section to say the pickers offer `low`..`xhigh` while `none` stays a valid stored value, and amend its `README.md` summary line. (Depends on #30 landing `changelog/0.2.0/`.)

**[medium] `packages/web/src/lib/strings.ts:191`, `:199`, `:468` — new Chinese code comments, which the language rule does not exempt.**
The carve-out for `packages/*/src/lib/strings.ts` covers the zh dictionary *values*, not the surrounding comments — CLAUDE.md spells this out for zh test fixtures ("the comments around them are still English"). On `main` this file has 43 comment lines of which only 2 are Chinese residue; this PR adds two more (`/** 可选档位不含 none…*/`, `/** 仅当已存配置为 none 时…*/`) and appends Chinese to the residue comment at `:468` instead of translating it while touching it.
*Fix:* write the three in English, e.g. "Selectable tiers exclude `none`; a stored `none` still displays — see `thinkingLevelNoneKept`."

**[low] `design/specs/06-PROTOTYPE.md:58`, `:76` — the UI behavior change lands with no spec update.**
Note `01-PRINCIPLES.md:141` enumerates `thinking_level: none | low | medium | high | xhigh` as the **system_config.yaml** value set, which this PR deliberately preserves — that is *not* a divergence. But `06-PROTOTYPE.md:58` (chat input toolbar) and `:76` (Agent settings 运行参数) describe these surfaces at comparable granularity and never mention the picker at all. Inherited debt — #28 added the picker without a spec line either.
*Fix:* add one clause to each noting the picker offers `low`+ only while `none` remains a valid stored value.

**[low] `packages/web/src/features/agents/agent-settings-page.tsx:507-508` — the legacy-`none` row is gated on local edit state, not the stored config, so an accidental pick cannot be undone in place.**
Agent with `model.thinking_level: none`; open Settings → Runtime, misclick `low` (nothing saved yet) → `thinkingLevel === "low"`, the appended `none` row disappears, and the user cannot restore the value they are actually storing without switching tabs (which remounts `RuntimeTab`) or reloading.
*Fix:* gate on the persisted value — `cfg.model?.thinkingLevel === "none"`.

**[low] `packages/web/test/thinking-level.test.ts` / `packages/web/e2e/draft.spec.mjs` — the only genuinely new backward-compat path has no test.**
`thinking-level.test.ts` re-pins `thinkingLevelLabel`, whose body is unchanged from `main` (it already covered `none`), plus the trivial contents of `SELECTABLE_THINKING_LEVELS`. The new conditional — append a display-only `none` row iff the current value is `none` — is exercised by nothing: the e2e seeds `medium`, so it only asserts 无 is absent, never that a stored `none` still renders on the trigger instead of `—`. A regression that drops the append would ship green.
*Fix:* extract the option assembly into a pure helper in `thinking-level.ts` (e.g. `thinkingLevelOptionsFor(current)`) and unit-test both branches.

### Notes

- Backward compatibility checks out end to end: `ThinkingLevelName` (`core/src/interfaces.ts:84`) still includes `"none"`; the server's write validation `THINKING_LEVELS` (`agent-config-service.ts:40`) still accepts it; the read path at `:119-120` passes it through; `thinkingLevelLabel` still resolves it; and the save guard (`agent-settings-page.tsx:470`) only emits `thinkingLevel` when it differs from `cfg`, so a stored `none` is never silently rewritten. The draft picker's value comes from the agent-config API, not localStorage, so no cached `none` can strand a draft.
- i18n shape is intact — `thinkingLevelNoneKept` exists in both dictionaries, and `Strings = typeof zh` would have failed typecheck otherwise. `models.en.md`/`models.zh.md` got equivalent sentences; `configuration.{en,zh}.md:95` still lists all five as valid config values, which remains correct.
- Pre-existing, not caused here: `agent-config-service.ts:119-120` casts an arbitrary YAML `thinking_level` to `ThinkingLevelName` without validating on read, so a hand-edited garbage value renders as `—` with no selected row. Also `""` (inherit) can never be re-selected once an override exists.

hiyouga:
评审意见已在最新提交处理：

- **中文代码注释**：strings.ts 中本分支新增/追加的三处注释已改回英文（中文例外仅适用于词典值本身，不适用于注释）。
- **legacy none 行的判定改为已存配置**：此前挂在本地编辑态上，误点其他档位会让 none 行在保存前消失、无法点回磁盘上实际存储的值；现改为 `cfg.model?.thinkingLevel === "none"`（已持久化的配置），保存前始终可回到已存值。
- **兼容分支补测试**：选项组装抽为纯函数 `thinkingLevelOptionsFor`（thinking-level.ts 导出），单测覆盖两条分支——已存 none 时追加仅展示行（含行序与文案），否则无 none 行且字典顺序与缺省档映射保持不变。

验证：build/typecheck 7/7，单测全绿（web 354，新增 2 例），e2e 14 通过，format 通过。

nicolaepocroianu:
I looked at the workspace with real content in it rather than trusting it, and found three things worth fixing. Pushed in ac605edf.

**The workspace was not actually filling its pane.** The detail route still wrapped it in `mx-auto max-w-4xl ... p-6` inside an `overflow-auto` div. So the full-height layout sat inset by about eighty pixels on each side, inside a second scroll container wrapped around a child that manages its own scrolling. That is exactly the shape `styles.css` warns about. The route is now a bare `h-full min-h-0`, and the back link and the two access notices moved into the workspace header and its notices strip where they belong.

**Upload, media library and reuse stacked full width**, which read as three unrelated blocks and pushed the generate controls below the fold. They are three ways of choosing one file, so they now sit in one row, as in the prototype you picked.

**Scene identifiers broke mid-word** in the rail (`scene-3-letter-d-lowercas / e`). They truncate with the full text in a tooltip now.

The editor fits without scrolling at 1440 by 900 with five scenes and fifteen assets loaded.

Re-verified after the changes: `activities.spec.mjs` 22 passed, typecheck and formatting clean, and the guardrail suites plus the workspace model tests at 57 passed.

The earlier visual audit I gave you was against the old stacked page. Most of those nine points were symptoms of the wrong container and are gone with it; these three are what replaced them.

nicolaepocroianu:
Correcting a verification claim of mine, and reporting what a fuller check found.

**What I claimed earlier was thinner than it sounded.** I said the document-scroll sweep passed with `/activities` added, but I had only run that one test in isolation with `--grep`. Running the whole file is a different thing, so I went back and did it, on this branch and on unmodified `main`.

| Run | Result |
| --- | --- |
| `layout.spec.mjs` on `main` | 6 failed, 0 passed |
| `layout.spec.mjs` on this branch | 5 failed, 1 passed |

The one that now passes is `layout: no page grows the document`, the test this branch repaired and added `/activities` to. The remaining five fail identically on `main` and are untouched by this branch:

- `layout: en draft + context gauge + mobile models`
- `models: group header actions collapse to icons instead of disappearing`
- `layout: collapsed rail — order, bilingual tooltips, last conversation`
- `layout: mobile chat dropdowns stay inside the viewport`
- `layout: login — blank start, non-crossing traces, lang/theme controls`

`layout.spec.mjs` is in the same state `activities.spec.mjs` was in before #29: it still drives the app through Chinese labels the app stopped shipping when the translations were removed, and at least one test name still promises "bilingual tooltips". I fixed only the two label uses that stood between me and the invariant this branch needs to satisfy. Repairing the rest is its own change, like #29 was, and I did not fold it in.

**One thing worth knowing about the environment.** My first attempt at this comparison was invalid and I nearly reported it. The test server died early with `EPERM: operation not permitted, rename ... .skill-porting.incoming`, a Windows file-lock during skill seeding, after which every remaining test failed with `ECONNREFUSED`. That produced a plausible-looking 6-failure list that had nothing to do with the code. Worth knowing if a run here ever looks inexplicably bad.

**Also re-verified on this branch after the last commit:** full web unit suite at 2509 passed with only the known `model-grouping` baseline failing, and `activities.spec.mjs` at 22 passed.

nicolaepocroianu:
Recording a known gap between the approved prototype and this branch, so it is not lost.

**The shell matches the prototype; the inside of the detail pane does not.** The artboards showed Loom's asset editor as current media against a replacement, side by side, each with its own metadata. This branch keeps the existing single-column arrangement inside the new pane: bound preview, then accepted preview, then the candidate lists below.

**Why it is not in this change.** A real side-by-side means moving the candidate lists into the comparison, and those lists carry the accept buttons, staleness notices and accessible names that most of the 22 browser tests target by role and label. That is a restructure of the pane the suite depends on most, and it deserves its own change with the shipped layout reviewed first.

**It also carries a scope question.** Loom's metadata cards show format, dimensions, file size and length. Format and size are honestly available for uploaded files, and dimensions for an image once it has been previewed. For media read from the WAF checkout none of it is available without a new endpoint. So the choice is to scope the metadata to uploads or to add server support, rather than render a table of dashes.

**Also still unbuilt from Loom, recorded earlier and unchanged:** audio trim, per-word pronunciation clips for book activities, the media library as a standalone browsing destination, and the diff minimap.

PR #35, opened 2026-07-22
feat(web): sidebar group polish — even hover geometry, folder open/close icons, group pinning

hiyouga:
## Hover-geometry fix (diagnosis)

The group header row is `flex items-center` whose height is set by its fixed `h-7` (28px) action buttons, while the collapse toggle on the left was content-sized: `py-0.5` plus its tallest child — the 18px avatar in agent mode (~22px pill) and only a 15px folder / 16px text line in workspace mode (~20px pill). `items-center` then floats that shorter pill next to the 28px button pills, so the workspace header's hover highlight visibly didn't line up with the chevron/"+" buttons beside it. Fix: the collapse toggle now carries `self-stretch` (its own `items-center` keeps the content centered), so every hover surface in a header row spans the same 28px height. Applied to both modes — the agent header had the same, milder mismatch (22px vs 28px), and "one consistent hover height" should hold sidebar-wide; visually this only makes its hover pill slightly taller.

## Folder open/close

The workspace group's folder icon now tracks the group state: the existing closed-folder glyph when collapsed, a new `FOLDER_OPEN_ICON` (lucide folder-open, same 24-grid line-art family and stroke settings as the other `Icon` glyphs) when expanded. The grouping-mode toggle in the section header keeps the closed glyph — there it denotes the mode, not a state.

## Group pinning (both modes)

- A pin toggle sits in each group header between the collapse toggle and "+": revealed on header hover / keyboard focus while unpinned, and permanently visible once pinned — the visible pin doubles as the subtle pinned indicator. `aria-pressed` + new strings in both dictionaries (`nav.pinGroup`/`nav.unpinGroup`: 置顶分组/取消置顶, Pin group/Unpin group). The header row got a named `group/header` hover scope so the reveal can't interact with the session rows' plain `group` scope.
- Ordering: a new pure `pinnedFirst(items, keyOf, pinned)` in `session-grouping.ts` — a stable partition putting pinned keys first while each partition keeps its existing order (recency for workspace groups, the configured order for agents). Applied via `orderedAgents` / `orderedWorkspaceGroups` memos; pinning the merged temp group lifts it above named groups too.
- Persistence: `penguin.sidebarPinnedGroups.<projectId>` in localStorage, exactly mirroring the collapse-state design — per Project, one key set shared across both grouping modes (agent ids and workspace paths cannot collide), storage-failure tolerant, corrupted data degrades to "nothing pinned". The collapse load/save pair was generalized into `loadGroupSet`/`saveGroupSet` so both concerns share one implementation.

Kept the diff surgical around `renderGroupBody`/`renderRows` (untouched) to minimize overlap with the in-flight automated-sessions-subgroup branch.

## Verification

- `pnpm -r build` (7/7), `pnpm typecheck` (7/7): clean.
- `pnpm test`: all green — web 29 files (session-grouping now 13 tests incl. 4 new `pinnedFirst` cases), core 27+1, server 33, cli 13, skills, docs, landing.
- Playwright e2e ran locally (`SKIP_BUILD=1 pnpm --filter @prismshadow/penguin-web test:e2e`): **14 passed (38.1s)** — draft.spec now asserts pin → geometric reorder (bounding-box y, the layout.spec technique) → localStorage write → order and indicator surviving a reload.
- `pnpm format` + `pnpm format:check`: clean.

### DIFF
--- a/packages/web/e2e/draft.spec.mjs
+++ b/packages/web/e2e/draft.spec.mjs
@@ -223,11 +223,27 @@ test("draft: pick model/approval -> reload restores them -> send creates the ses
     )
     .toContain("temp-workspaces");
   await expect.poll(async () => (await yOf(tempLabel)) < (await yOf(namedLabel))).toBe(true);
-  // Pinned state and order survive a reload.
+  // Pinned state and order survive a reload. The pin button's accessible name is STATIC
+  // ("置顶分组"); the state lives in aria-pressed alone (review: a name swapping to 取消置顶
+  // alongside aria-pressed announces the state twice in conflicting ways).
   await page.reload();
   await expect(tempLabel).toBeVisible();
   await expect.poll(async () => (await yOf(tempLabel)) < (await yOf(namedLabel))).toBe(true);
-  await expect(tempHeader.getByRole("button", { name: "取消置顶" })).toBeVisible();
+  const tempPin = tempHeader.getByRole("button", { name: "置顶分组" });
+  await expect(tempPin).toHaveAttribute("aria-pressed", "true");
+  // The pinned pin doubles as the always-visible indicator. toBeVisible() ignores opacity,
+  // so assert computed opacity directly: pinned = 1 with the mouse elsewhere; an unpinned
+  // header's pin stays hover-gated at 0. Park the mouse first — after the reorder the named
+  // header sits where the temp header was clicked, which would otherwise hover-reveal it.
+  await page.mouse.move(640, 500);
+  await expect.poll(() => tempPin.evaluate((el) => getComputedStyle(el).opacity)).toBe("1");
+  await expect
+    .poll(() =>
+      wsHeader
+        .getByRole("button", { name: "置顶分组" })
+        .evaluate((el) => getComputedStyle(el).opacity),
+    )
+    .toBe("0");
 
   // —— Switch the sidebar to agent mode via the section-header toggle (persists in localStorage) ——
   await page.getByRole("button", { name: "按 Agent 分组" }).click();
--- a/packages/web/e2e/draft.spec.mjs
+++ b/packages/web/e2e/draft.spec.mjs
@@ -8,6 +8,8 @@
  *   "临时工作区" group, a named directory groups under its basename, and that group header's
  *   "+" pre-fills the draft's Workspace (via router state, applied once per navigation — a
  *   manual change made afterwards survives a reload instead of being re-overridden);
+ * - the group header's hover pin toggle lifts a group above the unpinned ones in its mode and
+ *   persists per Project (order re-checked after a reload);
  * - after switching the sidebar to agent mode (toggle persisted in localStorage), the agent
  *   group header's "+" creates a draft scoped to that group's Agent (explicitly set via router
  *   state, overriding the cache).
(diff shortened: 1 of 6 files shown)


rank-Yu (approved):

hiyouga (commented):
**Verdict: needs-changes** — the implementation is genuinely sound (pinning is pure, stable and non-mutating; i18n complete), but the PR ships a new user-visible, persisted feature with no `changelog/0.2.0/` entry and no `design/specs/` update.

> **Merge conflict:** this branch conflicts with #40 on `packages/web/src/components/layout/sidebar.tsx` and `packages/web/test/session-grouping.test.ts`. The conflict is **textual only** — `pinnedFirst` reorders whole groups while #40's `partitionSessions` splits sessions inside a group; they compose cleanly (pin groups, then partition each group's rows).

### Findings

**[medium] `changelog/0.2.0/2026-07-22-web-app.md` — no changelog entry.**
The branch touches nothing under `changelog/`, and #30's Web App entry ends at "Group collapse state persists per Project across reloads. Agent mode is unchanged." — it never mentions pinning, the folder open/close icons, or the header hover geometry. 0.2.0 would ship claiming the sidebar only gained workspace grouping, while a user finds a pin icon and a new `penguin.sidebarPinnedGroups.<projectId>` localStorage key with no record of where they came from.
*Fix:* extend that entry's "Sidebar groups by Workspace" section with a pinning/folder-icon paragraph and update its one-line summary in `README.md` — grouping by surface means extending the Web App file, not adding a new one. (Depends on #30.)

**[medium] `design/specs/06-PROTOTYPE.md:45` — the Session-list contract is not updated.**
That bullet still describes the list as flat ("选择历史 Session 或新建 Session … 每条 Session 支持重命名、归档与删除"), with no mention of grouping modes, per-Project collapse persistence, or pinning. The design repo is clean, so no spec change is in flight either. Per CLAUDE.md the spec is ground truth — as written it says the sidebar has no group pinning while the code persists pin state per Project, so someone resolving a conflict in favor of the spec would delete the feature.
*Fix:* add one sentence to the `Session 列表` bullet covering the two grouping modes, collapse persistence and pinned-first ordering (this also closes the gap #26 left).

**[low] `packages/web/src/components/layout/sidebar.tsx:919-921` — `GroupPinButton` sets `aria-pressed={pinned}` *and* swaps its accessible name with state, the documented toggle-button anti-pattern.**
With a group pinned, a screen reader announces "Unpin group, button, pressed" — the state and the label contradict each other, so the user cannot tell whether the group is pinned. The same file's grouping-mode toggle at line 604 does it correctly with a static label plus `aria-pressed`.
*Fix:* keep the name static (`S.nav.pinGroup`) and let `aria-pressed` carry the state, or keep the action-name swap and drop `aria-pressed`.

**[low] `packages/web/src/components/layout/sidebar.tsx:637`, `:702` — the group collapse toggles have no `aria-expanded`, while this PR makes the open/closed state richer visually only.**
A screen-reader user tabbing the workspace header gets "Collapse, button" with no expanded state exposed; the folder-open affordance sighted users just gained has no non-visual equivalent. Pre-existing from #26, but this is the change that touches the state rendering.
*Fix:* add `aria-expanded={!collapsed}` to both toggles.

### Notes

- The pinning logic is correct where the risk was highest, and each hypothesis was checked rather than assumed: `pinnedFirst` (`session-grouping.ts:99`) is a partition, not a comparator — no total-order hazard, stable by construction, never mutates its input (`[...items]` on the empty fast path, `[...pin, ...rest]` otherwise; the test asserts `not.toBe(items)`). Pins survive reload via `penguin.sidebarPinnedGroups.<projectId>` with the same corrupt-data-degrades-to-empty handling as the collapse set, and the effect keys on the derived storage keys so a Project switch swaps both. A pin whose group disappears is simply skipped and re-applies if the workspace returns — a stray key, not a crash or phantom group.
- The `self-stretch` fix is geometrically sound: header rows are 28px because of the `h-7` action buttons, the collapse toggle was content-sized, and `align-self: stretch` makes its hover pill span the row while `items-center` keeps the glyphs centered. The pin button reserves 28px permanently (`opacity-0`, not removed), so there is no hover layout shift.
- Test coverage is real but asymmetric: the 4 new `pinnedFirst` unit tests hit empty-set, front-move stability, unmatched pinned keys, all-pinned and the merged-temp-group case, and the e2e addition is load-bearing. Only workspace mode is exercised end to end — `orderedAgents` has no test — and the post-reload `toBeVisible()` does not actually verify the "pinned stays visible" opacity claim, since Playwright visibility ignores opacity.

hiyouga:
两条 a11y 发现均已在 6ad80ae 修复：

1. **GroupPinButton 切换按钮反模式**：accessible name 固定为「置顶分组」（`aria-label={S.nav.pinGroup}`），状态只由 `aria-pressed` 表达，与 :604 分组模式切换按钮的模式一致；`title` 提示仍随状态切换（仅面向指针用户，aria-label 存在时不参与 accessible name）。e2e 相应改为断言 `aria-pressed="true"`。
2. **折叠开关缺 `aria-expanded`**：两处（agent 模式 / workspace 模式的分组头）均已加 `aria-expanded={!collapsed}`。

顺带按 review 备注补齐：
- `orderedAgents`（agent 模式置顶排序）补了单测：agentId 作 key 的分区、多个置顶项保持原相对顺序（session-grouping.test.ts，现 14 例）。
- e2e 的置顶指示器断言不再依赖 `toBeVisible()`（Playwright 忽略 opacity）：改为直接断言 computed opacity —— 置顶按钮为 1、未置顶分组的按钮悬停门控为 0（断言前先把鼠标移开，避免重排后旧点击位置误触 hover）。

未加 changelog（等 #30 之后第二波统一处理），未动 design 仓库。验证：`pnpm -r build` / `typecheck` / `test` 全绿，e2e 14/14（1.1m），`format:check` 干净。

PR #36, opened 2026-07-22
feat(web): collapsed-rail nav with last conversation, new chat, and localized tooltips

hiyouga:
## Summary

Implements the product requirement for the collapsed sidebar (narrow rail): it now shows, top to bottom, **last conversation, new chat, Agents, Skills, Models, Costs, Traces, Benchmark**, and every icon's hover tooltip is bilingual (follows the UI language).

Before this change the rail showed a generic `/chat` entry plus five page links (Benchmark missing) with hardcoded English `title`s ("chat", "agents", ...).

## Rail order and behavior

| # | Entry | Behavior |
|---|-------|----------|
| 1 | Last conversation | Navigates to the newest **non-archived** Session across the current Project (by `createdAt`; the flat store list is only ordered per Agent). Scopes the current Agent to that Session's Agent, matching `Sidebar.openSession`. When no such Session exists the button renders disabled (dimmed, tooltip kept) — falling back to new-chat would duplicate the adjacent button's action. |
| 2 | New chat | Mirrors the pinned sidebar's "New chat": `default_agent` draft via `/chat/new` (falls back to the first Agent), same gray active fill while on the draft page. |
| 3-8 | Agents / Skills / Models / Costs / Traces / Benchmark | `NavLink`s with the pinned nav's active-fill convention; Benchmark added. |

## Tooltip i18n

Each entry gets `title` + `aria-label` from the string tables (`strings.ts` zh / `strings-en.ts` en), read at render time so locale switching applies. Copy per the product owner's zh wording:

- New keys: `nav.lastConversation` (最近一次对话 / Last conversation) and `nav.railAgents` (智能体 / Agents — the rail uses the owner's short wording; the page keeps its full name 智能体仓库).
- Reused keys: `chat.newSessionMenu` (新建对话 / New chat) and the existing page names `nav.skills/models/usage/traces/benchmark` (技能库 / 模型仓库 / 成本中心 / 轨迹观测 / 评估中心), avoiding duplicated copy.

## Implementation notes

- The rail moved into a dedicated `CollapsedRail` component in `app-layout.tsx`; it reuses `Sidebar`'s `NAV_ICONS` and the new-chat pencil glyph (both now exported) plus the shared `GlyphIcon`, replacing the previously duplicated inline SVG paths.
- No changes to expand/collapse mechanics, persistence (`penguin.sidebarCollapsed`), or the pinned sidebar besides the icon-constant extraction.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm test` (docs 11, skills 16, landing 26, core 396, server 297, cli 121, web 351 — all green), `pnpm format` + `pnpm format:check`.
- `SKIP_BUILD=1 pnpm --filter @prismshadow/penguin-web test:e2e`: 15/15 passed, including a new `layout.spec.mjs` test covering rail order + en/zh tooltips (asserted via `aria-label`/`title`), the disabled empty state, last-conversation targeting the newest non-archived Session (with a newer archived one present), and draft/page navigation with active highlighting.

### DIFF
--- a/packages/web/e2e/layout.spec.mjs
+++ b/packages/web/e2e/layout.spec.mjs
@@ -12,7 +12,7 @@
  * - the collapsed rail shows, in product-specified order, last conversation / new chat /
  *   Agents / Skills / Models / Costs / Traces / Benchmark with localized (en + zh) hover
  *   tooltips; "last conversation" targets the newest non-archived session and is disabled
- *   while none exists;
+ *   while none exists; expanding from the rail restores the pinned sidebar;
  * - login page: a single brand penguin logo above the form (part of the form area; the
  *   background still only has the trace animation), the trace animation grows in after a
  *   delayed blank first paint, no two trace segments cross or touch (except where a fork shares
@@ -189,20 +189,33 @@ test("layout: collapsed rail — order, bilingual tooltips, last conversation", 
   expect(patched.ok(), "archive newest session").toBeTruthy();
   await page.reload(); // the session store loads on mount; the collapsed state persists via localStorage
 
-  await rail.getByRole("button", { name: "Last conversation" }).click();
-  await expect(page, "last conversation → newest non-archived session").toHaveURL(
-    `${BASE}/chat/${target}`,
-  );
+  // A click during the fetch window is a graceful no-op (the entry stays enabled while the
+  // list loads), so retry click+assert until the store has settled.
+  await expect(async () => {
+    await rail.getByRole("button", { name: "Last conversation" }).click();
+    await expect(page, "last conversation → newest non-archived session").toHaveURL(
+      `${BASE}/chat/${target}`,
+      { timeout: 1000 },
+    );
+  }).toPass({ timeout: 15_000 });
+  // Active fill = the *unprefixed* bg-gray-200/70 token (the resting state carries hover:bg-gray-200/70, which a bare substring match would also hit).
+  const ACTIVE_FILL = /(^|\s)bg-gray-200\/70(\s|$)/;
+  // On a conversation, the entry lights as "you are here" (any non-draft /chat/:id).
+  await expect(rail.getByRole("button", { name: "Last conversation" })).toHaveClass(ACTIVE_FILL);
 
   // --- New chat enters the draft page and shows the rail's gray active fill there ---
   await rail.getByRole("button", { name: "New chat" }).click();
   await expect(page).toHaveURL(`${BASE}/chat/new`);
-  await expect(rail.getByRole("button", { name: "New chat" })).toHaveClass(/bg-gray-200/);
+  await expect(rail.getByRole("button", { name: "New chat" })).toHaveClass(ACTIVE_FILL);
+  // The draft belongs to the new-chat entry: the last-conversation one must not stay lit here.
+  await expect(rail.getByRole("button", { name: "Last conversation" })).not.toHaveClass(
+    ACTIVE_FILL,
+  );
 
   // --- Page entries navigate and highlight like the pinned nav ---
   await rail.getByRole("link", { name: "Skills" }).click();
   await expect(page).toHaveURL(`${BASE}/skills`);
-  await expect(rail.getByRole("link", { name: "Skills" })).toHaveClass(/bg-gray-200/);
+  await expect(rail.getByRole("link", { name: "Skills" })).toHaveClass(ACTIVE_FILL);
 
   // --- zh: tooltips follow the product-specified wording ---
   await page.addInitScript(() => localStorage.setItem("penguin.lang", "zh"));
@@ -220,6 +233,11 @@ test("layout: collapsed rail — order, bilingual tooltips, last conversation", 
   ];
   expect(await attrs("aria-label"), "rail order (zh)").toEqual(ZH);
   expect(await attrs("title"), "rail tooltips (zh)").toEqual(ZH);
+
+  // --- Expand: the rail's top button (localized) restores the pinned sidebar ---
+  await page.getByRole("button", { name: "展开侧栏" }).click();
+  await expect(page.locator("aside")).toHaveClass(/w-64/);
+  await expect(page.getByRole("button", { name: "收起侧栏" })).toBeVisible();
 });
 
 test("layout: login — blank start, non-crossing traces, lang/theme controls", async ({ page }) => {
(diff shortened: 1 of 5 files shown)


rank-Yu (approved):

hiyouga (commented):
**Verdict: needs-changes** — the rail logic itself is sound (no `undefined` navigation, correct archived filter), but the two mandatory repo rules (design-spec sync, changelog entry) are unmet and the empty state fires during initial load.

### Findings

**[high] `design/specs/06-PROTOTYPE.md:13-25` — the PR redefines the left-nav contract without touching any spec.**
It drops the `/chat` rail entry, adds "last conversation" with explicit resolution semantics plus a disabled empty state, and adds Benchmark — while the 导航结构 section still lists 对话 as a nav entry and never mentions a 最近一次对话 entry or the rail's contents. The prior rail decision was recorded at `08-CHANGELOG.md:90` ("⑦ 侧栏收起态仅留按钮 + 头像"), so a reader following CLAUDE.md's "design docs are ground truth" gets the old contract. (`design/` is a symlink to `penguin-harness-design`, so the companion update lands there.)
*Fix:* update `06-PROTOTYPE.md` 导航结构 with the collapsed-rail order and the last-conversation/disabled semantics, and add a dated line at the top of `08-CHANGELOG.md`.

**[medium] `changelog/0.2.0/2026-07-22-web-app.md` — no changelog entry.**
The branch carries only the frozen `changelog/0.1.0/` tree, and #30's Web App entry (workspace sidebar / model picker / letter avatars / chat rendering) has no rail section, nor does `README.md`. 0.2.0 would ship with a reworked collapsed rail appearing nowhere in the release notes.
*Fix:* add a "## Collapsed rail" section to that entry and extend its README line. (Depends on #30 landing `changelog/0.2.0/`.)

**[medium] `packages/web/src/components/layout/app-layout.tsx:41` — `CollapsedRail` destructures only `sessions` from `useSessions()` and ignores `loading`, so the "no conversations yet" disabled branch (lines 96/101) is also what renders while the session list is still being fetched.**
With the rail collapsed, press F5 on a project that has plenty of chats: `SessionsProvider` starts at `sessions: []` / `loading: true` and only fills after projects → agents → per-agent `listSessions` round-trips (`ProjectProvider` renders children immediately, `state/project.tsx:152`). "Last conversation" paints dimmed and `disabled` on **every** load; a click in that window does nothing, then the button silently pops enabled.
*Fix:* pull `loading` from `useSessions()` and only apply the dimmed/disabled empty state when `!loading` (skeleton or plain enabled styling while loading).

**[low] `packages/web/src/components/layout/app-layout.tsx:70-77` — the `pages` list drops the old `{ to: "/chat" }` entry and nothing replaces its highlight role.**
The last-conversation button only lights when `activeSessionId === lastSession.sessionId` (line 100) and new-chat only on the draft (line 112). Collapse the sidebar and open the 2nd-newest conversation — no rail entry carries the gray active fill at all, so the rail shows no "you are here". Before this PR the `/chat` NavLink (react-router v7, `end` defaults to false) was active for every `/chat/*` URL.
*Fix:* light the last-conversation entry for any `/chat/:id` that isn't the draft, or keep a dedicated chat entry.

**[low] `packages/web/e2e/layout.spec.mjs:150` — the new test collapses the rail but never expands it again.**
Neither `onExpand` wiring (the top chevron nor the avatar button) is exercised: replace `onClick={onExpand}` with a no-op on both and the suite still reports 15/15 green — while the user would be permanently stuck in the rail with no way back to the pinned sidebar.
*Fix:* after the zh assertions, click `Expand sidebar` and assert the pinned sidebar returns.

### Notes

- The two things most likely to break in a rewrite this size are fine. `lastSession` (line 45) can only ever be a non-archived session and `openLastSession` early-returns when null, so there is no `/chat/undefined` path. Collapse state is untouched — same `penguin.sidebarCollapsed` read in a `useState` initializer — and the app is CSR-only (`createRoot` in `main.tsx`, no `hydrateRoot`), so there is no hydration-mismatch surface. Diffing responsibilities, `AppLayout` lost nothing: mobile header, `md:` breakpoint, initial-password banner, `ChangePasswordDialog`, Drawer and the `main` scroll container are byte-identical, and the old file had no effects or keyboard handling to drop.
- i18n is complete — both `lastConversation` and `railAgents` exist in both dictionaries (2 new keys, not 3). Note `en.nav.railAgents === en.nav.agents === "Agents"`: the new key only differs in zh, so the English side now has two copies that can drift.
- e2e ordering is not flaky as written (`createdAt` is ms-resolution and the three `mkSession()` calls are awaited sequentially; the fresh `mktemp -d` data root guarantees the "no sessions yet" precondition). Separately, pre-existing: `chat-page.tsx:190` picks `sessions[0]` for the `/chat` landing, which may be archived — a different answer from the rail's project-wide newest non-archived.

hiyouga:
Thanks for the review — all four points addressed in 20aaab2:

1. **Loading state (app-layout.tsx)**: `CollapsedRail` now pulls `loading` from `useSessions()`. The dimmed/disabled empty state applies only once the list has settled with no non-archived Session (`!lastSession && !loading`); while loading the entry keeps its plain enabled styling (no flash) and a click is a graceful no-op (`openLastSession` returns early). Went with plain styling rather than a skeleton — the fetch window is short and a one-off shimmer on a 32px icon reads as noise.

2. **Active highlight**: took the first option — the last-conversation entry now lights for any non-draft `/chat/:id` route (`onConversation`), restoring the rail-level "you are here" the old `/chat` NavLink provided. The draft page still belongs to the new-chat entry only.

3. **e2e**: after the zh assertions the test clicks the localized expand button (展开侧栏) and asserts the pinned sidebar returns (`aside` back at `w-64`, collapse button visible), so `onExpand` is exercised. Also added on/off assertions for the last-conversation highlight, and two test-robustness fixes this surfaced: the active-fill regex is now anchored to the unprefixed `bg-gray-200/70` token (a bare `/bg-gray-200/` also matched the resting `hover:` variant), and the last-conversation click retries via `toPass` since it may now legitimately land in the loading window.

4. **`railAgents` duplicate**: kept as a separate key with a comment in `strings-en.ts` stating it deliberately equals `nav.agents` and exists for the zh-only wording difference (智能体 vs 智能体仓库). A value reference isn't possible inside the object literal without hoisting a constant, which seemed worse than the comment.

No changelog entry per the coordination note (second wave after #30); design-repo docs untouched.

Verification: `pnpm -r build` / `pnpm typecheck` / `pnpm test` (all 7 packages) / `pnpm format` + `format:check` green; full Playwright suite `SKIP_BUILD=1 pnpm --filter @prismshadow/penguin-web test:e2e` 15/15 (run on private ports — 8930/8931 were held by a concurrent worktree's e2e server, which briefly produced misleading stale-bundle failures).

PR #37, opened 2026-07-22
fix(web): usage chart tooltip at the pointer's lower-right + cache hit rate in the Cache Read bubble

hiyouga:
## What

Two improvements to the Cost center's daily token chart tooltip (requirements: hover bubble at the mouse cursor's lower-right; Cache Read popup also shows the cache hit rate).

### 1. Tooltip position: pointer lower-right, edge-aware

**Before:** the bubble was pinned to the top of the chart (`top: 0`), horizontally anchored to the hovered column's x position and clamped by a hardcoded 160px width estimate.

**After:** the bubble follows the mouse pointer at a **+12px/+12px lower-right offset**. Positioning is computed by a new pure helper `bubblePosition()` in `chart-geom.ts`.

**Edge strategy — flip first, clamp as a last resort:**
- near the **right edge** the bubble flips to the pointer's lower-left; near the **bottom edge** it flips above the pointer; in the bottom-right corner it flips on both axes (upper-left). Flipping keeps the bubble clear of the pointer/hovered mark, where pure clamping would slide it back over them;
- a final clamp confines the bubble to the visible window in the degenerate case (window narrower than the bubble on both sides of the pointer) — covering is then unavoidable, and a window narrower than the bubble itself would still clip (unreachable at real card widths; the clamp just keeps the failure graceful);
- bounds are the scroll container's **currently visible window**, not the full canvas — the daily token canvas is wider than its card at 30 days (fixed 25px bars) and scrolls horizontally, and flipping against the visible right edge is what keeps the bubble on screen mid-scroll.

Placement is **imperative** (post-review): mousemove records the client position and coalesces into a single rAF that batches the layout reads (svg origin, visible window, real bubble size) and writes `style.left/top` directly — no React state, so a 60-120Hz pointer doesn't re-render the svg; a layout effect keyed on the hover anchor and bubble-renderer identity re-places before paint when content changes size. Implemented in the shared `ChartFrame`, so the daily cost chart (same frame, same page) gets identical behavior instead of diverging. Hit targets, hover semantics, and bubble styling are unchanged.

### 2. Cache hit rate in the Cache Read bubble

Hovering a `cache_read` segment now shows an extra line:

```
2026-07-21
cache_read 1.2M
命中率 98%   (en: "Hit rate 98%")
```

- **Formula:** `cacheRead / (cacheRead + cacheWrite)` — the **single shared helper `cacheHitRate()` in `lib/format.ts`**, also adopted by the Trace page's turn/global summaries (its local `hitRateOf` is now a thin adapter), so the metric cannot drift between pages.
- **Format:** whole percent via the existing `formatPercent`, identical to the Trace page.
- **Label:** reuses the existing pair `traces.hitRate` (zh `命中率` / en `"Hit rate"`) — exactly one label pair for this metric (cross-feature `S.*` reuse follows existing convention in this codebase).
- **Guard:** `cacheHitRate` returns `null` when `cacheRead + cacheWrite` is 0 and the bubble omits the line (in practice a cacheRead segment only exists when its value > 0, but the public helper must not divide by zero; `formatPercent(null)` would render `—`).

## Verification

- `pnpm -r build` — pass
- `pnpm test` — pass (web 361, core 396, server 297, cli 121, landing 26); new coverage: `bubblePosition` in `test/usage-charts.test.ts` (default lower-right, right/bottom/corner flips, exact-fit boundary, scrolled-window bounds, degenerate clamp) and `cacheHitRate` in `test/format.test.ts` (ratio values, whole-percent rendering via `formatPercent` incl. 33%/67%/99.9→100%, the 0/0 → null guard)
- e2e: checked — the usage page has no chart/tooltip spec coverage (only an incidental Cost-center navigation in an auth-redirect test), so no e2e applies to this change


rank-Yu (approved):

hiyouga (commented):
**Verdict: needs-changes** — the math and geometry are correct and well-tested, but the new Cost-center stat conflicts with the spec's own placement of that metric (PRN-015) with no spec update, and the change ships with no changelog entry.

### Findings

**[medium] `packages/web/src/features/usage/usage-charts.tsx:311` — adds a cache-hit-rate stat to the Cost center, a surface whose stat items the design spec enumerates exhaustively and which the spec assigns to the Trace page instead.**
`design/specs/01-PRINCIPLES.md:281` (PRN-015, 各观测面的统计项与顺序) puts 命中率 on the Trace page — "Trace 页：… Token 细分（输入 / 其中命中缓存与命中率 / 输出）" — while `:282` lists the Cost center's items with no hit rate. `06-PROTOTYPE.md:57` states the rule outright: "不拆缓存构成（命中缓存属排障口径，归 Trace 页）". `06-PROTOTYPE.md:115` also still describes the daily Token bar as "每段可单独高亮并显示该桶数值" (that bucket's value only).

This is the one finding in the batch that is a **design conflict rather than design lag** — CLAUDE.md says to confirm the design change and update the specs first, not implement around them. The tooltip-placement change is a separate matter: the specs are silent on bubble placement, so there is no conflict there, only a missing update.
*Fix:* confirm the placement decision, update PRN-015 + `06-PROTOTYPE.md:115` (and the placement behavior) in `penguin-harness-design`, then land the code.

**[medium] No changelog entry for a user-visible Web App change.**
`changelog/` is unchanged on this branch, and #30's `2026-07-22-web-app.md` (sidebar grouping / model picker / letter avatars / chat rendering) never mentions the usage chart tooltip or the cache hit rate, nor does `README.md`.
*Fix:* extend that entry with a section for the tooltip + hit rate and refresh its README summary line. (Depends on #30 landing `changelog/0.2.0/`.)

**[low] `packages/web/src/features/usage/chart-geom.ts:123` — third implementation-and-label of the same metric, at a different precision from the existing one.**
`packages/web/src/features/traces/trace-file-view.tsx:105-111` already defines `inputOf = cacheRead + cacheWrite` and `hitRateOf = cacheRead / input` — the identical formula — rendered through `lib/format.ts:81` `formatPercent` as `Math.round(ratio*100)%`. A ratio of 1,182,000 / (1,182,000 + 20,000) = 0.98336 renders "98%" on the Trace page and "98.3%" in the new bubble, under two different labels (`traces.hitRate` = "Hit rate"/"命中率" vs the new `usage.cacheHitRate` = "Cache hit rate"/"缓存命中率").
*Fix:* reuse `hitRateOf` + `formatPercent`, or move the shared pure formula into one place; pick one precision and one label.

**[low] `packages/web/src/features/usage/chart-svg.tsx:159` — `setPointer` on every `mousemove` re-renders the whole chart subtree, plus two forced synchronous layouts per move.**
The 30-day token chart reconciles ~225 SVG elements per mousemove (90 mark rects + 90 hit rects + 5 grid groups × 2 + up to 30 x-labels) at 60–120 events/s, where previously it re-rendered only when the hovered segment changed. Each pass also does `getBoundingClientRect()` in the handler and `offsetWidth`/`offsetHeight` in the dependency-less layout effect at `chart-svg.tsx:123`, which runs after *every* render.
*Fix:* hold the pointer in a ref and write `style.left/top` imperatively (or rAF-throttle `setPointer`), and give the measuring effect a dependency on the hovered key.

### Notes

- Coordinate spaces check out — this was the suspected bug and it isn't one. `pointer.x/y = clientX/Y − svgRect.left/top` is content-space; the bubble is an abspos child of the same scroll container (no padding/border, svg is `display:block` at the content origin) so its `left/top` share that space and translate with `scrollLeft`. Page scroll cancels (both terms viewport-relative). No viewBox scaling: `viewBox="0 0 w 200"` with `width={w} height={200}` and no `svg` rule in `styles.css`, so 1 user unit = 1 CSS px.
- The denominator is the right one for this codebase: `usage_records` has no uncached-input column (`db/schema.ts:64-67` — only `cache_read`, `cache_write`, `output`, `total`), and `trace-file-view.tsx:105` already treats cacheWrite as the miss, so `reads/(reads+writes)` really is hits/(hits+misses), matching `06-PROTOTYPE.md:123`. Degenerate inputs verified: (0,0) → `null` → the line is omitted, so no NaN/Infinity ever reaches the DOM; (0,100) → "0.0%"; (100,0) → "100.0%"; 99.99% → "100.0%" (asserted as intended). The tests do cover the 0/0 case.
- Geometry verified at every boundary with a 495px card and a 180×58 bubble: right edge, exact fit, one-past-fit, bottom, bottom-right corner, `px=0,py=0`, and a scrolled window — all land fully inside, bubble never covering the pointer, and the flipped `top` can never go negative for any bubble shorter than `CHART_H−2·12`. The one case that still clips is a visible window narrower than the bubble (150px window, 180px bubble), which contradicts the "clipping still isn't allowed" claim in the doc comment at `chart-geom.ts:145` and is untested — but unreachable at real card widths, so not filed as a finding.

hiyouga:
没问题 cache放这里可以的

hiyouga:
Review findings addressed in 817556b — disposition:

1. **Duplicate metric implementations — fixed (consolidated to one formula, one format, one label).** `cacheHitRate(cacheRead, cacheWrite): number | null` now lives in `lib/format.ts` next to `formatPercent`; the Trace page's `hitRateOf` is a thin `Buckets` adapter over it, and the Cost center bubble renders the same helper through the same `formatPercent` (whole percent) under the existing `traces.hitRate` label (zh 命中率 / en "Hit rate"). The one-decimal helper in `chart-geom.ts` and the new `usage.cacheHitRate` strings are removed, so exactly one formula and one label pair remain. Tests moved to `format.test.ts` with integer expectations (33%/67%, 99.9% → 100% at whole-percent precision); the 0/0 → null → line-omitted guard stays.

2. **Per-mousemove re-render — fixed (ref + rAF, imperative placement).** `mousemove` now only records `clientX/Y` into a ref and schedules a single `requestAnimationFrame`; the rAF callback batches all layout reads (svg origin, scroll window, real bubble `offsetWidth/Height`) and writes `style.left/top` directly. No React state is touched, so pointer movement no longer re-renders the svg, and read-then-write is coalesced to one batch per frame. The measuring/placement layout effect is keyed on `[bubble, hover]` (the bubble-renderer identity changes exactly when the parent's hover state — e.g. the hovered segment key — changes) instead of running per render; it re-places before paint when content changes size, so flip decisions always use the real size. The bubble mounts hidden and is revealed by its first placement.

3. **Comment honesty — fixed.** `bubblePosition`'s doc (and the matching test) now state that in the degenerate case the bubble covers the pointer, and a window narrower than the bubble itself still clips on the right — unreachable at real card widths; the clamp just keeps the failure graceful.

**Placement question:** settled per the owner's issue comment ("没问题 cache放这里可以的") approving the hit rate's placement in the Cache Read bubble — no code change needed; spec updates are handled centrally, so this PR does not touch the design docs. No changelog entry yet per the coordination plan (second wave after #30 lands).

Validation after the fixes: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (web 361 / core 396 / server 297 / cli 121 / landing 26), `pnpm format:check` — all pass.

PR #38, opened 2026-07-22
feat(core): subagents inherit the parent session's model and thinking level

hiyouga:
A spawned subagent must follow the PARENT session's model choice, thinking level, and workspace — never the Project default. This wires that inheritance into the spawn site in core.

## Before / after

| Aspect | Before | After |
| --- | --- | --- |
| Model | `run_subagent` with `model_id`/`provider` omitted fell through to the **Project `default_model`**, so a parent running a pinned model spawned children on a different model. | The spawn site forwards the parent session's resolved `(provider, model_id)` pair. The child re-resolves the same Project-config entry, so per-model `max_tokens` / `context_window` / `vision` follow automatically. |
| Thinking level | The child always used **its own Agent config** (`system_config.model.thinking_level`) — on cross-agent spawns that could differ from the parent. | The parent session's **effective** level is passed down via the new `CreateSessionOptions.thinkingLevel` and used for both the `session_meta` echo and the LLM config. Self-invocation is semantically unchanged. |
| Workspace | Already inherited (the spawn closure passes the parent's `workspaceDir`). | Unchanged — now locked by an explicit test assertion. |

## Rules

- **Explicit tool args still win**: a complete `(provider, model_id)` pair in the tool call selects that model; inheritance only fills a fully omitted pair. Half a reference is still rejected (both by the tool and by `createSession`) — it is never silently completed from the parent's half.
- The tool surface is unchanged: no new `run_subagent` parameters; workspace/thinking inheritance is implicit.
- `resumeSession` keeps reading the Agent config (no per-session override exists on resume).

## Changes

- `packages/core/src/agent.ts` — `CreateSessionOptions.thinkingLevel`; the effective level is resolved in `createSession` (option wins over Agent config) and threaded through `buildRuntime` into the meta echo and `llmConfig`; the spawn closure inherits the parent's model pair and effective thinking level.
- `packages/core/src/state/default-config.ts` — `run_subagent`'s `model_id` description now says "Omit both to inherit the parent session's model."
- `packages/core/src/environment/tools/run-subagent.ts` — header comment and half-pair error message updated to the same wording.
- `packages/core/src/interfaces.ts`, `packages/core/src/omnimessage/types.ts` — doc comments updated (`SubagentRunner.spawn`, `SessionMetaPayload.thinking_level`).
- `packages/docs/content/{tools,interfaces}.{en,zh}.md` — stale "Project default model" wording replaced; a bullet documents the three inherited aspects.
- Tests: `agent.test.ts` drives the **real** spawn closure (an Environment wrapper captures the injected `SubagentRunner`; reading the first `handle.run` message yields the child `session_meta` before any LLM request, so no network) — parent pair/thinking/workspace inheritance, explicit-args-win, and a cross-agent spawn where the child's own config loses to the parent. `subagent.test.ts` reframes the tool-level case: an omitted pair is forwarded as omitted (the tool never invents one; the runner inherits).

## Seed caveat

Tool descriptions are seeded into each agent's `system_config.yaml` at agent creation, so **existing agents keep the old `model_id` description text** (same caveat as #24). The behavior change applies to them regardless — inheritance happens at the spawn site, not in the tool description.

## Verification

- `pnpm -r build` — all packages build.
- `pnpm typecheck` — clean across the workspace.
- `pnpm test` — core 401 passed / 2 skipped (e2e), server 297, cli 121, web 351, skills 16; includes 5 new inheritance tests and the rewritten tool-level case.

### DIFF
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -70,6 +70,26 @@ import type { ModelEntry } from "./state/index.js";
  */
 const MAX_SUBAGENT_DEPTH = 1;
 
+/** The five valid thinking level names (session_meta additionally records the literal "default" for "no level"). */
+const THINKING_LEVEL_NAMES: readonly ThinkingLevelName[] = [
+  "none",
+  "low",
+  "medium",
+  "high",
+  "xhigh",
+];
+
+/**
+ * Narrows a recorded session_meta `thinking_level` back to a ThinkingLevelName; the literal
+ * "default" (no level recorded), a missing field (legacy Trace), and anything unknown are
+ * not levels and yield undefined.
+ */
+function asThinkingLevelName(value: unknown): ThinkingLevelName | undefined {
+  return typeof value === "string" && (THINKING_LEVEL_NAMES as readonly string[]).includes(value)
+    ? (value as ThinkingLevelName)
+    : undefined;
+}
+
 export interface CreateAgentOptions {
   agentId?: string;
   projectId?: string;
@@ -91,12 +111,16 @@ export interface CreateSessionOptions {
    */
   provider?: string;
   /**
-   * Thinking level for this Session; when unspecified, falls back to the Agent config's
-   * `model.thinking_level`. Subagent spawning passes the parent Session's effective level
-   * here, so a child Session thinks at the parent's level rather than the child Agent's
-   * own default.
+   * Thinking level for this Session — a tri-state:
+   * - a `ThinkingLevelName` pins the level;
+   * - `undefined` (omitted) falls back to the Agent config's `model.thinking_level`;
+   * - `null` means "no thinking level": the config fallback is suppressed entirely
+   *   (nothing goes into the LLM config; session_meta echoes `"default"`).
+   * Subagent spawning always passes the parent Session's effective level (its level, or
+   * `null` when the parent has none), so a child Session thinks at the parent's level and
+   * never falls back to the child Agent's own config.
    */
-  thinkingLevel?: ThinkingLevelName;
+  thinkingLevel?: ThinkingLevelName | null;
   /** Explicit credentials; if unspecified, falls back to credentials in the Project config, then to AgentHub reading environment variables. */
   apiKey?: string;
   baseUrl?: string;
@@ -214,9 +238,13 @@ export class Agent {
     }
     const sessionId = formatSessionId();
     const subagentDepth = opts.subagentDepth ?? 0;
-    // Effective thinking level: an explicit option (how subagent spawning passes down the
-    // parent Session's level) wins over this Agent's own system_config.
-    const thinkingLevel = opts.thinkingLevel ?? this.state.systemConfig.model?.thinking_level;
+    // Effective thinking level (tri-state option): a value wins over this Agent's own
+    // system_config; `null` — how subagent spawning says "the parent has none" — suppresses
+    // the config fallback entirely; only `undefined` (no option) reads the Agent config.
+    const thinkingLevel =
+      opts.thinkingLevel === null
+        ? undefined
+        : (opts.thinkingLevel ?? this.state.systemConfig.model?.thinking_level);
 
     // Agent-level vault (agent_state/.vault.toml) and installed Skills: read the current values each time a Session is created.
     const vault = await loadAgentVault(this.state.root, this.state.projectId, this.state.agentId);
@@ -361,14 +389,20 @@ export class Agent {
     // original history); the vault uses current values (it's injected into the
     // subprocess environment, not the history, so a resumed Session should get the
     // latest keys too).
+    // The recorded thinking level is restored with the rest of session_meta (like model,
+    // Workspace, and system prompt): a resumed subagent session keeps its inherited level
+    // instead of re-reading this Agent's config. The literal "default" (no level recorded)
+    // — and a legacy Trace without the field — falls back to the Agent's current config.
+    const thinkingLevel =
+      asThinkingLevelName(meta.thinking_level) ?? this.state.systemConfig.model?.thinking_level;
+
     const rt = await this.buildRuntime({
       workspaceDir,
       modelEntry,
       apiKey,
       baseUrl,
       systemPrompt: meta.system_prompt,
-      // Resume has no per-Session override: the level follows this Agent's current config.
-      thinkingLevel: this.state.systemConfig.model?.thinking_level,
+      thinkingLevel,
       subagentDepth: 0,
       vault: await loadAgentVault(this.state.root, this.state.projectId, this.state.agentId),
     });
@@ -407,7 +441,7 @@ export class Agent {
         model_context_window: modelEntry.context_window ?? "unknown",
         system_prompt: meta.system_prompt,
         tools: rt.tools,
-        thinking_level: this.state.systemConfig.model?.thinking_level ?? "default",
+        thinking_level: thinkingLevel ?? "default",
         agent_state: this.state.stateDir,
         workspace: workspaceDir,
       },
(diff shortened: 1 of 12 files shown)


rank-Yu (approved):

hiyouga (commented):
**Verdict: needs-changes** — the inheritance logic itself is correct and well-tested, but the change rewrites a contract that `05-ARCHITECTURE.md` states verbatim in two places with no spec update, and there is no changelog entry.

> **Merge conflict:** this branch conflicts with #40 on `packages/core/test/agent.test.ts`. Whichever lands second needs a rebase.

### Findings

**[high] `design/specs/05-ARCHITECTURE.md:167` and `:506` — the spec still states the pre-change subagent-spawn contract that this PR inverts.**
`:167` reads `modelId?: string;    // 省略 = Project 默认 Model` and `:506` reads `派生时可指定 modelId，未指定时用 Project 默认 Model。`, while `packages/core/src/agent.ts:527-529` now resolves an omitted pair to the parent session's `{ modelId, provider }` — the Project `default_model` path is unreachable from a spawn. `design/` is a gitignored symlink, so a spec change would not appear in `gh pr diff`; I checked `penguin-harness-design` directly — clean on `main` at `afb539d "Sync the specs with the 0.1.0 release"`, no branch and no PR for this. Two further gaps in the same spec: `:95` enumerates `createSession`'s options without the new public `thinkingLevel` option (`agent.ts:99`), and `01-PRINCIPLES.md:141` documents `thinking_level` purely as a `system_config.yaml` Agent field — the spec has no notion of a per-Session level or of a child overriding its own Agent config.
*Fix:* land a spec PR in `penguin-harness-design` updating `05-ARCHITECTURE.md:167`, `:506` and `:95` (child follows the parent session's model + thinking level + Workspace; explicit pair wins; half a pair rejected) before or alongside this one.

**[medium] `packages/core/src/agent.ts:371`, `:410` — `resumeSession` re-reads the child Agent's own `system_config.model.thinking_level`, so resuming a subagent session silently discards the level its Trace recorded — even though model, Workspace and system prompt are all restored from `session_meta`.**
Parent `default_agent` at `xhigh` spawns a child on `helper_agent` (own config `low`). The child runs at `xhigh` and its Trace `session_meta.thinking_level` is `"xhigh"`. The server registers that child in the session index with `source: "subagent"` (`session-manager.ts:893`) and shows it in the sidebar with no read-only guard; the user opens it and sends a message → `createCoreSessionLoader.load` calls `resumeSession` (`session-manager.ts:119`) → `thinkingLevel` becomes `helper_agent`'s `low`. The session now runs at `low` while its own persisted `session_meta` (what the trace view renders) says `xhigh`. The PR body calls this intentional, but its rationale ("no per-session override exists on resume") is stale — this PR creates exactly such an override and already persists it.
*Fix:* in `resumeSession`, seed `thinkingLevel` from `meta.thinking_level` when it is one of the five `ThinkingLevelName` values (treat `"default"` as undefined), matching how model/Workspace/system prompt are restored, and use the same resolved value at line 410.

**[medium] `changelog/0.2.0/` — no changelog entry.**
`git diff origin/main...HEAD -- changelog/` is empty. On #30's branch neither `2026-07-22-model-settings.md` nor `2026-07-22-agent-guardrails.md` mentions subagent inheritance, and `README.md` has no line for it — so a user-visible behavior change (subagents stop using the Project default model) would ship unrecorded.
*Fix:* extend `2026-07-22-model-settings.md` with a "Subagents follow the parent session" section and update its README summary line, rebasing once #30 lands.

**[low] `packages/core/src/agent.ts:219`, `:539` — when the parent's effective thinking level is `undefined`, the spread omits it and the child falls back to *its own* Agent config, so the PR's stated invariant ("the child Agent's own config loses to the parent") does not hold on that path.**
A user removes `model.thinking_level` from `default_agent`'s `system_config.yaml` (the field is optional — `state/default-config.ts:69` — and the server's PUT can only set, never unset, so hand-editing is the route). The parent then runs with no thinking level. It spawns `helper_agent`, whose config pins `xhigh`: line 539 omits `thinkingLevel`, line 219 resolves `undefined ?? "xhigh"`, and the child reasons at `xhigh` while the parent does not reason at all — the exact cross-agent divergence the PR set out to remove, plus an unexpected reasoning-token bill.
*Fix:* pass the parent's level as an explicit tri-state (e.g. `thinkingLevel: thinkingLevel ?? null` meaning "parent has none, do not fall back").

**[low] `packages/core/test/agent.test.ts:223-380` — the new tests pin the happy paths but not the two edge cases the PR's own rules turn on.**
Nothing exercises (a) a parent whose Agent config has no `thinking_level` — the fallback hole above would pass CI unnoticed — or (b) `runner.spawn({ modelId })` with the provider missing; the PR states "half a reference is never silently completed from the parent's half", but that is only asserted at the tool layer in `subagent.test.ts`, which returns before reaching the runner.
*Fix:* add one test spawning after `delete agent.state.systemConfig.model.thinking_level`, and one asserting `runner.spawn({ modelId: "claude-sonnet-4-6" })` rejects with `/must be given as a \(provider, model_id\) pair/`.

### Notes

- Resolved precedence, read off the code — **model**: (1) explicit complete pair from the tool → `resolveModelRef` on the child's project config; (2) both omitted → parent session's resolved `modelEntry` pair; (3) Project `default_model` now unreachable via spawn; (4) half a pair rejected twice (`run-subagent.ts:87`, `agent.ts:163`). **Thinking level**: (1) `CreateSessionOptions.thinkingLevel` → (2) the created Agent's `system_config.model.thinking_level` → (3) omitted from `llmConfig`, meta `"default"`. Both match the stated intent except step (2) of the thinking chain, flagged above. No explicit per-subagent setting is silently overridden.
- Nesting and staleness are fine. A grandchild would inherit transitively because the child's `buildRuntime` receives the already-resolved `modelEntry` and `thinkingLevel`, though `MAX_SUBAGENT_DEPTH = 1` makes it unreachable today. Capturing the parent's level in the closure rather than re-reading is correct: per #28 a level change takes effect at the *next* session creation, so the parent's own LLM is likewise fixed for the session's lifetime.
- No falsy-swallow and no persistence break. Every new defaulting site uses `??` or an explicit `!== undefined` guard, and `mapThinkingLevel` maps `"none"` behind a `thinking !== undefined` check, so `"none"` survives. `omnimessage/types.ts` and `default-config.ts` are comment/description-only edits, so pre-change Traces resume unchanged. The seeded `run_subagent` description does go stale for existing agents (`agent-state.ts:416` takes stored `tools.builtin` wholesale), which the PR acknowledges as the #24 precedent. Language rule is clean.

hiyouga:
Thanks for the thorough review — addressed in 7345221. Disposition per finding:

**[medium] `resumeSession` discards the recorded thinking level — fixed as specified.** `resumeSession` now narrows `meta.thinking_level` back to a `ThinkingLevelName` (new `asThinkingLevelName` guard over the five names) and seeds the resolved value into both `buildRuntime` and the `session_meta` echo. The literal `"default"` — and a legacy Trace without the field — still falls back to the Agent's current config, exactly as before. A resumed subagent session therefore keeps its inherited level (parent `xhigh` stays `xhigh` even when `helper_agent`'s own config says `low`), consistent with how model/Workspace/system prompt are restored. Pinned by a new test in `resume.test.ts` covering both the restoration and the `"default"` fallback.

**[low] undefined-parent fallback hole — fixed with the tri-state you proposed.** `CreateSessionOptions.thinkingLevel` is now `ThinkingLevelName | null`: a value pins the level, `undefined` (omitted) falls back to the Agent config (unchanged for non-spawn callers), and `null` suppresses the config fallback entirely (nothing in the LLM config, meta echoes `"default"`). The spawn site always passes `thinkingLevel: thinkingLevel ?? null`, so a parent with no level produces a child with no level — the child's own config can never apply on the spawn path. Doc comment on the option states the tri-state semantics.

**[low] missing edge tests — added.** (a) spawn after deleting the parent's `model.thinking_level` with a cross-agent child whose config pins `medium` → the child meta must read `"default"`; (b) `runner.spawn({ modelId: "claude-sonnet-4-6" })` (and the provider-only mirror) rejects with `/must be given as a \(provider, model_id\) pair/` at the runner layer.

**[high] spec (`05-ARCHITECTURE.md` / `01-PRINCIPLES.md`) — handled centrally.** The `penguin-harness-design` updates are being coordinated in a separate central pass rather than per-PR; not included here.

**[medium] changelog — deferred to the second wave.** Per the current plan, the changelog entry lands after #30 merges (extending `2026-07-22-model-settings.md` with a "Subagents follow the parent session" section and its README line), so this branch intentionally carries no `changelog/` diff yet.

Resulting precedence (thinking level): explicit `CreateSessionOptions.thinkingLevel` value → `null` = no level (spawn's "parent has none") → `undefined` = Agent config → nothing (meta `"default"`); resume reads the recorded level first, `"default"`/legacy falls back to the Agent config.

Validation: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (core 404 passed / 2 skipped e2e; all other packages green), `pnpm format:check` clean.

PR #41, opened 2026-07-22
fix(web): keep chat dropdown menus inside the viewport on mobile

hiyouga:
## Summary

对话页面的有些下拉框在手机端会超出屏幕宽度 — reproduced with Playwright at 375×667 and 390×844 (en locale, the wider copy): three chat-page dropdown panels ran past the viewport because their width clamps ignored the trigger's own anchor offset (`max-w-[calc(100vw-2rem)]` caps the width, but a panel anchored 8rem into the row can still cross the edge). Note the app's root font size is 18px, so all rem figures below are 18px-based.

**Measured overflows (before)**

| Dropdown | Anchoring | Overflow |
| --- | --- | --- |
| Model select (`chat-input.tsx`, draft toolbar) | `right-0` to a button ~70px from the viewport's right edge, `w-max` | left edge **−34px** off-screen @375 (−35px with the show-all expander open) |
| Skills select (`chat-input.tsx`, draft + session toolbar) | `left-0` to a button ~128px into the toolbar, `w-[26rem]` | right edge **+92px** past 375 — and the menu's autofocused search box then scrolled the whole draft page sideways by ~68px, cutting off the left edge of the UI (the most visible symptom) |
| Workspace select (`draft-view.tsx`, ownership pill) | `left-0` to a pill whose offset varies with the agent pill's width | right edge **+143px** past 390 when both pills share one row |

Approval mode, thinking level, agent select, and the slash/@-mention menus measured in-viewport at both sizes and are untouched.

## Fix

- **Model select** — the panel's right edge is docked to the button, whose offset from the viewport's *right* edge is fixed (page + card padding, gap, send button), so the clamp just has to reserve it: `max-w-[calc(100vw-2rem)]` → `max-w-[calc(100vw-6rem)]`.
- **Skills select** — same idea from the left: `max-w-[calc(100vw-10rem)]`, plus `@md:max-w-[calc(100vw-17rem)]` for the container width at which the toolbar buttons grow text labels (the anchor offset roughly doubles). Applied to both the downward (draft) and upward (session) variants.
- **Workspace select** — the anchor offset is content-dependent (the agent pill's name width), so no static clamp is safe: on each open the pill measures the room on both sides, keeps the desktop `left-0` docking whenever the panel fits, and otherwise docks to whichever side has more room, capping the width to that room. Carried by a new optional `menuStyle` passthrough on the shared `Dropdown` (backward-compatible; no other caller changes).

**Desktop unchanged**: all three clamps resolve far above the panels' fixed/content widths at desktop viewports, and the workspace menu always fits left-anchored there (verified the clamps only bite when `100vw` is small).

## Verification

Full chain on node v24 — `pnpm install --frozen-lockfile` → `pnpm -r build` → `pnpm typecheck` → `pnpm test` (cli 121, web 351, all green) → e2e → `pnpm format` / `pnpm format:check`:

- `layout.spec.mjs` gains **"mobile chat dropdowns stay inside the viewport"**: opens every chat dropdown at 375×667 and 390×844 on the draft page (including the model menu's show-all state; the two widths exercise both the wrapped and one-row ownership-pill geometries) plus approval/skills on the session composer, and asserts the panel's left/right edges stay on-screen, no ancestor is scrolled sideways, and `scrollWidth ≤ clientWidth`.
- The new assertions **fail against the previous code** (verified: `skills @375: panel right edge on-screen — Received: 398.5`) and pass with the fix.
- Full e2e suite: **15 passed** (`SKIP_BUILD=1 pnpm --filter @prismshadow/penguin-web test:e2e`).

### DIFF
--- a/packages/web/e2e/layout.spec.mjs
+++ b/packages/web/e2e/layout.spec.mjs
@@ -8,6 +8,11 @@
  * - the models page at 390x844 must not overflow horizontally, and text must not overlap
  *   (the group header's provider name used to get pushed out of the button box and overlap
  *   the group-level actions);
+ * - every chat-page dropdown menu, opened at phone widths (375/390), must keep its panel
+ *   inside the viewport and must not shove the page sideways (the model menu used to run
+ *   ~34px off-screen left, the skills menu ~92px off-screen right — with its autofocused
+ *   search box then horizontally scrolling the whole draft page — and the workspace menu
+ *   ~143px off-screen right when the ownership pills share one row);
  * - the sidebar's "New chat" button has no background fill (same gray-scale style as nav items);
  * - login page: a single brand penguin logo above the form (part of the form area; the
  *   background still only has the trace animation), the trace animation grows in after a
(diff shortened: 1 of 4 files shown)


rank-Yu (approved):

hiyouga (commented):
**Verdict: approve-with-nits** — the diagnosis is right, all three clamps are correct by construction (including under the user-configurable root font size), and the new e2e is genuinely load-bearing. The gaps are the missing changelog entry, a design-doc line this PR proves false, and a stale-measurement edge case on rotation.

### Findings

**[medium] `changelog/0.2.0/` — no changelog entry for a user-visible fix.**
The branch touches only the four source/test files. `changelog/0.2.0/` does not exist on `main` — #30 creates it — and #30's `2026-07-22-web-app.md` has no section this belongs to.
*Fix:* after #30 lands, extend `2026-07-22-web-app.md` with a mobile-dropdown paragraph and refresh its `README.md` summary line. Per the grouping rule this extends the existing Web App surface entry rather than opening a new file.

**[medium] `design/specs/08-CHANGELOG.md:30` records a claim this PR falsifies, and is not corrected.**
That line states 技能下拉加宽至 26rem（视口钳制，手机不溢出）— the skills dropdown widened to 26rem, *viewport-clamped, no overflow on phones*. This PR measures 92px of right-edge overflow at 375px from exactly that `w-[26rem] max-w-[calc(100vw-2rem)]` combination, so the recorded claim was wrong when written: the clamp capped the width but ignored the ~8rem anchor offset. Since the design docs are ground truth, a reader today concludes the phone case is already handled and would not expect this fix to be needed.
*Fix:* add a corrective dated line in `penguin-harness-design`'s `08-CHANGELOG.md` noting the clamp must reserve the anchor offset, not just the width. (`design/` is a gitignored symlink, so this cannot ride this branch.)

**[low] `packages/web/src/features/chat/draft-view.tsx:819-833` — the workspace dock is measured once on open and never re-measured, so a viewport change while the menu is open leaves both the docking side and the pixel `maxWidth` stale.**
`Dropdown` closes only on outside-mousedown and Escape (`dropdown.tsx:30-45`) — there is no resize or orientation handler anywhere in the chain — so the menu survives a rotation. The damaging direction is landscape → portrait: opened at 844px wide the pill has room, so `toggle` takes the first branch and sets `{ right: false }` with **no** `maxWidth`; rotate to 390px and the panel keeps `left-0` at its full `w-80`, and with the pill reflowed to ~150px in, its right edge lands ~114px past the viewport — reintroducing exactly the symptom this PR fixes. Portrait → landscape is only cosmetic (a menu needlessly capped to the frozen portrait width).
*Fix:* while `open`, re-run the measurement on `resize`/`orientationchange` (or simply close the menu on viewport resize, which also matches how the other two clamps behave, since pure-CSS `calc(100vw-…)` reflows for free).

**[low] `packages/web/src/features/chat/draft-view.tsx:826` — the JS re-derives a width contract that lives in the `menuClass` three lines below, with no coupling between them.**
`Math.min(20 * rem, window.innerWidth - 2 * rem)` hard-codes `w-80` and `max-w-[calc(100vw-2rem)]` from `menuClass` at line 869. The inline comment does flag the dependency, which helps, but editing either side silently desynchronizes them — and a wrong `width` here does not fail loudly, it just picks the wrong dock or caps to the wrong room.
*Fix:* hoist the two numbers to shared constants used to build both the class string and the measurement, so the pair can only change together.

### Notes

- **The clamps are font-size-robust, which is worth recording because the PR body's "root font size is 18px" is only the default.** `state/theme.tsx:28` defines `FONT_PX = { sm: "16px", md: "18px", lg: "20px" }` and `:95` writes it to `documentElement`, driven by a user-facing Font size setting. Expressing the clamps in `rem` is what makes this safe: the anchor offsets they reserve (page/card padding, gaps, the send button) are Tailwind spacing utilities, which are themselves rem-derived, so offset and clamp scale by the same factor and `6rem ≥ offset` holds at all three tiers. A `px` clamp would have broken at `lg`. The runtime path is independently safe — `toggle` reads the live `getComputedStyle(documentElement).fontSize` rather than assuming 18. Only the default tier is exercised by the e2e, but the invariant holds by construction, so I'm not filing it.
- **The `right-0`/measurement pairing checks out**, which was the one thing that could have invalidated the whole approach. `right-0`/`left-0` resolve against the `Dropdown` root (`dropdown.tsx:48`, `relative`), not the button, while `toggle` measures `e.currentTarget` (the button) — those agree only if the root shrink-wraps. It does: the pills row is `flex flex-wrap items-center gap-2` (`draft-view.tsx:546`), so the root is a `flex: 0 1 auto` item sized to the button. `toggle` also has exactly one call site (`onClick={toggle}` at line 878), so the new `ReactMouseEvent` parameter is never undefined. The dock branches are sound: branch 2 only right-docks when `roomLeft > roomRight`, and leaves `maxWidth` unset only when `roomLeft ≥ width`, which guarantees `r.right − width ≥ 12 > 0`.
- **The `@md:` variant is correctly container-scoped**, not viewport-scoped: the panel renders inside the card that carries `@container` (`chat-input.tsx:1229`), and `@md` is the same breakpoint at which the toolbar buttons already reveal their text labels (`@md:block` at `chat-input.tsx:149/295/428/529` on `main`) — which is precisely the transition that roughly doubles the anchor offset. The two clamp values track that.
- **The new e2e is load-bearing and well-targeted** — it asserts panel edges, no horizontally-scrolled ancestor, and `scrollWidth ≤ clientWidth`, covers six dropdowns at two widths plus the upward session-composer variants, and deliberately seeds a long model id and a key-less model to force the `w-max` panel and the show-all expander to their widest states. Two small things: the `waitForTimeout(200)` for the pop-in animation is a hard sleep in an otherwise assertion-driven suite, and nothing exercises a viewport change while a menu is open (the gap behind finding 3).
- Cosmetic: the PR description opens in Chinese while the rest of the repo's PR bodies are English. Not covered by the CLAUDE.md language rule (which governs repo contents, not PR metadata), so noted only for consistency.

PR #43, opened 2026-07-22
feat: AgentHub 0.4.1 and a model-catalog refresh across every provider group

hiyouga:
Bumps `@prismshadow/agenthub` from `^0.4.0` to `^0.4.1` (npm `latest`), documents the new API in the `agenthub-models` skill, and refreshes the built-in model catalog so it offers **every chat model AgentHub 0.4.1 can route** — 11 new entries across five provider groups, not just OpenRouter.

## What 0.4.1 actually changed

Verified against the package itself, not the release notes. Three pieces of evidence, retraceable with:

```bash
npm pack @prismshadow/agenthub@0.4.0 @prismshadow/agenthub@0.4.1   # then untar both
diff -rq v040/package/dist v041/package/dist
node -e 'console.log(require("@prismshadow/agenthub").listSupportedModels().length)'   # 49
```

**Type-compatible with 0.4.0.** `diff -rq` reports `dist/types.d.ts`, `dist/autoClient.d.ts` and `dist/baseClient.d.ts` identical between the two tarballs: `UniConfig`, `ThinkingLevel`, `UniEvent`/`UniMessage`, the content-item union and the `AutoLLMClient` surface are all unchanged. `packages/core/src/llm/generative-model.ts` needed no adaptation, and `pnpm typecheck` passed on this repo before any source change.

New public surface, all additive in `dist/index.d.ts`:

- **`listSupportedModels(currency?: "USD" | "CNY"): SupportedModel[]`** — a 49-entry supported-model registry. Each entry is `{ model, base_url, client, input_modalities, output_modalities, context_window?, pricing? }`; the `(model, base_url, client)` triple maps straight onto the `AutoLLMClient` constructor. `pricing` is per million tokens keyed by AgentHub's usage buckets (`prompt_tokens`, `thoughts_tokens`, `response_tokens`, optional `cached_tokens`), stored in USD and converted at 7 CNY/USD on request. Coverage is the official vendor endpoints plus OpenRouter and SiliconFlow. New exported types: `Currency`, `Modality`, `ModelPricing`, `SupportedModel`. *Evidence: signature and doc comments from `dist/registry.d.ts` (a "Only in v041" file in the diff); the entry count, modality vocabulary, currency conversion and every price in this PR come from running it.*
- **`UnsupportedParameterError`** (extends `AgentHubError`, carries `client` and `parameter`) — every client's bare `throw new Error(...)` for a rejected `temperature` / `tool_choice` / `prompt_caching` value became this typed error. Thinking levels never raise: each client maps a `ThinkingLevel` onto the closest one the model supports. *Evidence: `dist/errors.d.ts` and `dist/index.d.ts` diffs; the per-client rules from grepping every `dist/*/client.js` for `UnsupportedParameterError` and reading each guard (`!== undefined` for Gemini 3.6 vs `!== undefined && !== 1.0` for the rest).*
- **Three new protocol clients** wired into `AutoLLMClient._createClientForModel`: `gemini3_6` (matched *before* the broader `gemini-3` prefix, covering `gemini-3.6-*` and `gemini-3.5-flash-lite`, rejecting `temperature` outright), `glm5_2` (sends `reasoning_effort` alongside its `thinking` block), `kimi_k3` (maps `ThinkingLevel` onto `reasoning_effort` low/high/max, always reasons, allows `tool_choice: "required"`, caches context automatically). *Evidence: `dist/gemini3_6/`, `dist/glm5_2/`, `dist/kimi_k3/` are "Only in v041"; precedence from the `dist/autoClient.js` diff.*
- The deprecated image endpoint `gemini-3.1-flash-image-preview` is now `gemini-3.1-flash-image`. *Evidence: the id `listSupportedModels()` returns. Caveat: `dist/integration/playground.js` still hardcodes the old `-preview` id, unchanged from 0.4.0 — an upstream inconsistency nothing here depends on.*

No impact on this repo's request path: `buildUniConfig` only ever sets `tools`, `system_prompt`, `max_tokens` and `thinking_level`, never `temperature` / `tool_choice` / `prompt_caching`, so `UnsupportedParameterError` is unreachable from PenguinHarness today.

`packages/cli/package.json` also declared `@prismshadow/agenthub` and was bumped alongside core.

## Model catalog: 59 -> 70 entries

Method: dump `listSupportedModels("USD")` from the installed 0.4.1 inside `packages/core`, map each `base_url` to a provider group, and set-difference against `MODEL_CATALOG`. That found 11 routable chat models with no catalog entry. Every row below was then re-checked field by field against the registry (a scripted comparison of context window, `input_modalities` and all three price buckets) — no row was hand-transcribed.

| Group | Model id | Display name | Context | `usd`/`cny` triple (cache_read, cache_write, output) | Vision |
| --- | --- | --- | --- | --- | --- |
| `google` | `gemini-3.6-flash` | Gemini 3.6 Flash | 1,048,576 | `usd(0.15, 1.5, 7.5)` | yes |
| `google` | `gemini-3.5-flash-lite` | Gemini 3.5 Flash-Lite | 1,048,576 | `usd(0.03, 0.3, 2.5)` | yes |
| `anthropic` | `claude-fable-5` | Claude Fable 5 | 1,000,000 | `usd(1, 12.5, 50)` | yes |
| `anthropic` | `claude-sonnet-5` | Claude Sonnet 5 | 1,000,000 | `usd(0.2, 2.5, 10)` | yes |
| `moonshot` | `kimi-k3` | Kimi K3 | 1,048,576 | `cny(2, 20, 100)` | yes |
| `openrouter` | `google/gemini-3.6-flash` | Gemini 3.6 Flash | 1,048,576 | `usd(0.15, 1.5, 7.5)` | yes |
| `openrouter` | `google/gemini-3.5-flash-lite` | Gemini 3.5 Flash-Lite | 1,048,576 | `usd(0.03, 0.3, 2.5)` | yes |
| `openrouter` | `moonshotai/kimi-k2.6` | Kimi K2.6 | 262,144 | `usd(0.144, 0.684, 3.42)` | yes |
| `openrouter` | `qwen/qwen3.6-35b-a3b` | Qwen 3.6 35B A3B | 262,144 | `usd(0.14, 0.14, 1)` | yes |
| `openrouter` | `z-ai/glm-5.1` | GLM-5.1 | 204,800 | `usd(0.1794, 0.966, 3.036)` | no |
| `siliconflow` | `Pro/moonshotai/Kimi-K2.6` | Kimi K2.6 | 262,144 | *(none — see below)* | yes |
| `siliconflow` | `Pro/zai-org/GLM-5.1` | GLM-5.1 | 200,000 | *(none — see below)* | no |
| `siliconflow` | `Qwen/Qwen3.6-35B-A3B` | Qwen 3.6 35B A3B | 262,144 | *(none — see below)* | yes |

`vision` is `input_modalities.includes("Image")` from the registry, never a guess — that is why `z-ai/glm-5.1` and `Pro/zai-org/GLM-5.1` are text-only while the Qwen 3.6 rows are multimodal.

### Judgement calls

- **Anthropic keeps `cache_write = 1.25 x input`.** The registry's `prompt_tokens` maps to `cache_write` in every other group, but the `anthropic` block documents and uses Anthropic's 1.25x cache-write surcharge (`claude-opus-4-8` stores `usd(0.5, 6.25, 25)` against a registry input of 5). The two new Claude rows follow the neighbours: input 10 -> 12.5 and input 2 -> 2.5. Pinned by a test.
- **Moonshot stays on `cny()`.** The `moonshot` block stores official CNY prices. `cny(2, 20, 100)` divides to exactly the registry's `0.285714 / 2.857143 / 14.285714`, and matches the `kimi/kimi-k3` row already under Qwen Pay-As-You-Go.
- **`clientType`** follows each block: unset for the auto-routed direct-vendor groups (`google`, `anthropic`, `moonshot`), `"openai"` plus the preset gateway base URL for `openrouter` and `siliconflow`. A test now asserts the direct-vendor groups carry neither `clientType` nor `baseUrl`.
- **`qwen/qwen3.6-35b-a3b` has no published cache price** in either the registry or the OpenRouter API, so `cache_read` repeats the input price — the block's documented fallback.
- **Three SiliconFlow rows ship unpriced.** The registry publishes no pricing for them, `https://api.siliconflow.cn/v1/models` returns 401 without a token, and the console model page is client-rendered with no price in the HTML. No number could be sourced, so rather than invent one the entries omit `pricing`, exactly like `qwen3.8-max-preview`. Their cost reads as 0 until a published rate exists; the code comment and the test's `UNPRICED` set both say so.
- **Ordering** follows the documented rule (dictionary order, newer versions of a series first). SiliconFlow's dictionary order is case-insensitive — `Pro/` and `Qwen/` sort between `moonshotai/` and `zai-org/`, matching how `ZHIPU/GLM-5.2` sorts last under Qwen Pay-As-You-Go. The ordered-id tests pin all of it.
- **No duplicates introduced.** The pre-existing cross-group repeats are `deepseek-v4-pro`, `glm-5.2`, `qwen3.7-max` and `qwen3.7-plus`; none of the 11 new ids collides within its own group, which is the catalog's real key.
- **Display names are shared across groups** for the same model (`Kimi K2.6` under `moonshot`, `openrouter` and `siliconflow`; `GLM-5.1` under `zhipu`, `openrouter` and `siliconflow`). This also renamed the OpenRouter Gemini Lite row from "Gemini 3.5 Flash Lite" to **"Gemini 3.5 Flash-Lite"**, matching the existing `gemini-3.1-flash-lite` entry. A test pins the cross-group equality.

### Excluded: non-chat models

Six registry rows are deliberately not added — `gemini-3.1-flash-image`, `gemini-3.1-flash-tts-preview`, `gemini-embedding-2`, `text-embedding-3-large`, `qwen/qwen3-embedding-4b`, `Qwen/Qwen3-Embedding-8B`. This is not a new decision: the catalog's own header comment already scopes it to chat models ("excludes ... non-chat models (embedding / image generation / TTS)"), and the catalog drives chat-model selection and cost reporting. Detected mechanically as `output_modalities != ["Text"]`, so nothing was judged by name.

### Also fixed

- **`cache_read` now stores the published cache-hit price** for the Gemini rows (review finding). Upstream does publish one — $0.15 and $0.03, agreed by the OpenRouter models API and the 0.4.1 registry — and `cache_read` is billed as its own bucket in the cost center, so repeating the input price would overstate cache-heavy spend 10x and "sync presets" would push that into existing Projects. The direct-vendor rows already set this precedent.
- **`google/gemini-3.5-flash`'s context window: 1000000 -> 1048576**, matching the direct-vendor row for the same model and both upstream sources.
- **`resolveModelEnv` learned `kimi-k3`.** It mirrors AgentHub's routing but was written against v0.3.3; `kimi-k3` matches none of the `kimi-k2.x` substrings, so the id resolved to no env pair at all. AgentHub 0.4.1's `KimiK3Client` reads `MOONSHOT_API_KEY` / `MOONSHOT_BASE_URL` (verified in `dist/kimi_k3/client.js`), so the new branch returns the same pair. Without it the catalog-invariant test would fail on the new `kimi-k3` row.

### Follow-up (deliberately not in this PR)

14 pre-existing OpenRouter rows disagree with the 0.4.1 registry — all of them added before this branch, none touched here. The full list, produced by a scripted comparison:

- `cache_read` still repeats the input price where a real cache price exists: `anthropic/claude-fable-5` (10 vs 1), `anthropic/claude-opus-4.8` and `4.7` (5 vs 0.5), `anthropic/claude-sonnet-5` (2 vs 0.2), `moonshotai/kimi-k3` (3 vs 0.3), `openai/gpt-5.6-sol` (5 vs 0.5), `openai/gpt-5.6-terra` (2.5 vs 0.25), `openai/gpt-5.5` (5 vs 0.5), `x-ai/grok-4.5` (2 vs 0.3), `deepseek/deepseek-v4-pro` (0.435 vs 0.003625).
- Stale context windows: `deepseek/deepseek-v4-flash` and `-pro` (1000000 vs 1048576), `openai/gpt-5.6-sol` / `-terra` / `gpt-5.5` (1000000 vs 1050000), `stepfun/step-3.7-flash` (256000 vs 262144), `xiaomi/mimo-v2.5` (1048576 vs 1050000), `z-ai/glm-5.2` (1000000 vs 1048576), `moonshotai/kimi-k3` (1000000 vs 1048576).
- Full price drift: `deepseek/deepseek-v4-flash` (0.09/0.09/0.18 vs 0.0196/0.098/0.196) and `z-ai/glm-5.2` (0.93/0.93/3 vs 0.15236/0.8204/2.5784).

Correcting those changes money reporting for models people are already using, so it deserves its own reviewed pass rather than riding along here. The block comment now records the split and says to treat those `cache_read` values as an upper bound until then.

## Skill: `agenthub-models` (version 7 -> 8, `updated: 2026-07-22T00:00:00Z`)

- New **Supported-model registry** section: `listSupportedModels`, the `SupportedModel` shape, the usage-bucket pricing keys and the USD/CNY switch, with a note that the registry is narrower than the routing rules (older ids like `gpt-5.4` still route but are no longer listed).
- New **Config parameters the model may reject** section: `UnsupportedParameterError`, which clients reject `temperature` (Gemini 3.6 any value; GPT-5.5, Claude 4.8/5, DeepSeek V4, Kimi K2.6/K3 only `1.0`), the per-family `tool_choice` limits, and `prompt_caching` needing `ENABLE`.
- Model-ID table gains `Gemini 3.6`, `Kimi K3`, `GLM 5.2` and `Qwen 3.6` rows, splits Claude into a `Claude 5` row (`claude-fable-5`, `claude-sonnet-5`) and a `Claude 4` row, and fills in gateway variants; the image id is corrected to `gemini-3.1-flash-image` in the table and the example.
- Routing section lists the real 0.4.1 branch order, including the `gemini-3.6` precedence rule and a warning that a Gemini id on an OpenAI-compatible gateway must pass `clientType: "openai"` explicitly.

## README: curated to the latest generation per family

The tables were drifting toward a mirror of the catalog. They are now explicitly a shortlist, with the full preset list left to the app.

**Rule applied uniformly:** one row per vendor family, naming the newest generation the catalog carries. Where that generation ships several tiers the row names the generation; a single-model generation keeps its own name.

| Model | Providers |
| --- | --- |
| DeepSeek V4 | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan |
| Kimi K3 | Moonshot AI, OpenRouter, Qwen Pay-As-You-Go |
| GLM 5.2 | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
| Hunyuan 3 | OpenRouter |
| Qwen 3.8 Max | Qwen Token Plan (preview) |
| GPT 5.6 | OpenRouter |
| Gemini 3.6 Flash | Google Gemini, OpenRouter |
| Claude 5 | Anthropic, OpenRouter |

Consequences of applying the rule to rows nobody named:

- `Kimi K2.6` and `Gemini 3.5 Flash` dropped, as requested. `Gemini 3.6 Flash` now carries the Gemini row and lists both Google Gemini and OpenRouter, which it only can because of the catalog additions above.
- `GPT 5.5` -> **`GPT 5.6`**. The newest GPT generation in the catalog is `openai/gpt-5.6-sol` / `-terra`, which exist only on OpenRouter, so the providers column honestly reads "OpenRouter" and the direct OpenAI group (`gpt-5.5`, `gpt-5.5-pro`, `gpt-5.4*`) is no longer visible in the README. That is the rule working as intended — the app's Models page is where the full picture lives.
- `Claude Opus 4.8` -> **`Claude 5`**, the multi-tier generation name (`claude-fable-5` + `claude-sonnet-5`), available on both Anthropic and OpenRouter.
- Families the table never listed (MiniMax, Nemotron, Step, Grok, MiMo, LongCat) stay out — this is a shortlist, not an index.

Every providers column was regenerated from `MODEL_CATALOG` after the additions, so e.g. Kimi K3 now correctly shows the Moonshot official endpoint alongside OpenRouter and Qwen Pay-As-You-Go.

The sentence under each table now points at the app's **Models** page for the complete preset list, keeping the existing "any OpenAI-protocol endpoint / 1000+ online and local models" point. Both READMEs stay structurally 1:1; the first column returns to its original 16-character width, and since `*.md` is in `.prettierignore` the padding is by hand.

**Other places that list models, checked:**

- The landing site does **not** enumerate supported models — `packages/landing/src/lib/benchmark-data.ts` names the three models used in a published benchmark comparison (historical measurements, not a support list) and `strings.ts` carries a one-line news banner. Neither is a catalog mirror, so neither follows this curation and neither was touched.
- `packages/landing/content/blog/introducing-penguinharness.{en,zh}.md` does carry a copy of the old table, but it is a dated launch announcement and is deliberately left unchanged.
- `packages/docs/content/models.{en,zh}.md` lists a handful of ids explicitly labelled "not exhaustive"; still accurate, so unchanged.

## Lockfile

Minimal: `pnpm install` (no `--frozen-lockfile`) touched only the four agenthub lines — the two importer specifier/version pairs, the resolution integrity hash and the snapshot key. No transitive dependency moved.

## Verification

- `pnpm -r build` — all 7 packages OK
- `pnpm typecheck` — all 7 packages OK
- `pnpm test` — **1260 passed, 2 skipped, 0 failed** (skills 16, docs 11, landing 26, core 407 + 2 skipped, server 302, cli 121, web 377). Core gained one case, a new `direct-vendor groups` test; the existing catalog tests were extended in place (ordered-id arrays for `openrouter` and `siliconflow`, the `UNPRICED` set, the Kimi K3 CNY round-trip, the Anthropic 1.25x convention, cross-group display-name equality, and `resolveModelEnv` for `kimi-k3` / `gemini-3.6-flash` / `gemini-3.5-flash-lite` / the Claude 5 ids).
- Playwright e2e not run: no web behaviour changed.

Changelog: covered by PR #30 (commit `a7a9271`) — its Models/core entry covers the 0.4.1 bump and the catalog rows, its Skills entry covers the `agenthub-models` v7 -> v8 rewrite, and both index lines are updated. This PR deliberately adds no changelog file of its own.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,15 +82,17 @@ Four Skill groups ship in the box ([docs](https://penguin.ooo/docs/skills)); Age
 | Model                 | Providers                                                                        |
 | --------------------- | -------------------------------------------------------------------------------- |
 | DeepSeek V4           | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
-| Kimi K3               | OpenRouter, Qwen Pay-As-You-Go                                                   |
-| Kimi K2.6             | Moonshot AI                                                                      |
+| Kimi K3               | Moonshot AI, OpenRouter, Qwen Pay-As-You-Go                                      |
+| Kimi K2.6             | Moonshot AI, OpenRouter, SiliconFlow                                             |
 | GLM 5.2               | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
 | Hunyuan 3             | OpenRouter                                                                       |
 | Qwen 3.8 Max          | Qwen Token Plan (preview)                                                        |
 | GPT 5.5               | OpenAI, OpenRouter                                                               |
-| Gemini 3.6 Flash      | OpenRouter                                                                       |
+| Gemini 3.6 Flash      | Google Gemini, OpenRouter                                                        |
 | Gemini 3.5 Flash      | Google Gemini, OpenRouter                                                        |
-| Gemini 3.5 Flash Lite | OpenRouter                                                                       |
+| Gemini 3.5 Flash-Lite | Google Gemini, OpenRouter                                                        |
+| Claude Fable 5        | Anthropic, OpenRouter                                                            |
+| Claude Sonnet 5       | Anthropic, OpenRouter                                                            |
 | Claude Opus 4.8       | Anthropic, OpenRouter                                                            |
 
 Any OpenAI-protocol endpoint is supported: pick a preset above, or point a custom endpoint at any of the 1000+ online and local models.
--- a/README.md
+++ b/README.md
@@ -79,17 +79,19 @@ Four Skill groups ship in the box ([docs](https://penguin.ooo/docs/skills)); Age
 
 ## Supported Models
 
-| Model            | Providers                                                                        |
-| ---------------- | -------------------------------------------------------------------------------- |
-| DeepSeek V4      | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
-| Kimi K3          | OpenRouter, Qwen Pay-As-You-Go                                                   |
-| Kimi K2.6        | Moonshot AI                                                                      |
-| GLM 5.2          | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
-| Hunyuan 3        | OpenRouter                                                                       |
-| Qwen 3.8 Max     | Qwen Token Plan (preview)                                                        |
-| GPT 5.5          | OpenAI, OpenRouter                                                               |
-| Gemini 3.5 Flash | Google Gemini, OpenRouter                                                        |
-| Claude Opus 4.8  | Anthropic, OpenRouter                                                            |
+| Model                 | Providers                                                                        |
+| --------------------- | -------------------------------------------------------------------------------- |
+| DeepSeek V4           | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
+| Kimi K3               | OpenRouter, Qwen Pay-As-You-Go                                                   |
+| Kimi K2.6             | Moonshot AI                                                                      |
+| GLM 5.2               | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
+| Hunyuan 3             | OpenRouter                                                                       |
+| Qwen 3.8 Max          | Qwen Token Plan (preview)                                                        |
+| GPT 5.5               | OpenAI, OpenRouter                                                               |
+| Gemini 3.6 Flash      | OpenRouter                                                                       |
+| Gemini 3.5 Flash      | Google Gemini, OpenRouter                                                        |
+| Gemini 3.5 Flash Lite | OpenRouter                                                                       |
+| Claude Opus 4.8       | Anthropic, OpenRouter                                                            |
 
 Any OpenAI-protocol endpoint is supported: pick a preset above, or point a custom endpoint at any of the 1000+ online and local models.
 
--- a/README.md
+++ b/README.md
@@ -79,23 +79,18 @@ Four Skill groups ship in the box ([docs](https://penguin.ooo/docs/skills)); Age
 
 ## Supported Models
 
-| Model                 | Providers                                                                        |
-| --------------------- | -------------------------------------------------------------------------------- |
-| DeepSeek V4           | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
-| Kimi K3               | Moonshot AI, OpenRouter, Qwen Pay-As-You-Go                                      |
-| Kimi K2.6             | Moonshot AI, OpenRouter, SiliconFlow                                             |
-| GLM 5.2               | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
-| Hunyuan 3             | OpenRouter                                                                       |
-| Qwen 3.8 Max          | Qwen Token Plan (preview)                                                        |
-| GPT 5.5               | OpenAI, OpenRouter                                                               |
-| Gemini 3.6 Flash      | Google Gemini, OpenRouter                                                        |
-| Gemini 3.5 Flash      | Google Gemini, OpenRouter                                                        |
-| Gemini 3.5 Flash-Lite | Google Gemini, OpenRouter                                                        |
-| Claude Fable 5        | Anthropic, OpenRouter                                                            |
-| Claude Sonnet 5       | Anthropic, OpenRouter                                                            |
-| Claude Opus 4.8       | Anthropic, OpenRouter                                                            |
+| Model            | Providers                                                                        |
+| ---------------- | -------------------------------------------------------------------------------- |
+| DeepSeek V4      | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
+| Kimi K3          | Moonshot AI, OpenRouter, Qwen Pay-As-You-Go                                      |
+| GLM 5.2          | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
+| Hunyuan 3        | OpenRouter                                                                       |
+| Qwen 3.8 Max     | Qwen Token Plan (preview)                                                        |
+| GPT 5.6          | OpenRouter                                                                       |
+| Gemini 3.6 Flash | Google Gemini, OpenRouter                                                        |
+| Claude 5         | Anthropic, OpenRouter                                                            |
 
-Any OpenAI-protocol endpoint is supported: pick a preset above, or point a custom endpoint at any of the 1000+ online and local models.
+Each family's latest generation only — the app's **Models** page lists every built-in preset, and any OpenAI-protocol endpoint works too: pick a preset, or point a custom endpoint at any of the 1000+ online and local models.
 
 ## Requirements
 
--- a/README.zh.md
+++ b/README.zh.md
@@ -79,17 +79,19 @@ https://github.com/user-attachments/assets/aec49ae9-b743-467b-b247-37bedfeaa36e
 
 ## 支持的模型
 
-| 模型             | 可用供应商                                                                       |
-| ---------------- | -------------------------------------------------------------------------------- |
-| DeepSeek V4      | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
-| Kimi K3          | OpenRouter, Qwen Pay-As-You-Go                                                   |
-| Kimi K2.6        | Moonshot AI                                                                      |
-| GLM 5.2          | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
-| Hunyuan 3        | OpenRouter                                                                       |
-| Qwen 3.8 Max     | Qwen Token Plan（预览）                                                          |
-| GPT 5.5          | OpenAI, OpenRouter                                                               |
-| Gemini 3.5 Flash | Google Gemini, OpenRouter                                                        |
-| Claude Opus 4.8  | Anthropic, OpenRouter                                                            |
+| 模型                  | 可用供应商                                                                       |
+| --------------------- | -------------------------------------------------------------------------------- |
+| DeepSeek V4           | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
+| Kimi K3               | OpenRouter, Qwen Pay-As-You-Go                                                   |
+| Kimi K2.6             | Moonshot AI                                                                      |
+| GLM 5.2               | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
+| Hunyuan 3             | OpenRouter                                                                       |
+| Qwen 3.8 Max          | Qwen Token Plan（预览）                                                          |
+| GPT 5.5               | OpenAI, OpenRouter                                                               |
+| Gemini 3.6 Flash      | OpenRouter                                                                       |
+| Gemini 3.5 Flash      | Google Gemini, OpenRouter                                                        |
+| Gemini 3.5 Flash Lite | OpenRouter                                                                       |
+| Claude Opus 4.8       | Anthropic, OpenRouter                                                            |
 
 只要是 OpenAI 协议的端点都可以接入：从上表选择预置，或用自定义端点连接 1000+ 在线与本地模型。
 
--- a/README.zh.md
+++ b/README.zh.md
@@ -82,15 +82,17 @@ https://github.com/user-attachments/assets/aec49ae9-b743-467b-b247-37bedfeaa36e
 | 模型                  | 可用供应商                                                                       |
 | --------------------- | -------------------------------------------------------------------------------- |
 | DeepSeek V4           | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
-| Kimi K3               | OpenRouter, Qwen Pay-As-You-Go                                                   |
-| Kimi K2.6             | Moonshot AI                                                                      |
+| Kimi K3               | Moonshot AI, OpenRouter, Qwen Pay-As-You-Go                                      |
+| Kimi K2.6             | Moonshot AI, OpenRouter, SiliconFlow                                             |
 | GLM 5.2               | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
 | Hunyuan 3             | OpenRouter                                                                       |
 | Qwen 3.8 Max          | Qwen Token Plan（预览）                                                          |
 | GPT 5.5               | OpenAI, OpenRouter                                                               |
-| Gemini 3.6 Flash      | OpenRouter                                                                       |
+| Gemini 3.6 Flash      | Google Gemini, OpenRouter                                                        |
 | Gemini 3.5 Flash      | Google Gemini, OpenRouter                                                        |
-| Gemini 3.5 Flash Lite | OpenRouter                                                                       |
+| Gemini 3.5 Flash-Lite | Google Gemini, OpenRouter                                                        |
+| Claude Fable 5        | Anthropic, OpenRouter                                                            |
+| Claude Sonnet 5       | Anthropic, OpenRouter                                                            |
 | Claude Opus 4.8       | Anthropic, OpenRouter                                                            |
 
 只要是 OpenAI 协议的端点都可以接入：从上表选择预置，或用自定义端点连接 1000+ 在线与本地模型。
--- a/README.zh.md
+++ b/README.zh.md
@@ -79,23 +79,18 @@ https://github.com/user-attachments/assets/aec49ae9-b743-467b-b247-37bedfeaa36e
 
 ## 支持的模型
 
-| 模型                  | 可用供应商                                                                       |
-| --------------------- | -------------------------------------------------------------------------------- |
-| DeepSeek V4           | DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan                 |
-| Kimi K3               | Moonshot AI, OpenRouter, Qwen Pay-As-You-Go                                      |
-| Kimi K2.6             | Moonshot AI, OpenRouter, SiliconFlow                                             |
-| GLM 5.2               | Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go |
-| Hunyuan 3             | OpenRouter                                                                       |
-| Qwen 3.8 Max          | Qwen Token Plan（预览）                                                          |
-| GPT 5.5               | OpenAI, OpenRouter                                                               |
-| Gemini 3.6 Flash      | Google Gemini, OpenRouter                                                        |
-| Gemini 3.5 Flash      | Google Gemini, OpenRouter                                                        |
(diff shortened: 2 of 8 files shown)


hiyouga (commented):
**Verdict: needs-changes** — the catalog work is accurate and well-ordered, and the lockfile/README/skill mechanics all check out, but the stated reason for skipping the changelog is false, and the new rows store a cache price the PR body itself shows to be 10× the published one behind a comment asserting no such price exists.

### Findings

**[medium] The PR body's justification for omitting a changelog entry — "No changelog entry added — PR #30 aggregates them" — does not hold.**
Checked all seven files on `origin/docs/changelog-0.2.0`: zero mentions of `agenthub`, `0.4.1`, `gemini-3.6` or `flash-lite` (the single `agenthub` hit in `2026-07-22-llm-requests.md` is about the empty-tools fix). #30 was opened at 10:55 today and aggregates *already-merged* work; this branch was created at 17:24 and is not in it. CLAUDE.md requires an entry for every update, and this one touches two surfaces (Models catalog + skills).
*Fix:* once #30 lands `changelog/0.2.0/`, extend the Models entry with the AgentHub 0.4.1 bump plus the two new catalog rows, extend `2026-07-22-skills.md` with the `agenthub-models` v7→v8 rewrite, and refresh both README index lines.

**[medium] `packages/core/src/state/model-catalog.ts:286-287`, `:308` — `cache_read` is stored at 10× the published cache price, behind a comment that states no cache price exists.**
Both comments read "the page publishes no cache price, so cache_read carries the standard input price", but the PR body states that the OpenRouter models API *and* AgentHub 0.4.1's own registry both publish a cache-hit price — $0.15 for 3.6 Flash and $0.03 for 3.5 Flash Lite. So `usd(1.5, 1.5, 7.5)` and `usd(0.3, 0.3, 2.5)` overstate cached-input cost by exactly 10× in both rows. `cache_read` is billed as its own bucket in the cost center, so for a cache-heavy workload — the common case once a system prompt is stable — reported spend is materially inflated, and "sync presets" will push the same number into existing Projects.

The consistency argument is defensible (the neighbouring `google/gemini-3.5-flash` at `:303` stores `usd(1.5, 1.5, 9)` under the same convention, and changing the whole OpenRouter block is rightly a separate PR). The comment is the real problem: it asserts a fact the PR body contradicts and will convince the next reader no cache price is available upstream. Note the direct-vendor row at `:635` already stores the real discounted price, `usd(0.15, 1.5, 9)` — so the two conventions coexist in one file with nothing explaining why.
*Fix:* reword both comments to say a cache price *is* published upstream but the OpenRouter block uniformly stores `cache_read` = input price pending a block-wide correction — or just store the real $0.15 / $0.03 and accept the local inconsistency.

**[low] `design/specs/08-CHANGELOG.md` gets no line, breaking the established pattern for catalog changes.**
Lines 33, 34, 35 and 43 each record a MODEL_CATALOG change at this granularity — new gateway groups with their per-entry pricing and context windows, and the dictionary-order re-sort whose rule this PR follows. A two-entry addition plus a core dependency bump is the same class of change and leaves no trace in the specs. (`design/` is a gitignored symlink to `penguin-harness-design`, so this cannot ride this branch.)
*Fix:* add a dated line covering the 0.4.1 bump, the two OpenRouter Gemini rows and the skill rewrite.

**[low] `packages/core/src/state/model-catalog.ts:290`, `:302`, `:312` — the new rows disagree with the sibling row between them on context window.**
After this PR the OpenRouter Gemini block reads 1048576 (3.6 Flash) / **1000000** (3.5 Flash) / 1048576 (3.5 Flash Lite), while the direct-vendor `gemini-3.5-flash` at `:638` uses 1048576 for the same model. The PR body notes AgentHub 0.4.1's registry lists 1,048,576 for both new ids, which suggests the existing 1000000 is the stale value rather than a real gateway difference. Not introduced here, but this change places the discrepancy on adjacent lines.
*Fix:* either correct `google/gemini-3.5-flash` to 1048576 in passing, or add a comment recording why the gateway row differs.

### Notes

- **Verified clean.** Ordering follows the documented rule (`model-catalog.ts:198-202`): newer series first, then tiers alphabetically → `3.6-flash`, `3.5-flash`, `3.5-flash-lite`, with the vendor prefix keeping the block between `deepseek/*` and `minimax/*`; the test's ordered id array pins it. The lockfile diff is exactly the four claimed agenthub lines (two importer pairs, the integrity hash, the snapshot key) with no transitive movement. Skill frontmatter is correctly `version: 7 → 8` with `updated` bumped, and its `description` was extended for the new registry section. Both README tables were re-padded consistently — every row's second pipe lands at the same column in both files — and stay 1:1.
- **The `UnsupportedParameterError` unreachability claim holds against this repo's code.** `buildUniConfig` (`generative-model.ts:1108-1122`) sets only `tools`, `system_prompt`, `max_tokens` and `thinking_level`; a repo-wide grep finds no `temperature`, `tool_choice` or `prompt_caching` anywhere in `packages/core/src` outside a comment at `:1104` saying as much. So the new typed error genuinely cannot fire from PenguinHarness today, including on Gemini 3.6.
- **The launch blog post was correctly left alone.** `packages/landing/content/blog/introducing-penguinharness.{en,zh}.md:84` carries a copy of the same supported-models table, but it is a dated announcement (`date: 2026-07-17`, `category: news`) and is already divergent from the README on the Kimi rows — it reads as a historical record, not a live list. The docs site does not enumerate individual models, so nothing else needed syncing.
- **What I could not verify.** `@prismshadow/agenthub` is not present in this checkout's `node_modules`, so the 0.4.1 API claims that the skill now documents — `listSupportedModels`, the `SupportedModel` shape and its pricing keys, `UnsupportedParameterError`, and the `gemini3_6` / `glm5_2` / `kimi_k3` clients and their branch order — rest entirely on the author's tarball diff and were not independently checked here. Since the skill is what agents read to write AgentHub code, a wrong signature there surfaces as broken generated code rather than a failing test; worth a second pair of eyes on that section specifically.

rank-Yu (approved):

hiyouga:
Thanks — all four addressed. Pushed as `9d79243`.

### 1. [medium] `cache_read` 10x the published price behind a comment denying the price exists — **fixed**

You were right, and the comment was the worse half of the bug: it asserted the page lists no cache price while the OpenRouter models API and AgentHub 0.4.1's own registry both publish one.

Re-verified both sources before writing anything; they agree exactly, no disagreement to report:

| | context | cache hit | input | output |
| `google/gemini-3.6-flash` | 1,048,576 | **$0.15** | $1.50 | $7.50 |
| `google/gemini-3.5-flash-lite` | 1,048,576 | **$0.03** | $0.30 | $2.50 |

Source A — `https://openrouter.ai/api/v1/models`, `pricing.input_cache_read` / `prompt` / `completion` and `context_length`. Source B — `listSupportedModels()` from the installed 0.4.1 package, `cached_tokens` / `prompt_tokens` / `response_tokens` and `context_window`, for the same models on the official Google endpoint.

Now stored as `usd(0.15, 1.5, 7.5)` and `usd(0.03, 0.3, 2.5)`. `cache_write` keeps the input price: upstream's `input_cache_write` field is a Gemini cache-*storage* rate, not a per-token write fee, and the direct-vendor rows already resolve it the same way.

Rationale, recorded in the code comment as well as the PR body: `cache_read` is billed as its own bucket in the cost center and "sync presets" pushes it into existing Projects, so accuracy of money reporting beats block-local consistency — and as you noted, `:635` already set the precedent.

The false comment is gone. In its place: the entry comment says the upstream cache price is published and stored, and the block header now records that two conventions coexist, that several 2026-07-20 rows likely have published cache prices too, and that their `cache_read` should be treated as an upper bound until a block-wide re-read. **Flagged as a follow-up in the PR body rather than done here** — `google/gemini-3.5-flash` still stores `usd(1.5, 1.5, 9)` against a published $0.15, and correcting that block deserves its own deliberate pass.

### 2. [low] Context-window disagreement on adjacent lines — **fixed**

`google/gemini-3.5-flash` corrected from 1000000 to **1048576**, matching the direct-vendor row for the same model and both upstream sources above. The gateway row was simply stale. Locked in by a test assertion that the two rows for that model now agree.

### 3. [medium] Changelog — **already covered, no change here**

Correct that #30 didn't cover this branch at review time; it does now. #30 was extended in commit `a7a9271`: the Models/core entry covers the 0.4.1 bump and both catalog rows, the Skills entry covers the `agenthub-models` v7 -> v8 rewrite, and both index lines are updated. This PR deliberately adds no changelog file of its own — #30 aggregates them.

### 4. [low] `design/specs/08-CHANGELOG.md` — **handled in the design repo**, outside this worktree.

### On "What I could not verify"

Fair — none of the 0.4.1 claims were retraceable without `node_modules`. The PR body now names the evidence per claim; the short version, reproducible from a clean checkout:

```bash
npm pack @prismshadow/agenthub@0.4.0 @prismshadow/agenthub@0.4.1   # then untar both
diff -rq v040/package/dist v041/package/dist
node -e 'console.log(require("@prismshadow/agenthub").listSupportedModels().length)'   # 49
```

- **"Type-compatible with 0.4.0"** — `diff -rq` reports `dist/types.d.ts`, `dist/autoClient.d.ts` and `dist/baseClient.d.ts` identical between the tarballs; `pnpm typecheck` passed on this repo before any source change.
- **`listSupportedModels` shape and semantics** — signature and doc comments from `dist/registry.d.ts` (a "Only in v041" file in the diff); the 49-entry count, the modality vocabulary, the 7 CNY/USD conversion and every price quoted come from actually running it.
- **`UnsupportedParameterError`** — `dist/errors.d.ts` and `dist/index.d.ts` diffs show the class and its export are the only additions; the per-client rejection rules in the skill come from grepping every `dist/*/client.js` for `UnsupportedParameterError` and reading each guard condition (that is where "Gemini 3.6 rejects any value, the others accept only `1.0`" comes from — the guards are `!== undefined` vs `!== undefined && !== 1.0`).
- **Three new clients and the routing precedence** — `dist/gemini3_6/`, `dist/glm5_2/`, `dist/kimi_k3/` are "Only in v041"; the ordering rule (`gemini-3.6` matched before the broader `gemini-3` prefix) is read from the `dist/autoClient.js` diff.
- **Image id rename** — from the id `listSupportedModels()` returns. One caveat worth recording: `dist/integration/playground.js` still hardcodes the old `-preview` id, unchanged from 0.4.0. Upstream inconsistency; nothing here depends on it.

The npm tarball ships only `dist/` and `README.md` — there is no CHANGELOG inside the package, so the repo changelog and the release PR were used for orientation only, never as the source of an API claim.

### Validation after the fix

`pnpm -r build` 7/7 · `pnpm typecheck` 7/7 · `pnpm test` **1259 passed, 2 skipped, 0 failed** (skills 16, docs 11, landing 26, core 406 + 2 skipped, server 302, cli 121, web 377) · `pnpm format` then `pnpm format:check` clean. No web behaviour changed, so Playwright e2e was not run.

PR #46, opened 2026-07-22
feat(server,web): serve Workspace HTML previews from a separate origin

hiyouga:
Fixes the `SecurityError` on "open in new tab" for Workspace HTML, and enables third-party embeds.

Design PR: **Prism-Shadow/penguin-harness-design#3** (spec added first, per the repo rule).

## The bug

Previews were served from the app's own origin under `sandbox` without `allow-same-origin`. That correctly kept Agent-generated markup away from the session cookie and the API, but it also meant an opaque origin, so `localStorage` and `document.cookie` threw and every embedded widget died.

While designing the fix I found a second, quieter bug: the preview URL was a query-parameter endpoint, so a page's relative subresources resolved to `/api/sessions/<id>/files/app.js` — a route that does not exist. **Multi-file generated apps could never load their own CSS or JS in preview.** The path-based URL fixes that too.

## The fix

`GET /files/preview-redirect?path=` authenticates, mints a short-lived HMAC token bound to (Session, preview host, expiry), and 302s to `/preview/<token>/<path>` on a **different origin**:

- **Locally**: the loopback counterpart of whatever you browse to (`127.0.0.1` ↔ `localhost`). Cookies are keyed by host and ignore port, so those are real separate cookie jars — a second port would **not** have worked.
- **Deployments**: `PENGUIN_PREVIEW_ORIGIN`.
- **Neither**: falls back to the old sandbox, and `previewIsolated` on `/api/me` reports `false` so the Files panel marks the link rather than letting the page break silently.

## Why it stays safe without the sandbox

- **The token binds the preview host, and `/preview/...` refuses to serve on the app origin.** The same process answers there; without this check it would be a same-origin XSS with full API access. Covered by a test.
- **`Referrer-Policy: no-referrer`** — otherwise the token-bearing URL leaks via `Referer` to exactly the third parties this change enables.
- Path is still re-resolved against the Workspace; `..` and symlink escapes rejected as before. Bad/expired token, wrong host and out-of-bounds path all answer a bare 404.
- The link is a plain anchor with `rel="noopener noreferrer"`, not `fetch` + `window.open` — the latter trips popup blockers and leaves an `opener` handle back to the app, the very reference this design denies.

## Verification

14 new server tests, plus an end-to-end run against a **real** server (not just `app.request()`), confirming: redirect lands on the counterpart origin, page served with no CSP and `no-referrer`, relative subresource loads `200`, and the same token on the app origin returns `404`.

Full suite green: **1,277 tests**, typecheck and format clean.

Also binds the IPv6 loopback alongside IPv4 — `localhost` commonly resolves to `::1` first, and every preview URL would otherwise refuse the connection. Smoke-tested on `127.0.0.1`, `localhost` and `[::1]`.

Docs updated bilingually (`server-api`, `configuration`) and a changelog section added.

https://claude.ai/code/session_01CxXYAeprZaqX1GJKUH1cpQ


hiyouga:
## Code review

Read the security-critical source, the design spec (`05-ARCHITECTURE.md §Workspace 文件预览`), `WorkspaceFilesService`, and `auth/middleware.ts`; validated the token logic and ran the tests' path logic by hand.

This is careful work — the `no-referrer` header, `rel="noopener noreferrer"` anchor, GET-mint reasoning, IPv6 companion listener, the `previewIsolated` degrade signal, and the reuse of the TOCTOU-safe `read()` confinement are all well-considered. One boundary is weaker than the PR/design claims, and deserves a decision before merge.

### 🔴 The "separate origin" still serves `/api`, so isolation rests on an undocumented precondition

Both listeners share the **same `app.fetch`** — the full app, including `/api` and the auth middleware — on the **same port**. So the preview origin is not an isolated file surface; it's the whole application on a sibling hostname.

The token's host check correctly stops the server from serving *Agent HTML on the app origin*. But it does nothing about the reverse: **Agent HTML running on the preview origin can call `/api` on the preview origin same-origin.** With the default `HOST=127.0.0.1`:

- User browses `localhost:7364` → session cookie is host-only for `localhost`.
- Preview counterpart = `127.0.0.1`; preview page origin = `http://127.0.0.1:7364`.
- The page does `fetch("http://127.0.0.1:7364/api/…")` — **same-origin**. SameSite=Lax doesn't apply (not cross-site), and `jsonOnlyWrites` is satisfied by a same-origin `application/json` fetch, so **reads *and* writes succeed** — *if a session cookie exists for `127.0.0.1`*.

`auth/middleware.ts` is cookie-only with **no Origin/Referer check**, so the whole boundary rests on "the user has no session cookie for the preview host." Because the two hosts are `127.0.0.1` and `localhost` — used interchangeably, with `127.0.0.1` the default bind and `localhost` what people type — a user who has logged in on both (a mundane, invisible act) gives any previewed HTML full authenticated API access.

This matters precisely because preview isolation exists to contain *untrusted* content (an agent that ingested a malicious page and echoed markup into a preview file). The design's invariant *"预览 Origin 永不收到 App 的会话 Cookie"* is technically true but misleading — the preview host is a fully functional app origin that can hold its own cookie — and the PR's "can reach neither this origin's cookies nor the API" is only conditionally true.

Local single-user tool, so **moderate, not critical**, but it's a security-hardening PR whose headline claim doesn't fully hold. Options, most→least robust:
1. **Serve only `/preview/*` on the preview surface** (dedicated listener/port that never mounts `/api`) — closes it regardless of cookies.
2. **Canonicalize auth to one loopback name** — redirect the app UI to a single canonical host and only set/accept the session cookie there, reserving the counterpart purely for previews.
3. **Document the precondition** (pick one loopback name; deployments route only `/preview/*` to the app on the preview origin) and soften the "cannot reach the API" claim in spec + PR.

### 🟡 Design-doc sync
`05-ARCHITECTURE.md §签名令牌` lists the token as binding *sessionId / Workspace subtree / expiry* — it omits the **host binding**, which the PR itself calls the load-bearing check. Per the repo rule (design is ground truth, kept in sync), the spec's security model should document the host-binding + app-origin-refusal, and clarify the `/api`-reachability caveat above.

### 🟡 IPv6 companion listener vs. wildcard binds
`index.ts` starts the `::1` listener only for `host ∈ {127.0.0.1, localhost}`, but `resolvePreviewTarget`'s `LOOPBACK_BINDS` treats `0.0.0.0`/`::` as loopback-serving. With `HOST=0.0.0.0` (IPv4-only), a `127.0.0.1`→`localhost` counterpart the browser resolves to `::1` has nothing listening, yet `/api/me` reports `previewIsolated: true` — the tab dies instead of falling back. (Also `HOST=localhost` resolving to `::1` first can land *both* listeners on `::1`, leaving `127.0.0.1` unbound.) Minor, edge deployments.

### 🟡 Path-traversal test asserts a route-miss, not confinement
In `workspace-preview.test.ts`, `…/${token}/../../etc/passwd` is normalized by WHATWG URL parsing *before* the handler, so the path no longer matches `/preview/:token/*` — the 404 is a route miss, not the server-side `..` rejection. A percent-encoded traversal (`%2e%2e/%2e%2e/…`) survives parsing, reaches `decodeURIComponent` → `workspaceFiles.read`, and actually exercises the confinement. Worth adding.

### 🟢 Minor — mint reads the whole file just to validate
`preview-redirect` calls `workspaceFiles.read(...)` only to validate the path, which `fs.readFile`s up to 50 MB into memory and discards it, on every click. A stat-based check (as `statExisting` already does) gives the same validation without reading contents.

hiyouga:
Fixed in 994a61a (design PR #3 updated in Prism-Shadow/penguin-harness-design@c3dd0d3).

**Primary — the `/api`-on-preview-host gap → canonicalized.** The App is now canonicalized onto one loopback name (`localhost`); the counterpart (`127.0.0.1`) is reserved for previews. A guard makes the preview host serve **only** `/preview/*`: its `/api` answers `401` (closing both the login and the stale-cookie paths, so a cookie is never set or honored there) and every App route `302`s to the canonical host. The guard is off when `PENGUIN_PREVIEW_ORIGIN` is set — deployments enforce the equivalent at the reverse proxy (route only `/preview/*` to the App on the preview origin), now documented.

**Also addressed:**
- **Design-sync** — spec now documents the token's load-bearing preview-host binding and the canonical-host rule.
- **Wildcard-bind finding** — loopback previews are limited to loopback-name binds (`127.0.0.1`/`localhost`); a `0.0.0.0`/`::` bind no longer claims a counterpart it can't serve on `::1` while `/api/me` reports isolation.
- **Test coverage** — the traversal test now uses `..%2f..%2f…` (a literal `../` normalizes away before the handler, so the old assertion only exercised a route-miss); added coverage for the guard (`/api` → 401 even authenticated, App route → 302) and the `PENGUIN_PREVIEW_ORIGIN`-off case.
- **Efficiency** — the mint validates the path with a stat instead of reading up to 50 MB of file bytes to discard them.
- **CLI** — startup advertises the canonical host; `waitForReady` probes with `redirect: "manual"` so the guard's 302 counts as ready.

Verification: `pnpm typecheck` clean; server suite **321 passing**, CLI serve tests passing; prettier clean.

<sub>🤖 via Claude Code</sub>

PR #47, opened 2026-07-22
chore(changelog): collect unreleased entries in changelog/unreleased/

hiyouga:
Migrates the tracked changelog docs to the `unreleased/` convention.

## What this changes

Two tracked docs still described the **old** rule — put entries in the upcoming version's numbered folder and rename it if the number changed:

- `changelog/README.md` (the layout doc)
- `CONTRIBUTING.md` (the contributor guide)

That guess isn't knowable when an entry is written: the batch that accumulated as `0.2.0/` shipped as **0.1.1**, so a numbered folder only ever created a rename to get wrong later. Both now describe the `unreleased/` flow — work in progress lives in `changelog/unreleased/`; at release it's renamed to the shipped version, its `# Unreleased` heading becomes `# Version X.Y.Z` + `Released on <date>.`, the root `CHANGELOG.md` line is added, and a fresh `unreleased/` is started.

## Scope note (the original description was stale)

The original write-up predated the 0.1.1 release and no longer matched the diff: there is **no `changelog/0.2.0/` folder to move** (that batch already shipped as `0.1.1/`, renamed at release — exactly the failure this convention prevents), so nothing is renamed here. This is purely the tracked-doc migration.

The `changelog/unreleased/` folder itself is seeded by the first pending entry — **#46** creates it with the `# Unreleased` heading and its first line — so this PR no longer adds an empty `unreleased/README.md` (which would have collided with #46 on merge).

> Merge **before #48**: both touch `changelog/README.md`. There's no longer any contradiction (this PR owns the release-flow paragraph; #48 only adds the `RELEASE.md` bullets), but expect a trivial adjacent-line merge.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -1,10 +1,10 @@
 # Changelog Details
 
-The root [../CHANGELOG.md](../CHANGELOG.md) keeps one brief line per release. Each release owns a folder here:
+The root [../CHANGELOG.md](../CHANGELOG.md) keeps one brief line per release. Work in progress lives in [`unreleased/`](unreleased/); each shipped release owns a numbered folder here:
 
-- `<version>/README.md` — the release summary: one brief line per change, each linking its detail file, e.g. `- [YYYY-MM-DD] Brief description. ([details](YYYY-MM-DD-short-slug.md))`
-- `<version>/YYYY-MM-DD-short-slug.md` — one detail file per entry, named by the entry date: an H1 title, then what changed and why, using `##` sections once the entry covers more than one thing.
+- `<folder>/README.md` — the summary: one brief line per change, each linking its detail file, e.g. `- [YYYY-MM-DD] Brief description. ([details](YYYY-MM-DD-short-slug.md))`
+- `<folder>/YYYY-MM-DD-short-slug.md` — one detail file per entry, named by the entry date: an H1 title, then what changed and why, using `##` sections once the entry covers more than one thing.
 - Entries are grouped by the surface they change — Models, Web App, landing site, skills, docs, tooling — rather than one file per commit. A related change extends the existing file and its summary line instead of opening a new one.
-- Changes not yet released go into the upcoming version's folder; during release preparation, rename the folder if the number changed and add the release line to the root file. Released folders are frozen.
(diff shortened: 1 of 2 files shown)


hiyouga:
## Code review

The change is fine and still needed, but **the description no longer matches the diff** — worth reconciling before merge.

### The description is stale
Checking `origin/main` directly: there is **no `changelog/0.2.0/` folder and no "six pending entries"** (0.1.1 is released, and `AGENTS.md` already documents the `unreleased/` convention). The actual diff is **2 files**: it edits `changelog/README.md` and adds an empty `changelog/unreleased/README.md`. It does **not** rename six entries and does **not** touch `AGENTS.md`, despite the body claiming both.

What almost certainly happened: the `0.2.0/` batch shipped *as* 0.1.1 (renamed at release), overtaking this PR's substantive operation and leaving only the doc fix.

- The `changelog/README.md` fix **is still correct and needed** — `main` still carries the old "upcoming version's folder" rule.
- Please **rewrite the PR description** to match the real diff, so a reviewer isn't reconciling a phantom `0.2.0/` rename.

### Overlaps #46
Both PRs add `changelog/unreleased/README.md` as a *new file* with different content (#46 includes its web-app entry line; this one is empty), so they collide. Given `main` + #46 already establish the `unreleased/` folder, the cleanest path is to **fold this PR's `changelog/README.md` fix into #46** (which already edits that folder) and close this — or land this first and rebase #46 to *append* to the existing README rather than create it.

hiyouga:
Updated per the review:

- **Description rewritten** to match the real diff — the original predated the 0.1.1 release (no `0.2.0/` folder to move). Note `AGENTS.md`/`CLAUDE.md` are **gitignored** local files (`.gitignore:35-36`), not repo docs, so they were never in scope; the tracked contributor doc is `CONTRIBUTING.md`.
- **Completed the migration**: `CONTRIBUTING.md` also described the numbered-folder rule (`changelog/<version>/… under the next unreleased version`) and now points at `changelog/unreleased/` — so the two tracked docs finally agree.
- **Dropped the empty `unreleased/README.md`**: #46 already seeds the folder with the `# Unreleased` heading + its first entry, so two PRs no longer add the same file (order-independent now).

Merge **before #48**. Prettier clean.

PR #48, opened 2026-07-22
ci(release): publish curated release notes, never an empty body

hiyouga:
## The failure

`v0.1.1` was released with a **completely empty Release body** — 12 assets and no text. The release step passed neither `body`/`body_path` nor `generate_release_notes`, so `softprops/action-gh-release@v2` had nothing to write and left the body unset. `v0.1.0` only reads well because a maintainer hand-wrote its notes on the Release page after the fact; for `v0.1.1` that manual step was forgotten. The failure is silent — nothing in the workflow logs flags it, and it is only visible once the Release is public.

## The design

The announcement becomes a repo artifact instead of a post-release chore.

- **`changelog/<version>/RELEASE.md`** holds the curated announcement — the same shape both releases already used: a lead sentence, `## Install`, `## Highlights`, `## Notable in this release`, `## Requirements`, and a closing link to the version's changelog folder. It is written during release preparation, alongside marking the folder `Released on <date>.`
- **A new `Resolve release notes` step** derives the version from the tag (`${TAG#v}`, with the tag resolved the same way as everywhere else in this workflow, so both the `push` tag path and the `workflow_dispatch` `inputs.tag` path work) and selects the body source:
  - present and non-empty → `body_path` points at it, `generate_release_notes: false` → the curated text is published verbatim;
  - absent → `generate_release_notes: true` → GitHub composes a body from the merged PRs.

The fallback is the point. Curated notes are better than generated ones, but the failure mode being fixed is *forgetting* — so the branch that runs when someone forgets has to produce something rather than nothing. Generated notes are a worse release page; an empty one is a broken release page.

The check is `[ -s ]` rather than `[ -f ]`, so a file that exists but is empty also falls back — an empty committed file would otherwise reproduce the exact bug this fixes.

Nothing else moves: the uploaded files, the `check-release` gating, and the npm publish job are untouched.

## What was verified about the action

Read from `softprops/action-gh-release` at the commit the `v2` tag currently resolves to (`3bb1273`), not from the docs or the default branch (which is now v3):

- **Empty `body_path` is ignored, not a read failure.** `parseConfig` assigns `input_body_path: env.INPUT_BODY_PATH` with no `|| undefined` normalization, so an empty input stays `""`; `releaseBody()` then guards with `if (config.input_body_path)`, and `""` is falsy, so `readFileSync` is never called (`src/util.ts`). Passing `body_path: ${{ steps.notes.outputs.body_path }}` with an empty output is therefore a clean no-op, and the fallback branch does not need a second step or a conditional `uses:`.
- **A *missing* file is also non-fatal at v2** — the read is wrapped in a `try` that warns and falls back to the `body` input. That is precisely why the existence check is load-bearing rather than decorative: pointing `body_path` at a file that is not there would not fail the job, it would quietly publish an empty body again.
- **`body_path` and `generate_release_notes` together concatenate, they do not conflict.** `prepareReleaseMutation` (`src/github.ts`) calls GitHub's `generateReleaseNotes` API, sets `generate_release_notes = false`, and then joins: `body = ${body}\n\n${releaseNotes.data.body}` when a body is present, else the generated notes alone. So setting both would append GitHub's auto-changelog underneath the curated text — hence `generate_release_notes` is explicitly `false` on the curated branch. Both the create and update paths funnel through this same function.
- The action's own unit tests cover `body_path` as `undefined` and as a non-existent path, but **not** as an empty string; that behaviour is established by reading the source, and confirmed by running the `releaseBody` function's logic directly (empty string → no read attempted, no warning; missing path → `ENOENT` warning, then fallback).

## Frozen-folder caveat — please weigh in

`changelog/README.md` says released folders are frozen, and this PR adds `changelog/0.1.1/RELEASE.md` to one. The content is fetched verbatim from the live Release (`gh release view v0.1.1 --json body`), so it records what was published rather than revising it — documentation, not history editing, and without it the convention has no worked example. Line endings were normalized from the API's CRLF to LF to match every other `.md` in the repo; the text is otherwise byte-identical. Happy to drop the backfill if you read the freeze rule more strictly.

## The one way this can still go wrong

The release job runs on the tag's checkout, so `RELEASE.md` must be committed **before** the tag is created — a file added afterwards is invisible to the workflow. That is stated explicitly in both `changelog/README.md` and `CONTRIBUTING.md`.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm test` (all suites), `pnpm format` + `pnpm format:check` — all green. `.github/**` is within prettier's scope here and the workflow is unchanged by `--write`.
- No actionlint in this repo, so the workflow was parsed with the `yaml` package: it parses, all 12 `${{ }}` expressions are balanced and non-empty, `Resolve release notes` (step 9) precedes `Publish GitHub Release` (step 10), and the tag expression is the same one used in the other five places in the file. Both trigger paths were evaluated by hand: `push` of `v0.2.0` → `TAG=v0.2.0`; `workflow_dispatch` with `inputs.tag: v0.2.0` on `main` → `TAG=v0.2.0` (not `main`).
- The step's `run:` block was extracted from the workflow and executed standalone under `bash -eo pipefail` with a fake `GITHUB_OUTPUT`:

  | input | stdout | outputs |
  | --- | --- | --- |
  | `TAG=v0.1.1` (file exists) | `Publishing curated release notes from changelog/0.1.1/RELEASE.md` | `body_path=changelog/0.1.1/RELEASE.md`, `generate=false` |
  | `TAG=v9.9.9` (no file) | `No changelog/9.9.9/RELEASE.md; falling back to GitHub-generated notes.` | `body_path=`, `generate=true` |
  | `TAG=v8.8.8` (file exists, 0 bytes) | falls back | `body_path=`, `generate=true` |

  Identical results whether the tag arrives from `github.ref_name` or `inputs.tag` — the step only sees `TAG`.

**Not verifiable without cutting a release:** the end-to-end path only proves itself at the next tag. What has been checked is the step logic, the YAML, and the action's documented and source-level behaviour for each branch; what has not been observed is an actual Release page produced by either branch.


hiyouga:
## Code review

Correct and unusually well-verified. I independently confirmed: the YAML parses; in the `release` job `actions/checkout` is present (step 0) and **Resolve release notes (step 9) precedes Publish GitHub Release (step 10)** with `body_path`/`generate_release_notes` wired through step outputs. The `[ -s ]` empty-file guard, the `env:`-based `TAG` (avoids shell injection), and the falsy-`body_path` / missing-file reasoning about `softprops/action-gh-release@v2` all check out. `gh release view v0.1.1` now returns a 2974-char body matching the backfilled `RELEASE.md` — i.e. the empty body was patched by hand after the fact, exactly the fragility this removes.

### 🟡 Conflicts with #47 on `changelog/README.md` — and they tell contradictory stories
Both PRs rewrite the *same* "Changes not yet released…" paragraph. #47 replaces it with the `unreleased/`-folder convention; this PR **keeps** the old "upcoming version's folder" language and inserts "write `RELEASE.md`." Merging both naively yields an incoherent doc. **Merge #47 first, then rebase this** so `RELEASE.md` is described *within* the unreleased→version rename flow (and reconcile this PR's `<version>/RELEASE.md` bullet with #47's `<folder>/` generalization).

### 🟡 `AGENTS.md` not updated
`CONTRIBUTING.md` gets the `RELEASE.md` rule (good), but `AGENTS.md` — the agent-facing changelog working-rules, authoritative per the repo — describes the whole changelog process yet says nothing about `RELEASE.md`. Add a line there for symmetry, or the next agent preparing a release won't know to write it.

### 🟢 Frozen-folder backfill — a policy call, well-flagged
Adding `changelog/0.1.1/RELEASE.md` to a folder `changelog/README.md` calls "frozen" is a real (self-flagged) tension. The content is verbatim from the live release (documentation, not history rewriting) and gives the convention a worked example — I'd accept it, but it's genuinely the maintainer's call, and asking in the PR body was the right move.

hiyouga:
Updated per the review:

- **`changelog/README.md` contradiction removed**: this PR was editing the "changes not yet released" paragraph (re-stating the numbered-folder rule) that **#47** rewrites to the unreleased convention. Reverted that edit — #47 owns the paragraph — and kept only the additive `RELEASE.md` bullets. Merge **after #47** (trivial adjacent-line merge, no contradiction).
- **On the "update AGENTS.md" review point — that was my error.** `AGENTS.md`/`CLAUDE.md` are gitignored local files, not repo docs. The tracked contributor doc is `CONTRIBUTING.md`, which this PR **already** updates with the `RELEASE.md` rule — so that's covered, no change needed.
- **Frozen-folder backfill** (`changelog/0.1.1/RELEASE.md`): kept, per the PR's own reasoning — it records the already-published notes and gives the convention a worked example. Still happy to drop it if you read the freeze rule strictly.

The workflow, `check-release` gating, and npm publish are untouched.

PR #53, opened 2026-07-23
fix(web): sidebar list expansion made the whole page scroll

hiyouga:
## Summary

Expanding the sidebar session list ("更多" / the subagent · scheduled · archived folders) made the **whole page** scroll: the composer could be pushed up out of place and a large blank area appeared under the lower half of the page.

Root cause: each session row renders its Agent name for screen readers as a `sr-only` span, and Tailwind's `sr-only` is `position: absolute`. The session-list scroller was not positioned, so the nearest containing block of those spans was the initial containing block (`html`). Boxes belonging to rows past the fold therefore escaped the scroller's `overflow-y-auto` clip and stretched the *document* (measured: `scrollHeight` 1930px vs 721px viewport with 40 rows) — the page itself became scrollable while the sidebar looked fine.

Fix: make the scroller its own containing block (`relative` on the session area), the same pattern the Files panel already documents for its clipping window. Absolute descendants now anchor and scroll inside the list.

Also repairs the e2e harness so the suite (and the new regression) can run at all: since #46 the server canonicalizes the App onto `localhost` and reserves `127.0.0.1` as the preview host, where `/api` answers 401 — `run.sh` still pointed `BASE_URL` at `127.0.0.1`, so every spec failed at login.

## Changes

- `packages/web/src/components/layout/sidebar.tsx` — `relative` on the session-area scroller (+ comment stating the constraint).
- `packages/web/e2e/paging.spec.mjs` — regression: the document must stay unscrollable both before and after clicking "更多" (fails on `main`).
- `packages/web/e2e/run.sh` — `BASE_URL` uses the canonical App host `localhost`.

## Verification

- `pnpm test:e2e` (SKIP_BUILD=1 after a full build): **17/17 specs pass**, including the extended paging spec. On `main` the suite cannot log in (401 on the preview host) and, with only the `BASE_URL` line fixed, the new document-overflow assertions fail; with the sidebar fix they pass (window.scrollY stays 0 after expansion).
- `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check` — all green (Node v24).

### DIFF
--- a/packages/web/e2e/paging.spec.mjs
+++ b/packages/web/e2e/paging.spec.mjs
@@ -5,6 +5,12 @@
  * page, after which all 21 rows are visible and the "更多" row disappears (no more hidden
  * rows, no more server pages).
  *
+ * A list taller than the viewport must scroll INSIDE the sidebar: the document itself
+ * stays unscrollable before and after "更多". Each row's sr-only Agent name is
+ * position:absolute, and without a positioned scroller those boxes anchored to the
+ * initial containing block, stretched the document, and let the whole page scroll (the
+ * composer could be pushed up, leaving blank space below).
+ *
  * Standalone spec: shares one server with the other specs, so it registers its own user
  * (auto-provisions a default Project) and seeds sessions via the API.
  */
@@ -59,8 +65,19 @@ test("sidebar shows 20 sessions plus a More row; More loads the 21st and then di
   const more = sidebar.getByRole("button", { name: "更多" });
   await expect(more).toBeVisible();
 
+  // The 20-row list already exceeds the 720px viewport: it must scroll inside the
+  // sidebar, never stretch the document (the sr-only regression described up top).
+  const docScrollable = () =>
+    page.evaluate(
+      () => document.documentElement.scrollHeight > document.documentElement.clientHeight + 1,
+    );
+  expect(await docScrollable(), "document scrollable before 更多").toBe(false);
+
   // More: raises the display cap and fetches the next server page → all 21 rows, no More left.
   await more.click();
   await expect(rows).toHaveCount(TOTAL);
   await expect(more).toHaveCount(0);
+
+  // Still only the sidebar scrolls after the list grew past one page.
+  expect(await docScrollable(), "document scrollable after 更多").toBe(false);
 });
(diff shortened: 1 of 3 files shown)


hiyouga:
## Code review

Tight, well-diagnosed fix. Three parts — the `relative` scroller fix, an e2e regression, and an e2e harness repair. LGTM; details below.

### The fix is correct and minimal
The root cause holds: `sr-only` is Tailwind's `position: absolute` utility (the session row renders `<span className="sr-only">{agentHint}</span>` at `sidebar.tsx:1118`), and an abspos box contributes its overflow to its **containing block** — the nearest positioned ancestor, or the ICB (`html`) when there is none. With the scroller unpositioned, those spans anchored to `html` and stretched the document past the `overflow-y-auto` clip. `relative` makes the scroller their containing block, so they clip/scroll inside it.

I checked the blast radius: `sidebar.tsx` has **no other abspos elements, dropdowns, popovers, or z-indexed overlays** in that subtree — the `sr-only` spans are the only absolute descendants — so the containing-block change re-anchors exactly what needs it and nothing else. `relative` with no `z-index` neither creates a stacking context nor shifts layout, so there's no visual side effect.

### Regression test — well targeted
`document.documentElement.scrollHeight > clientHeight + 1` must be `false` both **before and after** "更多" (fails on `main`, measured 1930 vs ~721). The `+1` sub-pixel tolerance is fine and won't mask a real 1930px overflow. Good invariant to lock in.

### e2e harness fix — correct and precisely scoped
`BASE_URL` → `localhost` is the right consequence of #46: `127.0.0.1` is now the preview host, so `/api` login 401s there. Nicely scoped — `MOCK_URL` and the `HOST=127.0.0.1` server bind are correctly left alone (the mock endpoint and the bind address, not the app origin the browser hits).

### Minor
- **Precedent claim**: the body says "the same pattern the Files panel already documents for its clipping window," but I don't see a matching `position: relative` containing-block note in `workspace-browser.tsx` — the `relative` there is about markdown link resolution. Trivial doc-accuracy nit; the fix stands on its own reasoning.
- **Process (not this PR's fault)**: #46's canonicalization silently broke the e2e harness — good that it's caught here. I swept the `e2e/` dir: only this `BASE_URL` line assumed the app answers on `127.0.0.1`, so nothing else needs the same change.
- **Future-proofing**: the fix relies on `sr-only` being the only non-portaled abspos descendant. If a later session-row feature adds an inline absolute element meant to overflow the sidebar (a hover card, say), the scroller would now clip it — the added comment helpfully flags that constraint.

**Verdict: approve.** Small, correct, and guarded by a regression that fails on `main`. (Reviewed by reading + reasoning; I did not re-run the e2e suite — trusting the PR's 17/17.)

PR #54, opened 2026-07-24
refactor(web): unify form controls, one notification rule, dedupe i18n copy

hiyouga:
Consolidates the web app's organically-duplicated UI, makes notifications follow one rule, prunes the i18n catalog — and, in follow-up commits, reworks the sidebar session list onto category-filtered paging, generalizes write confirmations, and localizes error prompts. Also carries `describeError` (core: LLM request errors unwrap their `cause` chain), the skill-update confirm dialog, the Cost Center error expansion, and the removal of the runtime dropdowns' "Reset to default" link. No behavior change for the collapse/i18n parts (values verified identical); the notification and sidebar changes are the intended UX shift. The changelog entry (`changelog/unreleased/2026-07-22-web-app.md`) is the complete, current description.

## Unified form controls
- **New `field.tsx`** — one `Field` wrapper (label/hint/error) + shared `controlBase`/`menuRowClass` class strings. `Input`, `Textarea`, `Select`, `OptionMenu`, `PasswordInput` now read from it instead of each carrying its own copy.
- **New `icons.tsx`** — `ChevronDown`, `CheckIcon`, `PlusIcon`, and a `CloseButton` that replaces the X button triplicated across `Modal`/`Drawer`/`Sheet`; inlined chevron/caret SVGs across `sidebar`/`select`/`option-menu` swapped in.
- **`Select` now reuses the shared `usePortalPanel` hook** — deleted its own ~40-line positioning + outside-click/Esc/scroll effect, so it flips up/down and closes exactly like `OptionMenu`.
- Promoted `models-page`'s local `FieldError`/label spans to the shared components.

## One notification rule
- Added an **`info` toast tone** (auto-dismiss, slides from the top) alongside success/error.
- **Field-value errors now show inline** — the input turns red with the message directly beneath it — across **8 forms** (login, change-password, create-user/reset-password, create-project, create-agent, agent runtime/tools, schedules, vault), mirroring the existing `models-page` pattern.
- Ad-hoc emerald "saved" banners and transient action-failures → **toasts**; persistent page-load-with-retry states left inline (a disappearing toast would be wrong there).

## Deduplicated i18n copy
- Removed **41 dead/orphaned keys**; folded **~24 both-locale duplicate labels** (Name, Username, Role, Actions, Created, Cost, Time…) into shared `S.common.*`.
- Lowercased mid-sentence **"agent"** in English strings (it's a common noun) and fixed the catalog header comment that mandated the old convention. Domain terms (Workspace/Token/Task/Session/Project/Trace) and proper names (Agent State, AgentHub) keep their capitals.
- The two catalogs stay strictly 1:1 (TypeScript-enforced).

## Sidebar session list: active-only default load, lazy folders, per-folder "More"
Follow-up to the earlier "More counts only active sessions" commit, which fixed the symptom but not the fetch: server pagination was source-blind, so a page could arrive full of subagent/scheduled/archived rows that vanish into collapsed folders — the visible active list stayed short, and a group whose first page held no active row could never reach its later active pages at all.

- **Server**: `GET .../sessions` takes `category=active|subagent|schedule|archived` (archived wins over the origin — the same precedence the sidebar renders by) and `counts=1` for per-category totals **plus a per-Workspace breakdown**; paging applies within the category. `readTraceMeta` now reads a bounded 256KB head window instead of whole Trace shards, so classification stays cheap.
- **Store**: pages per (Agent, category), 10 rows per page. The default load fetches **active rows + counts only**; a folder's first page is fetched on first expand; a reload refetches page 1 of every category already on screen (an open folder refreshes instead of blanking). Totals (overall and per Workspace) adjust locally on add/delete/archive.
- **Sidebar**: the Subagents/Scheduled/Archived folders share one lazy path — every group (workspace groups included) shows a folder only for its **own exact share**, fetched on click, paged by its **own** "More" row that lasts exactly until the share is fully loaded. This kills the phantom folders a workspace group used to grow from an Agent's content in *other* Workspaces (empty folder + a "More" whose click made it vanish). The group's "More" and the workspace header count key off the group's active share only.
- **Auto-opened "last conversation"** (chat home auto-select, collapsed rail, after-delete jump) follows one rule: the newest loaded **active or schedule** Session — archived rows and subagent children are never auto-opened.

## Every file-writing action confirms; empty saves say so
- Settings saves (AGENTS.md / system prompt / runtime / tools), skill **uninstall**, vault add over an existing key, and workspace uploads overwriting existing files now **confirm first** — previously one-click. All confirmations (new and existing) share one polished `ConfirmModal` with a tone icon badge (red triangle = delete, pencil = save/overwrite); chat quick-set controls stay one-click by design.
- Saving with nothing changed now toasts **"No changes to save"** (agent settings tabs, schedule edit, model edit) instead of a silent no-op or a misleading "Saved".

## Docs & comments
- Fixed **3 stale code comments** (model gateway count 3→5, `exampleTasks` `lol` card, models-page "top-level hint").
- Extended the Web App changelog entry + index.

> **Note:** design-spec syncs (`05-ARCHITECTURE.md`, `06-PROTOTYPE.md`) live in the design repo as separate commits and are **not** part of this PR.

## Verification
Full CI chain in the worktree: `pnpm -r build` ✓ · `pnpm typecheck` ✓ · `pnpm test` ✓ (core 413 · server 322 · cli 164 · web 381 · docs/skills/landing 63) · `pnpm format:check` ✓. e2e (Playwright) not run.

https://claude.ai/code/session_015sQWJZUdjdi4ESiWVaA2qi


hiyouga (commented):
## Review

Strong PR overall. The consolidation is real (one `Field`/`controlBase` source of truth, `Select` finally on `usePortalPanel`, one `CloseButton`), and the sidebar rework fixes a genuine architectural flaw — source-blind server pagination — rather than papering over the symptom. The `category` + `counts` endpoint design (archived-wins precedence shared verbatim between `SessionService.categoryOf` and `sessionCategory`, so filtering and rendering cannot disagree) is the right call, and it is well tested.

**Verified locally on `worktree-ui-unify` (30e8724):** `pnpm typecheck` ✓ · `pnpm test` ✓ (core 413 +2 skipped · server 322 · cli 164 · web 381 · docs 11 · skills 16 · landing 36) · `pnpm format:check` ✓ — matches the claims in the description.

---

### Should fix

**1. Server errors are pinned to the wrong field in two dialogs** (inline comments on `change-password-dialog.tsx` and `admin-users-page.tsx`). Both attach *any* server rejection to one specific field based on a code comment asserting the server only rejects for one reason — and in both cases that assertion is false, so a `invalid_password` length error renders under the wrong label. This is a regression: before this PR both were a single form-level line, which was correct for every code.

**2. `counts=1` re-walks the whole Traces directory once per Session** (inline comment on `session-service.ts`). The 256KB head window bounds the *read* but not the *walk*, and the default sidebar load now always requests counts, for every Agent in parallel.

### Worth a look

**3. `Select` has no `error` prop.** `Input`/`Textarea`/`PasswordInput` all take one and `Field` already renders it, but `SelectProps` stops at `label`/`hint`/`required`. Given this PR's premise — a field-value error belongs under its field — a required `<Select>` can't participate, so `schedules-tab`'s target/model selects still fall back to the form-level line. Two lines each on `Select` and `OptionMenu` to close the gap.

**4. Inline field errors aren't announced to screen readers** (inline comment on `field.tsx`). The input gets `aria-invalid`, but the message has no `id`/`aria-describedby` link and no live region, so an error appearing on submit is silent for AT users. This PR is what makes inline errors the app-wide rule, so wiring `useId()` in `Field` once fixes all 8 forms at the source.

**5. Workspace-mode "More" can click with no visible effect.** `renderFolder`'s "More" fans out per (Agent, category), not per Workspace, so a click can page in 10 rows that all belong to *other* groups — this group grows by zero and the button stays put. The comment acknowledges it ("one click may grow this folder by fewer than a full page"), and the button is correctly gated on the group's own share so it terminates. But for an Agent with a few hundred subagent Sessions and a 3-row share in this Workspace, that's a lot of clicks that look like a dead button. A busy/disabled state during the fetch would at least make it feel responsive; a `workspace=` filter on the list endpoint would make it exact.

**6. `loadMoreFor`'s offset counts rows the server never paged** (inline comment on `sessions.tsx`).

### Nits

- **Orphaned doc comment**: `/** Source badge on session list rows … */` now sits above `emptyGreeting` in both catalogs (`strings-en.ts:645`, `strings.ts:632`) after `sourceNames` was removed.
- **`aggregateWorkspaceCounts` guard** (`session-grouping.ts:132`): `if (n <= 0) continue` lets an `undefined` count through (`undefined <= 0` is `false`) and poisons `totals` with `NaN`. `if (!(n > 0)) continue` is the same length and total-proof. Defensive only — the server always sends all four keys.
- **`Buffer.alloc(256 * 1024)`** per `readTraceHead` call zero-fills 256KB every time; `allocUnsafe` is safe here since only `subarray(0, bytesRead)` is ever read.
- **PR description is stale**: it says "Net −43 lines (609 insertions / 652 deletions)" and describes only the first commit. The branch is now 10 commits, +1863/−1012, and also carries `describeError` (core), the skill-update confirm dialog, the Cost Center error expansion, and the "Reset to default" removal. The changelog entries are accurate and complete — the body just never caught up.
- **Changelog index date**: `changelog/unreleased/README.md` dates the Web App line `[2026-07-24]` while linking `2026-07-22-web-app.md`.

### Test coverage

The server-side endpoint test is genuinely good — cold-registry source derivation from a Trace head, archived-wins-over-schedule, paging *within* a category, `workspaceCounts` summing back to `counts`, and 400s for junk values. `describeError` covers cycles, dedup and non-`Error` tails.

The gap is `packages/web/src/state/sessions.tsx`: the offset arithmetic, `adjustCount` across add/remove/archive-toggle, and "a reload refetches page 1 of every category already on screen" are the most intricate new logic in the PR and have no tests. The pure helpers extracted into `session-grouping.ts` are all covered, which suggests the store logic could be extracted the same way (e.g. a pure `nextOffsetFor(state, agentId, category)`) and tested without React.

### Things I checked that are fine

- `Field` returning bare `children` when there's no label/hint/error preserves `OptionMenu`'s old `if (!label) return control` behaviour exactly.
- `PasswordInput` passes `error` down only as `invalid`, so no nested `<label>` and no double asterisk.
- `skills-page.tsx`'s `confirmUpdate` can't get stuck with `updating: true` — `updateOutdated` uses `Promise.allSettled` and never rethrows.
- `readTraceHead`'s UTF-8 boundary: the window is cut at its last newline, so a partial multi-byte sequence at the 256KB edge is always discarded; a first line longer than the window has no newline at all and correctly falls back to `readTraceTolerant`.
- `AgentSettingsPage` still renders its load error via the `if (error && !data)` early return after the bottom-of-page `{error && …}` line was deleted.
- Client/server category agreement holds within a request: `categoryOf` populates the `sources` registry before `toInfo` reads it again.

packages/web/src/components/account/change-password-dialog.tsx
```diff
@@ -17,34 +17,37 @@ export function ChangePasswordDialog({ open, onClose }: { open: boolean; onClose
...
     setBusy(true);
-    setError(null);
+    setErrors({});
     try {
       await api.changePassword({ oldPassword, newPassword });
       await refresh();
       onClose();
     } catch (e) {
-      setError(e instanceof ApiError ? e.message : S.common.unknownError);
+      // The server only rejects here when the old password is wrong — attach it to that field.
```
hiyouga: **The comment is wrong, and the field attribution follows it.**

`AuthService.changePassword` (`packages/server/src/auth/service.ts:91-99`) rejects for **two** reasons:

```ts
if (!row || !(await verifyPassword(oldPassword, row.passwordHash))) {
  throw new HttpError(400, "password_mismatch", "Current password is incorrect.");
}
if (newPassword.length < MIN_PASSWORD_LENGTH) {
  throw new HttpError(400, "invalid_password", "Password must be at least 8 characters.");
}
```

Repro: enter the **correct** current password and a 5-character new password → the server returns `invalid_password`, and "Password must be at least 8 characters." renders in red under **Current password** while the New password field stays clean. Before this PR it was one form-level line, which was correct for both codes — so this is a regression.

`apiErrorText` is already code-driven; the field should be too:

```ts
} catch (e) {
  const code = e instanceof ApiError ? e.code : null;
  const field = code === "invalid_password" ? "new" : "old";
  setErrors({ [field]: apiErrorText(e) });
}
```

A tiny shared `fieldForCode(err, map)` helper would cover this, `admin-users-page.tsx`, and the `vault-tab.tsx:106` / `agents-page.tsx` / `project-dialogs.tsx` cases (those are lower risk — the id really is the only realistically-rejectable field there — but they'd inherit the fix for free).

packages/web/src/features/admin/admin-users-page.tsx
```diff
@@ -138,33 +138,35 @@ function CreateUserDialog({
...
     }
     setBusy(true);
-    setError(null);
+    setErrors({});
     try {
       await api.adminCreateUser({ userId: id, password });
       onDone();
     } catch (e) {
-      setError(e instanceof ApiError ? e.message : S.common.unknownError);
+      // The server rejects a duplicate id here — surface it on the id field.
```
hiyouga: **Same mis-attribution as `change-password-dialog.tsx`, and here the wrong branch is hit *first*.**

`AdminService.createUser` (`packages/server/src/services/admin-service.ts:50-55`) validates the password **before** the duplicate check:

```ts
if (password.length < MIN_PASSWORD_LENGTH) {
  throw new HttpError(400, "invalid_password", "Password must be at least 8 characters.");
}
if (/* exists */) {
  throw new HttpError(409, "user_exists", `User already exists: ${userId}.`);
}
```

Repro: a brand-new valid username + a 3-character initial password → "Password must be at least 8 characters." renders under **Username**, while the password field it's actually about shows nothing.

Route by code: `invalid_password` → `password`, `user_exists` → `userId`, default → `userId`.

packages/server/src/services/session-service.ts
```diff
@@ -131,8 +198,39 @@ export class SessionService {
...
+      };
+    }
+
+    const want = paging ? paging.offset + paging.limit : Infinity;
+    const counts: SessionCategoryCounts = { active: 0, subagent: 0, schedule: 0, archived: 0 };
+    const workspaceCounts: Record<string, SessionCategoryCounts> = {};
+    const matched: SessionRow[] = [];
+    for (const row of sorted) {
+      if (!withCounts && matched.length >= want) break;
+      const cat = await this.categoryOf(row, traceIds.has(row.sessionId));
```
hiyouga: **`withCounts` classifies every row, and each cold classification re-walks the entire Traces tree.**

`categoryOf` → `sourceOf` → `readTraceMeta`, and `readTraceMeta` (line 392) opens with its own full walk:

```ts
for (const dateDir of await listDirsSafe(dir)) {
  for (const file of await listFilesSafe(path.join(dir, dateDir))) {
```

There's no memoization in `listDirsSafe`/`listFilesSafe` (they're bare `readdir` wrappers, line 459-475), so this runs **per Session**. With N Sessions over D date directories that's N×(1+D) `readdir` calls — ~27k for 300 Sessions across 90 days — and the sidebar's default load now always sends `counts=1`, for **every Agent in the Project, in parallel**, on the first load after every server start.

The `TRACE_HEAD_BYTES` window is a good optimization but it bounds the *read*, not the *walk*, so it doesn't help here.

The fix is nearly free: `discoverTraceSessionIds` (line 375) already iterates exactly these files once per request and throws away everything but the id. Have it return `Map<sessionId, { path, index }>` for the earliest shard and pass that into `readTraceMeta`/`categoryOf` — N walks collapse to 1, and `toInfo`'s lazy `sourceOf` gets the same speedup for free.

(Pre-existing on the unpaged full-list path, but paging was introduced precisely to bound it — `counts=1` reintroduces the unbounded cost on the *default* load.)

packages/web/src/state/sessions.tsx
```diff
@@ -82,79 +114,176 @@ export function SessionsProvider({ children }: { children: ReactNode }) {
...
+        const loaded = pageStateRef.current.get(pageKey(agentId, category));
+        if (loaded === undefined) return (countsRef.current.get(agentId)?.[category] ?? 0) > 0;
+        return loaded;
+      });
       if (targets.length === 0) return;
       const g = gen.current;
       const results = await Promise.all(
         targets.map(async (agentId) => {
-          const offset = sessionsRef.current.filter((s) => s.agentId === agentId).length;
+          const offset = sessionsRef.current.filter(
```
hiyouga: **The offset is derived from what's in `sessions`, but not everything in `sessions` came from a page.**

`add()` prepends Sessions that were never part of any fetched page — `chat-page.tsx:196` does exactly that for a deep link beyond the loaded pages (`api.getSession(routeSessionId).then(res => addSession(res.session))`).

Concrete sequence:
1. Page 1 of `active` loads rows 0–9 → 10 rows in `sessions`.
2. User opens a bookmark to an old active Session at the server's index 50 → the probe adds it → 11 rows.
3. Click "More" → `offset = 11` → the server returns rows 11–21.

Server row **10** is never requested, so it silently disappears from the sidebar until a full reload. The dedup-by-`sessionId` below handles the duplicate case but can't recover a skipped row.

`pageState` is already the per-pair source of truth for "has more" — storing a fetched-row count next to it (`Map<string, { hasMore: boolean; fetched: number }>`) makes the offset exact and removes the `sessionCategory` re-derivation from the hot path.

Low severity — self-heals on the next reload — but it's the kind of thing that reads as "the sidebar randomly drops a conversation".

packages/web/src/components/ui/field.tsx
```diff
@@ -0,0 +1,76 @@
...
+    </span>
+  );
+}
+
+export function FieldHint({ children }: { children: ReactNode }) {
+  return <span className="mt-1 block text-xs text-gray-500 dark:text-gray-500">{children}</span>;
+}
+
+/** Field-level error text: placed directly below the offending input (the input itself is highlighted red via the control's error/invalid state). */
+export function FieldError({ children }: { children: ReactNode }) {
```
hiyouga: **The error text isn't programmatically associated with its input.**

The control gets `aria-invalid`, which tells AT *that* something is wrong, but there's no `id` + `aria-describedby` linking this `<span>` to it and no live region — so on a failed submit a screen-reader user hears nothing and has no way to reach *what* is wrong. (The wrapping `<label>` doesn't help: it names the control, it doesn't describe it.)

Since this PR is what makes inline field errors the app-wide rule across 8 forms, fixing it once in `Field` covers all of them:

```ts
const errorId = useId();
// on the control (needs Field to clone or take a render prop):
//   aria-describedby={error ? errorId : undefined}
<FieldError id={errorId}>{error}</FieldError>
```

`Field` currently takes `children` opaquely, so this needs either `cloneElement` or handing the id down — worth the plumbing given it's the single chokepoint.

At minimum, `role="alert"` on `FieldError` would get the message announced when it appears, even without the association.

packages/web/src/components/ui/select.tsx
```diff
@@ -2,15 +2,18 @@
...
+import { Children, isValidElement, useState } from "react";
 import type { ChangeEvent, ReactNode, SelectHTMLAttributes } from "react";
 import { createPortal } from "react-dom";
 import { sizeClass } from "./input";
 import type { ControlSize } from "./input";
+import { Field, controlBase, menuRowClass } from "./field";
+import { CheckIcon, ChevronDown } from "./icons";
+import { usePortalPanel } from "./use-portal-panel";
 
 export interface SelectProps extends Omit<SelectHTMLAttributes<HTMLSelectElement>, "size"> {
```
hiyouga, line 18: `SelectProps` takes `label` / `hint` / `size` (+ `required` via the native attrs) but no `error`, while `Input`, `Textarea` and `PasswordInput` all do — and `Field` below already renders `error` and would need nothing new.

The consequence shows up in `schedules-tab.tsx`: `name`, `prompt`, `startAt` and `sessionId` all got inline errors this PR, but anything selected through `Select`/`OptionMenu` still has to fall back to the form-level `formError` line. Adding `error?: string` here (and on `OptionMenu`) and threading it into `<Field label={label} hint={hint} error={error} required={required}>` — plus the red-border treatment `Input` already defines — would make "one notification rule" actually hold for every control.

hiyouga:
Review addressed in 491ae62 — all six findings plus the nits:

**Should fix**
1. **Field mis-attribution** (change-password / create-user): both dialogs now route by error code — `invalid_password` → the new-password / password field, everything else → current-password / username. (Skipped the `fieldForCode` helper for now: two call sites, and the other candidates you listed are single-realistic-field cases.)
2. **`counts=1` walk-per-Session**: `discoverTraceSessionIds` became `discoverTraces` returning `Map<sessionId, earliest-shard location>` from the one walk the list already does; classification, adoption and `toInfo` read the head from it. Single-session GET/PATCH paths locate on demand exactly as before. N×(1+D) readdirs → 1 walk.

**Worth a look**
3. **`Select`/`OptionMenu` `error` prop**: added (shared `errorClass` red border + message below + `aria-invalid`), so dropdowns follow the inline-error rule.
4. **AT-silent inline errors**: `FieldError` now carries `role="alert"` + an `id`, and every shared control (Input/Textarea/PasswordInput/Select/OptionMenu) links it via `aria-describedby` — wired once in the control layer, all forms inherit.
5. **Dead-feeling workspace "More"**: the clicked row disables and reads "loading…" while its fetch is in flight (both the active list's and each folder's). The `workspace=` filter stays future work — the per-group share gate already guarantees termination.
6. **Offset counting never-paged rows**: `pageState` is now `Map<pair, { hasMore, fetched }>` — the offset is the pair's fetched-row cursor, so an `add()`-prepended deep-link row can't skip server row N anymore. This also deleted the `sessionCategory` re-derivation from the hot path you flagged.

**Nits**: orphaned `sourceNames` doc comment removed (both catalogs); `!(n > 0)` guard in `aggregateWorkspaceCounts`; `Buffer.allocUnsafe` in `readTraceHead`; changelog index date now matches its file; PR body intro rewritten to cover the whole branch.

**Test coverage**: with the `fetched` cursor the offset arithmetic collapsed to a map lookup, which removes the most error-prone piece you called out. The remaining store logic (reload's loaded-category refetch, `adjustCount`) is still hooks-bound — extracting it into pure helpers is a fair follow-up but I kept it out of this round to avoid churning the store twice in one PR.

Full chain green: `pnpm -r build` · `typecheck` · `test` (core 413 · server 322 · cli 164 · web 381 · docs/skills/landing 63) · `format:check`.

PR #59, opened 2026-07-24
fix(web): use the separate-origin preview for in-app HTML rendering

hiyouga:
## Problem

The Files panel's in-app HTML rendered view fed the file's text into an iframe via `srcDoc`. A `srcdoc` document has no real base URL, so a page's relative subresources — ``, `app.js`, `style.css` — could never resolve and simply failed to load. The "open in new page" preview did not have this problem, because it goes through `GET /files/preview-redirect`, which mints a signed token and 302s to the separate preview origin where `/preview/<token>/<path>` serves the document and its relative subresources under the same token.

## Change

- **Unified serving path.** When a separate preview origin is available (`previewIsolated` from `/api/me`), the in-app rendered view now loads the exact same `api.workspaceFilePreviewUrl(...)` redirect URL in its iframe as "open in new page". The document gets a real base URL on the preview origin, so relative images/assets resolve and load, and storage/cookies work — identical behavior to the new-page preview.
- **Sandbox.** The in-app iframe uses `sandbox="allow-scripts allow-same-origin allow-forms allow-popups allow-modals"`. `allow-same-origin` is safe here precisely because the document lives on the separate preview origin: it grants that origin's identity, not the app's, so the frame still cannot reach the app's cookies or DOM. Popups stay sandboxed (no `allow-popups-to-escape-sandbox`), keeping the in-app frame strictly tighter than the unsandboxed new-tab preview.
- **Fallback kept.** With no separate preview origin (`previewIsolated === false`), the previous `srcDoc` + storage-shim + `allow-scripts` rendering remains, and the existing ⚠ warning semantics on the "open in new tab" link are unchanged.
- **Reload semantics.** A per-`previewPath` nonce keys the isolated iframe, so re-opening the same file (file chip click / back-and-reopen) remounts the iframe and refetches fresh content — the `src` string alone would not change. The preview responses are `Cache-Control: no-store`, so a remount always yields current content.
- **Source view.** The raw-content fetch in `previewPath` is kept: it still feeds the Source toggle, the truncation notice, and the non-isolated fallback; the isolated rendered view loads the file itself from the preview origin.
- **Docs.** `server-api` (en/zh) updated minimally: `preview=1` is described as the redirect's same-origin fallback (it never was the panel's mechanism), and the separate-origin section now says both the panel's rendered view and "open in a new tab" go through `preview-redirect`.

## Verification

Run from the worktree root with Node v24 (`node --version` → v24.18.0):

- `pnpm -r build` — all packages built
- `pnpm typecheck` — clean (core, server, cli, web)
- `pnpm test` — all pass: docs 11, skills 16, landing 36, core 414 (+2 skipped), server 322, cli 164, web 381
- `pnpm format` + `pnpm format:check` — "All matched files use Prettier code style!"

No web component-test infra exists (web unit tests are node-environment pure-logic only), so no DOM test was added.

## Review follow-up

All review points addressed in the follow-up commit (the design-spec sync noted in the review already landed in the design repo):

1. **Stale `previewIsolated` after UI login (blocking).** `login()` in `packages/web/src/state/auth.tsx` now refetches `GET /api/me` after a successful login and applies `previewIsolated` (and the fresh user). A UI login on a deployment with no separate preview origin previously kept the optimistic `true` (navigation is client-side; only the mount-time fetch ever asked), sending the panel down an isolated path the server cannot serve — an opaque-origin document without the storage shim. A transient `/me` failure never fails the login itself. Logout/login cycles stay correct: every login refetches.
2. **Server hardening.** `resolvePreviewTarget` returns `null` when the configured preview origin's host equals the App request's host (host-only, case-insensitive comparison — cookies ignore ports, so a port difference separates nothing). A `PENGUIN_PREVIEW_ORIGIN` misconfigured to the App's own host previously reported `previewIsolated=true` and produced an `allow-same-origin` iframe on a same-origin document. With `null`, `/api/me` reports `false` and the safe `srcDoc` fallback engages. Covered by a new case in `packages/server/test/workspace-preview.test.ts` (same host, host+port/case variants, and a legitimately different subdomain still accepted).
3. **Lazy source fetch.** Isolated HTML previews now mount the iframe with no upfront content fetch; the source text is fetched lazily on the first Source toggle, with its own error state (`sourceError`) and a skeleton while loading. Big files are no longer downloaded twice, and a transient fetch failure can no longer downgrade the preview to "unsupported" — it only affects the Source view (re-entering Source retries). Supersedes the "Source view" bullet above.
4. **Staleness guard.** `previewSeq` now doubles as an interleaving guard: `previewPath` captures its nonce up front and publishes via a `present()` helper only while still the latest call, so a slower earlier call can no longer overwrite a newer preview (the oversized-Markdown `setRichView("source")` is guarded the same way).
5. **`allow-downloads`.** Added to the isolated iframe's sandbox so download links inside a previewed page work in-panel as they do in the new tab. The sandbox is now `allow-scripts allow-same-origin allow-forms allow-popups allow-modals allow-downloads`.
6. **Docs.** `server-api.{en,zh}.md` now name the iframe sandbox set explicitly, describe the non-isolated `srcdoc` fallback (`allow-scripts` only, in-memory storage shim, no relative subresources), and note that some browsers partition or block storage inside a cross-site iframe, so behavior may differ slightly from the top-level tab.
7. **e2e comments.** The stale comments in `packages/web/e2e/chat.spec.mjs` claiming the storage shim is what makes `#shim-ok` pass are fixed: with `run.sh` binding `127.0.0.1`, the spec exercises the isolated branch, where the page runs on the separate preview origin and `localStorage` works natively. No new e2e scenarios were added.

### Follow-up verification

- `pnpm install --frozen-lockfile` / `pnpm -r build` — OK
- `pnpm typecheck` — clean (all 7 packages)
- `pnpm test` — all pass; server is now 323 (+1 for the new `resolvePreviewTarget` case)
- Web e2e `chat.spec.mjs` (covers the isolated preview flow, `#shim-ok`, files card) — passes against the rebuilt dist. Three unrelated specs (`draft-user-scope`, `paging`, `skills`) fail identically on a pristine `origin/main` checkout on this machine (verified in a throwaway worktree) — pre-existing local flakes, not caused by this branch.

## Owner bug report: switching conversations broke the rendering UI

Reported as 「切换对话时候渲染界面会出错」. Reproduced with a new e2e spec, root-caused, and fixed in `89a51a6`.

**Mechanism.** The Files panel stays mounted across a sidebar conversation switch (client-side navigation; the chat page's `selected` lookup flips the `session` prop synchronously, no remount). Inside `WorkspaceBrowser`, the openRequest locate effect depends on `[openRequest, previewPath]`, and `previewPath`'s identity changes with `session.sessionId` — so a session switch re-runs the effect while `openRequest` is still the OLD session's request (`useFilesPanel`'s `setOpenRequest(null)` reset is a parent effect, which flushes only after child effects). The replayed `previewPath(oldPath)` runs against the new session and overrides the session-switch reset that had just cleared the preview. On this branch the isolated-HTML path publishes the preview without an upfront existence fetch, so the panel committed an iframe at `preview-redirect(newSession, oldPath)` rendering the raw `{"error":{"code":"file_not_found",...}}` body — the broken rendering the owner saw.

**Pre-existing vs branch-introduced.** The resurrection itself predates this branch: the same spec run against a pristine `origin/main` build also fails, with the panel stuck on a quiet 「该类型不支持预览」 preview (main's upfront fetch 404s into kind `unsupported`). This branch escalated the visible damage (raw 404 page in a committed iframe) but did not introduce the replay. Fixed here since it is squarely in this PR's area.

**Fix** (`packages/web/src/features/chat/workspace-browser.tsx`):

- Each `openRequest` object is handled exactly once (`handledOpenRequest` ref guard), so a `previewPath` identity change alone can no longer replay it; `browsePath` creates a fresh object per click, so re-clicking the same file still re-triggers.
- The session-switch reset now runs during render (React's documented adjust-state-when-a-prop-changes pattern) instead of an effect, so no frame ever commits the old session's preview against the new session (previously the isolated iframe's `src` flipped to newSession+oldPath for one frame and fired a doomed request). The reset also bumps `previewSeq`, invalidating in-flight `previewPath` calls from the old session via the existing `present()` guard, so late results can't resurrect old-session content.

**Regression coverage.** New `packages/web/e2e/files-switch.spec.mjs`: opens an HTML preview in session A via the message file card (the `browsePath`/`openRequest` path the bug requires), switches to session B via the sidebar, asserts the panel is back on B's (empty) list — no preview header, no iframe — then switches back to A and asserts the list is usable. It fails on the unfixed build (iframe with the 404 body captured in the Playwright snapshot) and on `origin/main` (unsupported-preview resurrection); it passes with the fix, and `chat.spec.mjs` still passes.

### Session-switch fix verification

- `pnpm install --frozen-lockfile` / `pnpm -r build` — OK
- `pnpm typecheck` — clean (all 7 packages)
- `pnpm test` — all pass (docs 11, skills 16, landing 36, core 414+2 skipped, server 323, cli 164, web 381)
- e2e: `files-switch.spec.mjs` + `chat.spec.mjs` — both pass against the final build


hiyouga:
## Review: REQUEST CHANGES

Reviewed `git diff origin/main...` (base `3f868ac`) — 3 files, +56/-22.

The relative-subresource fix is real and the mechanism is right. Two things block: the design spec still asserts the opposite of what this implements, and the `previewIsolated` precondition the fallback depends on is frequently wrong. Separately, the change materially widens the trigger surface for agent-written HTML.

---

### Blocking

**1. `design/specs/05-ARCHITECTURE.md` §"Workspace 文件预览" contradicts this PR, and no companion design change exists.**

CLAUDE.md makes `design/specs/` ground truth and requires code and design to move in sync. The spec currently says:

- L1159 — 「**内联预览**（Files 面板内）：html / svg 一律以 `text/plain` 返回，杜绝同源脚本执行」. After this PR the Files-panel rendered view serves real `text/html` from the preview origin and executes scripts.
- L1161 — names the separate-origin form 「**新标签页预览**（html）」; it is now also the in-panel form.
- L1176 (安全不变量) — 「内联预览与下载的既有行为不变」. Inline-preview behavior is exactly what changed.

The spec also never documents the new security-relevant interface: the iframe `sandbox` attribute set, which is now the only boundary between the app and the framed document. `git -C design log --all` shows no matching change (latest `5a557c4`).

**2. `previewIsolated` is stale-`true` after a UI login, so the isolated branch is taken on deployments that have no preview origin — and the storage shim is lost.**

`packages/web/src/state/auth.tsx`: `previewIsolated` initializes to `true` (L35); the only writers are the mount-time `getMe()` effect (deps `[]`) and `refresh()`. `login()` (L65-68) sets `user` and **not** `previewIsolated`. `pages/login.tsx` calls `login()` then `navigate("/chat", {replace:true})` — client-side, no reload. `refresh()` is only called from `change-password-dialog`.

Failure scenario, on the deployment the design explicitly designates as the fallback case (LAN IP or domain with `PENGUIN_PREVIEW_ORIGIN` unset):

1. User loads the app logged out → `getMe()` 401s → `previewIsolated` keeps its default `true`.
2. User logs in → SPA navigates to `/chat`. `/api/me` is never refetched.
3. Files panel → click `report.html` → `workspace-browser.tsx:477` takes the isolated branch.
4. Server has no target → `preview-redirect` 302s to `files/content?path=…&preview=1` on the app origin.
5. That response carries CSP `sandbox allow-scripts allow-popups allow-modals allow-forms` (no `allow-same-origin`), so the document lands in an opaque origin — safe, but **the storage shim is not injected** (only the `srcDoc` branch at L504 injects it). The page's first `localStorage` access throws `SecurityError` and all scripts halt.

That is precisely the regression the shim exists to prevent (`design/specs/08-CHANGELOG.md:95`). Relative subresources still don't resolve either, so this state gets none of the PR's benefit and loses working scripts. The root cause predates this PR — but under `main` its only symptom was a missing ⚠ glyph; here it becomes functional. The PR's stated invariant ("without a separate origin the srcDoc + storage-shim fallback stays as before") does not hold in this state.

---

### Security: the change widens the trigger surface

Not a defect in the diff so much as a decision that deserves an explicit design call rather than an implementation detail.

Before (`origin/main`): the in-app rendered view was `srcDoc` + `sandbox="allow-scripts"` — opaque origin, no token, only its own bytes. Inert.

After: `src=<preview-redirect URL>` + `sandbox="allow-scripts allow-same-origin allow-forms allow-popups allow-modals"`, mounted automatically on file click.

Composed with the existing preview design:

- `services/preview-token.ts:22` — `PREVIEW_TOKEN_TTL_MS = 10 min`.
- The payload binds `{sessionId, host, expiresAt}` only — **no path scoping**. A valid token reads any path under that session's Workspace (`routes/preview.ts:36-84`).
- `routes/preview.ts:10-11` — the response deliberately carries **no CSP**: 「Because the boundary is now the origin itself, the response does NOT carry the CSP sandbox」. Egress is unrestricted.

With `allow-same-origin`, the framed document reads its own `location`, extracts the token, `fetch`es `/preview/<token>/<any path>`, and POSTs the bytes anywhere. So: prompt injection makes the agent write `report.html` with an exfil script → one ordinary click in the Files panel exfiltrates the whole Session Workspace. Under `main` this required an explicit click on the ⚠-marked "open in new tab" link.

The design's justification for token-in-URL — 「令牌对页面自身可见…但它授予的正是该页面已经拥有的东西」(L1169) — was written for an explicit user click, and is not accurate for a whole-subtree token on an auto-rendered frame.

Related: `resolvePreviewTarget` never compares the configured origin against the app's own host, and `app.ts:301` *disables* the canonical-host guard when `previewOrigin` is set. So `previewIsolated` guarantees only that a setting resolved — never that the origin actually differs. If `PENGUIN_PREVIEW_ORIGIN` is misconfigured to the app's host (or a proxy also routes the app there), `allow-scripts allow-same-origin` on a same-origin document means no sandbox at all. Suggested fix: return `null` from `resolvePreviewTarget` when the configured hostname equals `hostOnly(requestAuthority(...))`, so `previewIsolated` goes false and the safe fallback engages.

In the correct loopback default the boundary does hold — verified: `localhost` vs `127.0.0.1` are cross-site, the guard 401s `/api` and 302s everything else off the preview host, cookies are `HttpOnly; SameSite=Lax`, and `jsonOnlyWrites` + absent CORS blocks writes.

---

### Non-blocking

- **Third-party storage.** The comment claims "storage works, exactly as in the new-page preview". In a top-level tab the preview origin is first-party; framed inside the app it is third-party. With Chrome third-party cookies blocked, Firefox strict ETP, or Brave, `localStorage` in a cross-site frame throws `SecurityError` — and the isolated branch injects no shim. Elsewhere it is partitioned, i.e. a different store than the new tab sees. (Browser specifics reasoned, not executed; the load-bearing half — no shim on this path — is verified in code.)
- **e2e coverage silently inverted.** `e2e/run.sh` sets no `PENGUIN_PREVIEW_ORIGIN` but `HOST=127.0.0.1`, so `previewIsolated` is genuinely `true` and `chat.spec.mjs:237-239,265-266` now exercises the isolated branch. `localStorage` succeeds natively rather than through the shim, so `#shim-ok` passes vacuously, the comment ("without the shim it would throw SecurityError") is now wrong, and the srcDoc/shim path has **zero** e2e coverage.
- **Unused blocking fetch.** For `kind: "html"` the isolated path still awaits a full `res.text()` it only needs for the Source toggle (not the default view). A 30MB file is downloaded twice and the frame doesn't mount until the first completes. Worse, a transient failure (`!res.ok` → `catch`) downgrades the file to `kind: "unsupported"` and renders no preview at all, even though the isolated path would have worked. Lazy-load the source on first toggle.
- **`previewSeq` isn't used as a staleness guard.** Two rapid `previewPath(A)`, `previewPath(B)` calls both run to completion; a late-resolving A overwrites B. Pre-existing, but this PR introduces exactly the monotonic counter that would fix it — `if (nonce !== previewSeq) return;` before each `setPreview`.
- **`allow-modals`** lets agent HTML tab-block the app with `alert()` on file click. `allow-downloads` is absent, so an in-page download link silently does nothing in-panel while working in the new tab.
- **Docs overstate.** `server-api.{en,zh}.md:172` now says both the panel's rendered view and "open in a new tab" go through `preview-redirect` — true only when `previewIsolated`. The non-isolated in-panel `srcDoc` path is not mentioned, nor is the iframe's sandbox set.

---

### Verified as claimed

- `srcDoc` did make relative subresources unresolvable — a srcdoc document's base URL is its parent's.
- The in-app iframe now loads the same `api.workspaceFilePreviewUrl(...)` redirect as "open in new page".
- The per-`previewPath` nonce genuinely remounts: bumped at L262, set at all six `setPreview` sites, `key={preview.nonce}` at L489. Switching files also remounts; no stale state leaks.
- Hook correctness is clean — `previewPath`'s `useCallback` deps `[session.sessionId]` are right (`previewSeq` is module-level, `onPreviewOpen` goes through a ref, `previewIsolated` is read in render not captured). No cleanup needed.
- No non-HTML regression: image/pdf/text/md/unsupported paths only gained the required `nonce` field. The truncation notice is correctly absent from the isolated branch and still present in source and md views.
- `server-api` en/zh are mirrored and the zh is a faithful translation. No new user-facing strings, so no `strings.ts`/`strings-en.ts` change was required.

---

### Cross-PR note

No merge conflicts with #60-#65. `server-api.{en,zh}.md` is also touched by #63 and #64, but in different sections — #63↔#64 conflict there, #59 does not.

hiyouga:
切换对话时候渲染界面会出错，这个需要修复一下

hiyouga:
已修复切换对话时渲染界面出错的问题。根因：文件面板跨会话保持挂载，openRequest 定位 effect 因 previewPath 身份随 sessionId 变化而重放（父组件清空 openRequest 晚于子组件 effect），把旧会话的文件路径按新会话重新打开，隔离预览分支无前置存在性检查，iframe 直接渲染出 preview-redirect 的 404 响应体（该重放在 origin/main 上同样存在，表现为"该类型不支持预览"）。修复：每个 openRequest 对象只处理一次（ref 守卫），并把会话切换重置移到渲染期执行、同时递增 previewSeq 使旧会话在途请求失效；新增回归用例 packages/web/e2e/files-switch.spec.mjs。提交：89a51a6。

PR #61, opened 2026-07-24
fix(core): present the project dir to the model as the App Data Dir in the default system prompt

hiyouga:
## Problem

The default system prompt's Environment section exposed the project directory under the label `Project Dir`. Models kept mistaking it for the user's working directory (which is actually `CWD`) — treating its contents as task input and writing task deliverables there.

## Change

Same value, clearer name and semantics (commit 497363f, superseding the earlier `{{AGENTS_DIR}}` approach on this branch — that placeholder never shipped and is fully removed):

- The Environment line is now `- App Data Dir: {{PROJECT_DIR}}` — `{{PROJECT_DIR}}` remains THE placeholder, with the same substitution as main.
- The `# File system` section states the semantics explicitly: the App Data Dir is PenguinHarness's application data root, holding every agent's data files (`<app_data_dir>/agents/<agent_id>/agent_state/…`) plus the project-level data files; it is NOT the current task's directory and was not provided by the user — never treat its contents as task input, never place task deliverables there (the working folder is `CWD`).
- Angle-bracket notation is `<app_data_dir>`: own Agent State, other agents' assets, scratchpad, skills path and the PLAN.md suggestion are all spelled `<app_data_dir>/agents/…`; the secrets rule points at `.project_config.toml` directly under the App Data Dir.
- Library skills (`agent-creation`, `agent-evaluation`, `agent-optimization`, `benchmark-design`, `penguin-cli`, `penguin-sdk`) read "the App Data Dir value from your Environment section"; the eval-family's project-root derivation simplifies to `PROJECT_DIR = <app_data_dir>`. Frontmatter `version`/`updated` bumped per the library convention.
- Web system-prompt editor: the placeholder hint list documents `{{PROJECT_DIR}}` with the app-data-root description ("not the task working directory"), matches the prompt's placeholder order, and includes the previously missing `{{SKILL_METADATA}}`, `{{PROVIDER}}`, `{{MODEL_ID}}` (zh/en mirrored).
- Docs (`configuration` en/zh § System prompt placeholders) document `{{PROJECT_DIR}}` with the App Data Dir semantics. No legacy story is needed: the placeholder is unchanged, so existing custom prompts are unaffected.
- Tests assert the `App Data Dir` label and `{{PROJECT_DIR}}` substitution, and guard that the raw "Project Dir" label (and `<project_dir>` notation) never reaches the model.
- The `# System markers` section and the compaction prompt are untouched (handled by a parallel PR).

## Restore default configuration

[...]

PR #62, opened 2026-07-24
feat(core,web,cli): add file tools and per-tool call descriptions

hiyouga:
## What

Three related changes to the builtin toolset (core), surfaced through the server config API, the Web settings UI, and both renderers (CLI + Web).

### 1. New builtin file tools: `read_file` / `edit_file` / `write_file`

- `read_file` (permission `r`, timeout 30s, output cap 64k): `cat -n` style line-numbered output; `offset`/`limit` (default 2000 lines) page through long files with a trailing continuation note; overlong single lines are truncated; binary content (NUL bytes) is rejected with advice to use shell/image tools; empty files, missing files, and directories all produce clear explanations.
- `edit_file` (`rw`, 30s, 16k): exact-string replacement mirroring Claude Code's Edit semantics — the file must exist, `old_string` must occur exactly once (0 occurrences fails with "old_string not found"; N>1 fails with the count and a hint to add context or set `replace_all`), `old_string === new_string` fails; success echoes "Replaced N occurrence(s)" plus a line-numbered snippet around the change for verification.
- `write_file` (`rw`, 30s, 16k): whole-file create/overwrite, creating parent directories as needed; output distinguishes "Created" vs "Overwrote" and reports lines/bytes; an empty string is valid content.

All three resolve relative paths against the workspace, allow absolute paths (same trust model as the shell tool), are non-streaming, and never throw — failures come back as explanatory text with `stop_reason: failed`.

### 2. ~~Rename `exec_command` → `run_command`~~ (reverted)

Reverted per owner feedback (「还是改回去exec_command吧」): the shell tool stays **`exec_command`** — tool name, default config entry, renderer special-cases, mock harnesses, e2e specs, test fixtures, and docs are all back on `exec_command`, and the legacy-alias machinery this PR briefly introduced (registry alias, deprecated re-export, alias tests) is deleted again. One improvement is kept because it is independently more robust: assembled tools take their runtime name from `definition.name` (the config entry stays the single source of truth for naming).

### 3. Optional `description` call argument + per-tool `call_description` toggle

> Reworked per owner review — the description below is superseded by the design in "## Review follow-up" (A): the argument now lives in the config schema and the toggle is per-tool.

- `exec_command` / `input_command` / `run_subagent` / `input_subagent` accept an optional `description` string — one model-written sentence about what the call is doing, shown while it runs.
- Injected into `parameters.properties` at assembly time in `buildToolConfig` (in-memory only — the stored YAML is never rewritten; never added to `required`; entries without a schema or with their own `description` property pass through untouched). The injected schema flows into the LLM tool list and `session_meta.tools` automatically.
- Controlled by the new `SystemConfig` field `tools.call_descriptions` — **missing = enabled**, so all existing agents get the feature; explicit `false` turns it off. Plumbed through the server config API (`callDescriptions` in `AgentConfigDto` / update request, boolean-validated) and a switch row in the Web agent settings Tools tab (zh + en strings).
- Rendering needs no config plumbing: if the toggle is off the model never sends the argument. CLI appends a `— <description>` suffix once the argument JSON is complete (keeps the append-only preview contract); Web shows it as the collapsed-header subtitle, while the approval-row preview deliberately keeps the real command (users must approve the actual command, not the model's summary of it). The file tools get `file_path`-based previews in both renderers.
- The three file tools do not take `description` — their path argument is self-describing.

### Docs

`tools.{en,zh}.md` rewritten framing (file tools for precise reading/editing, shell as the general-purpose fallback), tool table now 9 tools, file-tool arg schemas, call-descriptions section, YAML example; counts and mentions updated in introduction / interfaces / skills / sessions-and-traces / architecture / web-app / configuration (zh + en). Landing blog posts and changelog/ left untouched (historical).

## Verification

All run at the worktree root with Node v24:

- `pnpm -r build` — OK
- `pnpm typecheck` — OK (all 7 packages)
- `pnpm test` — OK: docs 11, skills 16, landing 36, core 445 (+2 skipped e2e needing live keys, pre-existing), server 324, cli 168, web 389 — all passing
- `pnpm format` + `pnpm format:check` — OK

New tests (final state): per-tool behavior for read_file/edit_file/write_file (happy paths, every error case, workspace-relative resolution, offset/limit, replace_all, parent-dir creation, plus the hardening cases from the review follow-up); buildToolConfig per-tool `call_description` filter tests (default-on, false filters without mutating the stored config, no-op cases); server config API round-trip for the per-tool field (defaults, PUT false, 400 on non-boolean); CLI renderer tests (formats, append-only streaming, file-tool previews, approval payloads); Web preview/subtitle/pending-payload helper tests.

## Review follow-up

Second commit (`refactor(core,cli,web): per-tool call_description, bounded read_file, exact CLI formats`) addresses every point of the owner review:

- **A — toggle redesign (per-tool, config-resident).** The `description` property is written directly into the four command/subagent entries' `parameters.properties` in `defaultBuiltinTools()` (all tool parameters belong in the editable config), and each `ToolDefinitionConfig` gains `call_description?: boolean` (doc-commented; missing = true), set to `true` on those four defaults. `buildToolConfig` now *filters* the property out of an in-memory clone when an entry says `call_description: false` — the YAML is never rewritten, and old configs without the property/flag are a no-op. The global `tools.call_descriptions` is fully removed (SystemConfig, server DTO/`applyConfigUpdate`/validation, the web global switch, its tests and docs). The Tools tab now has a per-row switch (like the permission column), shown only for entries whose parameters actually declare a `description` property; the server carries the per-tool field on `toolsBuiltin` rows (`validateToolsBuiltin` boolean-checks it). Tests became filter tests (default-on, false filters without mutating the stored config, no-op cases) plus a server round-trip (defaults carry `call_description: true`, PUT false lands in the YAML, 400 on non-boolean); configuration/tools docs updated (zh+en).
- **B — exact CLI formats + `->` output gutter.** `renderPartialToolCall` now emits exactly `exec_command <- $ {cmd}` / `exec_command <- {description} ($ {cmd})`, `input_command <- {process_id} << {chars}` / `input_command <- {description} ({process_id} << {chars})`, `run_subagent <- {prompt}` / `run_subagent <- {description} ({prompt})`, `input_subagent <- {subagent_id} << {prompt}` / `input_subagent <- {description} ({subagent_id} << {prompt})` (payload `<< …` only when non-empty, as before). Append-only streaming is preserved: the description is adopted only once its own value is complete, and the parenthesis closes when the wrapped payload settles. All tool **output** gutters in `render.ts` (streaming, history, image placeholders) unified on `[tool-xxx] -> `.
- **C — shortened file paths.** Both the CLI preview and the web card preview/subtitle show at most one parent directory + filename (`…/parent/file.ts`; shared `shortenPath` rule, full path stays in the expanded arguments). The CLI renders the file-tool preview only once the path argument is complete, so shortening never rewrites the append-only line.
- **D — ordering.** `read_file` / `edit_file` / `write_file` now lead both `defaultBuiltinTools()` and the registry map.
- **E — CLI startup banner.** `t.header` (chat and run) now prints `PenguinHarness {kind} v{version}` plus `Agent:` / `Workspace:` / `Model:` (zh: `模型：`) each on its own line, per-language via the existing i18n pattern; `VERSION` comes from core.
- **F — bounded `read_file`.** Complete rework: stat first, then an incremental fd-based scan (256KB chunks, abort-signal checked per chunk; `readFile`/`writeFile` calls elsewhere pass `ctx.signal`) that splits lines at the byte level, stores only the requested window (per-line byte retention cap), and counts the rest. A hard 8MB scan cap: hit before the window produces anything → clean failure with `sed -n` guidance; hit after → partial window plus a lower-bound `[file has more than N lines; …]` note. Output is **self-budgeted** below the tool's own `maxOutputLength` (256-char note reserve): the window is trimmed and the shown range adjusted so the trailing `[file has N lines total; showing X-Y — call again with offset to continue]` note always survives Environment's front-keep truncation (`(output limit reached)` infix when trimmed). Covered by tests incl. an 8MB+ single-line file and an 8MB+ many-line file.
- **G — approvals show the payload.** Web: the pending-approval block now renders the decoded file-tool arguments (`old_string`/`new_string`/`content`, plus `file_path`/window args) in a scrollable `max-h-72` pre while pending. CLI: `beginUserPrompt` prints the decoded payload (multi-line fields gutter-prefixed `  | `, bounded to 24 lines / 200 chars per line with an explicit `[… N more lines not shown]` note) right before the prompt, without duplicating the call line. Tests updated/added on both sides (`tool-call-preview.test.ts`, `tool-render.test.ts`, `render.test.ts`).
- **H — mock harnesses.** `packages/web/e2e/mock-llm.mjs` and `packages/landing/scripts/capture-shots.mjs` emit the canonical shell tool name (`exec_command` after the revert; the e2e specs assert the same).
- **I — atomic writes.** `edit_file`/`write_file` write via a shared `atomicWriteFile` (temp file in the target directory + rename, preserving the original permission bits; temp removed on failure). Tests assert no temp files remain and that an overwritten file keeps mode 0600.
- **J — numeric-string offset/limit.** Coerced (`"3"` → 3); non-numeric values fail with `Invalid "offset"/"limit"` instead of silently defaulting.
- **K — empty `old_string`.** Its own message ("old_string must not be empty — … use write_file"), distinct from the missing-argument error.
- **L — CRLF.** `read_file` strips trailing `\r` from displayed lines and appends `(file uses CRLF line endings)` when detected; `edit_file`'s not-found error mentions CRLF when the file contains `\r\n`.
- **M — empty-snippet newline.** A whole-content erase now reports `Replaced 1 occurrence in "x".` with no dangling newline.
- **N — doc-comment tone.** The three tools' module docs now say failures are finalized as `failed` and anything unexpected that still throws is caught by Environment and finalized as failed (no more absolute "never throws").
- **O — secret guard.** `read_file` refuses paths whose basename is `.vault.toml` or `.project_config.toml` (it is auto-approved under read-only approval, so it needs its own guard; the shell tool stays rw-gated) + tests.
- **P — system prompt Skills line.** `# Skills` now says to read `SKILL.md` "in full with the read_file tool" (was "with a shell command"). ⚠ Merge note: this line is also touched by #61/#63 (system-prompt markers/dirs PR) — whoever merges second must keep all edits to this template.
- **Q — docs: frozen toolsets.** tools.{en,zh}.md now note that an existing agent's persisted `tools.builtin` list is frozen (the settings UI adds no rows) — hand-edit its `system_config.yaml` to adopt the file tools.

### Revert (third commit)

Commit `f7e9d64` reverts the rename per the owner's comment while keeping everything else; repo-wide grep confirms zero `run_command` residue in sources/tests/docs (changelog/ and design/ untouched). Full chain re-ran green after the revert (core 456+2 skipped, server 324, cli 175, web 393, docs 11, skills 16, landing 36).

### Named output gutter (fourth commit)

Owner follow-up: the CLI shows the tool name on **both** sides of a call, and (second round of feedback) the numbered pairing tag must stay. Commits `99f4d0e` + `b6a0823` make every tool-output line carry the same `[tool-xxx] <toolName>` prefix as its call line — streaming deltas, resumed-history rendering, and image placeholders alike:

```text
before:  [tool-473] exec_command <- $ date      after:  [tool-473] exec_command <- $ date
         ✓ [approved]                                   ✓ [approved]
         [tool-473] -> Sun Jul 26 …                     [tool-473] exec_command -> Sun Jul 26 …
```

The renderer resolves the name through a `tool_call_id → name` map fed by the preceding call (partial and complete forms; history keys include the origin chain so parent/child ids cannot collide); output whose call was never seen keeps the bare `[tool-xxx]` tag. The call-side format is untouched. Branch verified mergeable against the moved main (`git merge-tree` clean after #59/#60).

### Rebased again onto main (1e23452)

#64 (handoff-style /model + per-turn thinking) merged; rebased once more (force-with-lease, shas refreshed above). The only conflicts were the flagged session-meta fixtures in `render.test.ts` / `trace.test.ts`, resolved as the union — main's fixture shape (no `thinking_level`) with this branch's tool-name assertions; the rebased tree is byte-identical to git's clean auto-merge of the two branches, and the full chain re-ran green (core 470+2 skipped, server 331, cli 178, web 397). Still MERGEABLE.

### Rebased onto main (816a823)

#59/#60/#61 merged while this PR was open, so the branch was rebased onto `main` (force-with-lease; commit shas above are post-rebase). One conflict, resolved as agreed for the #61 collision: `DEFAULT_SYSTEM_PROMPT`'s `# Skills` line now carries **main's** `<app_data_dir>/agents/…` path **plus this PR's** "read that skill's SKILL.md in full with the read_file tool" phrasing; everything else (defaultBuiltinTools / call_description work, strings, tests) merged cleanly. The rebased tree was verified byte-identical to a pre-verified test merge with main, and the full chain re-ran green on it (core 459+2 skipped, server 328, cli 176, web 395, docs 11, skills 16, landing 36). GitHub now reports the PR MERGEABLE.

### Git-style diff outputs + lead-arg schema order (commit `3431ae4`)

Owner follow-up (streaming visibility + Claude-Code-style outputs):

- **Schema order** verified already correct — `file_path` is the first property of the three file tools and `description` the first of the four command/subagent tools, so models emit the user-visible argument first and the CLI/web previews surface it as it streams. No changes needed.
- **edit_file** output is now a git-style unified diff after the summary line (new shared `diff.ts`): one hunk per replacement site (nearby sites merged, exact by construction via regional re-replacement — same-line multiple hits diff correctly), capped at 5 hunks for replace_all storms with an `…and N more replacements` note; self-budgeted under `maxOutputLength` so the summary/note always survive; CRLF `\r` stripped and overlong diff lines capped:

  ```text
  Replaced 1 occurrence in "src.ts".
  @@ -1,3 +1,3 @@
   const a = 1;
  -const b = 2;
  +const b = 20;
   const c = 3;
  ```

- **write_file** overwrites append a small unified diff vs the previous content (real LCS line diff with prefix/suffix trimming and a size guard; previous content read back bounded ≤1MB, binary skipped), a one-line `+X/−Y lines vs the previous content (diff too large to show)` summary past ~60 diff lines, or `(content unchanged)`; created files carry no diff. read_file stays cat-n style.
- **CLI coloring**: `+` green / `-` red / `@@` dimmed for edit_file/write_file outputs in both the streaming and history renderers (other tools' output untouched; coloring keys off the line's first character, safe because file-tool outputs arrive as one whole-line delta).
- Approval views unchanged; tools docs (zh+en) document the new output formats; tests cover the exact hunks, the cap/budget/CRLF cases, write_file's diff/summary/unchanged/created variants, and the renderer coloring.

### Verification (rework)

Re-ran the full chain at the worktree root with Node v24: `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck`, `pnpm test` (docs 11, skills 16, landing 36, core 457 (+2 skipped live-key e2e), server 324, cli 176, web 393 — all passing), `pnpm format` + `pnpm format:check` — all green.


packages/core/src/state/default-config.ts
```diff
@@ -76,6 +76,14 @@ export interface SystemConfig {
...
     /** MCP Server configuration. */
     mcpServers?: MCPServerConfig[];
+    /**
+     * Whether the command/subagent tools (run_command / exec_command / input_command /
+     * run_subagent / input_subagent) accept an optional `description` argument — one
+     * model-written sentence about what the call is doing, surfaced by the frontends while
+     * the call runs. Injected into the tool schemas at assembly time (the stored YAML is
+     * never rewritten); defaults to true when missing, set false to turn it off.
+     */
+    call_descriptions?: boolean;
```
hiyouga: 这个要细粒度，逐个工具选择。

packages/core/src/state/agent-state.ts
```diff
@@ -411,11 +411,61 @@ export function selectBuiltinToolsForModel(
...
+      ...params,
+      properties: { ...properties, description: callDescriptionProperty() },
+    },
+  };
+}
+
 export function buildToolConfig(state: AgentState): ToolConfig {
   const systemTools = state.systemConfig.tools;
   const builtin = systemTools?.builtin ?? defaultSystemConfig().tools?.builtin ?? [];
+  // tools.call_descriptions defaults to ON: a missing field (all pre-existing configs)
```
hiyouga: 工具的所有参数必须放到system config文件里面（符合principle：可编辑性），如果用户关闭，再过滤掉这个参数。

packages/cli/src/tool-render.ts
```diff
@@ -114,38 +121,69 @@ function extractPartialStringField(argsJson: string, field: string): string | nu
...
+ * tools fall back to name(args-prefix). The command/subagent tools append the optional
+ * `description` argument as a ` — <description>` suffix once the arguments are complete.
  */
 export function renderPartialToolCall(name: string, argsJson: string): string | null {
   if (!argsJson) return null;
-  if (name === "exec_command") {
+  if (name === "run_command" || name === "exec_command") {
     const cmd = extractPartialStringField(argsJson, "cmd");
-    if (cmd !== null) return `$ ${toSingleLine(cmd)}`;
+    if (cmd !== null) return `$ ${toSingleLine(cmd)}${descriptionSuffix(argsJson)}`;
```
hiyouga: 对于 run command ，如果有 description，就把他写成 run_command <- description ($ 实际 bash command) ，没有 description 就是 run_command <- $ 实际 bash command

packages/cli/src/tool-render.ts
```diff
@@ -114,38 +121,69 @@ function extractPartialStringField(argsJson: string, field: string): string | nu
...
+    }
     return null;
   }
   if (name === "input_command") {
     const pid = extractPartialStringField(argsJson, "process_id");
     if (pid === null) return null;
     const chars = extractPartialStringField(argsJson, "chars");
     const payload = chars ? ` << ${capPreview(visualizeControlChars(chars))}` : "";
-    return `⌨ input_command → ${toSingleLine(pid)}${payload}`;
+    return `⌨ input_command → ${toSingleLine(pid)}${payload}${descriptionSuffix(argsJson)}`;
```
hiyouga: 对于 input command ，如果有 description，就把他写成 input_command <- description (pid << 实际 chars) ，没有 description 就是 input_command <- pid << 实际 chars

packages/cli/src/tool-render.ts
```diff
@@ -114,38 +121,69 @@ function extractPartialStringField(argsJson: string, field: string): string | nu
...
     const cmd = extractPartialStringField(argsJson, "cmd");
-    if (cmd !== null) return `$ ${toSingleLine(cmd)}`;
+    if (cmd !== null) return `$ ${toSingleLine(cmd)}${descriptionSuffix(argsJson)}`;
     return null;
   }
   if (name === "run_subagent") {
     const prompt = extractPartialStringField(argsJson, "prompt");
-    if (prompt !== null) return `run_subagent << ${capPreview(toSingleLine(prompt))}`;
+    if (prompt !== null) {
+      return `run_subagent << ${capPreview(toSingleLine(prompt))}${descriptionSuffix(argsJson)}`;
```
hiyouga: 对于 run subagent ，如果有 description，就把他写成 run_subagent <- description (实际 prompt) ，没有 description 就是 run_command <- 实际 prompt

packages/cli/src/tool-render.ts
```diff
@@ -114,38 +121,69 @@ function extractPartialStringField(argsJson: string, field: string): string | nu
...
-    return `⌨ input_command → ${toSingleLine(pid)}${payload}`;
+    return `⌨ input_command → ${toSingleLine(pid)}${payload}${descriptionSuffix(argsJson)}`;
   }
   if (name === "input_subagent") {
     const sid = extractPartialStringField(argsJson, "subagent_id");
     if (sid === null) return null;
     const prompt = extractPartialStringField(argsJson, "prompt");
     const payload = prompt ? ` << ${capPreview(toSingleLine(prompt))}` : "";
-    return `⌨ input_subagent → ${toSingleLine(sid)}${payload}`;
+    return `⌨ input_subagent → ${toSingleLine(sid)}${payload}${descriptionSuffix(argsJson)}`;
```
hiyouga: 对于 input subagent ，如果有 description，就把他写成 input_subagent <- description (sid << 实际 prompt) ，没有 description 就是 input_command <- sid << 实际 prompt

把所有工具输出都改成 ->

packages/cli/src/tool-render.ts
```diff
@@ -114,38 +121,69 @@ function extractPartialStringField(argsJson: string, field: string): string | nu
   return out;
 }
 
+/** The three file tools: previewed as `<name> <file_path>`. */
```
hiyouga: 应该显示最多一级父目录和文件名，太多就很长了

packages/core/src/environment/tools/registry.ts
```diff
@@ -32,8 +35,16 @@ export type BuiltinToolFactory = (
...
 export const BUILTIN_TOOL_FACTORIES: Record<string, BuiltinToolFactory> = {
-  [EXEC_COMMAND_NAME]: createExecCommandTool,
+  [RUN_COMMAND_NAME]: createRunCommandTool,
+  // Legacy alias: an on-disk system_config.yaml is loaded verbatim (no default-merge), so
+  // agents created before the exec_command -> run_command rename still list "exec_command".
+  // Both names map to the same factory, and the assembled tool takes its runtime name from
+  // the config entry — old agents and old traces keep working unchanged.
+  [EXEC_COMMAND_NAME]: createRunCommandTool,
   [INPUT_COMMAND_NAME]: createInputCommandTool,
+  [READ_FILE_NAME]: createReadFileTool,
```
hiyouga, line 38: 这三个放到最前面

hiyouga (commented):
web渲染file工具也只给出部分文件路径，现在太长了

hiyouga:
把CLI进入的提示也优化一下：显示版本，agent Workspace model都换行显示，现在太简陋了

hiyouga:
## Review: REQUEST CHANGES

Reviewed `git diff origin/main...` (base `3f868ac`) — 66 files, +1760/-274.

The implementation is well-built and the back-compat story genuinely holds — I traced the `exec_command` path end to end and it survives everywhere. Four things block: the mandated `design/specs/` update is entirely absent, `read_file` does an unbounded whole-file read in the shared server process, the approval preview hides `edit_file`/`write_file` payloads, and both mock harnesses now request a tool that new agents don't have.

---

### Blocking

**1. `design/specs/` was not updated — explicit project rule.**

Grepping all 8 spec files: `run_command`, `read_file`, `edit_file`, `write_file`, `call_descriptions` → **zero hits each**. `exec_command` → 15 hits. The design repo is clean with no related commit. Most load-bearing:

- `05-ARCHITECTURE.md:691` — the normative `model_msg`/`tool_call` OmniMessage example still has `"name": "exec_command"`.
- `05-ARCHITECTURE.md:222,224,231,233,237,261,411` — `exec_command` presented as *the* canonical shell tool; `:224` also omits the new `description` argument.
- `05-ARCHITECTURE.md:1143` and `06-PROTOTYPE.md:119` — error-code example `tool_failed:exec_command`, now `tool_failed:run_command` for new agents.
- `01-PRINCIPLES.md:143-152` — the ground-truth `system_config.yaml` schema block lists `tools.builtin` / `mcpServers` but not the new sibling key `tools.call_descriptions`.
- Nothing anywhere in `specs/` describes `read_file`/`edit_file`/`write_file`; `05-ARCHITECTURE.md:216-256` are the only per-tool sections.

**2. `packages/core/src/environment/tools/read-file.ts:86` — unbounded whole-file read, no size guard, in the shared server process.**

`bytes = await readFile(resolved)` reads the entire file regardless of `offset`/`limit`. `st` is already in hand at `:79` and `st.size` is never consulted. There is no equivalent of `read_image`'s `MAX_IMAGE_BYTES`, and `ctx.signal` is not passed to `readFile` (which accepts one).

Failure: `read_file({file_path: "/var/log/app.log", offset: 1, limit: 10})` on a 1.5 GB log allocates a 1.5 GB Buffer in the process that also serves the Web app and every other session, then `bytes.toString("utf8")` at `:115` exceeds V8's max string length and throws `ERR_STRING_TOO_LONG` — from **outside** the try/catch. The declared 30s timeout cannot stop any of this (see NB-3), so the allocation runs to completion.

`run_command` has no equivalent exposure: it streams through a bounded pipe and is killable via the signal.

**3. The approval preview for `edit_file`/`write_file` shows only `file_path`, hiding the payload being approved.**

`packages/web/src/features/chat/tool-call-card.tsx:56-59` — `previewArguments` returns the bare path for the three file tools. That same value renders inside the `{pending && (` branch (`:233`, `{preview}` at `:240`). The full arguments live only in the expanded section, collapsed by default. `packages/cli/src/tool-render.ts:183-186` does the same for the line preceding `promptApproval`.

Failure: approval mode `read-only`; model calls `write_file({file_path: "src/index.ts", content: "<1000 lines of rewritten code>"})`. `write_file` is `rw`, so it correctly stops at approval — and the user is shown `write_file  src/index.ts` with no indication of what is being written. Same for `edit_file`'s `old_string`/`new_string`.

This contradicts the module's own stated principle three lines above the code: *"the user must approve the actual command, not the model's summary of it."* `packages/web/test/tool-call-preview.test.ts:22-30` locks the behavior in.

**4. The Web e2e mock and the landing screenshot mock still emit `exec_command` against a freshly-provisioned default agent → "Unknown tool".**

Both harnesses use a fresh data root (`packages/web/e2e/run.sh:7` `DATA="$(mktemp -d)"`; `packages/landing/scripts/capture-shots.mjs:510` `mkdtempSync`), so `default_agent` is provisioned from the *new* defaults, which list `run_command` (`default-config.ts:185`). The mocks hardcode `name: "exec_command"` — `mock-llm.mjs:148,162,233`; `capture-shots.mjs:289,349`. `Environment.executeTool` keys `this.tools` by config entry → miss → `Unknown tool: exec_command`, `stop_reason: failed`.

The e2e stays **green**: approval runs before dispatch (`context-engine.ts:599` approves every complete `tool_call` regardless of whether the tool exists), and `chat.spec.mjs` only asserts the card label and the turn-2 reply. So the suite silently stopped covering real shell execution. `capture-shots.mjs:635` also still passes (it waits on the *requested* name), but the published marketing screenshots will show a failed tool call and a tool name that no longer exists. Neither script is CI-gated, so this surfaces the next time either is run.

---

### Non-blocking

- **NB-1. `read_file`'s 2000-line default and its 64000-char cap are unreconciled, and the continuation note is what gets cut.** `read-file.ts:137` joins the whole window into one delta; Environment front-truncates at `maxOutputLength` (64000, `default-config.ts:270`). A 2000-line source file averaging 45 chars/line → ~106000 chars → the model gets ~lines 1-1207 plus `[output truncated…]`, and the trailing `[file has N lines total; showing X-Y — call again with offset to continue]` (appended last, `:131`) is discarded. A model following the tool description would then call `offset: 2001` and skip 800 lines.
- **NB-2. "Never throws" isn't literally true.** Sites outside the try/catch: `read-file.ts:115,118,137`, `edit-file.ts:142` (RangeError on oversized results), `write-file.ts:31`. Environment's `catch (err)` around `gen.next()` does catch them and finalizes as `failed` + `[tool error] <msg>`, so nothing escapes into the agent loop — but the module doc comments overstate it.
- **NB-3. The declared 30s timeout cannot interrupt these tools.** Environment aborts a shared `AbortController` and relabels the result, but awaits `gen.next()` without racing it. None of the file tools pass `ctx.signal` to `readFile`/`writeFile`/`stat` (all accept one), so a slow NFS read blows past 30s and only then gets stamped `[tool timeout…]`. Compounds blocking-2.
- **NB-4. Writes are non-atomic.** `edit-file.ts:147` / `write-file.ts:84` call `writeFile` in place, no temp + rename. ENOSPC or a crash mid-write leaves the file truncated with the original unrecoverable.
- **NB-5. CRLF breaks the read→edit round-trip.** `read-file.ts:118` splits on `"\n"` only, so numbered output carries a trailing `\r`; `edit-file.ts` does exact `indexOf`. A model copying a line back as `old_string` gets `old_string not found` with an error telling it to check whitespace — which it did correctly. Neither tool normalizes or reports line endings.
- **NB-6. `offset`/`limit` silently ignore strings.** `read-file.ts:65-74` requires `typeof === "number"`. Models frequently emit `"offset": "500"`; that silently re-reads from line 1 rather than erroring, which can loop.
- **NB-7.** `edit_file` reports an explicitly-empty `old_string` as "Missing required argument". The guard is necessary (`countOccurrences` would infinite-loop, since `indexOf("", i)` returns `i`) — only the message is wrong.
- **NB-8. `read_file` is `r`, so auto-approved under `read-only`, and can read any absolute path.** `approval.ts:61` allows `permission === "r"` without a prompt; `read-file.ts:76` is `path.resolve(workspaceDir, filePath)` with no containment. `agent-vault.ts:1-7` states vault values "never enter the model context" — `read_file` on `agent_state/.vault.toml` puts them there, auto-approved. Mitigating: `read_image` was already `r` with the same resolve, `run_command` could always `cat` it (at `rw`, so gated), and `read_file` can't list directories, so exploitation needs a known path. Flagging as a deliberate-decision check, not a defect.
- **NB-9. Landing blog copy is now wrong.** `local-agents-on-amd-gpus.{en,zh}.md:97` states "no `read_file`, no `write_file`, no `edit_file`… That is intentional". Also `natural-language-training-loop.*:69`, `ai-infrastructure-past-present-future.*:38`, `simple-harness-is-all-you-need.*:56` describe the shell as the only file interface and name `exec_command`. Dated opinion posts, so arguably historical — but `:97` reads as a current architecture claim.
- **NB-10.** When an edit empties the file, `buildSnippet` returns `""` and the message ends with a dangling newline.

---

### Verified as claimed

**The rename's back-compat is solid** — this is the part I checked hardest. Dispatch is even stronger than the PR claims: `environment.ts` keys `this.tools` by the **config entry's** name and `listTools()` maps over `toolConfig.customTools`, so `BuiltinTool.name` isn't load-bearing at all; only self-referential error strings read `definition.name` (`run-command.ts:68,74`). Approval and permission are name-agnostic (`approval.ts:61`, `server/src/runtime/approvals.ts:109` both key on `toolPermission(name)`, reading the config array). `stream-error-watcher.ts` caches `p.name` dynamically, so an old agent correctly emits `tool_failed:exec_command`. Trace copies `p.name` verbatim. No allowlist, denylist, or icon map keyed on tool names exists anywhere. `packages/skills/**` never names the shell tool; the system prompt says "your shell tool" generically. Reverse direction is clean too — nothing hardcodes `exec_command` alone in product source. All four renderer call-sites include both literals.

**The one genuine user-visible gap**: an existing agent can only get the file tools by hand-editing `system_config.yaml` with the full JSON Schema — the Web Tools tab renders `toolsBuiltin` as a fixed table with no add-row control (`agent-settings-page.tsx:748-800`; rows come from `data.config.toolsBuiltin`, only permission/timeout/maxOutput editable). `tools.en.md:78` documents the alias but not this gap. Worth an explicit doc sentence.

`description` injection is correct: `agent-state.ts:427-451`, fresh objects at all three touched levels (`{...def}`, `{...params}`, `{...properties}`), `callDescriptionProperty()` returns a new object per call — no aliasing. `state.test.ts:317-397` includes a non-mutation assertion. `missing = enabled` holds through the round trip: `agent-state.ts:463-465` (`=== false` is the only off switch), `agent-config-service.ts:152-155` surfaces the key only when a boolean is stored, `setIfProvided` (`:194`) skips `undefined` so a PUT omitting `callDescriptions` leaves the YAML untouched, `optionalBoolean` (`validate.ts:190-199`) 400s on non-booleans. Turning it off gates the injection itself, upstream of `listTools()` — not just the UI.

File-tool semantics check out in detail. `read_file`'s offset arithmetic is correct 1-based (`end = min(total, offset+limit-1)`, `lines[n-1]`, trailing-newline pop at `:119` giving correct `cat -n` totals). `edit_file`'s sharp edges are handled right: `countOccurrences` (`:36-44`) uses `indexOf` + `index + needle.length`, not regex — no special-char breakage, no overlapping double-counts; `replace_all` uses `split/join` (`:142`) so a `new_string` containing `old_string` isn't re-matched; `buildSnippet` line numbers are correct for both paths. `write_file` rejects overwriting a directory at `:73`.

The "same trust model as the shell tool" justification is **true for this product**: `session-manager.ts:52-58` spawns `bash -lc` with `{...process.env, ...vault, ...HARDENED_ENV}` — no container, no chroot, host Node process. The file tools don't widen the boundary. (Except the `r` vs `rw` asymmetry, NB-8.)

CLI append-only contract holds: `descriptionSuffix` (`tool-render.ts:133-144`) returns `""` unless `JSON.parse` succeeds, which for incrementally-appended deltas only happens at the end. No double-append — `renderPartialToolCall` is pure, and `render.ts:709-717` writes only `preview.slice(lastPreview.length)` when the new preview is a prefix-extension, starting a fresh line otherwise. No XSS: `{subtitle}` is a React text child.

Strings: zero zh/en asymmetry at depth 2 (and `en: Strings` derived from `zh` enforces it at compile time). Docs: both `tools.{en,zh}.md` say 9 tools and 9 is correct — the registry's 10 entries include `exec_command` as an alias. All the cross-doc count updates landed, mirrored zh+en, with no stale count or `exec_command` left in `packages/docs/`.

---

### Cross-PR notes

- Conflicts with **#63** on `packages/core/test/engine.test.ts` and `packages/web/src/features/chat/tool-call-card.tsx` (different regions; the card collision is in the import/constant block near the top). Conflicts with **#64** on `packages/cli/test/render.test.ts` and `packages/core/test/trace.test.ts` (test churn).
- **The default system prompt is now stale w.r.t. this PR's own toolset.** This PR rewrites the shell tool's description to "everything the file tools don't cover", but `DEFAULT_SYSTEM_PROMPT`'s `# Skills` section still instructs the model to read SKILL.md "in full **with a shell command**" — exactly the read `read_file` was added for. That line is also being edited by #61 and #63, so it needs a coordinated fix.

hiyouga:
还是改回去exec_command吧，run command不好听

hiyouga:
已改回 exec_command（工具名、渲染特判、默认配置、mock、e2e 断言、测试与文档全部还原；别名机制与其测试一并移除，工具运行时名字仍取自配置条目）。文件工具、逐工具 call_description、CLI `<- / ->` 渲染格式、启动横幅及各项加固均保持不变。见 18aea98，全链路验证（build / typecheck / test / format）通过，仓库内已无 run_command 残留。

hiyouga:
命令行在工具调用和工具输出时候都要显示工具名，比如 exec_command -> xxxx exec_command <- xxx

hiyouga:
已按要求让命令行在工具输出侧也显示工具名：输出行前缀由配对标签改为工具名（流式输出、--resume 历史回放、图片占位行统一生效），调用侧格式不变。示例：之前 `[tool-653] -> a.txt`，现在 `exec_command -> a.txt`（调用行仍为 `[tool-653] exec_command <- $ ls`）。名字经 tool_call_id→name 映射从前置调用解析，未见调用时回退为 `[tool-xxx]` 标签。见 b28bb12，全链路验证通过，与最新 main 可干净合并。

hiyouga:
分支已 rebase 到最新 main（816a823，含 #59/#60/#61），--force-with-lease 推送，PR 已恢复可合并（MERGEABLE）。唯一冲突在默认系统提示词的 # Skills 行：按约定保留 #61 的 `<app_data_dir>/agents/…` 路径并叠加本 PR 的「用 read_file 工具完整读取 SKILL.md」表述；其余（文件工具、逐工具 call_description、输出行工具名前缀等）均干净合入。rebase 后各提交为 cdad1f4 / fd4cf10 / 2f4fadd / e2c51ff，全链路验证（build / typecheck / test / format）通过。

hiyouga:
[tool-473] exec_command <- $ date
✓ [已批准]
[tool-473] exec_command -> Sun Jul 26 15:01:28 CST 2026
这个编号不能丢掉

hiyouga:
已保留编号：输出行改为与调用行相同的 `[tool-NNN] 工具名` 前缀（流式、历史回放、图片占位行一致；未见调用时回退为裸 `[tool-NNN]`）。现在的效果：

```text
[tool-473] exec_command <- $ date
✓ [已批准]
[tool-473] exec_command -> Sun Jul 26 15:01:28 CST 2026
```

见 f22d918，全链路验证通过，PR 保持 MERGEABLE。

hiyouga:
和文件相关的操作，以及如果有description的时候，要把path或者description放到schema的最前面，让模型先输出这个，用户更可见。文件相关操作的工具输出要参考claude code的设计，比如编辑后要输出类似git diff一样，编辑前后的内容diff

hiyouga:
两点都已落实（043996a）：

1. **参数顺序**：核对后 schema 已把用户可见参数放在最前——三个文件工具 file_path 居首、四个命令/Subagent 工具 description 居首，模型会先输出它们，CLI/Web 预览随流即显，无需改动。
2. **文件工具输出参照 Claude Code**：edit_file 在摘要行后输出 git 风格 unified diff（每个替换点一个 hunk、相邻合并；replace_all 大量命中截断为 5 个 hunk 并附「…and N more replacements」；在 maxOutputLength 内自我预算，摘要行永不被截掉）；write_file 覆写时附与旧内容的小型 diff，改动过大改为一行 +X/−Y 摘要，内容未变提示 (content unchanged)；read_file 维持 cat -n。CLI 里 diff 行按 git 配色（+ 绿 / − 红 / @@ 淡）。示例：

```text
[tool-473] edit_file <- …/src/app.ts
[tool-473] edit_file -> Replaced 1 occurrence in "src/app.ts".
[tool-473] edit_file -> @@ -1,3 +1,3 @@
[tool-473] edit_file ->  const a = 1;
[tool-473] edit_file -> -const b = 2;
[tool-473] edit_file -> +const b = 20;
[tool-473] edit_file ->  const c = 3;
```

全链路验证通过，PR 保持 MERGEABLE。

hiyouga:
现在cli会重复输出两次，是有问题的：
> 当前目录有什么文件
The user is asking what files are in the current directory. Let me list them.
[tool-446] exec_command <- $ ls -la
[tool-446] exec_command <- 列出当前目录的文件 ($ ls -la)

应该有描述的时候只输出下面的内容。

另外把 path 或者 description 这种放到工具 schema 前面的设计写到 design 里面

hiyouga:
已修复重复输出（见 f67378a）。

根因：带描述的形态不是纯形态的前缀扩展——该模型先发 \`cmd\` 后发 \`description\`，预览切换形态时渲染层只能另起一行，于是两行都留在屏幕上。

改法：在「是否会有 description」这个问题尚未确定前**不渲染**（未见到 \`"description"\` 键、参数尚未解析完、且不是最后一片）；description 本身则可**流式领先**显示，其载荷等到描述值收尾后再追加——两种发出顺序下预览都保持追加式。参数始终没收尾（如中途被打断）时，在 stop 处仍会补出一行，调用不会凭空消失。

现在的效果（两种顺序都只有一行）：

\`\`\`text
[tool-446] exec_command <- 列出当前目录的文件 (\$ ls -la)
\`\`\`

另：schema 首位排序（文件工具 \`file_path\`、命令/子代理工具 \`description\`）的设计已按要求记入设计仓库 specs/05-ARCHITECTURE.md 的「工具」节（含「有描述时 CLI 只显示描述形态」这一口径）。

验证：build / typecheck / test（cli 180 项，含两条新回归：payload-first 不出纯形态、参数未收尾仍出行）/ format 全部通过。

hiyouga:
已按此思路重做（见 cb272c5）：渲染路径改为**先验决定**，不再等参数流到一半再判断。

- 依据取自 `session_meta.tools`——即**装配后**的工具 schema（逐工具 `call_description` 开关已经生效），CLI 在建 renderer 时用 `renderer.useToolSchemas(sessionMetaTools(session))` 注册；流中出现的 `session_meta`（恢复会话、子会话）也会同样登记。
- 未开启（schema 里没有 `description` 参数）→ 直接**流式渲染命令**，逐字增长，与改动前一致；出现异常的 description 也不再切形态。
- 已开启 → **必须等 description**：它自己流式领先显示，载荷在其收尾后追加，纯形态永不出现在屏幕上（模型先发载荷也一样）。
- **不存在该选项 / 工具未知 → fallback 到关闭**，即走流式命令那条路。
- 兜底：参数始终没收尾（中途被打断）时，stop 处仍补出一行，调用不会消失。

验证：build / typecheck / test（cli 182 项，新增「关闭时逐字流式」「开启时等待且不出纯形态」「未收尾仍出行」三条回归）/ format 全部通过。

hiyouga:
已改为**开关开启时必填**（见 1c53b62）：

- 四个命令 / Subagent 工具的 `parameters.required` 里加入 `description`，参数说明中另明确要求**最先输出**——从源头避免"最后才发"。
- `call_description: false` 时，装配期把该属性**连同其 `required` 项**一起滤除（仅内存内，不改写 YAML），不会出现"要求一个不存在的参数"。
- 由此渲染侧的等待是**有界**的：schema 里有该参数就必然会收到，前端据 schema 先验选路径即可；参数收尾仍无描述的兜底分支只用于兜住违反 schema 的模型。
- 双语文档（tools / configuration）与设计仓库 specs/05 已同步（206fd89）。

验证：build / typecheck / test（core 470、cli 182 等全绿，含 required 与过滤的断言）/ format 全部通过。

PR #60, opened 2026-07-24
feat(models,landing): add OpenRouter free models and a free-models article

hiyouga:
## Part A — catalog: OpenRouter free rows

All ids and context lengths were verified against the public OpenRouter models API (`GET /api/v1/models`) on 2026-07-24.

- **`inclusionai/ling-3.0-flash:free`** — "Ling 3.0 Flash (free)": inclusionAI's free tier of Ling-3.0-flash (released 2026-07-23), a 124B-param MoE with ~5.1B active params per token. Text-only (`supportsVision: false`), context 262,144, `usd(0, 0, 0)`. Sits in dictionary order between `google/gemini-3.5-flash-lite` and `minimax/minimax-m3`.
- **`openrouter/free`** — "Free Models Router": OpenRouter's unified free-tier endpoint that routes each request to a random free model, filtering by required features (tool calling, structured outputs, …). Context window is a deliberately conservative 128,000 (see follow-ups below); `supportsVision` deliberately `false` with a code comment — the harness must not send images to a router whose target may be text-only. `usd(0, 0, 0)`. Sits between `openai/gpt-5.5` and `poolside/laguna-m.1:free`.
- **`poolside/laguna-m.1:free`** — "Laguna M.1 (free)": the free tier of Poolside's flagship coding-agent model (agentic coding workflows with tool calling and reasoning). Text-only, context 262,144, `usd(0, 0, 0)`. Sits between `openrouter/free` and `qwen/qwen3.6-35b-a3b`.

Supporting changes:

- OpenRouter block header comment now covers the router in the genuine-$0 pricing convention.
- `model-catalog.test.ts`: all three ids inserted at their dictionary positions in the ordered OpenRouter list; the free-model pricing predicate is `m.modelId.endsWith(":free") || m.modelId === "openrouter/free"`.
- Docs `models.en.md` / `models.zh.md`: one short paragraph under the gateway groups noting the `:free` variants and the `openrouter/free` router — zero cost, subject to OpenRouter's free-tier rate limits and data policy.

`state.test.ts` and `server/test/models.test.ts` iterate the catalog without hardcoded counts and pass unchanged.

## Part B — landing article + announcement

[...]

PR #63, opened 2026-07-24
feat(core,server,web,cli): mid-run steering messages and square-bracket markers

hiyouga:
## Summary

Two related changes that share one principle: **`session_meta` holds per-session invariants only** — anything a user can change mid-conversation is a per-turn parameter or opens a new session.

### Per-turn thinking level

- `thinking_level` is removed from `SessionMetaPayload`; `createSession`/`resumeSession` no longer record it. **Back-compat:** old traces still carry the field in their meta JSON — `resumeSession` reads it loosely and keeps honoring it as the session's default level (the legacy literal `"default"` and a missing field fall back to the Agent config, as before); rebuilt metas never re-record it. The traces view keeps showing a `thinking_level` row for legacy metas that still carry it.
- The level is no longer frozen into the LLM's `UniConfig` at construction. `GenerativeModel` keeps the construction value as a default and resolves the effective level **per request**: `GenerativeModelParameters.thinkingLevel ?? default` (`openStream` now receives the per-request config).
- `RunOptions.thinkingLevel` threads the level into every LLM request of a run — reconnect retries included; compaction requests deliberately keep the default. `Session.run` forwards it.
- Server: `TaskCreateRequest.thinkingLevel` (optional; validated against the five names, 400 otherwise) → `SessionManager.startTask` → `session.run`.
- Web: in an active session the read-only thinking tag becomes an **editable picker** listing only the levels (per owner review, no "follow config" row): its display initializes to the Agent config's level and auto-follows it while untouched — tasks then omit `thinkingLevel`, so the server/core fallback applies and mid-session Agent-config edits keep taking effect; once the user picks a level it sticks for the session and rides on every subsequent `postTask`. It never writes through to the Agent config (that switch-becomes-default behavior stays draft-only).

### `/model` switch — handoff-style (owner redesign)

The original history-carrying fork was rejected in review and is fully removed: **no history is ever injected into a new session** — some models require thinking payloads and provider `fidelity` byte-for-byte when history is replayed, and that can never cross models, so a "sanitized" replay is not a faithful continuation either. `/model` now works exactly like the `@` handoff:

- Picking a model in the `/model` picker creates a NEW session for the SAME agent via the ordinary session-creation API — **deliberately with the SOURCE session's Workspace** (files the conversation refers to must stay reachable for a real continuation) and the source's approval mode — then posts a first task whose input opens with a `<model_switch_from>` source block (modeled exactly on `handoffMessage`/`parseHandoffMessage`): source session id + title, the absolute path of its latest trace file, the workspace, and the previous `(provider, model_id)` pair. The model reads the source trace (JSONL) itself when it needs the earlier context.
- The trace path is exposed the same way handoff gets its fields — from the Session DTO: new optional `SessionInfo.tracePath` (latest shard, populated on the **single-session GET only**; list rows omit it since locating it costs a directory walk per row).
- The flow mirrors `onHandoff` exactly: create → awaited first `postTask` → navigate; on failure the orphan session is deleted and the composer draft is kept. The remainder text left after the command rides into that first task like a normal send — skills wrapping and images included, with an interface-language auto-line when the body is empty.
- In the new session the block collapses into a "switched model — continued from the earlier conversation" banner (zh+en) linking back to the source session, same interaction as the handoff banner. The new marker is added to core's `stripConversationMarkers` so it never leaks into generated titles. Trace-service and stream-model need nothing special — the block is the task's first user message, exactly like `<handoff_from>` (verified: only `<context_summary>` is special-cased anywhere).
- Removed accordingly: `Agent.forkSession` + `trace/fork.ts` (sanitizer), `POST /api/sessions/:id/fork` + DTOs, `forked_from` in `session_meta`, the meta-based banner/capture, and the fork docs.
- CLI: no interactive command in this PR (the `@` precedent is web-only); a CLI command is a possible follow-up.

Docs updated (zh+en): omni-message, sessions-and-traces (new "/model" section), configuration, models, web-app, server-api, interfaces, agent-loop.

## Review follow-up

- `/model` picker robustness: the slash entry is gated on the model list being loaded; the command **consumes its `/model` token immediately** (same as `/compact`), which at once makes Escape/click-outside/current-model-no-op close without the slash menu re-opening and eliminates the stale open-time token ref (nothing to recompute at pick); the run state is re-checked at pick time (a picker surviving a running/compacting flip is a no-op); the `@` mention menu is suppressed while the picker is open.
- Traces view: a legacy meta's `thinking_level` still renders (loose read) so old traces keep their display; the `forked_from` row is gone with the field.
- `packages/web/e2e/draft.spec.mjs`: the stale `thinking_level`-in-meta and read-only-tag assertions now assert the new behavior (field absent from meta; the session picker shows the config-derived level — "思考等级：高" after the draft wrote "high" through).
- Owner review on the in-session picker addressed (e9bbd8b): the "跟随 Agent 配置" row is gone — the picker initializes to the Agent config's level and auto-follows it until the user pins one.
- Comment drift fixed: `thinking-level.ts` header documents both picker variants (draft write-through vs per-turn override without write-through); `handoff-banner.tsx` header documents all exported banners.

## Rebase note (onto main @ 816a823 — #59/#60/#61 merged)

Rebased cleanly; the one semantic merge point was #60's light-yellow **Free badge** in the chat model dropdown vs this branch's extraction of the shared `ModelMenuList`: the badge now lives **inside `ModelMenuList`**, so BOTH the draft model selector and the in-session `/model` switch picker render it (this closes the standing review caveat that the switch picker lacked the badge). Docs/strings unions (#59 preview edits, #60 free-model copy, #61) merged alongside this branch's additions; verification re-run green after the rebase.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (all 7 packages)
- `pnpm test` — pass (skills 1, docs 4, landing 7, core 28+1 skipped, server 35, cli 14, web 30 test files (post-rebase counts))
  - new/extended: `agent-mentions.test.ts` (`<model_switch_from>` round-trip incl. misdetection guards), `session-title.test.ts` (marker stripped from titles), `session-index.test.ts` (`tracePath` on the single GET only + latest shard; per-task `thinkingLevel` 400), `llm.test.ts` (per-request thinking override), `engine.test.ts` (RunOptions pass-through incl. reconnect; compaction keeps default), `resume.test.ts` (legacy `thinking_level` compat), `agent.test.ts` (meta invariance + subagent level inheritance), `session-manager.test.ts` (startTask pass-through), `stream-model.test.ts`

### Popover mobile fix (`c9954b7`)

The More-settings panel could still cross a phone viewport's right edge: it anchored `left-0` to a mid-toolbar trigger, and the `max-w` clamp caps width, not position — a fixed `w-64` panel overflowed. The panel now sizes to content (`w-max`, 176px cap) with docking **measured on each open** (same pattern as the draft Workspace picker): left-anchored while it fits, otherwise docked to the trigger's right edge with the width capped to the available room — verified on-screen at 320/375px in both draft ("down") and session ("up") directions. The mode pills shrink to the compact control size (h-6, tight padding, sized to their labels; the trigger keeps the toolbar's h-8 icon size), and the always-visible explanation line is removed — the full meaning is hover-only via each pill's `title`, per the toolbar's existing convention.

## Rebase note 2: reconciliation with #64 (`d1407bf`)

Rebased onto main `1e23452` (#64: handoff-style `/model` switch + per-turn thinking level). Conflicts and unions:
- **Marker unification completed**: #64 landed `model_switch_from` in the angle form; the producer (`modelSwitchMessage`) now emits `[model_switch_from]…[/model_switch_from]` and the parser accepts **both** forms (angle kept for traces produced while #64-on-main ran), the same dual-form pattern as every other origin block. Comments/docs/tests reference the square form; the angle form survives only in the dual parser and the legacy regression tests.
- `session-title.ts`: MARKER_TAGS now includes `model_switch_from` and the dual-form stripper covers all tags in both forms.
- Tasks route / `TaskCreateRequest` / `SessionManager.startTask`: union — `thinkingLevel` (from #64) and `queueIfBusy` (this branch) coexist; `/steer` unchanged. **Queued follow-ups store their `thinkingLevel` alongside the input and apply it at auto-start** (the queue entry is now `{input, thinkingLevel?}`), so a level-carrying send queued mid-run is not silently downgraded to the session default.
- Engine/`RunOptions`: #64's per-turn `thinkingLevel` and this branch's steering plumbing merged without conflict; `launchTask` forwards the level into `session.run`.
- chat-input: #64's `/model` command + ModelMenuList + levels-only thinking picker coexist with the More-settings popover and the enabled-while-running input; the slash menu (including `/model`) stays suppressed while running (`!running && !compacting` gating preserved).
- Docs (server-api/web-app/models/sessions-and-traces, zh+en): unions of #64's model-switch/thinking text and this branch's steer/queue rows, with the `/tasks` row now documenting `{input, thinkingLevel?, queueIfBusy?}`.

Verified post-rebase: build + typecheck clean; tests all passing (core 433, server 338, web 394, cli 168, docs 11, skills 16, landing 36); prettier clean.

## Rebase note 3

Rebased onto main `abd5b13` (#62: file tools, per-tool call descriptions, CLI tool-name gutters and diff coloring). Conflicts in `default-config.ts` (the `# Skills` line now keeps main's `read_file` wording **and** the square `[use_skills]` marker), `cli/src/render.ts` (main's `[tool-NNN] name -> output` gutters and diff colors + this branch's steering rendering and input hold — steering now renders from the complete `[user_steering]` user text alongside main's `tool_call` name-map case) and `tool-call-card.tsx` (main's described-tools/file-path preview helpers kept; this branch's obsolete tool-output steering chip dropped, since steering is its own message). Verified: build + typecheck clean, tests all passing (core 482, server 340, web 406, cli 186, docs 11, skills 16, landing 36), prettier clean.

## Follow-up: composer layout + marker consolidation

**Input area (`252c180`)** — the running-state toolbar overflowed on phones and carried two action buttons. It is now one `justify-between` row: the settings controls (attach / approval / skills / more-settings / hint) sit left in a `min-w-0` horizontally scrolling group, and status + model + the action button sit right and never shrink, so a narrow viewport scrolls the left group instead of pushing the button off-screen; the context ring's numbers hide below `@md` (the title keeps the exact usage), which is what the 320px running state needed. **Stop and Send are one button**: while running, an empty composer is Stop (abort) and any content turns the same button into Send — steering or queue-as-follow-up per the remembered mode, 409 fallback unchanged — with icon, title, aria-label and disabled state switching on content. Idle keeps the ordinary send button.

**Markers (`8b3e423`)** — every special-marker producer/parser now lives in `packages/core/src/omnimessage/markers/`, exported through the omnimessage barrel and as its own `@prismshadow/penguin-core/markers` subpath (web resolves core only via subpaths):
- `block.ts` — the canonical `[tag]…[/tag]` builder plus the dual-form (square + legacy angle) matcher/stripper, so the compatibility policy exists once instead of a regex pair per call site;
- `tags.ts` — the tag list and the title-noise subset;
- `engine-blocks.ts` — `[turn_aborted]`/`[turn_retried]` transcripts, the `[context_summary]` wrapper, `extractSummary`, `unwrapSyntheticBlock`;
- `origin-blocks.ts` — `[use_skills]` / `[handoff_from]` / `[scheduled_task]` / `[model_switch_from]` build+parse with their origin types;
- `steering.ts` — `[user_steering]` (moved).

Call sites import from it: the engine and Trace replay (replay no longer deep-imports the engine just to reach `extractSummary`, and the duplicated `[context_summary]` template is gone), `session-title`, the web `skill-use`/`agent-mentions` modules (re-exporting under their existing names, keeping their UI-only helpers), and the server scheduler — whose producer now shares one definition with the web parser that reads it. Behavior is unchanged including dual-form compat; marker tests moved to `packages/core/test/markers.test.ts` (legacy-form regressions included), with the web tests keeping the wiring checks.

Verified after both: build + typecheck clean, tests all passing (core 492, server 340, web 383, cli 186, docs 11, skills 16, landing 36), prettier clean.

### Dropdown clipping fix (`0d5d0ce`)

The scrolling left group introduced above turned out to be a clipping ancestor: setting one axis to a non-`visible` overflow makes the other compute to `auto`, so the panels anchored inside it were clipped and the upward-opening ones disappeared entirely on phones. Rather than reverting the layout, the panels now escape the container: `Dropdown` gained an opt-in `portal` mode following the pattern `use-portal-panel` already established for `OptionMenu` — the panel mounts on `document.body` at fixed viewport coordinates measured from the trigger. It flips direction when the requested side lacks room, caps its height to the space actually available (scrolling internally rather than running off-screen), clamps both edges into the viewport, and **re-places instead of closing** when something scrolls, since tapping a partially visible control scrolls the toolbar and must not dismiss the panel it just opened. Measurement uses the untransformed offset box — the pop-in scale animation otherwise reports a narrow panel and under-clamps the left edge. All five composer pickers moved onto it and dropped their hand-tuned width clamps and measured docking.

Verified with the e2e harness (`packages/web/e2e/layout.spec.mjs`), which grew a real clipping assertion — hit-testing a point inside each panel's visible area, since an on-screen rect alone doesn't prove the panel is painted — and now covers the session state at **320 and 375** plus a genuine **running** state (a pending approval parks the run): every enabled picker opens, is fully visible and unclipped, the page never scrolls sideways, and the toolbar reports no horizontal overflow with the merged Stop/Send button reachable. The harness caught two real defects in the first cut of the fix (the transform-skewed width measurement and the scroll-closes-panel race).


packages/cli/src/i18n.ts
```diff
@@ -475,9 +481,11 @@ const zh: Messages = {
 
   header,
   chatHints: () =>
-    "输入消息发起对话；行尾 \\ 续行；/compact 压缩上下文；/exit 退出；Ctrl-C 中断对话。",
+    "输入消息发起对话；行尾 \\ 续行；运行中输入可插话引导；/compact 压缩上下文；/exit 退出；Ctrl-C 中断对话。",
```
hiyouga, line 517: 用户在输入文字时候应该锁住cli界面

packages/cli/src/i18n.ts
```diff
@@ -475,9 +481,11 @@ const zh: Messages = {
...
   confirmExit: () => "确认退出 penguin？[y/N] ",
   taskInterrupted: () => "[已中断当前对话]",
+  steerQueued: () => "» 插话已排队（随下一个工具结果送达）",
+  steerLinePrefix: () => "↪ 用户: ",
```
hiyouga, line 521: 这个现在没显示

packages/cli/src/i18n.ts
```diff
@@ -475,9 +481,11 @@ const zh: Messages = {
 
   header,
   chatHints: () =>
-    "输入消息发起对话；行尾 \\ 续行；/compact 压缩上下文；/exit 退出；Ctrl-C 中断对话。",
+    "输入消息发起对话；行尾 \\ 续行；运行中输入可插话引导；/compact 压缩上下文；/exit 退出；Ctrl-C 中断对话。",
   confirmExit: () => "确认退出 penguin？[y/N] ",
   taskInterrupted: () => "[已中断当前对话]",
+  steerQueued: () => "» 插话已排队（随下一个工具结果送达）",
```
hiyouga: 这个要显示steer的文本内容

packages/core/src/engine/context-engine.ts
```diff
@@ -257,6 +270,62 @@ export class ContextEngine {
...
+    }
+  }
+
+  /**
+   * Queues a steering message for the running Task: it is appended to the next completed
+   * tool output as a `[user_steering]` block, or — if the turn ends without tool calls
+   * first — delivered as the next user turn. Returns false when no Task is running (the
+   * host should then submit the text as a normal task instead).
+   */
+  steer(text: string): boolean {
```
hiyouga, line 318: 记得考虑如果是没有工具输出时候怎么做？我觉得不要黏到工具输出后面了，就单独做个Usertext，随着工具调用一起发吧。

hiyouga:
## Review: REQUEST CHANGES

Reviewed `git diff origin/main...` (base `3f868ac`) — 63 files, +1255/-274.

The engine's steering delivery is largely correct and genuinely well-tested — the atomic drain, the trace/stream/next-input consistency, and the dual-form parsers all hold up. Three things block: a multi-second silent-loss window at the compaction checkpoint, existing agents receiving new markers their prompt never explains, and the absent `design/specs/` update.

---

### Blocking

**1. `packages/core/src/engine/context-engine.ts:452-506` — a Task-boundary compaction silently swallows steering queued *during* the compaction request, while `steer()` returns `true` and the server answers 202.**

The sequence, verified by reading the loop:

- `:452` — `const steeringNext = midTask ? [] : this.drainSteering().map(...)` drains the queue.
- `:454` — `const atBoundary = !midTask && steeringNext.length === 0` is computed from that drain.
- `:466` — `yield* this.summarizeContext(...)` runs a **full LLM request**. `taskRunning` is still `true` for its entire duration (cleared only in `run`'s `finally`, `:278-281`), so `steer()` accepts.
- `:483-487` — on completion, tests the now-**stale** `atBoundary` and returns.
- `run`'s `finally` at `:280` then sets `this.steeringQueue = []`.

Failure scenario: the agent finishes its answer, context is over `compaction.maxContextLength`, summarization takes ~10s. The user — UI still showing "running" — types "actually, use TypeScript" → `POST /steer` → **202** → input cleared → the message is discarded and appears in no Trace, no stream, and no next turn.

The same shape exists in `discard` mode at `:461-464` (shorter window: `discardContext` awaits `startNewContext()` file I/O). The documented contract at `:236-243` only excuses "a plain race with completion" — this is a seconds-long window. Fix is a re-drain after the compaction result.

**2. Existing agents receive the new markers while their persisted prompt documents the old ones, and are never told about `[user_steering]` at all.**

`packages/core/src/state/agent-state.ts:127-145` loads `system_config.yaml` verbatim when it exists; `defaultSystemConfig()` is reached only on the init branch. `grep -rn "migrat" packages/core/src/state` finds only `project-config.ts`'s explicit "no migration" note. So every agent created before this PR keeps a `system_prompt` describing `<turn_aborted>` / `<turn_retried>` / `<context_summary>` / `<use_skills>` / `<developer_instructions>` — while `context-engine.ts:957,1157,1189` and `web/src/features/chat/skill-use.ts:29` now emit the square-bracket forms.

Failure scenario: an existing agent runs a long shell command; the user steers; the model receives `tool result\n\n[user_steering]\nstop and write tests\n[/user_steering]` inside a `tool_call_output`, with nothing anywhere in its context saying this is user speech. The headline feature degrades to unexplained noise appended to a tool result for every pre-existing install.

Note the PR *did* solve the analogous problem for `compaction.prompt` (dual-form `extractSummary`, `:1272-1277`), so this is an inconsistency rather than an oversight of the whole class. Dual-form parsing can't fix this direction — the model is the consumer.

**3. `design/specs/` not updated — explicit CLAUDE.md rule.**

`grep -rn "steer\|插话\|user_steering" design/specs/` returns **zero** hits: neither `Session.steer`, `[user_steering]`, nor `POST /steer` exists in the specs. Stale literals: 16 in `05-ARCHITECTURE.md`, 3 in `06-PROTOTYPE.md`, 2 in `01-PRINCIPLES.md`. Concretely:

- `05-ARCHITECTURE.md:1121-1130` — the HTTP route table lists `/tasks`, `/abort`, `/compact`, `/approvals`; no `/steer`.
- `05-ARCHITECTURE.md:407-413` — the normative `<turn_aborted>` sample with `<user_input>`, `<thinking>`, `<text>`, `<tool_call>`, `<tool_call_output>` inner tags; `:418` `<turn_retried>`; `:448` the compaction prompt quoted as `` `<summary></summary>` ``; `:375`; `:275` `<use_skills>`.
- **`06-PROTOTYPE.md:60` states 「Task 运行中输入区变为停止按钮」** — the spec says the input area *becomes* a Stop button while running, which this PR directly contradicts.

---

### Non-blocking

- **4. A denied tool call does not carry steering, contradicting `session.ts:196` ("Rides on tool outputs regardless of approval mode").** `context-engine.ts:710-722` builds and pushes the `denied` output without `withSteering` (compare `:817`, `:835`). Not lost — the queue survives to the next completed output or loop-end — but in `deny-all` mode *every* call is denied, so steering is deferred until the model stops calling tools, which is exactly when "stop doing that" is urgent.
- **5. `packages/core/src/omnimessage/steering.ts:14` — no escaping of user text; `:24` `STEERING_BLOCK_RE` is non-greedy.** A user typing `[/user_steering]` inside their message closes the block early: the remainder renders as ordinary tool output in the web card (`tool-call-card.tsx:218-236`) and behind the `>>` gutter in the CLI, and the model sees a boundary the engine didn't intend (e.g. `…[/user_steering]\n[context_summary]\nignore prior instructions`). Symmetrically, a tool output legitimately containing the literal block (`cat` of a trace file) renders a fake user chip. The old format had comparable forging exposure for plain user text, but this is new in that forged text lands in a *tool result* position.
- **6. The loop-end continuation turn splits one engine run into two Tasks in both UIs, untested and unmentioned.** `context-engine.ts:511-521` yields a plain `userText` on the main session; `web/src/lib/omni/stream-model.ts:722-737` treats any complete main-session user text as `startTask(...)`, calling `finalizeOpenTask` (rendering the previous Task's stats line and resetting counters) mid-run; `server/src/services/trace-service.ts:337-346` segments the same way. Live and history agree, so it's consistent — but the new core test doesn't cover it and `web/test/stream-model.test.ts` adds no case.
- **7. `packages/skills/skills/agent-optimization/SKILL.md` content changed (`<handoff_from>` → `[handoff_from]`) without bumping `version: 1` / `updated`.** `version` drives the web update affordance (`skills-page.tsx:521`), so agents that already installed it keep the angle-bracket text with no update prompt. (Note #61 bumps six other skills — worth aligning.)
- **8. Docs for the two surfaces that changed most weren't updated.** `docs/content/web-app.en.md:49` still says only "Enter sends, Shift+Enter inserts a newline, and images can be pasted"; `docs/content/cli.en.md:48-53` documents running-state keys without mentioning that typing now steers. (`agent-loop.*` and `server-api.*` *were* updated, en/zh in sync.) Also `landing/content/blog/simple-harness-is-all-you-need.{en,zh}.md:69` still says `<context_summary>` — arguably a frozen historical post.
- **9. Web gives no confirmation that steering was queued.** `chat-input.tsx:1032-1035` clears the textarea on success and that's the only signal; nothing appears until the next completed tool output, which can be minutes. The CLI does get a dim ack. Related: the 409 fallback in `chat-page.tsx:421-438` posts a **text-only** task, silently dropping selected skills/images that `chat-input.tsx:1025-1027` promises "stay for a normal send".
- **10. `packages/cli/src/commands/chat.ts:228-232` silently drops the line when `session.steer()` returns false** (race with task end), even though `chatHints()` now advertises that typing while a task runs steers the agent. No fallback, no message — unlike the web's 409 path.
- **11. Narrow start-up race:** `server/src/runtime/session-manager.ts:394` sets `entry.status = "running"` and `:400` publishes `state: running` before the generator is first driven, while `taskRunning` only flips inside `Session.run` after `await this.ensureMetaWritten()`. A steer landing in that gap gets 409 → web falls back to `onSend` → `assertIdle` throws → error toast. Sub-millisecond in practice.
- **12. No length cap on `/steer`'s `text`** (`routes/sessions.ts:370-376`) — but `POST /tasks` has none either, so this matches existing convention rather than regressing it.

---

### Verified as claimed

**Every regex was checked for the `[` metacharacter trap** — this was the most likely place for a silent bug and it's clean. `extractSummary` (`context-engine.ts:1275-1276`), `unwrapSyntheticBlock` (`:1229-1231`), `stream-model.ts:723` (prefix `startsWith`, both forms), `skill-use.ts:42-44`, `agent-mentions.ts:121-123,169-171`, `session-title.ts:47-52` (`new RegExp` with `\\[`/`\\]`), and `steering.ts:24`. No unescaped bracket anywhere.

**Producers emit only the new form** — grepping `pr/63 -- packages/*/src` for all twelve angle-bracket literals returns zero producer hits; what remains is deliberate "old form still recognized" prose in docs plus the stale landing blog. Angle-bracket *value* placeholders (`<project_dir>`, `<agent_id>`, `<session_id>`) are correctly untouched and called out as notation at `default-config.ts:116`.

**Delivery is correct at the complete `tool_call_output`**: `:809-824` is `isCompleteOutput` → `withSteering` → `queue.push` → `await this.write` → `toolOutputs.push`, so one rewritten message reaches Trace, SSE and next-turn input. **No duplication across parallel tool calls**: `drainSteering` (`:296-302`) and `withSteering` (`:304-322`) contain no `await`, so whichever concurrent `executeOne` gets there first takes the whole queue atomically and the rest see `[]` — correct by construction, though not covered by a test. Steering does ride on errored/timed-out tools (`:833-841`). Ordering is preserved. Abort discards the queue (`:278-281` plus a test).

Max-turns interaction is right: `:374-383` stores the steering `userText` as `pendingCarryOver`, resent on the next `run`, deliberately not re-written to Trace (`:343-351`) so there's no duplicate record. The continuation is both traced and yielded (`:515-518`, `:496-499`).

Server: `sessions.ts:370-376` calls `resolveSession` (404s via `requireProjectAccess`) before `manager.steer`; `session-manager.ts:466-479` throws typed 409 `not_running` when the entry is absent, not running, or `steer()` returns false. Covered by `server/test/steer.test.ts`.

Web: the 409 fallback can't double-send — a 409 means the server never called `steer`, and the input clears only on `ok`. It follows #54's one-notification rule (`chat-page.tsx:431` uses `toastError(apiErrorText(...))`, identical to the five sibling handlers). No @-handoff regression: `agent-mentions.ts` changes are producer literals and dual-form parsers only, `splitLeadingMention` untouched, and with an @ target `canSteer` is false so Enter does nothing — same as pre-PR. The steering chip is escaped (React child, no `dangerouslySetInnerHTML`).

CLI renders steering once and never repeats streamed output: `render.ts:513-518` handles the complete output only to call `renderSteering` (`:766-782`), which filters to `kind === "steering"` and returns early when empty.

`stripConversationMarkers` strips both forms, and `[user_steering]` can't reach title material anyway (`session.ts:165-176` collects user text from `newMessages`; engine-yielded messages are scanned for `role: "assistant"`).

zh/en parity: 4 new keys in both web catalogs, 2 in both CLI locales, all 6 changed doc pairs symmetric.

---

### Cross-PR notes

- **Conflicts with #61 on `packages/core/src/state/default-config.ts`, and the resolution needs care.** Both PRs edit the `# Skills` line in different places: #61 changes `<project_dir>/agents/<agent_id>/…` → `<agents_dir>/<agent_id>/…`, this PR changes `<use_skills>` → `[use_skills]`. Taking either side wholesale silently reverts the other's goal. Correct result keeps both. (The rest of the prompt was divided cleanly — this PR correctly leaves `# File system` alone and #61 correctly leaves `# System markers` alone.)
- Conflicts with **#62** on `core/test/engine.test.ts` and `web/src/features/chat/tool-call-card.tsx`; with **#64** on `docs/content/server-api.{en,zh}.md` and `server/src/http/routes/sessions.ts` (both insert a route before the approvals route — keep both).
- **#63 and #64 together change the persisted trace format in one release** — this PR changes marker literals inside message text, #64 drops `thinking_level` from `session_meta`. Each tested its own migration in isolation; worth reading one genuinely old trace through both before shipping.

hiyouga:
应该让用户选择是steer还是no-steer，如果是no-steer，就要等结束后再自动发一条新的follow up message

hiyouga:
现在对话界面在手机端有问题，按钮会溢出。把插话排队的选项放到对话框下面吧，先做一个弹出分组，存放一些更多设置，之后还有一些开关选项也会放到这里面。

hiyouga:
这个弹出分组在对话前也可以设置并且记忆

hiyouga:
已按意见调整（e5d6f08）：移除了发送键旁的内联「插话/排队」切换（手机端溢出的来源），改为输入框工具条上的「更多设置」弹出分组——设置行式布局，后续开关选项可继续放入；沿用工具条现有下拉的向上展开与视口收敛，窄屏不再溢出。草稿态同样可设置，选择通过 localStorage（penguin.steerMode，与侧栏分组方式同一机制）记忆，运行中发送直接遵循记忆的方式；「N 条已排队」提示保留在输入框附近。全链路验证通过。

hiyouga:
现在这个弹出框在手机显示会超出屏幕宽度，这个按钮太大了，也要小一点，说明文字放到hover出来的文字，不要默认显示

hiyouga:
已修复（c9954b7）：溢出根因是面板以工具条中部的按钮为锚点向左对齐，max-width 只约束宽度不约束位置，固定 w-64 会越过屏幕右缘。现改为面板按内容收窄（上限 176px），并在每次展开时测量停靠方向（与草稿态 Workspace 选择器同一套逻辑）——放得下就左对齐，放不下改右对齐并按剩余空间封顶，320/375px 下草稿态与会话态均完整在屏。模式按钮缩小为紧凑规格（h-6、按文字收窄），触发按钮保持工具条 h-8 图标尺寸；默认不再显示说明文字，完整含义移入按钮 hover（title）。全链路验证通过。

hiyouga:
已与 #64 会师（d1407bf）：model_switch_from 统一为方括号——产出端只发 [model_switch_from]，解析端双格式兼容存量（含 #64 在 main 上运行期间产生的尖括号 Trace），与其余源块同一模式；标题剥离的 MARKER_TAGS 全表双格式。路由与引擎合并完成：/tasks 同时携带 thinkingLevel 与 queueIfBusy，排队跟进消息随条保存 thinkingLevel 并在自动发出时生效；输入区 /model 命令、思考等级拾取器与「更多设置」弹出分组共存，运行中 slash 菜单（含 /model）仍被抑制。全链路验证通过（core 433 / server 338 / web 394 / cli 168）。

packages/core/src/omnimessage/markers/steering.ts
hiyouga, line 1: 可以把一些特殊 marker 的解析都放到一个文件或者文件夹里面

hiyouga:
按钮应该和标题在同一行，形成左右两边的布局。现在对话过程中，手机端web的按钮还是会溢出屏幕，修一下。停止按钮和发送按钮可以合并一下，如果用户输入消息，则从停止按钮变发送按钮，如果输入框是空的，就是停止按钮

hiyouga:
两点均已处理：

1. 输入区布局（252c180）：工具条改为一行左右分组——设置类控件在左（窄屏内部横向滚动，不再把右侧挤出屏幕），状态/模型/操作按钮在右且不收缩；上下文环的数字在 @md 以下隐藏（悬停仍显示精确值），这是运行态在 320px 下所需的余量。停止与发送已合并为一个按钮：运行中输入框为空时是停止，一旦有内容即变为发送（按记忆的插话/排队方式发出，409 回退逻辑不变），图标、标题、aria 与禁用态随输入内容切换。

2. Marker 解析集中（8b3e423）：所有特殊 marker 的产出与解析统一到 packages/core/src/omnimessage/markers/——block.ts（方括号规范形式 + 双格式兼容匹配/剥离，兼容策略只写一处）、tags.ts（标签表与标题噪声子集）、engine-blocks.ts（turn_aborted/turn_retried 转写、context_summary、extractSummary、unwrapSyntheticBlock）、origin-blocks.ts（use_skills/handoff_from/scheduled_task/model_switch_from）、steering.ts（user_steering）。引擎、Trace 回放（不再为 extractSummary 深引 engine，context_summary 模板重复也消除）、session-title、Web 的 skill-use/agent-mentions（按原名再导出，仅保留 UI 专属逻辑）、服务端 scheduler（产出端与 Web 解析端现在共用同一定义）均改为从该模块引入；行为不变（含双格式兼容），测试迁至 packages/core/test/markers.test.ts 并保留历史格式回归。

全链路验证通过（core 492 / server 340 / web 383 / cli 186）。

hiyouga:
现在手机端对话时候的弹出选项都看不到了，你仔细看看

hiyouga:
已修复（0d5d0ce）。根因：上一版把工具条左侧分组设为 overflow-x-auto，CSS 中一个轴设为非 visible 会让另一个轴也变成 auto，于是锚定在该容器内的下拉面板被裁切——向上展开的更是整块看不见。

修法不是回退布局，而是让面板脱离该容器：Dropdown 新增 portal 模式（沿用 OptionMenu 已有的 use-portal-panel 思路），面板挂到 document.body、按触发按钮实测坐标以 fixed 定位；空间不足时自动翻转方向、按可用高度封顶（内部滚动而非溢出屏幕）、左右边缘收敛进视口；滚动时重新定位而不是关闭（手机上点击半露的控件会带动工具条滚动，不能把刚打开的面板关掉）。测量使用未受 transform 影响的 offset 盒——弹出缩放动画期间的 rect 会偏窄，导致左边缘收敛不足。五个输入区选择器全部改用该模式，原先手工调的宽度 clamp 与测量停靠逻辑一并移除。

验证：扩展了 packages/web/e2e/layout.spec.mjs——新增真正的裁切断言（在面板可见区域内做命中测试，仅靠 rect 在视口内并不能证明被绘制），覆盖会话态 320/375 两个宽度，以及真实运行态（always-ask 会话停在待审批）；每个可用选择器都能打开且完整可见，页面不横向滚动，工具条无溢出、合并后的停止/发送按钮可达。该测试在修复过程中还抓出了两个真实缺陷（缩放动画导致的宽度误测、滚动误关面板）。

PR #64, opened 2026-07-24
Handoff-style /model switch command and per-turn thinking level

hiyouga:
## Summary

Two related changes that share one principle: **`session_meta` holds per-session invariants only** — anything a user can change mid-conversation is a per-turn parameter or opens a new session.

### Per-turn thinking level

- `thinking_level` is removed from `SessionMetaPayload`; `createSession`/`resumeSession` no longer record it. **Back-compat:** old traces still carry the field in their meta JSON — `resumeSession` reads it loosely and keeps honoring it as the session's default level (the legacy literal `"default"` and a missing field fall back to the Agent config, as before); rebuilt metas never re-record it. The traces view keeps showing a `thinking_level` row for legacy metas that still carry it.
- The level is no longer frozen into the LLM's `UniConfig` at construction. `GenerativeModel` keeps the construction value as a default and resolves the effective level **per request**: `GenerativeModelParameters.thinkingLevel ?? default` (`openStream` now receives the per-request config).
- `RunOptions.thinkingLevel` threads the level into every LLM request of a run — reconnect retries included; compaction requests deliberately keep the default. `Session.run` forwards it.
- Server: `TaskCreateRequest.thinkingLevel` (optional; validated against the five names, 400 otherwise) → `SessionManager.startTask` → `session.run`.
- Web: in an active session the read-only thinking tag becomes an **editable picker** listing only the levels (per owner review, no "follow config" row): its display initializes to the Agent config's level and auto-follows it while untouched — tasks then omit `thinkingLevel`, so the server/core fallback applies and mid-session Agent-config edits keep taking effect; once the user picks a level it sticks for the session and rides on every subsequent `postTask`. It never writes through to the Agent config (that switch-becomes-default behavior stays draft-only).

### `/model` switch — handoff-style (owner redesign)

The original history-carrying fork was rejected in review and is fully removed: **no history is ever injected into a new session** — some models require thinking payloads and provider `fidelity` byte-for-byte when history is replayed, and that can never cross models, so a "sanitized" replay is not a faithful continuation either. `/model` now works exactly like the `@` handoff:

- Picking a model in the `/model` picker creates a NEW session for the SAME agent via the ordinary session-creation API — **deliberately with the SOURCE session's Workspace** (files the conversation refers to must stay reachable for a real continuation) and the source's approval mode — then posts a first task whose input opens with a `<model_switch_from>` source block (modeled exactly on `handoffMessage`/`parseHandoffMessage`): source session id + title, the absolute path of its latest trace file, the workspace, and the previous `(provider, model_id)` pair. The model reads the source trace (JSONL) itself when it needs the earlier context.
- The trace path is exposed the same way handoff gets its fields — from the Session DTO: new optional `SessionInfo.tracePath` (latest shard, populated on the **single-session GET only**; list rows omit it since locating it costs a directory walk per row).
- The flow mirrors `onHandoff` exactly: create → awaited first `postTask` → navigate; on failure the orphan session is deleted and the composer draft is kept. The remainder text left after the command rides into that first task like a normal send — skills wrapping and images included, with an interface-language auto-line when the body is empty.
- In the new session the block collapses into a "switched model — continued from the earlier conversation" banner (zh+en) linking back to the source session, same interaction as the handoff banner. The new marker is added to core's `stripConversationMarkers` so it never leaks into generated titles. Trace-service and stream-model need nothing special — the block is the task's first user message, exactly like `<handoff_from>` (verified: only `<context_summary>` is special-cased anywhere).
- Removed accordingly: `Agent.forkSession` + `trace/fork.ts` (sanitizer), `POST /api/sessions/:id/fork` + DTOs, `forked_from` in `session_meta`, the meta-based banner/capture, and the fork docs.
- CLI: no interactive command in this PR (the `@` precedent is web-only); a CLI command is a possible follow-up.

Docs updated (zh+en): omni-message, sessions-and-traces (new "/model" section), configuration, models, web-app, server-api, interfaces, agent-loop.

## Review follow-up

- `/model` picker robustness: the slash entry is gated on the model list being loaded; the command **consumes its `/model` token immediately** (same as `/compact`), which at once makes Escape/click-outside/current-model-no-op close without the slash menu re-opening and eliminates the stale open-time token ref (nothing to recompute at pick); the run state is re-checked at pick time (a picker surviving a running/compacting flip is a no-op); the `@` mention menu is suppressed while the picker is open.
- Traces view: a legacy meta's `thinking_level` still renders (loose read) so old traces keep their display; the `forked_from` row is gone with the field.
- `packages/web/e2e/draft.spec.mjs`: the stale `thinking_level`-in-meta and read-only-tag assertions now assert the new behavior (field absent from meta; the session picker shows the config-derived level — "思考等级：高" after the draft wrote "high" through).
- Owner review on the in-session picker addressed (e9bbd8b): the "跟随 Agent 配置" row is gone — the picker initializes to the Agent config's level and auto-follows it until the user pins one.
- Comment drift fixed: `thinking-level.ts` header documents both picker variants (draft write-through vs per-turn override without write-through); `handoff-banner.tsx` header documents all exported banners.

## Rebase note (onto main @ 816a823 — #59/#60/#61 merged)

Rebased cleanly; the one semantic merge point was #60's light-yellow **Free badge** in the chat model dropdown vs this branch's extraction of the shared `ModelMenuList`: the badge now lives **inside `ModelMenuList`**, so BOTH the draft model selector and the in-session `/model` switch picker render it (this closes the standing review caveat that the switch picker lacked the badge). Docs/strings unions (#59 preview edits, #60 free-model copy, #61) merged alongside this branch's additions; verification re-run green after the rebase.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (all 7 packages)
- `pnpm test` — pass (skills 1, docs 4, landing 7, core 28+1 skipped, server 35, cli 14, web 30 test files (post-rebase counts))
  - new/extended: `agent-mentions.test.ts` (`<model_switch_from>` round-trip incl. misdetection guards), `session-title.test.ts` (marker stripped from titles), `session-index.test.ts` (`tracePath` on the single GET only + latest shard; per-task `thinkingLevel` 400), `llm.test.ts` (per-request thinking override), `engine.test.ts` (RunOptions pass-through incl. reconnect; compaction keeps default), `resume.test.ts` (legacy `thinking_level` compat), `agent.test.ts` (meta invariance + subagent level inheritance), `session-manager.test.ts` (startTask pass-through), `stream-model.test.ts`


packages/core/src/omnimessage/types.ts
```diff
@@ -86,14 +92,18 @@ export interface SessionMetaPayload {
...
   /** Absolute path to the Workspace. */
   workspace: string;
   /** Session origin: spawned by a subagent / triggered by a scheduled task; absent = user-created. */
   source?: "subagent" | "schedule";
+  /**
+   * Present when this Session was forked from another Session (the model-switch command):
+   * the source session_id. The forked Session carries the source conversation as real history
+   * (sanitized — no thinking payloads, no provider fidelity) and continues on its own model.
+   */
+  forked_from?: string;
```
hiyouga: 这个不要加 forked from 了，和 @ 指派一样，就只是在第一条对话里面加一下历史 Session id。然后也不要去继承历史。

packages/core/src/trace/fork.ts
```diff
@@ -0,0 +1,54 @@
...
+ *     the run_subagent tool text remains part of the conversation).
+ * Everything else — request_begin/request_end pairs, abort events, compaction events, and
+ * all other complete model_msgs — is kept, so the forked record list stays a well-formed,
+ * resumable, renderable Trace with correct task/turn attribution.
+ */
+import { isCompleteModelMessage, isEventMessage, isSessionMeta } from "../omnimessage/index.js";
+import type { OmniMessage } from "../omnimessage/index.js";
+
+/** Payload types dropped wholesale: thinking content never replays across models. */
+const DROPPED_MODEL_TYPES = new Set(["thinking", "inline_thinking"]);
```
hiyouga: 不要这么做，有些模型必须加Thinking和fidelity，还是按照 @ handoff的方式照常渲染就行

packages/core/src/agent.ts
```diff
@@ -487,6 +515,185 @@ export class Agent {
     });
   }
 
+  /**
+   * Fork a Session onto another model: creates a NEW Session for the same Agent that
+   * **carries the source Session's conversation as real history** (not a reference note)
```
hiyouga: 不要 carry history，就和 handoff 一样，只给 session 路径，需要时候让他去读文件

hiyouga:
## Review: REQUEST CHANGES

Reviewed `git diff origin/main...` (base `3f868ac`) — 58 files, +1609/-284.

The per-turn thinking-level half is clean and genuinely well-tested — I traced it end to end including the reconnect path and found no defect. The fork's record sanitization is also sounder than I expected: I specifically tried to construct a sequence where dropping thinking leaves a malformed request, and could not. Three things block.

---

### Blocking

**1. `packages/core/src/trace/fork.ts:23` + `packages/core/src/agent.ts:652,672` — the fork carries image content from the source history into a target model that may not support vision.**

`sanitizeForkRecords` drops `session_meta`, `thinking`, `inline_thinking`, `token_usage`, `subagent`, and strips `fidelity`. It **keeps** `text`, `tool_call`, `tool_call_output`, **`image_url`**, **`inline_data`**, `request_begin`/`request_end`, `abort`, `approval_decision`, `compaction_begin`/`compaction_end`.

So image payloads survive into `resumeTrace`'s `history` (role `user`), `agent.ts:652` calls `rt.llm.setHistory(resumed.history)`, and `payloadToContentItem` (`generative-model.ts:122-131`) emits `{type:"image_url", image_url}` / `{type:"inline_data", …}` straight to the provider.

The `modelEntry.vision === false` branch at `agent.ts:672` only sets `inputImagesDir`, and `Session.run` applies that exclusively to **new** input (`session.ts:158`: `newMessages = await imagesToScratchpadPaths(newMessages, this.inputImagesDir)`). Injected history is never touched. `session.ts:63` states the contract this breaks verbatim: *"images never enter the session history directly (some providers return a 400 outright on image input)"* — and `agent.ts:854` repeats it: *"images are never allowed into that session's history."* The codebase already knows a non-vision model needs the `describe_image` / `vision_model` detour (`agent.ts:856`); the fork just never applies it to carried-over history.

Failure scenario: chat with Claude, paste a screenshot — the route turns it into an `image_url` OmniMessage (`routes/sessions.ts:91-96`) and `Session.run` writes it to the trace unmodified because `inputImagesDir` is unset for a vision model. Run `/model` and pick an entry with `vision: false`. `forkSession` copies the `image_url` record into the fork's trace and injects it via `setHistory`. The fork's first request carries an image block to a model that rejects image input → hard 400.

Worse than a one-off: the image is now baked into the fork's own trace file, so `resumeSession` replays it on every subsequent load. **The forked session is permanently unusable and can't be recovered without hand-editing the JSONL.** Same applies to `tool_call_output.images` (e.g. `read_image` output), which `payloadToContentItem:159-160` forwards as `tool_result.images`.

Note this composes badly with #60, which adds two `supportsVision: false` models to the catalog — both are plausible `/model` targets.

**2. `design/specs/` not updated — explicit CLAUDE.md rule.**

`design/` is a gitignored symlink to the separate design repo, so this can never appear in the PR diff. Checked that repo directly (HEAD `5a557c4`, clean tree):

- `05-ARCHITECTURE.md:603` — the canonical `session_meta` payload still lists `"thinking_level": "<thinking_level>"`, the exact field this PR deletes. No `forked_from`.
- `grep -rni "fork|分叉" design/specs/` returns **zero hits**. `Agent.forkSession`, `POST /api/sessions/:sessionId/fork`, the sanitization rules, and `forked_from` are entirely undocumented.
- `05-ARCHITECTURE.md:126` (`GenerativeModelParameters`) and `:97` (RunOptions prose) have no `thinkingLevel`; `:1118-1129` (route table) has no `/fork`; `:63-95` (Agent SDK prototype) has no `forkSession`.
- `06-PROTOTYPE.md:60,78` still assert that changing the thinking level inside a conversation writes back to Agent settings — now false for an active session — and list only `/compact` as a slash command.

**3. `packages/web/e2e/draft.spec.mjs:156-157` — two assertions this PR breaks, in a file it doesn't touch.**

- `:156` `expect(meta?.payload?.thinking_level).toBe("high")` now reads `undefined` — `createSession` no longer records it (`agent.ts:310-323`), and `agent.test.ts:281` asserts `"thinking_level" in payload === false`.
- `:157` `await expect(page.getByTitle("思考等级：高")).toBeVisible()` no longer matches. It asserted the read-only tag the PR replaced with a picker. The picker's title is `` `${S.chat.thinkingLevel}：${label}` `` (`chat-input.tsx:463`); in the session variant `value` defaults to `""`, `thinkingLevelLabel` returns `null` (`thinking-level.ts:33-37`), so the rendered title is `思考等级：跟随 Agent 配置`.
- The comment at `:148-150` is stale on both counts.

Mitigating: the web e2e isn't CI-gated (`pnpm test:e2e` maps to core only, `package.json:15`, `ci.yml:57`). `layout.spec.mjs:313-321` is safe.

---

### Non-blocking

Ranked most to least severe.

1. **Anthropic thinking-block replay risk after dropping thinking** (`fork.ts:23`). The fork can produce an injected history whose last assistant turn is `tool_use` with no thinking block, immediately followed by `tool_result`. `fork.test.ts` "keeps carry-over semantics" exercises exactly this shape. With thinking enabled on the target — the fork's default comes from the Agent config (`agent.ts:593`) — Anthropic rejects a `tool_result` whose preceding assistant turn lacks its thinking blocks. Verified the code produces the sequence; could not execute a provider request to confirm the rejection. Most relevant to a Claude→Claude fork, a likely `/model` use.
2. **`chat-page.tsx:399-406` — typed text is unrecoverable if the fork's first task fails.** `discardSessionDraft()` runs, the fork returns `true`, `ChatInput` clears (`chat-input.tsx:1017-1020`) and is remounted by the session-keyed wrapper, while the un-awaited `postTask` may reject with only a toast. Compare `onHandoff` (`:441-454`), which awaits, returns `false`, and deletes the orphan session. The fire-and-forget is deliberate; the text loss isn't.
3. **`handoff-banner.tsx:47-60` — a dangling `forked_from` navigates the user into an unrelated conversation.** `ForkedBanner` renders unconditionally and does `navigate(\`/chat/${sessionId}\`)`. If the source was deleted, the probe effect (`chat-page.tsx:197-212`) 404s and sets `probeFailedId`; the auto-select effect (`:218-225`) then runs `navigate(last ? \`/chat/${last.sessionId}\` : …, {replace:true})`. So clicking "continued from an earlier session" silently drops the user into the newest unrelated chat with no explanation.
4. **`chat-input.tsx:1013,1019` — `/model` silently discards attached images.** `input` is built from remaining text only, yet `setImages([])` clears them on success. The `@` target chip and `selectedSkills` are likewise ignored.
5. **`session-service.ts:408` — TOCTOU on the 409 idle guard.** `statusOf` is a bare read. `SessionManager.startTask` runs under `withLock(sessionId)` (`session-manager.ts:391,953`), but `forkSession` never takes that lock and `SessionService` has no lock API. A `POST /tasks` landing between the check and `readTraceTolerant` lets the fork snapshot a half-written turn.
6. **`web/test/stream-model.test.ts:282-294` — the origin-routing assertion is vacuous.** The final case builds the child meta via `meta("child")`, and the helper spreads `forked_from` only when defined (`:82`), so the message carries no `forked_from` at all — `expect(m2.forkedFrom).toBe("session-src")` would pass even if the main-session origin guard at `stream-model.ts:405-408` were deleted. Should be `meta("child", "other-src")`. Also untested: a rotated second main meta re-asserting the value, which `agent.ts:481-484` implements.
7. **`session-service.ts:424-432` — every core error becomes 400 `session_fork_failed`,** including I/O failures that should be 5xx, and the raw message is returned to the client. Those messages embed absolute filesystem paths (`agent.ts:566` interpolates the traces dir).
8. **Dropping `token_usage` zeroes the compaction guard for one turn.** `resumed.lastRequestTotal` is 0, so `compactionTrigger()` sees an empty context on the fork's first turn even though a full history was just injected. Self-corrects after one turn, but forking into a much smaller context window has no pre-emptive guard.
9. **`/model` is offered before `models` loads.** Gated only on `onSwitchModel` (`chat-input.tsx:936`), but the panel renders under `{modelSwitchOpen && models && …}` (`:1303`) and `chat-page.tsx:570` passes `models` only once the fetch resolves. In that window `/model` opens an invisible picker *and* suppresses the slash menu (`:963`) — Escape is the only way out.
10. **A legacy session's recorded thinking level silently decays at the first compaction.** Resume honors `meta.thinking_level` (`agent.ts:423`), but the rebuilt meta drops it; compaction's `rotate()` writes that field-less meta into `_002`, so a later resume falls back to the Agent config. Notably affects subagent sessions that inherited a non-default level.
11. **`chat-input.tsx:1009` — Escape re-opens the slash menu.** Closing the picker leaves the `/model` token under the caret, so `slashTok` matches again on the next render. Same on the "picked the current model" no-op path.
12. **`chat-input.tsx:942` — stale captured token range.** `modelSwitchMatchRef` stores `{start,end}` at open time; editing *before* the token while the picker is open makes `removeSlashToken` cut the wrong span and mangle the draft.
13. **`pickSwitchModel` doesn't re-check running/compacting** (`:1006-1009`). The command can't be invoked while running, but an already-open picker survives: open → Enter to send → status flips to running → click a model → 409 → toast.
14. **`mention` is gated on `!slashOpen` but not `modelSwitchOpen`** (`:973`), so typing `@` with the picker open paints the mention menu over it in the same `bottom-full left-0 z-40` slot.
15. **Doc drift (repo rule).** `thinking-level.ts:4-11` still says the picker "writes a picked level straight through to the Agent config" and `:29-30` still refers to hiding "the session read-only tag" — both now false for the session variant. `handoff-banner.tsx:1-6` documents only the handoff banner though the file exports two components.
16. **Legacy traces lose their thinking-level display.** Old traces still carry the field; the viewer no longer has a row for it (`trace-event-row.tsx:166`). Cosmetic, but data the viewer can no longer surface.
17. **Comment/doc overclaims.** (a) `agent.ts:~535` and `sessions-and-traces.{en,zh}.md` say the fork guarantees "the Trace on disk equals the injected history exactly" — resume synthesizes pairing placeholders never written to disk and drops uncommitted turns, so equality is approximate. (b) `chat-page.tsx:~385` says the fork mirrors "the @ handoff's create→postTask flow" — the order is inverted (handoff: `postTask` then `navigate`; fork: `navigate` then `postTask`). (c) `fork.test.ts` "rejects an unknown model pair before touching anything" — the source trace is read before model validation. (d) docs don't name the `session_fork_failed` error code, unlike `nothing_to_compact`.
18. **`fork.ts:35-53` passes through records that are neither a complete model_msg nor an event_msg** (e.g. a stray `partial_*`). `Writer.write`'s `isRecordable` filters them; `resumeTrace` sees the unfiltered array — file ≠ replayed array in that practically unreachable case.
19. **The fork's first task never carries `thinkingLevel`** (`chat-page.tsx:404`), unlike `onSend`. Consistent with the reset-on-session-switch rule, but worth a deliberate note.
20. **Orphan trace on partial failure.** The trace file is written before `sessions.insert(newRow)`, so a failed insert leaves an orphan. Low impact — the trace-discovery walk backfills.

---

### Verified as claimed

**Fork well-formedness after dropping records — no defect.** This was the deepest check, and the design holds. Dropping thinking cannot leave an empty assistant content array: `groupHistoryToUniMessages` (`generative-model.ts:180-197`) groups by *adjacent same-role runs*, so a turn reduced to zero assistant payloads emits no assistant group at all and the neighbouring user messages merge — no empty array, no role-alternation violation. `tool_call`, `tool_call_output` and both request-boundary events are kept, so `resumeTrace`'s eligibility filter and pairing backfill run identically to a normal resume; an aborted trailing turn is dropped wholesale by `dropUncommittedRound`; an unanswered committed `tool_call` gets its in-memory placeholder (tested). The dropped `token_usage`/`subagent` events play no role in turn determination. Origin-carrying child messages are never in a parent trace, so "keep all other model_msgs" is safe.

`fidelity` stripping is safe — it's opaque wire data, not a truncation or summarization marker, and absent fidelity is a supported shape throughout.

Compaction events kept: only the source's highest-index file is forked (one file = one context), so a fork of a compacted session gets the post-compaction context, with the same semantics as resuming one. No file/context contradiction beyond the pre-existing one.

**Thinking-level plumbing is clean.** `thinking_level` removal is correct and the loose legacy read works because `SessionMetaPayload` is only a structural cast over `JSON.parse`, never a runtime validator — removing the field doesn't affect parsing old traces. `asThinkingLevelName` rejects `"default"`, then falls back to the Agent config. `??` vs `||` doesn't matter: no `ThinkingLevelName` is falsy. `"default"` is purely a legacy sentinel and never a valid level. Per-request resolution at `generative-model.ts:909-913,990,1115-1122`; `buildUniConfig` no longer sets it. `RunOptions.thinkingLevel` reaches every request including reconnect retries (`context-engine.ts:277→347→535,566`), with `engine.test.ts` asserting the exact sequence `["high","high",undefined]` — the third being compaction, which deliberately keeps the default. (Consequence worth stating: a user who picks `none` still pays for thinking during compaction.)

I grepped `thinking_level`/`thinkingLevel` across the whole tree at the PR head. The only readers were the web trace row and `stream-model.ts:404-408`, both updated. The CLI never renders `session_meta`. The one miss is blocking finding 3.

Server validation matches `ThinkingLevelName` exactly, and `optionalEnum` returns `undefined` only for a literal missing key, so omitted and explicit are distinguishable and `null` 400s. The web picker is genuinely local per-session state — reset on session switch, `...(turnThinkingLevel ? {thinkingLevel} : {})` means the `""` sentinel can never serialize, and there is no write-through path.

Route authorization matches siblings — the fork route uses the same `resolveSession`, which 404s on both unknown and unauthorized, never leaking existence. `forked_from` round-trips end to end including rotation; forking a fork points at the immediate parent.

**No existing test lost coverage.** All the single-line deletions across `cli/test/render.test.ts:197`, `compaction.test.ts:118`, `replay.test.ts:53`, `session-title.test.ts:58`, `trace.test.ts:28,90` and the server test files are `thinking_level:` fixture-field removals from `sessionMeta({...})` literals — no assertion was removed. The one rewritten test (`resume.test.ts`) **gains** coverage: it now asserts the legacy read, the field's absence in the rebuilt meta, the no-field fallback, and a new case for the legacy literal `"default"`.

New tests are meaningful, not tautological — `fork.test.ts` does a real write→resume round trip off disk, `llm.test.ts` captures the actual per-request `UniConfig` through the `openStream` seam. The exception is non-blocking 6.

i18n parity holds (a full indentation-aware key-path diff of both dictionaries returns empty both directions). Docs en/zh parity is complete across all eight changed pairs.

`/model` handles empty trailing text, doesn't collide with `@` or `/compact`, is correctly suppressed while running and absent in drafts. SSE ordering is fine: the fork navigates then posts, but `use-session-stream.ts:141-151` is connect-first with buffered events plus a post-subscribe replay with dedupe — the shipped `@` handoff is strictly more exposed.

---

### Cross-PR notes

- Conflicts with **#63** on `docs/content/server-api.{en,zh}.md` and `server/src/http/routes/sessions.ts` (both insert a route before the approvals route — keep both); with **#62** on `cli/test/render.test.ts` and `core/test/trace.test.ts` (test churn).
- **This PR and #63 both change the persisted trace format in one release** — #63 changes marker literals inside message text, this one drops `thinking_level` from `session_meta`. Each tested its own migration in isolation; worth reading one genuinely old trace through both before shipping.
- Blocking finding 1 composes with **#60**, which adds two `supportsVision: false` models — both plausible `/model` fork targets.

packages/docs/content/models.zh.md
```diff
@@ -79,11 +79,11 @@ api_key = "sk-..."
 
 ## 思考等级
 
-思考等级共五档：`none | low | medium | high | xhigh`，按 Agent 在 `system_config.yaml` 的 `model.thinking_level` 配置，默认 medium。Web 拾取器只提供 `low` 及以上档位（多数模型不支持关闭思考；`none` 仍是合法的已存值，能正常显示）。对话草稿页在模型选择器旁提供快捷拾取器：选定档位立即写回所选 Agent 的该项配置（切换后的档位即成为该 Agent 的新默认，自下一个 Session 生效；进行中的 Session 沿用创建时的档位）。见 [配置参考](/configuration)。
+思考等级共五档：`none | low | medium | high | xhigh`，按 Agent 在 `system_config.yaml` 的 `model.thinking_level` 配置，默认 medium。Web 拾取器只提供 `low` 及以上档位（多数模型不支持关闭思考；`none` 仍是合法的已存值，能正常显示）。对话草稿页在模型选择器旁提供快捷拾取器：选定档位立即写回所选 Agent 的该项配置（切换后的档位即成为该 Agent 的新默认，自下一个 Session 生效）。进行中的会话里，思考等级是**逐轮参数**：输入区的拾取器默认“跟随 Agent 配置”，显式选定的档位随每次发送下发（仅作用于该会话的后续 Task，不写回 Agent 配置）。见 [配置参考](/configuration)。
```
hiyouga: 不要有“跟随 Agent 配置”的选项，用户设置好以后自动跟随下发

hiyouga:
已按意见去掉「跟随 Agent 配置」选项（e9bbd8b）：会话内的思考等级选择器现在只列出各档位，初始即显示 Agent 配置的档位并自动跟随——用户未手动选择时发送不携带档位（服务端回退到 Agent 配置，配置修改持续生效）；用户选定后即固定为该会话档位，之后每次发送自动携带。草稿页的写回行为不变；相关文档（models / web-app zh+en）与 e2e 断言已同步。

PR #67, opened 2026-07-24
fix(tooling): separate the dev data root from the installed one

hiyouga:
## Problem

Development entry points that run the server or CLI from source shared the same data root as the production install: core's `resolveRoot()` returns `PENGUIN_HOME ?? ~/.penguin/data`, and the server derives its SQLite `web.db` from that root. So `pnpm dev` / `pnpm dev:server` / `pnpm penguin` operated on the very same agents, sessions, and `web.db` as the user's real installed penguin, causing state conflicts while hacking on the repo.

## Fix (scripts-only)

Prefix the data-touching dev scripts with `PENGUIN_HOME="${PENGUIN_HOME:-$HOME/.penguin/dev-data}"` (POSIX `${VAR:-default}` substitution; pnpm runs scripts via `sh`, and the repo is POSIX-only):

- `packages/server/package.json` `dev` — covers `pnpm dev` and `pnpm dev:server` (both delegate here)
- root `package.json` `penguin` — the CLI-from-source runner
- `packages/cli/package.json` `penguin` — the same runner at package level

Semantics: dev entry points default to a separate root `~/.penguin/dev-data`, and an explicitly exported `PENGUIN_HOME` still wins. `dev:web` / `dev:docs` / `dev:landing` are Vite-only and untouched.

Docs: CONTRIBUTING.md's development-commands section explains the dev data root, and the installation page (en/zh) notes it in the from-source section.

## What stays unchanged

- Product code defaults: `resolveRoot()` (`PENGUIN_HOME ?? ~/.penguin/data`), server config, CLI `--root` semantics
- The globally linked / npm-installed `penguin` binary and installed server keep `~/.penguin/data`
- The "single data directory" architecture wording (it describes the product default, still true)

## Verification

- `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck`, `pnpm test` (all suites pass), `pnpm format` + `pnpm format:check` — all green (Node 24)
- Substitution smoke test: `sh -c 'PENGUIN_HOME="${PENGUIN_HOME:-$HOME/.penguin/dev-data}" node -e "console.log(process.env.PENGUIN_HOME)"'` prints `$HOME/.penguin/dev-data` with the variable unset and `/tmp/x` with `PENGUIN_HOME=/tmp/x` exported
- `pnpm penguin config model list --root /tmp/does-not-matter-$$` still starts the CLI from source and lists models
- Checked for tests asserting the exact script strings (none) and README dev sections (no data-root claims; they defer to CONTRIBUTING.md)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,6 +31,11 @@ core) with back-to-back builds deduped: starting `dev:server` and `dev:web` at t
 time (or just `pnpm dev`) installs and builds exactly once. `dev:docs` / `dev:landing`
 run the install check only (`--install-only`).
 
+Dev entry points that touch data (`pnpm dev`, `pnpm dev:server`, `pnpm penguin`) default
+to a separate data root, `~/.penguin/dev-data`, kept apart from the installed CLI/server's
+`~/.penguin/data` — hacking on the repo never mixes state with your real agents. Export
+`PENGUIN_HOME` to point them anywhere else; an explicit value always wins.
+
 Copy `.env.example` to `.env` for model credentials in development.
 
 ## Repo layout
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "test:e2e": "pnpm --filter @prismshadow/penguin-core test:e2e",
     "build": "pnpm -r build && pnpm link:cli",
     "link:cli": "pnpm --dir packages/cli link --global || echo '[link:cli] pnpm global directory not configured; skipping the global penguin link (run pnpm setup once first)'",
-    "penguin": "tsx packages/cli/src/index.ts",
+    "penguin": "PENGUIN_HOME=\"${PENGUIN_HOME:-$HOME/.penguin/dev-data}\" tsx packages/cli/src/index.ts",
     "dev": "concurrently -n server,web -c cyan,magenta \"pnpm dev:server\" \"pnpm dev:web\"",
     "dev:server": "pnpm --filter @prismshadow/penguin-server dev",
     "dev:web": "node scripts/dev-prebuild.mjs && pnpm --filter @prismshadow/penguin-web dev",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -19,7 +19,7 @@
     "typecheck": "tsc --noEmit -p tsconfig.json",
     "test": "vitest run --passWithNoTests",
     "build": "tsup",
-    "penguin": "tsx src/index.ts"
+    "penguin": "PENGUIN_HOME=\"${PENGUIN_HOME:-$HOME/.penguin/dev-data}\" tsx src/index.ts"
   },
   "dependencies": {
     "@prismshadow/agenthub": "^0.4.1",
--- a/packages/docs/content/installation.en.md
+++ b/packages/docs/content/installation.en.md
@@ -60,7 +60,7 @@ cd penguin-harness
 pnpm install && pnpm build
 ```
 
-After the build, run `pnpm penguin <args>` inside the repo as the dev runner, or use the globally linked `penguin` command.
+After the build, run `pnpm penguin <args>` inside the repo as the dev runner, or use the globally linked `penguin` command. Dev entry points (`pnpm penguin`, `pnpm dev`) default to a separate data root `~/.penguin/dev-data`, while the linked/installed `penguin` keeps `~/.penguin/data`; export `PENGUIN_HOME` to override.
 
 ## Published npm packages
 
--- a/packages/docs/content/installation.zh.md
+++ b/packages/docs/content/installation.zh.md
@@ -60,7 +60,7 @@ cd penguin-harness
 pnpm install && pnpm build
 ```
 
-构建完成后，在仓库内用 `pnpm penguin <args>` 作为开发入口运行，或使用全局链接的 `penguin` 命令。
+构建完成后，在仓库内用 `pnpm penguin <args>` 作为开发入口运行，或使用全局链接的 `penguin` 命令。开发入口（`pnpm penguin`、`pnpm dev`）默认使用独立数据根目录 `~/.penguin/dev-data`，全局链接或安装的 `penguin` 仍使用 `~/.penguin/data`；可通过环境变量 `PENGUIN_HOME` 覆盖。
 
 ## 已发布的 npm 包
 
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -25,7 +25,7 @@
     "node": ">=24"
   },
   "scripts": {
-    "dev": "node ../../scripts/dev-prebuild.mjs && tsx watch src/index.ts",
+    "dev": "node ../../scripts/dev-prebuild.mjs && PENGUIN_HOME=\"${PENGUIN_HOME:-$HOME/.penguin/dev-data}\" tsx watch src/index.ts",
     "start": "node --disable-warning=ExperimentalWarning dist/index.js",
     "typecheck": "tsc --noEmit -p tsconfig.json",
     "test": "vitest run --passWithNoTests",


hiyouga:
## Review: APPROVE WITH NITS

Reviewed `git diff origin/main...` (base `1e23452`) — 6 files, +10/-5.

The mechanism is correct and I verified it empirically rather than by inspection: ran the exact prefix string from the diff through `pnpm run` in a scratch package on this machine. Default → `~/.penguin/dev-data`; `PENGUIN_HOME=/custom/x` → `/custom/x`; `PENGUIN_HOME=` (empty) → falls back to the default, which is actually *better* than `resolveRoot()`'s `??`, since that would accept `""` as a root. The assignment prefix also works on the right side of `&&`, which is the shape `packages/server`'s `dev` uses.

Nothing blocking. The nits below are doc-accuracy and two undocumented first-run surprises.

---

### Non-blocking

**1. `CONTRIBUTING.md:36` — "hacking on the repo never mixes state with your real agents" is too strong.** Three repo-sanctioned paths still write to `~/.penguin/data` after this PR:

- **`pnpm build` → `link:cli`** (root `package.json:16-17`) globally links `packages/cli`, whose bin is `"penguin": "./dist/index.js"`. After a build, the global `penguin` *is* the repo's own build, running against the real root. `installation.en.md:63` is factually right about this ("the linked/installed `penguin` keeps `~/.penguin/data`"); CONTRIBUTING's absolute claim is not.
- **`examples/*/package.json` `start`** — `"start": "tsx build-agent.ts"` and `"start": "tsx self-evolve-recursive.ts"`, both unprefixed, calling `createAgent({ agentId })` with no root → `resolveRoot()` → `~/.penguin/data`. This looks deliberate: `examples/build-agent-with-agent/README.md` says outright "This writes the model into `~/.penguin/data/default_project/.project_config.toml`". But it contradicts the new sentence. (`examples/*` aren't workspace members — `pnpm-workspace.yaml` lists only `packages/*`.)
- **`packages/server/package.json:29` `start`** is unprefixed. Defensible as the production entry, but `pnpm --filter @prismshadow/penguin-server start` from a checkout runs repo code against the real root.

Suggest softening to "the dev commands above never mix state…" and naming the exception.

**2. Cross-doc inconsistency the split introduces.** `installation.en.md:63` calls `pnpm penguin` "the dev runner", while `examples/build-agent-with-agent/README.md` step 2 says to configure a model with `penguin config model add …` and step 3 runs the example from the repo root. A developer who follows installation.md and uses `pnpm penguin config model add …` now writes into `~/.penguin/dev-data`, and the example — still resolving `~/.penguin/data` — won't see the model, failing with a no-model error. Nothing in the PR reconciles the two documents.

**3. Two first-run surprises, undocumented.** The split does **not** error — I traced the bootstrap and it's complete: `db/database.ts:21` `mkdirSync(dirname(dbPath))` creates the dir, `server/src/index.ts:25` `seedAdmin()` provisions admin + `default_project` + built-in agents, and `provisionInitialProject` → `ensurePresetModels` seeds the preset catalog. But:
- The dev `web.db` is new, so **the admin password resets to the initial `penguin-2026`** and the change-password nag returns. A developer who changed their admin password will simply fail to log into `pnpm dev`.
- **No API keys carry over.** Preset model *entries* are seeded but no credentials, so the dev app opens with nothing selectable until a key is re-added.

Both deserve one sentence in the new CONTRIBUTING paragraph — that's exactly where the developer hits them. Note `.env` is not the answer: `loadDotenv()` resolves against `process.cwd()`, which under `pnpm --filter @prismshadow/penguin-server dev` is `packages/server`, not the repo root where `.env.example` lives. (Pre-existing, not caused by this PR.)

**4. Missing changelog entry.** No entry under `changelog/unreleased/`, no index line. Precedent for this exact class exists — `changelog/0.1.0/2026-07-20-tooling.md` has a full section "Serialize the dev prebuild to fix concurrent dev:server / dev:web clobbering", a dev-scripts-only change, and `changelog/0.1.1/2026-07-22-tooling.md` exists too. **Honest counterweight:** recent practice has drifted — of the last 12 commits on main only 3 touched `changelog/`; #64, #61, #60, #59, #56, #53, #52, #50 all merged without one. So this violates the written rule but matches what's actually been merging.

**5. Trivial list mismatch.** CONTRIBUTING lists three entry points (`pnpm dev`, `pnpm dev:server`, `pnpm penguin`); `installation.{en,zh}.md:63` list two. Neither mentions `packages/cli`'s own `pnpm penguin`.

---

### Verified as claimed

**"pnpm runs scripts via `sh`, and the repo is POSIX-only"** — supported. `packageManager: pnpm@10.26.2`, and there is **no `.npmrc`** anywhere in the tree, so neither `shell-emulator` nor `script-shell` is overridden → `/bin/sh -c`. POSIX-only holds today: `README.md:152` lists "[ ] Windows support" as an *unchecked roadmap item*, `installation.en.md` Requirements says Linux/macOS, `cli.en.md:169` says neither upgrade path works on Windows, and CI is `ubuntu-latest` with **no OS matrix**. Forward-looking note only: Windows is on the roadmap and `VAR=… cmd` prefixing breaks under `cmd.exe`; the escape hatch would be `shell-emulator=true`. Not a finding against this PR.

**Coverage of data-touching scripts** — I enumerated all 10 `package.json` files. The three prefixed are the right three (root `dev`/`dev:server` inherit by delegation, so no double-prefix needed). Confirmed rather than assumed for the ones that correctly need no prefix:
- `packages/web` `test:e2e` → `e2e/run.sh` sets `PENGUIN_HOME="$DATA"` from `mktemp -d`, plus its own `PENGUIN_WEB_DB`.
- `packages/landing` `shots` → `capture-shots.mjs:521` sets `PENGUIN_HOME` under `mkdtempSync`; `blog-shots` → `capture-blog-shots.mjs:104`, same pattern.
- `packages/core` `test:e2e` only constructs `GenerativeModel`, never touches the state layer.
- **No test anywhere calls `resolveRoot()` or `createAgent()`** — server tests use `makeTempRoot()` with `dbPath: ":memory:"`, core/cli state tests set `PENGUIN_HOME` to a temp dir in `beforeEach`. No unit test can write to the real root.
- `scripts/dev-prebuild.mjs` only uses `os.tmpdir()`; `dev:web`/`dev:docs`/`dev:landing` are Vite-only. There is no `benchmark` script.

**"Deliberately does not change the single-data-directory architecture wording" — correct call.** `design/specs/01-PRINCIPLES.md:53` (PRN-005) states the shared root defaults to `~/.penguin/data` *and names `PENGUIN_HOME` as the documented override*. This PR uses exactly that override, for dev scripts only, so no spec text goes stale. `05-ARCHITECTURE.md:67,264,1026,1041` all describe installed/product behavior and stay accurate. One soft precedent: `design/specs/08-CHANGELOG.md` is a decision log that has recorded comparable dev-script decisions (line 49 "根级 dev 命令补全", line 50 "dev 启动构建互斥"), so a one-liner there would match convention — but `design/` is a separate repo and can't ride in this PR, and that log hasn't been updated since 2026-07-22 either. **Net: no design change required; the PR's judgment is right.** Worth stating explicitly since every other PR in this batch got flagged for design drift — this one legitimately doesn't need it.

**en/zh parity** — `installation.zh.md:63` is a faithful rendering of the en line (same three facts), both one-line edits at the same line in the same section. `configuration.{en,zh}.md:14` and `installation.{en,zh}.md:41` remain correct — they describe the shipped product. README needs no update; its Development section defers to CONTRIBUTING.

---

### Cross-PR note

Merges cleanly onto `1e23452`. No overlap with #66 or #68. For context, #62 (2 files) and #63 (7 files) no longer merge cleanly onto current main after #59/#60/#61/#64 landed, and #68 conflicts on four SKILL.md files — this PR is unaffected by all of that.

PR #68, opened 2026-07-24
feat: add isolated agent tuning pipeline example

zzzbitz:
## Summary

Users can now start a complete Agent creation, Benchmark construction, and score-driven optimization experiment from the Web App's draft screen. Each phase runs in an independent CLI Session, while private Rubrics and Gold answers remain confined to leaf Evaluators instead of leaking into the Test Agent or Optimizer.

The built-in tuning Skills now define clearer phase ownership, complete Evaluator request contracts, bounded parallel Case execution, Trace provenance checks, and versioned accept-or-rollback optimization. The example derives its CLI environment from the active Project and leaves orchestration Skills on `default_agent` rather than installing them on the Test Agent.

## Validation

- `pnpm --filter @prismshadow/penguin-skills test` - 20 tests passed
- `pnpm --filter @prismshadow/penguin-skills build` - passed
- `pnpm --filter @prismshadow/penguin-web test` - 379 tests passed
- `pnpm --filter @prismshadow/penguin-web typecheck` - passed
- `pnpm format:check` - passed
- Manual Chrome check: the fourth example card is visible, submits the full prompt, creates a Session without implicit Skill injection, preserves the shell command, and produces no console warnings or errors

The full Playwright suite was not run because the repository-specific Chromium binary was unavailable locally.


hiyouga:
## Review: REQUEST CHANGES

Reviewed `git diff origin/main...` (merge-base `3f868ac`) — 13 files, +733/-221 — and separately `git diff origin/main pr/68` to assess the conflict.

Thanks for this — the phase-ownership tightening, the `benchmark_id` request contract, and the Trace provenance binding are real improvements over what's on main. The blockers below are mostly a consequence of the branch being based on `3f868ac`, which is 4 merged PRs behind; a rebase fixes several of them mechanically, but not all.

**Please read the merge note at the end before rebasing** — three of these will *not* surface as conflict markers and will land silently broken if the rebase is done mechanically.

---

### Blocking

**1. All four `version` values are lower than `origin/main`, so this rewrite can never reach an existing user.**

| skill | `3f868ac` (this PR's base) | `origin/main` | `pr/68` |
|---|---|---|---|
| agent-creation | 2 | **5** | 3 |
| agent-evaluation | 1 | **4** | 2 |
| agent-optimization | 1 | **4** | 2 |
| benchmark-design | 1 | **4** | 2 |

`updated` regresses too (agent-creation `2026-07-24T17:34:13Z` vs main's `2026-07-25T00:00:00Z`).

`packages/web/src/features/skills/skills-page.tsx:76-86` (`outdatedAgentIds`) filters on `v !== undefined && v < libraryVersion`, and its own doc comment is explicit: *"a locally newer copy … doesn't [count] — reminders fire only on a strictly lower version."*

Failure scenario: a user whose `default_agent` already has `agent-optimization` v4 installed gets library v2. `4 < 2` is false, so the agent never appears as outdated, the "Update" affordance never renders, and the installed copy stays on the pre-#68 text permanently. The library file and every installed copy diverge with no UI path to reconcile. Versions must be bumped from **main's** values — 6 / 5 / 5 / 5 — not from the stale base.

**2. The rewrite reverts merged #61 across all four skills.**

`816a823` ("fix(core): present the project dir to the model as the App Data Dir…") renamed the Environment line. On `origin/main`, `default-config.ts:150` emits:

```
- App Data Dir: {{PROJECT_DIR}}
- Agent ID: {{AGENT_ID}}
```

There is **no `Project Dir` line in the Environment section on main.** Branched pre-#61, this PR reintroduces the old wording:

- `agent-creation/SKILL.md:20,23` — `PROJECT_DIR="<project_dir>"  # the Project Dir value from your Environment section`, and "All agents of this project live side by side in the project directory" (main: "under `agents/` in the App Data Dir")
- `agent-evaluation/SKILL.md:40` — "Environment Project Dir"; `:78` drops #61's clarifying comment
- `benchmark-design/SKILL.md:32` and `agent-optimization/SKILL.md:32` — "Use the Environment's Project Dir"

Failure scenario: an agent following `benchmark-design` looks for "Project Dir" in its Environment, finds only "App Data Dir", and either guesses or asks — the exact ambiguity #61 was filed to fix. A mechanical "take theirs" resolution reintroduces this across four files.

**3. `strings-en.ts:584` / `strings.ts` — the same stale field name is baked into the example prompt, and this file does *not* conflict.**

```
PROJECT_DIR="<Project Dir from the current Session Environment>"
PROJECT_ID="$(basename "$PROJECT_DIR")"
export PENGUIN_HOME="$(dirname "$PROJECT_DIR")"
```

The orchestrating agent must resolve `PROJECT_DIR` to derive `PROJECT_ID` and `PENGUIN_HOME` for all three `penguin run` launches, and the named field is absent from its prompt on current main. Unlike finding 2, **nothing flags this during merge** — it lands broken silently.

**4. `agent-optimization/SKILL.md:84` reverses a snapshot-ownership rule that `design/specs/` states as ground truth.**

`design/specs/05-ARCHITECTURE.md:302`: agent-optimization 要求当前版本的系统快照已由现有 Web 导出能力准备好，但**不自行创建或导入快照包**.

Both `3f868ac` and `origin/main` match the spec verbatim: *"The system owns Agent State snapshot archives… **Do not create, import, extract, or replace snapshot archives yourself.** Require `<target>/snapshots/v<version>.tar.gz` to exist; if it is missing, stop and ask the user to export…"*

This PR replaces it with the opposite: *"When the current-version snapshot is absent, **create it** from `agent_state/` without `.vault.toml`."* That is this PR's own change, present at the base and deliberately removed — not a merge artifact. Per CLAUDE.md, a design change must be confirmed and land in `design/specs/` **first**. Failure scenario: the model hand-rolls a `tar.gz` the Web import path never validated; if its layout differs from the exporter's, the user's documented recovery route fails exactly when rollback is needed.

**5. One-shot feedback mode is deleted, contradicting the spec and leaving docs stale.**

`design/specs/05-ARCHITECTURE.md:522` still specifies: 如承担优化角色的 Agent…未指明目标时默认优化调用方. This PR removes the entire "Pick the target Agent" / "One-shot feedback mode" path including the `<handoff_from>` @-mention route — which the web app still generates (`agent-mentions.ts:105`) and core still recognizes (`session-title.ts:36`).

`self-improvement.{en,zh}.md` was updated but **not** `skills.{en,zh}.md`, whose tables mirror the SKILL.md `description` verbatim. After this PR, `skills.en.md:75` still reads "Improve an Agent State from **direct feedback or** versioned multi-Case Benchmark scores…" while the shipped description says "from versioned Benchmark scores". Confirmed via `git diff --name-only origin/main...pr/68 -- packages/docs/` — only the two `self-improvement` files changed.

**6. The example prompt instructs the Optimizer to do what `agent-optimization` forbids, in the same changeset.**

The Benchmark is defined so the answer is unobtainable from public information — *"The private mapping is a fixed permutation from the three markers to the three actions"*, *"Public Statements must not provide enough information to derive the private mapping"*. Phase 3 then authorizes: *"This experiment permits writing a context-to-action mapping **learned from black-box scores** into the Agent State when it is reusable across similar instances."*

But `agent-optimization/SKILL.md:66-69` — in this same PR — says: *"Do not encode Case ids, expected answers, **Benchmark-specific constants**, private scoring conditions… Prefer improving general analysis, validation, and execution methods over memorizing Benchmark answers."* The mapping is a permutation the Builder invented for this Benchmark and that exists nowhere else; "reusable across similar instances" means reusable within this Benchmark only.

So the score curve the example asks the user to chart is a record of how many rounds it took to brute-force a hidden permutation from a 3-cell score oracle — presented as agent self-improvement, on the first screen a new user sees.

Compounding it: **"Baseline hard gate: below 70"** plus *"It may choose mappings that discriminate against the Test Agent's public default policy"* tells `benchmark-design` to keep restructuring until the Test Agent scores low, and `benchmark-design/SKILL.md:180-184` will honor a hard gate by continuing "credible structural adjustments" until met. Phase 1 pins the default policy being discriminated against. The improvement is manufactured at both ends.

*(The framing judgment is mine; the flat contradiction between the two texts is mechanical.)*

**7. `Runs: 1` makes accept/reject single-sample, defeating "accept-or-rollback".**

`benchmark-design/SKILL.md:79` sets `runs = 3` by default precisely to average noise; the example overrides to 1. `agent-optimization/SKILL.md:175` accepts a Candidate on "strictly higher" score and `:180` promotes it to the next Reference. A 3-case × 1-run matrix against a non-deterministic LLM accepts a no-op edit on sampling noise and permanently promotes it. The docs change asserts "the candidate version is kept only if the total score strictly improves" — true mechanically, meaningless at n=1.

---

### Non-blocking

1. **The isolation claim is weaker than stated — and the spec agrees.** `design/specs/05-ARCHITECTURE.md:303`: 约束靠 Prompt 与 Trace 审计，**不引入文件系统沙箱**. Verified there is none: `packages/cli/src/approval.ts:30` defaults to `allow-all`, and grepping `sandbox|jail|confine|restrictPath` across `packages/core/src` + `packages/cli/src` returns only approval-mode strings. The Test Agent's CWD is `<app_data_dir>/agents/<test_agent_id>/workspaces/tmp-<8hex>`; gold answers sit at `<app_data_dir>/agents/<test_agent_id>/benchmarks/<id>/CASE-nnn/rubric/README.md` — `../../benchmarks/` from CWD, inside the Test Agent's *own* tree, while the system prompt hands it the App Data Dir and says "Reach these paths directly." **To be clear: I found no step that puts gold answers into the Test Agent's prompt or workspace** — `agent-evaluation:67` genuinely copies only `statement/`, so the accidental-exposure boundary is real. But the PR description and `changelog/…:11` describe it as structural when it is prompt-level plus a post-hoc self-audit by the same model. Please reword to match the spec's own honest framing.
2. **`design/specs/05-ARCHITECTURE.md:329`** says the full matrix dispatches in one parallel batch (同批并行). "Use bounded batches that fit the available subagent capacity" (`benchmark-design:146`, `agent-optimization:122`) is a genuine improvement, but it is a spec divergence not landed in `design/specs/`.
3. **`agent-creation/SKILL.md`** drops the `## Set name and description` section and the "Prefer configuring an agent the user already created" preference. `name`/`description` guidance now lives only under "Creating a brand-new agent", so the configure-existing path no longer sets them, and the skill nudges toward creating new agents. Likely unintended.
4. **`example-tasks.test.ts` is thin** — `expect(task?.skills).toEqual([])` restates the constant, and the `buildSkillsMessage` assertion re-derives the empty-list branch. It also touches `zh` only; the higher-value assertion (en/zh key parity for `exampleTasks`, which this PR roughly doubles) is untested.
5. **Hardcoded `provider: deepseek` / `deepseek-v4-flash`** in both catalogs. The id is real (`model-catalog.ts:207`) and matches the default project config, but a user on OpenRouter (added in #60) gets a one-click card that fails at the first Evaluator launch.
6. **Changelog surface grouping.** CLAUDE.md says "default to extending the existing file for that surface"; `changelog/unreleased/2026-07-22-web-app.md` exists on main and the new entry's `## Web App` section duplicates that surface. Defensible as one cross-surface feature — maintainer call. The entry also omits the `self-improvement.{en,zh}.md` docs change. Format and index line are otherwise correct.
7. **`draft-view.tsx:555`** — comment still says "24×24 line icons (gamepad / music note / sparkle)" with four cards now rendering. Layout itself is fine (`flex flex-col`, no 3-item assumption).

**i18n is clean.** No Chinese in the `strings-en.ts` addition, no untranslated English in `strings.ts`; both add exactly `selfImprovement: { label, desc, prompt }` and the prompts are semantically parallel. `self-improvement.en.md` / `.zh.md` changed identically.

---

### Verified as claimed

The fourth card renders and submits the **full** prompt (`skills: []` → `buildSkillsMessage` returns the body unchanged), matching the repo convention that example cards submit full prompts and only marketing copy is condensed. Each phase really is an independent `penguin run` via `exec_command`. All referenced tools and flags exist. `benchmark_id` replaced `benchmark_dir` consistently across all three skills, with identity fields matching in `agent-evaluation:27-36`, `benchmark-design:152-162`, `agent-optimization:128-138`. Trace provenance (`agent-evaluation:96-106`) binds one unique root from `session_meta` on four fields. Phase ownership is genuinely clearer, with no contradiction between the four skills on ownership. Orchestration skills do stay on `default_agent` ("Install no Skills"; all three phases `--agent-id default_agent`).

---

### Merge note — what a correct rebase must produce

Three of these produce **no conflict marker**, so please do them by hand:

- **All four skills** — take #61's App Data Dir wording from `origin/main`, not this branch's `Project Dir`. Specifically: `agent-creation` "under `agents/` in the App Data Dir" + `APP_DATA_DIR=` in both bash blocks (including the `cp …/default_agent/…` line); `agent-evaluation` "Environment App Data Dir" + the `PROJECT_DIR="<app_data_dir>"` comment; `benchmark-design` / `agent-optimization` "Environment's App Data Dir".
- **All four skills** — `version` must exceed main's: **6 / 5 / 5 / 5**, `updated` after `2026-07-25T00:00:00Z`.
- **`agent-optimization`** — keep main's snapshot paragraph unless `design/specs/05-ARCHITECTURE.md:302` changes first.
- **`agent-optimization`** — decide deliberately on one-shot feedback mode. If it goes, `design/specs/05-ARCHITECTURE.md:522` and `packages/docs/content/skills.{en,zh}.md:75` must change in the same PR.
- **⚠ No conflict marker:** `benchmark_dir` → `benchmark_id` must be taken **all-or-nothing** across `agent-evaluation`, `benchmark-design`, `agent-optimization`. Main still has `benchmark_dir` in `agent-evaluation`; resolving one file toward main and another toward this branch yields Evaluators that reject every request as `invalid_request`.
- **⚠ No conflict marker:** `strings-en.ts` / `strings.ts` merge cleanly and still need the `Project Dir` → `App Data Dir` fix by hand (blocking 3).

For context on sequencing: #65, #66 and this PR all add lines to `changelog/unreleased/README.md`, and #66 also lands a `2026-07-24`-dated entry, so expect a trivial index conflict whichever order they merge in.

---

*Findings 6 and 7 are judgments about model behavior under these prompts rather than mechanical defects — flagged as such. Everything else was verified by reading the files at `origin/main` (`1e23452`), `pr/68` (`9c0f151`), and the merge base `3f868ac`. No builds or tests were run.*

PR #71, opened 2026-07-26
feat(core,server,web): images reach every input — steering, goal objectives, one fold

### LINKED ISSUE #108: [Bug] 运行中插话只发文字不发图片，残留的图片还会让「停止」按钮消失
## 问题描述

运行期间的「插话」只能带文本。输入框里附了图片时，点「发送给运行中的 Agent」只会把文字发出去，图片被留在输入框里，界面上没有任何提示——用户以为图片一起发过去了，实际没有。这跟模型支不支持视觉无关，任何模型都一样丢。

图片被留下之后还会连带把按钮打死：运行期间输入框右下角是一个单按钮，草稿为空时是「停止」，草稿有内容时是发送。而只剩一张图片的草稿既发不出去（插话不带图片，按钮是灰的），又不再是「停止」——这次运行没法从输入框停止，只能先把图片删掉。

## 复现步骤

1. `pnpm dev` 起 Server 和 Web，打开一个会话。
2. 发一条消息让 Agent 跑起来，保持 RUNNING（用 always-ask 审批模式停在工具确认上最容易复现）。
3. 保持默认的「插话」发送模式（不切到「跟进」）。
4. 运行期间输入一段文字，并附一张图片，点「发送给运行中的 Agent」。

## 实际结果

- 只有文字被发给了运行中的 Agent，图片仍留在输入框里，没有任何提示。
- 图片留下之后，按钮变成灰的「发送给运行中的 Agent」，「停止」消失，这次运行无法再从输入框停止，必须先手动删掉图片。
- 只附一张图片、不写文字：按钮从一开始就是灰的，同样顶掉了「停止」。

## 预期结果

- 插话时附着的图片，要么跟文字一起发给运行中的 Agent，要么明确告诉用户没发。
- 运行期间按钮任何时候都应该有一个可用的动作：要么能把草稿发出去，要么是「停止」，不应该出现「按不动、同时也停不了」。

rank-Yu:
Closes #108.

## Summary

Images could only enter a session through one door: a Prompt that starts a Task. Every other input either dropped them or refused them outright — steering silently left them sitting in the composer, a goal objective rejected them with a 400. This makes all three inputs carry images, through one fold with three call sites.

**Mid-run steering** — the moment an image is most useful is mid-run, when the agent has just produced the wrong thing and the fastest correction is a screenshot. `session.steer(text, images?)` queues them with the text; the engine delivers them as ordinary user image messages directly behind the `[user_steering]` block, the same shape a Prompt uses, so the LLM client, Trace and replay need no new case. An image with no caption is a complete steering message on its own.

**Goal objectives** — "make the page match this mockup" is a goal, and a screenshot states it better than a paragraph. Accepted now, with one deliberate difference from every other input: folded to scratchpad path lines **whatever the model's vision**. The objective is re-injected as the text of every round's `[goal]` block, so an image cannot ride along as an image; sending it in round 1 alone would leave every later round pointing at something compaction has since removed — a silent failure, since the objective still *reads* correct. As a path line it survives every round and every compaction, and the model spends tokens on it only when it actually looks. Text stays required: a picture alone states no objective.

**The mid-run action button** — while a task runs, the composer's single button flips from Stop to its send face as soon as the draft has any content. Steer mode couldn't carry every draft, so one it couldn't carry (an image with no text, back when images weren't steerable) produced a **permanently disabled** send button that had also displaced Stop: the message couldn't be sent and the run couldn't be stopped from the composer. The only exit was deleting the attachment. That draft now simply steers; what steering still can't carry — skills selected with no text or image, or an `@` handoff target — falls back to the **follow-up queue** for that one send, so the button stays live. Stop keeps its original rule: it is the button's face exactly when the draft is empty.

## One fold, three call sites

The conversion has always lived in core (`imagesToScratchpadPaths`, since the initial commit). What changed is who calls it and when. `Session` takes `imagesDir` (always) plus an explicit `modelVision` flag in place of the optional `inputImagesDir`, and binds the fold **once** — it is the only layer that knows both the scratchpad and the model's capability. The gate moves to each call site:

| Input | Folds when | Applied at |
| --- | --- | --- |
| Prompt (`run`) | no vision | run entry, before Trace and title material |
| Steering (`steer`) | no vision | delivery, at the turn boundary |
| Goal objective | **always** | before the objective is extracted |

Steering folds at delivery rather than at queue time for two independent reasons: `Session.steer` must stay synchronous (the CLI calls it inside a sync readline handler and prints its acknowledgment from the boolean it returns), and the queue is discarded when the run exits — folding early would write orphan files for a message that never reaches the model. A fold that fails on an unwritable scratchpad drops the images with a note rather than taking the running Task down with it.

The old encoding — "was the directory configured" standing in for "the model lacks vision" — worked only while there was exactly one rule to express. Goal mode is the second.

For steering the fold happens **before** the wrapping, so the lines land inside the block: `parseUserSteeringText` only recognizes a text that is exactly one block, and a line appended after the closing tag would cost the message its steering identity — every render layer would read it as a new Task. For the goal objective it happens before the objective is extracted, so the lines sit at the end of the text where `stripLeadingMarkerBlocks` keeps them through every round.

## Rendering

Steering images land **inside** the steering chip, not in bubbles of their own, and neither the chat stream nor the Trace timeline opens a new Task for them (both apply the grouping rule they already use for a Prompt's images); an images-only Prompt sent later still starts a real turn. Goal rounds show the attachments in full under round 1's bubble and collapse them into a one-line chip from round 2 that expands on click — they are in every round's input, but a twenty-round goal shouldn't repeat the same picture twenty times.

Both shapes render identically whether the images arrived as real image messages (vision model) or as `[attached image: …]` path lines, since `splitImageAttachments` reconstructs a fetchable `/api/sessions/:id/scratchpad/:file` URL from the path.

One thing deliberately **not** unified: the objective the server records for the status card and title material stays image-free (`isPlainText` filters them out before the derivation). Core folds them into its own copy at run time, which is what every round re-injects. Folding in both places would grow a string of absolute scratchpad paths in the UI.

## Verification

- Full-repo typecheck ✓ · unit tests green across all packages (core 577 · server 401 · web 454 · cli 200 · …) ✓ · prettier ✓
- New tests: **core** ×6 (steering image order behind the text on a vision model; the in-block fold without vision; a failing fold degrading instead of killing the Task; goal fold + re-injection across rounds on a vision model; goal fold without vision; a text-less objective rejected), **server** ×4 (steer route image routing/validation; Trace-timeline segmentation; goal route accepts text+images and refuses image-only; the recorded objective stays path-free), **web** ×3 (images join the steering chip; an images-only steering message; attachment lines survive the goal → skills → attachments render chain)
- New e2e `steer-midrun.spec.mjs` ×2: an image-only draft mid-run **steers** and comes back as a chip with the image inside it (still one Task), and a skills-only draft falls back to the queue and drains. Reverting the grouping logic was confirmed to make the first one fail, so the assertions aren't vacuous.
- Full Playwright e2e: **24 passed, 4 failed** — `draft-user-scope` / `layout` / `paging` / `skills`, each verified to fail identically on this branch's merge commit baseline (pre-existing, unrelated to this change)
- Docs (`goal-mode`, `agent-loop`, `server-api`, `web-app`, en + zh) and a `changelog/unreleased/` entry updated


hiyouga (changes requested):
图片是可以 steering 的

hiyouga:
@rank-Yu 这个得有issue吧？是个bug

rank-Yu:
> @rank-Yu 这个得有issue吧？是个bug

我提个 issue

packages/core/src/engine/context-engine.ts
```diff
@@ -424,35 +442,77 @@ export class ContextEngine {
...
+   * path lines at delivery (see deliverSteering). Returns false when no Task is running (the
+   * host should then submit the message as a normal task instead).
+   *
+   * A message with neither text nor images queues nothing and still returns true: `false` is
+   * specifically "send this as a normal task", which would be the wrong advice for an empty
+   * one. Every host guards against this already; the check is here so an empty
+   * `[user_steering]` block can't reach the model through a host that forgets.
    */
-  steer(text: string): boolean {
+  steer(text: string, images: string[] = []): boolean {
```
hiyouga: 这个输入应该是 omni message [] 的 array，SDK 的接口应该都用 Omni message

packages/core/src/agent.ts
```diff
@@ -451,15 +448,12 @@ export class Agent {
...
-              sessionId,
-            ),
-          }
-        : {}),
+      // Where an input image lands when it becomes a path line (see SessionConfig.imagesDir).
+      imagesDir: path.join(
+        scratchpadDir(this.state.root, this.state.projectId, this.state.agentId),
+        sessionId,
+      ),
+      modelVision: modelEntry.vision !== false,
```
hiyouga: modelHasVision

hiyouga (changes requested):

hiyouga (approved):

PR #72, opened 2026-07-26
refactor(core): ignore the recorded thinking level when resuming a session

hiyouga:
## What

When resuming a Session from its Trace, a `thinking_level` recorded in the session_meta is now ignored. The resumed Session's default thinking level comes from the Agent's current config (`model.thinking_level`) only.

## Why

Requested by the product owner: resume no longer needs to stay compatible with a thinking level stored in the session meta — if the field appears, it is simply ignored. Current Traces stopped recording the field when the thinking level became a per-turn run parameter (#64); this PR removes the last resume-time compatibility read of the retired field.

- `packages/core/src/agent.ts`: `resumeSession` reads the level from `this.state.systemConfig.model?.thinking_level` only; the `asThinkingLevelName` helper and the `THINKING_LEVEL_NAMES` table (whose sole caller was the resume path) are removed. Per-turn thinking-level overrides (`RunOptions.thinkingLevel`) and spawn-time subagent inheritance are untouched.
- `packages/core/src/omnimessage/types.ts`: the `SessionMetaPayload` doc comment now says a legacy `thinking_level` is ignored on resume.
- `packages/core/test/resume.test.ts`: the resume test now asserts a recorded `"xhigh"` is ignored and the config level (`medium`) wins; the no-field and legacy-literal-`"default"` cases keep reading the config as before.
- `packages/docs/content/omni-message.{en,zh}.md`: the session_meta section no longer claims resume honors the legacy field.

Behavioral consequence: a resumed legacy subagent session no longer keeps the level it inherited at spawn time; it falls back to the resuming Agent's configured level.

## Verification

All run from the worktree root:

- `pnpm build` — pass (the trailing `link:cli` global-link step is skipped in this environment: no pnpm global bin dir; unrelated to this change)
- `pnpm typecheck` — pass (core, server, cli, web)
- `pnpm test` — pass: core 492 passed / 2 skipped (the offline-by-default live LLM e2e), server 340, cli 186, web 383 (includes `thinking-level.test.ts`), docs 11, skills 16, landing 36
- `pnpm format && pnpm format:check` — clean ("All matched files use Prettier code style!")

### DIFF
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -70,26 +70,6 @@ import type { ModelEntry } from "./state/index.js";
  */
 const MAX_SUBAGENT_DEPTH = 1;
 
-/** The five valid thinking level names (legacy session_meta additionally recorded the literal "default" for "no level"). */
-const THINKING_LEVEL_NAMES: readonly ThinkingLevelName[] = [
-  "none",
-  "low",
-  "medium",
-  "high",
-  "xhigh",
-];
-
-/**
- * Narrows a legacy session_meta `thinking_level` back to a ThinkingLevelName; the literal
- * "default" (no level recorded), a missing field (current Traces no longer record one), and
- * anything unknown are not levels and yield undefined.
- */
-function asThinkingLevelName(value: unknown): ThinkingLevelName | undefined {
-  return typeof value === "string" && (THINKING_LEVEL_NAMES as readonly string[]).includes(value)
-    ? (value as ThinkingLevelName)
-    : undefined;
-}
-
 export interface CreateAgentOptions {
   agentId?: string;
   projectId?: string;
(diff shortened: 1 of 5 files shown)


hiyouga:
## Review — LGTM

Focused and complete: the helper and its lookup table go away with their only caller, the doc comment on `SessionMetaPayload`, both docs pages and the test all move in lockstep, and the test still covers all three shapes (legacy `"xhigh"`, no field, legacy literal `"default"`) rather than just the one that changed.

Two notes, neither blocking:

- **`ThinkingLevelName` is still imported and used** (`RunOptions.thinkingLevel`, `buildRuntime`), so removing `asThinkingLevelName` leaves no dangling import — confirmed on the branch.
- **This is a deliberate back-compat removal against on-disk Traces.** The body says the product owner asked for it, so the decision is made; per the repo's changelog rule the batch changelog should carry the one-line consequence (a resumed legacy subagent session no longer keeps its inherited level) so it is discoverable outside this PR.

hiyouga:
Both notes acknowledged. The second is already covered: the batch changelog PR #80's core entry (2026-07-26-core.md) carries the consequence verbatim — "a resumed legacy subagent session no longer keeps the level it inherited at spawn time — it follows the resuming Agent's configuration". Per the repo's new backward-compatibility rule, #80 will additionally gain a dedicated backward-compatibility file for this batch recording the ignore-legacy-field decision (owner-directed), its scope, and that no compat code needs future retirement.

PR #73, opened 2026-07-26
feat(server,web): import and export Trace files in the trace viewer

hiyouga:
## Summary

Trace files can now be moved across deployments straight from the web trace viewer ("Trajectory" page):

- **Export** — a download button next to the selected Trace file serves the raw JSONL verbatim (`GET /api/projects/:p/agents/:a/traces/:sessionId/:index/download` with attachment + nosniff headers). Available to any project member, like the agent snapshot export.
- **Import** — an upload icon-button on each agent row in the left tree (owners only, mirroring the agent snapshot import; 14MB cap). The uploaded file must be valid Trace JSONL whose first record is a `session_meta` with a filename-safe `session_id`; it lands in the date directory of its first record's timestamp at the session's next free index — existing files are never overwritten. On success the tree refreshes and the imported file is selected; failures show inline, with the `invalid_trace` code localized.
- Shared DTOs `TraceImportRequest` / `TraceImportResponse`, zh/en i18n keys, docs updates (`server-api`, `sessions-and-traces`), and server integration tests covering export bytes/headers, 404s, access control, index allocation, and every invalid-content rejection.

## Verification

- `pnpm build` — pass
- `pnpm test` — pass (server 349, core 492, cli 186, web 383, docs 11, skills 16, landing 36 — includes 9 new tests in `packages/server/test/trace-import-export.test.ts`)
- `pnpm format && pnpm format:check` — pass
- Playwright e2e for the traces page does not exist — skipped.


hiyouga:
导入时候应该检查重复，如果重复就提醒无法导入重复的

hiyouga:
## Review

The transport/content validation split is right (route checks base64 + size, service checks JSONL + leading `session_meta` + filename-safe `session_id` next to the path construction), the owner/member split mirrors the snapshot endpoints, and the tests cover the path-traversal case. Three findings, the first is the substantive one.

### 1. An import whose `session_id` already exists under the target Agent merges into that session

`importTraceFile` allocates `max(existing index) + 1` and writes. That never overwrites a byte, but "next free index of an existing session" is not a neutral place to put an unrelated file — index *N* of a session means "context segment *N* of that session", and three existing code paths read it that way:

- **History**: `readMessages` → `readMessagesExpanded` concatenates *every* index file of the session in order, so the imported content is spliced into the existing session's transcript in the Web UI.
- **Resume**: `findLatestTraceFile` (core `trace/resume.ts`) picks the **highest-index** file. After an import, resuming that session replays the *imported* file and takes its `session_meta` as runtime config — `agent.resumeSession` reads `workspace`, `system_prompt` and the model ref straight from it (`packages/core/src/agent.ts`). So the imported file silently becomes the session's resume point.
- **A live session**: `Writer.rotate()` is just `index += 1`, and `write()` uses `appendFile` with no existence check. If the session is running at index 1 when an import lands at index 2, the next compaction appends live content onto the imported file (or, when the import's date dir differs from the writer's, produces two files with the same index that `locateAll` then both returns).

None of this needs a hostile actor — exporting a Trace from a deployment and importing it back into the same Agent to test it is the obvious first thing someone does with this feature.

Since the stated purpose is moving Traces across deployments, the simplest fix is to make the collision explicit rather than absorbing it:

```ts
const existing = await this.locateAll(projectId, agentId, sessionId);
if (existing.length > 0) {
  throw invalid(`This Agent already has a Session with id ${sessionId}.`);
}
const index = 1;
```

If importing additional segments of a session you already have is a real use case, that should be an explicit opt-in flag rather than the default, and the docs sentence ("lands at the session's next free index") should say what that means for resume.

### 2. `fs.writeFile` should be `{ flag: "wx" }`

`locateAll` → compute index → `writeFile` is a TOCTOU window: two concurrent imports of the same `session_id` both compute *N*, and the second overwrites the first. `MessagesLiveTail`-style claims are fine to make in a doc comment, but `TraceImportResponse.index` and both docs pages state "never overwrites an existing file" as a guarantee, so it is worth making the open enforce it — `wx` fails with `EEXIST` instead, which you can surface as a retry or a 409.

### 3. Date directory is UTC here, local everywhere else

`importTraceFile` derives the date dir with `new Date(ts).toISOString().slice(0, 10)`, but `Writer` uses `formatLocalDate` (`internal/dates.ts`, `getFullYear/getMonth/getDate`). On any server not running UTC, exporting a Trace and importing it back files it under a different date than the one it was recorded under — cosmetic (`locateAll` scans all date dirs), but it makes export→import visibly non-round-tripping in the drill-down tree. `formatLocalDate(new Date(ts))` would match the writer.

### Minor

- The web import path reads the file with `readAsDataURL`, which materializes a ~19MB string for a 14MB file and gives no client-side size check — a user picking a 100MB file uploads ~133MB before the server 400s. A `file.size > 14MB` guard before reading would fail fast.

hiyouga:
All findings addressed in fe0c85b:

1. **Duplicate session rejected** — `importTraceFile` now throws `HttpError(409, "trace_session_exists", ...)` when `locateAll` finds any existing file for the session id; the imported file is always index 001 of a new Session. The web surfaces the code through `errors.byCode` (zh + en).
2. **TOCTOU** — the write uses `{ flag: "wx" }`; `EEXIST` maps to the same 409, so two concurrent imports of the same new session id cannot overwrite each other.
3. **Date dir timezone** — the import now uses the server's `formatLocalDate` (`src/internal/dates.ts`, same convention as core's Trace Writer; core does not export its copy) instead of `toISOString().slice(0, 10)`, so export → import round-trips into the same date dir on non-UTC servers.
4. **Client-side size guard** — `onPickFile` rejects picks over `14 * 1024 * 1024` bytes before any read, showing a localized message through the existing inline import-error path.

Docs: the four sentences in `server-api.{en,zh}.md` and `sessions-and-traces.{en,zh}.md` now state the new semantics (duplicate session id → 409; an import is always index 001 of a new Session, local date dir).

Tests: the next-free-index test is now "duplicate session id → 409 trace_session_exists, nothing written" (tree unchanged, existing file byte-identical); the roundtrip test keeps index 1 and absorbed the trailing-newline normalization check; added a concurrent-import test — the loser is rejected by either the pre-check or the `wx` write, so the observable outcome (one 200, one 409, exactly one stored file) is deterministic regardless of interleaving. Trace fixtures moved to 02:00 UTC with the expected date computed via the same local-date formula, so a regression back to UTC date dirs fails on any runner west of UTC.

Verification: `pnpm build`, `pnpm typecheck`, `pnpm test` (server 350/350, incl. 10 trace-import-export tests; all other packages green), `pnpm format && pnpm format:check` all pass.

hiyouga:
错误提示框要用之前统一的弹窗

hiyouga:
Follow-up in 577a900 — import failures now use the app's shared notification surface.

**Adopted mechanism**: the toast established by #54's one-notification-rule — `toastError(...)` from `packages/web/src/components/ui/toast.tsx` (module-level pub/sub, single `<Toaster />` mounted in `app.tsx`), the same call shape the refactored pages use for a failed mutation (`toastError(apiErrorText(e))`, as in models-page / skills-page / agent-settings-page).

**Replaced**: the `importError` state and the inline red `<p>` under the agent row. All three failure paths now toast: the server rejection (`apiErrorText(e)` — localizes `trace_session_exists`, `invalid_trace`, the 400 cap, etc. via `errors.byCode`), the client-side 14MB pre-check (`S.traces.fileTooLarge`), and the FileReader failure (`S.common.unknownError`).

**Intentionally left**: the tree's `getAgentTraces` load error stays inline — it predates this PR and is a page-load-failure state, which the #54 rule explicitly keeps with its content rather than in a disappearing toast (same as the trace-file-view load error). The export path is a plain `<a download>` anchor with no script-side error handling, so there was nothing to migrate there.

**Tests**: none added — the web suite is node-env pure-logic by design ("no DOM tests" per `vitest.config.ts`), and these calls live in component event handlers; correctness is covered by typecheck (the toast import and i18n keys are type-checked) and the paths are one-line substitutions of the previously verified handlers.

Verification: `pnpm build`, `pnpm typecheck`, `pnpm test` (all packages green), `pnpm format && pnpm format:check` all pass. Branch not rebased, changelog untouched, per instruction.

PR #74, opened 2026-07-26
feat(core,server,web,cli): show version in the web UI with update check and admin self-update

hiyouga:
The web app now tells you what version you are running and when a newer release exists, and lets an admin install it without leaving the browser.

## What's new for users

- **Version footer** — the sidebar user menu shows `PenguinHarness vX.Y.Z`, plus the release's build date (stamped by release CI; dev builds show no date). Fetched lazily on first menu open, never on app load.
- **Update reminder** — opening the menu also checks GitHub for a newer release (cached; the server's only outbound call, fail-soft in every case). When one exists: a small accent-colored dot on the user button, and a highlighted "New version vX.Y.Z" row linking to the release notes.
- **Admin self-update** — admins additionally get an "Update now" action: a dialog explains that the new release is installed into the install directory (data untouched) and runs the CLI's existing `penguin update` machinery on the server host, showing the command's output tail. **The service must be restarted afterwards** (`penguin web` / `penguin server`) for the update to take effect. Source checkouts, unrecognized layouts, and Windows are reported as unsupported with the CLI's own explanation.
- **Opt-out** — `PENGUIN_UPDATE_CHECK=off` disables the release lookup entirely (no network call), for air-gapped or privacy-sensitive deployments.

## Implementation notes

- Core gains `BUILD_DATE` next to `VERSION` (both sed-stamped by `.github/workflows/release.yml` in the release and npm jobs); `normalizeVersion`/`compareVersions` moved from the CLI into core's `internal/version.ts` (re-exported from both) so the server shares them.
- New authenticated routes: `GET /api/version`, `GET /api/version/update-check` (always HTTP 200; `error: network|rate_limited|bad_response`; success cached 1 h, failure 10 min; injectable fetch/clock), `POST /api/version/update` (admin only). `penguin server|web` exports `PENGUIN_CLI_ENTRY` before importing the server; the route spawns `node <entry> update --yes` with a fixed argv (no request data reaches the spawn), forces `PENGUIN_LANG=en`, and classifies refusals via the CLI's own message fragments (safe: server and CLI ship in lockstep). Concurrent update requests share one run.
- Docs updated: server-api (Version section), web-app (version & updates), configuration (env var) — en + zh.

## Verification

- `pnpm build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (skills 16, docs 11, landing 36, core 492+2 skipped, server 352 incl. new `version.test.ts` with 12 tests, cli 188, web 383)
- Release-workflow sed stamping simulated locally (initial stamp and re-stamp both work); built CLI smoke-tested from a source checkout (`update --yes --release v99.0.0` → refusal text, exit 0, classified "unsupported"); live GitHub latest-release API shape verified (`tag_name`/`html_url`/`published_at`).
- e2e: skipped (not part of this change's verification set).


hiyouga:
加上更新日期吧，另外新建对话的页面也可以在副标题下面加一行小字，写版本和更新日期（以及可选的更新提醒

hiyouga:
## Review

The server side is careful in the ways that matter for a self-update endpoint: fixed argv with nothing from the request flowing into the command, admin gate, in-flight dedup so two admins cannot race two installers, timeout with a SIGKILL escalation, and an update check that is genuinely fail-soft (never throws, caches both outcomes, has an env opt-out). One real problem, in the docs.

### 1. The "Projects and Members" section heading was deleted, not preceded

In both `packages/docs/content/server-api.en.md` and `server-api.zh.md` the diff replaces the existing heading instead of inserting above it:

```diff
-### Projects and Members
+### Version and Self-Update
+
+| Method | Path | Description |
...
```

On the branch, `server-api.en.md` now reads: `### Version and Self-Update`, the three version endpoints, the `update-check` paragraph, and then — with no heading of its own — the `/api/projects` + members table and its "Member writes are owner-only." line. `### Projects and Members` no longer exists in the file (same for `### Project 与成员` in the zh page).

That is user-visible on the docs site, not just cosmetic: `extractToc` collects `##`/`###` headings, so the page's table of contents loses its Projects and Members entry, and the anchor it generated (`#projects-and-members`) 404s for anyone who linked it. Fix is to keep the old heading and add the new section above it:

```diff
+### Version and Self-Update
+
+| Method | Path | Description |
+| ... |
+
+`update-check` is ...
+
+### Projects and Members
+
 | Method | Path | Description |
```

### 2. `classifyUpdateRun` keys off English message fragments

The doc comment states the reasoning (exit codes cannot separate "refused" from "updated", `PENGUIN_LANG=en` pins the language, CLI and server ship from the same install) and I agree it holds today. Worth making it hold tomorrow too: the five strings in `REFUSAL_MARKERS` are duplicated from `packages/cli/src/i18n.ts` with no link between the two, so a copy-edit to `update.sourceCheckout` silently reclassifies a refusal as `updated` + `needsRestart: true`. A test in the CLI package asserting each `update.*` refusal string still contains its marker substring would fail loudly at the point of edit — cheaper than exporting a shared constant across the dependency boundary.

### Minor

- `BUILD_DATE` stamping: `grep`/`sed` both use BRE, so the unescaped `|` in `string | null` is a literal and matches — correct, though it reads like a regex bug at first glance. A short comment saying "BRE: `|` is literal here" would save the next reader the double-take.
- `useVersionInfo`: on failure the shared promise is cleared so a later dropdown open retries, but `versionCache`/`updateCache` are module-level and never invalidated, so a long-lived tab keeps showing a stale "new version available" after the admin has updated and restarted. Given the dot is only a reminder, that is probably fine — flagging it as intent rather than a bug.
- Self-updating in place swaps `web-dist` under the running server, so a browser session that reloads between the update and the restart can pull mixed assets. The dialog's restart hint covers it in practice.

hiyouga:
Addressed the review in 62937b9:

1. **Docs heading deleted** — restored `### Projects and Members` / `### Project 与成员` in both `server-api.en.md` and `server-api.zh.md`; the new Version section now sits above it and the projects/members table keeps its heading, TOC entry, and anchor.
2. **Update date for current installs** — `UpdateCheckService` now also resolves the running version's own release date from `releases/tags/v<version>` (same headers/timeout, shared cache entry and TTLs, strictly fail-soft; skipped when `PENGUIN_UPDATE_CHECK=off` or when `BUILD_DATE` is already stamped). `UpdateCheckResponse` gains `currentPublishedAt: string | null` (JSDoc + docs field lists). The sidebar footer date is `buildDate ?? currentPublishedAt` (date-only): the version renders as soon as `/api/version` resolves and the date appears when the update check lands — the footer never waits for it.
3. **Draft page version line** — a quiet `PenguinHarness vX.Y.Z · <date>` line under the brand subtitle (`text-xs text-gray-400`), plus the localized "new version vX.Y.Z" release link (`target="_blank" rel="noopener noreferrer"`) when one is known. Reuses `useVersionInfo(true)` (module cache prevents refetch storms) and the existing `S.update.*` strings; nothing renders until the version resolves.
4. **REFUSAL_MARKERS drift guard** — new `packages/cli/test/update-refusal-markers.test.ts` hardcodes the five marker fragments (with a pointer at `packages/server/src/http/routes/version.ts` REFUSAL_MARKERS) and asserts each en `update.*` refusal string still contains its marker, so editing a refusal message fails loudly at edit time.
5. **release.yml** — a comment at both BUILD_DATE stamp spots notes the unescaped `|` is literal under POSIX BRE.

Not changed, per review triage: the `useVersionInfo` module-cache staleness (flagged as intended — the cache lasting the browser session is by design), and mixed assets between update and restart (already covered by the restart hint).

Verification: `pnpm build`, `pnpm typecheck`, `pnpm test` (skills 16, docs 11, landing 36, core 492, server 355 incl. new current-date/skip/fail-soft cases, cli 193 incl. the drift guard, web 383 — all green), `pnpm format && pnpm format:check` clean.

hiyouga:
用人可读的方式写 Jul 26 或者 7 月 26 日，前面加上 最近更新日期 六个字

hiyouga:
Done in 2cab4c8 — the version date is now human-readable with the requested label, in both surfaces (sidebar user-menu footer and the draft page's version line):

| Locale | Before | After |
| --- | --- | --- |
| zh | `PenguinHarness v0.1.2 · 2026-07-26` | `PenguinHarness v0.1.2 · 最近更新日期 7 月 26 日` |
| en | `PenguinHarness v0.1.2 · 2026-07-26` | `PenguinHarness v0.1.2 · Last updated Jul 26` |

Implementation: a new `formatMonthDay(iso, locale)` in `packages/web/src/lib/format.ts` (en `Jul 26`, zh `7 月 26 日`, no year; parses the date fields from the string so a viewer's timezone can never shift the day, and keeps the exact spacing of the requested wording, which `Intl` would drop) feeding a shared `S.update.lastUpdated(date)` string (zh 「最近更新日期 …」 / en "Last updated …"). Unit tests cover the wording, ISO-timestamp inputs, and passthrough of unparsable values; the web-app docs paragraph was updated to match. `pnpm build` / `typecheck` / `test` / `format:check` all green.

hiyouga:
ci-windows on the merge ref caught a platform-naive test: `cliEntryFor`'s expectations compared `path.resolve` output against POSIX literals, which gain a drive prefix on win32 (the job reaches this branch via the merge with current main, which now carries #79's Windows CI). Expectations now resolve through `path.resolve` themselves — the contract under test is "resolved absolute", not a literal. CLI suite green locally; awaiting the re-run.

hiyouga:
这个最新更新日期不需要联网，以后每个包发布时候就要存储发布日期，直接显示。这个最新更新日期值得是用户当前所用版本的更新日期。如果有新版本，就在后面上标按钮写一个有新版本可用。

hiyouga:
Done in 6ddb2ba.

**Reverted (the date no longer touches the network):** the `releases/tags/v<current>` fetch in the update-check service, the `currentPublishedAt` field on `UpdateCheckResponse`, its tests, and the web fallback `buildDate ?? currentPublishedAt` are all gone. The displayed date now comes only from core's `BUILD_DATE`, which the release workflow stamps at publish time — verified that both jobs stamp it: the `release` job (tarball artifact) and the `publish-npm` job, so every future release carries its date offline. When `BUILD_DATE` is null (dev builds, and v0.1.2 which predates the stamping) the line shows the version alone. The update check itself stays — it only powers the new-version reminder now. Docs and JSDoc state the semantics explicitly: 最近更新日期 / "Last updated" is the **running version's release date, stamped at build time**.

**New-version superscript badge** — when the check finds a newer release, both version lines gain a tiny raised accent pill right after the date (or the version when no date): zh 「有新版本可用」 / en "New version available", a real button/link with aria-label naming the concrete version.
- Sidebar footer: admins → opens the update dialog; other users → release page in a new tab (mirrors the dropdown affordance above it; without a release URL it degrades to a plain pill).
- Draft page: release-page link (`_blank` + `noopener noreferrer`), replacing the previous inline text link on that line.

Line examples: `PenguinHarness v0.1.3 · 最近更新日期 7 月 27 日 ⁽有新版本可用⁾` / `PenguinHarness v0.1.3 · Last updated Jul 27 ⁽New version available⁾`.

Gates all green: `pnpm build` / `typecheck` / `test` (server back to 352 with the reverted cases removed, cli 193 incl. the platform-agnostic cliEntryFor expectations, web 386) / `format` + `format:check`.

hiyouga:
在用户设置里可以有一个手动检查更新按钮吗？

hiyouga:
Added in 96545d3 — a manual "check for updates" action in the user settings menu.

**Where it sits:** in the sidebar user menu's bottom section, directly above the version footer line it refreshes (same action-row style as "Change password"). Label 「检查更新」 / "Check for updates"; while running the row disables and reads 「检查中…」 / "Checking…".

**Behavior:**
- The click calls `GET /api/version/update-check?force=1` — a new optional query that bypasses the server's TTL cache and performs a fresh GitHub lookup (the cache exists to shield the unauthenticated rate limit from passive sidebar opens; an explicit user click is rare enough to press through it). The outcome lands in the same cache, so subsequent passive checks reuse it; concurrent callers still share one in-flight lookup, and `PENGUIN_UPDATE_CHECK=off` stays authoritative — force never dials out then.
- When a newer release is found, the result is broadcast to every mounted surface at once: the menu's new-version row, the avatar dot, and the superscript badges on both version lines appear immediately — that visible change is the feedback.
- When nothing changes visibly, exactly one toast fires: up to date → 「已是最新版本」 / "You're on the latest version"; checks disabled → info toast naming `PENGUIN_UPDATE_CHECK=off`; a failed lookup (fail-soft `error` field) or an unreachable server → error toast.

Server tests cover the route wiring, the warm-cache bypass + recache, and the opt-out; docs updated (server-api row notes `?force=1`, web-app paragraph gains the manual-check sentence, both languages). Gates all green: build / typecheck / test (server 355, web 386, cli 193) / format + check.

hiyouga:
检查更新放到修改密码下面吧，版本号放到检查更新右侧，不用写PenguinHarness

hiyouga:
Done in 70e145b — the menu block now looks like this:

**Before**
```
──────────────────────────────
 修改密码
 用户管理                (admin)
 退出登录
──────────────────────────────
 检查更新
 PenguinHarness v0.1.2 · 最近更新日期 7 月 26 日 ⁽有新版本可用⁾
```

**After**
```
──────────────────────────────
 修改密码
 检查更新              v0.1.2 ⁽有新版本可用⁾
 用户管理                (admin)
 退出登录
```

- "Check for updates" sits directly below "Change password" in the account block (its own bordered section is gone); the running version renders muted on the right of the same row as plain `v0.1.2` — no "PenguinHarness" prefix anywhere in the menu. While checking, the left label swaps to 「检查中…」/"Checking…" and the version stays put.
- The 最近更新日期 / "Last updated Jul 26" date is now the row's hover tooltip rather than visible text — chosen to keep the row uncluttered; easy to adjust (e.g. render it inline next to the version, or drop it) if you'd prefer.
- The superscript 「有新版本可用」 badge rides along to the right of the version as a passive indicator there: HTML forbids nesting a button/link inside the row's button, and whenever the badge shows, the clickable paths (release-notes link, and "Update now" for admins) are already present as the menu's reminder rows directly above. Also adjustable if you'd rather have the whole row split into two click targets. The avatar dot and the draft page's version line are unchanged.

Gates all green: build / typecheck / test (server 355, cli 193, web 386) / format + check.

PR #75, opened 2026-07-26
feat(web): live header statistics while a task runs

hiyouga:
## Summary

The chat page header's top-right statistics (Tokens / Cost / Elapsed chips) previously only refreshed when a task finished. They now update in real time while a task runs:

- **Elapsed** ticks once per second during a run, showing the settled cross-task cumulative plus the running task's wall clock so far. When idle it renders exactly the same settled value as before, with no timer running. The formula lives in a new pure helper `liveSessionElapsedMs` (`task-stats.ts`); at task end the model folds the task's elapsed into the cumulative in the same update that stops the tick, so nothing is double-counted.
- **Cost** shows the last fetched session cost plus a live estimate of the current task, converted from the task's live usage buckets with the session model's pricing via the new shared `bucketCostUsd` (which now also backs the per-reply turn cost, so the two conversions cannot drift). Subagents may run on different models — the estimate applies the main model's pricing and may be slightly off mid-task; the existing idle-gated `getUsage` refetch still reconciles to the authoritative server-recorded value at task end. Without pricing the live addition is skipped and today's uncosted (`*`) semantics are unchanged.
- **Tokens** already advanced per completed LLM request; its data source is untouched.
- The chip row and the header info dropdown now render the same three values from one shared computation (`headerStats`) instead of duplicating them.

Out of scope (unchanged by design): mobile visibility of the chip row (`hidden sm:flex`), the per-reply stats footer, and the stream protocol (`token_usage` cadence stays per-request, so cost/tokens advance per completed request while elapsed ticks every second).

## Verification

All commands run at the worktree root on top of `f6ebf5d` (v0.1.2):

- `pnpm build` — pass (only the pre-existing non-blocking `link:cli` global-bin-dir warning).
- `pnpm typecheck` — pass (all packages).
- `pnpm test` — pass: server 340, cli 186, web 389 tests (includes 7 new unit tests in `packages/web/test/task-stats.test.ts` covering elapsed accumulation across running/idle boundaries with no double count, and cost bucket conversion incl. the uncosted path).
- `pnpm format && pnpm format:check` — clean.
- E2E: `e2e/run.sh` does not forward single-spec arguments to Playwright, so the full web suite was run: `SRV_PORT=8960 MOCK_PORT=8961 pnpm --filter @prismshadow/penguin-web test:e2e` (non-default ports; another process occupied 8930 on this machine). Result: **15 passed, 3 failed** — `chat.spec.mjs` passes, including a new live-tick assertion (the header elapsed chip text advances while the task is running/awaiting approval) and the existing header/footer cost assertions. The 3 failures (`draft-user-scope.spec.mjs`, `paging.spec.mjs`, `skills.spec.mjs`) are pre-existing in this environment: a baseline run of the identical suite on clean `origin/main` (changes stashed, web dist rebuilt) fails the exact same 3 specs with identical errors, all in areas untouched by this change (login form load, sidebar paging count, skills install modal).

### DIFF
--- a/packages/web/e2e/chat.spec.mjs
+++ b/packages/web/e2e/chat.spec.mjs
@@ -59,6 +59,14 @@ test("chat + tool approval + stats/cost/copy + traces + files", async ({ page })
   await expect(page.getByText("exec_command").first()).toBeVisible();
   // Thinking + tool calls are wrapped in a work group; header shows running/done status.
   await expect(page.getByText("运行中").first()).toBeVisible();
+
+  // Live header statistics: the elapsed chip ticks once per second while the task runs (the
+  // pending approval below keeps it running), so its text must advance with no further server
+  // event. Scoped to the header stats container (div.hidden …): the per-reply footer reuses
+  // the same 用时 ("elapsed") label once the turn's stats line lands.
+  const headerElapsed = page.locator('div.hidden span[title="用时"]');
+  const elapsedBefore = await headerElapsed.textContent();
+  await expect(headerElapsed).not.toHaveText(elapsedBefore);
   // The user takes control of the running work group (toggle = userToggled), keeps it open, and
   // opens the exec_command card to watch the arguments. Both must survive the end of the turn.
   //
--- a/packages/web/src/features/chat/chat-page.tsx
+++ b/packages/web/src/features/chat/chat-page.tsx
@@ -12,6 +12,7 @@
  * created. The Session list and the new-chat entry point live in the global sidebar.
  */
 import { useCallback, useEffect, useRef, useState } from "react";
+import type { ReactNode } from "react";
 import { useNavigate, useParams } from "react-router";
 import type {
   AgentSummary,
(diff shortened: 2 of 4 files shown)


hiyouga:
## Review — LGTM

Extracting `bucketCostUsd` so the per-reply turn cost and the header's live estimate cannot drift is the right move, and the test that asserts the live buckets land exactly on `endTask`'s settled buckets pins the property that actually matters. `liveSessionElapsedMs` is pure and its no-double-count-across-the-boundary test covers the one case that would be easy to get wrong.

I checked the double-count risk on the cost side: the `getUsage` effect early-returns unless `stream.taskState === "idle"`, and its cleanup sets `cancelled` on the transition into running, so `sessionCost` cannot absorb the running Task's usage while `liveCost` is also being added. Good.

Two cosmetic consequences of the new `costUsd = (sessionCost ?? 0) + liveCost`, both worth a moment's thought since the old code went out of its way to avoid exactly this shape of thing:

- On a session with no recorded usage yet (`sessionCost === null`) and a Model that *does* have pricing, the chip now appears the instant a Task starts, showing a formatted zero, because `liveCost` is `0` before any token lands. The comment right above it says a "—" was rejected because "that would take up space while saying nothing, only making people think the cost is zero or something's broken" — a literal `$0.00` reads the same way for the second or two before the first usage event. Gating the live addition on `liveCost > 0` would keep the chip hidden until there is something to show.
- At task end `taskOpen` flips false immediately, so the chip drops back to the pre-task `sessionCost` and then jumps to the reconciled total when the idle `getUsage` returns — a visible backwards step. Minor, and it self-corrects.

Nit: `SessionElapsed` recreates its 1s interval whenever `taskOpen` flips and also calls `setNow(Date.now())` on entry, so the first render after a task starts is not a stale `now` — worth a word in the comment that the `setNow` is load-bearing rather than redundant.

hiyouga:
Addressed in 7c2b2b7:

- **$0.00 flash at task start** — the live cost path is now gated on having something to say: `liveCost > 0 || sessionCost != null`. A brand-new session's chip stays hidden until the first token_usage lands (no formatted zero in the gap), while a session with an already-recorded cost keeps showing it from the task's first instant. The gate sits next to the `headerStats` computation with a comment, since that logic is component-local rather than a testable helper; `bucketCostUsd` and its tests are unchanged.
- **`setNow(Date.now())` on effect entry** — commented as load-bearing: `now` still holds mount time or a previous task's final tick, and the first interval callback is a second away, so the first live render after a task starts must re-anchor immediately.

Not changed: the brief step back at task end (chip dropping to the pre-task session cost until the idle getUsage lands). It is transient (~1s) and self-correcting; holding the last live estimate instead would risk displaying a wrong number for longer than the true value takes to arrive.

Verification: `pnpm build`, `pnpm typecheck`, `pnpm test` (all suites green, incl. web unit tests), `pnpm format && pnpm format:check` clean.

PR #76, opened 2026-07-26
fix(web): keep running-state rows to one line on mobile

hiyouga:
While a task is running, the running-state UI accumulated enough text (per-tool call descriptions from #62, the steering-era composer from #63, approval labels, header stats) that phone-width layouts wrapped or squeezed. This PR keeps every running-state row to a single line below the `sm` breakpoint by hiding text and keeping the colored icon, with `title`/`aria-label` preserving the meaning and accessible names. At `sm` and above nothing changes.

## Per-surface changes

- **Per-reply stats footer** (`task-stats-line.tsx`) — the real height bug: the row is `h-5` but was `flex-wrap`, so at 390px the cost/elapsed chips wrapped onto a second row that painted over the content below (clearly visible on hover). Now: no wrap + `overflow-hidden` + `whitespace-nowrap`; the input/output/TPS chips hide below `sm`, keeping timestamp + cost + elapsed + copy on the one fixed line.
- **Chat page header** (`chat-page.tsx`) — the "Running" text plus the "Open workspace" label squeezed the session title to a few characters at 390px. Below `sm` the running indicator is just the pulsing emerald dot (title keeps the wording) and the workspace button is icon-only (`title` + `aria-label` keep its name). Single row always; the title gets the reclaimed room.
- **Approval buttons** (`approval-buttons.tsx`) — Allow/Deny drop their text below `sm` and keep a colored glyph (✓ on the accent primary button, red ✕ on the secondary one). `aria-label` pins the accessible name at every width, so `getByRole("button", { name: "允许" / "Allow" })` in chat.spec / compact-abort.spec / layout.spec keeps resolving.
- **Work-group header** (`work-group.tsx`) — kept: status icon + Running/Done label + live elapsed. The steps count hides below `sm`, and the awaiting-approval pill collapses to a bare amber dot with `title`/`aria-label`. One uncut line with no clipped chevron even with long durations.
- **Tool-call card** (`tool-call-card.tsx`) — the collapsed row's "awaiting approval" text hides below `sm` (the labeled amber hourglass StatusIcon already marks the wait); the pending-approval block's name+arguments row drops `flex-wrap` so the preview truncates (`min-w-0`) instead of wrapping under the name pill.
- **#63 steering UI** — inspected at 390px: the steer-queued hint and follow-up-queued chip are single short text lines that fit, and the More-settings control is already icon-only; no change needed.

## Regression test

`layout.spec.mjs`'s running-state block now also checks, at 390x844 with a pending approval: no horizontal document overflow, the work-group header stays single-line (clientHeight ≤ 40), Allow/Deny stay reachable by role+name, no overlapping text; then approves and asserts the stats footer keeps to its one fixed line (`scrollHeight ≤ clientHeight + 1`). Verified red on the pre-fix build (footer scrollHeight 39 vs 23) and green with the fix.

## Verification

- `pnpm build` — pass (exit 0)
- `pnpm typecheck` — pass
- `pnpm test` — pass (web: 31 files / 383 tests; all other packages green)
- `pnpm --filter @prismshadow/penguin-web test:e2e` — 15/18 pass, including chat.spec, subagent.spec, and all four layout.spec tests with the new running+approval assertions. The 3 failures (draft-user-scope, paging, skills) are pre-existing in this environment: re-running exactly those three against a clean origin/main build (diff stashed, web rebuilt) reproduces the same three failures with identical signatures (login form load timeout / sidebar paging count 10 vs 20 / skills install modal).
- New layout assertions verified red on the pre-fix build (stats footer scrollHeight 39 vs clientHeight 23) and green with the fix.
- Before/after screenshots (zh + en, 390x844 and 1280x720) taken against a live server driven by the e2e mock and inspected: at 390 the header title is readable again, approval buttons are icon-only, and the stats footer no longer wraps onto a clipped second row; the 1280 before/after pairs are identical apart from run-dependent timers.


hiyouga:
现在手机端显示的对话统计信息不全，而且已批准·自动的按钮还是会出现换行，导致高度被撑大，需要修复一下

hiyouga:
## Review — LGTM

The one-line-on-mobile pass is consistent and the e2e additions test the right invariants (document scroll width, header `clientHeight`, footer `scrollHeight <= clientHeight + 1`) rather than screenshots.

I checked the two spots where dropping `flex-wrap` could have caused clipping instead of truncation, and both are safe:

- `tool-call-card.tsx` pending row: the name pill is `shrink-0` and the preview is `min-w-0 flex-1 truncate`, so the preview truncates as the comment claims.
- `task-stats-line.tsx`: the copy button gained `shrink-0`, so `overflow-hidden` clips the stat chips rather than the button.

Keeping the accessible names via `aria-label` on the icon-only Allow/Deny buttons is the right call — the e2e specs resolve them by role+name and keep working.

One nit: the collapsed approval indicator in `work-group.tsx` uses `role="status"`, which makes it an ARIA live region — screen readers will announce it every time it appears, and it re-appears on re-render of the collapsed header. Since the amber pill above it (`sm:inline`) is the same information for wider screens and carries no live semantics, `role="img"` (or just `title` + `aria-label`) would match the pill's behavior and be less chatty.

hiyouga:
Addressed all three review points in bb47e4c:

1. **Full stats on mobile** — every chip (input / output / TPS / cost / elapsed) is back below `sm`. The footer keeps its one fixed `h-5` line: the timestamp + chips span scrolls horizontally with a hidden scrollbar (the existing `.no-scrollbar` utility), and the copy button sits **outside** the scroll area, so it stays pinned at the row's end instead of scrolling out of reach. `layout.spec` now asserts all chips present at 390, single-line footer height, the copy button on-screen, and no document-level horizontal overflow (inner scroll containers only).
2. **"已批准 · 自动" no longer wraps** — the decision pill is `shrink-0` + `whitespace-nowrap` (the subtitle remains the row's only shrinking column), and below `sm` it draws only the decision half ("已批准"/"Approved") with the full wording kept on `title`/`aria-label` (`role="img"`, one non-live node). Verified at 390x844 with harness screenshots for both the zh auto case (allow-all session) and the en manual case: the tool-card header measures 35px — one line, no growth.
3. **Collapsed approval dot in work-group** — `role="status"` dropped; it now uses the same non-live semantics as the text pill: `role="img"` + `title`/`aria-label`.

Also recorded the PR's mobile one-line batch in `changelog/unreleased/` per the repo changelog rules.

Verification: `pnpm build`, `pnpm typecheck`, `pnpm test` (all unit suites green, 383 web tests), `pnpm format:check` clean. Web e2e (ports 8970/8971): 15 passed, 3 failed — only the known pre-existing failures on clean main (draft-user-scope, paging, skills); `layout.spec` and `chat.spec` pass including the new 390px assertions.

hiyouga:
现在手机端的对话统计信息都不显示，修一下。另外已批准 已拒绝可以变成对勾和×？手机端审批时候命令显示不全。 aborted 和已拒绝不需要重复写两遍。

hiyouga:
All four points addressed in 7d47a8a:

1. **Stats not showing on phones — diagnosis: pre-existing, not a regression of this PR.** The footer has always been `opacity-0` + `group-hover:opacity-100`, and Tailwind v4 scopes `hover:` variants to `@media (hover: hover)` — on a touch-primary device (`hover: none`) the reveal can never fire. Reproduced on the pre-fix build with real phone emulation (390x844, touch): computed opacity stayed `0` at rest **and after tapping the reply**. Fix: below `sm` the footer is now always visible (`sm:opacity-0` keeps the ≥sm hover-reveal exactly as it was — chat.spec still asserts hidden-at-rest / visible-on-hover at desktop; layout.spec now asserts opacity `1` at rest at 390).
2. **已批准/已拒绝 → ✓/✕.** Below `sm` the decision indicator is a bare colored glyph — ✓ emerald for approved, ✕ red for denied — the same glyph language as the Allow/Deny buttons; the full wording (`已批准 · 自动` / `Approved · manual` …) stays on `title`/`aria-label`. ≥sm keeps the text pill (asserted in chat.spec).
3. **Full command readable while approving.** While a call is pending, the args/command preview wraps in whole below `sm` (expanded-args style: `pre-wrap` + `break-all`, the block may grow, no inner scroll) so the user reads exactly what they approve; once decided the pending block unmounts and the one-line treatment resumes. Desktop verified unchanged: the preview keeps its one-line truncate (`white-space: nowrap` asserted in chat.spec). e2e asserts the wrapped, un-clipped preview at 390.
4. **aborted + 已拒绝 written twice.** The duplication: a denial feeds back a tool output with `stop_reason: "aborted"` (core deny path), so the card header showed both the amber `aborted` badge and the Denied pill — plus the status icon's `aborted` hover label. Rule implemented: when `decision = deny` and the output stop reason is that denial-shaped `aborted`, the decision indicator is the card's single marker (badge and icon label dropped, the icon stays as a decorative red glyph); a genuine user-abort of a RUNNING tool carries no deny decision and keeps its own `aborted` marker. e2e asserts a denied card renders the ✕ indicator and zero `aborted` texts.

Verification: `pnpm build` / `typecheck` / `test` / `format:check` all green. Web e2e (ports 8970/8971): 15 passed, 3 failed — only the known pre-existing failures (draft-user-scope, paging, skills); layout.spec and chat.spec pass including all new assertions. Touch-emulated screenshots (hover:none, taps only) inspected for all four items at 390x844, zh and en.

hiyouga:
现在用户消息的复制还是不可见，在触屏模式下。移动端可以不显示TPS，减少小数位数，避免统计信息宽度溢出

hiyouga:
Both points addressed in 91f886e:

1. **User-message copy invisible on touch** — same root cause as the reply footer (hover-gated opacity never fires under `hover: none`): the message footer (time + copy, bubble bottom-right) is now always visible below `sm` (`sm:opacity-0`), with the ≥sm hover-reveal untouched. Verified under touch emulation: computed opacity `1` at rest, no tap needed; layout.spec asserts it at 390 and chat.spec keeps asserting the desktop hover behavior.
2. **Stats width on mobile** — the row is now slimmed to *fit* rather than scroll:
   - **TPS dropped below `sm`** (still shown from `sm` up; the chip stays in the DOM, `hidden sm:flex`).
   - **Fewer decimals below `sm`** via opt-in compact formatter variants — concrete rules, open to redirection:
     - **Cost**: sub-unit amounts keep **2 significant digits** instead of 4 fixed decimals — `$0.1234` → `$0.12`, `$0.0012` stays `$0.0012`, `$0.00047` stays `$0.00047`. Significant digits (not fixed decimals) so a nonzero cost is never rounded down to `$0.00`; amounts ≥ 1 are already short (`$1.50`) and unchanged.
     - **Elapsed**: settled durations from **10s up drop the tenths** (`12.7s` → `13s`); below 10s the tenth is significant and stays (`1.7s`); `ms` and `1m3s` forms are already whole numbers.
     - **Tokens**: `humanizeTokens` is already compact (`260`, `16.3k`) — unchanged; dropping the `.3` would lose real information for one character.
     - **Desktop formatting is byte-identical** — compact is an opt-in parameter used only below `sm`.
   - The chip gap also tightens one step below `sm` (`gap-x-2`): with TPS gone and compact decimals the priced en row was still ~7px over 390 (timestamp width), and this closes it honestly. Measured at 390 under touch emulation, the full priced row (`Jul 27, 8:19 AM · ↑260 · ↓50 · $0.0012 · 694ms · copy`) now fits with zero scroll (307/307 px en, 295/295 px zh); the hidden-scrollbar sideways scroll remains only as a fallback for extreme values.

Verification: `pnpm build` / `typecheck` / `test` (new unit tests cover both compact variants) / `format:check` green. Full web e2e: 15 passed, 3 failed — only the known pre-existing failures (draft-user-scope, paging, skills); layout.spec asserts the user footer's at-rest visibility, the hidden TPS chip, and the no-scroll fit at 390, with all desktop assertions unchanged. Touch-emulated screenshots inspected for both items, en and zh.

hiyouga:
工具不用在耗时右侧显示已放行还是已拒绝了，左边图标就够用了。把这个可以删了。另外移动端的允许和拒绝（用户使用的按钮）还是要写文字，不能写符号。

hiyouga:
Both points addressed in 4de794b:

1. **Decision indicator next to the elapsed time: deleted** — at every breakpoint (the ≥sm text pill and the below-sm glyph both, qualifier included). The card's **left status icon is now the single carrier**: on an approved card it reads `已批准 · 手动/自动` ("Approved · manual/auto") via `title`/`aria-label` (`role="img"`), on a denied card `已拒绝 · …` — the denial's forced abort is the decision itself, so no `aborted` badge repeats it. A genuine user-abort of a *running* tool still keeps its own `aborted` label, and an approved call that later fails keeps the failure stop reason (the failure is then the story). Zero visible decision text, exactly one accessible source of truth. e2e now asserts the icon's accessible name (approved and denied) plus zero visible decision text on the card, at 390 and desktop.
2. **Allow/Deny buttons are words again** — the icon-only-below-sm treatment is reverted: `允许/拒绝` ("Allow/Deny") render as text at every breakpoint (action buttons the user presses are textual; iconic shorthand stays reserved for passive indicators). They sit on their own row beneath the full-wrap command preview, so the pending card fits 390 unchanged — verified under touch emulation (`Allow`/`Deny` innerText, document width 390/390, no overflow) and asserted in layout.spec via the buttons' rendered innerText.

Verification: `pnpm build` / `typecheck` / `test` / `format:check` green; full web e2e (isolated ports): 15 passed, 3 failed — only the known pre-existing failures (draft-user-scope, paging, skills). Touch-emulated screenshots at 390x844 inspected: pending card with text buttons + wrapped command, and approved/denied cards showing only the labeled status icon.

PR #77, opened 2026-07-26
fix(server,web): live in-progress output survives refresh and reconnect

hiyouga:
## Bug

While a Task runs, a long tool call (or a long reply) streams its output live. If the user refreshed or closed/reopened the session mid-stream, the in-progress message disappeared and only came back once it completed: streaming `partial_*` messages never reach the Trace (so `GET /messages` cannot contain them), and a fresh EventSource carries no `Last-Event-ID`, so the channel ring buffer replays nothing.

## Design

The server now keeps a live tail of every open streaming fragment per running session (`runtime/live-tail.ts`, fed by `SessionManager.drive` in the same synchronous tick as each publish, cleared when the run ends), and `GET /messages` returns it as an optional `live: {cursor, fragments}` field — the channel cursor and the fragment snapshot are captured atomically before the trace read, so every published event with id <= cursor is already reflected in the fragments. The client tags its buffered SSE events with their event ids; when `live` arrives with a matching epoch it drops buffered partials at/or before the cursor, feeds the synthetic `partial_* start` fragments (full accumulated prefix, origin preserved) through the normal reducer path at the cursor's position, and replays the rest — complete messages are never cursor-dropped, the existing overlap dedup decides for them.

## User-visible outcome

After a refresh or reopen mid-stream, the currently in-progress message is always there — assistant text, thinking, tool-call arguments, and partial tool output — with the already-streamed prefix present and streaming continuing live; the final state matches a never-reloaded run.

## Verification

All commands run at the worktree root:

- `pnpm build` — all packages build.
- `pnpm typecheck` — all packages pass.
- `pnpm test` — all suites pass: core 492, server 353 (incl. new `live-tail.test.ts` 10, `messages-live.test.ts` 2, extended `channel.test.ts`), web 393 (incl. new live-tail seeding cases in `stream-controller.test.ts` / `stream-model.test.ts`), cli 186, docs/skills/landing.
- `pnpm format && pnpm format:check` — clean.
- `SRV_PORT=8980 MOCK_PORT=8981 SKIP_BUILD=1 pnpm --filter @prismshadow/penguin-web test:e2e` — 17 passed / 3 failed:
  - New `reload-midstream.spec.mjs` **passed both tests**: (1) reload during a real slowly-streaming `exec_command` (one line per 200ms for ~8s) — prefix visible again within 4s of reload while the command is still running, output keeps growing live, and the final tool output is byte-identical to a control run that was never reloaded; (2) reload during slowly-streamed assistant TEXT — prefix back promptly, streaming continues to the full reply, every chunk appears exactly once (no duplicated overlap).
  - Regression proof: `chat.spec.mjs` and `subagent.spec.mjs` (plus compaction/compact-abort/malformed/draft/layout/sheet/files-switch/project-switch) all pass.
  - The 3 failures — `draft-user-scope.spec.mjs` (login form load timeout), `paging.spec.mjs` (sidebar paging count), `skills.spec.mjs` (skills modal) — are pre-existing failures on clean `origin/main` in this environment (identical signatures confirmed against a clean-main baseline run in a parallel worktree), unrelated to this change.

Note: the default e2e port (8930) was occupied by a foreign server in this environment; the suite was run on dedicated ports as shown above.


hiyouga:
## Review

The design here is the strong part: capturing cursor + fragments in one synchronous tick before the trace read, dropping only *partial* buffered events at/before the cursor while letting complete messages fall through to the normal overlap dedup, and skipping the whole weave on an epoch mismatch. That contract is stated identically in `MessagesLiveTail`, the route, and the docs, and the unit tests exercise each branch of it including the epoch-mismatch and trace-append-in-flight cases. The two e2e specs comparing a reloaded run against a never-reloaded control are exactly the right shape of test for this.

One concrete problem, and it is the reason this file did not actually get reviewed.

### `packages/server/src/runtime/live-tail.ts` contains a literal NUL byte

The PR renders it as:

```
diff --git a/packages/server/src/runtime/live-tail.ts b/packages/server/src/runtime/live-tail.ts
new file mode 100644
Binary files /dev/null and b/packages/server/src/runtime/live-tail.ts differ
```

Git is classifying a 230-line TypeScript source file as binary. The cause is line 80, in `fragmentKey` — the separator between the origin chain and the kind is a **raw U+0000 byte in the source text**, not the two-character escape `\0`. Confirmed by decoding the blob: exactly one control byte, at offset 4105, value `0x00`; the file is otherwise valid UTF-8.

The runtime behaviour is fine — U+0000 is legal inside a template literal, and it makes a good separator precisely because it cannot appear in a session id. The cost is entirely in tooling, and it is not small:

- GitHub shows "Binary file not shown" for the whole file, so the 230 new lines of the tracker — the core of this PR — are invisible in the PR diff.
- `git diff`, `git blame`, `git add -p` and conflict resolution all degrade to whole-file replacement for every future change to it.
- Grep-based tooling and editors that refuse to open binary files skip it.

The fix is one character — write the escape instead of the byte:

```ts
return `${(origin ?? []).join("/")}\0${kind}::${toolCallId}`;
```

`\u0000` reads more explicitly if you prefer, and `\u001f` (unit separator) would also work while keeping the file plainly text under any tool. Worth a short comment noting the separator cannot occur in a session id, since that is the property the key relies on.

Note this will not be rescued by the `.gitattributes` that #79 adds: `* text=auto` still classifies a file containing NUL as binary.

### Everything else I checked holds

- `Channel.lastEventId` returns `${epoch}-${nextSeq - 1}`, so it is `epoch-0` before any event — below every real seq, so a capture taken before the first publish drops nothing. Test covers it.
- `weaveLiveTail` compares epochs before seqs and bails out entirely on a mismatch, so a recycled channel cannot cause a silent mis-drop; buffered events with `id === null` always land in `post` and replay.
- Seeded fragments are filtered out when a mid-replay rebuild hands the remainder to the next round — they are snapshot-bound and the new round refetches fresh ones. Easy to get wrong, good catch.
- `LiveTailTracker` state is per-session and dropped in `drive`'s `finally` *before* the idle flip, so `GET /messages` cannot see a stale tail once `statusOf` reads idle.
- The docs change from 1000/2MB to 10,000/8MB matches what `channel.ts` already had on `main` — this corrects a stale doc rather than introducing a mismatch.

### Minor

- The fragment cap keeps the tail, which is right for text (it rejoins the live deltas seamlessly) but yields a syntactically broken prefix for `partial_tool_call`, where the buffer is JSON arguments. Unreachable in practice at 512KB of arguments, and the complete message reconciles — just noting the asymmetry looks incidental rather than chosen.

hiyouga:
Addressed the review: the raw U+0000 in `fragmentKey` is now the `\0` escape (file is plain text again — this commit's diff still renders as binary→text because the old blob contained the NUL, but every future diff/blame/grep works normally), with a comment recording why the separator cannot collide with session ids. The fragment-cap note now states the `partial_tool_call` tail-keeping asymmetry as a deliberate choice, per the minor. Server suite 353/353, typecheck + format clean.

PR #78, opened 2026-07-26
feat(web): subagent side panel with a live call-graph, compact chips in the stream

hiyouga:
Subagent conversations move out of the main chat flow into a dedicated right-hand side panel (same docked/resizable pattern as the Workspace files panel, bottom Sheet on mobile). The main stream keeps only a compact per-subagent shortcut chip.

## What changed

- **Compact chip in the stream** (`subagent-chip.tsx`): replaces the inline nested subagent card at both render sites (bound `run_subagent` tool cards and standalone items). One-line pill with the agent's avatar, resolved agent name, short session-id suffix, a spinner while the child runs, and an amber dot — also announced in the accessible name — while any approval is pending anywhere in the child's subtree. Clicking opens the panel focused on that child.
- **Subagents panel** (`subagents-panel.tsx` + `subagents-view.tsx` + `use-subagents-panel.ts`): docked at >=1024px with drag-to-resize (own persisted width key), bottom Sheet on narrow viewports. Top: a call graph of the **latest Task** — nodes carry the agent's avatar/name and a running/done glyph, edges show who spawned whom; clicking a node switches the conversation below (the root node shows a "main conversation stays in the chat area" note). Bottom: the selected child's live conversation, rendered with the same `MessageStream`/`MessageItems` machinery as the main chat, so streaming, nested tool cards, deeper chips, and approvals all keep working inside the panel. Starting a new Task resets the graph to the new Task's agents; a child focused from an older turn keeps its conversation visible with no highlighted node.
- **Topology as pure modules** (`agent-topology.ts`): latest-Task extraction (walks bound cards + standalone items, recurses into child models, dedupes by session id) and a layered-tree layout (depth = column, leaves take rows, parents center; orthogonal elbow edges) with fixed-size node boxes — no text measurement, unit-tested. Rendering (`agent-topology-view.tsx`) draws edges in SVG with real `<button>` nodes layered above for keyboard/AT support; wide trees scroll horizontally.
- **Nested identity capture** (`stream-model.ts`): a nested model now stashes `meta` (agentId parsed from the child session_meta `agent_state` path, plus provider/modelId/source) — the panel's live source for "which agent runs this child". Binding logic is unchanged. New pure helpers `agentIdFromStatePath` and `hasPendingWithinOrigin` (approval-key prefix match driving the chip/toolbar badges).
- **Chat page wiring**: a new toolbar toggle ("Agents panel", network glyph, `aria-expanded`, amber dot when any approval sits inside any subagent) next to the Workspace toggle; the two docked panels are mutually exclusive (opening one closes the other — coordinated in `chat-page.tsx`, not in the hooks). `StreamRenderContext` gains `onOpenSubagent`.
- **i18n**: new `subagentPanel` group + `chat.openAgents` in the zh and en dictionaries (same-shape enforced).
- **Docs**: the web-app page's subagent bullet now describes the chip + panel (en + zh).

## Walkthrough (no display available for screenshots)

Ask an agent to delegate: the spawning `run_subagent` tool card shows a small pill — avatar, "General Agent", `1a2b3c01`, spinner — instead of the old full nested transcript. Clicking it (or the "Agents panel" toolbar button) docks a panel on the right: "Call graph" on top shows root -> child nodes with a running ring on the active child and a brand-blue ring on the selected node; below it, the child's own conversation streams live. If a child's tool call awaits approval, the chip and the toolbar button gain an amber dot, and allow/deny work from inside the panel. On a new prompt the graph resets to just the root until the new Task spawns again.

## Verification

All commands run at the worktree root on the final code:

- `pnpm build` — pass (all packages; the trailing global `link:cli` step is skipped in this environment: no pnpm global bin dir).
- `pnpm typecheck` — pass (7 packages).
- `pnpm test` — pass (web 32 files / 401 tests incl. the new `test/agent-topology.test.ts`, 18 tests; cli 186; others green).
- `pnpm format && pnpm format:check` — pass.
- Full web e2e suite (`SKIP_BUILD=1 SRV_PORT=8990 MOCK_PORT=8991 pnpm --filter @prismshadow/penguin-web test:e2e`; dedicated ports because a foreign server occupies the default 8930 in this environment): **16 passed, 3 failed — the 3 failures are pre-existing on clean origin/main in this environment** (confirmed signatures: `draft-user-scope.spec.mjs` login-form load timeout, `paging.spec.mjs` sidebar count 10 vs 20, `skills.spec.mjs` install-button not found) and are unrelated to this change. The rewritten `subagent.spec.mjs` passes both tests: chip -> panel -> call graph (2 nodes) -> node switching -> mid-run and after-completion reloads -> child title + sidebar folder; and an always-ask scenario proving a nested approval stays discoverable (chip badge + toolbar dot) and actionable from the panel (the mock child now runs one `exec_command` before reporting).

## Design decisions where the spec left room

- Panel open state persists across session switches (Files-panel parity) but not across page reloads — the Files panel deliberately persists only width, and the e2e covers "chip reopens the panel after reload".
- Topology nodes are HTML buttons absolutely positioned over an edges-only SVG (rather than all-SVG nodes): keyboard focus, accessible names, and CSS truncation come for free; geometry stays in the pure layout module.
- The tool card's old "auto-expand on nested pending approval" is removed: the chip renders outside the collapsed area and carries the pending signal itself, so there is nothing left to reveal by expanding.
- Node label fallback order: agents list name -> agent id -> child session title -> short session id (chips fall back to the generic "Subagent" label since they always show the short id alongside).
- With no explicit selection the panel auto-focuses the current Task's first child (derived, not stored), which also covers "a new task spawns its first subagent while the panel is open".


hiyouga:
子智能体面板和工作区每次只显示一种，不要两个都显示。
如果用户点击过去会话的subagent，就可以显示过去的拓扑图。但是每次切换 Session，都默认只展示最新的拓扑。

hiyouga:
在主对话区里面的按钮不要用现在的样式，最好也变成一个长条，和别的思考中一样。

hiyouga:
在子整体拓扑图上也要显示运行时间，就用开始出现到当前的时间就行，不用计算里面的 items

hiyouga:
## Review — LGTM

The pure-extraction / pure-layout / rendering split (`agent-topology.ts` vs `agent-topology-view.tsx`, mirroring `chart-geom.ts` / `chart-svg.tsx`) keeps the interesting part unit-testable, and it is tested. Fixed-size node boxes to avoid text measurement is the right simplification at this graph size.

I checked the two places where this could plausibly be wrong, and both are correct:

- **`hasPendingWithinOrigin` prefix matching is collision-safe.** `approvalKey` is `` `${origin.join("/")} ${toolCallId}` ``, and the check requires the chain to be followed by either a space or a slash. So for `chain = ["c1"]`, a sibling child `c1x` (key `"c1x tool"`) does not match, a nested approval `"c1/c2 tool"` does, and a root approval `" tool"` correctly does not. The delimiter is doing real work here — worth a comment saying so, since a future refactor of `approvalKey` to a different separator would silently break it.
- **`layoutTopology`'s recursive `assignRow` cannot loop.** `extractTopology`'s `seen` set means every node is pushed exactly once with exactly one `parentId`, so `childrenOf` is a forest by construction. Fine as-is; no cycle guard needed.

Two nits:

- `resolveAgentLabel` uses `agent.name ?? agent.agentId`, so an agent whose `name` is the empty string renders an empty label rather than falling back to the id. `||` (or an explicit length check) matches the intent of the doc comment, which says "name, falling back to the agent id itself".
- `use-subagents-panel.ts` is a ~160-line near-clone of `use-files-panel.ts` — the header says so, which is honest, but the duplicated surface is the whole state machine (breakpoint listener, drag-resize, width clamp-on-resize, localStorage round-trip), not just the shape. Since there are now two of them, a `useSidePanel({ storageKey, dockQuery })` returning the shared parts would leave each hook with only its own focus/selection command. Not a blocker for this PR, but the third copy should not happen.

hiyouga:
Addressed the review round — the four directives plus the nits (commit db502bf):

1. **Panels never show together.** Audited every open path: both toolbar toggles, the file card (`onOpenFile`), the subagent row (`onOpenSubagent`), the mobile Sheets; there is no auto-open path. Each was already coordinated, but site by site — it is now centralized in chat-page: the panel objects handed to consumers wrap `setOpen`, so opening either panel from **any** path (including future call sites) closes the other. e2e asserts open files → agents closes, and the reverse.

2. **Historical topology.** `use-subagents-panel` keeps a `taskScope` (null = latest Task). Clicking a subagent row pins the graph to that row's Task via the new `extractTopologyForChild`: the Task slice containing the main-stream item that references the child, bounded by the nearest user messages on either side (falls back to latest when unreferenced, e.g. after a resync). Opening from the toolbar and switching Sessions reset to the latest Task; an older Task's root reads done even while the newest one runs. Unit-tested on a two-task fixture; e2e drives a second Task live (latest graph empties → the old turn's row pins task 1's tree back, node highlighted → toolbar reopen returns to latest).

3. **Bar-style rows.** The in-stream chip is now a full-width bar matching the other collapsed rows (work-group header / thinking row idiom: rounded-md border, muted background, px-3 py-2, left-aligned avatar + name + short id + spinner / amber pending dot). Accessible name and click behavior unchanged; both render sites updated, and the e2e selectors were role+name based, so they survived.

4. **Per-node elapsed time.** `routeNested` stamps each nested model on creation and on every routed message (cheap assignments — wall clock of the whole spawn, never a sum of its items). Two stamp pairs are kept: local wall clock and message timestamps. A running node ticks with the LiveDuration 1s idiom from its first appearance; a done node freezes at its last activity. Done durations derive from the **message-timestamp** pair, so a reloaded session shows the same values as the live view (a replay sets all local stamps at load time, which would collapse every span to ~0 — the local pair is only the unparseable-timestamp fallback). The root shows no duration. Unit tests cover the live, replayed, and fallback paths; e2e asserts the settled duration after a reload.

5. **Nits.** Empty-string agent names now fall back to the id (`resolveAgentLabel` and the root label in the view), and the `approvalKey` space/slash delimiter dependency is documented on both sides of `hasPendingWithinOrigin`'s prefix match. On extracting a shared `useSidePanel`: agreed on the rule — the third copy must not happen — but deferring as marked non-blocking: with two stable copies the extraction would churn the Files panel for zero behavior change, and its right shape will be clearer the moment a third panel actually appears.

Also added the missing changelog entry for the panel feature, and updated the web-app docs (bar row, per-node elapsed, historical topology, exclusivity). One follow-up outside this repo: `design/specs/06-PROTOTYPE.md` still describes the old inline nested rendering inside the `run_subagent` card — the design repo needs a matching update for the panel + row UX.

**Verification:** `pnpm build`, `typecheck`, `test` (408 web unit tests incl. the new topology/duration suites), `format:check` all green. Full web e2e on a clean server: 16/19 passed — subagent (extended), chat, layout, files-switch, sheet, compaction, draft, malformed, project-switch, compact-abort green; draft-user-scope / paging / skills fail identically on clean main (pre-existing, not chased).

hiyouga:
现在subagent面板里面的session不显示用户消息，修一下。另外如果用户新建对话再发送，那么智能体面板是不显示的。

hiyouga:
Both reports fixed (commit ebf1ad6):

**1. Child conversation showed no user messages.** Root cause in core, live path only: the engine writes a run's input to the child's own Trace but never replays it to its consumer (a session's normal caller typed that input itself) — for a spawned child that consumer is the parent, whose frontend had no other live source for the child's user side. So the panel showed user bubbles after a reload (child-Trace expansion splices them) but not while streaming — which one you got depended on a history-fetch vs task race, and with a real (slow) model the live view always lost. Fixed in the message flow itself, not synthesized: the subagent runner now forwards each run's input message (origin-tagged, after `session_meta`, before the run's output); `input_subagent` follow-up prompts ride the same generator. The parent engine drops origin messages from the parent Trace, so a reload never duplicates the bubble. Locked by a core unit test on the real runner (the input yield precedes any LLM request), plus e2e assertions live and after reload.

**2. New chat → send → the panel doesn't show.** Verified every mechanical path first — toolbar button, dock and Sheet mounts, the draft→real-session transition, mid-run and post-run timing, panel-open persistence across the draft — all worked (8 scripted variations, fast and slow model timings). What a draft-born session actually lacked was the panel showing *itself*: it starts closed, nothing opens it when the child spawns, and once the turn ends the collapsed step row hides the only trace. The chat page now **auto-opens the panel on the session's first live spawn** — only while the Task is running (opening an old conversation never hijacks the layout), at most once per session (closing it again is respected), desktop dock only (no uninvited mobile Sheet), and never over an open files panel. If you meant something else by 不显示, tell me and I'll chase that instead — but with this, a fresh conversation's first spawn puts the panel on screen with the child's prompt and live conversation in it.

Tests: core runner unit test (input forwarding), `latestTaskHasSubagent` unit tests (trigger arms only for the latest Task's spawns), and a new e2e driving the real draft flow: `/chat/new` → send → panel auto-opens with **no clicks** → child conversation shows its own user prompt live → reload → same prompt from the Trace-expansion path. The mock child's command now sleeps ~1s so the Task reliably outlives the draft navigation (deterministic, closer to a real model).

**Verification:** build / typecheck / unit tests (core 493, web 410, all packages green) / format:check clean; full web e2e 17/20 passed — subagent (3 tests incl. the new draft flow), chat, layout and the rest green; draft-user-scope / paging / skills still the known pre-existing failures.

hiyouga:
我的意思是新建任务以后，面板默认不要显示，只有用户手动打开，或者调用subagent时候才显示。现在如果我在旧任务打开了子智能体面板，新建任务发送以后，这个任务可能和子智能体没关系，他的面板还是打开状态，我认为他应该是关闭才对。

hiyouga:
Done (commit 2249a3f). The panel's visibility is now task-scoped — final lifecycle:

1. **A new task starts** (you send a message) → the panel closes, even if it was open during the previous task.
2. **Entering a session** → starts closed (no inherited open state; this panel drops the Files panel's cross-session persistence — the Files panel itself is unchanged).
3. **It reappears only via**: a manual open (toolbar toggle, or a subagent row — row clicks still pin that turn's historical graph), or the **current task spawning a subagent** — the auto-open is re-armed per task (desktop dock only; never over an open files panel).
4. **Manual close mid-task** → respected until the next task boundary; the next spawning task auto-opens again.

Boundary signal: the count of Task-starting user messages in the stream — the same user_text/user_image predicate the reducer's Task segmentation uses, so steering and internal injections never count while queued follow-ups do. All decisions live in one pure, unit-tested tracker (`advancePanelTaskScope`); the chat page just applies its actions under the layout guards.

e2e: a dedicated reload-free test drives the full cycle in one conversation — spawn → auto-open (stays open past the task's end); plain follow-up → closes at the boundary, stays closed; manual reopen → the next spawning task's boundary closes it, its spawn auto-opens it again (new mock branch "run another subagent", delayed so close→reopen is observable in order); manual close mid-task → holds through the task's end. The chip/graph/reload test keeps the historical-topology flow (a row click is a manual open).

**Separate pre-existing bug found while stabilizing this (out of scope, filing here for the record):** a follow-up task sent from a page that has been RELOADED can occasionally go dead — repro: subagent task → reload mid-run → reload after completion → send a plain message. Under load, ~1 in 4-6 runs either (a) the POST /tasks returns 202 but the task never runs (no messages, no task_state change, the LLM is never called), or (b) the input area is left in a stale running/steer state so the send takes the steering detour. Reproduced with a distilled script that never touches the subagents panel, and identically with this branch's core change reverted — it lives in the session-runtime/stream layer, not in this PR's UI work. The subagent spec now avoids asserting through that window (boundary assertions moved to the reload-free test, URL pin before the post-reload send).

Verification: build / typecheck / unit (core 493, web 417 incl. the new tracker suite) / format:check green; subagent spec 4/4 twice on fresh servers; full web e2e 18/21 passed — only the 3 known pre-existing failures (draft-user-scope / paging / skills).

hiyouga:
Rebased onto main (now over #66 goal mode, #74, #76, #82, #87 and the rest) and squashed to one commit, `26e3cad`.

**Where the seams landed:**

- `mock-llm.mjs` was the only true conflict — resolved as the **union** of both scenario chains. Order audited: the subagent branches ("run a subagent" whole-history keyed, the last-message-keyed "run another subagent", the sleep-1 child turns) and main's new branches (auth-401 / quota-403 key- and phrase-gated pre-header branches, "slow stream test" / "slow text test" for reload-midstream) have disjoint trigger phrases, so nothing shadows anything; my last-message-keyed branch still precedes the whole-history `hasToolResult` fallthrough, and main's slow branches sit after the subagent turns exactly as both sides intended.
- `agent.ts` auto-merged: the subagent runner's input forwarding (`yield withOrigin(userText(prompt), hop)` before `childSession.run`) is intact; goal mode's changes are confined to `goalFilePath` wiring in session construction — orthogonal to the runner. Core's runner test still passes.
- `chat-page.tsx` auto-merged: the exclusivity wrappers, the task-scope tracker effect, and the panel mounts coexist with goal mode's `goal` send parameter and `GoalStatusBanner`, #74's version UI and #76's toolbar changes. One deliberate semantic note: each goal-mode iteration starts a new Task, so the panel follows the task-scoped rule there too — it closes at every iteration boundary and auto-opens only if that iteration spawns a subagent, which is exactly the rule you asked for.
- Docs (`web-app.en/zh.md`) auto-merged with both sides' bullets intact; #76's responsive row tweaks don't affect the subagent row (its content already fits one line on mobile).

**Verification on the rebased tree:** `pnpm install --frozen-lockfile`, build, typecheck, format:check clean; full unit suites green including main's growth (core 568, server 398, cli 200, web 451). Full web e2e: **22/26 passed** — the subagent spec 4/4, and main's new llm-errors (3/3) and reload-midstream (2/2) specs green. Failures are all pre-existing: the known trio (draft-user-scope / paging / skills) plus `layout.spec.mjs:334` (mobile chat dropdowns, from #76) — that one fails **identically on a clean origin/main checkout** in this environment (deterministic, the draft page's Workspace dropdown reports viewport ratio 0 at 375px), so it's not introduced by this branch.

PR #79, opened 2026-07-26
feat(core,tooling): Windows support — shell selection, install.ps1, win-x64 release package, Windows CI

hiyouga:
## Summary

Windows could not run PenguinHarness at all — `exec_command` spawned `bash` unconditionally (ENOENT on every tool call), npm scripts used POSIX env prefixes, and there was no install method or Windows release artifact. This PR makes Windows a supported platform end to end:

- **Install one-liner**: `irm https://penguin.ooo/install.ps1 | iex` (pin with `$env:PENGUIN_VERSION`). `install.ps1` mirrors `install.sh`: sha256 verification, staged rename-then-delete swap of `bin`/`lib`/`web`/`node` that never touches `data\`, launcher shim generation, user `Path` registration, clear locked-file error. The `penguin.ooo/install.ps1` forwarder downloads fully and runs the installer as an in-memory script block, so the default `Restricted` execution policy cannot break it mid-install.
- **Release package**: `penguin-win32-x64.zip` — same `lib/` + `web/` layout as the tarballs, official win-x64 Node runtime (`node.exe` at the archive root, as nodejs.org ships it), `bin\penguin.cmd` + `bin\penguin.ps1` launchers (CRLF for cmd), `.sha256`, listed in `SHA256SUMS`, and `install.ps1` uploaded alongside `install.sh`.
- **Shell resolution** (core): `PENGUIN_SHELL` (explicit path/name, arg shape by basename) → non-Windows: `bash -lc` (today's behavior, bit for bit) → Windows: Git-Bash from PATH (best skill-ecosystem compatibility; the WSL System32 launcher is ignored) → `pwsh` → `powershell` (`-NoLogo -NoProfile -Command`). The chosen shell is shown to the model via a new `{{SHELL}}` placeholder (`Shell:` line in the default system prompt).
- **Kill path** (core): Windows has no signals/process groups — every escalation becomes `taskkill /pid <pid> /t /f` (whole-tree kill, `child.kill()` fallback, sync variant for the process-`exit` hook). `input_command`'s Ctrl-C degrades to a hard tree kill on Windows (documented in code).
- **Sandbox fix found by the Windows CI** (server): `fs.constants.O_NOFOLLOW` does not exist on win32, so the Workspace upload path's final-segment symlink guard silently vanished there — an upload onto an agent-preset symlink would overwrite a file outside the Workspace. Now guarded via `lstat` on win32; POSIX keeps the atomic `ELOOP` behavior.
- **Scripts**: `scripts/run-with-env.mjs` replaces the four POSIX env-prefix script lines (`${VAR:-default}` semantics with `~/` expansion, cmd-safe); `dev-prebuild.mjs` spawns pnpm through the shell on win32 (.cmd shims, CVE-2024-27980).
- **Line endings**: `.gitattributes` (`* text=auto eol=lf`, `*.sh` LF, `*.ps1` CRLF, png/webp binary) + Prettier `endOfLine: lf` (a no-op on the existing tree).
- **CI**: new `ci-windows` job on `windows-latest` (install → build → typecheck → test) plus a PowerShell parse gate for both installer scripts. POSIX-only tests (0600 mode-bit assertions, POSIX-layout `detectInstall` fixtures) skip via `process.platform` guards in the tests themselves, not CI filters.
- **Docs/landing/README**: Windows section in `installation.{en,zh}.md` (requirements Win10+/PS5.1+, npm alternative, shell resolution + `PENGUIN_SHELL`, upgrade = re-run installer, honest caveats), landing hero + quickstart show both OS install rows, READMEs gain the one-liner and the roadmap checkbox is flipped.

## What was audited

Fixed in this PR: `bash -lc` hard dependency (A1), signal/kill path (A2), POSIX-only npm scripts (A3), missing Windows release artifact + launcher (A4), `dev-prebuild` spawnSync without shell (A5), missing `.gitattributes`/`endOfLine` (A6), `/tmp` literals + chmod-mode assertions in tests, `config lang` dying on a `/bin/zsh` spawn (now a clear en/zh message pointing at `setx PENGUIN_LANG`), and the win32 `O_NOFOLLOW` symlink-upload gap the new CI surfaced.

Deliberately deferred: in-place `penguin update` on Windows (stays refused; documented upgrade path is re-running `install.ps1` — `update.ts` untouched, another PR owns it), `config lang` rc-file persistence on Windows (guarded refusal, not ported), server `SIGTERM` note on Windows (left as is), web browser e2e runner stays POSIX-only (`run.sh`; documented in `e2e/README.md`, not part of `pnpm test`).

Known caveats on Windows (also in the docs): config/vault files get default NTFS ACLs, not 0600 semantics; `penguin.ps1` shim can be blocked by a `Restricted` execution policy (use `penguin.cmd` or `Set-ExecutionPolicy -Scope CurrentUser RemoteSigned`); Ctrl-C to a session is a hard tree kill; only an x64 package (ARM64 runs via emulation).

## Verification

- **ci-windows: PASS** (all steps: install → build → typecheck → full `pnpm test` across all packages → PowerShell parse gate) — https://github.com/Prism-Shadow/penguin-harness/actions/runs/30220428080/job/89841822102 . The ubuntu `ci` job stayed green on every push. It took four iterations, each fixing a real Windows finding: (1) EBUSY on temp-dir cleanup after killing process trees → `fs.rm` retry options in core tests; (2) the tool-timeout test's premise (timeouts > 250ms can never fire for exec — it yields to background first) → kept-prior-output assertion made POSIX-only; (3) the `O_NOFOLLOW` sandbox gap → real server fix; (4) path-separator expectations in cli tests → `path.join`/skip.
- Linux: `pnpm build`, `pnpm typecheck`, `pnpm format:check` — pass; `pnpm test` — all green: skills 16, docs 11, landing 36, core 503 (+2 skipped), server 340, cli 186, web 383.
- New `shell-resolver.test.ts`: 11 pure-function tests (injected platform/env/prober; no real shells spawned).
- `pnpm penguin --version` through `run-with-env.mjs`: default, preset-override, and exit-code propagation verified.
- `install.ps1` + forwarder validated with real PowerShell 7.4.6 on Linux: parser clean; full functional runs against a local fake release server — fresh install, upgrade (data/ preserved, lib/ swapped, no `.old`/`.staging` leftovers), checksum-mismatch abort, 404 abort, and the complete `irm | iex` forwarder chain incl. `-Version` arg forwarding and shim generation.
- Release packaging block simulated locally: correct zip layout (`penguin/{bin,lib,web,node}`), `penguin.cmd` with CRLF, checksums.
- Web e2e (`bash e2e/run.sh`, full suite): 15 passed with the same environment-flaky failures as a pristine `origin/main` control build (paging/skills/draft-user-scope fail identically on main here) — no regressions from this branch.


hiyouga:
## Review

Broad but coherent, and the risky mechanical parts are right. Things I specifically verified rather than assumed:

- The `release.yml` heredocs are inside a `run: |` block scalar, so YAML strips the common 10-space indent and both `<<'EOF'` and the closing `EOF` land at column 0 in the actual script — the batch file is generated correctly, and the extra 2-space indent inside the `if` block survives harmlessly.
- `%~dp0` includes a trailing backslash, so `set "DIR=%~dp0.."` yields `...\bin\..`, and `%DIR%\node\node.exe` resolves. `Split-Path -Parent $PSScriptRoot` gives the install root for the `.ps1` shim. Both correct.
- The win-x64 Node zip really does have `node.exe` at the archive root under a `node-vX-win-x64/` top dir, which is what the `unzip` + `mv` handles.
- The win packaging step runs *after* the universal tarball is built, so `rm -f out/penguin/bin/penguin` cannot affect the other artifacts.
- `resolveShell`'s WSL guard: `Sysnative` also lives under `SystemRoot`, so the WOW64 path the comment mentions is caught by the same prefix test.

One thing needs a decision before merge, plus two notes.

### 1. Existing Agents will never see the `Shell:` line (existing-config compat)

`{{SHELL}}` is added to `DEFAULT_SYSTEM_PROMPT` in `default-config.ts`, and `assembleSystemPrompt` substitutes it. But the prompt is baked into `system_config.yaml` at Agent creation (`loadOrInitAgentState` writes `defaultSystemConfig()` once) and is never auto-upgraded afterwards — the only path that rewrites it is `resetSystemConfigToDefaults`, which is an explicit user action.

So the outcome splits:

- **New Agents on Windows**: prompt contains `- Shell: pwsh`, model writes PowerShell. Works.
- **Every existing Agent** — including a Linux/macOS user's Agents, and any Agent created on Windows before this ships: prompt has no `{{SHELL}}` placeholder, so the substitution is a no-op and the model is never told which shell `exec_command` speaks. On Windows it will keep emitting bash syntax into `powershell -Command` and the commands fail.

That is the "existing on-disk config" case in AGENTS.md, so per the repo rule this should not be decided inside the PR. The options as I see them:

1. **Do nothing, document it** — Windows is new, so arguably only Agents created *before* a user's first Windows run are affected, and those users are told to run the reset. Cheapest, but it silently degrades any Agent carried over.
2. **Targeted migration on load** — when the stored `system_prompt` contains `- OS Version: {{OS_VERSION}}` and no `{{SHELL}}`, splice the line in once and rewrite. Narrow and idempotent, but it is a real migration with the usual "how long do we keep it" cost.
3. **Fall back at assembly time** — if the assembled prompt contains no `{{SHELL}}` and the platform is win32, append the `Shell:` line to the Environment block. No disk write, no migration to retire; slightly magic.

I would lean 3 for the Windows case specifically, but that is your call — flagging it rather than picking.

### 2. Ctrl-C becomes a hard kill of the whole tree on Windows

`interrupt()` routes to `signalGroup("SIGINT")`, which on win32 becomes `taskkill /pid <pid> /t /f`. So `input_command`'s Ctrl-C, which on POSIX interrupts the foreground command and leaves the session alive, terminates the shell and every descendant on Windows. The code comment explains why (no shared console, no real signals on a piped child) and I agree there is no clean alternative without a native helper.

It is currently only documented in the source. Since this changes what a documented user-facing gesture does, it belongs in the Windows section of the installation/tools docs too — a user who Ctrl-Cs a long `npm install` expecting to get their session back will instead lose it.

### 3. `install.ps1` rewrites the user PATH as a literal string

```powershell
$UserPath = [Environment]::GetEnvironmentVariable("Path", "User")
...
[Environment]::SetEnvironmentVariable("Path", $NewPath, "User")
```

`GetEnvironmentVariable` returns the **expanded** value when the registry entry is `REG_EXPAND_SZ`, and `SetEnvironmentVariable` writes back `REG_SZ`. So any `%USERPROFILE%\...`-style entry the user had in their PATH is silently hard-coded to today's expansion — the classic installer footgun, and it is not reversible from the user's side because the original text is gone.

It is what a lot of installers do, so this may be an accepted cost; if you want to avoid it, read/write via `Microsoft.Win32.Registry` under `HKCU:\Environment` preserving the value kind, then broadcast `WM_SETTINGCHANGE`. At minimum the existing "added ... to your user Path" notice could say it rewrote the variable.

### Minor

- The generated `penguin.ps1` shim is left LF while `penguin.cmd` gets `sed -i 's/$/\r/'`. Harmless (PowerShell accepts both) but inconsistent with the `*.ps1 text eol=crlf` rule the same PR adds for tracked files.
- `.gitattributes` adding `* text=auto eol=lf` normalizes the working tree going forward but does not renormalize what is already committed; worth confirming `git add --renormalize .` is a no-op on `main` so a Windows contributor's first commit does not carry an unrelated line-ending diff.
- CI's PowerShell parse step passes `(Resolve-Path $f)` (a `PathInfo`) where `ParseFile` wants a string. It works via `ToString()`, but `.Path` would be explicit.

hiyouga:
Review follow-up: all four findings addressed in 33d668b (plus CI hardening commits 93d07e1, 3e75967, 76ab4a2).

**1. Existing agents never saw the `Shell:` line** — implemented as **option 3, the assembly-time fallback**, per the delegated decision. `assembleSystemPrompt` now injects `- Shell: <shell>` at render time when the platform is win32 and the agent's template carries no `{{SHELL}}` placeholder: right after the `# Environment` heading when the template has one (any heading level), else as a minimal final line. In-memory only — no disk write, no migration. POSIX output stays byte-identical (bash was always implied there), and the injection is deterministic and idempotent: a template that renders `{{SHELL}}` itself, or already carries any `- Shell:` line (including a deliberately hardcoded different value), is left untouched. Unit tests cover: pre-`{{SHELL}}` template + win32 → line present exactly once inside the Environment section; POSIX → byte-identical; template with `{{SHELL}}` → no duplicate; hardcoded line → no duplicate; no Environment section → appended final line. The fallback is commented with its **retirement condition**: removable once pre-`{{SHELL}}` agent configs (created before this PR shipped the placeholder) are no longer expected in the wild.

**2. Ctrl-C semantics in user docs** — now user-facing in both places, en + zh: a "Ctrl-C semantics" bullet in the Windows caveats of `installation.{en,zh}.md`, and a paragraph in the command-sessions section of `tools.{en,zh}.md` right after the `input_command` argument block. Both state that on Windows Ctrl-C terminates the whole command session tree (`taskkill /t /f`) instead of interrupting the foreground command, because no console signal can be delivered to a piped child.

**3. install.ps1 PATH clobbering** — user Path registration now goes through the registry preserving the value kind: `OpenSubKey("Environment", $true)`, `GetValueKind("Path")` (missing value → created as `REG_EXPAND_SZ`, the kind Windows itself uses for Path), `GetValue(..., DoNotExpandEnvironmentNames)`, membership check and append on the **raw** string, `SetValue(..., $kind)`. Membership is checked per entry after expansion, so a `%USERPROFILE%`-style spelling of the bin dir counts as already present and never gets a duplicate appended. No WM_SETTINGCHANGE broadcast; the restart-your-terminal notice stays and says the Path was appended. The block self-skips off-Windows (`$env:OS` guard) — the registry does not exist there, and this keeps the Linux functional harness runnable (the old API was a silent no-op on Linux). Neither the forwarder nor the docs described the old expand-and-rewrite behavior, so nothing else needed mirroring.

Verification for (3): the `pwsh -NoProfile` parse gate passes locally with real PowerShell 7.4.6, and the functional harness was reproduced against a local fake release server — fresh install, upgrade (`data\` preserved, `lib\` swapped, no `.old`/`.staging` leftovers), checksum-mismatch abort, 404 abort, and a StrictMode in-memory script-block run (the forwarder path). The registry read/write itself is Windows-only and cannot execute on a Linux box, so exactly that portion is covered by static parse + review plus isolated pwsh tests of the raw-string membership/append expressions (unexpanded `%USERPROFILE%` entry counts as present, append keeps raw entries verbatim, empty value, trailing semicolon, case-insensitivity).

**4. Minors** — (a) release.yml now CRLF-converts the `penguin.ps1` shim like the `.cmd`, matching the `*.ps1 eol=crlf` attribute (the installer-side fallback shims already come out CRLF on Windows via `Set-Content`). (b) `git add --renormalize .` against a pristine HEAD checkout of this branch is a **no-op** (empty staged diff, verified in a detached worktree). (c) the CI parse step passes an explicit string — `(Resolve-Path $f).Path` — to `ParseFile`.

Also, per the repo's working rules: added the missing `changelog/unreleased/` entry for the Windows-support surface, and the design specs on the companion branch (`penguin-harness-design@docs/2026-07-26-impl-sync`, commit 86c421b) now document the assembly-fallback exception (with its retirement condition) and the kind-preserving Path write.

CI: slow Windows runners flaked ci-windows three times, each time on pre-existing timing-sensitive tests untouched by this change set, hardened in the established pattern. (1) `state.test.ts`'s first-init test timed out at vitest's 5s default during a slow-I/O moment, and the cleanup rm then raced the still-running init into ENOTEMPTY → 93d07e1 gives that cold-start init test a 20s failure deadline and the cleanup fs.rm retry options. (2) `engine.test.ts`'s overlapping-execution test races a fast printf against `sleep 0.4`, and a cold Git-Bash spawn (1-2s on Windows) swamped the 400ms delta — the two completed 5ms apart in the wrong order → 3e75967 makes only the completion-order expectation POSIX-only (the overlap/non-blocking-approval assertion, file contents, and final-reply checks still run on Windows), same as the earlier kept-prior-output relaxation. (3) A third run then timed out two more first-in-file process-spawning tests that are sub-second normally — at that point clearly a degraded runner-pool day, not individual tests → 76ab4a2 adds a core `vitest.config.ts` (the same separate-config convention docs/landing/web already use) whose only non-default setting is a platform-aware `testTimeout`: 30s on win32, the 5s default kept on POSIX — a failure deadline only, passing tests are unaffected. ci-windows is green on the final sha: https://github.com/Prism-Shadow/penguin-harness/actions/runs/30257472014/job/89949300663

Nothing was declined. Local gates: `pnpm build` / `typecheck` / `test` / `format:check` all pass.

PR #82, opened 2026-07-27
fix(core,web): retry transport and quota LLM errors, disable dead sessions on auth failure

hiyouga:
## Problem

Two classes of LLM request failures observed in the field aborted the whole turn with `[已中断]: llm request error: …` even though nothing about the conversation was broken:

- `terminated: other side closed (UND_ERR_SOCKET)` — Node fetch wraps a dropped socket as `TypeError: terminated` with the real undici code on `cause`; the old classifier only looked at the top-level `code`, so it fell through to `failed`.
- `403 …订阅额度不足… (insufficient_user_quota)` — a gateway quota/subscription rejection; the blanket "4xx → non-retryable" rule discarded it before any quota signal was considered.

Both are transient: they must go through the **existing** in-run reconnect flow (resend input + `[turn_retried]`, tools never re-run), exactly like a network drop. Conversely, an authentication error can never be fixed inside a Session — its model + credentials are locked at Session creation — so retrying (or sending the next message) is pointless; the web chat needed to say so and point at a new Session.

## Classification (as implemented in `packages/core/src/llm/generative-model.ts`)

| Class | Signals | Outcome |
| --- | --- | --- |
| Retryable transport | status 429/408/5xx; a code in {`ECONNRESET`, `ECONNREFUSED`, `ETIMEDOUT`, `EPIPE`, `EAI_AGAIN`, `ENOTFOUND`, `ECONNABORTED`, `UND_ERR_SOCKET`, `UND_ERR_CONNECT_TIMEOUT`, `UND_ERR_HEADERS_TIMEOUT`, `UND_ERR_BODY_TIMEOUT`} on the error **or anywhere on its `cause` chain**; message keywords now also match `terminated` / `other side closed` / `fetch failed` | `timeout` → engine reconnects |
| Retryable quota | status 402/403 (or no status) **and** a code in {`insufficient_user_quota`, `insufficient_quota`} on the error, its cause chain, or the parsed provider body (`err.code`, `err.error.code`, `err.error.error.code` — OpenAI and Anthropic SDK shapes); **or** status 403 with message matching `/quota|subscription|额度/i`. Checked before the blanket 4xx rule | `timeout` → engine reconnects |
| Auth-dead | status 401 (own or wrapped); code/type in {`invalid_api_key`, `authentication_error`, `unauthorized`} (own, cause chain, or provider body); or SDK class name `AuthenticationError` | `failed` + `code: "auth"` → abort event carries `code: "auth"` |
| Plain failed | everything else, including a bare 403 with no quota/auth signal | `failed`, no code |

- Reconnect cap: `maxReconnects` default **2 → 5** (linear 250ms backoff unchanged; exhaustion message now reads "failed after 5 retries"). The compaction-request loop shares the cap by design.
- Web: an abort with `code: "auth"` on the main session sets a sticky `modelAuthDead` flag (rebuilt from Trace replay on reload; subagent-origin auth aborts only mark the nested model and never disable the parent composer). The composer disables/grays, the placeholder swaps, and a slim notice explains the Session is dead with a **New Session** button (`/chat/new`). The existing `[已中断]` abort line is unchanged. CLI behavior is unchanged.
- Protocol: `LLMOutcome.code?: "auth"` and `AbortPayload.code?: string` are **additive optional** fields — no compat shims, nothing reads legacy data differently, no server/DTO changes (abort events already reach clients via the stream and the Trace).
- Docs updated (en+zh): agent-loop (trigger set, cap 5, auth note, ASCII diagrams), interfaces (LLMOutcome.code), omni-message (StopReason table + AbortPayload.code), architecture, message-flow, web-app.

## Verification

All from the worktree root:

- `pnpm build` — pass
- `pnpm test` — pass: docs 11, skills 16, landing 36, core 505 (+2 skipped), server 340 (incl. `errors.test.ts` unchanged), cli 186, web 386
- Web e2e (`SKIP_BUILD=1 SRV_PORT=8975 MOCK_PORT=8976 bash packages/web/e2e/run.sh`, dedicated ports because 8930 is squatted in this environment):
  - **quota retry test** (new): mock answers the first 2 requests with `403 {"error":{"code":"insufficient_user_quota","message":"no active subscription"}}` then streams normally — asserts two amber retry lines with attempt numbers (`已发起第 1/2 次重试`), the turn completes, no abort, Trace shows two `request_end(timeout)` — **pass**
  - **auth dead test** (new): mock always answers `401 {"error":{"code":"invalid_api_key"}}` — asserts the abort line renders, the abort event in the Trace carries `code: "auth"`, the composer is disabled with the notice (still disabled after a reload via Trace replay), and the New Session CTA lands on `/chat/new` with a usable draft composer — **pass**
  - Full suite: **17 passed, 3 failed** — the 3 failures are exactly the pre-existing ones on clean main in this environment (`draft-user-scope` login-form fill timeout, `paging` sidebar count, `skills` install-modal button), unrelated to this change.


hiyouga:
现在出错了我只能看到超时，但是我想在成本中心看到具体背后的问题（详细信息）

hiyouga:
## Review

The shape of this is right: walking the `cause` chain instead of only the top level is the correct fix for `TypeError: terminated`, `anyInCauseChain` is cycle-safe, `providerSignals` handling both the OpenAI (`err.error.code`) and Anthropic (`err.error.error.code`) body shapes is the kind of thing that is usually discovered in production instead, and both new protocol fields are additive-optional with no compat shims. The e2e specs assert the observable outcome (retry lines with attempt numbers, `request_end(timeout)` in the Trace, the composer state surviving a reload) rather than the classifier's internals, which is the right level.

Three findings. The first two are, I think, blocking.

### 1. The auth-dead session is unrecoverable in practice, and the notice tells the user the wrong thing

Two separate things compound here.

**(a) The credentials are not fixed at Session creation.** That claim appears in three places — `LLMOutcome.code`, `AbortPayload.code`, and the user-facing string ("its model and credentials were fixed at creation"). Only the model *reference* is fixed. `agent.resumeSession` locks `provider` + `model_id` from the Trace's `session_meta`, but reads the credentials from the current project config every time:

```ts
const modelEntry = getModel(this.projectConfig, ref);
const apiKey = opts.apiKey ?? modelEntry.api_key;
const baseUrl = opts.baseUrl ?? modelEntry.base_url;
```

and the server never passes an override — `createCoreSessionLoader.load` does a fresh `createAgent({ root, projectId, agentId })` (so a fresh config read) and then `agent.resumeSession({ sessionId: row.sessionId })`.

**(b) But a loaded Session never goes back through that path, because updating a key does not invalidate anything.** `SessionManager.entries` holds the built `RuntimeSession` in memory. The rebuild-on-config-change machinery exists and works — `RuntimeEntry.generation` against `agentGenerations`, discarded on the next idle access — but `invalidateAgentRuntimes` has exactly one caller, `http/routes/vault.ts:54`. The route where an API key is actually edited does not call it:

```ts
app.put("/", async (c) => {
  const projectId = requireValidId(c, "projectId");
  deps.projectService.requireProjectOwner(c.var.user.userId, projectId);
  const req = parseModelsUpdate(await readJson(c));
  return c.json(await deps.projectConfigService.updateModels(projectId, req));   // no invalidation
});
```

So a cached entry keeps the old key until the 30-minute idle sweep (`ENTRY_IDLE_MS`), a server restart, or an unrelated vault update happens to bump the same Agent's generation.

That invalidation gap is pre-existing rather than introduced here, but this PR is what makes it user-visible, and together the two turn the feature into a trap. The sequence a real user hits: key expires → Session marked auth-dead → they update the key on the Models page → the cached runtime still holds the old key, so nothing changes → the notice tells them the Session is unrecoverable anyway → they abandon the conversation. Every step confirms a story that is not true. And the flag cannot clear on its own: it is sticky, rebuilt from Trace replay on every reload (the abort event lives in the Trace forever), it only clears on a later successful request, and no request can be made while the composer is disabled.

**The invalidation fix belongs in this PR.** It is the precondition for the rest of the feature being honest — "your credentials failed, fix them and carry on" is only true once a key update actually reaches a loaded Session — and it is small. `SessionManager` already iterates entries by Project elsewhere (`activeCountForAgent`), so no agent-list lookup and no async call is needed in the route:

```ts
/**
 * After a Project's models/credentials change: invalidate every cached runtime in this
 * Project, so the next Task re-resumes with the new api_key / base_url. Same effective-value
 * semantics as invalidateAgentRuntimes — no hot swap into a Task already in flight.
 */
invalidateProjectRuntimes(projectId: string): void {
  const agentIds = new Set<string>();
  for (const e of this.entries.values()) {
    if (e.projectId === projectId) agentIds.add(e.agentId);
  }
  for (const agentId of agentIds) this.invalidateAgentRuntimes(projectId, agentId);
}
```

Iterating entries is complete here, not a shortcut: the generation map only matters for runtimes that are already cached, and an Agent with no cached entry builds fresh through the loader anyway. The route then mirrors the vault one:

```ts
const res = await deps.projectConfigService.updateModels(projectId, req);
deps.manager.invalidateProjectRuntimes(projectId);
return c.json(res);
```

The rest of finding 1 then follows:

- Fix the copy in all three places: the model reference is fixed at creation, the credentials are not.
- Point the CTA at the action that actually fixes it — the Models page — and keep "New Session" as the secondary escape. As written, the primary action discards the conversation *and* leaves the key broken, so the next Session fails identically.
- Give the flag a way out. Clearing it on a Models-page visit works; so does simply letting the user dismiss the notice and try again.

### 2. An auth failure can be swallowed by the quota heuristic and retried instead of marked dead

In the catch chain, `isRetryableError` is consulted before the `failed` branch that computes `isAuthenticationError`:

```ts
} else if (isRetryableError(error)) {
  outcome = { status: "timeout" };
} else { /* only here is isAuthenticationError consulted */ }
```

and `isRetryableError` now calls `isQuotaExhaustedError`, whose fallback branch is a message keyword test with no check that an auth signal is absent:

```ts
return status === 403 && typeof err.message === "string" && /quota|subscription|额度/i.test(err.message);
```

So a 403 whose body carries a definitive auth code but whose message happens to mention a subscription or quota — e.g. `403 {"error":{"code":"invalid_api_key","message":"subscription key invalid"}}`, and SDK errors routinely put the body's message on `err.message` — classifies as retryable quota. It then burns every reconnect, and the exhaustion path emits its abort with no code at all:

```ts
yield* this.emitAbort(`${reason} failed after ${this.maxReconnects} retries`);
```

Net result on the one error class this PR exists to handle: a run of pointless requests against a dead credential, then a plain `failed` abort, and the Session is never marked auth-dead. This gets worse, not better, with the longer retry schedule in finding 3.

A bare 403 + `invalid_api_key` (no quota words) does classify correctly today, so the trigger is narrow — but the ordering is the real problem, and it matters more given that quota is *deliberately* retryable here: this ordering is the only thing keeping a dead credential out of the retry path. A definitive signal should not be overridable by a keyword heuristic:

```ts
export function isRetryableError(error: unknown): boolean {
  if (error == null) return false;
  if (isAuthenticationError(error)) return false; // definitive: never retry, never reclassify
  ...
```

Either way, `emitAbort` at the exhaustion site should forward a code once one is available, or the distinction is lost precisely when it matters most.

### 3. The single retry schedule is too short to be worth retrying quota at all

Quota being retryable is the product call, and one shared schedule for every retryable class (rather than a quota-specific one) is the design call — I am not relitigating either. But then that one schedule has to be long enough for the slowest class it serves, and right now it is sized only for the fastest.

`backoff` is `reconnectBackoffMs * attempt`, so 5 reconnects span **250 + 500 + 750 + 1000 + 1250 = 3.75 seconds** of wall clock. Whatever a quota retry is waiting out — a gateway's per-minute bucket, a top-up landing, a billing cycle rolling over — none of them close inside four seconds. As written the retry path spends 6 requests against the provider and gives up just short of the point where it could have helped, while the user watches retry lines that imply recovery is in progress.

A single exponential-with-ceiling ladder covers both classes without splitting them:

```ts
const ms = Math.min(this.reconnectBackoffMs * 2 ** (attempt - 1), this.reconnectBackoffMaxMs);
```

With the existing 250ms base, a 30s ceiling and ~8 attempts that is 250, 500, 1000, 2000, 4000, 8000, 16000, 30000 — about 62 seconds of total patience. The first two attempts stay exactly as fast as today, so a transport blip still recovers in under a second with no regression; the tail is where the quota case gets its window.

Two consequences worth deciding deliberately rather than discovering:

- A genuinely dead endpoint now takes ~1 minute to give up instead of ~4 seconds. It is visible the whole time (retry lines with attempt numbers) and the user can abort, so it is a slow failure rather than a hang — but it is a real change in how a hard outage feels.
- The compaction-request loop shares `maxReconnects`, so a failing compaction stretches to roughly the same minute. If that is not wanted, compaction is the one place worth giving its own cap.

### Minor

- `terminated` in the keyword regex is broad enough to catch unrelated provider copy (`"Request terminated by content filter"` would now retry). It is only a fallback — the real `UND_ERR_SOCKET` case is already caught by the cause-chain code check above it — so it could be dropped or tightened to `terminated` co-occurring with a transport hint.
- `pointer-events-none select-none` on the disabled composer card means a user whose long draft failed to send cannot select or copy it back out — which is exactly what they will want to paste elsewhere. `disabled` alone already blocks editing, so the textarea could stay selectable.

hiyouga:
*Revision note: the review above was updated after discussion with the author. Decisions taken so far, so the thread is not misread:*

- **Quota stays retryable**, and **one shared retry schedule** serves every retryable class — no quota-specific backoff. Both are the author's calls; my arguments against them are withdrawn. Finding 3 now only says that the single schedule has to be sized for the slowest class it serves, and proposes one exponential ladder that keeps the first attempts as fast as today.
- **The runtime-invalidation fix lands in this PR** rather than separately. It is folded into finding 1 with a concrete `invalidateProjectRuntimes` — `SessionManager` already iterates entries by Project, so it needs no agent-list lookup and no async call in the route.
- **Finding 1 also covers the runtime cache**, which the author was right to ask about: `PUT /models` never calls `invalidateAgentRuntimes` (only the vault route does), so a key update does not reach a loaded Session until the 30-minute idle sweep or a restart. That compounds with the incorrect "credentials are fixed at creation" claim rather than replacing it.
- **Finding 2 is unchanged** and gains weight twice over: with quota deliberately retryable, the auth-versus-quota ordering is the only thing keeping a dead credential out of the retry path — and a longer schedule means more wasted requests when it goes wrong.

hiyouga:
All findings from the revised review are addressed in 0d6eeb3, plus the owner's follow-ups: error details in the Cost center, key updates auto-unlocking auth-dead sessions, and — in 56dfd09 — the live retry countdown with retry-now / give-up controls.

### 1. Recoverable auth-dead sessions (finding 1 + the auto-unlock addendum)

- **`invalidateProjectRuntimes` landed in this PR**, exactly the reviewed shape: `SessionManager` collects agentIds from its own entries table for the Project and bumps each generation via the existing `invalidateAgentRuntimes` — sync, no agent-list lookup, complete because uncached Agents build fresh through the loader anyway. `PUT /api/projects/:projectId/models` calls it right after `updateModels`, mirroring the vault route's effective-value semantics (no hot swap into an in-flight Task; rebuilt on next idle access). Server tests: a unit test (both of a Project's Agents invalidated, other Projects untouched) and a route-level integration test mirroring the vault one (cached runtime rebuilt on the next Task after a key PUT; reads don't invalidate).
- **Key update → auto-unlock** (owner addendum). Two paths, both landed:
  - *Live*: the models PUT publishes a new `credentials_updated` ServerEvent to every **existing** channel of the Project's Sessions (`peek` — no channels are created for cold sessions); the web's stream-controller clears the auth-dead state on receipt, so open tabs unlock without a reload. Covered by a server test (event on the subscribed channel; cross-Project channels silent; no channel materialized for unsubscribed sessions).
  - *Reload*: the boolean became a timestamp. stream-model records `lastAuthAbortMs` from the abort event's envelope time; the models response now carries `updatedAt` (the config file's mtime — persistent, survives restarts, rewritten by every models update). The composer is dead only when `lastAuthAbortMs` is set AND (`updatedAt` unknown OR abort > update) — so after a key fix the replayed Trace's trailing auth abort no longer resurrects the dead state, while a wrong replacement key re-arms naturally (its own abort is newer than the update).
- **Auto-clear on success**: a completed main-session `request_end` clears the state — live and history replay share the code path, so a Trace with auth-abort followed by a completed request stays alive after reload. Subagent-origin auth aborts still only mark the nested model. Unit-tested in all four orders plus the time gate's four cases.
- **Corrected copy** in all three places (`LLMOutcome.code`, `AbortPayload.code`, web zh+en strings): the model *reference* is fixed at creation; credentials come from the current Project config — update the key on the Models page and the session unlocks by itself.
- **CTAs**: primary = "打开模型配置 / Open Models page" (navigates to `/models`), plus a "重试 / Retry" escape hatch that clears the state for another attempt (re-arms on the next auth abort — for credential changes the timestamps can't see), and the existing "New Session" as secondary.

### 2. Auth-first classification (finding 2)

`isRetryableError` now consults `isAuthenticationError` **first** — a definitive credential signal is never retried and never reclassified by the quota/message heuristics — and as a belt `isQuotaExhaustedError` itself refuses anything carrying an auth signal. The review's adversarial case is a test at every level: `403 + body code invalid_api_key + message "subscription key invalid"` → not retryable, not quota, is auth; the streamGenerate funnel yields `failed` + `code:"auth"`; the abort event carries the code. The retry-exhaustion `emitAbort` also forwards the last outcome's `code` when present (defense-in-depth — with the ordering fix auth can't reach that loop, but if it ever did, hosts would still see "auth" rather than a plain failure; unit-tested). Existing quota/transport classifications stay green.

### 3. One shared exponential ladder (finding 3, as revised)

The linear ladder is replaced by `min(reconnectBackoffMs × 2^(attempt−1), reconnectBackoffMaxMs)` with a new `reconnectBackoffMaxMs` dep (default 30s), and `maxReconnects` default rises 5 → 8. With the 250ms base the schedule is **250, 500, 1000, 2000, 4000, 8000, 16000, 30000 ≈ 62s of total patience** — the first two attempts are exactly as fast as today for transport blips; the tail is where quota-class errors get a real window. The user interrupt still cancels a backoff mid-wait (the abort listener fires inside the timer), which matters more now that a single wait can be 30s. `LLMOutcome.code` stays `"auth"`-only and there are no per-class schedules, per the revision note.

**Compaction cap decision**: compaction requests use their own `compactionMaxReconnects` (default 3, ~1.75s of backoff) instead of the shared cap. Rationale: compaction failure already has graceful semantics — the original context is kept and compaction retries on the next trigger — so failing fast beats stalling the session for a minute on housekeeping; the main turn loop keeps the full ladder. Both caps are unit-tested (including that the compaction loop ignores `maxReconnects`), the exhaustion message/tests say 8, and the exact ladder is pinned by a test on the exported `reconnectDelayMs`.

#### Live countdown + retry-now / give-up on the wait (owner addenda, 56dfd09)

A 30s wait behind a static "waiting" line reads as a hang, so the schedule is now visible and steerable:

- **Protocol**: `request_end` gains an additive optional `retry_in_ms` — present only when the engine will retry that failure in-run (timeout/malformed with attempts remaining under the applicable cap). It is computed by the **same** exported `reconnectDelayMs` the sleep uses, so the announced wait and the actual one cannot drift. Seam: the reconnect counter lives in the outer loop while the event is built inside the turn, so the attempt index is threaded into `runTurn` / `runCompactionRequest` (moving the emission out would have split the request event pair across layers — threading was the cleaner cut). Compaction request events are Trace-only (never streamed), so their `retry_in_ms` lands in the Trace record and no countdown renders for compaction. Absent on final failures (the abort follows) and on completed requests; old Traces replay unchanged.
- **Countdown**: the waiting retry line ticks every 250ms with ceil'd whole seconds (zh 「第 N 次重试，X 秒后发起…」 / en "retry #N in Xs…", extending the existing `S.chat.reconnect` family) when the announced wait is ≥2s — sub-2s waits keep today's rendering, so the fast head of the ladder doesn't flash numbers. The anchor is the CLIENT arrival time of the event, so server clock skew can't bend the ticker; at zero it stops and falls back to the plain waiting text. History safety: replay delivers the following `request_begin`/`abort` immediately, so a replayed `request_end` with `retry_in_ms` never leaves a ticking item (unit-tested in both directions).
- **Retry now (立即重试)**: new `POST /api/sessions/:sessionId/retry-now` skips the remaining backoff and fires the retry immediately — the attempt counter is unchanged. Under the hood the engine's backoff became wakeable: timer, user abort and skip race into a single `settle` (idempotent, no double-fire), surfaced as `skipReconnectWait()` through core Session → RuntimeSession → `SessionManager.retryNow`, mirroring the steer seam. The route answers a benign `200 {skipped:false}` when no wait is in progress, so timing races never surface as errors; same auth semantics as the other session routes (404 for outsiders).
- **Give up (放弃)**: no new machinery — it is the ordinary session abort; the engine's existing abort-during-backoff path ends the turn (abort line + carry-over) and the composer is immediately usable again. Both buttons render only on the live waiting state and only for main-session items (a subagent's backoff belongs to the child session, which the route does not target).

### Owner directive: real error details in the Cost center

`request_end` gains an additive optional `message` (the failure detail from `LLMOutcome.message`, non-completed statuses only; old Traces replay unchanged). Timeout outcomes now carry `describeError(error)` too — previously the detail was dropped exactly on the retry path. Both the turn loop and the compaction loop emit it. The server's stream-error-watcher stages that detail with the pending failure: when a retried failure flushes on the next `request_begin` (no abort ever arrives), the record keeps the real reason — the new server test asserts a retried quota failure lands in the errors panel with "insufficient_user_quota" in the message, kind/code unchanged (`llm_timeout`/expected). The abort reason still outranks the staged detail on failed exits, and the user-interrupt distrust rule now falls back to the staged detail before the generic status text.

### Minors

- `terminated` no longer matches the keyword fallback on its own — "Request terminated by content filter" must NOT retry (counter-test added); it counts only alongside transport vocabulary (socket/connection/closed). The real socket case is already covered by the `UND_ERR_SOCKET` cause-chain code; `other side closed` / `fetch failed` keywords kept.
- The disabled composer drops `pointer-events-none select-none` — `disabled` on the textarea already blocks editing, and the stuck draft stays selectable/copyable.

### Verification

- `pnpm build`, `pnpm typecheck`: clean.
- `pnpm test`: all packages green — core 513 (incl. the skip-wakes-the-backoff and announced-ladder tests), server 350 (incl. the new invalidation/event/updatedAt/watcher tests and the retry-now route tests), web 390 (incl. the new lifecycle/time-gate and countdown-state tests), cli 186, docs/skills/landing.
- `pnpm format` / `format:check`: clean.
- E2E (SRV_PORT=8975 / MOCK_PORT=8976): the quota spec now has the mock fail five times (engine deps aren't env-configurable, so the failure count — not knobs — opens the window): the 4s wait before retry #5 shows the live countdown, the displayed seconds visibly DECREASE between polls, clicking 立即重试 fires the retry well before the scheduled delay (asserted via a 1.5s expectation window against >2s of remaining wait) and the turn completes; the Trace carries `message` with the quota code AND the announced ladder `retry_in_ms: [250, 500, 1000, 2000, 4000]`. A new give-up scenario clicks 放弃 mid-wait: the abort line renders, the waiting notice flips to "stopped", and the composer is immediately usable. The auth spec walks the full recovery: dead state with the corrected copy → Models-page CTA navigates to `/models` → New Session still works → Retry without a fix re-arms the dead state → key update via the API unlocks the open tab **live** (no reload, no Retry click), the same conversation continues on the new key, and a reload after the success does not resurrect the state. Full web suite: 18/21 passed; the 3 failures (draft-user-scope / paging / skills) are the known pre-existing failures on clean main, unrelated to this change; `llm-errors.spec.mjs` passes 3/3 in isolation as well.
- Docs updated in sync (en+zh): agent-loop (ladder numbers + diagram, compaction cap, auth-first ordering, the countdown + retry-now/give-up line), omni-message (`request_end.message` + `retry_in_ms`, corrected `code` wording), web-app (Models-page CTA, retry, auto-unlock, the countdown controls), server-api (models PUT invalidation + `credentials_updated` event + `updatedAt` + the `/retry-now` route), interfaces/architecture/message-flow (numbers + `LLMOutcome` comments).



### Follow-up (auth as a stop reason, cap 5, terse copy) — 885d5f5

Review round 3 is in; where the sections above disagree, this supersedes them:

- **Auth is now a StopReason, not a field.** `StopReason` gains a sixth value `"auth"`; `LLMOutcome.code` and `AbortPayload.code` are **removed** (abortEvent is reason-only again). End-to-end flow: the classifier funnels credential failures to `{status: "auth", message}` → the engine stops exactly like `failed` (the auth/failed branch is the second belt keeping it out of the retry loop; the auth-first classifier is the first; the compaction loop folds `auth` → `failed`, keeping the event pair's existing value set) → the turn's `request_end` streams `status: "auth"` → the web records `lastAuthFailureMs` from that event (envelope timestamp) and gates the composer against the models `updatedAt` exactly as before. `credentials_updated` live unlock, auto-clear on a completed request, Retry/dismiss, and subagent isolation all stand unchanged. `finishInterrupted` stamps `"auth"` on partials it closes (consistent with the other terminal reasons); replay/CLI/web key on `completed` or `timeout|malformed`, so `"auth"` falls through like `failed` everywhere; the stream-error-watcher maps it into the existing `llm_failed`/unexpected bucket — no new taxonomy. Docs updated to the six-value enum (omni-message, interfaces, architecture).
- **Cap back to 5** (default). The exponential ladder stays: **250, 500, 1000, 2000, 4000ms ≈ 7.75s** of total patience; docs numbers/diagrams and test fixtures updated. The quota e2e keeps 5 mock failures, which now sits exactly at the cap (attempt 6 is the last allowed and succeeds) — the 4s countdown window before retry #5 is intact.
- **Terse copy.** The auth notice is action-only: zh 「模型 API 认证失败：请在模型配置页更新该模型的 API key，或新建会话。」 / en "Model API authentication failed: update this model's API key on the Models page, or start a new Session."; the placeholder shrank to 「模型认证失败，请先更新 API key」 / "Model authentication failed — update the API key first". The countdown and button strings added by this PR were already action-only.
- **Verification (rerun on 885d5f5)**: build/typecheck/format clean; all packages green — core 512 (funnel/adversarial tests re-pointed at `status: "auth"`; the now-moot exhaustion code-forwarding test removed), server 351 (+ the auth→llm_failed watcher case), web 390 (lifecycle tests re-keyed to `request_end(status "auth")`), cli 186. `llm-errors.spec.mjs` 3/3 (the auth spec now asserts the Trace's `request_end.status === "auth"` and that the abort carries no code); full web suite 18/21 with the same 3 known pre-existing failures (draft-user-scope / paging / skills).

packages/web/src/lib/strings.ts
```diff
@@ -626,20 +626,35 @@ Penguin 视觉风格（见 web-design 技能），深色/浅色主题（<html da
     subagent: "子会话",
     subagentRunning: "运行中",
     aborted: (reason?: string) => `[已中断]${reason ? `：${reason}` : ""}`,
+    /** Auth-dead notice (abort with code "auth"): only the model reference is fixed at creation — credentials come from the current Project config, so fixing the key on the Models page auto-unlocks this Session. */
+    modelAuthDead:
```
hiyouga: 这个太长了，只要告诉用户怎么做就行，不需要解释原因，其他也一样。

packages/core/src/engine/context-engine.ts
```diff
@@ -274,8 +307,10 @@ export class ContextEngine {
 
   constructor(private readonly deps: ContextEngineDeps) {
     this.maxTurns = deps.maxTurns ?? 100;
-    this.maxReconnects = deps.maxReconnects ?? 2;
+    this.maxReconnects = deps.maxReconnects ?? 8;
```
hiyouga: 默认是5吧？

packages/core/src/omnimessage/types.ts
```diff
@@ -253,6 +253,14 @@ export interface ApprovalDecisionPayload {
...
   type: "abort";
   reason?: string | null;
+  /**
+   * Machine-readable failure class; currently only `"auth"`: the run ended on a
+   * credentials/authentication error that no in-run retry can fix. Only the model
+   * REFERENCE is fixed at Session creation — credentials are read from the current
+   * Project config when the Session loads — so hosts should point the user at updating
+   * that model's API key (Models page), after which this Session can continue.
+   */
+  code?: string;
```
hiyouga: 这里不要加 code，不应该用这个 event 来处理 auth 的问题

packages/core/src/interfaces.ts
```diff
@@ -153,7 +153,22 @@ export interface GenerativeModelParameters {
...
   message?: string;
+  /**
+   * Machine-readable failure class; currently only `"auth"`: set when the failure is a
+   * credentials/authentication error that no in-run retry can fix — the request would keep
+   * going out with the same dead credential. Only the model REFERENCE is fixed at Session
+   * creation; credentials are read from the current Project config when the Session loads,
+   * so updating that model's API key (Models page) lets this Session continue. Carried
+   * through to the abort event so hosts can disable input until the credential changes.
+   */
+  code?: "auth";
```
hiyouga: 这个不需要 code

hiyouga (commented):
现在 auth 的这个处理太奇怪了，可以把它作为一个单独的 stop reason?但是不要加新字段了

packages/core/src/omnimessage/types.ts
hiyouga: Done in 885d5f5: the `code` field is gone from the abort event entirely (abortEvent is reason-only again). Auth is now its own StopReason — hosts read it from the request's own terminal status (`request_end.status === "auth"`), not from the abort.

packages/core/src/interfaces.ts
hiyouga: Done in 885d5f5: `LLMOutcome.code` is removed. Credentials failures now funnel to `{status: "auth", message}` — the sixth StopReason value, treated exactly like `failed` by the engine.

packages/core/src/engine/context-engine.ts
hiyouga: Reverted to 5 in 885d5f5. The exponential ladder stays: 250, 500, 1000, 2000, 4000ms ≈ 7.75s of total patience; docs and tests updated to match.

packages/web/src/lib/strings.ts
hiyouga: Trimmed to action-only in 885d5f5: 「模型 API 认证失败：请在模型配置页更新该模型的 API key，或新建会话。」 (en mirrors it); the placeholder shrank the same way, and the other strings this PR added (countdown, buttons) were already action-only.

PR #84, opened 2026-07-27
fix(core): reject empty compaction summaries and offer no tools to compaction requests

### LINKED ISSUE #83: 上下文压缩产生空摘要时仍被判定成功，导致任务状态丢失
## 问题

当上下文压缩请求只返回 thinking 或工具调用、没有返回文本时，系统仍会将压缩判定为成功，并生成空摘要：

```text
<context_summary>

</context_summary>
```

随后旧上下文被丢弃，Agent 无法恢复原任务，导致长流程意外停止。

## 预期行为

只有生成有效且非空的摘要时，压缩才能成功。

如果摘要为空或模型调用了工具，应重试或判定失败，并保留原上下文。同时，压缩请求应禁用工具调用。

目前相关测试明确接受空摘要，也需要同步调整。

hiyouga:
Fixes #83

## Problem

When the context-compaction LLM request returned only thinking or tool calls and no text, the engine still judged the compaction successful: it committed an empty `<context_summary></context_summary>` as the new context's first input and discarded the old context. The Agent lost its task state, and long-running flows stopped cold.

## Behavior before / after

Before:

- Any `completed` compaction response was accepted: `extractSummary` output was wrapped as the summary without validation, empty or not; the old context was discarded either way — a model answering the summarize prompt with a tool call produced an empty summary.
- Trace resume mirrored the bug: a file ending in `compaction_end(completed)` with a textless compaction output reconstructed an empty summary.

After:

- **Tools stay on the compaction request, unchanged** (owner constraint, see below): the request goes out with the session's frozen config — tool list included — so the request prefix stays byte-identical to ordinary turns and the provider's prompt cache remains valid at the moment the context is largest. A test pins that the compaction request carries the identical config object as an ordinary request (tool list verbatim, no `tool_choice`).
- A completed compaction response counts as success only with a **valid summary**: non-empty extracted text and no tool calls. Anything else is rejected.
- **Rejected tool calls are answered before retrying** (review's blocking finding): the rejected assistant turn is already committed on the live LLM object, so each of its tool calls gets a synthesized failed `tool_call_output` (`[tool error] the compaction request expects a summary, not tool calls`) — written to Trace so resume replays the same pairing, prepended to the retried input so the provider sees `tool_use`/`tool_result` paired, and held as carry-over if the compaction is abandoned with calls still unanswered. No `tool_use` ever dangles; ordinary turns after a failed compaction run cleanly. Empty-text rejections need no repair (the committed turn is plain text/thinking).
- **Rejections retry under a dedicated cap of 5 attempts** (`MAX_SUMMARY_REJECTIONS`); on the 5th rejection the compaction ends with `compaction_end(failed)`, keeping the original context and Trace index — the established failure semantics. Transport `timeout`/`malformed` attempts keep the shared `maxReconnects` cap (they are uncommitted; rejections are committed, well-formed turns).
- Trace resume mirrors the contract: a "completed" summarize closure whose output extracts to an empty summary (a shape only a pre-fix engine could write) is voided — the original context replays — and committed requests inside a compaction span don't count as Session turns.

## Owner constraint honored

Per the follow-up comment, the compaction request must **not** drop tools: omitting them changes the request prefix and breaks the provider prompt cache exactly when the context is largest, multiplying cost by tens of times. This supersedes issue #83's original "disable tools" suggestion (and the first commit of this branch, which omitted the field). `tool_choice` is likewise never used. Instead, tool-calling responses are rejected, repaired, and retried up to 5 times, then the compaction fails loudly with the original context intact.

## Tests

- Tools-identical pin: compaction request `UniConfig` is the same object as ordinary requests' (captured at the `openStream` seam); tool list verbatim, no `tool_choice`.
- Empty (thinking-only) responses: exactly 5 attempts (with `maxReconnects: 1`, proving the dedicated budget), then `failed`; no LLM swap, no Trace rotation, next run's input has no `[context_summary]`.
- Tool-calling rejections: every retry input leads with the repair answering the previous rejection's call (exact output text, `stop_reason: "failed"`), all five repairs land in the Trace, the calls are never approved/executed, and after exhaustion the next ordinary turn on the same engine leads with the final repair — no dangling `tool_use` — while a later run is plain (carry-over spent).
- Repair-then-success: rejection, repaired retry, compaction completes with the retry's summary.
- Cap boundary: four rejections then a valid summary on the 5th and final allowed attempt → success; only the adopted attempt's `token_usage` is surfaced.
- `stop_reason` filter: a non-completed (interruption-closure shape) `tool_call` neither rejects the summary nor gets a repair.
- The replay test that accepted an empty summary is inverted: a textless "completed" closure voids the closure and replays the original context.

## Verification

- `pnpm build` — success (all packages).
- `pnpm test` — all green: core 499 passed / 2 skipped (llm e2e, needs API keys), server 340, cli 186, web 383, landing 36, skills 16, docs 11.
- `pnpm format && pnpm format:check` — "All matched files use Prettier code style!"

### DIFF
--- a/packages/core/src/engine/context-engine.ts
+++ b/packages/core/src/engine/context-engine.ts
@@ -191,6 +191,19 @@ export interface ContextEngineDeps {
 /** Whether compaction is possible; when not `ok`, `compact()` is a no-op and yields no messages (see ContextEngine.compactability). */
 export type CompactAvailability = "ok" | "unsupported" | "empty" | "just_compacted";
 
+/**
+ * Maximum summarize attempts when the compaction response is rejected as an invalid summary
+ * (empty extracted text, or the model answered with tool calls — issue #83/#84). Deliberately
+ * separate from (and larger than) `compactionMaxReconnects`: that cap governs transport-level
+ * timeout/malformed attempts that were never committed and back off exponentially, while a
+ * rejection is a well-formed committed response — the request itself works, the model just
+ * didn't produce a summary, so the repaired input is resent immediately with no backoff. And
+ * since the compaction request keeps the session's toolset (the prefix cache must stay valid,
+ * see summarizeContext), a model insisting on tools deserves several chances.
+ * Beyond this many rejected attempts the compaction fails (original context kept).
+ */
+const MAX_SUMMARY_REJECTIONS = 5;
+
 /** Result of executing one LLM turn (the return value of runTurn). */
 interface TurnResult {
   /** All tool outputs for this turn, reordered to match the original tool_call order (for the next turn's LLM input). */
@@ -1007,13 +1020,22 @@ export class ContextEngine {
    * `summarize` compaction: appends the compaction Prompt to the **old** LLM object (first
    * folding in all of this turn's tool results when mid-Task, to keep tool_use/tool_result
    * pairing), then extracts the `[summary]` and wraps it as `[context_summary]` user text. The
+   * compaction request carries the session's toolset **unchanged** — the request prefix must
+   * stay byte-identical to ordinary turns so the provider's prompt cache remains valid;
+   * compaction runs exactly when the context is largest, where re-billing the whole
+   * transcript uncached costs tens of times more (issue #84 — this is why tools are *not*
+   * omitted and no `tool_choice` override is used). The
    * compaction request's streamed output is not pushed to the Human output stream (it emits
    * paired compaction events, plus the compaction request's `token_usage` — positioned between
    * the two events, so the frontend can count compaction cost into its stats), but it is written
-   * to the old Trace. timeout/malformed reconnect via the existing retry mechanism under the
-   * compaction-specific cap (`compactionMaxReconnects`, tighter than the turn loop's ladder),
-   * collapsing to failed once retries are exhausted; on failure/abort, the original context and
-   * Trace index are kept — it does not fall back to discard.
+   * to the old Trace. Compaction succeeds only with a **valid summary** — non-empty extracted
+   * text and no tool calls in the response. An invalid summary is rejected: any tool calls the
+   * model issued are answered with synthesized failed outputs (pairing repair, see the loop
+   * body) and the repaired input is resent immediately, up to MAX_SUMMARY_REJECTIONS attempts,
+   * then the compaction fails. timeout/malformed reconnect via the existing retry mechanism
+   * under the compaction-specific cap (`compactionMaxReconnects`, tighter than the turn loop's
+   * ladder), collapsing to failed once retries are exhausted; on failure/abort, the original
+   * context and Trace index are kept — it does not fall back to discard.
    * Docs: /docs/agent-loop § "Compaction".
    */
   private async *summarizeContext(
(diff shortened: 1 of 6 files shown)


hiyouga:
## Review

The diagnosis in #83 is right and most of the fix is well-judged. Specifically worth calling out as correct:

- Omitting `tools` from the request rather than sending `[]` or a `tool_choice` override, and asserting *absence on the wire* at the `openStream` seam rather than trusting the config object — that test is the one that would actually catch a regression here.
- Reusing `malformed` instead of inventing a retry path. An unusable response that was never adopted is exactly that class, and it inherits the existing backoff and the `compaction_end(failed)` → keep-original-context contract for free.
- `extractSummary` already ends in `.trim()`, so `summaryText !== ""` genuinely rejects whitespace-only and empty-tag output, not just the empty string. The PR body's claim holds.
- Inverting the replay test rather than deleting it, and voiding the closure on resume so a pre-fix Trace replays the original context instead of fabricating an empty summary.

One finding I think is blocking, because it turns a data-loss bug into a worse one on the tool-call path.

### A hallucinated tool call leaves an unanswered `tool_use` committed on the live LLM object

`runCompactionRequest` records `sawToolCall` but never dispatches the call, so no `tool_call_output` is ever produced. `summarizeContext` then rejects the response and, once retries are exhausted (or the retry itself fails), returns `{ status: "failed" }` — which by design does **not** call `startNewContext()`. `this.llm` is only replaced there:

```ts
private async startNewContext(): Promise<void> {
  this.llm = this.deps.createLLM!(this.lastSessionTokens);
```

and at the `run` call site, a `failed` compaction matches neither the `aborted` nor the `completed` branch, so the loop simply falls through and keeps going. Every subsequent ordinary turn goes back through the same `this.llm.streamGenerate` — the same AgentHub object whose committed history now ends in an assistant `tool_use` that will never be answered.

That is the exact state three existing safety nets in this file are written to prevent, all citing the same consequence:

- the `maxTurns` path holds pending tool outputs as carry-over rather than discarding them, because "the previous turn's assistant `tool_call` has already been committed by AgentHub, so discarding its paired output and sending a fresh message would be rejected by the provider as an unanswered `tool_use` (400, see issue #33)";
- a throwing `approve` callback collapses to deny rather than propagating, so the committed `tool_use` is not "left unanswered";
- `executeOne` collapses a contract-violating Environment exception into a failed output, "keeping `tool_use` and `tool_result` paired".

And `generative-model.ts` states the end result plainly: "subsequent requests would all be rejected by the provider as an unanswered `tool_use` (400), and the engine has no fix-up path left that touches LLM history."

The PR is half-aware of this — the comment notes the *retry* will be rejected and "collapses to failed below, which keeps the original context all the same". The part that is not accounted for is that the damage outlives the compaction. Compare the two behaviours on a tool-calling compaction response:

- **Before:** the response was accepted, `startNewContext()` ran, and the dangling `tool_use` was thrown away with the old LLM object. The bug was an empty summary and lost task state — but the Session kept working.
- **After:** the summary is correctly refused and the context is correctly preserved, but the poisoned LLM object is preserved along with it. The next ordinary turn hits the same 400, and so does the one after that. In-process there is no repair path; only a resume rebuilds a clean object (replay synthesizes the paired placeholders), so the practical recovery is "restart the server".

Trading lost context for a Session that cannot take another turn is not obviously the better end of the deal, and it is silent — the user sees compaction fail (which the design says is survivable) and then unrelated-looking request errors.

The fix fits the file's existing shape: before retrying or giving up, answer the hallucinated call. `executeOne` already builds exactly the right message:

```ts
const failed = toolCallOutput({
  output: `[tool error] ${message}`,
  toolCallId: toolCall.payload.tool_call_id,
  stopReason: "failed",
});
```

So `runCompactionRequest` would collect the completed `tool_call` messages (not just a boolean), and `summarizeContext` would synthesize one failed output per call — something like `[tool error] compaction requests offer no tools` — write them to the Trace, and prepend them to the resent `input`. Pairing is then intact, the retry has a real chance instead of a guaranteed 400, and even on final failure the LLM object is left well-formed for ordinary turns.

Note the empty-text case does **not** have this problem: that committed turn is plain assistant text/thinking, so resending the compaction prompt on top of it is structurally sound exactly as the comment says. Only `sawToolCall` needs the repair.

### Minor

- `sawToolCall` is set for any complete `tool_call` message regardless of `stop_reason`, while the ordinary turn loop filters with `if (tc.payload.stop_reason !== "completed") continue;` — a synthesized close-out `tool_call` carries a non-`completed` stop reason. Unreachable today (those paths do not end `completed`, so the check is never consulted), but matching the turn loop's filter would keep the two from drifting.
- A rejected attempt's `token_usage` is still fed to `observeTokenUsage` — so it updates `lastRequestTotal` / `lastSessionTokens` and counts toward Session cumulative cost — but `attempt.usage` is only yielded on success. That is defensible (the frontend's compaction line should show the attempt that produced the summary), but it does mean the displayed compaction cost understates what was actually spent whenever a retry happened. Worth a sentence in the doc comment so the asymmetry reads as chosen.
- `requestConfig` shallow-copies before `delete`, so the shared frozen `uniConfig` is untouched and the later `{ ...config, thinking_level }` preserves the deletion. Correct as written — just flagging that it depends on `tools` never being nested deeper, which is true today.

hiyouga:
compact 时候绝对不能不发送 tools，这样会破坏前缀缓存，导致成本数十倍增加，你必须保证 compact 时候的工具和之前一样。所以当出现工具就重试，如果超过 5 次就报错退出

hiyouga:
Both the review and the follow-up constraint are addressed in 84c590a (the PR body is updated to match the final state).

## Review — blocking finding: unanswered `tool_use` on the live LLM object

Implemented per the prescribed shape. `runCompactionRequest` now collects the response's complete `tool_call` **messages** instead of a boolean, and on rejection `summarizeContext` answers each call with a synthesized failed output in exactly the shape `executeOne` uses:

- output `[tool error] the compaction request expects a summary, not tool calls`, the call's `tool_call_id`, `stopReason: "failed"`;
- written to the (old) Trace at synthesis time, so resume replays the identical pairing — replay reclaims them as pending input paired against the committed call, matching the in-process state;
- prepended to the retried input, so the retry sees `tool_use`/`tool_result` paired instead of a guaranteed 400;
- and when the compaction is abandoned (failed on exhaustion, failed attempt, or abort) while the latest rejection's calls are still unanswered, held as carry-over — the next request on the same object leads with them, so even the final-attempt-hallucinates case leaves the object well-formed. The exhaustion test asserts the next ordinary turn's input leads with the last repair and that a subsequent run is plain (carry-over spent).

The retried input is rebuilt from the original base rather than accumulated: everything a committed rejection carried (previous repairs included) is already in history and is not resent. Empty-text rejections get no repair, as the review notes — that committed turn is plain text/thinking.

## Review — minors

1. **`stop_reason` filter**: adopted — only `stop_reason === "completed"` tool_calls are collected, matching the turn loop's filter, with a test pinning that an interruption-closure-shaped call neither rejects the summary nor produces a repair.
2. **Usage asymmetry**: documented as chosen at the yield site — rejected attempts still feed `observeTokenUsage` (Session cumulative cost and context tracking stay correct) while only the adopted attempt's `token_usage` is surfaced, so the displayed compaction cost understates true spend when retries happened.
3. **`requestConfig` shallow-copy + delete**: moot — the reversal below removes that code entirely; `requestConfig` is back to its original shape.

## Owner constraint: tools must stay on the compaction request

The first commit's tools-omission is reverted in full (`withoutTools` parameter, the per-request config deletion, and the wire-absence test are gone): omitting the tool list changes the request prefix and invalidates the provider prompt cache exactly when the context is largest, re-billing the whole transcript uncached. The compaction request now goes out with the session's frozen config, byte-identical to ordinary turns, and the absence test is replaced by its inverse — a pin (captured at the `openStream` seam) that the compaction request carries the **same config object** as an ordinary request, tool list verbatim, no `tool_choice`. This supersedes issue #83's original "disable tools" line.

With tools genuinely offered, a tool-calling response is a live possibility rather than a hallucination, which makes the review's pairing repair the centerpiece: invalid summaries (empty text or tool calls) are rejected, repaired, and retried under a dedicated cap of **5 attempts** (`MAX_SUMMARY_REJECTIONS` — deliberately separate from the shared `maxReconnects`, which still governs uncommitted transport `timeout`/`malformed` attempts). On the 5th rejection the compaction fails loudly via the existing `compaction_end(failed)` path with the original context and Trace index kept. Tests pin the counting on both sides of the boundary: 5 rejections → failed; a valid summary on attempt 5 → success.

hiyouga:
Rebased onto current main (`a955046`) after #79 and #82 landed; the branch is now a single commit (`fa924ea`) — the earlier add-then-revert churn around tools omission is squashed away.

Semantic merge with #82's compaction changes — the loop now runs **two independent retry budgets**:

- **Transport failures** (`timeout` / `malformed`, never committed by the stateful LLM): #82's `compactionMaxReconnects` cap (default 3) with the exponential `reconnectDelayMs` ladder, each failing `request_end` announcing the planned wait as `retry_in_ms` (Trace-only for compaction). Unchanged from main.
- **Summary rejections** (committed, well-formed responses that just aren't summaries — empty extracted text or tool calls): this PR's dedicated cap of 5 (`MAX_SUMMARY_REJECTIONS`), with the pairing repair prepended to the resent input. Rejections resend **immediately** — a rejection is model behavior, not a transport failure, so no backoff applies and no `retry_in_ms` is announced (the rejected attempt's `request_end` carries status `completed`, for which `plannedRetryDelayMs` yields nothing). The distinction is documented at both the constant and the loop branch.

`runCompactionRequest` combines both sides: main's attempt threading and `requestEnd(status, message, retry_in_ms)` emission plus this PR's `toolCalls` collection (filtered to `stop_reason === "completed"`). #82's `auth` outcome keeps folding into `failed` in the compaction loop, now with the repair stash applied on that exit too. `generative-model.ts` is byte-equal to main (verified `git diff origin/main` empty for it). Both sides' tests are kept — main's compaction-cap/`retry_in_ms` ladder test runs alongside this PR's rejection suite, with the rejection tests pinning `compactionMaxReconnects: 1` to prove rejections never consume the transport budget, and a new assertion that no rejected attempt's `request_end` ever announces a `retry_in_ms`.

Verification on the rebased branch: `pnpm install --frozen-lockfile` ✓; `pnpm build` ✓; `pnpm typecheck` clean across all packages; `pnpm test` all green — core 533 passed / 2 skipped (llm e2e), server 364, cli 186, web 400, landing 36, skills 16, docs 11; `pnpm format:check` clean.

PR #87, opened 2026-07-27
fix(core): committed compaction attempts absorb the turn's pending input

### LINKED ISSUE #85: Mid-task compaction retries can resend already-committed tool outputs
Follow-up flagged during #84 (see its final review reply). Bounded — both flavors collapse into the survivable failed/abort paths — but worth a deliberate design pass:

- When a **mid-task** compaction attempt is rejected (empty summary / tool-calling response) and retried, the resent `baseInput` still contains the turn's already-committed `tool_call_output`s; strict providers may 400 them as stale `tool_result`s. The post-failure loop's re-send of `turn.toolOutputs` has the same shape.
- A mid-task abort in that window can overwrite the carry-over stash.

Both affect only the mid-task compaction flavor; idle-triggered compaction is unaffected. A proper fix means teaching the run loop that a committed compaction attempt has absorbed the turn's outputs — a small design change to the carry-over contract rather than a patch inside `summarizeContext`.

Refs: #83, #84.

hiyouga:
Fixes #85

A mid-task summarize compaction folds the turn's tool outputs into the compaction request. Once any attempt is **committed** by the stateful LLM — and #84's rejected-but-completed attempts (empty or tool-calling responses) are committed — those outputs live in the old context's history. The engine nevertheless kept resending them, which strict providers reject as stale/duplicate `tool_result`s.

## Design: the two-case carry rule

The compaction retry has exactly two cases, and every seam now states them in one small branch:

- **AgentHub committed** (≥1 `completed` attempt): whatever the caller folded into the compaction request — the mid-task tool outputs, or the carry-over a manual `compact()` folds in — is consumed into the committed history. It is never resent: retries carry only the pairing repairs and the Prompt, the post-abandonment continuation drops the outputs, and `compact()` disposes of the folded carry. Only the repair stash (unanswered `tool_use` pairing from a final rejection) remains pending.
- **AgentHub did not commit** (every attempt timeout/malformed/failed/auth/aborted): the folded input never reached the model context. The carry is untouched — restored verbatim at the `compact()` seam (same message objects, pinned by an object-identity test), resent by the run loop exactly as before.

`CompactionResult` carries a single `committed: boolean` for this. (`timeout`/`malformed` end an incomplete stream and `failed`/`auth`/`aborted` throw or cut off before a clean end, so only `completed` commits; zero committed attempts also implies zero repairs.)

## Behavior per scenario

| Scenario | Before | After |
| --- | --- | --- |
| Mid-task rejection retry (attempt 1 committed) | retry resends `[turn outputs, prompt]` (+repairs) — stale `tool_result`s | retry sends `[repairs, prompt]` only; later retries `[prompt]` |
| Abandonment after a committed attempt (rejection cap / transport exhaustion / failed / auth) | continuation resends `turn.toolOutputs` — duplicates | continuation delivers the stashed repairs (+steering); with nothing left to send the run ends at the failure and the next prompt continues from committed state |
| Abandonment with **no** committed attempt (all timeout/malformed) | outputs resent | unchanged — outputs resent (never absorbed, still unanswered) |
| Abort in the compaction window after a committed attempt | case-A carry-over **overwrites** the repair stash and re-holds absorbed outputs | stash **merged**: repairs first; case-A carry-over only when nothing committed |
| Manual `compact()` abandoned, nothing committed | folded carry-over restored | unchanged in effect — restored **verbatim** (same objects) |
| Manual `compact()` abandoned after a committed attempt | folded carry-over silently restored — resent next run | carry consumed; only the repair stash remains pending |
| Idle/boundary compaction (no pending input) | — | unchanged (the committed branch is vacuous there; existing #84 suite green) |
| Boundary compaction failed with queued steering | repair stash waited until run end | stash rides the very next request, ahead of other content |

## Resume symmetry

Replay already reconstructs the committed-consumption state (#84 groundwork): committed compaction rounds pull the pending outputs into replayed history, and the eligibility filter keeps exactly the unanswered repairs as carry-over. Two replay tests pin the abandoned (nothing left to resend) and aborted (the repair alone replays as carry-over) trace shapes against the live engine's state.

## Tests

Six new engine tests (mid-task consumption + empty-handed abandonment ends the run; trailing tool-calling rejection continues with the repair alone; all-transport abandonment resends as before; abort-in-window merges the stash; manual-compact verbatim restore by object identity; manual-compact consumption with trailing repairs) and two replay pins. The existing #83/#84 suite passes unchanged.

## Verification

- `pnpm build` — success (all packages).
- `pnpm test` — all green: core 541 passed / 2 skipped (llm e2e, needs API keys), server 364, cli 186, web 406, landing 36, skills 16, docs 11.
- `pnpm format && pnpm format:check` — "All matched files use Prettier code style!"


packages/core/src/engine/context-engine.ts
```diff
@@ -1052,10 +1106,20 @@ export class ContextEngine {
     // executed and aren't recorded again, while carry-over's not-yet-written synthetic content
     // (flatten text, backfilled placeholders) and the compaction Prompt are written now.
     const prompt = userText(settings.prompt);
-    const baseInput = [...pendingToolOutputs, prompt];
-    let input = baseInput;
+    // The resend base: shrinks to the Prompt alone once an attempt commits — the folded turn
```
hiyouga, line 1154: 如果压缩没有提交消息到 agenthub，就直接还原之前的 carry 吧，而不是返回 absorb 的内容。如果提交了，就把 carry 处理掉

hiyouga: Implemented in 9114fff as the binary you describe. `CompactionResult` now carries the raw `committedAttempt` fact, and the seam branches on it directly: nothing committed to AgentHub → `this.pendingCarryOver = folded` — the prior carry restored verbatim (same message objects; with zero committed attempts there are also zero repairs, pinned by a test); something committed → the carry is consumed, never restored — only the repair stash for a final rejection's unanswered tool calls remains pending. The previous restore-behind-the-stash ordering is gone; `absorbedTurnInput` stays as the derived signal the in-run loop uses. Two tests pin both sides: an all-timeout manual compact leaves the carry byte-identical (asserted by object identity on the next run's input), and a committed-attempt abandonment leaves only the repair riding the next run.

hiyouga:
Simplified per the feedback in 95d9593: `CompactionResult` now carries a single `committed: boolean` — `absorbedTurnInput` is deleted. The old signal was `committed && folded-input-existed`, and every consumer's committed branch is vacuous when nothing was folded (dropping zero outputs, clearing an empty carry), so branching on `committed` alone is behaviorally identical; the untouched test suite proves it.

Each seam now reads as the two-case rule directly: the run loop drops the turn outputs from the continuation only when something committed (the dedicated absorbed-continuation block dissolved into the shared next-input assembly), the abort branch adds case-A carry-over only when nothing committed, and `compact()` restores the carry verbatim or consumes it. The per-return-site `absorbed()` stamping in `summarizeContext` is gone — it just tracks the boolean. Net −12 lines in the engine; repair-stash mechanics, both retry budgets, and resume symmetry unchanged.

hiyouga:
Rebased over #66 (goal mode) onto current main; the branch is squashed to a single commit (`be46ddf`).

Goal mode drives Tasks from outside the engine (`goal-loop.ts` calling `session.run` per round), so the run loop's structure — and with it every seam this PR hooks — is unchanged: the two-case carry rule still lives at the compaction checkpoint (`if (result.committed) turnOutputs = []`), the abort branch (`if (!result.committed)` case-A merge), the shared next-input assembly (stash drain + end-of-run guard), and the `compact()` verbatim-restore/consume binary. The one textual conflict was at the `compact()` seam, where #66 added its dead-goal downgrade to the carry consumption: resolved by applying `downgradeCarriedGoalInput` to the drained `folded` snapshot — a no-commit restore keeps the downgraded copies (the downgrade is idempotent and every consumer applies it anyway), while non-goal messages keep object identity, so the verbatim-restore identity test still passes.

Audited goal mode's paths for the carry rule: no new engine continuation paths exist to cover. Two interactions verified rather than changed — the repair stash rides `pendingCarryOver` through goal mode's downgrade map untouched (repairs are `tool_call_output`s; the downgrade only rewrites user text that parses as a goal round), and the "run ends when nothing remains to send" outcome after an absorbed abandonment composes cleanly with the goal loop, which simply opens the next round with a fresh `[goal]` prompt on the intact context.

Verification on the rebased branch: `pnpm install --frozen-lockfile` ✓; `pnpm build` ✓; `pnpm typecheck` clean; `pnpm test` all green including goal mode's new suites — core 567 passed / 2 skipped, server 398, cli 200, web 417, landing 36, skills 16, docs 11; `pnpm format:check` clean.

PR #100, opened 2026-07-27
fix(core,tooling): keep the harness's own ports out of the Agent's environment and off the dev server

hiyouga:
Two ways PenguinHarness's listen port collided with something it should not have.

## 1. `PORT` reached every command the Agent ran

`penguin web` writes `PORT` / `HOST` into its own `process.env` as the channel to the server module (the CLI's `startServer`), and the command Environment built child environments straight from `process.env`:

```ts
env: { ...process.env, ...this.vault, ...HARDENED_ENV },
```

So every `exec_command` inherited `PORT=7364`. `npm run dev`, Vite, Next and most Express templates read `PORT`, which means an Agent asked to start a dev server tried to bind **the harness's own port** rather than picking one.

The host side is now stripped of `PORT`, `HOST`, and the internal `PENGUIN_CLI_ENTRY` / `PENGUIN_WEB_DIST` before the spread. Keys are *deleted*, not blanked — a program that tests `PORT` for presence rather than value must see nothing.

Stripping is right even when the value came from the user's shell (`PORT=3000 penguin web`): it still means "the port PenguinHarness is on", which is exactly the port a spawned server must avoid. The vault is applied *after*, so a user who genuinely wants `PORT` in commands can set it there — covered by a test.

Deliberately **not** stripped: `PENGUIN_HOME` and the other user-facing `PENGUIN_*` settings. An Agent working on PenguinHarness itself (the self-evolution case) may legitimately want the same data root, and that is a config decision rather than a leak.

`packages/core/src/environment/tools/command/session-manager.ts` is the only place a child environment is assembled — `run-subagent` runs in-process, and `session.ts` just forwards what the manager built — so one edit covers `exec_command` and `input_command` both.

## 2. The dev backend shared 7364 with the installed one

`pnpm dev:server` bound the same port as an installed `penguin web`, and the two routinely run at once. Either the dev server failed to bind, or — worse and quieter — the Vite proxy talked to the **installed** server instead of the one being worked on, so changes appeared to do nothing.

The development backend moves to **7368**; 7365 / 7366 / 7367 are already the web / landing / docs dev servers. The full allocation table now lives in `core/internal/ports.ts`, which is where a reader looks even though the dev ports themselves have to be literals (vite configs and package.json scripts cannot import core TS).

`run-with-env` applies `${VAR:-value}` semantics, so `PORT=9999 pnpm dev` still wins, and the Vite proxy reads `PORT` with 7368 as the fallback — overriding the backend port moves the proxy with it instead of silently pointing at the old one. Verified both directions:

```
$ node scripts/run-with-env.mjs … PORT=7368 -- node -e "…"      → PORT=7368
$ PORT=9999 node scripts/run-with-env.mjs … PORT=7368 -- …      → PORT=9999
proxy target: default → 127.0.0.1:7368 ; PORT=9999 → 127.0.0.1:9999
```

The dev *data root* was split from the installed one for exactly this reason; this finishes the job for the port.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (core, server, cli, web)
- `pnpm test` — pass: core 572 / 2 skipped, server 398, cli 200, web 451, landing 36, skills 16, docs 11

The new tests in `exec-session.test.ts` are end-to-end through the real `Environment`, not unit tests of a helper: they set `PORT=7364` etc. on `process.env` (the exact state a serving process is in), run a real command, and assert the child saw nothing. A passthrough test on an unrelated variable is the control, so `PORT=[]` cannot pass vacuously. The env is read via `node -e` rather than `echo $PORT` because the resolver may pick bash or PowerShell depending on the machine.

## Note

The two halves are separable — say the word and I will split them. I kept them together because they are the same defect seen twice: the harness's own port escaping into a context that should not know about it.


hiyouga:
## Review

Reviewed adversarially in a clean worktree. The core claim — that `session-manager.ts` is the only place a child environment is assembled — I verified independently and it holds (details in the "fine" list). But the Vite half reintroduces the exact silent-wrong-backend failure this PR exists to fix, and several of the stated rationales do not hold.

### 1. `PORT ?? "7368"` proxies to port 80 when `PORT` is set-but-empty — blocking

`packages/web/vite.config.ts:24`

```ts
target: process.env.PENGUIN_API_PROXY ?? `http://127.0.0.1:${process.env.PORT ?? "7368"}`,
```

`??` only falls back on `undefined`/`null`. An exported-but-empty `PORT=` yields `"http://127.0.0.1:"` → **port 80**, while `pnpm dev:server` binds 7368. Every `/api` call silently goes to whatever is on :80, with no error — which is precisely the "the proxy silently talked to the installed server" failure mode the PR opens with.

Every other `PORT` reader in this repo treats empty as unset, deliberately:

- `packages/server/src/config.ts:82` — `Number(env.PORT || DEFAULT_SERVER_PORT)`, with a comment naming "the common `.env` case of an empty `PORT=`"
- `packages/cli/src/commands/serve.ts:35` — `raw === ""` → default, pinned by `server/test/config.test.ts:14-16`
- this PR's own `scripts/run-with-env.mjs:40` — `if (!env[name])`

Fix: `process.env.PORT || "7368"`.

### 2. The collision is relocated, not eliminated

`package.json:18` gives `pnpm penguin` `PORT=7368`, and `packages/server/package.json:28` gives `dev:server` the same 7368. `core/internal/ports.ts:20` documents them on one row as though that were intended.

The scenario is the self-development case this PR itself invokes: a harness started as `pnpm penguin web` (dev data root, now 7368) is asked to run `pnpm dev` in the repo → `dev:server` wants 7368 → EADDRINUSE, or the Vite proxy reaches the harness instead of the dev backend. Pre-PR both were 7364, so this pair is no better than before; only the installed-vs-dev pair was fixed. Giving `pnpm penguin` its own port (7369) closes it.

### 3. The stated justification for stripping `PORT`/`HOST` is false on the CLI-only path

`session-manager.ts:43-45` asserts the value "still means 'the port PenguinHarness is on'". `startServer` (`cli/src/commands/serve.ts:84-85`) is the only writer, and it runs only for `penguin server` / `penguin web`. For `penguin run`, `penguin chat` and the REPL nothing listens at all — yet `cli/src/index.ts:15` does `import "dotenv/config"`, so a project's `.env` carrying `PORT=3000` *is* in `process.env` for those runs and is now dropped from the Agent's commands.

The behaviour may still be right; the reason given for it isn't, and it is now a permanent comment.

### 4. `hostEnvForChild` is a dead export whose comment describes a test that was never written

`session-manager.ts:56-64` — "Exported for the unit test — the spawn path itself cannot be asserted without starting a real process." Nothing imports it, and the tests that *were* written do exactly what the comment calls impossible: they assert the spawn path by starting a real process. The `source` parameter is never called with a non-default argument either. Drop the export and the comment, or add the unit test.

### 5. The comment's own rule contradicts the strip list

`session-manager.ts:47-52` calls `PENGUIN_WEB_DIST` "internal plumbing", then says "deliberately **not** stripped: `PENGUIN_HOME` and the other user-facing `PENGUIN_*` settings". But `PENGUIN_WEB_DIST` is a documented user-facing deployment variable (`docs/content/configuration.en.md:17`, `server/README.md:19`, read at `server/src/config.ts:85`) and *is* stripped — while `PENGUIN_WEB_DB`, equally documented and also a filesystem path, is not. The added docs paragraph gets this right; only the code comment is wrong.

### 6. The new docs sentence is false

`docs/content/configuration.{en,zh}.md:23` — "Everything else in the host environment passes through unchanged." `HARDENED_ENV` (`session-manager.ts:26-33`) overrides `GIT_EDITOR`, `GIT_TERMINAL_PROMPT`, `TERM`, `NO_COLOR`, `PAGER`, `GIT_PAGER` on every child, and applies *after* the vault, so those six cannot be set at all. This PR is the first place the docs make a passthrough claim, so it is the right place to state the exception.

### 7. PLAUSIBLE (Windows) — the strip is case-sensitive, Windows env names are not

`session-manager.ts:60-64` does `delete env[key]` on a plain-object copy. On Windows the copy preserves the OS's stored casing, so `set Port=3000` before `penguin web` survives `delete env["PORT"]` and the child — resolving case-insensitively — still sees it. Not verifiable on Linux. The new tests can't catch it either: they create `process.env.PORT` themselves in uppercase, so Windows CI passes regardless.

### Minor

- `default-config.ts:98` still tells the model to avoid "a PenguinHarness service port" while this PR removes the only in-band way to learn it. Not a regression (the guardrail was already numberless) but now backed by no readable data.
- `.env`'s `PORT` no longer takes effect for `pnpm dev:server` / `pnpm penguin`, which contradicts the documented priority at `cli/src/commands/serve.ts:10-11`. `PENGUIN_HOME` set the precedent and `.env.example` has no `PORT`, so the blast radius is small.
- `ports.ts:20` calls `pnpm penguin` one of "those packages' `dev` scripts" — it is a root script.
- The PR body's test counts don't match a clean run of this branch (core 569 passed / 2 skipped, landing 39). Everything passes, so the conclusion stands, but the numbers look written rather than copied.

### Checked and found fine

- **"Only place a child env is assembled" — verified.** Exactly one `new Environment` (`core/src/agent.ts:691`), one `new CommandSessionManager` (`environment.ts:96`), one `new ManagedSession` (`session-manager.ts:84`); `session.ts:79-81` forwards `opts.env` verbatim. Subagents go through the same Agent/Environment path in-process. MCP has no adapter layer yet (config only, no spawn). Background processes reuse the same manager. The other spawns in the repo are host-side, not Agent-side — including `version.ts:73`, which legitimately needs `PENGUIN_CLI_ENTRY`.
- **"The vault can put it back" — verified end-to-end.** `PORT` satisfies `isValidVaultKey` and the spread order puts vault after the host env. Worth stating in the docs though: the vault is per-Agent and its key names are disclosed to the model.
- **`run-with-env.mjs` semantics — verified by running it**: unset → 7368, `PORT=9999` → 9999, `PORT=` → 7368, `PORT=0` → `0` preserved.
- **The new tests are meaningful.** Reverting to `{...process.env, ...}` fails the first with `PORT=[7364], HOST=[127.0.0.1], …`. The passthrough test is a real control. The vault test passes with or without the fix — it documents rather than guards.
- **Every remaining `7364` is correct** (installed/production default), and the two dev-facing docs were updated. CI and e2e have no port assumptions — `web/e2e/run.sh` uses its own 8930/8931 with explicit `PORT=`.
- **Dev preview URLs follow the port automatically** (`preview-token.ts:199-204` derives from the actual bind port), so this half also fixes previews under `pnpm dev` alongside an installed server.
- **Unclaimed side benefit:** `skills/penguin-sdk/SKILL.md:311` does `.listen(Number(process.env.PORT ?? 4630))` while logging `localhost:4630` — pre-PR that template silently bound the harness's port. The strip fixes it.

### Suggested order

Fix 1 first — it is a one-character change and it defeats the PR's own purpose. Then 2, since the self-development collision is the case the PR argues from. 3–6 are comment/doc corrections, but they will otherwise sit in the tree as permanent wrong rationales.

PR #103, opened 2026-07-28
feat(server,web): page the cost center's error table, abbreviate the source, drop command-tool noise

hiyouga:
Three changes to the cost center's error panel.

## Paging

The table showed the newest 20 with no way back, so an error from an hour ago was simply gone.

New `GET /api/projects/:projectId/usage/errors?offset&limit&from&to&agentId` → `{items, total}`, newest first. Page 0 is the batch the dashboard response already carries, so stepping back is the only thing that costs a request. `total` is the filtered row count, so the pager knows where the end is.

It takes the dashboard's **date/agent filter only**. The model filter is deliberately not accepted: it never applied to errors in the first place (HTTP and process errors have no Model dimension), so honoring it here would imply a narrowing the summary above the table does not do. Admin-only visibility of unattributed errors carries over unchanged — a regular member paging back must not start seeing another tenant's login failures.

The pager sits outside the scroll box so it stays reachable without scrolling the rows, and appears only when there is more than one page. A failed page fetch shows its message and keeps the rows already on screen rather than blanking the table.

## `[env] tool_failed:exec_command`

The source is bracketed and abbreviated instead of `environment · tool_failed:exec_command`. Only `environment` needs shortening — every other source (`http`, `llm`, `session`, `usage`, `title`, `subagent`, `process`, `schedule`) already reads at a glance, so the map has one entry rather than a full abbreviation scheme. The "most common error code" stat uses the same shape, so the two places agree.

## Command-tool failures are no longer errors

`resultForExit` in core maps **any** non-zero exit to `failed`:

```ts
if (exit.code !== 0) return { stopReason: "failed", note: `[exit code: ${exit.code ?? "unknown"}]` };
```

But a non-zero exit is how shell commands return information, not a fault — `grep` exits 1 when nothing matches, `test -f` when the file is absent, `diff` when files differ. Recording those made the error table largely a log of ordinary Agent work, which buried real errors and consumed both the 20k row cap and the dedup window that exist to protect them. The Agent already sees the failure and adjusts; nothing there needs a human.

**Both command tools are excluded.** `input_command` ends in the *same* `resultForExit`, and it is how a backgrounded `exec_command` is polled for its eventual exit — so excluding only one would record the same command's non-zero exit when it finished in the background and drop it when it finished in the foreground. Its remaining failure paths (a missing or unknown `process_id`, Ctrl-C mixed with other content) are the model misusing the tool, which it likewise sees in the output and corrects on its own.

Excluded at the capture site rather than filtered at query time, so the noise never reaches the table, the cap, or the dedup key.

## Verification

- `pnpm test` — pass: server 402, core 569 / 2 skipped, cli 200, web 451, landing 39, skills 16, docs 11

New tests: `queryErrors` pages newest-first across a boundary with no overlap, past-the-end is empty rather than an error, and the filter is applied so a page cannot widen the summary; plus one pinning that both command tools' failures and timeouts are dropped while every other tool still records.

Existing tests used `exec_command` as their stand-in for "some tool that failed", so they move to `write_file` / `read_file`. Two of them pair a parent and a child session on the *same* `tool_call_id` — the point being that the per-origin name cache must not let one overwrite the other — so those keep two **distinct** tool names, which also keeps short-window dedup from suppressing the second record.

### DIFF
--- a/packages/docs/content/server-api.en.md
+++ b/packages/docs/content/server-api.en.md
@@ -141,6 +141,7 @@ On Session creation, `modelId` and `provider` are both-or-neither: send the comp
 | Method | Path | Description |
 | --- | --- | --- |
 | GET | /usage | Usage statistics; query parameters `from`, `to`, `groupBy`, `agentId`, `provider`, `modelId` |
+| GET | /usage/errors | One page of the error detail table (newest first): `offset`, `limit`, plus the same `from` / `to` / `agentId` filter → `{items, total}` |
 | GET | /agents/:agentId/traces | Date → Session drill-down structure of Trace files |
 | GET | /agents/:agentId/traces/:sessionId/:index | Read Trace events (`offset` / `limit` pagination) |
 | GET | /agents/:agentId/traces/:sessionId/:index/analysis | Trace performance analysis |
--- a/packages/docs/content/server-api.zh.md
+++ b/packages/docs/content/server-api.zh.md
@@ -141,6 +141,7 @@ Schedule 写操作仅限 Owner。新建 Session 模式的任务，`modelId` 与 
 | 方法 | 路径 | 说明 |
 | --- | --- | --- |
 | GET | /usage | 用量统计，查询参数 `from`、`to`、`groupBy`、`agentId`、`provider`、`modelId` |
+| GET | /usage/errors | 异常明细表分页（按时间倒序）：`offset`、`limit`，以及与看板一致的 `from` / `to` / `agentId` 过滤 → `{items, total}` |
 | GET | /agents/:agentId/traces | Trace 文件的日期 → Session 下钻结构 |
 | GET | /agents/:agentId/traces/:sessionId/:index | 读取 Trace 事件（`offset` / `limit` 分页） |
 | GET | /agents/:agentId/traces/:sessionId/:index/analysis | Trace 性能分析结果 |
--- a/packages/server/src/api/types.ts
+++ b/packages/server/src/api/types.ts
@@ -1085,8 +1085,20 @@ export interface UsageErrors {
   unexpected: number;
   /** The most frequent source · code (null when there are no errors). */
   topCode: UsageErrorCount | null;
-  /** Most recent N items (reverse chronological). */
+  /** Most recent N items (reverse chronological) — the first page; older ones come from `GET /usage/errors`. */
   recent: UsageErrorItem[];
+}
+
+/**
+ * GET /api/projects/:projectId/usage/errors — one page of the error detail table, newest
+ * first. The dashboard response above already carries the first page; this exists so
+ * "show me earlier ones" does not have to refetch the whole aggregate. It takes the same
+ * date/agent filter as the dashboard, so a page never widens what the summary counted.
+ */
+export interface UsageErrorsPage {
+  items: UsageErrorItem[];
+  /** Filtered row count, so the caller knows when it has reached the end. */
+  total: number;
 }
 
 export interface UsageResponse {
--- a/packages/server/src/db/repos/errors.ts
+++ b/packages/server/src/db/repos/errors.ts
@@ -194,15 +194,15 @@ export class ErrorsRepo {
   }
 
   /** The most recent `limit` entries (reverse chronological order). */
-  recent(projectId: string, f: ErrorFilter = {}, limit = 20): ErrorItem[] {
+  recent(projectId: string, f: ErrorFilter = {}, limit = 20, offset = 0): ErrorItem[] {
     const { where, params } = this.conds(projectId, f);
     const rows = this.db
       .prepare(
         `SELECT ts, source, code, kind, message
          FROM error_records WHERE ${where}
-         ORDER BY id DESC LIMIT :limit`,
+         ORDER BY id DESC LIMIT :limit OFFSET :offset`,
       )
-      .all({ ...params, limit });
+      .all({ ...params, limit, offset });
     return rows.map((r) => ({
       ts: r.ts as string,
       source: r.source as string,
--- a/packages/server/src/http/routes/usage.ts
+++ b/packages/server/src/http/routes/usage.ts
@@ -1,12 +1,14 @@
 /**
  * Usage statistics routes:
  * GET /api/projects/:p/usage?from&to&groupBy&agentId&provider&modelId
- * (model filter is paired: provider and modelId are given together).
+ * (model filter is paired: provider and modelId are given together);
+ * GET /api/projects/:p/usage/errors?offset&limit&from&to&agentId — one page of the error
+ * detail table, for paging back past the first page the dashboard already returns.
  */
 import { Hono } from "hono";
 import type { UsageGroupBy } from "../../api/types.js";
 import type { AppEnv } from "../../auth/middleware.js";
-import { badRequest, optionalDateParam, requireValidId } from "../validate.js";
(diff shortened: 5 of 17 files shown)


hiyouga:
## Review

Reviewed adversarially in a clean worktree; the stated verification reproduces (build / typecheck / format / full suite green). The endpoint and the SQL are sound. The pager has a real bug in its most common interaction, and two of the justifications in the code comments are factually wrong.

### 1. "Newer" back to page 0 leaves the previous page's rows on screen — blocking

`errors-panel.tsx:111-112`

```ts
useEffect(() => {
  if (page === 0) return;   // ← returns before restoring items
```

Page 0 is deliberately never fetched — but it is also never *restored*. The only other writer of `items` is the `[recent]` effect at `105-109`, and `recent` is `data.errors.recent`, a stable reference for the panel's whole mounted lifetime.

Repro: open the cost center with >20 errors → **Older** (page 2 loads correctly) → **Newer**. `page` becomes 0, the label flips to "Page 1 / N", the Newer button greys out — and the table still shows rows 21–40. The only escape is changing a filter, which remounts the panel. A stale `pageError` survives the same way, since `setPageError(null)` is also below the early return.

One line:

```ts
if (page === 0) { setItems(recent); setPageError(null); return; }
```

Worth saying plainly: `packages/web/vitest.config.ts` runs in the node environment with no DOM tests, so nothing in the suite could have caught this. The green run in the PR body is not coverage of the pager.

### 2. The exclusion drops genuine environment faults the rationale does not cover

`stream-error-watcher.ts:109-126`

The PR quotes `resultForExit`'s non-zero-exit line but not the one immediately above it (`command/session.ts:266`):

```ts
if (exit.signal) return { stopReason: "failed", note: `[terminated by signal ${exit.signal}]` };
```

`failed` from the command tools is not only "non-zero exit". It also covers:

- **signal termination** (`session.ts:266`) — an OOM-killed build, a segfaulting test binary, SIGBUS;
- **spawn failure** (`exec-command.ts:88, 114`, `input-command.ts:104`) — nonexistent `workdir`, EMFILE/ENOMEM, an unresolvable shell;
- **no command session manager configured** (`exec-command.ts:59`, `input-command.ts:51`) — a server misconfiguration.

Those are exactly the recorder's own documented *"a config/environment fault, needs a human"* category (`error-recorder.ts:14-16`), and an Agent cannot "see it and adjust" its way out of any of them. The permanent comment at `stream-error-watcher.ts:120-125` — that `input_command`'s remaining failure paths are all "the model misusing the tool" — is incomplete as written.

A narrower rule was available and is roughly the same size: `toolFailureText(p.output ?? "")` already carries the note, so the watcher can distinguish `[exit code: N]` from `[terminated by signal …]` / `[spawn error: …]` and drop only the first.

### 3. The dedup half of the justification is false

`stream-error-watcher.ts:114-115` and the PR body claim the noise "consumed both the 20k row cap and the dedup window that exist to protect [real errors]".

The dedup key is `source \0 code \0 projectId` (`error-recorder.ts:122`), and the code is `tool_failed:exec_command`. So exec_command failures could only ever suppress *other exec_command failures* — never another error's window. The row-cap half is true; the dedup half is not, and it is now a permanent comment in the file.

The same 2s dedup window also caps the pre-change noise at roughly one record per 2s per project, which softens "made the error table largely a log of ordinary Agent work" — real, but bounded.

### 4. An empty page > 0 strands the user with no controls

`errors-panel.tsx:133` (pageCount from the dashboard's `total`), `:166` (empty state gated on `page === 0`), `:222` (pager gated on `items.length > 0`).

The component discards `res.total` and pins `pageCount` to the dashboard snapshot. If a page > 0 comes back empty — rows pruned by the 20k cap, or a project/agent deletion between load and click — the table renders headers with no rows, *and* the pager is hidden, so there is no way back. Two lines: use `res.total`, and drop `items.length > 0` from the pager condition.

### 5. The new endpoint has no HTTP-level tests, including its headline security claim

The code reads correct — same `requireProjectAccess`, same `includeGlobalErrors: c.var.user.isAdmin`, and the non-admin predicate is `project_id = :pid` (`errors.ts:139`). I found no cross-tenant path. But the harness for exactly these assertions already exists and was not extended: `authz.test.ts:47` (outsider → 404) and `id-validation.test.ts:99` (traversal) both cover `/usage` and not `/usage/errors`, and `errors.test.ts` has the admin-only assertion for the dashboard but not the paged route.

The PR asserts "a member paging back must not see another tenant's errors" three times in prose and zero times in a test.

### 6. Two of the new tests are weaker than they look

- `errors.test.ts:539, 546` feed `stopReason: "timeout"` for the command tools. Environment never emits that on a `tool_call_output` — a tool timeout is finalized as `failed` plus a `[tool timeout: …]` note (`environment.ts:315-317`, asserted in `core/test/environment.test.ts:520`). Those two cases pin an unreachable state.
- The third paging test ("applies the dashboard's filter") passes without offset support at all — it only exercises `agentId` at offset 0. Reverting `recent()`'s offset fails the other two, so the core of the feature is genuinely covered; this one adds nothing.

### 7. Minor

- **Silent constant coupling**: web `PAGE_SIZE = 20` (`errors-panel.tsx:45`) must equal server `ERROR_RECENT_N = 20` (`usage-service.ts:33`). Neither references the other; changing one silently skips or repeats rows with no failing test.
- **Offset paging over a head-appended table**: errors arriving between load and an "Older" click shift every row down, so page 1 re-shows rows already seen. Standard tradeoff, but undiscussed — keying off the newest row's id avoids it.
- **Doc rot this PR introduces**: it makes `tool_failed:exec_command` an impossible code and, in the same diff, installs it as the canonical example at `errors-panel.tsx:35` (also stale at `:176` and `stream-error-watcher.ts:42`).
- **Not airtight**: `NOT_RECORDED_TOOLS.has(name ?? "")` — on a tool-name cache miss an exec_command failure still records as `tool_failed:unknown`.
- **`filters` identity** (`usage-page.tsx:297`, `errors-panel.tsx:117/131`): benign today, because the three deps are the entire object. But the PR text weighs adding the model filter; a fourth key would be read from a stale closure with no dep to catch it, and the repo has no eslint to flag it. Depend on `filters` and memoize at the call site.

### Checked and found fine

- **No infinite refetch loop and no stale closure today.** `recent` is stable (`data` is set once per load), and `setExpanded` on `[items]` settles after one extra render. Note the `[recent]` reset effect is effectively dead: `{data && <ChartCard>}` unmounts on every filter change, so the panel remounts with fresh state — `page` does reset, just not by the mechanism its comment claims.
- **`total` and `items` cannot disagree.** Both derive from the same `ErrorFilter` through the same `conds()` in one synchronous request. Off-by-one checks out; the model filter is correctly excluded on both sides.
- **Input validation.** `paginationQuery` bounds limit 1–1000 and offset ≥ 0; deep offsets are bounded by the 20k cap; `optionalDateParam` rejects impossible dates. Route mounting collides with nothing.
- **Nothing else consumed command-tool error records** — `ErrorsRepo` has one reader, and the CLI has no error panel. The `session-manager.test.ts` edits are stand-in renames, not behaviour changes.
- **The paging tests are meaningful** where it counts: binding `offset` to 0 fails two of three; neutralising the exclusion guard fails the command-tool test.
- **i18n parity** across `strings.ts` / `strings-en.ts`.

### Suggested order

Fix 1 and 4 (both one- or two-liners in the same component) before merge; they are user-visible. Then either narrow the exclusion per 2 or correct the two comments so the file does not carry a wrong rationale permanently. 5 is cheap and worth doing given how load-bearing the claim is.

PR #104, opened 2026-07-28
fix(core): retry on failed as well as timeout; only auth stops the run

hiyouga:
## The report

```
llm request error: Upstream HTTP/2 stream failed (upstream_http2_stream_error)
```

Plainly a transport fault. It aborted the turn.

Tracing it: no `status`, so the status rules do not apply; no `UND_ERR_*` on the cause chain; and the message matches none of `timeout|network|socket hang up|econnreset|connection reset|other side closed|fetch failed|rate limit|…`. It fell straight through `isRetryableError` to `failed`, and `failed` ended the run.

## The fix is in the policy, not the classifier

`isRetryableError` is an allowlist — known network codes, known status classes, known message vocabulary. An allowlist can only ever be as complete as the failures already seen, so every gateway that phrases a transient failure its own way is a new lost turn and a new patch.

The tempting fix is to make the classifier retry everything but auth. **That is what this PR did first, and it was wrong**: it collapsed every failure into the `timeout` status, which is not just a retry decision — it is what gets written to `request_end`, into the Trace, and into the cost center's expected/unexpected split. A genuine failure would have been recorded as a routine timeout everywhere. Those commits are reverted; `generative-model.ts` is untouched and still classifies honestly.

Instead the **engine** decides. The turn loop now reconnects on `failed / timeout / malformed`:

```ts
const TURN_RETRY_STATUSES: readonly StopReason[] = ["failed", "timeout", "malformed"];
```

and `auth` becomes the only LLM outcome that stops the run — a rejected credential cannot be retried into working.

So `failed` is retried while still being *called* `failed`. The classifier's judgement is a hint, not a verdict; its imperfection stops being turn-destroying, and observability keeps its distinctions. The asymmetry that justifies it: retrying a genuinely permanent error costs the ladder (~7.75s, with a live countdown and a give-up control on screen) and then ends the same way, while aborting a transient one destroys the turn and leaves the user to re-send and hope.

## Details worth flagging

**The countdown follows.** `plannedRetryDelayMs` gated `retry_in_ms` on `timeout|malformed`, so a retried `failed` would have shown no countdown. It now takes the retryable set explicitly, because it is shared with the compaction loop and an announced wait that does not match what the loop actually does is a lie.

**Compaction is deliberately unchanged.** It keeps the narrower `timeout|malformed` pair: a failed compaction is already graceful — it keeps the original context and retries at the next trigger — so failing fast beats stalling the session behind a ladder. Happy to widen it if you would rather it match.

**A `failed` turn's carry-over marker changes.** It now goes through the reconnect path, so an exhausted ladder produces `[turn_retried]` rather than `[turn_aborted]` — consistent with how exhausted timeouts already behaved, and more accurate ("the attempt failed, this is the retry" vs "you were interrupted"). The flatten path itself is unchanged and still reached by user aborts and `auth`.

## Verification

- `pnpm test` — pass: core 570 / 2 skipped, server 398, cli 200, web 451, landing 39, skills 16, docs 11

New test: a `failed` request carrying the reported message succeeds on retry, the run emits no abort, `request_end` still reports `failed` (not relabelled), and it announces `retry_in_ms` like any other retryable failure.

Two existing flatten-carry-over tests triggered on `failed`; they now trigger on `auth`, the one status still exiting straight to that path. The `[turn_retried]` carry-over of an exhausted ladder was already covered by its own test.

Docs updated in both languages: agent-loop's reconnect section, plus the `timeout`/`failed` rows in architecture, interfaces and message-flow.


hiyouga:
## Review

Reviewed adversarially in a clean worktree, judging the net diff vs main rather than the branch history. The highest-risk question — whether retrying `failed` can duplicate a turn or orphan a `tool_use` — I chased hard and it comes out **safe** (evidence below). But the PR's central justification for accepting the cost is false, and the observability story is inverted.

### 1. The retry is invisible in both frontends — the justification for the trade does not hold

The PR argues the cost of retrying a permanent error is acceptable because "the user watches a live countdown with a give-up control on screen the whole time". That is not true for the class this PR adds.

`packages/web/src/lib/omni/stream-model.ts:1272`

```ts
if (p.status === "timeout" || p.status === "malformed") {
  model.reconnectRun += 1;
  const item: ReconnectItem = { ... }
```

`failed` falls through to the `else` and sets `model.reconnectRun = 0`. No `ReconnectItem` ⇒ no countdown; `findLastWaitingReconnect` returns null ⇒ the "Retry now" / "Give up" controls (`message-item.tsx:154,167`) never render. The `retry_in_ms` this PR newly emits for `failed` is parsed and discarded.

Type-level confirmation that the client was never in scope: `ReconnectItem.status` is `"timeout" | "malformed"` (`stream-model.ts:203`) and `S.chat.reconnect` takes the same union (`strings-en.ts:676`). The CLI is identical — `render.ts:380` (`pendingRetry: "timeout" | "malformed" | null`), no `[Retry]` line.

So the reported case now stalls the session up to 7.75s with **no on-screen explanation and no way to abandon it**. Side effect: because `failed` hits the `else`, a mixed ladder (timeout → failed → timeout) resets the attempt counter to #1.

`docs/content/agent-loop.en.md:94` — which this PR edited — still asserts the countdown and give-up controls for the whole retry set.

### 2. Observability is inverted: a silently recovered blip is escalated as "needs a human"

`packages/server/src/runtime/stream-error-watcher.ts:80` still reads:

```ts
failed: { code: "llm_failed", kind: "unexpected", text: "LLM request failed (not retryable)." },
```

Trace it: `request_end(failed)` stages the pending record; the retry's `request_begin` flushes it and persists an `unexpected` record **even when the next attempt succeeds**. `error-recorder.ts:14-18` defines `unexpected` as "shouldn't happen … needs a human", and the cost center highlights it by default.

Net effect: every blip this PR now hides from the user is escalated to the operator, with fallback copy that literally says "(not retryable)".

Notably the abandoned first approach in this branch had diagnosed exactly this class of problem and added a fix; the rework reverted it without replacing it. Whatever the right answer is (a `kind` that depends on whether the ladder was survived, or a distinct code for an exhausted ladder), it should be decided rather than dropped.

Rider (PLAUSIBLE): `auth` shares the `llm_failed` dedup bucket, and dedup is `(source, code, project)` within a 2s window — so a genuine credential failure landing within 2s of a recovered `failed` is dropped outright.

### 3. `interfaces.{en,zh}.md` now list `failed` twice with contradictory reactions

The PR *added* a row instead of editing the existing one:

```
| `failed` | a failure the classifier did not judge transient | auto-reconnect within the run too (only `auth` stops) |   ← :64
| `failed` | non-retryable (params, …)                       | stop, hand back to the user |                          ← :67
```

Verified present in both locales. `agent-loop.en.md:94` also has a PR-introduced grammar break: "A reconnect the engine re-sends the original input…".

### 4. The LLM layer advertised as "deliberately untouched" now documents the retired policy

`generative-model.ts:19-21` still says credentials failures behave the "same engine behavior as `failed`", and `:1083-1085` says `failed` is "handed to `context_engine` to stop and return control to the user". Keeping the *classifier* honest was the right call; leaving its prose asserting the old policy is not.

### 5. Stale comments and one test whose name now asserts the opposite of reality

- `context-engine.ts:1468-1470` — `buildCarryOver` "Used only for the exit cleanup of aborted / **failed** / auth"; `failed` no longer reaches it.
- `context-engine.ts:938-943` — "…**failed**/auth/aborted exits directly."
- `context-engine.ts:526-529` — "A **timeout/malformed** attempt is never committed to history by AgentHub … so reconnect resends this turn's input unchanged." This is the load-bearing safety invariant for the resend, silently extended to a third status without being restated. (It does hold — see below — but the comment is now the only place it's written down, and it's incomplete.)
- `core/test/engine.test.ts:1732` — `it("an auth outcome stops immediately **like failed**: no retry…")`, with inline comments "Auth behaves like failed" and "the auth/failed branch is the second belt". Two other tests were renamed; this one was missed and now says the opposite of what the code does.

### 6. Orphaned JSDoc

`context-engine.ts:322-346` — the two new consts were inserted *between* the `/** Delay before reconnect attempt N … ≈ 7.75s … */` block and `export function reconnectDelayMs`. Two JSDoc blocks now sit back to back, `reconnectDelayMs` has no doc, and the ladder documentation dangles.

### 7. The exhaustion string for `failed` is ungrammatical and user-visible

`context-engine.ts:585-589` yields e.g. `llm request error: 400 unknown parameter 'foo' failed after 5 retries`, where the siblings are a short token plus the suffix. This is shown verbatim in the error panel and CLI and persisted as the error message. The modified test only asserts `toContain(...)` on two fragments, so nobody read the whole string.

### 8. `[turn_retried]` now tells the model something untrue

`core/src/state/default-config.ts:115` — "the previous attempt of this round failed on a **transient error** (timeout, disconnect, malformed response, or a temporary provider error)". A retried `failed` is by definition the class the classifier judged *non*-transient, and it now carries this block.

### 9. Cost, and a class that cannot possibly succeed

- `generative-model.ts:1228-1233`: a non-completed outcome yields no `token_usage`, so retried attempts are billed but never reach session totals or the cost center. Pre-existing for `timeout`; this PR applies it to the class most likely to be a hard provider verdict, up to ×6.
- PLAUSIBLE: `generative-model.ts:1104-1110` — a `mergeOmniToUniMessage` throw (empty input, mixed roles, bad tool-argument JSON) returns `failed` **before any request is issued**. The engine now retries it 5× with byte-identical input, burning 7.75s and six `request_begin`/`request_end` pairs in the Trace, masking a genuine engine bug behind a stall.

### Checked and found fine

- **Non-idempotency — the thing most likely to make this unsafe — genuinely holds.** AgentHub 0.4.1 `baseClient.js:199-217`: `streamingResponseStateful` pushes into `_history` only *after* the inner `for await` completes, and the validators run inside `streamingResponse` before that returns. **Any thrown error leaves history uncommitted**, so no `failed` path can duplicate a turn or orphan a `tool_use`. No server-side provider state either (`store: false`, no `previous_response_id`), and `generative-model.ts:1223-1226` already guards the committed-turn race.
- **Backoff math is exactly as claimed**: `min(250·2^(n−1), 30000)` over 5 reconnects = 250+500+1000+2000+4000 = **7750ms**, 6 requests total.
- **`auth` end-to-end unchanged**: the branch, the absent countdown (correctly `undefined`, since `auth` is not in `TURN_RETRY_STATUSES`), and the web composer gate all still work.
- **Compaction genuinely unaffected**: `failed`/`auth` still stop there, and `COMPACTION_RETRY_STATUSES` matches what that loop does — no drift between the announced wait and the actual behaviour.
- **Flatten coverage is genuinely preserved**, not quietly reduced: `auth` reaches the identical `buildCarryOver` → `flattenCarryOver` path including the goal-protocol downgrade. Only the now-nonexistent "failed ⇒ flatten" binding was dropped.
- **The new test is meaningful** — restoring main's `context-engine.ts` fails it with `expected 1 to be 2`.
- **Goal mode and subagents**: the dead goal's protocol is still neutralised at the consumer site; a child's ladder cannot blow the parent tool call.
- Typecheck, format and the full suite reproduce.

### Suggested order

1 is the one that undoes the argument for the whole change — either render the countdown for `failed` (extend `ReconnectItem.status` and the two string unions) or drop the claim from the PR and the docs. 2 needs a decision rather than a revert-and-forget. 3–8 are text, but they will otherwise ship as permanent contradictions in the tree.

PR #105, opened 2026-07-28
fix(web): stop rendering an empty bubble for a fidelity-only text or thinking message

hiyouga:
## What causes it

Not a stray empty string — core produces these on purpose.

`flushText` and `flushThinking` emit a **complete message with an empty body** when the provider attached an opaque `fidelity` payload to an otherwise empty part:

```ts
// A text segment with empty text but a fidelity payload (e.g. Gemini carrying a
// thoughtSignature on a text part, or a GPT-5 phase marker with no text) still produces a
// complete message — aligned with flushThinking, so the fidelity isn't lost…
if (this.textBuffer || hasFidelity(this.textFidelity)) {
  yield assistantText(this.textBuffer, stopReason, this.textFidelity);
}
```

That message has to exist: the fidelity must round-trip into history or the next request loses the provider's signature. It just has nothing to show.

Two details explain the exact symptom:

- **Why after thinking.** A thoughtSignature / encrypted-reasoning marker is attached at the thinking→text boundary, so the empty part lands immediately after a thinking segment.
- **Why only sometimes.** It depends on the provider and the shape of the turn — a turn where the model goes from thinking straight to a tool call, on a provider that signs text parts.

## Why it reached the screen

Core never opens a streaming fragment for these — it starts a text segment only once real text arrives (`if (!item.text) break;`). So on the web side there was no `pendingText` / `openText` to replace, and the complete message fell through to the "no open fragment (history / mid-stream join): append directly" path, which built an `AssistantTextItem` with `text: ""`. `message-item.tsx` then rendered it unconditionally as an empty `my-3` block.

The same hole existed on the thinking branch, from the same `flushThinking` case (GPT-5 encrypted reasoning), so both are fixed.

Worth noting the codebase already half-knew: `collectTaskAssistant` filters `assistant_text` items with `it.text.trim()` when gathering a reply's text for copy. The render path just never got the same treatment — now the two agree.

## The fix

Skip the **append** when the body is blank, for text and thinking alike. Deliberately placed after the fragment-replacement branch, so it only suppresses creating a fresh empty item: a fragment that streamed real content is still settled by its complete message exactly as before. (Under core's invariant an empty complete message can never accompany a non-empty fragment — the buffer holds the same deltas — but leaving that branch untouched means the guard cannot blank a bubble that had content.)

Whitespace-only counts as empty, matching what `collectTaskAssistant` already treats as nothing.

## Verification

- `pnpm test` — pass: web 455, core 569 / 2 skipped, server 398, cli 200, landing 39, skills 16, docs 11

Four new tests: an empty assistant text after thinking adds no item; whitespace-only counts as empty for both kinds; real content is unaffected; and a streamed fragment is still settled rather than dropped. **I checked they are not vacuous** — reverting the two guards fails exactly the first two:

```
× an empty assistant text after thinking adds no item
× a whitespace-only body counts as empty too
Tests  2 failed | 453 passed (455)
```

## Not changed

Core still emits the message, and the Trace still records it. That is correct — the fidelity is required on replay, and dropping it upstream would silently degrade the next request on signing providers. This is purely a view-model fix.

### DIFF
--- a/packages/web/src/lib/omni/stream-model.ts
+++ b/packages/web/src/lib/omni/stream-model.ts
@@ -905,6 +905,15 @@ function handleComplete(
         model.pendingText = null;
         return;
       }
+      // Fidelity-only message: core emits a complete text/thinking message with an empty body
+      // when the provider attached an opaque payload to an otherwise empty part — Gemini's
+      // thoughtSignature on a text part, GPT-5's encrypted-reasoning phase markers — both of
+      // which land right after a thinking segment. The message has to exist so the fidelity
+      // round-trips into history, but it has nothing to show, and no fragment was ever opened
+      // for it either (core only starts a segment once real content arrives). Rendering it
+      // produced a blank "assistant:" bubble; collectTaskAssistant already skipped these when
+      // gathering the reply text, so the two paths now agree.
+      if (!p.text.trim()) return;
       // No open fragment (history / mid-stream join): append directly.
       const doneMs = tsOf(timestamp);
       const item: AssistantTextItem = {
@@ -942,6 +951,9 @@ function handleComplete(
         model.pendingThinking = null;
         return;
       }
+      // Same fidelity-only case as the text branch above (GPT-5 encrypted reasoning): the
+      // message carries the payload, not a thought to show.
+      if (!p.thinking.trim()) return;
       const item: ThinkingItem = {
         kind: "thinking",
         id: nextId(model),
--- a/packages/web/src/lib/omni/stream-model.ts
+++ b/packages/web/src/lib/omni/stream-model.ts
@@ -895,6 +895,17 @@ function handleComplete(
       // The complete message usually follows right after a fragment's stop: prefer replacing an already-closed pending fragment, then a still-open one.
       const target = model.pendingText ?? model.openText;
       if (target) {
+        // A blank body discards the fragment instead of settling it (same fidelity-only case as
+        // below — core starts a text segment on the first *truthy* delta, so a whitespace-only
+        // segment does stream). Blanking it in place would leave the live view showing an empty
+        // bubble that a reload then drops. Removing the item and clearing both slots is what
+        // discardFragmentFor does for the dedup path, and it leaves no fragment stuck streaming.
+        if (!p.text.trim()) {
+          removeItem(model, target);
+          if (target === model.openText) model.openText = null;
+          model.pendingText = null;
+          return;
+        }
         // The complete message replaces the fragment's content (this guarantees consistency).
         target.text = p.text;
         target.streaming = false;
@@ -906,13 +917,15 @@ function handleComplete(
         return;
       }
       // Fidelity-only message: core emits a complete text/thinking message with an empty body
-      // when the provider attached an opaque payload to an otherwise empty part — Gemini's
-      // thoughtSignature on a text part, GPT-5's encrypted-reasoning phase markers — both of
-      // which land right after a thinking segment. The message has to exist so the fidelity
-      // round-trips into history, but it has nothing to show, and no fragment was ever opened
-      // for it either (core only starts a segment once real content arrives). Rendering it
+      // when the provider attached an opaque payload to an otherwise empty part — on this text
+      // branch a Gemini thoughtSignature or a GPT-5 `fidelity.phase` segment marker (GPT-5's
+      // encrypted reasoning rides the *thinking* branch instead) — which is why the blank bubble
+      // showed up right after a thinking segment. The message has to exist so the fidelity
+      // round-trips into history, but it has nothing to show, and usually no fragment was opened
+      // for it either (core only starts a segment once a truthy delta arrives). Rendering it
       // produced a blank "assistant:" bubble; collectTaskAssistant already skipped these when
-      // gathering the reply text, so the two paths now agree.
+      // gathering the reply text, and with the blank-fragment discard above both the live and the
+      // history path now agree with it.
       if (!p.text.trim()) return;
       // No open fragment (history / mid-stream join): append directly.
       const doneMs = tsOf(timestamp);
@@ -943,6 +956,13 @@ function handleComplete(
       const tsMs = tsOf(timestamp);
       const target = model.pendingThinking ?? model.openThinking;
       if (target) {
+        // Blank body: discard the fragment rather than settle it (see the text branch).
+        if (!p.thinking.trim()) {
+          removeItem(model, target);
+          if (target === model.openThinking) model.openThinking = null;
+          model.pendingThinking = null;
+          return;
+        }
         target.thinking = p.thinking;
         target.streaming = false;
         if (p.stop_reason !== undefined) target.stopReason = p.stop_reason;
--- a/packages/web/test/stream-model.test.ts
+++ b/packages/web/test/stream-model.test.ts
@@ -1516,4 +1516,56 @@ describe("fidelity-only messages render nothing (empty assistant bubble after th
     expect((texts[0] as AssistantTextItem).text).toBe("Hello");
     expect((texts[0] as AssistantTextItem).streaming).toBe(false);
   });
+
+  // A blank body can also arrive through a fragment: core starts a text segment on the first
+  // truthy delta (`if (!item.text) break;`), and "\n\n" is truthy, so a whitespace-only segment
+  // really does stream. Guarding only the append path would leave live and after-refresh
+  // disagreeing — the fragment kept a blank bubble that a reload then dropped.
+  it("a blank streamed text segment is discarded, so live matches the history rebuild", () => {
+    const live = createStreamModel();
+    pushMessage(live, thinkingMessage("pondering"));
+    pushMessage(live, partialText("start", "\n\n"));
+    pushMessage(live, partialText("stop"));
+    pushMessage(live, assistantText("\n\n"));
+
+    const history = createStreamModel();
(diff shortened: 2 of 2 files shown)


hiyouga:
## Review

Reviewed adversarially in a clean worktree. The diagnosis is correct and independently confirmed — `flushText` / `flushThinking` really do emit complete messages with empty bodies on `hasFidelity(...)`, no `partial start/stop` precedes them, so they land in the append path and rendered an empty `my-3` div. The guard's placement is also right. But the PR does not fully achieve what it claims.

### 1. The stated goal isn't reached: live and after-refresh now render differently

The guard sits *after* the fragment-replacement branch, so it never runs when a fragment was opened. And core does stream a whitespace-only text segment — `if (!item.text) break;` (`generative-model.ts:359`) lets `"\n\n"` through, because it is truthy.

So for a whitespace-only segment:

- **Live**: `start` / `delta "\n\n"` / `stop` + `assistantText("\n\n")` → the replacement branch (`stream-model.ts:896-907`) keeps the item (`text: "\n\n"`, `streaming: false`) → the blank div renders.
- **After F5**: the Trace holds only the complete message → the guard at `:916` drops it → the gap is gone.

Reproduced: live → `["user_text", "assistant_text"]`, history → `["user_text"]`; identical asymmetry for thinking (`:944-953` vs `:956`).

That makes the comment at `:914-915` and the PR body ("collectTaskAssistant already skipped these … the two paths now agree") an overclaim — they agree only on the append path. All four new tests exercise the append path; none covers a blank *streamed* segment, which is exactly the hole.

Fix: in the replacement branch, when the complete body is blank, **remove** the fragment item — `discardFragmentFor` (`stream-model.ts:1430`) already does this — rather than blanking it, then clear `openText`/`pendingText`. That makes both paths agree without reintroducing a stuck spinner.

### 2. The guard silently swallows a `[stopReason]` marker in a narrow case

`message-item.tsx:299-301` renders `[{stopReason}]` for any non-`completed` stop reason, and `finishInterrupted(reason)` → `flushText(reason)` emits `assistantText("", reason, fidelity)` (`generative-model.ts:517-519, 609-610`). So an interrupted turn whose only text was a fidelity marker used to show a lone `[failed]` / `[aborted]`, and now shows nothing.

Mitigations cover some reasons (abort → interruption item; timeout/malformed → retry-hint item), but **`failed` and `auth` render nothing from `request_end`** (`stream-model.ts:1244-1263` — `auth` only flips the composer gate). Narrow (needs fidelity present, no text, and no thinking in the turn), but the trade-off isn't mentioned.

### 3. The placement is right; the reassurance about it is hollow

I verified core's invariant: `textStarted` / `thinkingStarted` are set in the same statement pair that appends to the buffer, and `partial(stop)` is emitted only when started — so a stop implies a non-empty body. Guarding *before* the replacement branch would strand `openText.streaming = true` forever (a permanent blinking `▌`), so after is correct.

But the PR's "leaving that branch untouched means the guard cannot blank a bubble that had content" is hollow: feeding `start` / `delta "Hello"` then `assistantText("")` yields `{ text: "", streaming: false }` — the pre-existing replacement branch blanks it into exactly the bug this PR fixes. The guard isn't the cause, but it also isn't a defence. Worth saying plainly rather than implying protection that isn't there.

### 4. Surface divergence with the CLI

`cli/src/render.ts:189-204` (`renderHistory`, used by `--resume`) writes `${p.text ?? ""}${marker}\n` unconditionally, so the same message still prints a blank line — or a lone dim `[aborted]`. The CLI's live renderer ignores complete assistant text entirely, so it was never affected. Net: web and CLI-resume now disagree, unacknowledged. (The Trace viewer still showing it is correct — it is a raw event dump.)

### 5. Nit

`stream-model.ts:910-911` says "GPT-5's encrypted-reasoning phase markers", conflating two things: on the **text** branch core calls them `fidelity.phase` segment markers (`generative-model.ts:339-343`); encrypted reasoning (`{id, encrypted_content}`) is the **thinking** branch (`:376-378`). Minor, except this comment is now the only documentation of the renderer-side contract — `docs/content/omni-message.en.md:288` documents the empty-body message but says nothing about what a client should do with it.

### Checked and found fine

- **`touchTask` before the guard is correct** — the message really did occur in the round, `taskLastTsMs` is only a round-end fallback, and it matches the `[context_summary]` branch. `advanceLastTs` also runs regardless, so the next thinking item's `startedAtMs` and the next tool card's `argStartedAtMs` are unaffected (verified).
- **Nothing else is lost.** `AssistantTextItem` / `ThinkingItem` carry no `fidelity`; core still emits it and the Trace still records it. `collectTaskAssistant` already filtered on `.trim()`, so copy text and the stats footer's `atMs` are unchanged. Bonus: a replayed-but-not-deduped empty message no longer duplicates.
- **Live-tail is not a second source.** A synthetic `partial_text start` with `text: ""` *is* reachable (a capture can land between `start` and its first `delta`), but core always follows `start` with a `delta` and the client drops partials at/below the cursor — the seeded empty item is transient.
- **`.trim()` is a defensible test** for the append path: a whitespace-only body renders to nothing through `Md` anyway, and `collectTaskAssistant` already treated it as nothing. Its only cost is finding 1.
- **Multi-agent**: the guard lives in the shared complete-message handler, so origin-routed nested models get it too.
- **The tests are not vacuous, exactly as claimed** — stubbing out both guards gives `2 failed | 76 passed`, matching the two named tests. Test 4 is a genuine regression guard on the placement.
- Verification reproduced independently on Node 24, with the exact counts in the PR body.

### Suggested direction

Finding 1 is the one worth acting on — it is the difference between "fixed" and "fixed in one of the two paths", and the fix is a few lines in the branch immediately above the guard. 2 and 4 are worth a sentence in the PR body even if you decide not to act on them.

packages/web/src/lib/omni/stream-model.ts
```diff
@@ -905,6 +905,15 @@ function handleComplete(
...
       }
+      // Fidelity-only message: core emits a complete text/thinking message with an empty body
+      // when the provider attached an opaque payload to an otherwise empty part — Gemini's
+      // thoughtSignature on a text part, GPT-5's encrypted-reasoning phase markers — both of
+      // which land right after a thinking segment. The message has to exist so the fidelity
+      // round-trips into history, but it has nothing to show, and no fragment was ever opened
+      // for it either (core only starts a segment once real content arrives). Rendering it
+      // produced a blank "assistant:" bubble; collectTaskAssistant already skipped these when
+      // gathering the reply text, so the two paths now agree.
+      if (!p.text.trim()) return;
```
hiyouga, line 929: 这个return的话如何保证fidelity进trace，resume时候能回放正常？

packages/web/src/lib/omni/stream-model.ts
```diff
@@ -942,6 +951,9 @@ function handleComplete(
         model.pendingThinking = null;
         return;
       }
+      // Same fidelity-only case as the text branch above (GPT-5 encrypted reasoning): the
+      // message carries the payload, not a thought to show.
+      if (!p.thinking.trim()) return;
```
hiyouga, line 976: 这个return的话如何保证fidelity进trace，resume时候能回放正常？

hiyouga:
## Addressed: finding 1 (and the finding 5 nit)

Pushed `c37d3a0`.

**Finding 1 — reproduced, then fixed.** The three new tests fail on `8adb7d2` and pass on `c37d3a0`:

```
× a blank streamed text segment is discarded, so live matches the history rebuild
× a blank streamed thinking segment is discarded too
× discarding a blank fragment clears the open-fragment slots, leaving no stuck spinner
```

The fragment-replacement branch now discards the fragment when the complete body is blank — `removeItem` plus clearing `openText` / `pendingText`, which is exactly what `discardFragmentFor` does for the dedup path — rather than settling it in place. Removing instead of blanking is what avoids the stranded `streaming: true` the review flagged. Thinking gets the same treatment. Live and after-refresh now produce identical items for a whitespace-only segment; the tests assert that equality directly rather than each side in isolation.

`discardFragmentFor` itself is not reused here: it re-routes by `msg.origin`, and at this point the reducer has already resolved the nested model, so calling it would route a second time.

**Finding 5 — corrected.** The note above the append guard no longer conflates GPT-5's `fidelity.phase` segment markers (text branch) with its encrypted reasoning (thinking branch), and no longer claims the two paths agree — with the discard in place they now actually do, which the comment says instead.

**Findings 2 and 4 — acknowledged, not acted on.**

- **2 (a `[failed]` / `[auth]` marker on a fidelity-only interrupted turn is now silent).** Real but very narrow: it needs a turn whose *only* assistant output is a fidelity marker, no text and no thinking, ending in `failed` or `auth` specifically — `aborted` and `timeout` still render their own items. Showing a lone `[failed]` attached to nothing was not obviously better than showing nothing, and the request-level error surfaces are the right place for that signal. Worth a separate issue if it is ever seen in practice.
- **4 (CLI `--resume` still prints a blank line for these).** Correct, and out of scope for a web fix — `cli/src/render.ts:189-204` needs the same blank-body skip. Better as its own PR against the CLI renderer than smuggled into this one.

**Verification** (Node 24, worktree root): `pnpm -r build`, `pnpm typecheck` pass; `pnpm test` passes — web **458** (455 + 3), core 569 / 2 skipped, server 398, cli 200, landing 39, skills 16, docs 11; `pnpm format:check` clean.

Note the "The fix" section of the PR body describes `8adb7d2` only — this commit widens it from "skip the append" to "skip the append and discard a blank fragment".

PR #106, opened 2026-07-28
fix(web): drop the stop-reason pill from the tool card row; the status icon carries it

hiyouga:
## What

The tool card's collapsed row stated a call's outcome twice. The status icon on the left already distinguishes it:

```ts
failed: "M12 21a9 9 0 1 0 0-18 9 9 0 0 0 0 18zM9 9l6 6m0-6-6 6",   // red circled X
done:   "M12 21a9 9 0 1 0 0-18 9 9 0 0 0 0 18zm-3.5-9.2 2.4 2.5 4.6-4.8",  // grey check
```

…and then a pill to the right of the duration spelled the word out again (`aborted`, `failed`, `timeout`). Both pills are removed — the call's and the output's — along with the now-unused `Badge` / `stopReasonTone` import.

## Why nothing is lost

I checked each channel the removed text was carrying before deleting it:

- **The visual signal** — `state` already resolved to `"failed"` from *either* stop reason (`callStopReason` or `outputStopReason` being non-`completed`), so the red X was already showing for every case the pill covered.
- **The exact reason** — `stateLabel` already falls through to `item.outputStopReason ?? item.callStopReason`, and `StatusIcon` renders it as `title` + `aria-label`. So the literal word is still available on hover and to screen readers.
- **The data** — the view model keeps both fields untouched; the unit tests that assert them (`card.outputStopReason === "aborted"` and friends) still pass unchanged.
- **The literal value on a surface that wants it** — the Trace viewer still renders the raw stop reason per event via its own `Badge` / `stopReasonTone`, which is where an exact protocol value belongs.

## Consistency

The row had already taken this direction for the approval decision, whose existing comment reads: *"carried ONLY by the left status icon's title/aria-label — per review the row shows no visible decision text at any breakpoint; the icon is the single source of truth for how the call was decided."* This applies the same rule to the outcome. The `deniedByUser` guard that existed to stop the pill from repeating the deny decision is no longer needed for that purpose, though the flag itself still feeds `stateLabel`; its comment is updated rather than left stale.

Practical benefit beyond the redundancy: the left group of that row is the one that gives way on narrow screens, and the pill was competing for the width the call's subtitle needs.

## Verification

- `pnpm test` — pass: web 451, core 569 / 2 skipped, server 398, cli 200, landing 39, skills 16, docs 11

No test or e2e spec asserted the pill's rendered text (the existing assertions are on the model fields), so nothing needed updating. No docs describe it either.

## Scope note

Only the tool card. The assistant-text row has a similar `[{stopReason}]` marker in `message-item.tsx`, left alone deliberately: there is no status icon beside it, so removing it there would drop the information rather than deduplicate it. Say the word if you want that one reconsidered too.

### DIFF
--- a/packages/web/e2e/malformed.spec.mjs
+++ b/packages/web/e2e/malformed.spec.mjs
@@ -78,7 +78,14 @@ test("a malformed tool_call settles unpaired; the retry line shows and the retry
     .first()
     .click();
   const group = page.locator(".anim-msg.my-2").first();
-  await expect(group).toContainText("malformed");
+  // Visibility is asserted explicitly: `toContainText` walks every descendant except
+  // SCRIPT/NOSCRIPT/STYLE, so the StatusIcon's own SVG <title>malformed</title> satisfies it
+  // even when nothing is drawn. This call never ran, so it has no output block to expand
+  // into — the row's `[malformed]` marker is the only explanation a touch user can reach.
+  await expect(
+    group.getByText("[malformed]").filter({ visible: true }),
+    "the broken tool card shows a visible malformed marker",
+  ).toHaveCount(1);
   // A running spinner has role=status; none should remain after the turn ends.
   await expect(page.locator('[role="status"]')).toHaveCount(0);
 });
--- a/packages/web/src/features/chat/tool-call-card.tsx
+++ b/packages/web/src/features/chat/tool-call-card.tsx
@@ -17,7 +17,6 @@ import { S } from "../../lib/strings";
 import { humanizeDuration } from "../../lib/format";
 import { approvalKey } from "../../lib/omni/stream-model";
 import type { ToolCallItem } from "../../lib/omni/stream-model";
-import { Badge, stopReasonTone } from "../../components/ui/badge";
 import { Chevron } from "../../components/ui/chevron";
 import { ZoomableImage } from "../../components/ui/image-zoom";
 import { StatusIcon } from "../../components/ui/status-icon";
@@ -201,8 +200,8 @@ export function ToolCallCard({ item, ctx }: { item: ToolCallItem; ctx: StreamRen
     : null;
   // A user denial reports stop_reason "aborted" on the output it feeds back; that abort IS the
   // decision, not an independent outcome — the icon reads "Denied", and no separate "aborted"
-  // badge repeats it. A user-abort of a RUNNING tool carries no deny decision and keeps its own
-  // "aborted" marker (the stop-reason branch below).
+  // repeats it. A user-abort of a RUNNING tool carries no deny decision, so the label falls
+  // through to its stop reason below.
   const deniedByUser = item.decision === "deny" && item.outputStopReason === "aborted";
   const stateLabel = pending
     ? S.chat.approvalWaiting
@@ -216,7 +215,15 @@ export function ToolCallCard({ item, ctx }: { item: ToolCallItem; ctx: StreamRen
 
   return (
     <div>
-      {/* Collapsed row: status icon + tool name + total duration (generation + execution, excluding approval wait). Expand chevron on the right. */}
+      {/* Collapsed row: status icon + tool name + total duration (generation + execution,
+          excluding approval wait). Expand chevron on the right.
+
+          Outcome is the icon's job alone — a red X for failed/aborted/timeout, a grey check
+          for done — with the exact stop reason in its title/aria-label. There used to be a
+          pill spelling it out to the right of the duration as well; it said the same thing
+          twice, and on a phone it competed for the width that the call's subtitle needs. The
+          Trace viewer still shows the raw stop reason per event, which is where the literal
+          value belongs. */}
       <button
         type="button"
         aria-expanded={open}
@@ -262,14 +269,6 @@ export function ToolCallCard({ item, ctx }: { item: ToolCallItem; ctx: StreamRen
           <span className="hidden shrink-0 font-mono text-xs text-amber-600 sm:inline dark:text-amber-400">
             {S.chat.approvalWaiting}
           </span>
-        )}
-        {item.callStopReason && item.callStopReason !== "completed" && (
-          <Badge tone={stopReasonTone(item.callStopReason)}>{item.callStopReason}</Badge>
-        )}
-        {/* Writing "aborted" next to the Denied pill would state the same outcome twice — see
-            deniedByUser above. */}
-        {item.outputStopReason && item.outputStopReason !== "completed" && !deniedByUser && (
-          <Badge tone={stopReasonTone(item.outputStopReason)}>{item.outputStopReason}</Badge>
         )}
         <span className="min-w-0 flex-1" />
         {/* Expand indicator on the right */}
--- a/packages/web/src/features/chat/tool-call-card.tsx
+++ b/packages/web/src/features/chat/tool-call-card.tsx
@@ -1,20 +1,23 @@
 /**
  * Tool call card: collapses to a single line by
- * default — status icon + tool name + duration (a live-ticking timer while running) + status
- * badge; clicking expands full arguments and output; a bound subagent renders as a full-width
- * shortcut row below them regardless of collapsed state (the child conversation itself lives
- * in the subagents side panel; the row carries its own pending-approval dot, so the card no
- * longer needs to auto-expand for nested approvals).
+ * default — status icon + tool name + duration (a live-ticking timer while running) + a plain
+ * `[stop reason]` marker when the step did not finish cleanly; clicking expands full arguments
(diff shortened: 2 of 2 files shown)


hiyouga:
## Review

Reviewed adversarially in a clean worktree. The change is small and the state-machine claim at its core holds — `failed` (`tool-call-card.tsx:183-185`) is exactly the disjunction of the two old pill conditions, so the red X really does appear wherever a pill did. But "nothing is lost" does not survive contact with the details, and two statements in the PR body are wrong.

### 1. Cards with no output lose their explanation entirely on touch devices

The exact reason now lives only in `title` / `aria-label` (`status-icon.tsx:58-61`). The fallback for a phone — expand the card and read the output — works only when the tool actually ran; core appends `[tool timeout: …]` / `[tool error] …` into the output text. Two client-synthesised paths never produce output at all:

- `settleUndispatchedCall` (`stream-model.ts:1060-1065`) — a `malformed`/`timeout` tool_call that was never dispatched. It sets `outputComplete` and copies `outputStopReason`, but no output ever arrives.
- `closeExecutingToolCards` (`stream-model.ts:1042-1051`) — cards closed as `aborted` with no output.

In both, `tool-call-card.tsx:322` renders no output block, so a malformed tool call becomes a red X with **zero** explanation anywhere in the chat view, with no hover to recover it. That is the case the repo's own `malformed.spec.mjs` exercises, and it is the opposite of the PR's stated motivation (phone width).

### 2. The amber/red severity distinction is silently destroyed

`stopReasonTone` (`badge.tsx:31-39`) mapped `aborted` → **amber** and everything else → **red**, and `badge.tsx:12-15` documents amber explicitly as "warning semantics (aborted stop_reason…)". `StatusIcon` has exactly one failure tone — red (`status-icon.tsx:21`).

So a user-interrupted call is now visually identical to a hard tool failure. The PR's claim that "the red X was already showing for every case the pill covered" is true about *presence* and quietly drops *severity*. This is not mentioned anywhere in the PR.

### 3. It introduces the inconsistency it claims to remove — the scope note is factually wrong

`thinking-block.tsx:43` renders a visible `[{item.stopReason}]` marker on a row that **does** have a `StatusIcon`, using the identical `state` / `stateLabel`-as-title pattern (`thinking-block.tsx:19-23, 37`), inside the **same** WorkGroup, and whose own header comment says it "Shares the same row style and set of running-state icons (in progress / done / failed) as tool cards".

The PR's scope note names only `message-item.tsx` and justifies leaving it with "there is no status icon beside it". That justification does not apply here, and this file was not mentioned at all. After merge, a failed thinking step shows `[malformed]` while the failed tool step directly beneath it shows nothing.

Worth noting the better fix this suggests: the thinking row's marker is plain gray `[text]`, **not** a padded pill. A cheap-width treatment already exists in this codebase — aligning the tool row onto it would have addressed the width complaint without deleting the information.

The CLI also still prints ` [aborted]` on tool_call lines (`cli/src/render.ts:189, 216`), so the surfaces now disagree three ways.

### 4. "No test or e2e spec asserted the pill's rendered text" is false, and one assertion became a tautology

`packages/web/e2e/malformed.spec.mjs:74-81` — comment: "the broken tool card **shows** malformed"; assertion: `await expect(group).toContainText("malformed")`.

It still passes, but only because Playwright's `toContainText` walks all descendants and skips only SCRIPT/NOSCRIPT/STYLE — so it now reads the SVG `<title>malformed</title>`. Verified empirically against this DOM shape: `toContainText` passes while `getByText(...).filter({ visible: true })` yields 0. The assertion no longer verifies any visible UI.

Compounding it: the web Playwright suite is **not in CI**. `.github/workflows/ci.yml:58` runs the root `test:e2e`, which is `pnpm --filter @prismshadow/penguin-core test:e2e`; `packages/web`'s `test:e2e` (`bash e2e/run.sh`) is manual only. So the PR's verification list could not have covered this, and the claim was made without checking.

### 5. Stale module header

`tool-call-card.tsx:2-4` still describes the row as "status icon + tool name + duration … **+ status badge**". The inline comment was rewritten; the file header was not.

### Minor

- `deniedByUser`: the old code suppressed only the *output* pill for a denial — the *call* pill still rendered. The new `stateLabel` (`tool-call-card.tsx:212-213`) returns `decisionText` alone, so a non-`completed` `callStopReason` on a denied card would be invisible. I could not construct a reachable path (`settleUndispatchedCall` closes such calls before approval is requested), so this is a claim-accuracy gap rather than a live bug.
- `tool-call-card.tsx:202-203` reads "the icon reads "Denied", and no separate "aborted" repeats it" — the noun was deleted with the code, leaving a dangling sentence.

### Checked and found fine

- Typecheck passes; 451/451 web unit tests pass; prettier clean. Diff is genuinely one file (+11/−12).
- No dead code: `deniedByUser` is still read at `212`; `Badge` / `stopReasonTone` remain used by `traces/trace-event-row.tsx` and `chat-input.tsx`, so dropping the import here orphans no export.
- Screen-reader accessibility genuinely is preserved — `role="img"` + `aria-label` contributes to the enclosing button's accessible name. The gap is touch, not assistive tech.
- `layout.spec.mjs:576-612` still holds; its `filter({ visible: true })` correctly yields 0 because an SVG `<title>` has no box.
- The `pending`/`executing` precedence in the `state` ternary can in principle mask a failed `outputStopReason`, but core yields the partial stop and the complete output back-to-back, making that a sub-frame window rather than a defect.
- No docs describe the pill; omitting a changelog entry matches the project's separate-aggregate-PR convention.

### Suggested direction

Findings 1–3 are worth acting on before merge. The cheapest resolution that satisfies the original request — outcome carried by the icon, not competing for width — is to align the tool row on the thinking row's existing plain-gray `[reason]` treatment rather than removing it, keep an amber-vs-red distinction for `aborted`, and fix the `malformed.spec.mjs` assertion so it checks visible text.

hiyouga:
“待审批”这几个字移除掉吧，因为审批界面总是展示

hiyouga:
## Addressed: findings 1–5, plus the "待审批" request

Pushed `ff12f0d`.

**Findings 1 + 2 + 3 — one fix.** The stop reason is back on the row, but as the plain `[reason]` marker `thinking-block.tsx:43` already uses, not the padded Badge pill that competed for width. That is the direction the review suggested, and it resolves all three at once:

- **1** — a call that never ran (malformed, or closed undispatched) has a visible explanation again, no hover required.
- **2** — amber for `aborted` against red for a real failure, so a user interruption no longer reads identically to a hard failure. `StatusIcon` still has its single failure tone; the marker carries the severity.
- **3** — the thinking row and the tool row inside one work group now render their outcomes the same way, instead of `[malformed]` above and nothing below.

One thing the review did not catch, which only surfaced once the marker was restored and the e2e was actually run: `settleUndispatchedCall` does `card.outputStopReason ??= card.callStopReason`, so an undispatched call ends with **both** fields set to the same value. The two old pills therefore rendered `[malformed][malformed]` side by side. Identical reasons now collapse to one marker.

**Finding 4 — fixed, and confirmed by running it.** The assertion checks visible text now. You were right that it was a tautology: it passed on the pill-less build by reading the icon's own SVG `<title>malformed</title>`. And the observation that web Playwright is not in CI is correct — `package.json:15` scopes the root `test:e2e` to `@prismshadow/penguin-core`, so `.github/workflows/ci.yml:58` never reaches `packages/web`'s suite. I ran it by hand against a temp data root: **layout / malformed / compact-abort → 6 passed, 1 failed**. The failure is `layout: mobile chat dropdowns stay inside the viewport`, which fails identically with these changes stashed — it predates them.

**Finding 5 — fixed.** The module header no longer advertises a "status badge", and the stale "Waiting for approval badge" sentence is gone too.

**Minor — fixed.** The dangling `no separate "aborted" repeats it` now reads `no separate `[aborted]` marker repeats it`. The `deniedByUser` gap the review flagged closes on its own: the call's own reason still renders, only the output's is suppressed, exactly as before.

---

**On 「待审批」.** Removed from the tool card row (`tool-call-card.tsx`) — while a call is pending, the approval block below it is always on screen with the tool name, the arguments and the Allow/Deny buttons, so the row was pure repetition. The labeled amber hourglass icon still marks the wait, now at every breakpoint rather than below `sm` only.

The other three occurrences are deliberately kept:

| | condition | approval UI visible? |
|---|---|---|
| `work-group.tsx:143` pill | `pending && **!shown**` | **no** — renders only while the group is collapsed, where the cards inside are hidden. It is the sole signal there. |
| `tool-call-card.tsx` `stateLabel` | — | not visible; StatusIcon title/aria-label |
| `subagent-chip.tsx:40` | — | not visible; chip accessible name, and the subagent panel may be closed |

Removing the work-group pill would repeat the mistake finding 1 identifies — deleting information that has no other surface.

PR #111, opened 2026-07-29
fix(web): prevent duplicate docked panel dividers

laodouuu:
## Problem

Opening the Workspace panel could show two vertical dividers between it and the chat, while the Agents panel showed only one.

## Root cause

Both docked panel shells stay mounted at zero width to preserve their close animation and internal state. Their divider lived on that persistent animation shell, so the closed Agents panel could still paint a border immediately before the open Workspace panel's own divider.

## Fix

- Move the divider from each persistent animation shell to its fixed-width inner panel.
- Let the existing `overflow-hidden` clipping window hide the panel content and divider together at zero width.
- Apply the same structure to the Workspace and Agents panels without changing their animation, resize handling, or mounted state.
- Add an unreleased changelog entry.

## Validation

- `pnpm build` — passed
- `pnpm format:check` — passed
- `pnpm typecheck` — passed
- `pnpm --filter @prismshadow/penguin-web test` — 34 files, 451 tests passed
- `git diff --check` — passed

The full local `pnpm test` run is not reported as green because this Windows host does not permit unprivileged symlink creation: one Core and four Server symlink tests fail with `EPERM`. After selecting the installed Git Bash explicitly, all shell-related tests passed; the current upstream `main` commit also passes both Linux and Windows GitHub CI jobs.

### DIFF
--- a/changelog/unreleased/2026-07-29-docked-panel-divider.md
+++ b/changelog/unreleased/2026-07-29-docked-panel-divider.md
@@ -1,0 +1,8 @@
+# Keep docked panel dividers inside the visible panel
+
+The Web app no longer shows a duplicate divider between the chat and Workspace panels.
+
+## Web App
+
+- Move the divider from each docked panel's persistent animation frame to its clipped inner panel.
+- Preserve the existing open and close animation, resizing behavior, and mounted panel state.
--- a/changelog/unreleased/2026-07-29-docked-panel-divider.md
+++ b/changelog/unreleased/2026-07-29-docked-panel-divider.md
@@ -1,8 +1,0 @@
-# Keep docked panel dividers inside the visible panel
-
-The Web app no longer shows a duplicate divider between the chat and Workspace panels.
-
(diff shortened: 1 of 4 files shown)


changelog/unreleased/2026-07-29-docked-panel-divider.md
Youhai020616, line 1: 我不确实是不是要写changelog，我看他们之前提交的pr没有写，可以问问

Youhai020616 (approved):
LGTM

Youhai020616 (approved):

hiyouga:
#114 修过一次了，你看看怎么合并一下方案

laodouuu:
> #114 修过一次了，你看看怎么合并一下方案

我觉得可以关闭此PR，修复效果基本是一样的

hiyouga:
ok 那就关闭吧

laodouuu:
已在#114 实现

PR #113, opened 2026-07-29
fix(web): honor currency setting in benchmark costs

laodouuu:
## Summary

- honor the selected display currency in Benchmark cost trend axes and tooltips
- apply the same currency conversion to evaluation, case, and individual run costs
- keep score and duration formatting unchanged

## Root cause

Benchmark cost rendering called `formatMoney` without passing the currency selected in settings. Because `formatMoney` defaults to USD, Benchmark costs remained in dollars while the model library, cost center, chat, and Trace views followed the USD/CNY preference.

The currency preference and Benchmark page both entered the repository in the initial source import, so this appears to be an initial coverage omission rather than a later feature-ordering issue.

## Design note

If Benchmark costs are intentionally designed to remain fixed in USD and the Benchmark view is not meant to support display-currency switching, please close this PR.

## Validation

- `pnpm format`
- `pnpm format:check`
- `pnpm typecheck`
- `pnpm build`
- `pnpm --filter @prismshadow/penguin-web test` — 451 tests passed

The full `pnpm test` command was also attempted locally on Windows. Core shell tests initially resolved to PowerShell because Git Bash was not on `PATH`; explicitly selecting Git Bash removed those shell-syntax failures. One unrelated symlink-guard test still requires Windows symlink privileges, and one timing assertion passed when rerun in isolation.

### DIFF
--- a/changelog/unreleased/2026-07-29-benchmark-currency.md
+++ b/changelog/unreleased/2026-07-29-benchmark-currency.md
@@ -1,7 +1,0 @@
-# Web App: Benchmark costs follow the selected display currency
-
-Benchmark cost trends, evaluations, cases, and individual runs now use the display currency selected in settings.
-
-## Consistent cost formatting
-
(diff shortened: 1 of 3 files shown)


Youhai020616 (approved):
LGTM

hiyouga (approved):

PR #115, opened 2026-07-29
fix(core): preserve committed responses after stop

laodouuu:
## Problem

PR #107 guarantees that `streamGenerate` exits when an upstream iterator ignores its `AbortSignal`, but its new `STOPPED` branch classified every abort race as `aborted` or `timeout`.

If AgentHub had already delivered `finish_reason`, the response was committed even when the transport failed to close afterward. Downgrading that response could send ContextEngine down the incomplete-turn cleanup or reconnect path, risking broken tool-call/output pairing or duplicate work.

## Fix

- Preserve normal completion when `STOPPED` wins after `finish_reason` was already observed.
- Keep the existing `aborted` and `timeout` classification when no terminal event was received.
- Add deterministic regressions for both a user Stop and idle timeout after a committed terminal tool-call event whose upstream iterator never settles.

## Changelog

Intentionally omitted here; this change will be recorded in a separate batched changelog PR alongside #111 and #113.

## Validation

- Independent read-only agent review: no actionable or blocking findings.
- `pnpm --filter @prismshadow/penguin-core exec vitest run test/llm.test.ts` — 87 passed.
- `pnpm --filter @prismshadow/penguin-core typecheck` — passed.
- `pnpm format:check` — passed.
- Full core suite attempted: 571 passed, 4 skipped; 4 unrelated Windows-host failures remain (one symlink-privilege `EPERM` case and three long-running command-session timing cases).

### DIFF
--- a/packages/core/src/llm/generative-model.ts
+++ b/packages/core/src/llm/generative-model.ts
@@ -1188,9 +1188,13 @@ export class GenerativeModel implements LLMInterface {
         if (res === STOPPED) {
           // Upstream never settled after the abort. Abandon it rather than await it: ask it to
           // close (best effort — a stream that ignored the signal may ignore this too, so the
-          // rejection is swallowed and the promise is not awaited) and classify by trigger.
+          // rejection is swallowed and the promise is not awaited). A received finish_reason
+          // means AgentHub already committed the response, so preserve normal completion;
+          // otherwise classify the interrupted request by its trigger.
           void Promise.resolve(it.return?.(undefined)).catch(() => undefined);
-          outcome = userSignal?.aborted ? { status: "aborted" } : { status: "timeout" };
+          if (!translator.sawFinishReason()) {
+            outcome = userSignal?.aborted ? { status: "aborted" } : { status: "timeout" };
+          }
           break;
         }
         if (res.done) break;
--- a/packages/core/test/llm.test.ts
+++ b/packages/core/test/llm.test.ts
@@ -1429,6 +1429,23 @@ describe("GenerativeModel.streamGenerate outcome classification (PRN-013)", () =
     await new Promise(() => {}); // never settles, never observes the signal
   }
 
+  /** A committed terminal tool-call event whose transport never closes afterward. */
+  async function* committedThenDeaf(): AsyncGenerator<UniEvent> {
+    yield ev({
+      event_type: "stop",
+      finish_reason: "tool_call",
+      content_items: [
+        {
+          type: "tool_call",
+          name: "exec_command",
+          arguments: { cmd: "echo done" },
+          tool_call_id: "committed-call",
+        },
+      ],
+    });
+    await new Promise(() => {}); // response committed, but the transport never closes
+  }
+
   /** Fails the test loudly instead of letting a regression hang the whole suite. */
   function withTimeout<T>(p: Promise<T>, ms: number): Promise<T> {
     return Promise.race([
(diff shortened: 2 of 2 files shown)


hiyouga:
## The premise does not hold for agenthub 0.4.1 — this would introduce the failure it aims to prevent

I reached the same conclusion as this PR while reviewing #107, wrote a probe to confirm it, and then found the opposite once I read AgentHub's source. Posting the evidence rather than the conclusion, so it can be checked.

### Where the commit actually happens

`node_modules/@prismshadow/agenthub/dist/baseClient.js:199-217`:

```js
async *streamingResponseStateful(options) {
  const events = [];
  for await (const event of this.streamingResponse({ ... })) {
    events.push(event);
    yield event;
  }
  // ↓ the commit — AFTER the for-await drains
  if (events.length > 0) {
    this._history.push(tempMessages[tempMessages.length - 1]);
    this._history.push(this.concatUniEventsToUniMessage(events));
  }
}
```

The turn is pushed into `_history` **after the loop**, i.e. during the final `next()` that returns `done` — not when `finish_reason` arrives. `AutoLLMClient.streamingResponseStateful` (`autoClient.js:163`) is itself an `async *` delegating through its own `for await`, so it inherits the same shape.

So "`finish_reason` delivered" does **not** imply "committed". The real invariant is **the generator ran to completion ⟺ the turn was committed**.

### What that means for the STOPPED branch

`STOPPED` winning the race means `it.next()` had not settled — the generator is still suspended inside its `for await`, before the commit. `it.return()` then resumes it with a return completion, which exits the generator body without executing the post-loop code. In the deaf case the pending await never settles and nothing runs at all. Either way: **no commit**.

Reproducing `baseClient`'s exact shape:

```js
const history = [];
async function* innerStream() {
  yield { finish_reason: null, text: "hi" };
  yield { finish_reason: "tool_call" };   // terminal event delivered
  await new Promise(() => {});             // transport never closes — the #107 case
}
async function* streamingResponseStateful() {
  const events = [];
  for await (const ev of innerStream()) { events.push(ev); yield ev; }
  history.push({ role: "assistant", events });   // the commit
}
// …drive it exactly as streamGenerate does, aborting after the terminal event…
```

```
finish_reason observed : true
history committed      : false   []
```

### The consequence

| | AgentHub `_history` | reported outcome | consistent? |
|---|---|---|---|
| #107 as merged | not committed | `aborted` / `timeout` | ✅ both sides agree the turn did not happen |
| with this PR | not committed | `completed` | ❌ |

On the mismatch: ContextEngine keeps the `tool_call` as completed, dispatches it, produces a `tool_result`, and the next request carries a `tool_result` whose `tool_use` is absent from AgentHub's history. The provider rejects it (400), and per the comment this PR is trying to honour (`generative-model.ts:1244-1250`) *"the engine has no fix-up path left that touches LLM history"*. That is the exact failure mode the comment warns about — reached from the other direction.

### Why the new tests pass anyway

`SeamModel` overrides `openStream` with a bare async generator, which has no history commit at all. That makes "committed or not" unobservable, so both classifications look self-consistent under test. It is what misled me first time round, and I suspect it is why the independent review found nothing actionable.

### The race this PR guards against cannot occur

If the turn really was committed, the generator has already run to completion and `it.next()` is already resolved. `Promise.race` attaches its reaction to `it.next()` before `stopped`, so an already-settled `it.next()` schedules first and wins — including against an abort fired in the same tick. There is no reachable "committed but reported `aborted`" state.

### Caveat

All of the above is tied to where agenthub **0.4.1** performs the commit. If a later AgentHub moves it onto the `finish_reason` event, this PR's reasoning becomes correct and #107's becomes wrong. Whichever way it lands, the invariant deserves a test that fails loudly if a dependency bump flips it — that is the durable fix here, and it would be worth keeping from this PR even though I think the classification change itself should not be merged.

PR #119, opened 2026-07-29
fix(web): scroll the sidebar nav with the session list, in both sidebar states

hiyouga:
Follow-up to #116, which fixed one way a page grows the document (an absolutely positioned descendant escaping its scroller) and reported the other one: content that cannot shrink.

## The floor

The sidebar column had exactly one shrinkable block — the session list. Everything else was `shrink-0`: the Project switcher, New chat, the eight nav entries, the user row. Together they need roughly **412px**, so in a shorter window the session list collapses to zero and the nav is pushed out of the `<aside>`. Nothing clips it (`aside` has no `overflow`, nor does the layout row above it), so the overflow becomes the document's scrollable height: a second scrollbar appears and dragging it scrolls the entire shell.

Measured before this change: at a 1440x320 viewport every page reported the identical `+92px` of document overflow — identical across pages because the cause is the one thing they share.

A 412px-tall viewport is not exotic: browser zoom past 200% and bottom-docked devtools both get there.

## The change

**Pinned sidebar** — the nav joins the session list inside the same scroller, so it rides up as the list is scrolled. Pinned: the Project switcher (top), New chat (its primary action), the user row (bottom). The list separator moves inside the scroller and stays full-bleed (`-mx-2` undoes the scroller's padding, `px-3` restores the row's own inset), now travelling with the list rather than framing a pinned nav.

**Collapsed rail** — the same shape: the eight icons scroll between the pinned expand control and the account avatar. The scrollbar is hidden (`no-scrollbar`); at 48px wide a visible one would cost a third of the rail. The DOM order is unchanged, so the rail-ordering test still finds all eight entries in the same `nav`.

## Verification

- Measured in a real browser at 900 / 600 / 420 / 320 / 240px tall, both sidebar states: document overflow `0` everywhere (it was `+92` at 320 before).
- Screenshots at 900px show the sidebar unchanged — same separator position, same spacing.
- The `layout.spec.mjs` document sweep now asserts both states at 420 / 320 / 240, and the note explaining why it could not go below 420 is deleted along with the floor.
- `pnpm typecheck`, `pnpm test` (1686 passed / 2 skipped), `pnpm format:check` — green.

https://claude.ai/code/session_01XUMcjn2UJ7SSqjhEyvCeN6

### DIFF
--- a/packages/web/e2e/layout.spec.mjs
+++ b/packages/web/e2e/layout.spec.mjs
@@ -16,9 +16,11 @@
  *   search box then horizontally scrolling the whole draft page — and the workspace menu
  *   ~143px off-screen right when the ownership pills share one row);
  * - no page grows the **document**: the app shell is height-constrained and each page scrolls
- *   inside its own container, so a second scrollbar means an absolutely positioned descendant
- *   escaped its scroller (the Traces tree and the Agent settings page both had one, visible
- *   only with a second Agent below a long list);
+ *   inside its own container, so a second scrollbar means either an absolutely positioned
+ *   descendant escaped its scroller (the Traces tree and the Agent settings page both had one,
+ *   visible only with a second Agent below a long list) or something that cannot shrink no
+ *   longer fits — checked at 420/320/240px tall in both sidebar states, since the sidebar's
+ *   chrome used to stop fitting below ~412px;
  * - the sidebar's "New chat" button has no background fill (same gray-scale style as nav items);
  * - the collapsed rail shows, in product-specified order, last conversation / new chat /
  *   Agents / Skills / Models / Costs / Traces / Benchmark with localized (en + zh) hover
@@ -704,22 +706,42 @@ test("layout: no page grows the document (absolute descendants stay in their scr
 
   // A short viewport puts the second Agent's node below the fold with a handful of Sessions
   // instead of dozens — the same geometry a full-height window reaches with a longer list.
-  // Not shorter than 420: below ~412px the sidebar's own fixed chrome (project switcher, the
-  // eight nav entries, the user row) no longer fits, which grows the document for its own
-  // unrelated reason and would mask what this test is after.
   await page.setViewportSize({ width: 1440, height: 420 });
   const paths = ["/traces", "/chat", "/agents", "/agents/default_agent", "/skills", "/models"];
+  const grewBy = (p) =>
+    p.evaluate(() => {
+      const de = document.documentElement;
+      return de.scrollHeight - de.clientHeight;
+    });
   const overflowing = [];
   for (const path of paths) {
     await page.goto(`${BASE}${path}`);
     await page.waitForTimeout(1200);
-    const grew = await page.evaluate(() => {
-      const de = document.documentElement;
-      return de.scrollHeight - de.clientHeight;
-    });
+    const grew = await grewBy(page);
     if (grew > 0) overflowing.push(`${path} (+${grew}px)`);
   }
   expect(overflowing, "pages whose document scrolls").toEqual([]);
+
+  // The other way to grow the document: content that cannot shrink. The sidebar's own chrome
+  // (Project switcher, New chat, eight nav entries, user row) used to be fixed height and
+  // stopped fitting below ~412px — a window that short is reachable by browser zoom or docked
+  // devtools. The nav now scrolls with the session list, and the collapsed rail scrolls its
+  // icons the same way, so both states shrink to nothing instead of pushing the page out.
+  const railToggle = page.getByRole("button", { name: "收起侧栏" });
+  for (const height of [420, 320, 240]) {
+    await page.setViewportSize({ width: 1440, height });
+    await page.goto(`${BASE}/chat`);
+    await page.waitForTimeout(900);
+    expect(await grewBy(page), `pinned sidebar @${height}`).toBe(0);
+  }
+  await page.setViewportSize({ width: 1440, height: 900 });
+  await page.goto(`${BASE}/chat`);
+  await railToggle.click();
+  for (const height of [420, 320, 240]) {
+    await page.setViewportSize({ width: 1440, height });
+    await page.waitForTimeout(700);
+    expect(await grewBy(page), `collapsed rail @${height}`).toBe(0);
+  }
 });
 
 test("layout: login — blank start, non-crossing traces, lang/theme controls", async ({ page }) => {
--- a/packages/web/src/components/layout/app-layout.tsx
+++ b/packages/web/src/components/layout/app-layout.tsx
@@ -77,11 +77,16 @@ function CollapsedRail({ onExpand }: { onExpand: () => void }) {
         title={S.nav.expandSidebar}
         aria-label={S.nav.expandSidebar}
         onClick={onExpand}
-        className="flex h-8 w-8 items-center justify-center rounded-md text-gray-500 transition-colors duration-150 hover:bg-gray-200/70 hover:text-gray-800 dark:text-gray-400 dark:hover:bg-gray-800 dark:hover:text-gray-200"
+        className="flex h-8 w-8 shrink-0 items-center justify-center rounded-md text-gray-500 transition-colors duration-150 hover:bg-gray-200/70 hover:text-gray-800 dark:text-gray-400 dark:hover:bg-gray-800 dark:hover:text-gray-200"
       >
         <GlyphIcon d="M9 6l6 6-6 6M20 4v16" size={18} />
       </button>
-      <nav className="mt-1 flex flex-col items-center gap-1">
+      {/* The entries scroll as one block, like the pinned sidebar's nav + session list: the rail
+          keeps only the expand control and the account avatar at fixed height, so a window too
+          short for eight icons scrolls them here instead of pushing them out of the rail and
+          growing the document. Scrollbar hidden — at 48px wide it would cost a third of the
+          rail's width. */}
+      <nav className="no-scrollbar mt-1 flex min-h-0 flex-1 flex-col items-center gap-1 overflow-y-auto">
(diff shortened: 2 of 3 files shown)


hiyouga:
## Review

Reviewed in a clean detached worktree at `3a8e833`, scoped to `origin/main` (`4681135`) — the three files, one commit. The diagnosis is right and the pinned-sidebar half of the fix is correct and minimal: the `-mx-2`/`px-3` full-bleed trick preserves the old separator insets to the pixel and, contrary to what I expected, adds no horizontal overflow (`scrollWidth === clientWidth`, 255/255, at 900/420/320/240). But the PR breaks a pre-existing assertion in the very spec file it edits, the collapsed-rail half doesn't actually scroll at the heights the new test asserts, and two claims in the body are wrong.

### 1. `layout.spec.mjs:255` now matches zero elements — blocker

```js
const newChat = page.locator("nav").getByRole("button", { name: "New chat" });
```

This PR moves New chat out of `<nav>` into a plain `<div className="shrink-0 px-2 pt-2">` (`sidebar.tsx:780`), leaving the `<nav>` at `:808` containing only `NavLink`s — `<a>` elements, no buttons at all.

There are exactly two `<nav>` elements in the app (`sidebar.tsx:808`, `app-layout.tsx:89`), and the second only renders inside `CollapsedRail`. The test containing line 255 never collapses the sidebar (the only `Collapse sidebar` clicks in the file are at `:279` and the new `:739`, both in other tests, and Playwright's `page` fixture gives each test a fresh context). `Drawer` returns `null` when closed (`drawer.tsx:28`), so no second `Sidebar` is mounted either.

So the locator resolves to 0 elements, and `await expect(newChat).toBeVisible()` retries for the full 15s `expect` timeout and fails. Confirmed by reading the structure, not by an e2e run — but note the verification list in the body is `typecheck` / `test` / `format:check` with **no e2e run at all**, so the spec file this PR edits was never executed.

Fix: scope to the sidebar rather than to `nav` — `page.getByRole("complementary").getByRole("button", { name: "New chat" })`, the same handle `paging.spec.mjs:60` already uses.

### 2. The collapsed rail squashes instead of scrolling — major

`app-layout.tsx:24-29` (and the disabled variant at `:102`):

```ts
`flex h-8 w-8 items-center justify-center rounded-md …`
```

No `shrink-0`. The new nav (`app-layout.tsx:89`) is a column flex container (`min-h-0 flex-1 overflow-y-auto`), so its eight children default to `flex-shrink: 1` on the main axis and compress to min-content **before the scroller ever engages**. This PR added `shrink-0` to the two *pinned* rail controls (expand button `:80`, avatar `:135`) but not to the eight items it moved into the scroller.

Measured in headless Chromium on a structural replica using the exact computed values from `railItemClass`, at the app's default font scale (`theme.tsx:28,95` sets root `font-size: 18px`, so `h-8` is 36px, not 32px):

| viewport height | item height (nominal 36px) | scrolls? |
|---|---|---|
| 900 | 36px | no |
| 470 | 36px | no |
| **420** *(tested)* | **35.1px** | no |
| 380 | 30.1px | no |
| **320** *(tested)* | **22.6px** | **no** |
| 240 | 22px | yes |

At 1440x320 collapsed, the icons render 22.6px tall while staying 36px wide — a 1.6:1 distortion, with squashed `rounded-md` hover/active fills — and nothing scrolls. Squashing begins just under 470px at the default scale (≈520px at `lg`).

The new test still reports `collapsed rail @420` and `@320` as passing, because squashing *also* yields document overflow 0. Those two assertions therefore pass vacuously with respect to the behaviour they claim to prove. Adding `shrink-0` keeps overflow at 0 and makes the mechanism the intended one.

### 3. The rail's focus ring is now clipped — minor

`app-layout.tsx:89`. The nav is `align-self: center` inside `items-center`, so its content box is shrink-to-fit — exactly the width of its `w-8` items, zero padding, at any font scale. `overflow-y: auto` makes `overflow-x` compute to `auto` too (this codebase documents that itself at `dropdown.tsx:8-11`), so the global focus ring (`styles.css:120-126`: `outline: 3px solid; outline-offset: 2px`), which lies entirely outside the border box, is clipped.

Probe at a 16px root: nav padding box `263.5 → 295.5`, ring extent `258.5 → 300.5`, `clippedHorizontally: true`. Keyboard user tabs onto "Skills" → the left and right strokes are gone; for the first and last items even top and bottom are cut. Before this PR the nav had no overflow and the ring was fully visible. Fix: horizontal padding on the nav (and widen `w-12`), or a `ring-*`/`box-shadow` focus style, which paints inside.

### 4. "Screenshots at 900px show the sidebar unchanged" is wrong — minor

Before, New chat and the nav items were siblings inside one `space-y-0.5` nav → **2px** gap. After, New chat sits in `div.px-2.pt-2` with no bottom padding, the scroller has no top padding, and the nav carries `pt-2` → **8px** gap (`sidebar.tsx:780` / `:807` / `:808`). This applies at every viewport height. There are no screenshot baselines in the repo (`grep toHaveScreenshot` → nothing), so that claim was a manual eyeball and it is wrong on this detail. Measured from the Tailwind values, not screenshotted.

### 5. Stale module header — minor

`sidebar.tsx:3` still reads `… new chat (default_agent draft) + fixed nav (Agents / models / cost center / Trace) …`. The nav is exactly what stopped being fixed. The two inline block comments and the spec-file header were updated; the module header was missed.

### 6. Test hygiene in the new block — nit

- `layout.spec.mjs:739` clicks the collapse toggle with no assertion that the rail actually collapsed. If that click ever silently no-ops, the following three assertions re-test the pinned state and pass. `expect(page.locator("aside")).toHaveClass(/w-12/)` closes it.
- `layout.spec.mjs:665`: the test title ("absolute descendants stay in their scroller") no longer describes half of what the test asserts. The header comment at `:18-23` was updated; the title was not.

### Design-spec compliance — passes, with one gap

Against the invariant added in design `92d98d7` (`06-PROTOTYPE.md:11`, landed with #116):

- **Containing block, both new scrollers: satisfied.** The pinned scroller (`sidebar.tsx:807`) carries explicit `relative` and keeps the #116 comment explaining why. The rail scroller (`app-layout.tsx:89`) has no explicit `relative` and relies solely on the base rule `.overflow-y-auto { position: relative }` (`styles.css:86-101`) — which is compliant, since that rule is the project's chosen enforcement mechanism and nothing on the element out-ranks it. Minor inconsistency: explicit on one new scroller, implicit on the other.
- **Visually-hidden-but-focusable controls inside the new scrollers: none.** The only `sr-only` in `components/layout/` is `sidebar.tsx:1313`, a `<span>`, not focusable — the spec's second clause doesn't bite, and the first is satisfied since it sits inside the `relative` scroller. No file inputs anywhere in the sidebar or rail.
- **"A second scrollbar is a defect": satisfied by arithmetic, not by construction.** The irreducible pinned chrome drops from ~412px to ~137px at a 16px root (~154px at the default 18px, ~171px at `lg`), the rail's to ~92px — genuinely below any window a browser can produce. But nothing clips the shell: `app-layout.tsx:159` is still `<div className="flex h-full">` with no `overflow-hidden`, and the `<aside>` at `:161` has no overflow either. Any future pinned element puts the document straight back into overflow, and the only guard is a fixed list of viewport heights in one e2e test. `overflow-hidden` on the shell root or the `<aside>` would make the invariant hold by construction.

**Not recorded in the specs.** Moving the primary navigation from pinned to scrolling is a product change — at every viewport height, scrolling to an older session now hides the nav. `06-PROTOTYPE.md:17-25` enumerates the nav entries and `:27` fixes the collapsed-rail order, but nothing says whether the nav is pinned or scrolls. Per CLAUDE.md this belongs in `06-PROTOTYPE.md` under 导航结构. (The pattern matches ChatGPT/Claude sidebars, so the change itself is defensible — it just isn't written down, and `sidebar.tsx:3` still asserts the opposite.)

### Hypotheses I checked and could not land

Recording these so they don't get re-litigated:

- **Scroll-handler / paging regression** — no `onScroll`, `scrollTop`, `scrollIntoView`, `IntersectionObserver` or scroll-position restore anywhere in `components/layout/`. Paging is a "More" button (`sidebar.tsx:600,647`), not scroll-driven. Nothing observes the scroller, so moving its boundary cannot break a handler. This was my highest-value hypothesis and it does not apply.
- **`no-scrollbar` missing or WebKit-only** — defined at `styles.css:149-155` with `scrollbar-width`, `-ms-overflow-style` and `::-webkit-scrollbar`. Fine on Firefox.
- **Overlays clipped by the new boundary** — both Sidebar `Dropdown`s (`:713`, `:1007`) render inline-`absolute` without `portal`, but both sit outside the scroller in the pinned top and bottom blocks. The only things moved in are six `NavLink`s with no out-of-flow descendants. Rail tooltips are native `title`. No `sticky` anywhere.
- **`app-layout.tsx` regressing other routes** — the change is entirely inside `CollapsedRail`; `<main className="min-h-0 flex-1 overflow-hidden">` (`:212`) and the rest of the shell are untouched.
- **The new sweep being vacuous for the pinned state** — it is a genuine regression test there: the old chrome was ~412px, so 320 and 240 would have failed on the true base, and it measures `documentElement.scrollHeight - clientHeight`, which is the right thing. The zh literal `"收起侧栏"` is correct — this test doesn't `addInitScript` a locale, so it inherits `locale: "zh-CN"` from `playwright.config.mjs:13`.

### Verification I could and couldn't reproduce

`prettier --check` passes on all three files. `pnpm typecheck` and `pnpm --filter @prismshadow/penguin-web test` fail in this worktree only on a stale `@prismshadow/penguin-core` dist (`markers`, `parseUserSteeringText`, `StopReason`, `retry_in_ms`) — **no error in any file this PR touches**, and no web unit test renders a component, so neither is informative either way. A green run needs a workspace build I skipped on this shared box. The e2e claim in §1 is the one that matters and it is not backed by a run in the body.

PR #120, opened 2026-07-29
feat(core): tighten the default system prompt

hiyouga:
The built-in template (`DEFAULT_SYSTEM_PROMPT`) is read by the model on every turn, and had grown wordy in exactly the sections that repeat most. This trims them without dropping a rule, and adds two behaviours it was missing.

## What changed

| Section | Change |
| --- | --- |
| Personality | Now pins the reply language to the user's own — the tool schema already demands that of every call `description`, so the two agreed in practice but only one of them said it. |
| Constraints | The process/port bullet collapses from two sentences to one, with the same guarantees (don't kill what you didn't start, PenguinHarness services included; don't take a service port; pick another free port when yours is busy). The API-retry bullet moves out entirely. |
| Stop rules | API auth/key handling is stated **once**, as the special case of "an error you cannot resolve" instead of a rule of its own: retry at most once, then stop calling tools and send the user outside the chat to update the key. Same facts (no secret in the conversation, a new key only applies to the next conversation), one bullet instead of two spread across two sections. |
| Tool use | The internet bullet loses its redundant half. |
| System markers | Rewritten tight — roughly half the words, same four markers and same instructions. |
| File system | Eight bullets to six: the workspace-relative-path convention merges into the `CWD` bullet, another agent's state merges into your own. |
| Suggested workflows | The parallel-subagent recommendation was already there; what is new is saying that it is the fast path through a large task. Its Playwright/curl line folds into Tool use, which now carries both halves the two lines used to hold separately (prefer Playwright when installed, otherwise `curl`). |
| Skills | Drops the "There is no skill tool" filler, and gains a tooling convention: interpreter and tool environments (virtualenvs, pipx, caches) install **once** into the shared `<app_data_dir>/agents/<agent_id>/shared_env/`, one subdirectory each, and are reused across sessions; an environment already in the current directory wins. A project's own dependencies are the opposite case and stay in the project — Node resolves `node_modules` from there — with pnpm preferred so repeated installs share one store. |

Net effect on the template: same rule set, **−21.7%** words across the six rewritten sections (829 → 649; `# Stop rules` itself grew, absorbing the Constraints bullet), plus the two new behaviours (reply language, shared tooling directory).

## Existing agents

No migration and no compatibility shim: an Agent always runs with its on-disk `agent_state/system_config.yaml` verbatim, so this only reaches **newly created** Agents, and existing ones whose owner runs *Restore default configuration* on the settings page (already documented in [Configuration](https://penguin.ooo/docs/configuration)).

`shared_env/` is a prompt-level convention, not a code-created path — `paths.ts` is untouched, and Agent State snapshots still package `agent_state/` only, so a virtualenv can never bloat an export. The data-layout tree in the docs gains the directory with that caveat spelled out (EN + ZH).

## Review follow-up (second commit)

- **`shared_env/` is scoped to tooling.** As first written the rule sent npm installs there too, which cannot work — Node resolves `node_modules` from the project — and its escape hatch ("unless the current directory already has an environment") does not fire for a *fresh* project that has `package.json` and no `node_modules` yet, so a new app's very first install would have gone somewhere vite could not resolve. Project dependencies now explicitly stay in the project, with pnpm preferred there; the shared directory keeps virtualenvs, pipx tools and caches, one subdirectory per environment (which also settles the container-vs-venv ambiguity in the old `env/` name).
- **Two rules the trim dropped rather than shortened are back**: the Playwright *preference* (only the curl-first use-case split had survived, so an ordinary page went to `curl` where it used to go to Playwright), and `[user_steering]`'s "incorporate it **immediately**".
- **The reply-language rule says "reply"** and carves out code, identifiers and commit messages — this repo's own rule is English-only code.
- **The App Data Dir bullet gets its explicit "never leave deliverables there" back**, now that the prompt invites creating a directory inside it, and the scratchpad is named where its path is given so "this Session's scratchpad" resolves verbatim.
- **The vacuous assertion is fixed**: `toContain("# Stop rules")` passes on `main` too, so it is now an ordering assertion that fails if the retry rule moves back into Constraints.

Still outstanding and deliberately not in this PR: `shared_env/` is not yet in `design/specs/` — that is a design-repo change, per the project's own split.

## Verification

```
pnpm -r build     # ok
pnpm typecheck    # ok
pnpm test         # ok — 1736 tests passed across the 7 packages (2 skipped)
pnpm format && pnpm format:check   # clean
```

The prompt assertions in `packages/core/test/state.test.ts` still pass unchanged (port guardrails, `retry at most once`, `stop calling tools`, the marker names, the section order); two new assertions cover the reply-language line and the shared `env/` directory, and one stale comment about which section held the API rule is corrected.

### DIFF
--- a/packages/core/src/state/default-config.ts
+++ b/packages/core/src/state/default-config.ts
@@ -117,9 +117,10 @@ Some messages carry system-synthesized \`[tag]...[/tag]\` blocks — not user te
 # File system
 - Angle-bracket names such as \`<app_data_dir>\`, \`<agent_id>\` and \`<session_id>\` are placeholders — substitute the values from the Environment section.
 - You work inside the user's folder (\`CWD\` in Environment). When you create or update a file there, mention its workspace-relative path in backticks (e.g. \`src/app.py\`) so the user can open it from the message.
-- The App Data Dir is PenguinHarness's data root: every agent's files plus the project-level data. It is NOT the task's directory and was not provided by the user — never treat its contents as task input, and never leave task deliverables there.
+- The App Data Dir is PenguinHarness's data root: every agent's files plus the project-level data. Its contents were not provided by the user — never treat them as task input. \`CWD\` may itself be a temporary Workspace inside this tree: that one folder is the task's, the rest of it is not.
 - Your Agent State is \`<app_data_dir>/agents/<agent_id>/agent_state/\`; it holds assets such as \`skills/\`, and its \`AGENTS.md\` is already in your context. Another agent's is the same path under its id. Reach them directly.
 - Keep intermediates in this Session's scratchpad, \`<app_data_dir>/agents/<agent_id>/scratchpad/<session_id>/\`, but always place final deliverables in the workspace (under \`CWD\`) — files left in the scratchpad are not part of your output.
+- Install tooling once, not per task: keep interpreter and tool environments — Python virtualenvs, pipx tools, model and package caches — under the shared \`<app_data_dir>/agents/<agent_id>/shared_env/\`, one subdirectory per environment (create them as needed), and reuse them across sessions; an environment already present in the current directory always wins. A project's own dependencies are not tooling: they have to resolve from the project, so install them inside the project directory itself, and prefer pnpm there — its shared store keeps repeated installs from duplicating on disk.
 - Never read, copy or print \`.project_config.toml\` under the App Data Dir, or any agent's \`agent_state/.vault.toml\` — they hold the user's secrets. Configuration is CLI-only (\`penguin config ...\`); if a task seems to need them, say so and ask the user instead.
 
 # Suggested workflows
@@ -140,7 +141,6 @@ The vault holds this agent's per-agent secrets (agent_state/.vault.toml). Each e
 
 # Skills
 Skills are reusable instruction packages at <app_data_dir>/agents/<agent_id>/agent_state/skills/<skill_name>/SKILL.md. When a task matches one below, or the user asks for one (the message may start with a [use_skills] block naming them), read that SKILL.md in full with read_file, then follow it. If a request names a skill without a concrete task, ask the user what they need first.
-Install a skill's tooling once, not per task: keep interpreter and tool environments — Python virtualenvs, pipx tools, model and package caches — under the shared <app_data_dir>/agents/<agent_id>/shared_env/, one subdirectory per environment (create them as needed), and reuse them across sessions; an environment already present in the current directory always wins. A project's own dependencies are not tooling: they have to resolve from the project, so install them inside the project directory itself, and prefer pnpm there — its shared store keeps repeated installs from duplicating on disk.
 {{SKILL_METADATA}}
 
 # Environment
--- a/packages/core/src/state/default-config.ts
+++ b/packages/core/src/state/default-config.ts
@@ -115,13 +115,13 @@ Some messages carry system-synthesized \`[tag]...[/tag]\` blocks — not user te
 - \`[user_steering]\`: a user message sent mid-run, delivered between turns. Not a new task: incorporate it immediately and adjust course within the current one.
 
 # File system
-- Angle-bracket names such as \`<app_data_dir>\`, \`<agent_id>\` and \`<session_id>\` are placeholders — substitute the values from the Environment section.
-- You work inside the user's folder (\`CWD\` in Environment). When you create or update a file there, mention its workspace-relative path in backticks (e.g. \`src/app.py\`) so the user can open it from the message.
-- The App Data Dir is PenguinHarness's data root: every agent's files plus the project-level data. Its contents were not provided by the user — never treat them as task input. \`CWD\` may itself be a temporary Workspace inside this tree: that one folder is the task's, the rest of it is not.
-- Your Agent State is \`<app_data_dir>/agents/<agent_id>/agent_state/\`; it holds assets such as \`skills/\`, and its \`AGENTS.md\` is already in your context. Another agent's is the same path under its id. Reach them directly.
-- Keep intermediates in this Session's scratchpad, \`<app_data_dir>/agents/<agent_id>/scratchpad/<session_id>/\`, but always place final deliverables in the workspace (under \`CWD\`) — files left in the scratchpad are not part of your output.
-- Install tooling once, not per task: keep interpreter and tool environments — Python virtualenvs, pipx tools, model and package caches — under the shared \`<app_data_dir>/agents/<agent_id>/shared_env/\`, one subdirectory per environment (create them as needed), and reuse them across sessions; an environment already present in the current directory always wins. A project's own dependencies are not tooling: they have to resolve from the project, so install them inside the project directory itself, and prefer pnpm there — its shared store keeps repeated installs from duplicating on disk.
-- Never read, copy or print \`.project_config.toml\` under the App Data Dir, or any agent's \`agent_state/.vault.toml\` — they hold the user's secrets. Configuration is CLI-only (\`penguin config ...\`); if a task seems to need them, say so and ask the user instead.
+- Angle-bracket names such as \`<app_data_dir>\` and \`<session_id>\` are placeholders — substitute the values from the Environment section.
+- You work inside the user's folder (\`CWD\`). For each file you create or update there, mention its workspace-relative path in backticks (e.g. \`src/app.py\`) so the user can open it.
+- The App Data Dir is PenguinHarness's data root — every agent's files and the project-level data, none of it supplied by the user, so never treat it as task input. \`CWD\` may itself be a temporary Workspace inside it: that one folder is the task's, the rest is not.
+- Your Agent State is \`<app_data_dir>/agents/<agent_id>/agent_state/\`; it holds \`skills/\`, and its \`AGENTS.md\` is already in your context. Another agent's is the same path under its id — reach it directly.
+- Keep intermediates in this Session's scratchpad, \`<app_data_dir>/agents/<agent_id>/scratchpad/<session_id>/\`, but always place final deliverables in the workspace (under \`CWD\`) — what stays in the scratchpad is not part of your output.
+- Install into the project's own environment when it has one. Otherwise keep reusable ones — Python virtualenvs, model and package caches — under \`<app_data_dir>/agents/<agent_id>/shared_env/<name>/\` and reuse them across Sessions. For Node, prefer pnpm in the project itself: its shared store keeps repeated installs from duplicating on disk.
+- Never read, copy or print \`.project_config.toml\` or any agent's \`agent_state/.vault.toml\` — they hold the user's secrets. Configuration is CLI-only (\`penguin config ...\`); if a task seems to need them, say so and ask the user instead.
 
 # Suggested workflows
 Recommendations, not requirements — adapt them to the task.
--- a/packages/core/src/state/default-config.ts
+++ b/packages/core/src/state/default-config.ts
@@ -86,7 +86,7 @@ const DEFAULT_SYSTEM_PROMPT = `# Role
 You are PenguinHarness, an agent that completes the user's requests on their machine with the tools available to you.
 
 # Personality
-Communicate with the user precisely and concisely, yet with warmth. Do not repeatedly explain your tools or restate their results.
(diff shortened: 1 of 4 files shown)


packages/core/src/state/default-config.ts
```diff
@@ -143,7 +139,8 @@ The vault holds this agent's per-agent secrets (agent_state/.vault.toml). Each e
...
 
 # Skills
-Skills are reusable instruction packages stored under <app_data_dir>/agents/<agent_id>/agent_state/skills/<skill_name>/SKILL.md. There is no skill tool: when a task matches an installed skill below, or the user asks to use one (a message may start with a [use_skills] block listing skill names), first read that skill's SKILL.md in full with the read_file tool, then follow it. If a request only names a skill without a concrete task, ask the user what they need before starting.
+Skills are reusable instruction packages at <app_data_dir>/agents/<agent_id>/agent_state/skills/<skill_name>/SKILL.md. When a task matches one below, or the user asks for one (the message may start with a [use_skills] block naming them), read that SKILL.md in full with read_file, then follow it. If a request names a skill without a concrete task, ask the user what they need first.
+Install a skill's dependencies once, not per task: keep Python virtualenvs, npm installs and other reusable environments in the shared <app_data_dir>/agents/<agent_id>/env/ directory (create it if missing) and reuse them across sessions — unless the current directory already has its own environment, which always wins.
```
hiyouga: 这个不是针对skill说的，而是针对全部任务，所以应该放到文件系统部分。<app_data_dir>/agents/<agent_id>/shared_env/

hiyouga:
## Review

Reviewed in a clean detached worktree at `f13f1ce`, scoped to `origin/main` (`4681135`) — the four files, one commit. I read the prompt diff as specification text and tracked every removed or merged clause: **44 of 49 rules survive intact**, and the riskiest claims in the body are the ones that hold up best (see the last section). But two rules are genuinely gone rather than trimmed, the new `env/` convention is wrong for npm as written, and it isn't in the specs.

### 1. The `env/` rule will send project installs somewhere Node can't resolve — major

`default-config.ts:143`:

> Install a skill's dependencies once, not per task: keep Python virtualenvs, **npm installs** and other reusable environments in the shared `<app_data_dir>/agents/<agent_id>/env/` directory (create it if missing) and reuse them across sessions — unless **the current directory already has its own environment**, which always wins.

Two problems compound:

1. **Node resolution is project-local.** `node_modules` must sit at or above the importing file. Anything installed under `<app_data_dir>/agents/<id>/env/` is unreachable from code in `CWD`, and `NODE_PATH` does not work for ESM. This is unlike a virtualenv, which is location-independent by design — the rule generalizes from a case where it works to one where it can't.
2. **The escape hatch doesn't fire in the common case.** It's keyed on the current directory *already having* an environment. A fresh JS project has `package.json` and **no** `node_modules` yet — so by the letter of the rule, the very first `npm install` of a new project goes to the shared directory.

The opening clause scopes to "a skill's dependencies", but the rest generalizes ("npm installs and other reusable environments"), and the exception clause implies the rule would otherwise cover the current directory's own deps — otherwise why need the exception? The model reads this every turn.

Concretely: user asks for a React app in an empty `CWD` → model runs `npm install --prefix <app_data_dir>/agents/default_agent/env react` → `vite` can't resolve `react` → turns burned debugging its own install location.

Suggest restricting the rule to interpreter/tool environments (virtualenvs, pipx, model caches) and saying explicitly that project package installs stay in the project.

Two smaller things about the same line:
- **`env/` is ambiguous as container vs. venv.** `python -m venv env` conventionally *creates* a venv named `env`. "Keep Python virtualenvs … in the shared `.../env/` directory" reads both ways, and no naming scheme (e.g. `env/<skill_name>/`) is given. If the model makes `env/` itself a single venv, the next skill needing a different Python version has nowhere to go.
- **It races against the parallelism this same PR emphasises.** `:128` now says dispatching independent subtasks in parallel "is the fastest way through a large task"; `run_subagent`'s `agent_id` defaults to the current agent (`:357-360`), so parallel subagents resolve to the same `<agent_id>` and therefore the same `env/`. No locking, no per-skill subdirectory. Two concurrent `pip install`s into one venv can leave a half-written `site-packages`/`RECORD` that fails on import for every later session, with no GC path to repair it.

### 2. "Without dropping a rule" — two were dropped, one weakened

**Lost: the Playwright preference** (`:129`). `main` had two lines, and they were not the same rule twice:

- Tool use: "`curl` for pages and APIs, or Playwright (if installed) for dynamic sites" — a **use-case split**.
- Suggested workflows: "To visit web pages, **prefer Playwright when installed; otherwise `curl`**" — a **preference order**.

The second is deleted as "the duplicated Playwright/curl line … Tool use already carries it". It doesn't: only the curl-first split survives, so for an ordinary web page with Playwright installed the model now reaches for `curl` where it previously preferred Playwright. The two lines were genuinely in tension, so resolving them is defensible — it just isn't a dedup, and the body says no rule was dropped.

**Lost: `[user_steering]` immediacy** (`:115`). Old: "It is not a new task: **incorporate it immediately and** adjust course within the current task." New: "Not a new task: adjust course within the current one." User sends "actually, skip the tests" mid-run; nothing now says *immediately*, so finishing the in-flight tool batch first is compliant.

**Weakened: deliverables in the App Data Dir** (`:120`). The bullet dropped "and never place task deliverables there; the working folder is `CWD`". The guarantee survives via the positive blanket statement at `:122` ("always place final deliverables in the workspace (under `CWD`)"), so it is covered — but the explicit prohibition on the App Data Dir is now only inferable, in the same PR that invites the model to create a directory inside it.

Rationale-only losses, flagged but not defects: "so further retries cannot succeed" (`:104`), "which are none of your business" (`:123`), "to keep execution consistent" (`:127`). Qualifier trims: "unless the user **explicitly** asks" → "unless the user asks" (`:98`), "instead of killing the listener" (`:98`), "delivered between turns **alongside tool results**" (`:115`).

One narrowing I checked and it's fine: `:123` drops "or otherwise access" from the secrets bullet, so `grep`ping or `source`ing `.vault.toml` is no longer literally covered. `design/specs/05-ARCHITECTURE.md:281` specifies exactly 读取/复制/打印 — the new text matches the spec and the old text exceeded it. Flagging only so the narrowing is a conscious choice.

### 3. `env/` is not in `design/specs/` — major (project rule)

`design/specs/01-PRINCIPLES.md:55-90` holds the authoritative per-agent directory tree, and `:96-107` enumerates every per-agent path bullet by bullet (`agent_state/`, `benchmarks/`, `snapshots/`, `traces/`, `scratchpad/`, `workspaces/`). Neither gained `env/`, and the design repo tree is clean — so this isn't "lands separately", it's outstanding.

CLAUDE.md: "any code change that affects architecture, behavior, or interfaces must update the corresponding docs under `design/specs/`." `packages/docs/` now documents a seventh per-agent directory that the ground-truth spec doesn't contain. The next person reconciling code against the spec deletes the docs entry, or a future spec-driven change to the agent layout silently omits `env/`. (Precedent for doing it: design `92d98d7` landed the same day as impl #116.)

### 4. One of the two new test assertions is vacuous — minor

`state.test.ts:498-502`. The comment says "Auth/key failures live **entirely** in Stop rules", but the assertion is `expect(prompt).toContain("# Stop rules")` — a heading already present on `main`. Nothing pins that "retry at most once" is *inside* Stop rules, nor that it's *absent* from Constraints, so the exact revert it guards against passes green.

```ts
expect(prompt.indexOf("# Stop rules")).toBeLessThan(prompt.indexOf("retry at most once"));
expect(prompt.indexOf("retry at most once")).toBeLessThan(prompt.indexOf("# Tool use"));
```

The other new assertion is fine — neither `"in the user's language"` nor the `env/` path appears on `main`, so both genuinely pin new behaviour.

### 5. "Always answer in the user's language" has no artifact carve-out — minor

`:89`. Nothing in the prompt scopes this to prose. The only place in the repo that draws the line is a skill — `packages/skills/skills/web-design/SKILL.md:64`, "Code identifiers, CSS class names and code comments stay English" — and only for pages it generates. A Chinese-speaking user can now get Chinese variable names, commit messages, or log strings. Directly relevant here: this repo's own CLAUDE.md mandates English-only code, comments and tests.

Fix: "always **reply** in the user's language", or add the carve-out `web-design` already models.

Related, and worth knowing the claim is stronger than stated: the tool-schema precedent is real but narrower than the body implies. The string "…written in the user's language" is byte-identical on four call-description tools (`:280, 314, 350, 396`), each `required` — but that's 4 of 9 tools, it says nothing about the model's prose reply, and `applyCallDescriptionToggle` (`agent-state.ts:514-527`) can strip it. So `:89` is a genuine expansion, not a restatement of something already true.

### 6. Two body claims to correct

- **"Suggested workflows says out loud that dispatching to subagents in parallel is the fast path"** — `main` already said "dispatch independent subtasks in parallel". Only the rationale clause ("that is the fastest way through a large task") is new. The commit message is accurate; the PR summary overstates.
- **"~25% fewer words in the per-turn sections"** — measured: Role→Suggested workflows 1019→846 = **−17.0%**; the six rewritten sections 829→649 = **−21.7%**. Also worth noting `# Stop rules` *grew* (106→109 words) absorbing the Constraints bullet, while `# Constraints` shrank 101→72.

### Nits

- `:104` — the auth/key bullet lost "so further retries cannot succeed". Mandatory "stop calling tools" survives, so the rule stands; the reason a model would otherwise rationalise around it does not.
- `:127` — the literal `PLAN.md` path is replaced by "this Session's scratchpad", a phrase never defined verbatim; `:122` gives the path without naming it that. One extra resolution hop.
- `:10-12` — the header comment listing the sections still omits `# Tool use`, `# System markers`, `# Vault`, `# Skills`. Pre-existing, but this PR rewrote two of the omitted sections.
- `:121` — merging another agent's state into your own drops the explicit `<its_agent_id>` template; the model now derives it. The merge grants nothing new — the secrets bullet still forbids "any agent's `agent_state/.vault.toml`".

### Claims that held up

All of these were the ones most likely to be wrong, and all check out:

- **`env/` can never bloat an export** — `snapshot-service.ts:69-78` passes the literal list `["agent_state"]` to `tar.create`. No glob, no walk. Import replaces `agent_state/` by rename (`:166-184`).
- **Nothing else walks, watches, sizes or mis-deletes the agent dir** — no `readdir(agentDir(...))` anywhere in the repo; every walk is either one level above (listing agent ids) or jumps straight into a named `paths.ts` subdir. Zero hits for chokidar/`fs.watch`/fast-glob/globby/recursive `readdir`. Deletion is `fs.rm(agentDir(...), {recursive:true})` (`agent-service.ts:192`), so `env/` is reclaimed correctly. No exclusion list needs updating.
- **The path shape is real** — `paths.ts:46-48` `agentDir()` = `<root>/<projectId>/agents/<agentId>`, and `{{PROJECT_DIR}}` is labelled "App Data Dir", so the prompt path resolves to `agentDir() + "/env"` exactly.
- **`<app_data_dir>` is resolvable by the model** — this was my main "is this a live bug?" candidate and it isn't. `assembleSystemPrompt` substitutes only `{{…}}`; angle-bracket tokens reach the model verbatim by design (`:15-18`), the Environment section renders `App Data Dir: <concrete path>` + `Agent ID` (`agent-state.ts:444-480`), and `:118` tells the model to substitute. Same mechanism already load-bearing for `agent_state/`, `skills/` and `scratchpad/`, and specced at `01-PRINCIPLES.md:45`.
- **No migration needed** — `agent-state.ts:130-147` does `systemConfig = parsed as SystemConfig` with no merge against `defaultSystemConfig()`; the init path is the only writer of defaults. Only new agents and "Restore default configuration" pick this up. One caveat worth adding to the body: a parent with the new prompt can delegate to a pre-existing subagent that lacks it.
- **Docs EN/ZH asymmetry (+3 vs +2) is line-wrapping, not a missing caveat** — both convey all three facts (shared deps, created on demand, prompt-level convention not a code-created path). Tree alignment is exact in both: `#` at column 45 on the entry line, 41 on continuations, matching the existing `agent_state/` block; `env/` sits between `scratchpad/` and `workspaces/`, both `├──`, and `└──` still terminates at `snapshots/`.
- **All 13 placeholders intact, no orphans**, section order unchanged, no trailing whitespace, `prettier --check` clean on all four files, and no stale references anywhere in the repo to the ten removed clauses.

`packages/core/test/state.test.ts` passes **54/54**. I skipped `pnpm -r build` and the full suite on this shared box.

PR #121, opened 2026-07-29
feat(web,server,core): attach files to a message from the composer

hiyouga:
> **Stacked on #122** (`/agent` command). This PR's base is `feat/agent-slash-command`, so the diff shown here is the attachment work alone; both PRs edit `chat-input.tsx`, and the conflicts are already resolved in this branch. Merge #122 first — GitHub retargets this one to `main` automatically.

Adds a **file attachment** entry to the composer's "+" menu, directly below the image upload: any file type, several at a time, shown as removable chips (name + size) above the text body. Like images, a message with attachments and no text is sendable.

## Design

**Transport — no new upload endpoint.** A draft chat has no Session id yet, so there is nothing for the composer to upload to before sending. The file rides the existing task-input pipeline exactly like a pasted image: `TaskInputPart` gains

```ts
| { type: "file"; fileName: string; dataUrl: string }
```

a base64 `data:` URL, **≤10MB per file** (413 `file_too_large` beyond that). The server already caps request bodies at 20MB, which a single 10MB file fits after base64 inflation; several attachments in one message must fit that same body cap, so the per-file limit was chosen to fit rather than raising the global one. `parseTaskInput` validates the part synchronously and side-effect free (non-empty `fileName` with no path separators / no `..`, a real `data:…;base64,…` URL, non-empty payload) — nothing touches the filesystem until the whole request is known good.

**Where the bytes land.** Each file is written into the Session scratchpad (`<agent>/scratchpad/<sessionId>/`), which is already deleted with the Session, so cleanup is free. The name is sanitized to the character set the scratchpad read endpoint serves, keeping the extension; the write is an exclusive create (`wx`) with a short random suffix on collision, so a second `report.pdf` lands beside the first instead of overwriting it (the same convention core uses for image uploads).

**Marker.** One `[attached file: <absolute path>]` line per file is appended to the user's text message (a text message is created when the input is files-only). It **trails** the body rather than prefixing it — `[use_skills]` and the origin blocks are parsed at index 0, so a leading block would break that chain. The bytes never enter the conversation: the model opens the file by path with its ordinary file tools.

**One format, one placement rule.** The line format moved into core's markers module (`attachment-lines.ts`) next to its existing `[attached image: …]` sibling, and the placement rule is now the single shared `appendAttachmentLines` helper used by both core's image path and the server's file path. Producer (core / server) and parser (Web) import the same spellings, so the two conventions cannot drift.

**Frontend.** `lib/attachments.ts` now yields images *and* file paths in one pass (`splitAttachments`); image behavior is unchanged. A message carrying files renders an `Attached files: a.pdf, b.csv` banner (paperclip glyph), styled after `skills-banner.tsx`, and the remaining body text renders as usual.

**Consistency across send paths.** `canSend`, the Stop/Send switch, the follow-up queue, the @ handoff and the `/model` switch all treat attachments exactly as they treat images; every successful send clears them along with the images. Mid-run steering stays text-only, so attachments remain in the draft. Goal mode is text-only server-side, so the composer clears and blocks attachments there just like images — and the server rejects a `file` part in goal mode before writing anything.

**Incidental hardening.** The scratchpad read endpoint gained `X-Content-Type-Options: nosniff`: arbitrary user-uploaded bytes now land in that directory and are served from the App's own origin.

## Notes

- Paste still only picks up images (unchanged); files come from the menu entry.
- The composer does not pre-check the 10MB cap — same as images today; an oversize file surfaces the server's `file_too_large` toast, for which the Web already has copy.
- Non-ASCII file names sanitize to `-` (e.g. an all-CJK name becomes `file.ext`); the message banner shows the on-disk name, which is what the model sees.

## Follow-up commit — non-ASCII file names

The first version mapped every character outside `[A-Za-z0-9._-]` to `-`, so `报告.pdf` reached disk (and the model) as `file.pdf`, a second one as `file-3f9a1c.pdf`. The justification was the scratchpad read endpoint, whose traversal guard was that same character whitelist. Both ends changed:

- **the sanitizer** now only rewrites ASCII outside the old whitelist — a space or shell metacharacter in a path the model pastes into a command is a real footgun — and keeps everything above ASCII as typed, minus Unicode category C (invisible controls, and the bidi overrides behind file-name spoofing). Windows hygiene rides along: trailing dots/spaces dropped, device names prefixed (`con.txt` → `_con.txt`), and the stem capped by UTF-8 **bytes** on a character boundary, since a CJK character costs three and a character count would let a Chinese name blow the filesystem's real limit;
- **the read endpoint** drops the whitelist for a structural check (no separators, no control characters, not a relative marker) *confirmed* by resolving the path and requiring its parent to be exactly this session's directory. That containment is strictly stronger than the charset was — it also rejects shapes a character class misses, such as a Windows drive-relative `C:evil.png`.

`报告 2026.pdf` now lands as `报告-2026.pdf`. New tests cover the round trip (including the byte-boundary cut, the device-name prefix, the all-invisible fallback) and the endpoint's rejections.

## Verification

Run from the worktree root with Node 24, in this order:

| Command | Result |
| --- | --- |
| `pnpm -r build` | pass (all 7 packages) |
| `pnpm typecheck` | pass (all packages) |
| `pnpm test` | pass — core 584, server 418, web 477, cli 202, docs 11, skills 16, landing 39 |
| `pnpm format && pnpm format:check` | pass ("All matched files use Prettier code style!") |

New tests: `packages/server/test/task-attachments.test.ts` (7 cases — scratchpad write + marker line, files-only input, same-name collision, name sanitization, media-type parameters, malformed parts → 400 with nothing written, oversize → 413, goal mode → 400) and the extended `packages/web/test/attachments.test.ts` (file lines, mixed image+file, unrecognized lines untouched, plus the pre-existing image cases unchanged).

Docs updated in both languages: `web-app.{en,zh}.md` (composer "+" menu) and `server-api.{en,zh}.md` (`TaskInputPart` and the `/tasks` row).

## Review follow-up

Applied on top of the original commits (three commits: core, server, web). Each finding from the review, and what changed.

**1. `[handoff_from]` corrupted by the trailing line (major).** `appendAttachmentLines` appended to the last user text message, and the origin-block parsers require a *whole-message* match — so a files-only message with a staged handoff (whose only text message is the origin block) rendered the raw block in a user bubble. Fixed in core so every host inherits it: the pick now skips a message that is entirely a whole-message origin block and falls back to the line-only text message the no-user-text case already produced. `isWholeOriginBlock` is exported from the markers module and implemented by running `parseHandoffMessage` / `parseModelSwitchMessage`, so it cannot drift from what they accept. `[use_skills]` and `[scheduled_task]` are prefix+body blocks, deliberately excluded — appending after their body still works. Covered in both directions (`packages/core/test/input-images.test.ts`) plus a server-level test of the files-only-plus-origin-block request.

**2. No cap on attachment count or aggregate size (major).** `parseTaskInput` now enforces **20 files** and **12MB of decoded bytes** per request, checked after each `file` part so a hostile `input` stops at the item that crosses the line rather than being decoded in full first. The caps live next to the per-file one in `services/task-attachments.ts` with the rationale in the comment: 20 is far past any usable chip row, and 12MB (~16MB of body once base64 inflates it) is the bound a caller actually reaches, so the guarantee no longer depends on the body middleware. 413 `too_many_files` / `payload_too_large`, both with EN+ZH copy. Two new tests, each asserting nothing was written.

**3. The body cap was header-only (major).** Replaced the `content-length` check with hono's `bodyLimit`, which keeps the header fast path and otherwise counts bytes off the stream, aborting the moment the total crosses 20MB. `onError` throws the same `HttpError`, so the 413 `payload_too_large` body is unchanged. The harness *can* express this: `packages/server/test/body-limit.test.ts` posts a `ReadableStream` body with no `content-length` through `app.request`. The body is valid JSON on purpose — with the old middleware that test returns **202**, which is the reviewer's reproduction. Three cases: streamed over-cap → 413, declared over-cap → 413 via the fast path, streamed under-cap → 202 with the body intact.

**4. Files written before the busy check, write loop not atomic (minor).** Now transactional from the caller's side. `attachFilesToInput` removes what it already wrote if the batch fails part-way, and the route removes the whole batch if `startTask` throws — so a 409/500 never leaves bytes that no message references and a retry cannot land a second copy. The cheap pre-check was also possible: `SessionManager.assertCanAcceptTask` runs the lock-free half of `startTask`'s rejections (shutdown, deleting Agent/Session, and — without `queueIfBusy` — a busy Session) before anything is written. It is explicitly advisory; the authoritative check still runs under the Session lock, which is what the cleanup path covers. Both halves tested.

**5. Symlinked parent (minor).** The session scratchpad directory is now realpath'd after `mkdir -p` and required to resolve inside the Agent's scratchpad root, the same containment rule `workspace-files-service.resolveWriteParent` applies. The check is on the canonical path but the write stays on the logical one, so message paths do not shift on hosts where the data root sits behind a link. `wx` (`O_CREAT|O_EXCL`) still covers the final component — verified, and now stated in the comment rather than left implicit. Tested with a planted `pad → outside` symlink (skipped on win32, where creating one needs a privilege).

**6. A hand-typed `[attached file: …]` rendered as a system notice (minor).** The file line is now gated on the scratchpad shape exactly as the image line is: anything else stays plain text in the body, so neither a user nor another project member can plant a banner in someone's view, and `[attached file: /etc/shadow]` no longer reads as a genuine marker. The test that enshrined the old behaviour is inverted and extended with the review's own example. The banner still shows base names.

**7. Response hardening (minor).** Non-image types on the scratchpad read endpoint now carry `Content-Disposition: attachment`, so the allowlist is no longer the entire defense, and `MIME_BY_EXT` is marked as a security boundary — with why adding `.svg` / `.pdf` there is not the one-line change it looks like. Documented in `server-api.{en,zh}.md`.

**8. Composer nits (minor).** Oversize picks are refused from `File.size` before the file is base64-encoded in the tab, matching what trace import already does; a batch is staged only once every read resolves, so chip order — and the `[attached file: …]` line order that follows it — is pick order rather than `FileReader` completion order; and the banner takes the chip's truncation rule so a long list cannot grow it into a paragraph.

**9. Design docs.** Out of scope for this PR — the design repo is owned separately and its specs move in their own PR.

**Not addressed, deliberately:** persisting attachments in the draft cache. Base64 blobs would blow the localStorage quota, so the fix is an upload endpoint, not a bigger draft — a separate change.

### Follow-up verification

| Command | Result |
| --- | --- |
| `pnpm -r build` | pass (all 7 packages) |
| `pnpm typecheck` | pass (all packages) |
| `pnpm test` | pass — core 589, server 429, web 471, cli 202, docs 11, skills 16, landing 39 |
| `pnpm format && pnpm format:check` | pass ("All matched files use Prettier code style!") |

Docs updated in both languages: `server-api.{en,zh}.md` (measured body cap, the new per-request caps, the scratchpad endpoint's response headers) and `web-app.{en,zh}.md` (the caps and the client-side pre-check).


hiyouga:
## Review

Reviewed in a clean detached worktree at `7153222`, scoped to `origin/main` (`4681135`) — the 19 files, one commit. I ran the two new test files (7/7 and 8/8), plus three throwaway probes: 30+ hostile filenames through the real sanitizer onto a real temp filesystem, the origin-block parsers against a trailing attachment line, and a chunked-encoding body-cap probe against this worktree's `@hono/node-server`.

**The headline security question does not land, and the sanitizer is genuinely solid** — details in the last section, because it's worth recording what was chased and closed. But the aggregate-size story rests on a cap I demonstrated is bypassable, there is no cap at all on attachment count, and the trailing-marker rationale is wrong for origin blocks in a way that breaks a path this PR adds.

### 1. The 20MB body cap is header-only and bypassable — major

`app.ts:335-339`, relied on by `task-attachments.ts:377`:

```ts
const contentLength = Number(c.req.header("content-length") ?? 0);
if (contentLength > MAX_BODY_BYTES) throw new HttpError(413, "payload_too_large", …);
```

A chunked request carries no `content-length`, so `Number(undefined ?? 0)` → `0` → passes. Reproduced against this worktree's `@hono/node-server` with the identical middleware shape: a **30MB** chunked POST returned **200** with the body fully parsed (`seen: 31457291`).

That matters because the cap is the PR's entire justification for having no per-request aggregate limit — `task-attachments.ts:371-376`, repeated in `server-api.en.md` / `server-api.zh.md`:

> the server already caps request bodies at 20MB … several attachments in one message must fit that same body cap

The premise is false. Any authenticated project member can POST `/api/sessions/:id/tasks` with `Transfer-Encoding: chunked`, `Content-Type: application/json` and N `file` parts of 9.9MB each. Every part clears the per-file 413 individually; `attachFilesToInput` writes all N into the session scratchpad sequentially. Unbounded disk write, plus memory pressure — peak per request is roughly 3.5× the body (buffered chunks + JSON string + parsed string + decoded `Buffer`).

The bypass predates this PR, but the only sinks were `image_url` parts (which cost the attacker context and tokens) and the owner-only trace import (`requireString(..., maxLen)`). This PR turns it into a plain any-member arbitrary-size disk write.

### 2. No cap on the number of attachments — major

`sessions.ts:89-96` (`parseTaskInput` iterates `input` with no length check) and `task-attachments.ts:497-499`.

Even honouring a 20MB body, a minimal `file` part is ~60 bytes — `{"type":"file","fileName":"a","dataUrl":"data:;base64,YQ=="}` — so ~350,000 fit in one legitimate request. In a single POST that's 350k sequential `fs.writeFile` calls into one directory, 350k `[attached file: …]` lines appended to one user message (prompt and Trace record both explode), and after the first `a` every subsequent identical name burns up to 16 `EEXIST` opens (`MAX_NAME_ATTEMPTS`), ending in an uncaught 500 once the 6-hex space saturates.

Same fix as §1: enforce a real check inside `parseTaskInput` — max `file` parts per request and max total decoded bytes — rather than delegating to the body middleware.

### 3. The trailing-marker rationale is wrong for origin blocks — major

The body's core premise:

> It **trails** the body rather than prefixing it — `[use_skills]` and the origin blocks are parsed at index 0, so a leading block would break that chain.

Half true. `parseSkillsMessage` really is index-0-only (confirmed: a trailing line leaves `{skills:[…], rest:"hello\n\n[attached file: …]"}` intact). But the origin blocks require a **whole-message** match — `origin-blocks.ts:139`, and its own doc comment says so:

```ts
if (!block || block.index !== 0 || block[0].length !== trimmed.length) return null;
```

`parseModelSwitchMessage:261-264` is identical. So a *trailing* line breaks them exactly as a leading one would. Demonstrated:

```
parse(block)        = {"agentId":"alpha","agentName":"Alpha","sessionId":"s0"}
parse(block + line) = null
```

Reachable via exactly the feature this PR advertises ("a message with attachments and no text is sendable"): `sendNormal` pushes a text part only `if (body)`, so with no text, no skills and only attachments, `body === ""` → `input` is file parts only → `chat-page.tsx:663` posts `{ input: [origin, ...input] }` → the server's `appendAttachmentLines` finds the **origin block** as the last user text message and appends to it.

Result: instead of a one-line `HandoffBanner`, the new chat's first message renders the whole `[handoff_from] … agent: alpha (Alpha) … [/handoff_from]` block verbatim in a user bubble. No test covers it. `/model` escapes only by accident — `pickSwitchModel` always supplies a fallback body (`S.chat.modelSwitchAutoMessage`); the handoff path has none.

### 4. Files are written before the 409 busy check — minor

`sessions.ts:481-492` + `runtime/session-manager.ts:461-478`. `attachFilesToInput` is awaited, *then* `startTask` runs `assertOpen` / `assertAgentNotDeleting` / `assertSessionNotDeleting` / `ensureEntry` / `assertIdle` — any of which can 409. A busy session without `queueIfBusy` therefore 409s with the bytes already on disk. The Web correctly keeps the chips on failure, so the user's retry lands a second copy as `report-<hex>.pdf`. The comment above the call acknowledges the queued and self-heal cases but not this one.

Related, **`attachFilesToInput` is not atomic**: a non-`EEXIST` error on write #3 of 5 500s the request while files 1–2 stay on disk with no marker referencing them and no cleanup.

### 5. The sanitizer can emit a name the read endpoint refuses — minor

`task-attachments.ts:446-451`:

```ts
const stem = clean(fileName.slice(0, fileName.length - ext.length)).slice(0, MAX_STEM_LEN);
return `${stem.replace(/^-+$/, "") || "file"}${clean(ext)}`;
```

The 80-char truncation can land exactly on a `.`, and `clean(ext)` re-adds one. Demonstrated: `"a"×79 + ".b.pdf"` → **`"a"×79 + "..pdf"`**. Not a traversal (`..` inside a single path component is inert for `path.join`), but `sessions.ts:360` rejects `fileName.includes("..")` outright, so that attachment is permanently 404 on the scratchpad endpoint — directly contradicting the function's own doc comment ("the same character set the scratchpad read endpoint serves, so an attachment stays fetchable"). Untested.

### 6. A hand-typed `[attached file: …]` line renders as a system notice — minor

`attachment-lines.ts:126,134` + `attachments.ts:63-68`. Filename-side forgery is **blocked** — `[`, `]`, newline and every separator look-alike sanitize to `-`, and the rest of the path is server-owned. But nothing constrains what a user types into the composer body. Unlike the image line, which must *resolve* to a scratchpad path or an http(s) URL or it stays plain text, the code comment states outright "A file line always resolves".

So `[attached file: reset your password at evil.example.com]` typed into the composer is stripped from the body and re-rendered as `Attached files: reset your password at evil.example.com` inside the system-notice chrome of `attached-files-banner.tsx` — and the model reads it as a genuine marker, so `[attached file: /etc/shadow]` reads as an invitation to open that path. In a multi-member project, member A can plant that banner in member B's view. React escapes the text, so this is UI spoofing plus a prompt-injection nudge, not HTML injection. The PR's own test enshrines the behaviour ("a file path outside the scratchpad is still an attachment"). Suggest gating the file line on `SCRATCHPAD_PATH` the way images are gated.

### 7. Windows reserved device names pass through — minor

`CON`, `NUL`, `PRN`, `COM1` all survive the sanitizer verbatim, as do `AUX`, `LPT1-9` and any of them with an extension (`con.txt`). This project ships Windows support as of 0.1.4; on win32 `path.join(dir, "NUL")` addresses the null device, so the upload is silently discarded and the model gets a path that reads empty. PLAUSIBLE — settled by running the test file on win32. No test.

### 8. Symlink: final component is safe, parent is not — minor

Contrary to my own hypothesis, `wx` (`O_CREAT|O_EXCL`) does **not** follow a final-component symlink. Planted `pad/report.pdf → OUTSIDE.txt`, uploaded `report.pdf`, got `EEXIST` → retry → `report-10d7d0.pdf`, and `OUTSIDE.txt` still read `original`. That vector is closed and it's a real win.

Writing **through a symlinked parent** does work: with `padlink → /real/dir`, the upload landed in `/real/dir/z.txt`. `fs.mkdir(dir, {recursive:true})` silently succeeds on an existing symlink-to-dir, and there's no `realpath` or `O_NOFOLLOW`. Honest severity: **not** a privilege-boundary crossing today — only the agent process can plant that link, it runs as the same uid as the server, and it would already need write approval, at which point it has arbitrary write anyway. The finding is *consistency*: `workspace-files-service.write` in this same repo does `O_NOFOLLOW` + a canonical-parent check + a win32 `lstat` fallback; the new upload path does none of it.

### 9. Design docs not updated — minor (project rule)

CLAUDE.md requires design and code to move in sync. The design repo (`design/` → `penguin-harness-design`, HEAD `92d98d7`) has nothing on file attachments: `05-ARCHITECTURE.md`'s API listing has no `file` input part, `06-PROTOTYPE.md`'s composer section doesn't mention it, `08-CHANGELOG.md` has no line.

This is more than a missing entry. The specs frame the scratchpad as *"供模型存放本次运行的计划与临时文件"* (`01-PRINCIPLES.md:105`) and *"scratchpad 只放中间产物"* (`08-CHANGELOG.md:74`). This PR gives it a third role — user-uploaded **inputs** — which is a design-level decision, not an implementation detail. `packages/docs/` was updated correctly, but that's the shipped product doc, not ground truth.

### Nits

- **`.` as a filename lands as a hidden file.** `scratchpadName(".")` returns `"."`; `path.join(dir, ".")` *is* the directory, and the write only survives because `mkdir` ran first so `wx` hits `EEXIST` and retries → `.-e325b9`. Related: `.bashrc` → `.bashrc`, `-rf.txt` → `-rf.txt`. Nothing exploitable — the sanitizer just has no "result must not be a relative-path token" rule.
- **No client-side size pre-check, against this repo's own convention.** Acknowledged in the body, but `design/specs/06-PROTOTYPE.md:131` already establishes the pattern for trace import: *"上限 14MB，超限在前端读取前即拒绝"*. One line in `addAttachments` matches it, instead of base64-encoding the whole file in the tab first.
- **Attachments never reach the draft cache.** `draft-cache.ts` / `use-session-draft.ts` persist text + handoff + skills only, and `ChatPage` remounts `ChatInput` per `sessionId`. Picking three 8MB files and switching sessions loses them silently; `toggleGoal(true)` also clears them with no confirmation. Consistent with images, so pre-existing shape — but an un-recoverable file pick is a worse loss than a re-pastable image.
- **`addAttachments` ordering is completion-order, not pick-order.** `chat-input.tsx:1810-1820` pushes on `FileReader.onload`, so chip order and `[attached file:]` line order aren't guaranteed to match the order the user picked.
- **Banner has no max-width** (the chip has `truncate max-w-56`), so many long names wrap tall. Cosmetic.

### The XSS question, chased to a conclusion

`sessions.ts:374-390`. `MIME_BY_EXT` is an **allowlist of five image types** (`.png/.jpg/.jpeg/.gif/.webp`); everything else falls to `application/octet-stream`. With the PR's new `nosniff`:

- `evil.html`, `evil.svg`, `evil.xhtml`, `evil.xml`, `evil.mhtml` → `application/octet-stream` + `nosniff` → the browser downloads, never parses. No script execution, no same-origin access.
- A polyglot uploaded as `evil.png` → `image/png` + `nosniff` → rendered as an image, no script.
- The sanitizer's charset `[A-Za-z0-9._-]` matches the endpoint's own `^[A-Za-z0-9._-]+$` gate, so nothing else sneaks in.

**Residual hardening gap (minor):** unlike the sibling workspace read at `sessions.ts:584-608` — which downgrades `scriptable` types to `text/plain`, sets `Content-Disposition`, and sandbox-CSPs the preview — this response has no `Content-Disposition` and no CSP. The entire defence is "the map stays image-only". Adding `.svg` or `.pdf` later is a one-line, innocuous-looking change that becomes same-origin stored XSS immediately. Worth a `Content-Disposition: attachment` for non-image types and a comment marking that map as a security boundary.

### What else I checked and couldn't break

- **The sanitizer.** 30+ hostile inputs executed against the real functions and, where accepted, written to a real temp directory. `..`, `../../etc/passwd`, `sub/dir.txt`, `C:\x`, `a\u0000.txt` → 400. `a.txt:b` → `a.txt-b` (ADS neutralized). `a⁄b.txt` (U+2044) and `a／b.txt` (U+FF0F) → `a-b.txt`. `\u202Egnp.exe` → `-gnp.exe` (no extension spoofing in the banner). `a]b.txt` → `a-b.txt`, so `]` can never reach the marker line. `中文文件.pdf` → `file.pdf`. `"a"×300 + ".pdf"` → `"a"×80 + ".pdf"`. Only §5 produced anything unexpected. **Check-then-use is safe**: validation runs on the original name and the write uses only the sanitized name, but the sanitizer's output alphabet cannot contain `/`, `\` or `\0` and cannot create a `..` except via §5, which is inert for `path.join`. The sanitizer alone is sufficient.
- **`sessionId` / `agentId` in the path.** `sessions.ts:481` uses `row.projectId`/`row.agentId`/`row.sessionId` from `resolveSession` (DB lookup + `requireProjectAccess`), never a raw route param. All are `isValidId`-constrained. None can carry `]`, a newline or a separator, so the marker can't be forged through the path either.
- **Size check semantics are correct** (`:425-435`): the comparison is **decoded bytes vs 10MB**. 10MB of base64 *text* → accepted with `decoded = 7,864,320`; 10MB+1 of real bytes → 413 `file_too_large`. All validation completes before `attachFilesToInput` (`sessions.ts:475-481`), so "nothing touches the filesystem until the request is known good" holds and the goal-mode rejection genuinely precedes any write. The `[A-Za-z0-9+/=\s]+` payload class is anchored behind a comma-bounded prefix — no catastrophic backtracking.
- **"Producer and parser cannot drift" is real, not just claimed.** `attachments.ts:24-29` imports all four symbols from `@prismshadow/penguin-core/markers`; zero hardcoded strings remain. `appendAttachmentLines` is genuinely used by both producers (`session-support.ts:48`, `task-attachments.ts:500`). Edge cases verified: zero lines → unchanged; no user text → line-only message; body ending in `\n` → harmless triple newline; `[use_skills]` + trailing lines → parses fine.
- **`splitAttachments` image behaviour is unchanged** — diffed against `origin/main` line by line: same fast path, same `line.trim()` match, same `resolveAttachment` gate, same **untrimmed** `kept.push(line)`, same trailing-`\n+` cleanup.
- **Send-path state machine.** All five paths verified in code. `sendNormal` and `pickSwitchModel` both clear only under `if (ok)`; a 413/400 toasts (`file_too_large` and `payload_too_large` both have EN+ZH copy) and the chips survive. **No path clears on failure, no path forgets to clear on success.** Mid-run steering is text-only and attachments stay. Goal mode is blocked on both sides, and the server 400s at `sessions.ts:461-468` before any write.
- **i18n parity** verified programmatically, not by diff size: both `chat` blocks have exactly **149** keys, zero one-sided. Docs EN/ZH asymmetry (+7/−2 vs +6/−2) is comment line-wrapping only; content identical. Every added CJK line is in `strings.ts`.
- **Test quality is genuine, not mock theatre.** Real app over a real temp filesystem, bytes read back with `fs.readFile`, and "nothing written" asserted via `await expect(fs.access(dir)).rejects.toThrow()` on the **real** directory. None vacuous. The gap is coverage, not rigour: nothing from the hostile-input set above, and nothing for §3.

I skipped `pnpm -r build` and the full suites on this shared box; both touched test files pass.

### Merge-order note

This PR and #122 overlap on 8 files, including `chat-input.tsx` — `+167/−29` here against `+302/−225` there, which is close to a rewrite. That's a guaranteed hard conflict, and they interact semantically: #122 restructures the handoff send path that §3 travels through. Worth landing one and rebasing the other rather than resolving the conflict mechanically.

PR #122, opened 2026-07-29
feat(web): replace the @ mention with an /agent command, and stage both switches until send

hiyouga:
## What changes

The `@`-mention handoff is gone; `/agent` takes its place, and both switch commands now wait for Enter instead of firing at pick time.

**`/agent` replaces `@`.** It is an ordinary slash command: running it consumes its token (like `/compact` and `/model`) and opens an agent picker built exactly like the `/model` one — title bar, search box (agentId / display name, case-insensitive substring, same rule as the model search), capped scrolling list, click-outside and Escape to dismiss, upward height measured against the real ceiling. Rows carry the Agent avatar used everywhere else Agents are picked. Offered wherever there are candidates, so the draft page keeps its handoff.

**Both switches are staged, not immediate.** Picking an agent or a model sends nothing: it pins a chip above the text body, the user keeps typing, and **Enter/Send** performs the switch — the handoff to a new chat for that Agent, or the fork of this conversation onto the picked model. Only when the composer is empty does the default auto-message fill in (`modelSwitchAutoMessage` for a model fork; a handoff may still open with just its `[handoff_from]` block). Selected skills keep riding along in a `[use_skills]` block. The two chips are mutually exclusive with each other and with goal mode, and both are removed the same way as before (the chip's ×, or Backspace at the start of the text).

Every send path treats a staged model switch exactly like the handoff target: it alone makes the composer sendable, it keeps the single action button on Send rather than Stop, and it blocks mid-run steering (a staged switch is not a steering message; the follow-up queue keeps its old "a switch opens its new chat directly" behaviour). Draft caching is unchanged — the handoff target still survives a reload via `initialHandoffTargetId` / `onHandoffTargetChange`; a pending model pick was left as transient UI state and not cached (superseded — see Review follow-up §2).

**Dead code removed with the trigger:** `matchMention`, `splitLeadingMention`, `MentionMatch`, and the prefix-matching `filterAgents`. What survives is the picker's search filter plus the origin marker producers/parsers other modules import, so `agent-mentions.ts` becomes `agent-handoff.ts` (mechanical rename, its unit test follows). Retired copy: `mentionHint`, `mentionRemove`; new copy in both dictionaries: `switchAgent`, `switchAgentTitle`, `agentSearchPlaceholder`, `agentsNoMatch`, `handoffTargetTitle`, `handoffRemove`, `modelSwitchTargetTitle`, `modelSwitchRemove`.

Docs (`web-app`, `sessions-and-traces`, `server-api`, en + zh), the composer's file-header documentation and the neighbouring comments that described the `@` flow, and the draft e2e spec all follow the new interaction.

**Follow-up commit — the model-facing prose.** The `[handoff_from]` block still told the receiving model "This conversation was opened by @-mentioning you", naming an interaction that no longer exists. The tag and its field lines are a persisted format (old Traces parse through the same code), so only the prose changed: it is now trigger-agnostic, and the neighbouring core comments no longer name the `@` either.

**`/agent` is a session command now** (review comment, second follow-up commit). Like `/model`, it has nothing to act on in a draft: there is no conversation to hand over, and the draft page picks its Agent in its own selector — a handoff there would have opened a second chat whose `[handoff_from]` block pointed at a conversation that never happened. `onHandoff` became an optional prop supplied by the chat page alone, and the command is gated on it exactly as `/model` is gated on `onSwitchModel`; the draft page dropped its handoff callback, its cached target and the chip wiring, while the per-session draft keeps all of it. The draft e2e now asserts the command's *absence* — and asserts the slash menu did open on the same prefix (the installed `/agent-*` skills), so it is a real absence rather than a menu that never appeared.

## Verification

- `pnpm -r build` — pass (all 8 packages)
- `pnpm typecheck` — pass (all 7 projects)
- `pnpm test` — pass (1730 tests: core 584, server 411, cli 202, web 467, landing 39, skills 16, docs 11)
- `pnpm format` + `pnpm format:check` — pass ("All matched files use Prettier code style!")
- `pnpm --filter @prismshadow/penguin-web test:e2e` — the rewritten `draft.spec.mjs` passes (`/agent` → picker → staged chip → keep typing → reload restores the chip → remove it). 23 passed / 4 failed; the four failures (`draft-user-scope`, `layout:370`, `paging`, `skills`) fail identically on clean `main` in this environment and are unrelated to this change.
- Extra manual e2e (throwaway spec, not committed): in a live session `/model` stages a chip without navigating, the × cancels it, and re-staging + typing + Enter forks into a new session on the picked model whose first message carries `[model_switch_from]` followed by the typed text.

## Review follow-up

Review findings addressed in `196ccdc`. Numbering follows the review comment.

**§1 — the deleted run-state guard (major).** Restored, at both levels the review asked for. The staged-switch decision now lives in one pure function, `stagedSendRoute` (`agent-handoff.ts`): a staged `/model` fork is `blocked` while the session is `running` or `compacting`, and `blocked` is deliberately **not** `post` — falling through would deliver the draft to the very session the user was switching away from. `canFollowUp` reads it (eligibility, so Queue mode no longer forks off a live trace when a run starts from outside the composer), and `sendNormal` re-checks it (that path is deliberately ungated on run state — the steering completion race calls it directly). A staged **handoff** keeps its old freedom: it never touches the running session. Rather than a silently dead button, the composer's existing hint-line slot above the card now says why the fork is waiting — new key `modelSwitchBusyHint` in both string files, `chat` key sets still equal at 152/152.

**§2 — the chip lost on unmount while its text survived (major).** The staged model is now cached with the text, the way `handoffAgentId` already was: new `switchModelRef` field in `draft-cache.ts` (a paired reference, validated through a shared `parseModelRef` alongside `modelRef`), carried by `use-session-draft.ts` (`onPendingModelChange`, immediate write like the handoff target; also cleared in `discard`) and wired in `chat-page.tsx` via `initialPendingModelRef`. Restored only while that model is still in the available list **and** is not the one this session already runs (a cache that outlived a fork); otherwise discarded. This supersedes the original body's "deliberately not cached".

**§3 — no test proved the switch waits for Enter (major).** Covered on both halves. E2E (`draft.spec.mjs`, in a live session, so the `/model` path is genuinely exercised): `/model` → pick → assert the URL did **not** change and the chip is pinned → type the body **after** the pick → reload (chip *and* text both restored from one cache entry) → Enter → assert the new session exists on the picked model with the same Agent, the "switched model" banner naming the previous one, and the typed body carried in. Then the same shape for `/agent`: stage → still in the source session → type → Enter → new session on `default_agent` with the handoff banner and the typed body. Unit: `stagedSendRoute` gets a 6-case matrix in `agent-handoff.test.ts` (including "blocked never degrades into post"), and `draft-cache.test.ts` gets the new field plus a malformed-value case in the style of the existing ones.

**§4 — picking the current model didn't clear an already-staged chip.** It now clears the staging: forking onto the model already in use can only mean "never mind", so leaving an earlier pick armed was the opposite of the intent.

**§5 — two chips on a cold load.** The two restore effects fire off separately-fetched lists, so either can land after the user has staged something by hand. A restore now only fills an **empty** slot (no chip, no goal mode): live intent is newer than the cached one and wins, exactly as one pick drops the other. Only one of the two can be cached at a time anyway, so the ordinary case is unchanged.

**§7 — Escape and IME in the pickers.** The window-level handler now carries `!e.isComposing`, following the Enter guards already in the file: dismissing a Pinyin candidate list no longer closes the picker (which was unrecoverable — the command had already consumed its token). `closeAll` also returns focus to the textarea, on Escape and click-outside alike.

**§8 + §9 — the copied picker, keyboard selection, current-agent marker.** Extracted rather than patched: `PickerList` owns the search box, the `max-h-56` cap, the row chrome, the keyboard navigation and a current-entry marker slot; `SwitchPickerPanel` owns the upward popup frame and title bar. Both pickers use them, each keeping its own row (provider logo vs Agent avatar). That closes all three gaps at once — ArrowDown/Up/Enter/Tab navigation (IME-guarded, Enter/Tab committing the highlighted row or the top match), and a ✓ marker for the session's own Agent. Keyboard highlighting starts **unset** on purpose, so the model dropdown is pixel-identical until an arrow key is pressed. Picking the session's own Agent stays a real action (a fresh conversation with the same Agent), so it is marked but not made a no-op.

**Nits.** `handoff-banner.tsx` dropped the `@` sigil, so it names the source Agent the way the composer chip does. The two Chinese JSDoc comments this PR touched in `strings.ts` are English again (values unchanged, per #117). `agent-optimization/SKILL.md` no longer says "@-mention handoff", with the skill version bumped 5 → 6 and `updated` refreshed, per the convention the other skills use.

**Not addressed:** §6 (the design specs) — `design/` is owned by a separate change and is out of this worktree's scope.

`packages/docs/content/web-app.{en,zh}.md` gained the behaviour §1 and §2 changed (the fork waits for an idle session; both chips are cached with the draft), kept line-for-line in sync.

### Verification (follow-up)

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (web 474, +7 from the new `stagedSendRoute` and `switchModelRef` cases)
- E2E, isolated ports (the default ones are taken on this shared box): full suite **22 passed / 5 failed**, and the *same 5* fail on this PR's head (`572fb1c`) with the fixes reverted and the web bundle rebuilt — `draft-user-scope`, `layout:370`, `paging`, `skills`, and `subagent:261`. `subagent:261` is a fifth pre-existing failure worth recording: it only appears when the four broken specs run before it (they share one server and data root); with those excluded, the remaining 18 specs pass 3/3 consecutive runs. The extended `draft.spec.mjs` passes in every configuration (28.5s against a 90s budget).


hiyouga:
## Review

Reviewed in a clean detached worktree at `572fb1c`, scoped to `origin/main` (`4681135`) — the 26 files, both commits. Ran `markers.test.ts` (16/16), `agent-handoff.test.ts` + `skill-slash.test.ts` (24/24), and a throwaway parser probe; skipped `pnpm -r build` and Playwright on a shared box.

**The persisted-format claim is the best-verified statement in the batch and it is fully true** — details below, because it was the biggest risk here and it's worth recording how it was settled. What needs work is the staging half: a deleted run-state guard, a chip that's lost while the text it belongs to survives, and no test for the model side of the PR's own central claim.

### 1. A run-state guard was deleted and only half-replaced — major

`chat-input.tsx:1489-1497` + `chat-page.tsx:594`. The old `pickSwitchModel` opened with:

```ts
if (!onSwitchModel || busy || running || compacting) return;
```

commented *"the run state is re-checked — the picker may have survived a status flip"*. The new one has only the same-model check.

The guard did move to send time for the idle path — `canSend` requires `!running && !compacting && !busy`. It did **not** move for the mid-run follow-up path: `canFollowUp` (`:1319-1329`) requires `running` and now includes `pendingModel !== null`; `send()` (`:1704-1713`) routes to `sendNormal(onQueueFollowUp!)`; `sendNormal` (`:1745-1748`) short-circuits `post` entirely via `switchModel ? await onSwitchModel!(...)`. `chat-page.tsx:594`'s `onSwitchModel` has no run-state check of its own.

Scenario: idle session → `/model`, pick a model, keep typing → a run starts from outside this composer (queued follow-up auto-sends, scheduled task fires, another tab or the CLI posts) → user's mid-run mode is **Queue** → Enter. Instead of queueing, `onSwitchModel` fires: a new Session is created, `getSession().tracePath` returns the **still-being-appended** trace, the `[model_switch_from]` block points the new model at a truncated conversation, and the user is navigated away from a live run.

Fair mitigations: the staged *handoff* already behaved this way pre-PR, and in default **Steer** mode `canSteer` is false so the button is merely disabled. But `design/specs/06-PROTOTYPE.md:63` scopes `/model` to *"仅活跃**空闲**会话"* (idle only), and the deleted guard was written for exactly this case.

### 2. The staged model chip is lost on unmount while its text survives — major

`chat-input.tsx:1174`. `pendingModel` is component state and deliberately not cached; the composer text **is** cached (`use-session-draft.ts`, debounced and flushed on unmount). `chat-page.tsx:1092` keys the composer subtree by `selected?.sessionId`, so `ChatInput` unmounts on **every session switch**, not only on reload.

Scenario: in session A, `/model` → pick GPT-5 (chip pinned) → type *"redo the whole refactor, be aggressive"* → click another session and come back. Text restored verbatim, chip gone, no indication. Enter posts **to session A on the old model** — the opposite of what was staged.

The body frames this as "a pending model pick is transient UI state and deliberately not cached", which is a coherent choice in isolation. It stops being coherent once the text it belongs to is persistent: the two halves of one intent have different lifetimes. Either cache the pick alongside the text, or clear the cached text when the pick is dropped.

### 3. No test proves the switch waits for Enter, for the model half — major

`agent-handoff.test.ts` (new, 7 tests, passing) covers only `filterAgents` plus marker round-trips. `draft.spec.mjs` proves "the handoff didn't fire at pick time" only **implicitly** — it keeps driving the draft page afterwards, which would fail on a navigation — and then **removes the chip before sending**, so "Enter performs the handoff" is never exercised. The staged **model** switch has **zero** coverage.

Given this PR's whole point is converting `/model` from immediate to staged, that's the assertion that most needs to exist.

### 4. The specs still describe both removed behaviours — major (project rule)

CLAUDE.md requires design and code to move in sync, and `design/` is ground truth. Still present in the design repo (HEAD `92d98d7`, tree clean — so this isn't "lands separately"):

- `06-PROTOTYPE.md:63` — *"**选中模型即**…开新会话…**并跳转**"* (picking switches immediately and navigates). Directly contradicted by this PR.
- `06-PROTOTYPE.md:63` — *"slash 与 **@ 菜单**都向上展开"*.
- `05-ARCHITECTURE.md:95` — *"沿用 **@ handoff** 口径"*.
- `05-ARCHITECTURE.md:290` — *"（同 **@ handoff**）"*.

`08-CHANGELOG.md` is a historical log and correctly frozen. `packages/docs/` was updated properly — EN/ZH got corresponding edits line for line, and no doc still describes the `@` flow.

### 5. Picking the current model doesn't clear an already-staged chip — minor

`chat-input.tsx:1491`: `if (sameModelRef(m, modelRef)) return;` returns before touching `pendingModel`. Stage model B, reconsider, run `/model` and pick the current model A meaning "never mind" → the picker closes and **chip B is still staged**. Enter forks onto B.

### 6. Two chips can render at once on a cold load — minor

`chat-input.tsx:1614-1620`. The handoff-restore effect clears neither `pendingModel` nor `goalOn`. `/agent` is gated on `agents`, `/model` on `models` — separate fetches. On a cold `/chat/:id` load where `models` resolves first, a model chip can be staged before `agents` lands; then `setTarget(restored)` fires and **both** render. `sendNormal` takes the `target` branch and silently drops the model chip.

### 7. The picker's Escape has no `isComposing` guard, and doesn't restore focus — minor

`chat-input.tsx:1470-1500`. zh scenario: `/agent` → Enter → type Pinyin in the search box → press Escape to dismiss the IME candidate list. Candidates *and* picker close, and because `/agent` already consumed its token (`clearInput()` at `:1420`) the text is gone too. `closeAll()` never refocuses the textarea, so focus lands on `<body>`.

The pattern is pre-existing on `/model` and now duplicated onto `/agent`. Credit where due: the **Enter** guards were preserved correctly (`!e.nativeEvent.isComposing` at `:1721` and in the send branch). The Backspace-chip branch lacks the guard but needs `selectionStart === selectionEnd === 0`, which a live composition normally doesn't satisfy — PLAUSIBLE only; a manual Pinyin trial would settle it.

### 8. Keyboard agent selection regressed — minor

The deleted `@` menu had ArrowDown/Up/Enter/Tab navigation. `AgentMenuList` has no `onKeyDown`. This *is* genuine parity with `ModelMenuList` (also click-only), so the "built exactly like the `/model` one" claim holds — but the net effect of the change is that keyboard agent selection, which existed, no longer does.

### 9. The picker is a copy, not an extraction — minor

The search-box block is byte-identical to `ModelMenuList`'s apart from two string keys, and the panel wrapper duplicates `anim-pop absolute bottom-full left-0 z-40 mb-1.5 flex w-80 max-w-[calc(100vw-2rem)] flex-col overflow-hidden …` verbatim — ~40 duplicated lines that will drift. Two related parity gaps while it's being touched: no current-agent marker (the model list marks with ✓), and no no-op when picking the session's own agent.

### 10. One surviving `@`-mention reference — minor

`packages/skills/skills/agent-optimization/SKILL.md:22`: *"an **@-mention handoff** contains `[handoff_from]`"*. Model-facing skill content, and the only stale reference left outside frozen `changelog/0.1.2/` and `design/`. Needs a skill version bump per convention.

Otherwise the dead-code sweep is **clean**: zero repo-wide hits for `matchMention|splitLeadingMention|MentionMatch|agent-mentions|mentionHint|mentionRemove|mentionDismissed|mentionIndex|insertMention`, no stale e2e selectors, README/landing/docs clear.

### Nits

- `handoff-banner.tsx:14-19` still renders the source agent as `@id` ("Handed off from @data_analyst's conversation") while the composer chip dropped its `@` prefix in this PR (`:2124`). The two now disagree.
- Mid-run with a staged chip in Steer mode disables Send with no explanation (`canMidRunSend` `:1335`, `canSteer` `:1306`). Pre-existing for the handoff chip, now extended to the model chip.
- `strings.ts:718` adds a **new** Chinese JSDoc comment — `/** /agent 交接：命令描述、拾取器标题、搜索框、无匹配提示、暂存目标的说明与移除按钮。 */` — and edits the adjacent `/** /model 切换：… */` without translating it. CLAUDE.md: Chinese *values* are required here, comments aren't, and "translate it to English when you touch it". The file is 50 English JSDoc comments to 6 Chinese, and the comment immediately below the new one is English. #117's English-only sweep didn't reach `strings.ts`, so these are exactly the residue that rule targets — and this PR touched two of them. All other Chinese added is legitimate (`*.zh.md` docs, two `draft.spec.mjs` selectors matching zh UI output).
- `/AGENT` (case-sensitive `startsWith` at `:1459`) and `/agent ` with a trailing space (fails `CMD_PREFIX`, `slash-token.ts:10`) fall through to literal body text. Pre-existing and shared by every command — consistency, not a regression.

### Persisted format: old Traces still parse

Verified three ways, since this was the headline risk:

1. **Byte-level.** The correct-base diff touches exactly three things: `tags.ts:24` is a **JSDoc comment only** (`handoffFrom: "handoff_from"` unchanged, `TITLE_NOISE_TAGS` unchanged); `origin-blocks.ts:93,96,105-116` are comments; `origin-blocks.ts:129` is the one prose sentence. `HANDOFF_PATTERNS` (`:135`), `parseHandoffMessage` (`:141-161`) and the shared `fieldLines` (`:82-90`) are untouched.
2. **Structurally prose-agnostic.** `fieldLines` matches `^(agent|session|workspace): (.+)$` per line and discards everything else; the tag match keys on the tag. Neither the old sentence nor the new one has a line starting with those keys.
3. **Empirically.** Reproduced a block emitted by `origin/main` byte-for-byte (old prose, all three field lines) and ran it through this branch's parser: square form parses to the identical origin, legacy `<handoff_from>` angle form parses, title stripper still removes it — 3/3.

A backward-compat pin already exists: `markers.test.ts:220-225` parses a legacy block whose body contains a literal prose line plus field lines — genuine prose-agnosticism coverage, not just format coverage. `grep -rn "mentioning you\|opened by @"` returns nothing repo-wide, so the old prose is never compared or depended on.

### Staged-switch state matrix

| Scenario | Actual | |
|---|---|---|
| Stage model → run starts externally → Send | Steer: button disabled, state retained. Queue: **forks mid-run off a live trace** | §1 |
| Stage, run completes | No status-driven clear; chip survives, `canSend` flips true | correct |
| Backspace removal | Empty + caret 0 → removed; text non-empty + caret 0 → removed (intended); **selection active** → selection deleted, chip kept | correct |
| Stage agent → stage model (both ways) | Each pick clears the other and updates the draft cache | correct, except §5/§6 |
| Goal ↔ chip, both directions | `toggleGoal(true)` clears target + `onHandoffTargetChange(null)` + `pendingModel` + images; both picks `setGoalOn(false)`. Latest wins both ways | correct |
| Send fails (network / 4xx) | Clears only inside `if (ok)`; text, images, skills, target, `pendingModel` all retained; orphan session deleted + toast | correct |
| Reload / session switch with staged model | Text restored, chip gone | §2 |
| Chip staged while compacting | Can't stage; pre-staged → `canSend` false, `running` false → `send()` returns; button disabled | correct |

### Claims that held up

`/agent` consumes its token like `/compact`/`/model`; picker parity (title bar, search, `max-h-56` cap, click-outside + Escape, upward measure floors at `Math.max(96, …)` so it can't render off-screen); case-insensitive substring search, unit-tested; offered on `agents.length > 0` so the draft page keeps its handoff and there's no empty picker; staging sends nothing; the empty-body fallbacks exactly match the old `/model` logic; skills still ride in `[use_skills]`; chip × and Backspace both work; every send path treats a staged model like the handoff target; the handoff target still survives reload via `initialHandoffTargetId`/`onHandoffTargetChange`; dead-code removal complete in code.

**i18n parity** checked programmatically rather than by diff size: both `chat` key sets are **151 keys**, `only in EN: []`, `only in ZH: []`. All 8 new keys in both, both retired keys gone from both. The +9/−3 vs +11/−4 asymmetry is purely added JSDoc lines.

### What's good

The persisted-format touch is exemplary — comment-only in `tags.ts`, prose-only in `origin-blocks.ts`, plus a standing note at `origin-blocks.ts:113-115` explaining *why* the prose must stay trigger-agnostic. That comment is the right artifact to leave behind. Failure handling is correct throughout: nothing clears unless `ok`. And collapsing `pickSwitchModel`'s bespoke send assembly into the single `sendNormal` path removes a real duplicate of the skills/images/auto-message logic — that consolidation is most of the `−524`, and §1 is the only dropped guard I found in it.

### Merge-order note

This PR and #121 overlap on 8 files, including `chat-input.tsx` — `+302/−225` here against `+167/−29` there. Guaranteed hard conflict, and they interact semantically: #121 has a bug where a files-only handoff appends an attachment line to the `[handoff_from]` block and breaks its whole-message parse, and this PR restructures exactly that send path. Worth landing one and rebasing the other rather than resolving mechanically.

hiyouga:
这个和模型一样，在对话为空的时候，不应该显示这个选项，因为没有需要Switch的

PR #142, opened 2026-07-30
feat(installer,web): one canonical bundle per target; robust in-place upgrades; update-check feedback

laodouuu:
## Summary

Rebuilt from scratch on top of latest `main` (replaces the previous online/offline-unification design on this branch, which kept raw payloads and `*-offline` wrappers side by side).

There is now **exactly one artifact per target**. `penguin-<target>.tar.gz` (and `penguin-win32-x64.zip`) is a flat installer bundle sealing three things: the native installer, the program payload (`payload.tar.gz` / `payload.zip`), and the payload's SHA256 checksum. The same file serves both installation modes:

- **Online**: `install.sh` / `install.ps1` download the bundle, verify it against its published `.sha256` (now mandatory — the warn-and-skip fallback is gone), open the flat outer layer, verify the sealed payload checksum, then stage and swap as before.
- **Offline**: transfer that one file, extract once, run the bundled `./install.sh` (or double-click `install.cmd`). The installer finds the sibling payload by itself and verifies the same sealed checksum — no network, and no separate checksum file to carry.

The universal package uses the same shape. Raw platform archives and `*-offline` wrappers are no longer published.

## Design

- **Shape probing instead of name/fallback logic.** The installers probe archive content (tar listing / zip entry list — no extraction): a top-level `payload.*` means bundle; anything else is treated as a program archive (top-level `penguin/`). That one probe replaces the previous 404-fallback chain and keeps working in every direction:
  - pinned `--version` at a pre-0.1.6 release downloads the old raw archive under the same asset name and installs through the program-archive path;
  - `--archive` / `-ArchivePath` accepts a Release bundle (self-verifying via the sealed checksum; the outer `.sha256` is also checked when present), the inner payload itself (what `install.cmd` passes), or a legacy raw archive (adjacent `.sha256` required, as before).
- **Windows long paths**: the outer layer is flat (4 files), so the user- or installer-driven initial extraction never creates deep paths; only the installer expands the payload, straight into the short staging directory. The zip shape probe uses `System.IO.Compression` (PS 5.1-safe) without extracting.
- **No wrapper scripts**: the canonical `install.sh` / `install.ps1` / `install.cmd` travel byte-identical inside the bundles (the previous `penguin-installer.sh` + generated entry-point split is gone); sibling-payload detection makes the same script the online and offline entry.
- **Sibling pickup is hardened**, preserving 0.1.5's "never trust archives beside a temporary script" property without the wrapper: POSIX sibling detection engages only for a real file actually named `install.sh` (a forwarder temp file or `curl | sh` never qualifies), and the penguin.ooo forwarder now stages the downloaded installer in a private `mktemp -d` directory instead of a bare file in shared `/tmp`, so another local user cannot plant a payload next to it. The Windows forwarder already runs the installer as an in-memory script block (`$PSScriptRoot` empty), which keeps its sibling detection offline-only.
- **Packaging/validation**: `scripts/package-release-bundles.sh` replaces `package-offline-bundles.sh`; the release workflow builds payloads as intermediates, packages six bundles, and validates the real outputs before upload (outer checksum, exact flat member set, byte-identical installer and payload, sealed checksum passes). `SHA256SUMS` now covers exactly the six bundles.
- **Tests**: `scripts/test-installer.sh` and `scripts/test-installer.ps1` replace the offline-only scripts and hermetically cover: bundle layout, offline install with network access stubbed to fail, corrupted-payload rejection, upgrade rollback, local bundle/payload archives, and the online flow (canonical install, outer and inner checksum mismatches, latest-404 and network failures with no fallback, pinned legacy versions). Wired into the existing Linux and Windows CI jobs.

## Windows fixes shipped with this rework

- **`penguin` no longer trips over the PowerShell execution policy.** The payload previously shipped both `bin\penguin.cmd` and `bin\penguin.ps1`; PowerShell prefers `.ps1` on PATH and client Windows defaults to the Restricted policy, so a fresh install failed with *"running scripts is disabled — penguin.ps1 cannot be loaded"*. The `.ps1` launcher is gone (from the payload and from the installer's shim regeneration): batch files are exempt from the execution policy and both PowerShell and cmd.exe resolve `penguin.cmd`, which forwards args and the exit code. `bin\` is swapped wholesale on upgrade, so **users who already hit this error just re-run the installer** — the stale `penguin.ps1` disappears with the old `bin\`. The docs' execution-policy note now only covers npm-generated shims.
- **New terminals now see `penguin` on PATH.** The installer writes the user `Path` through the raw registry API (to preserve `REG_EXPAND_SZ`), and a raw write does not broadcast `WM_SETTINGCHANGE` — Explorer, and every terminal window launched from it, kept the stale Path until the next logon, so a brand-new PowerShell window still reported `penguin` not found. The installer now broadcasts `WM_SETTINGCHANGE ("Environment")` via `SendMessageTimeout` after the write (best-effort, never fails the install), and the completion note says to open a **new terminal window** — a new tab of an already-running terminal inherits the terminal process's old environment by design.

## Robust in-place upgrades (`mv … Device or resource busy`)

`penguin update` re-runs the installer while the CLI process is still executing out of `lib/`; on filesystems that pin in-use directories — overlayfs under Docker is the reported case — the plain `mv lib .old.<pid>/lib` fails with `Device or resource busy` and the upgrade rolls back. Directory moves in `install.sh` now degrade through three strategies: whole-directory rename → per-entry renames into a fresh **or existing** destination → per-entry copy with the source entry removed (POSIX keeps an unlinked-but-open file valid for the process using it). A pinned, now-empty directory husk is left in place and reused by the incoming move, rollback follows the same rules, and the final failure message tells the user to stop running penguin processes before retrying. Covered by a hermetic test that stubs `mv`/`rmdir` to refuse directory renames touching the installed `lib/`, and verifies the upgrade completes, the copied subtree survives, and no `.old.*`/`.staging.*` are left behind.

## Update-check feedback in the Web App

The sidebar's manual "Check for updates" row now gives explicit feedback for **every** outcome: a spinner + busy label while the check runs; a success toast when already up to date; a success toast naming the release when a newer one is found — and the same row immediately turns into the persistent "New version vX available" entry that opens the update dialog (release notes + admin self-update); error and checks-disabled toasts as before. The outcome mapping is extracted into a pure, unit-tested classifier (`updateCheckOutcome`), including the precedence rules (disabled ≻ failed ≻ found) and the guard that "found" must name a version.

## Backward compatibility

- New installers keep a **legacy program-archive path** so pinned pre-0.1.6 versions and old local archives stay installable. This is the compat surface to eventually remove once pre-0.1.6 releases stop being install targets.
- **Hard break**: installer scripts saved from releases ≤ v0.1.5 expect the old raw layout and cannot install v0.1.6+ assets (they fail with a clear layout error). The documented flows are unaffected — the penguin.ooo forwarders always fetch the latest released installer.
- **`penguin.ps1` launcher removed**: typing `penguin` is unaffected (resolves to `penguin.cmd` in both shells); only direct `.\penguin.ps1` invocations need to switch to `penguin` / `penguin.cmd`. Upgrading replaces `bin\` and removes the old shim.
- Changelog entries (including the backward-compatibility file) are intentionally not in this PR; they go into the separate changelog summary PR.

## Verification

- `sh scripts/test-installer.sh` — bundle layout, offline, rollback, pinned-directory upgrade and online stub tests all pass locally
- `pnpm -r build`
- `pnpm typecheck`
- `pnpm test` — skills 18, docs 32, landing 39, core 598/2 skipped, server 437, cli 206, web 508 — all green (incl. the new `updateCheckOutcome` unit tests)
- `pnpm format` + `pnpm format:check`
- `sh -n` on `install.sh`, `scripts/package-release-bundles.sh`, `scripts/test-installer.sh`
- GitHub Actions workflow YAML parses
- `install.ps1` / `scripts/test-installer.ps1` are exercised and parser-checked by the Windows CI job (no local PowerShell available)


install.ps1
```diff
@@ -25,6 +25,7 @@ $ProgressPreference = "SilentlyContinue" # Invoke-WebRequest progress rendering
 
 $Repo = "https://github.com/Prism-Shadow/penguin-harness"
 $Asset = "penguin-win32-x64.zip"
+$OnlineAsset = "penguin-win32-x64-offline.zip"
```
hiyouga: 这个命名不应该改一下吗？

install.sh
```diff
@@ -65,6 +66,7 @@ if [ "$UNIVERSAL" -eq 0 ]; then
   esac
   TARGET="$os-$arch"
   ASSET="penguin-$TARGET.tar.gz"
+  BUNDLE_ASSET="penguin-$TARGET-offline.tar.gz"
```
hiyouga: 为什么现在还要加 offline 后缀？

laodouuu: 当时想着保留offline的离线包语义，不过去掉offline作为标准产物的设计应该更合理

hiyouga (approved):

PR #145, opened 2026-07-31
feat(core): recover truncated tool output via the Session scratchpad

zzzbitz:
## Summary

Rebuilt from scratch on top of latest `main` (replaces the previous manager-injection design on this branch).

When tool output in an Agent Session exceeds its entry's `maxOutputLength`, Environment now saves the text it actually received into the Session scratchpad (`scratchpad/<sessionId>/truncated-tool-output/`) and appends the recovery path to the same truncated tool result seen by the model and the frontend. Existing truncation semantics are untouched: the Web/CLI stream, the complete `tool_call_output`, Trace, and the next model input remain byte-for-byte aligned, and no model-facing tool or schema is added — the model inspects the file with the existing `read_file` / shell tools.

## Design changes vs. the previous branch state

- **Generic Environment parameter instead of a feature-specific manager.** `EnvironmentConfig` gains one plain string field, `sessionScratchpadDir` — the Session-scoped storage root — and Environment derives `truncated-tool-output/` from it internally. No archive class crosses the public config surface; standalone embedders opt in by passing a directory, or omit it to keep truncation-only behavior. The new `sessionScratchpadDir()` path helper is now the single definition of `scratchpad/<sessionId>`, and `imagesDir` / `goalFilePath` derive from it too.
- **Windows path compatibility on the consuming side.** File tools keep accepting both spellings — `path.resolve` on win32 handles `C:\…` and `C:/…` alike, now pinned by win32-only tests — and tool messages echo the caller's own spelling; the one system-composed path in tool output (`read_file`'s not-found workspace hint) goes through `modelVisiblePath` too. The Web App's file-card path stripping (`toWorkspaceRelative`) previously required the assistant's separators to match the Workspace's exactly; on a Windows Workspace it now compares both sides on `/` with a case-insensitive drive letter, so forward-slash, backslash and mixed spellings all resolve to the same card, while POSIX workspaces still never touch backslashes (legal filename characters).
- **Clean model-visible paths, everywhere.** One helper, `modelVisiblePath` (exported from the core barrel), now covers every path core composes for the model: the system prompt's **App Data Dir / CWD** lines, `[attached image: …]` lines, the server's `[attached file: …]` lines, the goal-file line, and the recovery note. On Windows the spelling uses forward slashes: `exec_command` runs through (Git) Bash and Node's fs APIs accept them, so the model can re-emit the same spelling into JSON tool arguments and shell commands without backslash-escaping mistakes. Session paths are ordinary absolute paths (never `\\?\`-prefixed), so the swap is lossless; POSIX paths pass through unchanged (the previous JSON-quoted literal with UNC/extended-length carve-outs is gone). Marker constructors stay platform-pure — normalization happens at the producer call sites.

## Retained behavior

- Notes: `[output truncated: exceeded N chars]` + `[output archived: <path>]` (or `[output archived (8 MiB limit; head and tail kept): <path>]` / `[output archive failed: <errno>]`), all outside the visible-output cap.
- Per-call archive budget 8 MiB − 1 byte (stays under `read_file`'s 8 MiB scan cap so EOF remains confirmable); larger calls keep UTF-8-safe head/tail windows with an explicit middle-gap marker.
- Bounded capture memory: exact phase stores Buffer chunks (no giant-delta pre-encoding), post-promotion tail is a fixed-capacity ring buffer; cross-delta surrogate pairs are joined, confirmed-lone surrogates encode as U+FFFD.
- Terminal facts are frozen before archive I/O (a late abort cannot reclassify a completed tool); write failures report a short errno code to the result and stderr, never the raw error or path; repeated saves return `ALREADY_SAVED`.
- Files are written only on actual truncation, `0700`/`0600` where the platform supports it, and live/die with the Session scratchpad (no new cleanup lifecycle). Recovery preserves what Environment received; text a producer dropped upstream (e.g. a bounded unread buffer marker) is not recoverable.

Docs updated (`interfaces`, `tools`, `sessions-and-traces`, `configuration`, en/zh). Changelog intentionally not included — it goes into the separate changelog summary PR.

## Verification

- `pnpm -r build`
- `pnpm typecheck`
- `pnpm test` — core 617 passed / 5 skipped (3 win32-only); skills 18; docs 32; landing 39; server 437; cli 206; web 505 — all green
- `pnpm format` + `pnpm format:check`
- Windows-specific assertions (forward-slash recovery path) run in the existing Windows CI job.


packages/core/src/internal/model-visible-path.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Formats a filesystem path as one JSON string literal for model-visible prose.
+ *
+ * On Windows, ordinary paths use forward slashes so the same literal has no backslash-escaping
+ * ambiguity when it moves between JSON tool arguments and shell-oriented discussion. Extended
+ * `\\?\` paths are a Windows API carve-out: changing their prefix to `//?/` would change their
+ * meaning, so they remain byte-for-byte unchanged before JSON quoting. POSIX backslashes are
+ * valid filename characters and must likewise remain untouched.
+ */
+export function modelVisiblePathLiteral(filePath: string): string {
```
hiyouga: 系统提示词里面的都是 unix 路径，只有这里给 windows 比较奇怪？

hiyouga: 说错了，跑了一下，windows 上系统提示词里面是 windows 路径

hiyouga (approved):

hiyouga (commented):
Reviewed in a local worktree at 5cefa10 (full diff vs. base 5308002, all 30 files). Overall: solid design, careful implementation, and unusually thorough tests. No blocking issues found — a few minor observations below.

## Verification (local)

- Full CI-equivalent chain green on Node 24: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (core 617 passed / 5 skipped, server 437, cli 206, web 505, docs 32, skills 18, landing 39), `pnpm format:check` — matches the PR description. GitHub `ci` / `ci-windows` / `build` are green as well.
- Base is 2 commits behind `main` (740f101, 8494988); no changed-file overlap with either, so the squash-merge is semantically clean too.

## What I checked in depth

**Capture correctness and memory bounds** (`truncated-tool-output-archive.ts`)
- Exact phase: the `Buffer.from` stays inside the size guard, so a single huge delta never allocates before rejection; promotion encodes only bounded string slices (`utf8Prefix`/`utf8Suffix` bound the encode by code-unit count before applying the byte cap). Verified.
- Ring tail: `appendTail`'s modulus (`tailBudget`) always equals the allocated capacity since `fileLimitBytes` is fixed per archive and both `promoteToHeadTail` and `tailBudget()` derive the same split; overflow/wrap-around arithmetic and `serializedTail`'s partial-code-point skip check out. The test asserting the archived tail is a suffix of the concatenated 4096-delta pattern pins exactly this.
- Cross-delta surrogate handling (including the consecutive-high-surrogate case) is correct, and the lone-surrogate → U+FFFD policy is documented and tested. The backing-`ArrayBuffer` retention assertion in the huge-delta test is a nice touch — `subarray` aliasing would have passed a naive length check.

**Environment integration** (`environment.ts:273-284`, `335-356`)
- The capture is created on the first delta where `contentLen > maxOutputLength`, and at that point `streamed` is still the exact untruncated prefix (the first possible cut coincides with the first over-limit delta), so `append(streamed)` + per-delta appends reconstruct the complete received text. The stream == complete invariant is untouched and re-pinned at both Environment and engine level.
- Terminal facts (`aborted`) are frozen before the archive `await`, with a regression test that aborts from inside a mocked `save()`. Good.
- Failure path degrades cleanly: errno-only note, stderr warning, original `stop_reason` preserved.

**Path spelling** (`modelVisiblePath` and consumers)
- `sessionEnvironment.cwd` / `projectDir` feed only `assembleSystemPrompt` placeholder substitution, not fs operations — the forward-slash swap is prompt-facing only. Environment/tools keep the raw `workspaceDir`. Verified.
- Web: the `SCRATCHPAD_PATH` regex already accepts both separators, so forward-slash `[attached image/file: …]` lines still resolve to the session-file endpoint; `toWorkspaceRelative` now accepts forward/back/mixed spellings on a Windows workspace while POSIX workspaces still never touch backslashes. Old backslash-spelled traces keep rendering — the dual-spelling acceptance is effectively bidirectional compat.
- Archive filenames are sanitized + hashed (`sha256(toolCallId)`), files `0600`/`wx`, dir `0700`; the 8 MiB − 1 headroom under `read_file`'s scan cap is now pinned by the `scan-cap-minus-one` test. Docs (en/zh) are candid about retention/exposure trade-offs.

## Minor observations (non-blocking)

1. **Comment vs. edge behavior on the compatibility path** — `environment.ts:309-311`. If a legacy tool streams deltas past the cap (capture created from the delta stream) and then emits a full `tool_call_output` that is ≤ `maxOutputLength`, `replace()` doesn't run (it's gated on `toolOutput.length > maxOutputLength`), yet `truncated` is still true via `contentLen > streamed.length` — so the saved archive contains the abandoned delta text while the visible content basis is the complete full message. That's defensible under "the archive preserves what Environment received", but the comment "A compatibility tool's complete message … also becomes the recovery basis instead of any deltas it happened to emit" overpromises for this corner. Cheapest fix is adjusting the comment; alternatively cancel the capture when the basis itself wasn't cut.

2. **Name collision corner in the session scratchpad** — a composer upload named exactly `truncated-tool-output` (extensionless) can occupy the archive root's name (`task-attachments.ts` writes directly into `scratchpad/<sessionId>/`); a later truncation then degrades to `[output archive failed: EEXIST/ENOTDIR]`. The reverse order is harmless (the upload's `wx` retry allocates a suffixed name). Graceful either way, so informational — reserving the name in `scratchpadName` (like the `WINDOWS_RESERVED` prefix rule) would close it if you care.

3. **Changelog follow-up** — when the deferred changelog PR is written, the backward-compat entry should record that `toWorkspaceRelative`'s dual-spelling acceptance is what keeps pre-PR Windows traces (backslash spellings) rendering, and classify it as permanent robustness rather than removable compat (old traces persist indefinitely, and assistants may emit either spelling regardless).

Nice work — the test suite here (late-abort freeze, ring-suffix equality, backing-buffer retention, win32 spelling pins) is a model for this kind of change.

PR #152, opened 2026-08-03
fix(web): queue follow-ups across steer races

### LINKED ISSUE #89: Follow-up task from a reloaded page occasionally never runs (202 accepted, no execution)
Surfaced and root-isolated while stabilizing PR #78's e2e (see its final comment for the distilled repro): a follow-up task sent from a **reloaded** page intermittently goes dead under load (~1 in 4–6):

- `POST /tasks` answers 202, but the task never runs — trace-verified: no LLM call, no `task_state` change, no messages; or
- the input area is left in a stale running/steer state (a steer attempt 409s and the fallback detour strands the composer).

Reproduced with a Playwright script that never touches the subagents panel, and identically with #78's core input-forwarding change reverted — pre-existing session-runtime/stream behavior, independent of that PR. Suspect area: the post-reload stream/runtime handshake (steer-vs-new-task routing against a stale runtime entry).

March-77:
## Summary

- route the Web composer's `not_running` steering fallback through the server's existing `queueIfBusy` Task path
- keep the complete draft intact across the fallback, including skills and attachments
- add a Playwright regression that reloads a running Session, forces the stale steering response, verifies `202 queued`, and waits for the follow-up's model output

## Root cause

After a reload, the SSE snapshot can briefly leave the composer in running/steer mode while the core run is ending. If `/steer` returns `not_running`, retrying with a plain Task POST can reach the runtime before its idle transition and return `task_in_progress`, stranding the draft. `queueIfBusy` is safe on both sides of that boundary: it starts immediately when idle or queues behind the last part of the old run.

## Validation

- `pnpm -r build`
- `pnpm typecheck`
- `pnpm test` (all workspace unit suites green)
- `pnpm format:check`
- `playwright e2e/reloaded-followup.spec.mjs` (passed with a real server and mock LLM)

Closes #89

### DIFF
--- a/changelog/unreleased/2026-08-03-reloaded-followup-race.md
+++ b/changelog/unreleased/2026-08-03-reloaded-followup-race.md
@@ -1,0 +1,5 @@
+# Reloaded follow-ups survive the steering completion race
+
+When a reloaded page still showed a Session as running, the composer first tried the steering endpoint. If the core run ended in that narrow window, `/steer` returned `not_running` and the Web App retried the draft as a plain Task. That second request could arrive before the server runtime completed its own idle transition, return `task_in_progress`, and leave the draft unsent.
+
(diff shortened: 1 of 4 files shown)


hiyouga:

PR #154, opened 2026-08-03
feat(server,web): steer with file attachments

### LINKED ISSUE #140: Multiple steering message hints
在 steering 模式下，只发送文件和只发送图片会出现不同的提示：
```
插话已排队，将随下一轮送达

1 条跟进消息已排队，本轮结束后自动发送
```

只发送文件应该也是插话才对，不应该是跟进

另外最好在界面能看到 steering message 的内容

March-77:
## Summary

- extend `POST /api/sessions/:id/steer` with task-style file attachments, including file-only steering
- preflight the runtime before writing uploads and remove written files if the steering race is lost
- send attached files from the running Web composer while preserving the complete draft on `not_running`
- document the API and Web behavior in English and Chinese

## Root cause

The ordinary Task path already had a confined scratchpad attachment pipeline, but the steering contract accepted only text and images. Reusing that pipeline requires an advisory running-state check before disk writes plus cleanup after the authoritative steering check, otherwise a completion race can leave orphaned files.

## Validation

- `pnpm -r build`
- `pnpm typecheck`
- `pnpm test` (server 439 passed, Web 505 passed, all workspace unit suites green)
- `pnpm format:check`

Closes #140


hiyouga:

PR #169, opened 2026-08-04
fix(cli): report web readiness probe failures

laodouuu:
## Summary

- A Windows installation printed `Server is not responding yet; open http://127.0.0.1:7364/ manually once it is ready.` after `penguin web` had already reported that the local server started.
- Investigation showed that the readiness probe used an empty `catch` block, discarding every `fetch` exception and reducing connection timeout, refusal, reset, permission, and DNS failures to the same generic message. On the affected computer, port 7364 was listening but loopback requests timed out; the service became reachable after the local firewall was disabled, confirming that firewall filtering was the cause in that case.
- This PR preserves the last probe error and reports an actionable, localized diagnosis. For connection timeouts it asks the user to allow PenguinHarness to communicate on the configured local port; it does not instruct users to disable their firewall.

## Changes

- Return a structured readiness result and retain the final nested Node/undici error code and message.
- Classify timeout, connection-refused, reset/socket, permission, DNS, and unknown failures.
- Write failed readiness diagnostics to stderr while preserving the existing rule that any HTTP response counts as ready.
- Add English and Chinese guidance, including `请允许 PenguinHarness 在本机端口 7364 上通信。` for the default timeout case.
- Add focused tests for successful responses, nested timeout causes, error classification, and localized prompts.

## Test plan

- [x] `pnpm format:check`
- [x] `pnpm build`
- [x] `pnpm typecheck`
- [x] CLI tests: 205 passed, 8 skipped
- [x] Web tests: 532 passed
- [x] Docs, landing, and skills tests: 89 passed
- [x] Core timing test that failed under full parallel load passed on focused rerun (17/17)
- [ ] Six Windows symlink-dependent Core/Server tests require symlink privileges and reported `EPERM`; all other Core/Server tests passed
- [x] `git diff --check`

Generated with Codex.


hiyouga:
Reviewed; pushed a follow-up commit 7ca18df addressing three points found during review:

- **Keep the detail line useful for `AggregateError` failures.** On Node ≥ 24, a refused probe against a multi-address host (e.g. `--host localhost`) rejects with `TypeError("fetch failed") → AggregateError { code: "ECONNREFUSED", message: "" }` (verified empirically on Node 24), which degraded the detail to `ECONNREFUSED: Unknown readiness probe error`. `describeReadinessFailure` now walks `AggregateError.errors[0]` and skips empty messages, so it reports the underlying `connect ECONNREFUSED …` instead. The default `127.0.0.1` path was already fine.
- **Regression tests** for the keep-polling→ready recovery path, and for the probe's real timeout shape: the per-probe 1s `AbortSignal` always fires before undici's 10s connect timeout, so the field failure is `DOMException("…", "TimeoutError")` rather than `UND_ERR_CONNECT_TIMEOUT`; both shapes are now covered, plus the AggregateError case above.
- **Convention cleanups**: doc comments on the new exports, dropped the redundant `ReadinessFailureKind` alias in favor of `WebProbeFailureKind`.

Also verified the original diagnostics end-to-end with real failures (closed port → `refused`, accept-but-never-respond server → `timeout` with the firewall hint, `.invalid` host → `dns`) — all classify correctly.

Verification: `pnpm -r build` · `pnpm typecheck` · `pnpm format:check` · full `pnpm test` (cli 216, all packages green).

hiyouga (approved):

PR #174, opened 2026-08-04
fix(core,server,web,cli): compaction tolerance, retry parity, and failure classification

### LINKED ISSUE #170: [BUG] 频繁压缩失败
模型供应商opencode，模型为deepseek-v4-flash，mac系统

STR:
长上下文，反复的压缩都失败了。没法进行工作，因为调用一次就自动压缩

Image

他自己简单分析了一下原因，仅供参考
根本原因
1. 模型侧（主因）：deepseek-v4-flash 系统性地不把摘要放进标签里——它把 [summary]/[/summary] 当成"标题"先写出来，再把正文写在闭合标签之后。第一次是双开标签 [summary]\n[summary][/summary]，之后全是空标签 [summary]\n[/summary]。它的思考里也显示了困惑（_003:476）：上下文里已经有一个旧的 [summary][/summary] 块，它搞不清是"写进那个块"还是"新写一个"，最终都选择了"标签在前、内容在后"。

2. harness 侧（把问题变成致命的）：

提取是朴素的：取第一对标签之间的文本；内容在标签外时静默丢弃，没有"标签内为空就用完整回复兜底"的逻辑；
校验不一致：第一次压缩时空摘要被当成功（completed），后面又变成校验并重试——行为不统一；
重试无效且烧 token：每次重试模型看到自己上一轮的错误输出，只会复制同样的格式；一次压缩 5 次重试 × 每次约 3000 output tokens，白烧了几万 token；
没有修复机制：摘要内容其实一直在模型回复里（只是位置错了），但 harness 从不去捞；最后只能中止会话。
修复方向（如果让我改）
提取兜底：标签内为空/过短时，回退到取"最后一个 [summary] 之后的所有文本"或整段模型回复；
生成侧引导：压缩提示里明确"内容必须写在开标签之后、闭标签之前，闭标签后不要再写任何内容"；或改用结构化输出/后处理拼接；
重试前清理：重试时不要把上一次的失败输出留在上下文里，避免模型照着学；
校验统一：成功/失败判定一致（按提取结果是否非空，而不是按请求是否完成）。

hiyouga:
Fixes #170.

## Summary

Issue #170: with deepseek-v4-flash, every compaction failed — the model writes `[summary]\n[/summary]` as a "title" and puts the body **after** the closing tag; extraction took the (empty) text between the first tag pair, each retry re-showed the model its own committed bad output (which it copied verbatim), and the burned attempts were invisible in the stats. The session re-triggered a failing compaction on every turn and became unusable.

## Changes

**Extraction tolerance (core)** — `extractSummary` now applies a ladder: ① the first `[summary]…[/summary]` pair whose content is non-empty once stray inner tags are stripped, ② otherwise whatever text remains after stripping the empty summary blocks/tags (rescues the issue's exact shape on the *first* attempt), ③ tagless output stays used verbatim. Legacy `<summary>` stays accepted. Step ① preserves the historical result for every healthy output, so resume re-extracts old Traces byte-identically.

**One failure model, standard retries (core)** — a compaction is an ordinary LLM request. A committed-but-unusable response (empty summary, or tool calls) counts as one more **failed attempt** on the same `compactionMaxReconnects` budget (defaulting to the shared `maxReconnects`) and the same exponential backoff ladder as the transport statuses (`failed`/`timeout`/`malformed`); only `auth` stops without retrying; once the budget is exhausted the compaction fails and the original context is kept. Unusable-response retries still repair the tool_use pairing (synthesized failed outputs) and prepend a corrective note before the re-sent Prompt — the committed prefix is append-only, so correcting forward is the prompt-cache-safe fix for the copy-my-own-mistake loop (issue #84's invariants untouched).

**Cost visibility (core + server)** —
- `compaction_end` gains optional `attempts` and `output_tokens` (accounting across all attempts, failed ones included).
- Every attempt's `token_usage` is now surfaced between the compaction event pair (previously only the adopted attempt's), so frontend stats and `usage_records` carry the compaction's true spend.
- `StreamErrorWatcher` records a failed compaction as one cost-center error row: source `compaction`, code `compaction_failed`, message carrying attempts / burned output tokens / trigger. No double-counting: compaction request events are Trace-only and never hit the llm-source records.

**Prompt simplification (new agents only)** — the default compaction prompt is shorter and shows the format as a concrete example (`[summary]put the summary text here...[/summary]`) instead of describing tag placement in prose; the retry guidance uses the same example. `compaction.prompt` is persisted per-agent, so existing agents keep their stored prompt — they are covered by the extraction rescue and retry guidance instead.

**Retry observability (core + server + web + cli + docs)** — a shared `RetryDetail` block (`error_message` / `attempt` / `retry_in_ms`), stamped by the builders (the way `withOrigin` stamps `origin`) instead of accreting positional parameters; `RequestEndPayload` and `CompactionEndPayload` both extend it. The former `message` field is renamed `error_message`, one name across the stack (`LLMOutcome.errorMessage` internally) — in this protocol plain "message" means an OmniMessage/model output, which this field is not. `attempt` is the authoritative 1-based ordinal within the retry run (stamped on failures and on completions that needed retries; absent on a clean first try), and the CLI/Web retry displays read it directly — no client-side counting. `compaction_end.attempt` is the final attempt ordinal and `compaction_end.error_message` the last failure detail, shown on the web banner, the CLI line, and the cost-center error row (`output_tokens` dropped — usage_records already carry the full spend). Docs updated (omni-message payload block, agent-loop retry and compaction sections).

## Backward compatibility

- New `compaction_end`/`request_end` RetryDetail fields (`attempt`, `error_message`) are additive and optional: old Traces replay unchanged, no migration, no user action (pre-field retry notices render as attempt 1). The pre-PR released `request_end.message` field is superseded by `error_message`: old Traces keep `message` in their JSON, and nothing reads it semantically after the fact (error rows are recorded live at same-version), so no dual-read code is kept.
- Extraction: healthy old Traces re-extract identically (first-non-empty-pair preserved); Traces where an old build committed an empty summary now resume with the salvaged content instead of an empty `[context_summary]` — a strict improvement.
- New error source `compaction` is a new value in an existing TEXT column; no schema migration.
- No new long-lived compat code is introduced (the legacy `<summary>` acceptance predates this PR and stays per its documented policy).

## Verification

- [x] `pnpm install --frozen-lockfile`
- [x] `pnpm -r build`
- [x] `pnpm typecheck`
- [x] `pnpm format` + `pnpm format:check`
- [x] `pnpm test` — core 624 passed (incl. outside-the-tags rescue e2e, unified-budget counting pins, shared-default retry budget, explicit-override precedence), server 447 passed (incl. compaction error-row record, completed/aborted not recorded, old-core event fallback, child-session attribution), web 533 passed, cli 216 passed, docs/landing/skills all passed


hiyouga:
Revised per review (35542c7): dropped the per-cause `failure_cause` classification and unified the failure model — a compaction is treated as an ordinary LLM request. Empty-summary and tool-calling responses now count as failed attempts on the same reconnect budget and backoff ladder as transport failures; only `auth` stops immediately; the compaction fails once the budget is exhausted. `compaction_end` keeps `attempts`/`output_tokens`; the server records a single `compaction_failed` error row whose message still carries attempts and burned tokens; web/CLI show the plain failed/aborted line again. PR description updated; full chain re-verified (same totals: core 624 / server 447 / web 533 / cli 216).

PR #185, opened 2026-08-04
fix(web): align uploaded attachments with user messages

laodouuu:
## Summary

- Fix uploaded file attachments rendering on the left as a generic notice even though they belong to a user message.
- Place the attachment below the user text and reuse the same right-aligned container and timestamp behavior as uploaded images.
- Keep the existing filename truncation and light/dark presentation.

This is intentionally a UI design interpretation. If it does not match the intended design, please feel free to close this PR.

## Changes

- Make `AttachedFilesBanner` presentation-only so its caller controls alignment, animation, and metadata.
- Render uploaded file attachments after the user text with the same `items-end` layout used by uploaded images.
- Reuse the owning user message's `atMs` value for the attachment timestamp.

## Test plan

- [x] `pnpm build`
- [x] `pnpm format:check`
- [x] `pnpm typecheck`
- [x] `pnpm --filter @prismshadow/penguin-web test` (41 files, 533 tests)
- [ ] `pnpm test` — the final tree reaches 623 passing and 5 skipped Core tests, with one existing Windows environment failure because symlink creation returns `EPERM` in `packages/core/test/file-tools.test.ts`

Generated with [Codex](https://openai.com/codex/)

### DIFF
--- a/packages/web/src/features/chat/attached-files-banner.tsx
+++ b/packages/web/src/features/chat/attached-files-banner.tsx
@@ -3,8 +3,8 @@
  * server appends aren't shown verbatim, they collapse into a single line reading
  * "Attached files: a.pdf, b.csv" (paperclip icon + static text, no navigation — the files live
  * in the session scratchpad and the model opens them by path); the body text around them is
- * rendered as usual by the caller. Same shape as SkillsBanner, so the two notices a message
- * can carry read as one family.
+ * rendered as usual by the caller. It keeps the same visual language as message-level notices,
+ * but aligns right because an uploaded file belongs to the user's input.
  */
 import { S } from "../../lib/strings";
 import { attachmentFileName } from "../../lib/attachments";
@@ -22,7 +22,7 @@ export function AttachedFilesBanner({ files }: { files: string[] }) {
     // into a paragraph-tall block above the message. The full list stays reachable as a title.
     <p
       title={label}
-      className="anim-msg my-2 flex w-fit max-w-full items-center gap-2 rounded-md border border-gray-200 bg-gray-50 px-3 py-2 text-xs text-gray-600 dark:border-gray-800 dark:bg-gray-900 dark:text-gray-400"
+      className="anim-msg my-2 ml-auto flex w-fit max-w-full items-center gap-2 rounded-md border border-gray-200 bg-gray-50 px-3 py-2 text-xs text-gray-600 dark:border-gray-800 dark:bg-gray-900 dark:text-gray-400"
     >
       <GlyphIcon d={PAPERCLIP_ICON} className="shrink-0 text-gray-400 dark:text-gray-500" />
       <span className="min-w-0 truncate">{label}</span>
--- a/packages/web/src/features/chat/attached-files-banner.tsx
+++ b/packages/web/src/features/chat/attached-files-banner.tsx
@@ -3,8 +3,8 @@
  * server appends aren't shown verbatim, they collapse into a single line reading
  * "Attached files: a.pdf, b.csv" (paperclip icon + static text, no navigation — the files live
  * in the session scratchpad and the model opens them by path); the body text around them is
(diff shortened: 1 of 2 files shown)


hiyouga:
## Review

Reviewed from a clean worktree at ca30646, diffed against `origin/main`.

**What it does**: uploaded file attachments on a user message stop rendering as a left-aligned message-level notice above the bubble; they now render below the user text in the same right-aligned container as uploaded images, with the same hover timestamp footer. `AttachedFilesBanner` becomes presentation-only (its `anim-msg my-2` moves to the caller's wrapper), and the sole call site supplies alignment, animation, and `MessageMeta`.

### The design question (the real decision here)

The old placement was not an accident: `design/specs/06-PROTOTYPE.md` (input-area section) specifies that attached-file rows are "对话中折叠渲染为「附加文件」提示（同 `[use_skills]` 口径）" — the same treatment as the skills banner, i.e. the left-aligned notice family — and the deleted code comment said exactly that ("Same shape as SkillsBanner… one family"). So as the PR itself anticipates, this is a deliberate design change, and in this repo the design docs are ground truth: it needs a design-side decision and a matching `design/` update (06 prototype line + an 08 changelog entry) before or together with the merge — not a silent divergence.

On the merits, I think the new interpretation is defensible and arguably right: uploaded files are the user's *content* (like uploaded images, which already render right-aligned), while the scheduled/skills banners describe how the message was produced. The old asymmetry — user's images on the right, user's files on the left in system-notice chrome — was the odd one out. Recommend: if the maintainer agrees, update the design spec and land this; if not, close per the author's own note.

### Code

The mechanics are sound; I verified each point in the worktree:

- `AttachedFilesBanner` has exactly one caller (`message-item.tsx`), so making it presentation-only regresses no other call site, and the animation/margin removed from the banner reappear on the new wrapper (`anim-msg group my-4`, matching the sibling text/image blocks).
- The new files block mirrors the images block exactly: `items-end` container, `MessageMeta` with `align="right"`, the same `{...(item.atMs !== undefined ? { atMs } : {})}` optional-prop idiom, and no `text` prop (so no copy affordance — same as images).
- Render order becomes text → files → images; the goal-round branch (text and images only, by design) and the steering branch are untouched.
- A files-only message (spec: 只带附件、没有正文也可发送) now reads as user content with a timestamp instead of a floating system notice — the case that motivated the PR.
- Comments updated accurately, English-only, consistent style.

### Tests

- Full web suite in the worktree: 41 files, 533/533 pass — matches the PR description.
- No unit test asserts attachment placement before or after; that matches the surrounding code (image placement is equally untested), so I don't consider a new test mandatory for a pure-presentation change.

### Adjacent gap (pre-existing, not caused by this PR)

While verifying, I noticed the `user_steering` branch destructures only `text`/`images` from `splitAttachments` and silently drops `files` — yet the same spec section says steering can carry file attachments ("文件按发送附件的同一规则落盘、以 `[attached file: <路径>]` 行送达，只发附件也可插话"). A files-only steering message renders an empty chip and the filenames appear nowhere in the flow. Worth a follow-up issue if confirmed intended-to-show.

### Process

- No `changelog/unreleased/` entry — assuming the separate-batch convention, flagging so the Web App entry picks this up.

**Verdict**: implementation is clean and correct for what it intends; the only blocking item is the design call — this contradicts the current `design/specs` wording ("同 `[use_skills]` 口径"), so it should merge only together with a design-doc update saying attachments render as user content, or be closed if the original notice-family design is upheld.

hiyouga:
Pushed f179849 addressing the adjacent gap from the review above: the `user_steering` branch discarded the `files` split, so a files-only steering message rendered as an empty chip and the `[attached file: …]` rows vanished from the flow. The chip now collapses them into the same `AttachedFilesBanner` a full prompt uses, right-aligned inside the chip below the images row. Verified in a worktree: web suite 41 files / 533 tests pass, web `tsc --noEmit` clean, prettier applied.

hiyouga (approved):

PR #196, opened 2026-08-04
feat(cli): add update download source selection

laodouuu:
## Summary

- make `penguin update` reuse the stable installer forwarder's download-source policy instead of requiring GitHub at the start of every upgrade
- prefer the OSS `latest.json` and immutable release in `auto`, with same-tag GitHub fallback, while keeping forced `oss` and `github` modes strict
- preserve explicit HTTPS mirror precedence, clear stale fallback settings, and localize source-selection failures

## Changes

- add OSS manifest validation, immutable release resolution, and installer candidate fallback to the CLI updater
- pass the selected payload base and optional fallback into `install.sh`
- document the source-selection environment contract in both CLI guides
- cover source modes, metadata fallback, pinned releases, custom mirrors, stale fallback clearing, and localized errors

## Testing

- `pnpm format:check`
- `pnpm typecheck`
- `pnpm -r build`
- Windows package suites: 1,938 passed and 14 skipped; six symlink cases were blocked by Windows `EPERM`
- Linux container follow-up: Core 624 passed / 5 skipped; Server 471 passed
- post-rebase CLI suite: 227 passed / 8 skipped
- Docker black-box probes for `oss`, `github`, `auto`, and explicit mirror precedence

### DIFF
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -11,12 +11,14 @@
  * made — never by guessing. A source checkout is refused outright: overwriting a working tree
  * would destroy uncommitted work.
  *
- * The latest version comes from the GitHub Releases API, the same source of truth install.sh
- * resolves `releases/latest/download` against. The target-a-specific-release flag is spelled
- * `--release <tag>` rather than `--version <tag>`: commander's program-level `-v, --version`
- * intercepts a subcommand's own `--version` when it is written with a space, so
- * `penguin update --version 0.1.2` would silently print the CLI version and do nothing. A flag
- * that works only in its `--version=0.1.2` form is a trap, so it got an unambiguous name.
+ * Release discovery and installer download use the same environment contract as the public
+ * installer entry point: an explicit PENGUIN_DOWNLOAD_BASE_URL has highest download priority;
+ * otherwise auto prefers the OSS latest pointer and immutable release, then falls back to the same
+ * GitHub tag, while oss and github are strict. The target-a-specific-release flag is spelled
+ * `--release <tag>` rather than `--version <tag>`: commander's program-level
+ * `-v, --version` intercepts a subcommand's own `--version` when it is written with a space, so
+ * `penguin update --version 0.1.2` would silently print the CLI version and do nothing. A flag that
+ * works only in its `--version=0.1.2` form is a trap, so it got an unambiguous name.
  *
  * Self-replacement hazard, and how it is handled: for a tarball install the installer deletes and
  * replaces `lib/`, which is the directory this very process is executing from. Two things make
@@ -60,7 +62,7 @@ import { fileURLToPath } from "node:url";
 import { realpathSync } from "node:fs";
 import { VERSION, compareVersions, normalizeVersion } from "@prismshadow/penguin-core";
 import type { Command } from "commander";
-import type { Messages } from "../i18n.js";
+import type { InstallerSource, Messages } from "../i18n.js";
 
 // The version helpers live in core (internal/version.ts) so the server's update-check
 // endpoint shares them; re-exported because they are part of this module's public,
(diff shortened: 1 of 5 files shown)


hiyouga:
## Review

Reviewed from a clean worktree at 5b7a08d, diffed against `origin/main`.

**What it does**: `penguin update` now resolves the target release and downloads the installer with the same source policy as the stable installer forwarder — `PENGUIN_DOWNLOAD_SOURCE=auto|oss|github`, `auto` preferring the OSS `latest.json` and its immutable release with a same-tag GitHub fallback, forced modes strict, explicit HTTPS mirrors first, stale fallback env cleared.

### Contract parity — verified against both shell entry points

I compared the TS logic point-by-point with `install.sh` and `packages/landing/public/install.sh`; they agree:

- Source modes and the error text (`must be auto, oss, or github`) match `install.sh:129-132`.
- `parseOssLatestManifest` mirrors the forwarder exactly: `schemaVersion === 1`, tag shape identical to `is_release_tag` (`^v[0-9A-Za-z][0-9A-Za-z._-]*$`), and `releaseBaseUrl` pinned to the fixed bucket base.
- `auto` semantics match: OSS discovery → same-tag GitHub transport fallback; when OSS *metadata* failed, auto stays on GitHub with no fallback — same as the forwarder's `use_github` path clearing `FALLBACK_BASE`.
- Explicit `PENGUIN_DOWNLOAD_BASE_URL` outranks source selection, HTTPS-only, validated before anything is downloaded.
- Clearing a stale fallback by passing `PENGUIN_DOWNLOAD_FALLBACK_BASE_URL=""` is safe: `install.sh:141-144` guards with `[ -n ... ]` before validating, so the empty string reads as "absent" rather than "invalid URL". This was the one spot where the new env contract could have broken the child installer; it doesn't.

A nice improvement beyond the title: `version: target` is now always passed to the child installer, so the latest flow is pinned to the resolved tag instead of the mutable `releases/latest/download` — the version shown at the confirmation prompt is exactly what gets installed, even if a new release lands mid-upgrade.

### Tests

- `test/update.test.ts`: 56/56 pass in the worktree. The `FetchLike` injection keeps everything offline, and coverage is thorough — source parsing, HTTPS validation, manifest validation, all three modes, pinned releases, transport fallback order, env clearing, and the localized message shapes for every source combination that can actually occur.
- Full CLI suite: 212/212 collected tests pass; `test/serve.test.ts` failed to *collect* only because my worktree's `@prismshadow/penguin-server` dist wasn't fully built — that file is untouched by this PR, and 212 plus its 15 matches the 227 reported in the description.

### Minor issues

1. **Lost diagnostic in the download-failure message.** The old `installerFetchFailed` printed the failing URL; now `penguin update --release 9.9.9` (nonexistent tag → all candidates 404) prints "Could not download the installer from the OSS mirror or GitHub. Check your network and retry." — pointing at the network when the actual cause is the tag. Suggest including `release.tag` in the message (e.g. "…the installer for v9.9.9 from…") so a bad `--release` stays self-diagnosing.
2. **`installerUrl()` is now production-dead.** After this change only its own tests reference it, and the no-version branch (`releases/latest/download/install.sh`) is unreachable from the command. Suggest removing it together with its describe block, or noting why it stays exported.
3. Nit: the zh `installerFetchFailed` spacing (`join("或 ")` plus the boundary regexes) is correct for every combination that exists today and the tests pin them all, but it silently relies on GitHub being the only Latin-edged source name — worth revisiting if a source is ever added.

### Process

- **Design sync**: I couldn't find this behavior in `design/specs/` — the only `PENGUIN_DOWNLOAD_SOURCE` mention is the desktop electron-updater note in 06-PROTOTYPE.md, and recent design-repo commits are all desktop packaging. Per the design-and-code-in-sync rule, is there a companion design change covering `penguin update`'s source selection (05 install/release section + an 08 changelog entry)?
- **Changelog**: no `changelog/unreleased/` entry in this PR — assuming the separate-batch convention, just flagging so the batch extends the existing CLI entry for this surface.

**Verdict**: LGTM aside from the minor items above; nothing blocking. Docs (en/zh) and i18n are updated in sync, and the security posture improves — HTTPS-only mirror bases, the manifest pinned to the fixed bucket base, and immutable same-tag fallback throughout.

hiyouga (approved):

PR #209, opened 2026-08-04
fix(web): preserve Benchmark selection

### LINKED ISSUE #181: fix(web): preserve Benchmark selection after opening a Session trace
## Description

The Evaluation Center loses its selected Benchmark after navigating to a Session trace.

## Reproduction

1. Open Evaluation Center and select an Agent and Benchmark.
2. Expand an evaluation, Case, and Run.
3. Click the Session link to open Trace Observability.
4. Click Evaluation Center in the sidebar.

The page returns to the empty default state instead of the previously selected Benchmark.

## Expected

Restore the last selected Agent and Benchmark for the current Project when returning to Evaluation Center.

## Code context

BenchmarkPage keeps selection only in component state. Navigating to the trace page unmounts it, while the sidebar returns to the bare /benchmark route. The existing agentId query parameter controls only which Agent group is expanded.

Observed on v0.2.0 at e05dea5.

taking-lying-flat:
## Summary

- remember the selected Agent and Benchmark per user and Project
- resolve remembered ids against fresh Benchmark data and ignore invalid or stale browser storage
- preserve the existing `?agentId=` deep-link behavior and reset lazy Agent data across Projects
- add unit and browser regression coverage for the Evaluation Center → Session trace → Evaluation Center flow

Closes #181

## Root cause

`BenchmarkPage` kept its selection only in component state. Opening a Run's Session trace unmounted the page, and the sidebar returned to bare `/benchmark`, so the selection was recreated as `null`.

## Validation

- `pnpm build`
- `pnpm format:check`
- `pnpm typecheck`
- `pnpm --filter @prismshadow/penguin-web test` — 595 passed
- Playwright `packages/web/e2e/benchmark.spec.mjs` — 2 passed

## Existing suite notes

- `pnpm test` reaches two pre-existing `packages/landing/test/blog.test.ts` failures: `main` contains the 0.2.1 blog post while those assertions still expect 14 posts and omit its slug.
- The complete Web E2E suite also reached two unrelated existing test failures; the new regression passes independently.

### DIFF
--- a/changelog/unreleased/2026-08-05-benchmark-selection.md
+++ b/changelog/unreleased/2026-08-05-benchmark-selection.md
@@ -1,0 +1,5 @@
+# Evaluation Center restores the selected Benchmark
+
+The Evaluation Center now remembers the last selected Agent and Benchmark for each user and Project, so following a Run's Session link into Trace Observability and returning through the sidebar restores the Benchmark instead of showing the empty initial state.
+
+Only the stable Agent and Benchmark ids are kept in browser storage. The current Benchmark summary is fetched again before restoration, invalid or stale selections are ignored, account and Project changes are isolated, and explicit `?agentId=` deep links retain their existing focus behavior. Unit coverage exercises storage validation and isolation, while a browser regression test follows the original Evaluation Center → Session trace → Evaluation Center path.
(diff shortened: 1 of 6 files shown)


hiyouga:

PR #216, opened 2026-08-06
feat(core): wire MCP servers into Environment via adapter layer

field-claw:
## Summary
PenguinHarness declared `mcpServers` in tool config but never connected them (the docs note "enumerating concrete MCP tools is reserved for a later adapter layer"). This PR implements that layer so configured MCP servers become real, LLM-callable tools.

## What changed
- **New `McpToolAdapter`** (`packages/core/src/environment/mcp/client.ts`): connects each declared stdio/SSE MCP server, enumerates tools, and wraps them as `BuiltinTool` instances.
  - Tool names namespaced as `mcp__<server>__<tool>` (collision-safe).
  - MCP `inputSchema` (JSON Schema) passes straight into `ToolDefinitionConfig.parameters` — both are `Record<string,unknown>`, **no conversion layer**.
  - Per-server connect failures are isolated (logged to stderr, don't abort the Environment).
- **`Environment`** gains a `static async create()`: when `mcpServers` is non-empty it connects them and registers their tools. The sync constructor is preserved (callers without MCP are unaffected).
- **`agent.ts`** now uses `Environment.create()` so the real agent loop can reach MCP tools.
- **Abort forwarding**: `ToolExecutionContext.signal` is forwarded into the SDK `callTool` `options.signal`, so user-interrupt / Environment timeout cancels an in-flight MCP request.
- Adds `@modelcontextprotocol/sdk` (MIT/Apache-2.0, license-compatible with Apache-2.0).

## Verification
- `pnpm --filter @prismshadow/penguin-core typecheck` → passes.
- Unit test (mocked SDK) `test/mcp-client.test.ts` → 3/3 green (namespacing, inputSchema passthrough, BuiltinTool delegation, fault isolation).
- **Real e2e** `test/mcp-e2e.mts` (spins up a live local stdio MCP server via `tsx`, drives `Environment.create → listTools → executeTool`) → reaches and calls the tool, streams output, finalizes `completed`.

## Backward compatibility
Callers that don't configure `mcpServers` continue using the sync `new Environment(...)` with zero changes. Only the runtime agent path switched to `await Environment.create(...)`.

## Notes / follow-ups (out of scope)
- CLI/Web config UI for `mcpServers` not added; JSON edit + docs suffices for now.
- `sse` transport is wired in `buildTransport` (client-side `SSEClientTransport`) but is **not** covered by a live e2e — server-side SSE was removed from the MCP spec and the SDK's `SSEServerTransport` is unstable to stand up as a test fixture. The stdio path is fully e2e-tested.
- `signal` is now forwarded into `callTool`; `Environment` continues to enforce timeout/abort on the tool stream as a backstop.

Generated with Claude as a PenguinHarness MCP evaluation follow-up.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -40,6 +40,7 @@
     "build": "tsup"
   },
   "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.0.0",
     "@prismshadow/agenthub": "^0.4.1",
     "@prismshadow/penguin-skills": "workspace:*",
     "smol-toml": "^1.3.0",
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -711,7 +711,7 @@ export class Agent {
     // into command subprocesses (shared by createSession and resumeSession; the
     // caller reads the current agent_state/.vault.toml); a child Agent loads **its
     // own** vault via createAgent rather than inheriting the parent's.
-    const environment = new Environment({
+    const environment = await Environment.create({
       workspaceDir,
       toolConfig,
       // The Session's generic scratchpad root; Environment derives its truncated-tool-output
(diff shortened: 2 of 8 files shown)


hiyouga:

PR #217, opened 2026-08-06
feat(web): make MCP Servers editable in agent settings

field-claw:
﻿## Summary
Makes the agent settings page's **MCP Servers** section editable (it was a read-only JSON dump). Pairs with the core MCP adapter in PR #216 — this UI lets users configure the servers that adapter connects at agent run time.

## What changed
- Replaced the read-only <pre> with a per-server editor: name input + config JSON textarea, add/remove rows.
- Live JSON parse validation per row (inline error if the config isn't valid JSON).
- submit diffs the edited list against the loaded config and only sends mcpServers when changed (PUT config's partial-update model).
- config stays free-form JSON so users paste the exact stdio/sse shape the core adapter consumes — no UI-side schema assumptions.
- i18n (en + zh) updated; removed the "(read-only)" label.

## Notes
- Web typecheck shows pre-existing sibling-package subpath resolution noise (@prismshadow/penguin-core/omnimessage etc.) that resolves under a full monorepo build; gent-settings-page.tsx itself is clean.
- Not browser-verified in this environment (no display); logic is straightforward controlled inputs.

Generated with Claude as a PenguinHarness MCP evaluation follow-up.

### DIFF
--- a/packages/web/src/features/agents/agent-settings-page.tsx
+++ b/packages/web/src/features/agents/agent-settings-page.tsx
@@ -3,7 +3,7 @@
  * Overview (name/description/State path/active count/State version + snapshot
  * export-import + restore default configuration), Prompt (AGENTS.md and system_prompt editors + placeholder
  * reference), Runtime (max_turns, model.*, compaction.*), Tools (editable built-in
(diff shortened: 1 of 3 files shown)


hiyouga:

PR #227, opened 2026-08-06
fix(web): queue follow-ups across steering completion races

### LINKED ISSUE #89: Follow-up task from a reloaded page occasionally never runs (202 accepted, no execution)

March-77:
Closes #89.

## Problem

After reloading a running Session, the Web client can briefly retain a stale
`running` state while the Task is completing.

If the user sends a message in that window, the composer first calls `/steer`.
When `/steer` returns `409 not_running`, the current fallback retries the draft
with a plain `POST /tasks`.

That request can reach the server before the runtime finishes its idle
transition, receive `409 task_in_progress`, and leave the draft unsent.

## Fix

Route the `not_running` fallback through the existing `queueIfBusy` Task path.

This path is safe on both sides of the completion boundary:

- when the Session is idle, the Task starts immediately;
- when the runtime is still finishing, the Task is queued and starts afterward.

The fallback still uses the normal draft composition path, so text, selected
Skills, images, and file attachments are preserved.

No server behavior or API was added.

## Tests

Added a deterministic Playwright regression that:

- keeps the first Task running on tool approval;
- reloads the Session;
- forces `/steer` to return `409 not_running`;
- verifies the fallback request contains `queueIfBusy: true`;
- verifies the server returns `202 { queued: true }`;
- completes the first Task and waits for the follow-up's model response.

Local verification:

- `pnpm -r build`
- `pnpm format:check`
- `pnpm typecheck`
- `pnpm test` — 2,071 passed, 5 skipped
- Server follow-up and steering tests — 8 passed
- targeted `reloaded-followup.spec.mjs` Playwright regression — passed

Implemented with agent assistance. I reviewed the state transition, final diff,
and regression assertions before submission.

### DIFF
--- a/packages/web/e2e/reloaded-followup.spec.mjs
+++ b/packages/web/e2e/reloaded-followup.spec.mjs
@@ -1,0 +1,97 @@
+/** Regression for #89: a Task posted after reloading a completed Session must execute. */
+import { test, expect } from "@playwright/test";
+import { provisionAndLogin } from "./auth.mjs";
+
+const BASE = process.env.BASE_URL;
+const MOCK = process.env.MOCK_URL;
+const U = "reloadedfollowupuser";
+const P = "password123";
+
+async function createSession(page) {
+  await provisionAndLogin(page.request, U, P);
+  const projects = await (await page.request.get(`${BASE}/api/projects`)).json();
+  const projectId = projects.projects[0].projectId;
+  const models = await page.request.put(`${BASE}/api/projects/${projectId}/models`, {
+    data: {
+      defaultModel: { provider: "custom", modelId: "claude-4-8" },
+      models: [
+        {
+          provider: "custom",
+          modelId: "claude-4-8",
+          apiKey: "sk-mock",
+          baseUrl: MOCK,
+          contextWindow: 200000,
+        },
+      ],
+    },
+  });
+  expect(models.ok(), "put models").toBeTruthy();
+  const session = await page.request.post(
+    `${BASE}/api/projects/${projectId}/agents/default_agent/sessions`,
+    {
+      data: {
+        provider: "custom",
+        modelId: "claude-4-8",
+        approvalMode: "always-ask",
+      },
+    },
+  );
+  expect(session.ok(), `create session: ${await session.text()}`).toBeTruthy();
+  return (await session.json()).session.sessionId;
+}
+
+test("a stale steer response after reload queues the follow-up instead of stranding it", async ({
+  page,
+}) => {
+  const sessionId = await createSession(page);
+  await page.goto(`${BASE}/chat/${sessionId}`);
+  const composer = page.locator("textarea").first();
+  await composer.waitFor();
+
+  await composer.fill("first task before reload");
+  const firstPost = page.waitForResponse((response) =>
+    response.url().endsWith(`/sessions/${sessionId}/tasks`),
+  );
+  await page.getByRole("button", { name: "发送" }).click();
+  expect((await firstPost).status()).toBe(202);
+
+  // Reproduce the post-reload stale-state seam deterministically: the client still sees a
+  // running Task, but /steer reports that the core run just ended. The fallback must use the
+  // server's queue-if-busy path, which is safe on both sides of the completion race.
+  await expect(page.getByRole("button", { name: "允许" })).toBeVisible();
+  await page.route(`**/api/sessions/${sessionId}/steer`, (route) =>
+    route.fulfill({
+      status: 409,
+      contentType: "application/json",
+      body: JSON.stringify({
+        error: {
+          code: "not_running",
+          message: "This Session has no Task in progress; send a new task instead.",
+        },
+      }),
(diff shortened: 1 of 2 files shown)


March-77:
Sorry about my earlier PRs. I submitted several changes too quickly and did not give enough attention to the project’s contribution expectations or explain my verification clearly. I understand why they were closed.

For this PR, I focused only on #89. I reviewed the relevant client and server flow, reproduced the completion race, checked the final diff, and added a deterministic regression test to verify that a follow-up message is queued correctly when the page is reloaded and the steering request encounters stale session state.

During a previous internship, I worked with a similar internal platform, although it was much less complete than this project. That experience is part of why I’m interested in this project, and I hope to learn more about this kind of system while making useful contributions here.

This PR is ready for review. Thank you for taking another look.

hiyouga (approved):
Reviewed in a dedicated worktree, against the PR head and against a local merge with current `main` (13 commits ahead; the merge is clean — the only commit touching `chat-input.tsx` since, #241, doesn't overlap the steer/send region).

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` and `pnpm test` (2,071 passed / 5 skipped) all pass in both states.

Worth flagging: **the green checks on this PR do not cover the new spec.** The e2e step self-skips on fork PRs because `DEEPSEEK_API_KEY` isn't exposed to them (`DEEPSEEK_API_KEY not available; skipping e2e.` in the `ci` job log). So I ran it locally:

- `reloaded-followup.spec.mjs` passes;
- reverting just the one-line change back to `await sendNormal()` and rebuilding makes it fail at `expect(response.status()).toBe(202)` with **409** — so it's a real regression test, not a tautology;
- `steer-reload.spec.mjs` and `chat.spec.mjs` still pass alongside it.

## On the change itself

The routing is sound. `queueIfBusy` is a strict superset of the plain POST inside `startTask`: an idle session still falls through to `assertIdle` + `launchTask` and starts immediately (`queued: false`), so this fallback can only turn failures into successes. The queue has no gap for the message to fall through either — the idle flip (`session-manager.ts:1293`) and the drain check (`:1314`) sit in one synchronous block, so a concurrent `startTask` either pushes before the flip and gets drained, or reads idle and launches directly; idle eviction also explicitly skips entries with pending follow-ups (`:1007`).

The blast radius really is just the `post` argument: `midRunAction` closes both mid-run channels when `goalOn`, and steering requires that no `/agent` / `/model` chip is staged, so the goal / handoff / model-switch branches inside `sendNormal` are unreachable from here. `onQueueFollowUp ?? onSend` is also the right guard — unlike the queue branch's `onQueueFollowUp!`, `steerAction` does not imply the queue channel is wired.

One side effect the description doesn't mention, and it's a welcome one: `/steer` also answers `not_running` while the session is compacting, so steering during compaction now queues and auto-sends afterwards instead of failing with `task_in_progress` and stranding the draft.

## Two notes, neither blocking — merging as-is

**1. `Closes #89` overstates it.** #89 records two symptoms: (a) `POST /tasks` answers 202 but the task never runs (trace-verified: no LLM call, no `task_state` change) — which is the issue title — and (b) the steer 409 fallback stranding the composer. This PR fixes (b). The issue attributes (a) to the post-reload stream/runtime handshake, which a client-side change can't reach. I'll reopen #89 after merging with a note on what remains.

**2. The per-turn thinking level is dropped on this path** — a small behavior delta against the code being replaced. `onSend` sends `thinkingLevel: turnThinkingLevel`; `onQueueFollowUp` doesn't. The picker is `disabled={busy}`, not `disabled={running}`, so it's reachable mid-run, and the server supports it (`startTask` stores `thinkingLevel` with the queued follow-up and applies it on auto-start — see the `TaskCreateRequest.queueIfBusy` doc comment). The existing queue branch (`chat-input.tsx:1720`) has the same gap, so this PR makes the fallback consistent with it rather than introducing a new class of bug. Fixing it once in `chat-page.tsx`'s `onQueueFollowUp` covers both paths:

```ts
const res = await api.postTask(selected.sessionId, {
  input,
  queueIfBusy: true,
  ...(turnThinkingLevel
    ? { thinkingLevel: turnThinkingLevel as TaskCreateRequest["thinkingLevel"] }
    : {}),
});
```

(plus `turnThinkingLevel` in the dependency array). Happy to take that as a follow-up.

PR #234, opened 2026-08-07
fix(core): serialize Trace appends; tolerate malformed middle lines on read

### LINKED ISSUE #215: 并发写入大型多模态消息导致 Trace JSONL 损坏 。Trace JSONL corruption caused by concurrent writes of large multimodal messages
## Title

Trace JSONL corruption caused by concurrent writes of large multimodal messages

## Description

Trace files can become corrupted when multiple messages are written concurrently and one message contains a large payload, such as a Base64-encoded image.

The writer currently appends records without serialization:

```js
await appendFile(path, `${JSON.stringify(message)}\n`, "utf8");
```

Large strings may be split into multiple underlying writes. Another message can be appended between those chunks, producing invalid JSONL:

```text
<large JSON part 1><another complete JSON>
<large JSON part 2>
```

The history reader then fails with an error such as:

```text
SyntaxError: Expected ',' or ']' after array element in JSON
    at JSON.parse
    at parseTraceLines
    at readTraceTolerant
```

Multimodal conversations trigger this more frequently because image Data URLs can make a single record several megabytes long.

## Expected behavior

Each trace message should be appended as one complete JSONL record without interleaving.

## Suggested fix

Serialize `write()` and `rotate()` operations through the same promise queue or mutex. A failed write should not block subsequent operations.

As additional hardening, the history reader should return partial history or a diagnostic placeholder instead of throwing an unhandled exception for a malformed middle record.

## Regression test

Concurrently write one multi-megabyte record and several small records, then verify:

- Every non-empty line can be parsed independently with `JSON.parse`.
- The parsed record count and order match the submitted messages.
- Concurrent `write()` and `rotate()` operations preserve the correct shard boundaries.

---

## 标题

并发写入大型多模态消息导致 Trace JSONL 损坏

## 问题描述

当多个消息被并发写入，且其中包含 Base64 图片等大型数据时，Trace 文件可能损坏。

当前 Writer 在未做串行化处理的情况下直接追加记录：

```js
await appendFile(path, `${JSON.stringify(message)}\n`, "utf8");
```

大型字符串可能被拆分为多次底层写入。此时，另一条消息可能插入这些数据块之间，产生无效的 JSONL：

```text
<大型 JSON 的前半部分><另一条完整 JSON>
<大型 JSON 的后半部分>
```

随后读取历史记录时会出现类似错误：

```text
SyntaxError: Expected ',' or ']' after array element in JSON
    at JSON.parse
    at parseTraceLines
    at readTraceTolerant
```

多模态对话更容易触发该问题，因为图片 Data URL 可能使单条记录达到数 MB。

## 预期行为

每条 Trace 消息都应作为完整的 JSONL 记录写入，不应与其他消息交错。

## 建议修复

通过同一个 Promise 队列或互斥锁，将 `write()` 和 `rotate()` 操作串行化。单次写入失败后，后续操作仍应能够正常执行。

作为额外容错措施，当历史记录中间存在损坏数据时，读取端应返回部分历史记录或诊断占位信息，而不是抛出未处理异常。

## 回归测试

并发写入一条数 MB 的记录和多条小记录，然后验证：

- 每个非空行都能被 `JSON.parse` 独立解析；
- 解析出的记录数量和顺序与提交的消息一致；
- 并发执行 `write()` 和 `rotate()` 时，记录仍写入正确的分片文件。

hiyouga:
Fixes #215.

## Problem

The Trace writer appended records with a bare `await appendFile(path, JSON.stringify(msg) + "\n")` and no serialization. `fs.appendFile` splits large payloads into multiple underlying writes (512 KiB chunks), so a multi-megabyte record — typically a base64 image Data URL in a multimodal conversation — could be interleaved by a concurrent append:

```
<large part 1><other complete JSON>
<large part 2>
```

The history reader (`parseTraceLines` via `readTraceTolerant`) then threw `SyntaxError` on the torn middle line, making the whole session history unreadable.

## Where the concurrency actually happens

One process, multiple async producers on a single `Writer` instance: the engine's LLM stream driver writes messages while each approved tool runs as a fire-and-forget `executeOne` promise that writes its own outputs (`packages/core/src/engine/context-engine.ts`), plus Session-level writes (`session_meta`, `goal_finished`). Cross-instance/file concurrency does not occur by design **within a single server/CLI process**: a Session allows only one active run at a time, child sessions write their own Trace files, and server-side Trace import only ever creates brand-new files (`wx` flag, existing session ids are 409-rejected). Two processes pointed at the same agent data directory are outside this contract and outside supported usage. Serializing inside the `Writer` instance is therefore the correct layer.

## Fix

- **Writer serialization** (`packages/core/src/trace/writer.ts`): `write()` and `rotate()` are enqueued on one per-instance promise chain. Each record lands as a single uninterrupted append; rotation cannot land inside an append, and shard membership follows submission order (the target path is resolved when the operation runs). Error semantics: a failed operation rejects **that caller's** promise only — the chain absorbs the failure so later operations proceed (callers already treat Trace writes as best-effort and log the error). A **hung** append is the documented flip side: it blocks the instance's chain until it settles — deliberate head-of-line blocking, since ordering cannot be preserved around an append whose outcome is unknown. The visible contract is preserved: `await write(msg)` still resolves only after that record has been appended, so write-then-send ordering assumptions hold.
- **Reader hardening** (`packages/core/src/trace/resume.ts`): `parseTraceLines` now skips a malformed **middle** line by default and keeps every parseable record, instead of throwing; the existing trailing-truncation tolerance is unchanged (still silently ignored). The last-non-empty-line check is precomputed with one backward scan, so widely damaged content parses in O(n) instead of O(n²). `readTraceTolerant` reports skipped lines once per read on stderr with the path and the first 20 line numbers (capped with an ellipsis plus exact total). No diagnostic placeholder record is injected — the OmniMessage payload union is closed, and replay is already built to stay structurally valid with records missing (pairing backfill, uncommitted-turn drop), so skipping integrates cleanly.
- **`readTraceHead` rides the new default** (`packages/server/src/internal/trace-head.ts`): a malformed line inside the bounded head window is now dropped silently instead of throwing. This also improves the Session-index reconcile path: previously the throw yielded `records = []`, so the index never got its `session_meta` backfill and the same permanently damaged file was re-read on every reconcile pass; now the head's parseable records (typically `session_meta` and the first prompt) are extracted once and the file classifies normally.
- **Recovery story**: files already corrupted by the pre-fix interleaving become readable again through the tolerant reader — the torn record (and the small record that landed inside it) is lost, everything else survives. No migration needed.
- **Import stays strict** (`packages/server/src/services/trace-service.ts`): `importTraceFile` passes the new `onMalformed: "throw"` option, keeping the 400 `invalid_trace` contract — import is the gate for user-supplied files, where damage should be reported rather than silently dropped.

## Regression tests

- Concurrent writes of one 3 MB record plus small records: every non-empty line parses independently, count and order match submission (fails against the old writer — verified).
- Concurrent `write()` + `rotate()`: shard boundaries follow submission order, a large in-flight append is never split by rotation (also fails against the old writer).
- A failed write rejects its own caller without wedging subsequent writes.
- Reader: a synthetically torn middle record (interleaved fragment, both on-string and on-disk) yields the surviving records without throwing, with the skip diagnosed; strict mode still throws; trailing truncation stays tolerated in both modes.
- Wide corruption: 10,000 malformed lines around two survivors parse correctly (and fast — guards the O(n) tail check); the stderr diagnostic caps its spelled-out line-number list at 20.

## Review follow-ups (round 2)

Addressed from the review:

1. `parseTraceLines` skip path de-quadratified: `lastNonEmptyIndex` precomputed by one backward scan instead of `slice().every()` per malformed line (reviewer measured 5.4 s for a 50k-line all-malformed file); wide-corruption test added.
2. `readTraceTolerant` stderr diagnostic capped at the first 20 line numbers plus ellipsis and exact total.
3. Writer class doc now states the head-of-line trade-off (a hung append still blocks the instance's chain) and scopes the no-cross-instance-concurrency argument to a single server/CLI process.
4. This description now covers the `readTraceHead` behavior change (above).

## Docs / design sync

- `packages/docs/content/sessions-and-traces.{en,zh}.md` updated: serialized appends, malformed-middle-line skip on read.
- Design spec note (design repo is separate, not part of this PR): `design/specs/05-ARCHITECTURE.md` § 重放规则 currently says only "重放容忍残缺末行（忽略该行）". Suggested sync: also state that a malformed middle line (legacy files damaged by pre-serialization concurrent appends) is skipped best-effort while all parseable records are kept, and that Trace 追加在 Writer 实例内串行执行. The writer's append-only / one-file-per-context contract is unchanged.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (all 8 packages)
- `pnpm -r test` — pass (core 643, server 533, cli 235, web 603, desktop 43, docs 32, landing 39, skills 18)
- New concurrency tests confirmed to fail against the pre-fix writer and pass with the fix

https://claude.ai/code/session_01FAkj9ao8kGTiBjAcsxnC1x

### DIFF
--- a/packages/core/src/trace/index.ts
+++ b/packages/core/src/trace/index.ts
@@ -7,4 +7,4 @@ export {
   readTraceTolerant,
   resumeTrace,
 } from "./resume.js";
-export type { LocatedTraceFile, ResumeResult } from "./resume.js";
+export type { LocatedTraceFile, ParseTraceLinesOptions, ResumeResult } from "./resume.js";
--- a/packages/core/src/trace/resume.ts
+++ b/packages/core/src/trace/resume.ts
@@ -104,6 +104,17 @@ export interface ParseTraceLinesOptions {
 export function parseTraceLines(content: string, options?: ParseTraceLinesOptions): OmniMessage[] {
   const onMalformed = options?.onMalformed ?? "skip";
   const lines = content.split("\n");
+  // Index of the last non-empty line, found with one backward scan up front. The malformed
+  // branch below runs once per skipped line in skip mode, so rescanning the remainder there
+  // (slice + every) would make widely damaged content O(n^2) — seconds of synchronous CPU on
+  // the server history path for a large non-JSONL file.
+  let lastNonEmptyIndex = -1;
+  for (let i = lines.length - 1; i >= 0; i--) {
+    if (lines[i]!.trim().length > 0) {
+      lastNonEmptyIndex = i;
+      break;
+    }
+  }
   const out: OmniMessage[] = [];
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i]!.trim();
@@ -111,8 +122,7 @@ export function parseTraceLines(content: string, options?: ParseTraceLinesOption
     try {
       out.push(JSON.parse(line) as OmniMessage);
     } catch (err) {
-      const isLastNonEmpty = lines.slice(i + 1).every((l) => l.trim().length === 0);
-      if (isLastNonEmpty) break; // truncated tail: expected crash artifact, ignore silently
+      if (i === lastNonEmptyIndex) break; // truncated tail: expected crash artifact, ignore silently
       if (onMalformed === "throw") throw err;
       options?.onSkip?.(i + 1, err);
     }
@@ -120,10 +130,14 @@ export function parseTraceLines(content: string, options?: ParseTraceLinesOption
   return out;
 }
 
+/** Maximum count of skipped line numbers spelled out in the readTraceTolerant diagnostic. */
+const SKIPPED_LINES_SHOWN = 20;
+
 /**
  * Read and parse a Trace file, best-effort: tolerates a truncated last line, and skips
  * malformed middle lines (files damaged by the pre-fix concurrent-append bug, #215) while
- * keeping every parseable record. Skipped lines are diagnosed once per read on stderr.
+ * keeping every parseable record. Skipped lines are diagnosed once per read on stderr, with
+ * the spelled-out line-number list capped (a heavily damaged file can skip thousands).
  */
 export async function readTraceTolerant(path: string): Promise<OmniMessage[]> {
   const skipped: number[] = [];
@@ -131,8 +145,10 @@ export async function readTraceTolerant(path: string): Promise<OmniMessage[]> {
     onSkip: (lineNumber) => skipped.push(lineNumber),
   });
   if (skipped.length > 0) {
+    const shown = skipped.slice(0, SKIPPED_LINES_SHOWN);
+    const ellipsis = skipped.length > shown.length ? ", …" : "";
     process.stderr.write(
-      `[trace] skipped ${skipped.length} malformed line(s) [${skipped.join(", ")}] in ${path}\n`,
+      `[trace] skipped ${skipped.length} malformed line(s) [${shown.join(", ")}${ellipsis}] in ${path}\n`,
     );
   }
   return messages;
--- a/packages/core/src/trace/resume.ts
+++ b/packages/core/src/trace/resume.ts
@@ -77,12 +77,32 @@ export interface ResumeResult {
 /** Content of the pairing-backfill placeholder output (the tool hadn't finished and no output was persisted before the process exited). */
 const PROCESS_EXIT_PLACEHOLDER = "[interrupted: process exited before the tool finished]";
 
+/** Options for `parseTraceLines`. */
+export interface ParseTraceLinesOptions {
+  /**
+   * Handling of a malformed line **in the middle** of the content (the trailing truncated
+   * line is always tolerated and ignored, see `parseTraceLines`):
+   *   - "skip" (default): drop that line and keep every parseable record. Reading a Trace back
+   *     is best-effort — the corruption already happened, and throwing away the whole file
+   *     helps no one. Replay is built to stay structurally valid with records missing (crash
+   *     tolerance, pairing backfill), so surviving records still resume.
+   *   - "throw": rethrow the JSON parse error. For validating untrusted input where damage
+   *     should be reported instead of silently repaired (server-side Trace import).
+   */
+  onMalformed?: "skip" | "throw";
+  /** Called for each skipped malformed middle line (1-based line number). Not called for the tolerated trailing truncated line. */
+  onSkip?: (lineNumber: number, error: unknown) => void;
+}
+
 /**
  * Parse Trace JSONL content. Tolerates a **truncated last line** left behind by an abnormal
- * process exit (that line is ignored); corruption in the middle is outside the crash window
- * (append-only, single writer), so it throws loudly.
+ * process exit (that line is silently ignored). A malformed line in the **middle** — e.g. a
+ * record torn by the pre-fix concurrent-append interleaving (#215) — is skipped by default
+ * (reported via `onSkip`), keeping every parseable record; pass `onMalformed: "throw"` to
+ * make middle corruption a hard error instead.
  */
-export function parseTraceLines(content: string): OmniMessage[] {
+export function parseTraceLines(content: string, options?: ParseTraceLinesOptions): OmniMessage[] {
+  const onMalformed = options?.onMalformed ?? "skip";
   const lines = content.split("\n");
   const out: OmniMessage[] = [];
   for (let i = 0; i < lines.length; i++) {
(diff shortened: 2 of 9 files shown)


hiyouga:
## Review

Reviewed in a dedicated worktree off `pr-234`.

### Verification

| Check | Result |
|---|---|
| `pnpm -r build` | ✅ |
| `pnpm typecheck` | ✅ 8 packages |
| core tests | ✅ 641 passed |
| server tests | ✅ 533 passed |
| `pnpm format:check` | ✅ |

Regression value confirmed independently: with `packages/core/src/trace/writer.ts` reverted to `main` and `trace.test.ts` run in isolation, `keeps a multi-MB record intact under concurrent small appends` and `serializes rotate() against in-flight writes` both fail while the other 5 pass. These are real regression tests, not self-confirming assertions.

### Verdict

The diagnosis and the chosen layer are both right. I re-checked the concurrency argument rather than taking it on trust: `agent.ts:302` / `agent.ts:457` are the only two `new Writer` sites, `session-manager.ts` keeps an active-session table plus per-Session mutual exclusion, and `trace-service.ts:1245` does write imports with `flag: "wx"` — so "cross-instance concurrency does not occur" holds, and serializing inside the `Writer` instance is the correct place to fix it.

`enqueue()`'s error handling is clean: `this.chain = result.then(noop, noop)` both keeps a failure scoped to its own caller and incidentally attaches a handler to `result`, so a fire-and-forget `write()` cannot raise `unhandledRejection`.

### Findings

**1. `parseTraceLines` degrades to O(n²) in skip mode (worth fixing)**

The trailing-line check at `resume.ts:114` allocates a full copy of the remaining array for *every* malformed line:

```js
const isLastNonEmpty = lines.slice(i + 1).every((l) => l.trim().length === 0);
```

`.every` short-circuits; `slice` does not. Before this PR the branch ran at most once (it threw immediately after); now it runs once per skipped line. Measured on all-malformed content:

```
 5000 lines ->   52 ms
20000 lines ->  547 ms      (4x lines, 10x time)
50000 lines -> 5443 ms      (2.5x lines, 10x time)
```

`readTraceTolerant` sits on the server history read path (`trace-service.ts:187`, `:1158`) and this is synchronous CPU, so a badly damaged file blocks the event loop for seconds. The trigger is "a large fraction of the file is unparseable" — a non-JSONL file dropped into the traces dir, or wide corruption after a disk failure — not the #215 tearing itself, which only produces 2 bad lines per torn record. The fix is a few lines: scan backwards once before the loop for `lastNonEmptyIndex` and compare `i === lastNonEmptyIndex` at the malformed branch, which also drops the repeated `trim()`.

Same function: `readTraceTolerant` joins every skipped line number into one stderr line (`[${skipped.join(", ")}]`). On a heavily damaged file that line gets enormous — worth truncating (e.g. first 20 plus an ellipsis).

**2. Head-of-line blocking is a real trade-off; the class doc should say so**

With one chain per instance, a single stuck `appendFile` (network mount, FUSE) blocks all Trace writes for that Session — and `context-engine.ts:1724` awaits `this.deps.trace.write(msg)` inside the ReAct loop, where `try/catch` stops a throw but not a hang, so the whole session stalls. Previously a stuck append only blocked the producer that issued it. The amplification is bounded (the LLM driver has to write to the same disk anyway) and I think the trade-off is acceptable, but the class doc currently only promises "a failed operation must not wedge the chain" — worth one sentence noting that a *hung* one still does.

**3. `packages/server/src/internal/trace-head.ts` behaviour change is not mentioned in the PR body**

`readTraceHead` goes through the default `parseTraceLines(text)` path, so flipping the default to `"skip"` also changes it: a malformed line inside the head window now gets dropped silently (no `onSkip` passed, so no diagnostic) instead of throwing. Tracing the caller at `trace-index.ts:249`, this is an improvement — the old throw produced `records = []` → `metaRead=0` → every later reconcile pass re-read the same permanently damaged file. But the PR body scopes the reader change to resume/history, and it also touches Session index classification. Worth a line in the description.

**4. The "no cross-instance concurrency" argument in the writer doc is process-scoped**

All three reasons listed (one active run per Session, child sessions write their own files, import uses `wx`) hold inside one server process. CLI (`chat.ts:108` → `resumeSession`) and server pointed at the same agent state directory are two processes with two Writers on one file, which the chain cannot cover. That usage shouldn't happen anyway, but since the comment reads as an exhaustive proof, it's better to say "within a single process" explicitly.

### Follow-ups

- The `design/specs/05-ARCHITECTURE.md` § 重放规则 sync noted in the description still needs its own PR in the design repo.
- No changelog entry here is correct per the repo convention (changelog goes in a separate batch PR).

hiyouga:
Thanks for the thorough review — all four findings addressed in 0ccd08c:

1. **O(n²) skip path**: `parseTraceLines` now precomputes `lastNonEmptyIndex` with a single backward scan and compares `i === lastNonEmptyIndex` in the malformed branch, so widely damaged content parses in O(n) with no per-hit `slice().every()` allocation (and no repeated `trim()`). Added a wide-corruption regression test: 10,000 malformed lines around two survivors, asserting exact survivors and exact skip line numbers — runs in milliseconds.
2. **Unbounded diagnostic**: `readTraceTolerant` caps the spelled-out list at the first 20 line numbers plus an ellipsis while keeping the exact total (`skipped N malformed line(s) [1, 2, …] in <path>`); covered by a new test.
3. **Doc trade-offs**: the `Writer` class doc now states head-of-line blocking explicitly — a *failed* append doesn't wedge the chain, but a *hung* one blocks it until it settles (deliberate: ordering can't be preserved around an append whose outcome is unknown) — and the no-cross-instance-concurrency argument is now scoped to "within a single server/CLI process", with two-process setups called out as outside the contract.
4. **PR body**: updated to cover the `readTraceHead` behavior change — a malformed line in the head window is now dropped silently instead of throwing, which also stops the Session-index reconcile loop from re-reading the same permanently damaged file every pass (old throw → `records = []` → no `session_meta` backfill, ever).

Full chain re-verified: `pnpm -r build`, `pnpm typecheck` (8/8), `pnpm -r test` (core 643, server 533, all others green), `pnpm format:check` clean.

https://claude.ai/code/session_01FAkj9ao8kGTiBjAcsxnC1x

PR #235, opened 2026-08-07
fix(core,docs): window-derived output caps and compaction threshold, slower retry ladder, retry-all-but-auth

### LINKED ISSUE #218: vllm local model issue
在使用 vllm 模型接入 PenguinHarness 时候，除了需要注意 --enable-auto-tool-choice  --tool-call-parser，还需要把 context window 配置成 max_model_len，另外 当前的 context len + LLM 请求时候的 maxTokens 不能超过 context window， 这样就可以保证当前的 context+maxTokens 数量不会超过 max_model_len。

避免出现

400 This model's maximum context length is 32768 tokens. However, you requested 32000 output tokens and your prompt contains at least 769 input tokens, for a total of at least 32769 tokens. Please reduce the length of the input prompt or the number of requested output tokens. (parameter=input_tokens, value=769)

同时确保 agent 压缩的 token 上限也要给压缩留空间，不能高于 context window - 1000 吧

另外现在 LLM 失败重试好像都很快，还没反应过来就重试完了，能不能稍微慢一点的退避？

Image

hiyouga:
Fixes #218.

Running against a small-window OpenAI-compatible endpoint (a local vLLM with `--max-model-len 32768`, say) failed three ways: every request 400'd because the configured `max_tokens` (seeded 32000) went on the wire as a fixed value regardless of how much of the window the input already occupied; the compaction threshold (default 128000) was unreachable inside the window, so compaction could never rescue the session; and the LLM retry ladder (250ms base) burned all five retries in ~7.75s — faster than any provider hiccup can recover. A follow-up request folded in here also simplifies retry classification: every LLM error retries except explicit auth failures.

### Per-request output clamp (core)
New `packages/core/src/llm/context-limits.ts` holds the pure window arithmetic; `GenerativeModel` resolves each request's effective output cap as `min(configured max_tokens, max(context_window − estimated input − OUTPUT_SAFETY_MARGIN, MIN_OUTPUT_TOKENS))`:

- Input size = the last completed request's real `token_usage.request.total` (seeded at construction with a system-prompt + tool-schema estimate; `setHistory` re-seeds on resume) plus a character-heuristic estimate of the newly appended messages. No tokenizer — ASCII ~4 chars/token, non-ASCII 1 token/char (errs high, the safe direction); image content gets a flat per-image allowance, including the `images` data-URL arrays on tool outputs (see review remediation below).
- A no-op in practice for big-window models (`min` returns the configured cap unchanged and the frozen request config goes out untouched); compaction requests run through the same path — they fire exactly when the context is largest.
- The clamp only ever lowers a cap (pinned meta budgets stay pinned), never emits a non-positive value, and **derives nothing from an assumption**: with no usable `context_window` on the entry (unset, or below the 4096 sanity threshold) the clamp is off entirely.
- Meta requests (title generation, vision describing) also receive the entry's `context_window` for a uniform derivation, though their small budgets (300/2048) mean the clamp never binds there — their overflow risk is input-side, which no output cap can fix.
- A hard clamp (derived cap below half the configured one) logs one stderr line naming configured cap, derived cap, window and estimated input — deduped per value — so a shaved `max_tokens` is diagnosable from the log.

### Window-derived compaction threshold (core)
`effectiveMaxContextLength` (moved from `agent.ts` into `context-limits.ts`) is now `min(configured max_context_length, context_window − COMPACTION_HEADROOM)`. The headroom (2×`OUTPUT_SAFETY_MARGIN` = 2048) reserves the safety margin plus ~1k for the summary request's own output, which the per-request clamp grants it at the trigger point. A 32k-window model now compacts at ~30.7k instead of never; `<=0` keeps meaning "compaction disabled"; derived at use — stored config is never rewritten. Unlike the output clamp, an unconfigured window here falls back to the assumed 128000 (harmless-to-helpful: it only makes compaction fire by the assumed mark). This replaces the previous 75%-of-window rule and closes the unset-window gap where no clamp applied at all.

### Slower retry backoff (core)
The engine's reconnect backoff base goes 250ms → 2000ms: the ladder is now 2s/4s/8s/16s/30s (≈60s total patience; retry count and the 30s ceiling unchanged). Transient provider failures get a real recovery window, and every planned wait clears the web app's 2s countdown floor (`COUNTDOWN_MIN_MS`), so each retry renders as a visible countdown. Other schedules (stream idle timeout etc.) are untouched.

### Retry classification: everything retries except auth (core)
Classification keeps only the explicit auth detection (`isAuthenticationError`: 401, explicit credential codes/types, SDK `AuthenticationError` — logic unchanged, deliberately narrow). Everything else — bare 403, 400, 429, 5xx, quota/subscription rejections, transport errors — lands in the retryable statuses (`failed`/`timeout`/`malformed`) and rides the ladder, failing after exhaustion exactly like before. The quota-detection machinery is deleted: `isQuotaExhaustedError`, its `insufficient_*` code set, the `/quota|subscription|…/` message heuristic (with its CJK fixtures) and every branch that consulted them. `isRetryableError` survives purely as the honest `timeout`-vs-`failed` **label** (transport-shaped vs provider rejection) so observability keeps its taxonomy — it no longer gates retries.

**Behavior notes.** Quota/subscription rejections previously *labeled* `timeout` now label `failed` (they were already retried; the real provider message still rides `request_end.error_message`, but the server's error records will bucket them as `failed`). And genuinely permanent errors (a malformed request, a hard 400) now deterministically burn the full ~60s ladder before surfacing — the ladder multiplies the cost of deterministic-failure paths by ~8× versus the old base. Both are the uniformity-over-classification trade the user chose; the engine comment ("retrying a genuinely permanent error costs the ladder and ends the same way") already frames it.

### Review remediation (both review comments applied)
- **Image-estimate bug (review 1, finding 1)**: tool outputs carrying images (`read_image`, screenshots) hold them as `images?: string[]` data URLs, which the estimator serialized as text — a 1 MB base64 image estimated ~262k "tokens" (~163×) and floored the *next* request's cap to 512 even on a 128k window. Fixed by counting each `images` entry at the flat allowance and keeping the base64 out of the serialized text (complete and partial payloads); regression test mirrors the reviewer's repro through the `openStream` seam, plus a genuinely-huge-text counter-case.
- **Observability (review 1, finding 2)**: the hard-clamp stderr line above. The retry-compounding cost is acknowledged in the behavior notes.
- **Unset-window guard (review 1, finding 3)**: no per-request clamp is derived from the 128000 assumption; only the compaction threshold keeps using it.
- **Thinking levels (review 1, finding 4)**: documented in `context-limits.ts` as considered and out of scope (a small derived cap can be consumed by reasoning before any answer; the compaction threshold bounds the exposure).
- **Constant derivation (review 2)**: applied as proposed — `MIN_OUTPUT_TOKENS = floor(OUTPUT_SAFETY_MARGIN / 2)`, `COMPACTION_HEADROOM = OUTPUT_SAFETY_MARGIN * 2` (same values, invariants now structural), `MESSAGE_OVERHEAD_TOKENS` dropped (the serialized-JSON envelope already over-counts structure; the margin's job anyway), floor rewritten to `min(configured, max(remaining, MIN_OUTPUT_TOKENS))`, and `lastRequestTotal` seeded with `baseInputTokens` at construction, removing both `max()` fallbacks.
- **Deliberate call the reviewer flagged**: the tiny-window guard moved into `resolveContextWindow` — a configured `context_window` under 4096 is treated as unconfigured (no real model sits under 4096; the old `Math.ceil(window / 2)` floor in `effectiveMaxContextLength` only ever guarded typo'd windows and is dropped). This changes `effectiveMaxContextLength(128000, 2048)` from `1024` to the 128000-derived `125952`: a typo'd window now behaves like an unset one instead of compacting after every request, and per finding 3 it also gets no output clamp.

### Docs
`models.{en,zh}.md` gain a "Local / self-hosted OpenAI-compatible endpoints (e.g. vLLM)" section (start vLLM with `--enable-auto-tool-choice` + the matching `--tool-call-parser`; set the entry's `context_window` to the server's `max_model_len`) and the `context_window` / `max_tokens` rows describe the derivations and the unset-window behavior. `agent-loop`, `omni-message`, `interfaces` and `configuration` pages updated for the new ladder, the window-derived threshold, and the retry taxonomy (quota out of the `timeout` enumeration).

## Verification

From the worktree root (Node 24), after merging current `origin/main` (PR #232/#233 included):

- `pnpm -r build` — all packages build
- `pnpm typecheck` — clean
- `pnpm -r test` — all green: core 666 passed / 5 skipped, server 551, cli 235, web 610, desktop 43, skills 18, docs 32, landing 39

Deterministic unit tests (no real timers): `packages/core/test/context-limits.test.ts` (constant derivation, clamp arithmetic incl. margin/floor/degenerate/never-raise/unset-window, estimator incl. the tool-output image regression, compaction-threshold derivation incl. the typo'd-window call), the `GenerativeModel` suites in `packages/core/test/llm.test.ts` (per-request clamp through the `openStream` seam incl. the reviewer's image repro; outcome labeling: bare 403 → `failed`, quota-coded 403 → `failed`, explicit auth signal → `auth`, 400 → `failed`), and `packages/core/test/engine.test.ts` (2s/4s/8s/16s/30s ladder values; bare-403 `failed` retries and succeeds within the cap; a 400-flavored `failed` burns the ladder and surfaces verbatim; `auth` stops with no retry).

https://claude.ai/code/session_01FAkj9ao8kGTiBjAcsxnC1x


hiyouga:
## Review

Reviewed in a dedicated worktree off `pr-235`, focused on the per-request clamp in `generative-model.ts` and its collaborator `context-limits.ts`.

### Verification

| Check | Result |
|---|---|
| `pnpm -r build` | ✅ |
| `pnpm typecheck` | ✅ |
| `pnpm -r test` | ✅ core 654/5 skipped, server 533, cli 235, web 603, desktop 43, docs 32, landing 39, skills 18 |

### What I checked and found sound

Rather than take the estimator's assumptions on trust, I traced each one:

- **`lastRequestTotal` basis.** `usageToTokenCounts` builds `total = cache_read + cache_write + output`, and AgentHub's OpenAI client already subtracts cached from `prompt_tokens` (`src/openai/client.ts`: `promptTokens = usage.prompt_tokens - cachedTokens`). No double counting, and `total` really is "prompt + completion" = the context after that request. Correct basis for the next request's input.
- **"An interrupted or failed attempt was never committed."** True, and for a non-obvious reason: AgentHub's `streamingResponseStateful` pushes the user message and the assistant reply to `_history` only *after* its `for await` loop completes normally. An early `break` (interrupt) or a throw resumes the generator via `return()`/`throw()` and skips that code entirely — history stays untouched, so not updating `lastRequestTotal` is exactly right.
- **Post-compaction staleness.** `createLLM` builds a fresh `GenerativeModel` from `llmConfig` and only carries `sessionTokens` over, so `lastRequestTotal` resets to 0 and does not linger as a huge overestimate against the new, small context.
- **The 2s countdown-floor claim.** `message-item.tsx` uses `COUNTDOWN_MIN_MS = 2000` with `>=`, so a 2000ms wait is exactly at the threshold and does render.

The overall shape — derive at use, never rewrite stored config; clamp only ever lowers; `undefined` keeps the no-cap contract off the wire — is right, and the docs match the code.

### Findings

**1. Images carried by a tool output are counted as raw base64, flooring the output cap (bug)**

`approximateMessagesTokens` gives the flat `IMAGE_TOKEN_ESTIMATE` only when the *payload type itself* is `image_url` / `inline_data`. But `ToolCallOutputPayload` carries images in an `images?: string[]` field of `data:<mime>;base64,…` URLs (`omnimessage/types.ts:199`) — and that payload falls into the `else` branch, where `JSON.stringify(msg.payload)` counts every base64 character:

```
  64 KB base64 | image_url:   1608 tok | tool_call_output.images:  16424 tok
 256 KB base64 | image_url:   1608 tok | tool_call_output.images:  65576 tok
1024 KB base64 | image_url:   1608 tok | tool_call_output.images: 262184 tok
```

A 1 MB base64 payload (≈ a 750 KB PNG, well under `read_image`'s `MAX_IMAGE_BYTES = 5 MB`) is estimated at 262,184 tokens instead of ~1,608 — a 163× overestimate. Driving the real `openStream` seam with the same harness your new suite uses:

```
32k window:  request1 max_tokens=31714, request2 max_tokens=512   <-- floored
128k window: request1 max_tokens=32000, request2 max_tokens=512   <-- floored
same image as a user image_url message: request2 max_tokens=32000 <-- correct
```

So `read_image` (or any tool returning a screenshot) floors the very next request to `MIN_OUTPUT_TOKENS`, **including on a 128k window** — i.e. the clamp misfires on exactly the workload the safety margin was written for, not just on small-window endpoints. The threshold is roughly 250 KB of base64 for a 32k window and 1 MB for the assumed 128k default.

Fix: treat `images` the way the top-level image payloads are treated — add `IMAGE_TOKEN_ESTIMATE` per entry and exclude the `images` array from the serialized text, e.g. destructure it off before `JSON.stringify`. `PartialToolCallOutputPayload` has the same field; partials never reach this path today, but handling both keeps it from regressing. This also removes the incidental cost of `JSON.stringify`-ing and code-point-scanning a multi-megabyte string on the request path.

**2. Finding 1 compounds with the new retry ladder**

A floored 512-token cap almost certainly hits `finish_reason: "length"`, which `omniStopReason()` maps to `failed`, which is in `RETRY_STATUSES`. The retry re-sends the same tool output, re-triggers the same estimate, and floors again — so all five attempts burn deterministically. Before this PR that cost ~7.75s; with the new ladder it is ~60s. Worth being aware of when weighing the base bump: the ladder is a clear improvement for genuinely transient failures, but it also multiplies the cost of any deterministic-failure path by 8×.

Related, and useful independently of finding 1: **nothing is observable when the clamp binds.** A user hitting this sees a `failed` turn with no hint that `max_tokens` was shaved to 512. A one-line stderr warning when the derived cap drops below, say, half the configured value would have made #218 itself diagnosable, and would make the next clamp bug visible without a debugger.

**3. Unset `context_window` + disabled compaction is a new regression path**

Catalog entries carry `context_window` (`model-catalog.ts:1077`), so this is narrow, but: a custom entry against a large-window model with no `context_window` now assumes 128000. Normally self-consistent — the derived compaction threshold (125,952) fires before the clamp floors. With `compaction.max_context_length <= 0` (compaction off) there is nothing to fire, so past ~127k of real context the cap pins at 512 on a model that previously worked fine at 32000. Cheap guard: skip the clamp (or only apply the compaction-threshold half of it) when `context_window` was never configured, rather than deriving a hard output cap from an assumption.

**4. Reasoning models and a small derived cap**

I checked AgentHub and no client sends an explicit `budget_tokens`, so there is no `max_tokens > budget` provider constraint to violate — good. But a derived cap of 1–2k on a `high`/`xhigh` session gets consumed by thinking before any answer is produced, ending as `length` → `failed`. The clamp arithmetic has no notion of the thinking level. Probably out of scope here, but worth a sentence in `context-limits.ts` so the next reader knows it was considered rather than missed.

### Note on the changelog

The retry-ladder change (7.75s → 60s of total patience) is user-visible behaviour on existing, unchanged `system_config.yaml` files, as is the lower effective compaction threshold. Nothing stored is rewritten, so no migration is needed, but both belong in the changelog batch entry so the change in feel is attributable.

hiyouga:
## Follow-up: fewer constants in `context-limits.ts`

Separate from the bug in my previous comment — this is purely about reading cost. The arithmetic here is genuinely simple, but it currently presents as **six named constants plus a stateful field with two `max()` fallbacks**, so understanding it means holding more in your head than the math warrants.

Three of the six are not independent — the prose already says so, it just isn't in the code:

- `MIN_OUTPUT_TOKENS`'s doc says *"Kept below `OUTPUT_SAFETY_MARGIN` so a floored request still fits inside the window"*
- `COMPACTION_HEADROOM`'s doc says *"`OUTPUT_SAFETY_MARGIN` plus ~1k for the summary request's own output"*

Deriving them makes the invariant unbreakable instead of a comment someone can drift away from, and leaves **one** tunable knob for the whole margin story.

I tried it locally — this diff keeps every number identical and passes the full core suite (654 passed / 5 skipped):

```diff
-export const MIN_OUTPUT_TOKENS = 512;
+export const MIN_OUTPUT_TOKENS = Math.floor(OUTPUT_SAFETY_MARGIN / 2);

-export const COMPACTION_HEADROOM = 2048;
+export const COMPACTION_HEADROOM = OUTPUT_SAFETY_MARGIN * 2;

-/** Per-message allowance for chat-template structure (role markers, separators). */
-const MESSAGE_OVERHEAD_TOKENS = 8;

 export function approximateMessagesTokens(messages: OmniMessage[]): number {
   let total = 0;
   for (const msg of messages) {
     const p = msg.payload as { type?: string };
-    total += MESSAGE_OVERHEAD_TOKENS;
     if (p.type === "image_url" || p.type === "inline_data") total += IMAGE_TOKEN_ESTIMATE;
     else total += approximateTokens(JSON.stringify(msg.payload));
   }
   return total;
 }

-  return Math.max(Math.min(configured, remaining), Math.min(configured, MIN_OUTPUT_TOKENS));
+  return Math.min(configured, Math.max(remaining, MIN_OUTPUT_TOKENS));
```

and in `generative-model.ts`, seeding the tracker with the fixed prefix removes both `max()` fallbacks:

```diff
     this.baseInputTokens =
       approximateTokens(config.systemPrompt ?? "") +
       approximateTokens(JSON.stringify(this.uniConfig.tools ?? []));
+    this.lastRequestTotal = this.baseInputTokens;

-      Math.max(this.lastRequestTotal, this.baseInputTokens) +
-        approximateMessagesTokens(newMessages),
+      this.lastRequestTotal + approximateMessagesTokens(newMessages),

-    this.lastRequestTotal = Math.max(
-      this.lastRequestTotal,
-      this.baseInputTokens + approximateMessagesTokens(history),
-    );
+    this.lastRequestTotal = this.baseInputTokens + approximateMessagesTokens(history);
```

Notes on each:

- **The floor rewrite is provably equivalent**, not just equivalent on the test cases: I enumerated ~20M `(configured, remaining)` pairs across the whole relevant range with zero mismatches. It also reads as the sentence you'd say out loud — "floor the remaining window, then never exceed the configured cap" — where `max(min, min)` needs a moment.
- **`Math.floor`** on the derived floor so an odd margin can't put a fractional `max_tokens` on the wire.
- **Dropping `MESSAGE_OVERHEAD_TOKENS` is the one real behaviour change**: the estimate loses 8 tokens per new message. That padding was already dominated by what the same line over-counts anyway — `JSON.stringify(payload)` charges for `"type"`, `"role"`, `"tool_call_id"`, braces and quotes, which is ~18 tokens of envelope per tool output that never reaches the provider. And absorbing chat-template structure is literally what `OUTPUT_SAFETY_MARGIN`'s doc says the margin is for, so counting it twice buys nothing. If you'd rather keep the belt-and-braces, keeping it costs one constant.
- **The `max()` fallbacks were dead weight**: a measured `token_usage.request.total` always includes the system prompt and tool schemas, so `lastRequestTotal >= baseInputTokens` holds from the first completed request onward; seeding it at construction makes that true from the start. `baseInputTokens` stays as a field only because `setHistory` needs it.

Result: `OUTPUT_SAFETY_MARGIN` becomes the single dial for the whole safety story, `DEFAULT_CONTEXT_WINDOW` and `IMAGE_TOKEN_ESTIMATE` stay as the two independent facts they are, and `lastRequestTotal` has one meaning with no conditional fallback.

One more I did **not** include because it is not behaviour-preserving: the `Math.ceil(window / 2)` floor in `effectiveMaxContextLength` only ever binds when `window < 4096` (below that, `window − 2048 < window / 2`), which no real model has. If its job is guarding against a `context_window` typo'd as `0`/`1`, rejecting those in `resolveContextWindow` — treating anything under a sane minimum as unconfigured — puts the guard where the other window validation already lives and drops a `max()` from the derivation. It would change `effectiveMaxContextLength(128000, 2048)` from `1024` to the 128000-derived value, so it needs a deliberate call rather than a silent tidy-up.

Happy to push any subset of this to the branch.

hiyouga:
Both review comments are applied in f50dafe (after merging current main in e3a5f2c). Point-by-point:

**Review 1**

1. **Tool-output image estimate (bug)** — fixed. `approximateMessagesTokens` now recognizes the `images?: string[]` data-URL array on tool-output payloads (complete and partial): each entry counts at the flat `IMAGE_TOKEN_ESTIMATE` and the array stays out of the `JSON.stringify` text, which also removes the megabyte-scan from the request path. Regression tests mirror your repro: through the `openStream` seam, a 1 MB base64 `read_image`-style output no longer floors request 2 (`max_tokens` stays 32000 on a 128k window), with a genuinely-huge-text counter-case pinning that real text still counts.
2. **Observability when the clamp binds** — added: when the derived cap drops below half the configured one, `GenerativeModel` writes one stderr line with the numbers (`max_tokens X -> Y (context_window W, estimated input E)`), deduped per distinct value so retries don't repeat it. The retry-compounding cost you quantified (deterministic failures now burn ~60s of ladder, ~8× the old base) is acknowledged in the PR body as part of the uniformity trade — same bucket as the retry-all-but-auth follow-up that landed in this push.
3. **Unset window + compaction off** — guarded: the per-request clamp is never derived from the 128000 assumption anymore. `resolveContextWindow` returns undefined for an unconfigured window and the clamp switches itself off; only the compaction-threshold derivation keeps the assumption (its failure mode is compaction firing by the assumed mark, not a pinned 512 cap).
4. **Reasoning models** — documented in `context-limits.ts` (module doc): a small derived cap can be consumed by thinking before any answer; considered, out of scope, exposure bounded by the compaction threshold sitting `COMPACTION_HEADROOM` below the window.

**Review 2 (constant derivation)** — applied in full: `MIN_OUTPUT_TOKENS = Math.floor(OUTPUT_SAFETY_MARGIN / 2)`, `COMPACTION_HEADROOM = OUTPUT_SAFETY_MARGIN * 2`, `MESSAGE_OVERHEAD_TOKENS` dropped (your argument taken: the JSON envelope already over-counts structure, and absorbing template overhead is the margin's stated job), the floor rewritten to `Math.min(configured, Math.max(remaining, MIN_OUTPUT_TOKENS))`, and `lastRequestTotal` seeded with `baseInputTokens` at construction with both `max()` fallbacks removed. The one item you flagged as needing a deliberate call is taken too, deliberately: the sub-4096 sanity threshold now lives in `resolveContextWindow` (a window that small is a typo, not a model) and the `Math.ceil(window / 2)` floor is gone from `effectiveMaxContextLength` — `effectiveMaxContextLength(128000, 2048)` now yields the 128000-derived 125952 instead of 1024, i.e. a typo'd window behaves like an unset one (and per finding 3 also gets no output clamp) instead of compacting after every request. Tests updated to pin exactly that.

Also folded into the same push: the user-requested retry simplification (every LLM error retries except explicit auth; the quota detector, its code set and the message heuristic are deleted; `isRetryableError` survives only as the timeout-vs-failed label), and your changelog note is queued for the maintainer's batch entry — the ladder feel-change and the lower effective threshold are called out as user-visible on unchanged configs.

Full chain re-verified after the main merge: build / typecheck / all-package tests (core 666+5 skipped, server 551, cli 235, web 610, desktop 43, skills 18, docs 32, landing 39) / format:check all green.

https://claude.ai/code/session_01FAkj9ao8kGTiBjAcsxnC1x

PR #242, opened 2026-08-10
feat(core): MCP Server support — stdio, Streamable HTTP and SSE transports

### LINKED ISSUE #229: 支持MCP吗
rt

### LINKED ISSUE #239: MCP Bridge
Consume External MCP Servers  Tools
当前仅提供 9 个内置工具。这虽然保证了 Token 效率和极简体验，但也带来了生态封闭的问题：
用户需要 GitHub 集成时，必须手写 curl + jq 的 shell 脚本;每次新增工具能力都需要修改核心代码，社区无法独立贡献;工具定义的上下文消耗随工具数量线性增长.

hiyouga:
## Summary

Implements the MCP (Model Context Protocol) integration that the architecture reserved a seam for (`tools.mcpServers` in `system_config.yaml`), built on the official TypeScript SDK v2 (`@modelcontextprotocol/client` 2.0.0, spec revision 2026-07-28).

Closes #239. Closes #229.

**Transports** — one entry per server, `{ name, config }` (the stored shape is unchanged):

```yaml
tools:
  mcpServers:
    - name: filesystem
      config:                       # transport inferred: command -> stdio
        command: npx
        args: ["-y", "@modelcontextprotocol/server-filesystem", "."]
    - name: linear
      config:
        transport: http             # Streamable HTTP; `sse` covers legacy servers
        url: https://mcp.linear.app/mcp
        headers: { Authorization: "Bearer ..." }
```

- `stdio` (local process: `command`/`args`/`env`/`cwd`), `http` (Streamable HTTP: `url`/`headers`), `sse` (legacy HTTP+SSE: `url`/`headers`); `transport` is inferable (`command` → stdio, `url` → http), `sse` stays explicit. Shared optional `connectTimeoutMs` / `timeoutMs` / `maxOutputLength` budgets.

**Integration shape**

- `environment/mcp/` resolves config, connects all servers in parallel on the first `listTools()` (single-flight, per-server connect budget, stderr tail in spawn-failure warnings), and exposes discovered tools as `BuiltinTool` wrappers named `mcp__<server>__<tool>` — Environment's execution contract (framing, timeout, truncation, interruption, approval) applies to them unchanged.
- Failure containment: invalid entries and unreachable servers produce stderr warnings and are skipped; Session creation is never blocked.
- Permission mapping for the read-only approval mode: `readOnlyHint === true` → `"r"`, everything else `"rw"` (annotations are untrusted hints; the default takes the restrictive direction).
- stdio server env is the SDK's safe inherited defaults plus the entry's own `env` — the Agent vault is deliberately **not** injected into MCP server processes (unlike command subprocesses); a variable a server needs must be listed explicitly in its entry. `cwd` defaults to the Session Workspace. `Environment.dispose()` closes every client, stdio child processes with it.
- Result mapping: text blocks → output text; image blocks → data-URL images; audio/binary resources → placeholder lines; `structuredContent` serialized only when no text block exists; `isError` → `stop_reason: "failed"`.
- The SDK's own per-request timeout is parked at the setTimeout maximum so Environment stays the single timekeeper through the shared abort signal.

**Web / server surface**

- The agent settings page's Tools tab gains vault-style MCP Server management (was a read-only JSON dump): a table of configured servers plus an Add/Edit modal — transport tabs on top (http default — url / headers; stdio: command / args / env / cwd; shared budget fields prefilled with their effective defaults), delete behind a confirmation, immediate persistence, and an extras carry-over so unknown config keys survive editing (zh + en strings).
- Connectivity testing mirrors the models page: a standalone **Test connection** button in the modal's top action row probes the current form values (result as a toast — tool count + latency), and a section-level button tests every configured server sequentially behind a confirm dialog, landing a tone-colored badge on each row (the group speed-test idiom).
- The server's `mcpServers` PUT validation runs every entry through the core transport resolver plus a duplicate-name check — a broken entry is a precise 400 at save time instead of a warn-and-skip at the next Session start. A new `POST …/config/mcp-test` route backs both test buttons: validate (400 for malformed), connect, discover, close — `{ ok, tools, latencyMs }` or `{ ok: false, error, latencyMs }`, nothing written.
- Docs: `tools` / `configuration` / `interfaces` / `web-app` pages updated (zh + en).

**Visible connect phase + protocol split (breaking)**

Session creation no longer blocks on MCP connects: the toolset resolves lazily at the start of the first run, streamed as protocol —

- `session_meta` **drops its `tools` field**; the full schemas follow as a new `tool_list_ready` event once discovery completes. **Explicitly incompatible with pre-split Traces** (decision: no compat code): their embedded tool record is no longer read or displayed; everything else about old traces renders unchanged.
- The connect + discovery wait is bracketed by one `mcp_connect_begin` / `mcp_connect_end` pair (end carries an overall compaction-style `status` — completed / failed / aborted — plus the per-server results; connects run in parallel; wall time = the pair's timestamp difference). In the Trace the pair and `tool_list_ready` are written right AFTER the run's input, so the whole bootstrap phase belongs to the new turn — a resumed session's reconnect included. The web chat renders the phase as a **unified step row** (strictly the reasoning-&-tools group-header shell — sticky title bar, duration slots, collapse scroll-back — shared with the compaction row): running/success/failure keep one shape, the settled row leads with the discovered-tool count and names unavailable servers, and expands into one group per server (status, tool count, that server's connect time), each opening into its tool list or failure detail; the CLI prints paired `[mcp]` lines (failure reasons inline on the end line); the timeline renders an "mcp connect" span in its own **"other" legend category** (not a tool execution).
- The connect is **abortable and cancelled on abort**: pending SDK connects are aborted, the provider resets, and the next send reconnects from scratch (locked by test). Continuing a conversation never reconnects (engine survives across tasks). The aborted turn is **recorded**: the Session itself writes the input, the aborted connect pair and the abort event to the Trace (no engine exists to do it), and the input **carries into the next send** — delivered to the model once, never re-written (one-shot skip list on the engine's trace sink).
- Trace event rows render tool parameter schemas as a property table (name / type / required / description) instead of raw JSON. Connect failures surface their per-server error detail in the chat row and inline on the CLI end line (previously stderr-only). The CLI declares `@modelcontextprotocol/client` itself so tsup keeps the SDK external — bundling it (via core's noExternal) inlined CJS deps and crashed every CLI command at startup.
- Compaction rotation rewrites `tool_list_ready` next to `session_meta`; replay ignores the new events (live streams re-emit them on the next run). Credential validation stays at creation time (a discarded bare-LLM construction keeps the `model_credential_missing` 400); `compact()` on a never-run session stays a strict no-op.
- Leaving a still-running conversation (switch / "new conversation") no longer fires a phantom completion: the chat page's reload-on-idle trigger is now guarded by session identity — the stream resets to "idle" whenever it detaches, and treating that as a Task completion reloaded the sessions + agents contexts right after the click (an occasional app-wide flicker on entering the draft).
- The session manager holds the running task's published input envelopes AND its streamed bootstrap records (connect pair / `tool_list_ready`) until the run ends; `GET /messages` appends whichever the Trace read hasn't caught up to (exact-envelope dedup) — during the connect window a history rebuild used to lose the user's own message and show no connecting status at all (a silent blank while a slow server timed out); this also closes the narrower pre-MCP race.

Discovery is a Session-lifetime snapshot (`tools/list_changed` ignored) — a candidate follow-up, as is applying the admin proxy policy to MCP stdio processes.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (core 713 | server 567 | web 641 | cli 235, incl. new suites: core MCP config validation + stdio end-to-end fixture covering discovery, framing, isError, images, entry-env/no-vault, cwd default, per-server truncation and timeout, user interruption, unreachable-server fallback + Streamable HTTP end-to-end asserting header injection; server mcpServers PUT round-trip, 400 cases and mcp-test reachable/unreachable/malformed; web form-helper suite; session first-run bootstrap suite (session_tools_ready): event order and trace writes with/without MCP, no repeat on the second run, per-server connect outcomes incl. failures; pending-input serving during the bootstrap window incl. dedup and idle clearing)
- ci / ci-windows green as of the previous push; the follow-up commits rerun the same suites.
- Specs sync for the protocol split (review blocker): Prism-Shadow/penguin-harness-design#34.


hiyouga:
## Review

Reviewed at `ca1dd45` in an isolated worktree. Verification re-run locally, all green:

| Check | Result |
|---|---|
| `pnpm -r build` | pass |
| `pnpm typecheck` | pass |
| `pnpm format:check` | pass |
| `pnpm -r test` | pass — core 706 · server 561 · web 631 · cli 235 · desktop 43 |

Overall the shape is good: `environment/mcp/` joins purely through the `BuiltinTool` interface, so Environment's execution contract (framing, timeout, truncation, interruption, approval) covers MCP tools without a single change to the dispatch path; failure containment is consistent (warn-and-skip everywhere); the test suite is substantial (stdio end-to-end fixture, Streamable HTTP header injection, per-server budgets, interruption, vault isolation, first-run bootstrap ordering). Findings below.

---

## Blockers

### Merge conflict with `main`

The merge base is `047505d`; `main` has since moved to `3b493df`. A three-way merge conflicts in exactly one file — `packages/web/src/features/chat/chat-page.tsx` — everything else auto-merges cleanly. Needs a rebase.

### `design/specs` not synced

`specs/05-ARCHITECTURE.md:113` and `:656` still declare `tools: ToolDefinition[]` inside the `session_meta` payload — the field this PR removes — and there is no spec text for `tool_list_ready`, the `mcp_connect_*` pair, the three transports, or `otherSpans`. No open design PR covers it. Given the project rule that design docs are ground truth and must move with interface changes, and that this is an explicitly breaking protocol change, the spec sync should land alongside.

(Changelog is covered by #243, consistent with the batch convention — no issue there.)

---

## Findings

### High — `appendPendingInputs` duplicates the user's message on long runs

`packages/server/src/http/routes/sessions.ts:107`

The dedup window is `messages.slice(-50)`, but `entry.pendingInputs` / `entry.pendingBootstrap` are held for the **entire run**, not just the bootstrap window — they are only cleared at idle. Once the current Task has written more than 50 trace records after the input, the input falls out of the window, is judged "the Trace hasn't caught up", and gets appended to the end of history a second time.

The comment justifies the slice with *"only the history tail can contain them (inputs are the newest records when this races)"* — that assumption holds during the connect window but not for the rest of the run.

Reproduced with a scratch test (seed a trace of `[input, ...60 assistant messages]`, adopt a parked running session holding the same envelope, `GET /messages`):

```
total messages: 62 | hello count: 2
last 3: ["step 58","step 59","hello world"]
```

The user's own message renders again after all the assistant output. Trigger paths are routine: reloading the page mid-run, opening the session in a second tab, a resync rebuild. For MCP sessions the held `mcp_connect_begin` / `mcp_connect_end` / `tool_list_ready` are re-appended the same way, producing a second MCP connect row at the bottom of the conversation. Note this is not MCP-specific — it hits every running session whose turn is long enough.

Suggested fix: clear both holds when the drive loop observes the run's first `request_begin`. `ContextEngine.run` writes input → `bootstrapRecords` → `toolList` → issues the request, so by the time `request_begin` is on the stream the Trace write has definitely landed, and the whole bootstrap window is still covered. Much cheaper than widening the dedup to the full message array.

### Medium — no validation of MCP-discovered tool names

`packages/core/src/environment/mcp/provider.ts:361`

`mcpToolName()` concatenates without checking the result, and `toolDefinitionsToSchemas` (`packages/core/src/llm/generative-model.ts:1394`) passes it to the provider verbatim. MCP does not restrict tool names to `[a-zA-Z0-9_-]` — names like `slack.postMessage` or `github/create_issue` exist in the wild — and `mcp__` + server + `__` + tool can also cross the 128-character limit. Either case makes **every** LLM request in the session a 400, which inverts the "warn and skip, never block the Session" stance this module holds everywhere else. Suggest validating in `register()` and skipping with a warning, the same way the duplicate-tool-name branch just above already does.

### Medium-low — contradictory comments on abort-mid-connect

`packages/core/src/session.ts:293` says *"the connect finishes in the background for the next run to reuse"*, while `ensureReady`'s docstring — and the code, which calls `cancelBootstrap()` — says the attempt is cancelled and the next run reconnects from scratch. The inline comment looks like a leftover from an earlier design.

---

## Minor

- **`connectTimeoutMs` is granted twice per server** (`packages/core/src/environment/mcp/provider.ts:343-350`). `client.connect(..., { timeout })` and `client.listTools(..., { timeout })` each receive the full budget, and the SDK's `timeout` is per-request (the total knob is `maxTotalTimeout`). Worst-case wall time per server is ~2× the configured value, while `ResolvedMCPServer.connectTimeoutMs` and the tools doc both describe it as the total "connect + handshake + tool discovery" budget.

- **Duplicate-name error shows the wrong message** (`packages/web/src/features/agents/mcp-servers-section.tsx:230`). `setFieldErrors({ name: { code: "required" } })` renders "this field is required" under a name field that is clearly filled in; the real `mcpDuplicateName` text only appears in the modal footer. `McpFormErrorCode` has no `duplicate` code.

- **`mcp-test` does not match the runtime environment** (`packages/server/src/http/routes/agent-config.ts:82`). `new McpToolProvider([entry], { warn })` passes no `workspaceDir`, so a stdio server's cwd falls back to the server process's directory instead of the Session Workspace. The doc's own example (`args: ["-y", "@modelcontextprotocol/server-filesystem", "."]`) resolves `.` differently under test than at runtime, so "Test connection" can pass while the real run fails. Two smaller things in the same handler: `warnings.length > 0 → ok:false` also trips on the benign "listed tool twice" warning, reporting a healthy server as failed; and `connectTimeoutMs` is unbounded from config, so one probe can hold a request open arbitrarily long (the bulk test walks every server sequentially).

- **Stale terminology.** `session_tools` still appears in `packages/core/src/session.ts:499`, `:521` and `packages/core/test/agent.test.ts:368` — the event is `tool_list_ready` now. `packages/cli/src/render.ts:367` and `packages/cli/src/tool-render.ts:24,176` still reference `session_meta.tools`, the field this PR removes.

- **`writeTrace` is described as shared but has one caller** (`packages/core/src/session.ts:521`). The JSDoc says it is shared by the bootstrap records (`session_meta` / `mcp_connect` / `session_tools` / `goal_finished`), but only `ensureMetaWritten` calls it.

- **Hardcoded version** (`packages/core/src/environment/mcp/provider.ts:61`). `CLIENT_INFO.version: "0.2.1"` duplicates `VERSION` in `packages/core/src/index.ts:62` and will drift at the next release bump.

- `readMs()` is reused to validate `maxOutputLength`, which is a character count rather than milliseconds — naming only, the error text is correct.

- `packages/web/src/features/agents/mcp-servers-section.tsx:302` calls `serverToForm(entry)` per row just to read `transport`, building the whole form state each render; a small `transportOf()` helper would be clearer.

hiyouga:
All review items addressed at `af9f248` (rebased onto `3b493df`; the one conflict was both sides adding a type to the same chat-page import list).

**Blockers**

- **Rebase** — done; branch now based on current main.
- **Specs sync** — Prism-Shadow/penguin-harness-design#34: `session_meta` drops `tools` from the payload spec (with the explicit-incompatibility note), new event specs for `mcp_connect_begin/end` (per-server `results`, overall compaction-style `status`, post-input turn attribution, abort-cancels semantics) and `tool_list_ready` (rotation rewrite, ignored by replay), `MCPServerConfig.config` transport/budget/naming/vault/containment semantics, and the timeline's "other" category in 06. `GenerativeModel`'s constructor `tools` field was left as is — the engine still constructs the LLM with the assembled tool list, so that declaration remains accurate.

**High — pending-hold duplication**: fixed exactly as suggested — the drive loop clears `pendingInputs` + `pendingBootstrap` at the run's first origin-less `request_begin` (engine order guarantees input → bootstrap records → tool list are on disk by then); the idle-time clear stays as a backstop for runs that never issue a request. The pending-input test now drives a fake session through `mcp_connect_begin` → gate → `request_begin` → gate and asserts both holds are empty mid-run while the session is still running.

**Medium — tool-name validation**: `register()` now rejects names outside `^[a-zA-Z0-9_-]{1,128}$` (the prefixed form) with a warning, same stance as the duplicate branch. Per-server `tools` counts now report *registered* tools, so the connect row/`tool_list_ready` stay consistent. The stdio fixture gained a `dot.name` tool — every existing count assertion staying at 6 doubles as the skip proof, plus an explicit skip-and-warn test.

**Medium-low — contradictory abort comment**: the stale "finishes in the background" line now states cancel-and-reconnect.

**Minor**

- `connectTimeoutMs` is one budget: `listTools` gets the handshake's remainder (`max(1, budget − elapsed)`), not a second full grant.
- Duplicate server name renders `mcpDuplicateName` under the name field via a new `duplicate` error code (footer-only message removed).
- `/mcp-test` runs the probe in a fresh `mkdtemp` workspace (a new session's temporary workspace is exactly an empty dir, unlike the server process cwd), takes its verdict + error from `connectResults()[0]` instead of the warning count (benign warnings no longer fail a healthy server), and caps the probe's connect budget at 30s.
- Stale `session_tools` / `session_meta.tools` references in session.ts, agent.test.ts, cli render.ts/tool-render.ts updated to `tool_list_ready`; `writeTrace`'s JSDoc now describes its single caller.
- `CLIENT_INFO` version: read lazily from core's `VERSION` (the import is circular — index → agent → environment — so a top-level read would hit the TDZ; the release workflow's sed anchor in index.ts is untouched).
- `readMs` → `readPositiveInt`; per-row `serverToForm` → `transportOf()`.

Verification re-run: build / typecheck / format clean; core 713 · server 567 · web 641 · cli 235 all green.

PR #249, opened 2026-08-11
fix(core): single-write Trace appends and crash-torn tail healing

hiyouga:
## Problem

A field incident left a subagent's Trace shard with two unparseable JSONL lines: a ~1.48 MB PNG `image_url` record was split across two lines with a `request_end` event spliced into the middle of its base64 data, so the parent conversation's messages endpoint failed with `Expected ',' or ']' after array element in JSON at position 524290` (HTTP 500 on the pre-#234 reader; the position sits 2 bytes past 512 KiB).

Root cause: `fs.appendFile` splits payloads larger than 512 KiB into multiple underlying writes (Node's internal `kWriteFileMaxChunkSize = 524288`; verified empirically on Node 24.18: a 700 KiB append issues `write()` calls of `[524288, 192512]`). Any record bigger than 512 KiB therefore had a tear window at the chunk boundary:

- **Before #234** (all released builds up to v0.2.1): appends were not serialized, so a concurrent append (parallel tool events vs. the model stream) could land *between* the chunks of a big record — exactly the observed splice.
- **After #234**: concurrency is serialized, but a process death between chunks still leaves exactly the first 524288 bytes of the record on disk; since Session resumption keeps appending to the same file, the next record then glues onto the torn line and is swallowed together with it.

## Fix

- `Writer` now appends each record through a **single `write(2)`** on an O_APPEND handle instead of `appendFile`: a crash can at most truncate the *last* record, never split one mid-file. A short-write loop remains as a safety net; a mid-record failure (e.g. ENOSPC) marks the tail torn.
- Before the first append to a **pre-existing** shard (resumption), the writer probes the file's last byte: a file not ending in `\n` carries a crash-torn tail, and the next record is prefixed with a newline so it starts on a fresh line — the torn line stays isolated (skipped by the tolerant readers) and never corrupts records written after it. A spurious heal costs one blank line, which every reader already ignores.
- `rotate()` resets the torn flag (the next shard is always a brand-new file).
- Docs (`sessions-and-traces` zh/en) document the single-write and tail-heal guarantees.

No format change: shard files remain plain JSONL, readable by old and new readers alike (blank lines were already ignored by every reader, including v0.2.1's).

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (core 671, server 557, cli 235, web 627, desktop 43, docs 32, landing 39, skills 18)
- New tests: a FileHandle-level spy pins the >512 KiB record to exactly one underlying `write()`; resume-onto-torn-tail heals onto a fresh line and both intact records survive `parseTraceLines`; clean/empty pre-existing files get no spurious prefix; a healed shard is probed once and rotation leaks no heal prefix into the next shard.

### DIFF
--- a/packages/core/src/trace/writer.ts
+++ b/packages/core/src/trace/writer.ts
@@ -116,7 +116,7 @@ export class Writer {
   private index = 1;
   /** Index whose file is prepared (date directory created, pre-existing tail probed); avoids redoing either per append. */
   private preparedIndex = -1;
-  /** True while the current shard may end mid-record (crash-torn tail found by the probe, or a failed append in this process): the next record starts on a fresh line. */
+  /** True while the current shard may end mid-record — assigned per shard by the tail probe, set by a mid-record append failure, cleared once a record fully lands. The next record then starts on a fresh line. */
   private tornTail = false;
   /** Serialization chain: mutating operations run strictly in submission order (see class docs). */
   private chain: Promise<void> = Promise.resolve();
@@ -165,7 +165,7 @@ export class Writer {
       const path = this.currentPath();
       if (this.preparedIndex !== this.index) {
         await mkdir(dirname(path), { recursive: true });
-        await this.probeTornTail(path);
+        this.tornTail = await this.probeTornTail(path);
         this.preparedIndex = this.index;
       }
       const record = `${JSON.stringify(msg)}\n`;
@@ -175,25 +175,25 @@ export class Writer {
   }
 
   /**
-   * Detects a crash-torn tail in a pre-existing shard file (resumption continues the
-   * original file): a non-empty file whose last byte is not `\n` ends mid-record, and the
-   * next append must not merge into that line. A missing file is simply fresh; any other
-   * probe error surfaces to the caller like an append failure (and is retried with the next
-   * write, since the index stays unprepared).
+   * Whether a pre-existing shard file ends mid-record (resumption continues the original
+   * file): a non-empty file whose last byte is not `\n` carries a crash-torn tail, and the
+   * next append must not merge into that line. A missing or empty file is simply fresh; any
+   * other probe error surfaces to the caller like an append failure (and is retried with the
+   * next write, since the index stays unprepared).
    */
-  private async probeTornTail(path: string): Promise<void> {
+  private async probeTornTail(path: string): Promise<boolean> {
     let fh;
     try {
       fh = await open(path, "r");
     } catch (err) {
-      if ((err as NodeJS.ErrnoException).code === "ENOENT") return;
+      if ((err as NodeJS.ErrnoException).code === "ENOENT") return false;
       throw err;
     }
     try {
       const { size } = await fh.stat();
-      if (size === 0) return;
+      if (size === 0) return false;
       const { bytesRead, buffer } = await fh.read(Buffer.alloc(1), 0, 1, size - 1);
-      if (bytesRead !== 1 || buffer[0] !== 0x0a) this.tornTail = true;
+      return bytesRead !== 1 || buffer[0] !== 0x0a;
     } finally {
       await fh.close();
     }
@@ -261,10 +261,9 @@ export class Writer {
    */
   async rotate(): Promise<void> {
     return this.enqueue(async () => {
+      // No torn-tail bookkeeping here: the first write to the new index re-probes and
+      // reassigns the flag, so the previous shard's state cannot leak into the next file.
       this.index += 1;
-      // The next shard is a brand-new file: a torn tail of the previous shard must not leak
-      // a heal prefix into it (the probe will re-evaluate if the file somehow pre-exists).
-      this.tornTail = false;
     });
   }
 }
--- a/packages/core/src/trace/writer.ts
+++ b/packages/core/src/trace/writer.ts
@@ -16,7 +16,7 @@
  *     context_engine).
  *   - Path convention: `<tracesDir>/<yyyy-mm-dd>/<sessionId>_<index3>.jsonl`.
  */
-import { appendFile, mkdir, readFile } from "node:fs/promises";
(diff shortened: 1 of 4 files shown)


hiyouga:
## Review: single-write Trace appends and crash-torn tail healing

**Verdict: correct and well-scoped. No correctness defects found.** Two test-coverage gaps and a small robustness simplification are worth addressing; nothing blocking.

Reviewed in a dedicated worktree at `a19e8f0` (rebased on `origin/main` `a41034c`, 4 files, +212/−22).

### The premise and the fix were verified independently, at the syscall level

The central claim rests on Node internals, so I tested it rather than trusting it. Two independent *processes* appending ~900 KB records to the same file:

```
appendfile  lines=24 well_formed=6  torn=18
single      lines=24 well_formed=24 torn=0
```

`fs.appendFile` tore 18 of 24 records; the `open(…, "a")` + single `fh.write` path tore none. That confirms both halves — `appendFile` really does issue multiple `write(2)` calls per record, and the replacement really is one kernel-atomic O_APPEND write. Node 24.18, ext4.

A related detail the PR's own test can't show: `fs.appendFile` does **not** route through `FileHandle.prototype.write` at all (a spy records `[]` for it, `[1228801]` for the new path). The new test still catches a regression to `appendFile` — via the `toContain` assertion going empty — but it catches it by accident of that fact, not by observing chunking.

### Correctness analysis

Every state transition of the two new pieces of state was walked:

- `tornTail` is only ever read inside the serialized chain, and cleared only after the record *and* its heal prefix fully landed. A throw at any point leaves it set, so the heal survives to the next append. ✅
- `preparedIndex` gates mkdir+probe per shard; a probe failure leaves it unset so the next write retries. ✅
- `rotate()` clearing `tornTail` is required and correct — and the comment's parenthetical about re-probing a pre-existing next shard is right (that's the resume-then-compact case). ✅
- Records can't contain an interior `0x0A` (`JSON.stringify` escapes newlines), so "last byte isn't `\n`" is a sound torn-tail test. A partial write cutting a multi-byte UTF-8 sequence still can't produce a false-clean probe. ✅
- Reader side holds up: `parseTraceLines` skips blanks and tolerates the trailing truncated line; `readTraceHead` cuts at the last newline; import's `wx` never touches a live shard. Nothing in web/server splits trace JSONL independently. ✅
- `packages/core/src/agent.ts:448` confirms resumption really does construct the Writer with `dateDir`/`startIndex` pointing at the existing shard, so the heal path is live, not dead code. ✅

### Findings

**1. The two failure branches in `appendRecord` ship untested** (`packages/core/src/trace/writer.ts:218` and `:224`). The five new tests cover the probe and the happy path, but nothing exercises the in-process partial-append → `tornTail = true` → heal-on-next-write flow, or the `bytesWritten <= 0` bail-out. That's the subtlest logic in the change.

Both tests were written locally to check the code actually works — **it does**, both pass. They're ~90 lines using the same prototype-spy trick as the existing test (force a half-length write then reject with ENOSPC; force a zero-progress write). Worth landing so the branches don't rot.

**2. `probeTornTail` sets `tornTail = true` but never `false`** (`packages/core/src/trace/writer.ts:184`), so correctness depends on `rotate()` clearing it from another method. Having the probe *return* a boolean and assigning `this.tornTail = await this.probeTornTail(path)` makes the flag a pure function of the probe, and makes `rotate()`'s reset redundant (the new index re-probes and assigns `false` on ENOENT anyway). Same behavior, one less invariant held at a distance.

**3. Test-assertion precision** (`packages/core/test/trace.test.ts:250`): the comment says "never in 524288-byte chunks", but `filter((len) => len > 512 * 1024)` cannot see a chunk of *exactly* 524288 — that's `>`, not `>=`. The test still fails correctly on a regression, just not via the assertion that reads like the guard.

**4. No `changelog/unreleased/` entry.** Per `CLAUDE.md` this belongs in the existing `2026-08-07-core-runtime.md`, whose summary line already covers the #215 Trace-serialization fix that this directly follows up. Consistent with how this repo batches changelog updates into separate PRs, so likely intentional — flagged so it isn't forgotten in the next batch.

**Design docs:** correctly untouched. `design/specs` describes Trace only as "append-only JSON Lines, never modified in place" and says nothing about the write mechanism; #234 set the precedent of documenting this layer in `packages/docs/content/sessions-and-traces.{zh,en}.md` only. The zh bullet mirrors the en one faithfully.

### Verification claims in the PR description — all accurate

Everything was re-run: build ✅, `pnpm -r typecheck` ✅, `pnpm format:check` clean ✅, and the test counts match the description exactly — core 671, server 557, cli 235, web 627, desktop 43, docs 32, landing 39, skills 18, all passing. CI is green on all four checks including `ci-windows`.

One caveat for anyone reproducing locally: `npx vitest run` from the repo root sweeps the Playwright e2e specs into vitest and produces spurious failures. Use `pnpm -r test` (the root `test` script) — that's clean.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

hiyouga:
Findings 1–3 addressed in 30677c4:

1. Both `appendRecord` failure branches are now covered: a half-written record followed by ENOSPC (via the same prototype-spy trick) marks the tail torn and the next append heals onto a fresh line with both intact records surviving `parseTraceLines`; a zero-progress write bails out with the tail left clean (no spurious heal prefix).
2. `probeTornTail` now returns the verdict and `write()` assigns it — `tornTail` is a pure function of the per-shard probe plus mid-record failures, and `rotate()` no longer carries the reset invariant.
3. Chunk-size filter tightened to `>= 512 * 1024`.

Finding 4 (changelog): deferred to the next changelog batch PR as suspected — will extend `changelog/unreleased/2026-08-07-core-runtime.md` there.

Full chain re-verified: `pnpm -r build` / `pnpm typecheck` / `pnpm test` / `pnpm format:check` all clean; trace suite now 14 tests.

PR #254, opened 2026-08-11
feat(web): dismissible initial-password banner

hiyouga:
## Motivation

The initial-password notice banner (shown while `passwordIsInitial` is set, outside desktop mode) could not be dismissed — it nagged on every page until the password was changed. Users who understand and accept the risk can now permanently close it. The choice is persisted per-user in `ui_prefs` (new optional key `initialPasswordBannerDismissed`; the free-form prefs table needs no server route/storage change or migration).

## Behavior

- A small amber-toned close (X) button sits at the right edge of the banner. It reuses the shared `CloseButton` glyph and `aria-label` (`S.common.close`, no new i18n keys); the shared component itself hardcodes gray hover colors that clash with the amber banner, hence the local twin.
- Clicking it hides the banner immediately and fire-and-forgets `PUT /api/me/prefs { initialPasswordBannerDismissed: true }` (shallow merge; a lost write only costs persistence).
- **Hydrate before render to avoid flash**: the banner renders only after `GET /api/me/prefs` confirms the key is not `true`, so an already-dismissed banner never flashes and vanishes on load. The hydrate effect runs only when the banner would show at all (`passwordIsInitial && !desktopMode`), with a cancelled guard.
- **Fail-open on prefs error**: if the prefs fetch fails, the banner shows as before.

## Changes

- `packages/server/src/api/types.ts`: declare optional `initialPasswordBannerDismissed?: boolean` on `UiPrefs` (type-only; table is free-form).
- `packages/web/src/components/layout/app-layout.tsx`: hydrate state, dismiss handler, and the close button on the banner.

## Verification

All run at the repo root on Node v24:

- `pnpm -r build` — pass
- `pnpm test` — pass (exit 0; e.g. web 646/646, desktop 48/48)
- `pnpm format` + `pnpm format:check` — clean ("All matched files use Prettier code style!")

### DIFF
--- a/packages/server/src/api/types.ts
+++ b/packages/server/src/api/types.ts
@@ -180,6 +180,8 @@ export interface UiPrefs {
    * scanning (#139).
    */
   showCliSessions?: boolean;
+  /** The initial-password notice banner (app layout) was permanently dismissed by the user. */
+  initialPasswordBannerDismissed?: boolean;
   [key: string]: unknown;
 }
 
--- a/packages/web/src/components/layout/app-layout.tsx
+++ b/packages/web/src/components/layout/app-layout.tsx
@@ -270,15 +270,17 @@ export function AppLayout() {
               {S.account.changeNow}
             </button>
             {/* Amber-toned twin of the shared CloseButton (same glyph + aria-label) — its hardcoded
-                gray colors would clash here. Absolutely positioned at the right edge: near-full-height
-                hit area without growing the banner and without transform (see the stacking-context
-                note in the file header); the banner's symmetric px-8 keeps the centered text clear. */}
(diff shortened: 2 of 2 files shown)


hiyouga:
Follow-up per user feedback: the dismiss X is now flat — no background fill in any state (removed `hover:bg-amber-100` / `dark:hover:bg-amber-900/40`); hover feedback is icon-color darkening only, matching the change-now link's hover shades (`hover:text-amber-950` / `dark:hover:text-amber-100`). aria-label, position and size unchanged.

Verification rerun at the repo root (Node v24): `pnpm -r build`, `pnpm typecheck`, `pnpm test` (exit 0, all suites pass), `pnpm format` + `pnpm format:check` (clean) — all pass.

PR #255, opened 2026-08-11
fix(web): align agent tab order with stat icons and even out stat spacing

hiyouga:
## Summary

- Establish a single canonical order for the agent settings tabs: **Overview, System Prompt, Runtime, Tools, Skills, Memory, Vault, Schedules** (Memory moves from position 3 to after Skills). The `TabKey` union and file-header comment are updated to match; tab rendering and `?tab=` deep-link resolution are order-independent and unchanged.
- Reorder the agents-list stat row to mirror the tab order: sessions → tools → skills → memory → vault → schedules → last-modified. Sessions and last-modified are not tabs and keep their first/last positions; the five clickable counts in between now deep-link in the same order as the settings tabs they open. JSX blocks were moved only — no per-item logic changed.
- Rename the prompt tab label: zh `Prompt` → `系统提示词`, en `Prompt` → `System Prompt`.
- Even out stat-row spacing: removed `min-w-[2.25rem]` from `STAT_LINK_CLASS` and the sessions span so each item hugs its content and the container's uniform `gap-x-2.5` governs spacing (`tabular-nums` kept). This trades cross-card column alignment for consistent in-row gaps, as requested — short counts no longer leave oversized visual gaps, and the last item (last-modified, which never had a min-width) no longer looks inconsistently spaced.

## Verification

- `pnpm -r build` — pass
- `pnpm test` — pass (all packages; server suite initially hit hook timeouts under a transient load spike from parallel agents, clean 584/584 on rerun)
- `pnpm format` + `pnpm format:check` — pass, no formatting drift

### DIFF
--- a/packages/web/src/features/agents/agent-settings-page.tsx
+++ b/packages/web/src/features/agents/agent-settings-page.tsx
@@ -3,10 +3,10 @@
  * Overview (name/description form + two ruled sections in the skills import modal's
  * family: Agent State — State version, snapshot export-import and the copyable State
  * path — and Kernel — the defaults generation with its update / restore-defaults
- * actions), Prompt (AGENTS.md and system_prompt editors + placeholder
- * reference), Memory (memory-tab.tsx), Runtime (max_turns, model.*, compaction.*), Tools (editable built-in
+ * actions), System Prompt (AGENTS.md and system_prompt editors + placeholder
+ * reference), Runtime (max_turns, model.*, compaction.*), Tools (editable built-in
  * tools table + the MCP Server form, mcp-servers-section.tsx), Skills (skills-tab.tsx),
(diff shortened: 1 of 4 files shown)


hiyouga:
Follow-up per user feedback: with the min-widths gone the stat items sat too close together, so the stat row's container gap is widened from `gap-x-2.5` to `gap-x-4` (single change point — every item stays equidistant; the row doesn't wrap, so no `gap-y` counterpart is needed). The comment referencing the gap value is updated to match; min-widths stay removed.

Verification re-run after the change: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (all packages green), `pnpm format` + `pnpm format:check` — all pass.

PR #257, opened 2026-08-11
feat(core,server,web): memory-style prompt injection toggles for skills, vault and schedules

hiyouga:
## Motivation

The Vault and Skills sections were hardcoded template text with inline `{{VAULT_KEYS}}` / `{{SKILL_METADATA}}` tokens, and Schedules had no prompt presence at all — the model could fire-and-receive scheduled prompts but was never taught the file-based task system. This PR brings all three features to the same pattern Memory already uses:

- **Template holds only section placeholders**: `{{VAULT}}`, `{{SKILLS}}`, `{{MEMORY}}`, `{{SCHEDULES}}` (in that order, before `# Environment`). Each expands to an editable per-feature prompt (`vault.prompt` / `skills.prompt` / `schedules.prompt` in `system_config.yaml`) carrying its inner injection point (`{{VAULT_KEYS}}` / `{{SKILL_METADATA}}` / `{{SCHEDULE_LIST}}`).
- **Per-feature on/off toggle** (`vault.enabled` / `skills.enabled` / `schedules.enabled`, default on): off empties the whole block. The toggles govern prompt injection only — vault values still reach subprocess environments, installed skills remain invocable via `[use_skills]`, and the server still fires scheduled tasks; the tab copy states this explicitly.
- **Each feature's tab** (Skills / Vault / Schedules) mirrors the Memory tab: an immediate-write switch card, an amber alert with one-click insert/migrate when the template lacks the placeholder, and a mono prompt editor with placeholder chips and confirm-first save (shared `usePromptInjection` hook). Vault/Schedules follow their tabs' existing owner-only gating; Skills stays member-level.
- **New `schedules.prompt`** teaches the model file-based CRUD over `agent_state/schedule/*.toml`: the directory as a literal `<app_data_dir>` pattern, a TOML example, the exact field rules `schedule-file.ts` enforces (`enabled` defaults to false so it must be set true, `period` ≥ 5m, `session_id` vs `workspace`+`provider`/`model_id` exclusivity, paired model reference, ~30s automatic reconcile), hygiene rules against duplicates, and the current task-name roster via `{{SCHEDULE_LIST}}` (empty-roster note when none exist). Session creation now collects the task names (`listScheduleNames`, contents never read).
- **Single-pass expansion**: all four section placeholders expand last in one `replace` pass; replacement products are never rescanned, so tokens smuggled into memory indexes or into one section's prompt can neither re-expand nor trigger another section (preserving and extending the existing `{{MEMORY}}`-last anti-smuggling property).

## Backward compatibility

`system_config.yaml` is materialized at Agent creation and never auto-upgraded, so existing agents carry the old hardcoded sections. Dual-format support with one-click migration (as decided):

- **Legacy inline placeholders keep working**: `{{VAULT_KEYS}}` / `{{SKILL_METADATA}}` written directly in a template body are still substituted — and now honor the toggles (an off switch substitutes an empty string), so the new switches work on old templates too.
- **One-click migration**: the Vault/Skills tabs detect a template still carrying the legacy default section verbatim (`LEGACY_VAULT_SECTION` / `LEGACY_SKILLS_SECTION`, frozen copies of the old text) and offer replacing it in place with the new placeholder — the assembled prompt is byte-identical since the default prompts moved the wording without changing it. Customized templates fall back to plain insert-before-`# Environment` (the memory-tab convention). Schedules only ever needs the plain insert.
- **Retirement comments**: the legacy constants, the migration branches, and the template-level inline substitution all carry `withShellLineFallback`-style doc comments with an explicit retirement condition — remove once pre-`{{VAULT}}`/`{{SKILLS}}` templates are no longer expected in the wild.

Server DTOs report the template facts read-only (`templateHasPlaceholder`, `legacySectionPresent`) next to the effective `enabled`/`prompt` values; `POST …/{skills,vault,schedules}/template-placeholder` mirrors memory's endpoint (idempotent config write).

## Verification

- `pnpm -r build` — all packages green
- `pnpm typecheck` — all packages green
- `pnpm test` — all suites green (core 774 passed / 5 skipped, server 590, web 646, cli 235, skills/docs/landing/desktop all passing), including:
  - new `packages/core/test/prompt-sections.test.ts` (25 tests): section rendering, toggles, default-prompt fallbacks, no-placeholder templates, single-pass no-smuggling (between sections and via memory index), legacy inline substitution under both toggle states, verbatim migration reproducing the current default template exactly, insert positioning/idempotence, `listScheduleNames`, Session-prompt integration
  - new `packages/server/test/prompt-sections.test.ts` (6 tests): DTO reporting, toggle/prompt round-trips through PUT config, legacy-template detection + per-feature migration + idempotence, `{{SCHEDULES}}` insert positioning, permission gating (skills member-level; vault/schedules owner-only 403, outsiders 404)
  - `packages/web/test/placeholders-parity.test.ts` green against the new default template token order in both dictionaries

### DIFF
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -17,6 +17,7 @@ import {
   formatModelRef,
   getModel,
   listInstalledSkills,
+  listScheduleNames,
   loadAgentVault,
   loadOrInitAgentState,
   loadProjectConfig,
@@ -262,6 +263,13 @@ export class Agent {
       this.state.projectId,
       this.state.agentId,
     );
+    // Schedule task names for the {{SCHEDULES}} roster: read fresh like the vault and Skills
+    // above (names only; the files' contents are the server-side scheduler's concern).
+    const scheduleNames = await listScheduleNames(
+      this.state.root,
+      this.state.projectId,
+      this.state.agentId,
+    );
     // Memory for this Session: null when the Agent has Memory off; a temporary Workspace gets
     // the user scope only (nothing written against it could ever be read back). Reads the
     // current indexes every time, like the vault and Skills above.
@@ -278,7 +286,7 @@ export class Agent {
     // into the prompt (so the model knows which API keys are available); values only
     // go into the subprocess environment. Skills only inject metadata (name and
     // description); the model reads the body on demand via shell. Memory likewise injects
-    // only its index; topic bodies are read on demand.
+    // only its index; topic bodies are read on demand. Schedules inject only task names.
     const systemPrompt = assembleSystemPrompt(
       this.state,
       sessionEnvironment(workspaceDir, sessionId, {
@@ -290,6 +298,7 @@ export class Agent {
       Object.keys(vault),
       installedSkills,
       memory,
+      scheduleNames,
     );
 
     const rt = await this.buildRuntime({
--- a/packages/core/src/state/agent-state.ts
+++ b/packages/core/src/state/agent-state.ts
@@ -30,6 +30,11 @@ import {
   MODEL_ID_PLACEHOLDER,
   DATE_PLACEHOLDER,
   MEMORY_PLACEHOLDER,
+  VAULT_PLACEHOLDER,
+  SKILLS_PLACEHOLDER,
+  SCHEDULES_PLACEHOLDER,
+  SCHEDULE_LIST_PLACEHOLDER,
+  SCHEDULE_LIST_EMPTY_NOTE,
   WORKSPACE_MEMORY_DIR_PLACEHOLDER,
   WORKSPACE_MEMORY_INDEX_PLACEHOLDER,
   USER_MEMORY_INDEX_PLACEHOLDER,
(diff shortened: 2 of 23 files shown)


hiyouga:
Follow-up (93e4592) addressing review feedback:

1. **Toggle cards now match the memory tab exactly** — labels switched from "Inject … section" to enable-style ("启用技能 / 启用密钥保险柜 / 启用定时任务"; "Enable skills / Enable vault / Enable schedules"), and the injection-semantics hint line under the label was removed, leaving label + switch only. The unused `enableHint` keys were deleted from both dictionaries (zh/en stay key-aligned via the `Strings` type); `promptSectionHint` in the prompt editor below is untouched.
2. **Schedules table wrapping is now deliberate** — all headers and the compact columns (status, period, fire times, queue) are `whitespace-nowrap`; long-text columns truncate with the full value on hover (`name` gains `max-w-40 truncate` + title, target already had it); the next/last-fired cell keeps its two-line stack with each line nowrap; table `min-w` raised 640→720px, horizontal scroll still handled by the existing `overflow-x-auto` container. Header comment updated to describe the wrapping rules.

Verification: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (all suites green — core 774/5 skipped, server 590, web 646, cli 235, others passing), `pnpm format` + `pnpm format:check` clean.

PR #259, opened 2026-08-11
fix(web): refine toggle switch styling

hiyouga:
Restyles the shared toggle switch (`packages/web/src/components/ui/switch.tsx`) after feedback that it looked off. Component-only change: size, API (`checked`/`onChange`/`disabled`/`className` passthrough), `button` + `role="switch"` + `aria-checked` semantics and disabled behavior are unchanged; no caller was touched.

## What changed

- **Lingering focus ring after mouse click** (the main defect): the ring was `focus:ring-2 focus:ring-gray-400/30`, so clicking the switch left a gray halo until the button lost focus. It is now `focus-visible`-only — keyboard focus still shows it, a mouse click doesn't — and accent-tinted (`ring-[var(--accent-bg)]/40`) so it matches the theme accent instead of a flat gray.
- **Track edge definition**: the track now carries an inset hairline (`inset-ring inset-ring-black/10 dark:inset-ring-white/10`, Tailwind v4's composable inset-ring channel, so it never collides with the focus ring). Applied in both states: over colored accents it reads as a subtle darker/lighter rim, and over the near-black/near-white neutral accents it simply disappears, so the checked state stays clean.
- **Off-state track colors**: light `bg-gray-300` → `bg-gray-200` (airier, iOS-like, with the hairline supplying the outline); dark `bg-gray-600` → `bg-gray-700`. The dark palette overrides gray-700/800/900 to true neutrals, while default `gray-600` is Tailwind's cool-tinted gray — the one tinted surface in an otherwise pure-neutral dark theme. `gray-700` (#303030) sits on the app's own ramp and keeps clear contrast against the white knob.
- **Knob**: shadow converged from `shadow` to `shadow-sm`, plus a `ring-1 ring-black/10` hairline. The hairline matters most in the dark **neutral** accent theme, where `--accent-bg` is near-white (#f3f4f6) and the white knob would otherwise vanish into the checked track. Dropped the redundant v3-era `transform` class.
- **Transitions**: `ease-out` added to both the track color and knob translate transitions (200ms unchanged).
- Header comment updated to describe the new focus/hairline behavior.

## Usage sites checked (no caller changes)

- `packages/web/src/features/agents/memory-tab.tsx` — bordered settings row
- `packages/web/src/features/agents/agent-settings-page.tsx` — tools table cell
- `packages/web/src/features/models/models-page.tsx` — vision toggle inside a `<label>`
- `packages/web/src/components/account/proxy-settings-dialog.tsx` — two dialog rows
- `packages/web/src/components/layout/sidebar.tsx` — settings popover row

All sit on white cards / dark `gray-900` surfaces (plus light gray blocks); the new track colors and hairlines were checked against each combination, including all `data-accent` presets and the neutral defaults.

## Verification

- `pnpm -r build` — pass (also confirmed the built CSS bundle contains the compiled `inset-ring`, `focus-visible:ring-2`, and `color-mix(... var(--accent-bg) 40%)` rules, and that the inset-ring and focus-ring box-shadow channels compose)
- `pnpm typecheck` — pass
- `pnpm test` — pass

### DIFF
--- a/packages/web/src/components/ui/switch.tsx
+++ b/packages/web/src/components/ui/switch.tsx
@@ -3,9 +3,12 @@
  * button is keyboard-operable out of the box — Space/Enter activate it — and is labelable
  * content, so clicking an enclosing `<label>`'s text toggles it too). Sliding knob with an
  * ease-out color/transform transition; the on-state follows the theme accent variable (same
- * source as Button primary), the off-state is gray. Track and knob each carry an inset/outer
- * hairline so the edges stay defined on any surface (the dark neutral accent is near-white,
- * where the white knob would otherwise vanish); dark-mode aware; disabled dims and blocks.
+ * source as Button primary), the off-state is gray. The 18px knob rides in the 20px track
+ * with a symmetric 1px gap all around (the iOS knob/track proportion) and carries no offset
+ * shadow — a directional shadow reads as vertical asymmetry at this size. Track and knob
+ * each carry a hairline (inset ring / outer ring) that meet across that gap, so the edges
+ * stay defined on any surface (the dark neutral accent is near-white, where the white knob
+ * would otherwise vanish); dark-mode aware; disabled dims and blocks.
  * The focus ring is accent-tinted and `focus-visible`-only, so keyboard focus shows it but a
  * mouse click doesn't leave a lingering halo. Sized for compact (sm) form rows, matching the
  * dialogs' controls.
@@ -41,8 +44,8 @@ export function Switch({ checked, onChange, disabled, className, ...rest }: Swit
     >
       <span
         aria-hidden
-        className={`inline-block size-4 rounded-full bg-white shadow-sm ring-1 ring-black/10 transition-transform duration-200 ease-out ${
-          checked ? "translate-x-[18px]" : "translate-x-0.5"
+        className={`inline-block size-[18px] rounded-full bg-white ring-1 ring-black/10 transition-transform duration-200 ease-out ${
+          checked ? "translate-x-[17px]" : "translate-x-px"
         }`}
       />
     </button>
--- a/packages/web/src/components/ui/switch.tsx
+++ b/packages/web/src/components/ui/switch.tsx
@@ -3,15 +3,21 @@
  * button is keyboard-operable out of the box — Space/Enter activate it — and is labelable
  * content, so clicking an enclosing `<label>`'s text toggles it too). Sliding knob with an
  * ease-out color/transform transition; the on-state follows the theme accent variable (same
- * source as Button primary), the off-state is gray. The 18px knob rides in the 20px track
- * with a symmetric 1px gap all around (the iOS knob/track proportion) and carries no offset
- * shadow — a directional shadow reads as vertical asymmetry at this size. Track and knob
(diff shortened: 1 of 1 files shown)


hiyouga:
Follow-up to review feedback on this PR: the knob read as detached from the track edge ("circle and border don't touch") and vertically asymmetric.

- **Knob size**: `size-4` (16px) → `size-[18px]` — an 18px knob in the 20px track (the iOS knob/track proportion), shrinking the gap from 2px to a symmetric 1px all around. The knob's outer 1px hairline now meets the track's inset hairline across that gap, so the two edges visually connect.
- **Symmetric travel**: off `translate-x-0.5` → `translate-x-px`, on `translate-x-[18px]` → `translate-x-[17px]` (36 − 18 − 1); 1px gap on every side in both states.
- **Dropped the knob's `shadow-sm`** — its 1px downward offset shadow was the source of the vertical asymmetry at this size. The knob's `ring-1 ring-black/10` hairline stays (still needed for contrast on the near-white neutral-accent track in dark mode), as does the track `inset-ring`.
- Header comment updated with the geometry rationale. Confirmed the built CSS bundle contains the compiled `size-[18px]` / `translate-x-px` / `translate-x-[17px]` rules, and that none of the five Switch call sites pass a `className` that could interfere with the geometry.

Verification re-run: `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format`, `pnpm format:check` — all pass.

hiyouga:
Second follow-up: the knob's edge should keep the same distance from the outline everywhere, in both states. The previous 18px-knob construction couldn't deliver that — the knob's hairline was an **outer** ring (drawn outside the element), so on the arc nearest the track edge it landed on top of the track's inset hairline and stacked into one heavier line, while the far arcs showed a single line; and with the knob's outer edge flush with the track's outer edge there was no clearance at all at the closest point.

New construction — the knob sits entirely **inside** the hairline with strictly uniform clearance:

- **Knob back to 16px** (`size-4`), and its hairline is now a `border border-black/10` — a border is drawn inside the 16px box, so it can never extend outward or overlap the track's hairline.
- **Travel**: off `translate-x-0.5` (2px), on `translate-x-[18px]` (36 − 2 − 16). All integer pixels.
- **Resulting geometry**: 36×20px track, 16px knob, 2px inset on every side in both states. At either end the knob center coincides with the track's end-cap circle center (2 + 8 = 10px from the edge), so the two curves are **concentric** — the gap along the arcs is exactly the gap along the straight runs. Net clearance between the knob's edge and the inner edge of the track's 1px inset hairline: **uniform 1px, everywhere, in both states**.
- Header comment updated with the geometry rationale (and why border-not-ring).

Visually verified with a throwaway Playwright script (not committed) rendering the exact class strings against the built CSS bundle: off/on x light/dark captured at deviceScaleFactor 1 and 2, magnified nearest-neighbor — clearance is even on all sides and around the caps at both DPRs, and the knob border still keeps the white knob defined on the near-white neutral-accent track in dark mode.

Verification re-run: `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format`, `pnpm format:check` — all pass; compiled bundle contains the `size-4` / `border-black/10` / `translate-x-0.5` / `translate-x-[18px]` rules.

PR #260, opened 2026-08-11
feat(core,server,web): agent config kernel version with smart update

hiyouga:
Stacked on #257 — GitHub will retarget to main when #257 merges; review only this PR's commits.

## What

Adds a "config kernel version" to Agent configs: a date-based stamp (`kernel_version` in `system_config.yaml`) recording which generation of the built-in defaults the config is based on, plus a smart, lossless "Update kernel" action beside the existing "Restore default configuration". Skill-style outdated hints appear on the agent settings overview and the agent list cards.

## Design

- **Date-based kernel version** (`KERNEL_VERSION`, first value `2026-08-11`): stamped by `defaultSystemConfig()`, so creation, restore-defaults and kernel update all stamp it. User edits never change it; a missing stamp means the config predates the mechanism and reads as outdated. Unrelated to `version`, the optimization counter.
- **Manual bump enforced by a hash-pin test**: `core/test/kernel-version.test.ts` recomputes every default leaf hash and fails whenever `defaultSystemConfig()` drifts from the latest `KERNEL_HASH_HISTORY` entry — the mechanical definition of "substantive default change ⇒ bump KERNEL_VERSION (new date) + append a history entry". Same-day changes may revise that day's entry; older entries are frozen.
- **Smart merge with per-path hash history** (`applyKernelUpdate`): traverses the current defaults' leaves; a missing value, or one hash-matching *any recorded generation's* default, advances to the new default; anything else is treated as a user customization and kept, reported in the result. `tools.builtin` merges per tool name (an edited tool is kept individually while the rest advance; user-added entries survive; deliberate removals stay removed — a tool is only appended when it is new in the latest generation). `name` / `description` / `version` / `tools.mcpServers` are never touched. Writes go through yaml `parseDocument`, so comments survive; the config is stamped last.
- **Conservative for unknown generations**: two generations are seeded — the current (post-#257) defaults, and the pre-#257 generation reconstructed byte-exactly from the frozen `LEGACY_*` constants (with a proof test tying the pinned hashes to that reconstruction). Older generations cannot be reconstructed: their values match no recorded hash and are conservatively kept — restore-defaults remains the full-refresh recourse. The docs state this explicitly.
[...]

PR #261, opened 2026-08-11
fix(web): card-style agent overview with copyable state path

hiyouga:
Stacked on #260 (which stacks on #257); review only this PR's commits.

Unifies the agent settings Overview tab around the card style #260 introduced for the kernel card, plus three user-feedback fixes.

## Changes

1. **Unified card styling / grouping** — the Overview's remaining flat blocks are regrouped into one new **Agent State** card in the kernel card's visual family (`rounded-lg border … px-4 py-3`, left title+info / right action slot):
   - Left: "Agent State" title with the State version line under it.
   - Right action slot: the snapshot **Export / Import** buttons (moved per item 4).
   - Below: the State path row, then the transfer description and import error as card footnotes.
   The resulting Overview is: name/description form + Save, Agent State card, Kernel card (unchanged). Grouping kept deliberately coarse — two cards total.
2. **Active Sessions stat removed** — the Overview no longer shows the active/running Session count (the server DTO field stays; nothing else displayed it). The `activeSessions` key was dropped from both dictionaries, and the now-unused `transferTitle` key was cleaned up alongside (its buttons now sit in the card's action slot). New keys: `stateTitle`, `copyStateDir` (zh/en).
3. **Copyable State path** — the path row adopts the chat details card's Session id convention: shared `CopyButton` with optimistic copy and check + "Copied" feedback at the button, and a `title` tooltip carrying the full path.
4. **Export / import beside the State version** — the existing snapshot controls moved into the Agent State card's right action slot, mirroring the kernel card layout. Owner-only import visibility, the conflict confirm modal and error display are unchanged.

No behavior changes; layout and i18n only.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (core / server / cli / web / desktop)

### DIFF
--- a/packages/web/src/features/agents/agent-settings-page.tsx
+++ b/packages/web/src/features/agents/agent-settings-page.tsx
@@ -1,8 +1,8 @@
 /**
  * Agent settings page: eight tabs —
- * Overview (name/description/State path/active count/State version + a kernel card grouping
- * the defaults generation with its update / restore-defaults actions + snapshot
- * export-import), Prompt (AGENTS.md and system_prompt editors + placeholder
+ * Overview (name/description form + an Agent State card grouping the State version,
+ * snapshot export-import and the copyable State path + a kernel card grouping the
+ * defaults generation with its update / restore-defaults actions), Prompt (AGENTS.md and system_prompt editors + placeholder
  * reference), Memory (memory-tab.tsx), Runtime (max_turns, model.*, compaction.*), Tools (editable built-in
  * tools table + the MCP Server form, mcp-servers-section.tsx), Skills (skills-tab.tsx),
  * Vault (vault-tab.tsx), Schedule (schedules-tab.tsx).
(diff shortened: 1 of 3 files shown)


hiyouga:
Follow-up (user feedback on the card layout: boxes read poorly, numbers not prominent enough, no clear distinction): reworked in 4a8ac4b.

- **Cards → ruled sections**: the Agent State and Kernel blocks now use the skills import modal's section family (`border-t border-gray-200 pt-4 dark:border-gray-800`, no rounded boxes) — the kernel block converted too so no lone card remains. The name/description form keeps its form shape, separated from the sections by the first rule.
- **Prominent values, clear label/value distinction**: labels stay `text-xs` gray; values render dark — State version `v{n}` as mono `text-sm font-semibold`, kernel generation dates mono `font-semibold` with the connector words (current/latest/up-to-date) kept light. Composite `kernelVersions()` i18n line split into `kernelCurrent`/`kernelLatest` inline labels (zh/en) so the dates can carry their own weight.
- **Layout per section**: title row (title left, actions right — State: export/import; Kernel: update/restore), info rows below (version; path + copy), kernel merge report still inside its section. All behavior unchanged (copy, confirm modals, owner gating, keepStale report).

Verification rerun: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (landing 39, core 792, server 594, cli 235, web 650, desktop 48 — all pass), `pnpm format` + `pnpm format:check` clean.

PR #262, opened 2026-08-11
feat(core): web search tip in default system prompt

hiyouga:
Stacked on #260 (which stacks on #257); review only this PR's commits.

## What

Adds one guidance bullet to the default system prompt's **# Tool use** section, right after the browse-with-shell bullet:

```
- For web searches, `curl -sG 'https://html.duckduckgo.com/html/' -A 'Mozilla/5.0' --data-urlencode 'q=your query'` is the most reliable first stop, and `curl -sGL 'https://news.google.com/rss/search' --data-urlencode 'q=your query'` the fallback (best for news) — together they cover most research.
```

Guidance, not a mandate. Both endpoints were live-tested before wording the tip:

- **DuckDuckGo HTML** — HTTP 200, ~34 KB HTML for a real query, 10 `result__a` links with titles + redirect URLs. Works exactly as written.
- **Google News RSS search** — the bare URL 302-redirects to itself with locale params (`hl/gl/ceid`) and an empty body; with `-L` it returns HTTP 200 and a full RSS feed (102 `<item>`s for the test query). The tip therefore spells `-sGL`, and `--data-urlencode` covers the q-encoding requirement.

## Kernel bump flow (first real run of #260's mechanism)

- The pinned-hash guard went red exactly as designed (only `system_prompt` drifted) and — per the **same-day revision rule** — the `"2026-08-11"` `KERNEL_HASH_HISTORY` entry was revised in place with the new `system_prompt` hash; `KERNEL_VERSION` is unchanged. No later generation exists, so the entry was not yet frozen, and the superseded hash disappears legitimately: #260 is unmerged, nothing in the wild was ever stamped with it.
- The pre-#257 **reconstruction proof** was anchored (`it.runIf`) on a recipe that read the *current* template, which this change invalidates — handled per its documented design, upgraded rather than retired: the **complete** pre-toggles default config is now a frozen test fixture (`core/test/fixtures/pre-toggles-default-config.json`), generated from the still-unchanged toggles-generation defaults and verified leaf-by-leaf against the pinned `"2026-08-10"` entry before freezing. The proof test asserts snapshot ↔ pinned entry — two independent frozen artifacts, so it stays non-tautological and **permanently valid** however the current defaults evolve (the whole config is frozen, not just the template, so future non-template default changes cannot re-break it).
- The core merge tests seed their "old config" scenarios from the same frozen snapshot, so "an untouched pre-toggles config advances wholesale" genuinely proves the old template (hash-matching `"2026-08-10"`) advances to the **new** template.
- The server endpoint test previously aged its config via the live recipe; it now ages by de-stamping and dropping the vault/skills/schedules sections (template left at the current default), asserting section materialization + kept customization + flags — the old-template-advance path stays covered in core against the frozen snapshot.

No new `{{…}}` placeholders, so the web placeholder-parity test is untouched; docs need no change (the tip lives in the template itself).

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (8 packages)
- `pnpm test` — pass (all suites; kernel guard green against the revised entry, snapshot proof green, server endpoint tests reshaped and green)

### DIFF
--- a/packages/core/src/state/default-config.ts
+++ b/packages/core/src/state/default-config.ts
@@ -394,6 +394,7 @@ Communicate with the user precisely and concisely, yet with warmth, and always r
 # Tool use
 - Prefer solving problems with your tools: inspect the real files and environment and run real commands instead of answering from memory or guessing.
 - For anything on the internet, browse with your shell tool: prefer Playwright when it is installed — it handles dynamic sites — otherwise \`curl\` for pages and APIs.
+- For web searches, \`curl -sG 'https://html.duckduckgo.com/html/' -A 'Mozilla/5.0' --data-urlencode 'q=your query'\` is the most reliable first stop, and \`curl -sGL 'https://news.google.com/rss/search' --data-urlencode 'q=your query'\` the fallback (best for news) — together they cover most research.
 
 # System markers
 Some messages carry system-synthesized \`[tag]...[/tag]\` blocks — not user text to answer:
--- a/packages/core/src/state/default-config.ts
+++ b/packages/core/src/state/default-config.ts
@@ -394,7 +394,7 @@ Communicate with the user precisely and concisely, yet with warmth, and always r
 # Tool use
 - Prefer solving problems with your tools: inspect the real files and environment and run real commands instead of answering from memory or guessing.
 - For anything on the internet, browse with your shell tool: prefer Playwright when it is installed — it handles dynamic sites — otherwise \`curl\` for pages and APIs.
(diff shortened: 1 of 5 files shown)


hiyouga:
Follow-up on user feedback: DuckDuckGo can answer with a CAPTCHA/bot-check page. The search bullet now ends with the switch instruction — *"DuckDuckGo sometimes answers with a CAPTCHA/bot-check page instead of results; when it does, run the search through a Playwright-driven browser instead."* — still one bullet, guidance tone.

Kernel flow: the `"2026-08-11"` history entry was revised in place a second time under the same-day rule (only the `system_prompt` hash moved; `KERNEL_VERSION` unchanged; the frozen pre-toggles fixture pins the older generation and was untouched, as designed).

Verification rerun: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (all package suites), `pnpm format` + `pnpm format:check` — all pass.

hiyouga:
Closed at the user's request — the search-tip approach needs more thought. Branch kept in case this gets revisited.

PR #19, opened 2026-07-22
feat(skills): vLLM and Ollama deployment plus LlamaFactory fine-tuning skills

hiyouga:
## Summary

Adds three built-in skills to the **AI App Development** group, covering the local-model lifecycle end to end — serve, register, fine-tune, serve again:

- **`vllm`** — deploy and serve LLMs with vLLM behind an OpenAI-compatible endpoint (`vllm serve`, VRAM/context-length/quantization flags), including the tool-calling flags agent harnesses require (`--enable-auto-tool-choice --tool-call-parser <parser>`; without them, requests that set tool_choice fail with vLLM's 400).
- **`ollama`** — deploy and run local models with Ollama (install, pull/run/list/ps/stop, the `http://localhost:11434/v1` OpenAI-compatible endpoint), raising the context length via `OLLAMA_CONTEXT_LENGTH` or a Modelfile `num_ctx`, and reusing — never killing — an existing instance.
- **`llamafactory`** — fine-tune with LLaMA-Factory: `data/dataset_info.json` registration (alpaca/sharegpt), `llamafactory-cli train` with YAML configs, `webui`, LoRA merge/export, `chat`/`api`, then serving the export via the `vllm`/`ollama` skills.

All three close the loop by registering the endpoint with `penguin config model add --provider custom --client-type openai ...` (with `--root ./penguin_data` for app-local config, consistent with the `penguin-cli` skill), so PenguinHarness agents can build, evaluate and tune AI apps on the served or fine-tuned model.

## Registration

- `SKILL_GROUPS` in `packages/skills/src/index.ts`: `ai-app-development` now lists `penguin-sdk`, `penguin-cli`, `agenthub-models`, `vllm`, `ollama`, `llamafactory`.
- Matching rows added to the docs tables `packages/docs/content/skills.en.md` / `skills.zh.md` and the `packages/skills/README.md` group table.
- The hardcoded group expectations in `packages/skills/test/skills.test.ts` and `packages/server/test/skills.test.ts` were extended.
- Each skill ships the standard frontmatter (bilingual short descriptions), a `## Before you start` gate, and a 24×24 line-art `icon.svg`.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` and `pnpm test` all pass:

| Package | Tests |
| --- | --- |
| skills | 16 passed |
| docs (incl. skills-sync) | 11 passed |
| core | 372 passed, 2 skipped |
| server | 285 passed |
| cli | 119 passed |
[...]

Issue #16, opened 2026-07-21
Help Add Multimodal Input to This Harness


patrick-tssn:
Could you help add multimodal input to this harness? We’d like support for images, video, audio, and video with audio. 🎥🖼️🔊

hiyouga:
Images are already supported for most models

PR #271, opened 2026-08-12
fix(server): trace index gate misses compaction-rotated shards in non-newest date dirs

Ufodar:
## Problem

After context compaction, the UI stays stuck on the "compaction complete" banner — the full post-compaction conversation is invisible. This happens whenever the session's original date directory is older than the newest date dir on disk (e.g. a long-running session that started yesterday and compacts today).

## Root cause

`TraceWriter.rotate()` (compaction) creates a new shard (`_002`) in the session's **original** date directory — `dateDir` is set at construction and never changes. But the `TraceIndexService` mtime gate only stat-checks the **newest** date directory:

```js
const newest = [...cached.dates.keys()].sort().at(-1);
const m = await statMtime(path.join(dir, newest));
if (m !== null && m === cached.dates.get(newest)) return; // skip
```

The comment assumed "in practice always the newest one (the Writer names dirs by current local date)." Compaction breaks this assumption: `rotate()` writes to the original dir, whose mtime the gate never checks.

`locateAll()` only force-reconciles when the index returns **zero** rows for a session. Since `_001` is already indexed (`rows.length = 1`), the force path never triggers → `GET /messages` returns only up to `compaction_end`.

## Fix

Stat **every known date directory** instead of only the newest, breaking out on the first whose mtime differs:

```js
let changed: string | null = null;
for (const date of cached.dates.keys()) {
  this.counters.gateStats += 1;
  const m = await statMtime(path.join(dir, date));
  if (m === null || m !== cached.dates.get(date)) {
    changed = date;
    break;
  }
}
if (changed === null) return;
```

Still readdir-free (one root stat plus up to N date-dir stats, N = date dir count, typically 1–5). The remaining blind spot is a backdated write (an external process that changes a file and resets its dir's mtime to the cached value) — that is not covered by the gate or by `locateAll()`'s whole-session force-retry, and remains missing until an unrelated force, explicit forced reconciliation, or restart.

## Verification

Run at the worktree root with Node 24:

- `pnpm install --frozen-lockfile` — pass
- `pnpm -r build` — pass
- `pnpm test` — pass (core 666, server 558, zero failures); one pre-existing flaky timing test in `core/test/exec-session.test.ts` passed on retry
- `pnpm format:check` — pass

Regression test details:

- New test in `packages/server/test/trace-index.test.ts`: session starts in `2026-07-05/`, a second session creates `2026-07-06/`, then `_002` is added to the old dir — the gate catches the mtime change and registers the shard. Confirmed the test fails on `main` (1 failed / 4 passed) and passes with the fix (5/5).
- All 4 existing tests in the same file pass unchanged.


changelog/unreleased/2026-08-11-trace-index-compaction-rotate.md
```diff
@@ -0,0 +1,40 @@
...
+diff).
+
+## The fix
+
+The gate now stats **every known date directory** instead of only the newest, breaking
+out on the first whose mtime differs from the cached value. This is still readdir-free
+(N stat calls where N is the date-dir count, typically 1–5) and preserves all existing
+behavior: the unchanged-tree fast path still returns early, and the backdated-mtime blind
+spot (an external write that resets a dir's mtime to the cached value) is still covered
+by `locateAll()`'s force-reconcile fallback.
```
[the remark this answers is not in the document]
Ufodar: Good catch. Updated the changelog and PR body to state the limitation honestly: `locateAll()` forces only on a whole-session miss (zero indexed rows), so a backdated `_002` for a session whose `_001` is indexed is not covered. The file-level JSDoc on the forced-reconciliation path now documents this too (commit 611795d).

packages/server/src/services/trace-index.ts
```diff
@@ -12,14 +12,16 @@
...
- *     dir's mtime). Only on a change does it readdir — and only the date dirs whose
- *     mtime moved — registering new shards and classifying each new Session ONCE
- *     (bounded head-read of the earliest shard for origin/workspace/title). Restart
- *     forgets the in-memory gate, so the first request per Agent after an upgrade or
- *     restart runs one full diff (this is also how the index first populates: no
- *     migration step needed).
+ *     traces root plus every known date dir (N stat calls, no readdir; a new
+ *     date dir bumps the root's mtime, a new file in any existing date dir bumps
+ *     that dir's mtime — including compaction's rotate() which writes to the
+ *     session's original date dir, not necessarily the newest). Only on a change
```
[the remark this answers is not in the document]
Ufodar: Fixed in 611795d. The `reconcileAgent` JSDoc now says "one root stat plus up to N date-dir stats" instead of "two stat calls". The forced-reconciliation description now explains that the remaining blind spot is a backdated write (mtime reset to the cached value), and that `locateAll()` does not cover it for sessions with existing indexed rows.

packages/server/src/services/trace-index.ts
```diff
@@ -166,16 +172,23 @@ export class TraceIndexService {
...
+      // only the newest; this is still readdir-free (N stat calls, N = date dir count).
       // Cached sentinels (-1: the dir was fresh when last seen) never match, forcing a
       // re-diff until the tree has been quiet (see mtime-gate's FRESH_MS).
-      const newest = [...cached.dates.keys()].sort().at(-1);
-      if (newest === undefined) return;
-      this.counters.gateStats += 1;
-      const m = await statMtime(path.join(dir, newest));
-      if (m !== null && m === cached.dates.get(newest)) return;
+      let changed: string | null = null;
+      for (const date of cached.dates.keys()) {
```
hiyouga: **1. The gate's hot path goes from a fixed 2 stats to 1+N sequential `await statMtime` calls, where N is the Agent's lifetime date-dir count (one dir per calendar day, never pruned) — and every caller multiplies it.**

`Writer.dateDir` is fixed from `formatLocalDate` at construction (core/src/trace/writer.ts:127), so there is one date dir per calendar day the Agent ran, and grep finds no trace retention/pruning anywhere in packages/server or packages/core. The `break` never fires on the hot path — 'nothing changed' means all N stats run, serially. Measured at N=365: 366 stats / ~24ms per reconcile vs 2 stats / ~0.1ms on main. Callers multiply it: `GET /projects/:id/agents` runs sessionStats -> reconcileAgent per Agent (routes/agents.ts:25-31, session-service.ts:261), so 20 Agents x 365 dirs = ~7,320 stats per sidebar poll (was ~40); `GET /messages` runs locateAll -> reconcileAgent once per distinct subagent child (trace-service.ts:265), so 30 children = ~11,000 stats; session detail pays it twice (session-service.ts:386 and :413). The changelog's 'typically 1-5' holds only for a brand-new install. Cheaper: `Promise.all` the stats, and/or bound the watch set to the newest dir plus the date dirs of live sessions.

packages/server/src/services/trace-index.ts
```diff
@@ -166,16 +172,23 @@ export class TraceIndexService {
...
+        this.counters.gateStats += 1;
+        const m = await statMtime(path.join(dir, date));
+        if (m === null || m !== cached.dates.get(date)) {
```
hiyouga: **2. A single date dir whose mtime is in the future is cached as the -1 sentinel forever and now permanently defeats the gate for the whole Agent — a new regression, since previously only the newest dir was gated.**

`cacheable()` returns -1 whenever `Date.now() - mtimeMs <= FRESH_MS`, which includes any NEGATIVE difference — a future mtime from a restored backup, `rsync -t` from a host with a faster clock, NFS/container clock skew, or a backward NTP/VM clock step. Such a dir can never match a real stat, so the new loop breaks on it and falls into the FULL diff on every single reconcile, forever: listDirs + N stats + readdir of the dir + an fs.stat per file + head-reads. Pre-PR an old skewed dir was never stat'd, so the fast path still returned early. Confirmed empirically: quiet tree, two date dirs, the older one given a future mtime, three no-op reconciles -> PR head does 3 full dir scans, main does 0. There is no self-healing path short of restart.

packages/server/src/services/trace-index.ts
```diff
@@ -166,16 +172,23 @@ export class TraceIndexService {
...
+          changed = date;
+          break;
+        }
+      }
+      if (changed === null) return;
```
hiyouga: **10. The gate discards every mtime it just read and the diff loop re-stats all of them (up to 2N stats per changed pass), and it walks `cached.dates` in raw readdir order so the common newest-dir change is found at a random position instead of first.**

After the `break` falls through, line 206 stats every date dir again purely to populate `next.dates`, so a changed pass costs up to N (gate) + N (diff) where main cost 1 + N. The changed path is the COMMON path while an Agent is working, because an actively-written dir is cached as the -1 sentinel and never matches. On top of that, `cached.dates` is built from `listDirs` (line 204) and never sorted, so its insertion order is ext4/APFS hash order; main's `[...keys].sort().at(-1)` found the active dir in exactly 1 stat, while the new loop pays ~N/2 before breaking. Two cheap fixes with no behaviour change: seed `next.dates` from the gate's already-read values, and iterate `[...cached.dates.keys()].sort().reverse()`.

changelog/unreleased/2026-08-11-trace-index-compaction-rotate.md
```diff
@@ -0,0 +1,42 @@
+# Trace index: mtime gate catches compaction-rotated shards in non-newest date dirs
```
hiyouga, line 1: **15. Changelog files are bundled into a feature PR against an explicit repo rule, and the entry carries a `## Verification` section — unique in the whole changelog corpus — containing a private session id and internal message counts that freeze into the released folder.**

CLAUDE.md (symlink -> AGENTS.md), section 'Changelog 逻辑': 「changelog 用单独的汇总 PR 提交，不塞进各功能 PR」. This PR ships the new 42-line entry plus the changelog/unreleased/README.md index line alongside the code. Recent history confirms the rule is live practice, not dead text — changelog-only PRs #258, #250, #244, #243, #238, #228 — so the batch's aggregate changelog PR will collide on the single-line unreleased/README.md, and this entry is the only one carrying no PR reference. Separately, `grep -rl '## Verification' changelog/` matches only this file: lines 34-42 put 'All four existing trace-index tests continue to pass', the session id `f0e639aa`, and raw counts (1937 -> 2265 messages) into user-facing release documentation, and changelog/README.md states 'Released folders are frozen', so once unreleased/ is renamed at release this can no longer be cleaned up. changelog/README.md also reserves `##` sectioning for entries that 'cover more than one thing'; this single-topic entry opens three.

hiyouga:
## Code review — findings outside the diff hunks

These land on lines this PR does not touch (unchanged lines of the function it changes, or consumers of it), so GitHub will not accept them as inline comments. Ranked with the inline ones, most severe first.

### 3. `packages/server/src/services/trace-service.ts:193`

**`deleteSessionTraces` locates shards through the non-forced `locateAll`, so any shard on disk but missing from the index survives the delete and the next reconcile resurrects the Session.**

`locateAll` force-retries only when `rows.length === 0`; deleting a Session that has `_001` indexed never forces. Under exactly the class of miss this PR is about (a rotate-created `_002` not yet indexed, or the surviving backdated-write / transient-failure holes), `files` omits `_002`, the `fs.rm` loop leaves it on disk, and `removeSession` then deletes the rows AND drops the gate entry — which GUARANTEES the next reconcile full-diffs, re-upserts `_002`, sees `getSession(sid)` null so `metaRead !== true`, re-classifies it, and the deleted Session reappears in `agentTraces`/`discoverTraces` as a CLI-origin Session while its DB row is already gone. A destructive operation should not trust the derived cache: force the locate here, or sweep `<sid>_*.jsonl` across the date dirs.

### 4. `packages/server/src/services/trace-index.ts:207`

**`if (m === null) continue;` evicts a date dir from `next.dates` on a transient stat failure, which both purges its rows and silently removes it from the gate's watch set permanently — reintroducing the exact blind spot this PR closes.**

One EACCES during a permission change, EMFILE under fd pressure, EIO/ESTALE on a network mount, or a Windows sharing violation makes `statMtime` return null for a date dir that still exists. Two things compound: the 'whole date dirs deleted from disk' sweep (lines 237-241) deletes every index row for that date, and the date disappears from `cached.dates` — the very set the new gate loop iterates. The traces root's mtime never moves (the dir was not removed), so no later pass ever stats it again. Every future rotate or external write into that dir is invisible until an unrelated force or a restart. A stat failure should mark the dir dirty (store -1 and skip the row purge), not forget it.

### 5. `packages/server/src/services/trace-index.ts:204`

**`listDirs`/`listFiles` swallow every readdir error and return `[]`, so one transient failure mass-deletes index rows and then caches the tree as clean, making the loss sticky.**

Root level: `statMtime(dir)` succeeds so the `rootMtime === null` escape at 168 is not taken, then `listDirs` hits EMFILE/EACCES/EIO and returns []. `dateDirs` is empty, so lines 237-241 delete EVERY row for the Agent, and line 247 commits `{root: cacheable(rootMtime), dates: new Map()}` — the real, unchanged root mtime. On every later pass `cached.root === rootMtime` holds and the new gate loop iterates an EMPTY map, so `changed` stays null and it returns at 191. Date-dir level is the same shape one level down: line 208 records the dir's mtime BEFORE the readdir at 215, so a failure there deletes every row of that date and still caches it clean. Neither `agentTraces` (trace-service.ts:1020) nor `sessionStats` force-retries, so the Agent shows 0 sessions and a flat activity sparkline until the root mtime moves or the process restarts.

### 6. `packages/server/src/services/trace-index.ts:225`

**The per-file `fs.stat` catch runs `continue` BEFORE `known.delete(fileKey)`, so a transient stat error deletes the index row for a file that is still on disk.**

`listFiles` returns `<sid>_002.jsonl` and `fs.stat` on it fails with something other than ENOENT (EMFILE, EACCES, EIO). The `continue` skips `known.delete(fileKey)`, so the key survives into the 'rows whose files vanished' sweep at 232-234 and the row is deleted. The comment's promise that 'a later pass settles it' is false: `next.dates` already holds the dir's real mtime from line 208, so the gate declares the dir clean. Because the session still has `_001` indexed, `locateAll` never force-retries and `GET /messages` returns a silently truncated transcript — the exact symptom this PR set out to fix, reached through a different door. The catch needs to distinguish ENOENT (genuine vanish) from every other errno.

### 7. `packages/server/src/services/trace-index.ts:168`

**`rootMtime === null` conflates 'this Agent has no traces dir' with a transient stat failure, so one errno wipes both the file rows and all per-Session facts.**

`statMtime` catches every error, so an EMFILE/EACCES/EIO on `<agent>/traces` reaches `deleteByAgent`, dropping every `trace_files` row and every `trace_sessions` row — title, workspace, source, provider, model ref, and the cached `page_stats` derivations. The next pass then re-head-reads EVERY session (the opposite of the module's 'classify each new Session ONCE' contract) and re-scans every message window. In the window before that, the shared `SessionSources` registry has no entry for those sessions, so a concurrently-served listing classifies subagent/schedule sessions as `source: null` and files them into the wrong bucket.

### 8. `packages/server/src/services/trace-index.ts:229`

**`metaRead !== true` re-queues a head-read on every scanning pass for any Session whose earliest shard lacks `session_meta` or whose head failed to read, and such a Session can never be adopted into the list.**

`factsFromRecords` sets `metaRead = meta !== undefined`, and `classifySession`'s catch stores `records = []` -> `metaRead: false`. Line 229 therefore re-adds the session to `newSessions` every time its date dir is re-readdir'd, paying a `readTraceHead` each time — the PR's own new test fixtures (`writeTraceFile(..., [userText('first turn')])`, no `sessionMeta`) are permanently in this state. Meanwhile `adoptTraceSession` (session-service.ts:432) returns null on `!facts?.metaRead`, so the Session never appears in the Session list at all, and its comment 'retried by a later reconcile' is false once the dir is cached clean. Combined with the -1 sentinel on an actively written dir, this becomes a head-read of every unmeta'd session on every request.

### 9. `packages/server/src/services/trace-index.ts:145`

**Forced reconciles do not coalesce — K concurrent forced callers produce K SEQUENTIAL full diffs — and a rejected in-flight pass makes the forced retry never run at all.**

`existing.then(() => this.reconcileAgent(projectId, agentId, opts))`: when `existing` settles, each of the K queued callbacks re-enters `reconcileAgent`; the first installs a new in-flight promise and the rest chain behind it, so K forced callers run K full diffs back to back — and this PR makes each of those passes substantially more expensive. Several browser tabs opening the same Session whose rows are missing is enough to trigger it. Separately, `.then` (not `.catch`) means one SQLITE_BUSY or EIO in an unrelated concurrent pass propagates and the forced caller's scan never happens, breaking the documented guarantee that a stale index 'degrades to one extra scan, never to a false 404'.

### 11. `packages/server/src/services/trace-index.ts:298`

**The O(1) write-time seam that this module's own doc names as the gap was left unused; the fix pays unbounded per-request polling for an event the server already observes in-process.**

Lines 9-13 state the model ('server paths that create/delete shards with known identities register synchronously') and name the gap verbatim ('without telling the server which shard rotated'). The precedent exists in this file — `registerImportedFile` drops the gate with `this.seen.delete(key)` at line 298 — and the sibling cache already exposes it as a named primitive, `ScheduleFileCache.invalidate` (runtime/schedule-store.ts:93, 'Write-time invalidation: the next list re-scans regardless of mtimes'). The server also already observes the triggering event: `stream-error-watcher.ts:302` handles `compaction_end` with a ctx carrying projectId/agentId/sessionId. A row-preserving `TraceIndexService.invalidate()` called from there fixes the reported bug exactly, at zero steady-state cost. The gate widening is still needed for external/CLI writers, so keep it — but shipping only the polling half means the deterministic path stays missing AND is still mtime-dependent (the changelog admits a surviving hole).

### 12. `packages/server/src/services/trace-service.ts:171`

**The consumers that produce the reported symptom are still unprotected: `locateAll` forces only on a whole-session miss, so a session with `_001` indexed but `_002` missing silently returns partial data with a 200 OK.**

Under the surviving blind spots (backdated write per the PR's own changelog, plus the transient-failure holes above), `rows.length !== 0` so no force happens, and every consumer trusts the truncated list: `readMessagesPage`'s tail window takes `endPos = files.length - 1` (line 432), `readMessages` concatenates whatever the index holds, and `latestTracePath` takes `files.at(-1)` (session-service.ts:402) — which is the source shard handed to a `/model` switch, so the model gets the PRE-compaction history. None of these can tell 'this is the last shard' from 'the index lost the last shard'. Cheap defence-in-depth: force when the session's own date dir is not the newest, or when a higher index for that session exists on disk.

### 13. `packages/server/src/services/trace-service.ts:163`

**Documentation left stale by the change: `locateAll`'s JSDoc still calls 'an external write into an old date dir' a gate blind spot — the exact case this PR just closed — and the changelog under-reports the cost trade while silently invalidating a frozen released entry.**

The JSDoc now contradicts the new module header in trace-index.ts:25-32 (which says the remaining blind spot is a backdated write), and the PR carries a commit titled 'address review: fix stale comments' that missed it. A maintainer reading it either adds redundant force-reconciles or trusts the force-retry to cover the backdated case, which it does not. Separately, changelog/unreleased/2026-08-11-trace-index-compaction-rotate.md:27 claims 'still readdir-free (N stat calls where N is the date-dir count, typically 1-5)' — N is the Agent's lifetime calendar-day count with no pruning — and never states that the hot path went from constant to linear; and changelog/0.2.1/2026-08-04-traces-page-scaling.md:7 still reads 'The hot path is two stats — an mtime gate on the traces root and the newest date dir', which is now false and, per CLAUDE.md, frozen and uneditable. The new entry should supersede it explicitly.

### 14. `packages/server/test/trace-index.test.ts:86`

**The stat-cost contract this PR changes is asserted by no test: `counters.gateStats` exists precisely to pin it and is used in zero assertions, and the new regression test omits even the `dirScans` assertion its sibling carries.**

trace-index.ts:115-120 documents the counters as 'Test observability (asserting the hot path stays readdir/head-read free)', but a repo-wide grep shows only `dirScans` and `headReads` are ever asserted (lines 89, 90, 99, 171 here). The steady-state block at 86-90 therefore passes identically at 2 stats and at 366, which is how the 2 -> 1+N change landed silently; the sibling suite does pin its budget (test/scheduler.test.ts:174-183). The new test at line 103 also asserts only the row count, not that the widened gate stayed readdir-free. One line — `expect(h.traceIndex.counters.gateStats).toBe(before + 1 + dateDirCount)` after two no-op reconciles — would make the trade explicit in review and catch the next regression.

packages/server/src/services/trace-index.ts
```diff
@@ -12,18 +12,24 @@
...
+ *     that dir's mtime — including compaction's rotate() which writes to the
+ *     session's original date dir, not necessarily the newest). Only on a change
+ *     does it readdir — and only the date dirs whose mtime moved — registering
+ *     new shards and classifying each new Session ONCE (bounded head-read of the
+ *     earliest shard for origin/workspace/title). Restart forgets the in-memory
+ *     gate, so the first request per Agent after an upgrade or restart runs one
+ *     full diff (this is also how the index first populates: no migration step
+ *     needed).
+ *   - Forced reconciliation (`force`): ignores the gate and diffs every date dir.
+ *     The remaining gate blind spot is a backdated write — an external process that
```
hiyouga: **16. This rewritten sentence is false: an in-place append moves no directory mtime at all, so the remaining blind spot is much larger than "a backdated write" — and it is the normal resume path, not an exotic one.**

The old doc gave an open-ended example ("e.g. an external write into an OLD date dir"); the rewrite turns it into a closed claim — "The remaining gate blind spot is a backdated write". Verified on Linux/ext4: creating a file bumps the containing directory's mtime, **appending to an existing file does not** (`dir mtime before=…423.201 after-append=…423.201 after-create=…424.305`). A gate built entirely out of directory stats therefore cannot see an append, and no backdating is required.

That is exactly what a resumed Session does. `packages/core/src/agent.ts:480-485` pins both the directory and the shard:

```ts
const trace = new Writer({
  tracesDir: dir,
  sessionId,
  dateDir: located.dateDir,
  startIndex: located.index,
});
```

so a resumed run appends to an **existing** shard in an **existing** date dir — invisible to the widened gate.

Consequence beyond the doc: `trace_files.size_bytes` stays stale because `upsertFile` never re-runs, and `page_stats` invalidation is driven entirely by that size comparison (`page_stats = CASE WHEN excluded.size_bytes = trace_files.size_bytes THEN trace_files.page_stats ELSE NULL END`). `prefixStates` (`services/trace-service.ts:373-380`) then returns the cached prefix record for a shard that has since grown, with no re-stat at read time. `listTraceFiles` re-stats but never calls `updateFileSize`, so it does not invalidate either.

This is the house pattern's known gap, not a new discovery: `agentTracesPage`'s own comment says "an actively-appended shard grows without moving any directory mtime", and the sibling gate in `runtime/schedule-store.ts` added a per-file mtime/size check precisely to close it. Please restore the open-ended wording, or state the append case explicitly alongside the backdated one.

changelog/unreleased/2026-08-11-trace-index-compaction-rotate.md
```diff
@@ -0,0 +1,42 @@
...
+newest date directory** after confirming the root mtime was unchanged. The comment assumed
+"in practice always the newest one (the Writer names dirs by current local date)."
+Compaction's `rotate()` breaks this assumption: the new shard lands in an older dir whose
+mtime the gate never checks, so the index never registers it.
+
+`locateAll()` only force-reconciles when the index returns **zero** rows for a session;
+since `_001` is already indexed (`rows.length = 1`), the force path never triggers. The
+result: `GET /messages` returns messages only up to `compaction_end`, and the UI stays
+stuck on the "compaction complete" banner — the full post-compaction conversation is
+invisible until the process restarts (which forgets the in-memory gate and runs a full
```
hiyouga: **17. "invisible until the process restarts" over-states the pre-fix severity: `rotate()` did move its own date dir's mtime, so any later gate trip from any cause rescanned that dir and registered the shard.**

The old gate only failed to *look* at that dir. Once anything else tripped it, the diff loop's skip test is `!force && cached?.dates.get(date) === m`, which fails for the rotate-touched dir — so it was rescanned and `_002` registered. Triggers include: the next day's first Session creating a new date dir (root mtime moves), any new shard in the newest dir, an import or Session delete (both `seen.delete(...)`), and any consumer's forced pass.

So the pre-fix window was "until the next unrelated change anywhere in this Agent's tree" — bounded in practice by midnight — not "until the process restarts". Since this sentence is the severity half of the trade the PR asks readers to accept (a constant-cost hot path becoming 1+N sequential stats per request per Agent), over-stating it skews the decision. The same claim appears in the PR description.

hiyouga:
## Code review — addendum (final gap-sweep pass)

Two more findings from the last pass; both land on lines this PR does not touch, so they cannot be inline. Numbering continues from the earlier comments.

### 18. `packages/server/src/services/trace-index.ts:237` — a shard that moves between date dirs is re-registered and then deleted by the same pass, and when both dirs survive, raw `readdir` order alone decides whether the row lives

`upsertFile`'s conflict key is `(project_id, agent_id, session_id, file_index)` with `DO UPDATE SET date = excluded.date` — no `date` term — and `deleteFile` likewise matches on `project_id, agent_id, session_id, file_index` only (`db/repos/trace-index.ts:131-138`). `knownByDate` is snapshotted at pass start, holding the OLD date.

Trigger: any external reorganization that moves shards across date dirs — a hand-fixed timezone dir, `mv traces/2026-07-05 traces/2026-07-04`, or restoring an exported Trace under a different date.

- **Old dir gone:** the file is found under the new date and `upsertFile` updates the row's `date` in place; then the trailing "whole date dirs deleted from disk" loop (line 237) sees the OLD date missing from `next.dates` and calls `deleteFile(project, agent, session, index)` — deleting the row it just wrote.
- **Both dirs survive** (one shard moved B → A): the per-dir sweep at line 232 decides it, so `readdir` order is the whole story. A-then-B registers it and then deletes it; B-then-A deletes it and then re-registers it. The comment on that sweep literally says "external delete / **rename**".

Afterwards `next.dates` caches the surviving dir with a real (non-sentinel) mtime, so the gate reports the tree clean, and `locateAll` force-retries only on a **zero-row** session — a session that still has its other shards indexed then serves a permanently truncated transcript. That is precisely the failure class this PR set out to eliminate, reached through a different door.

### 19. `packages/server/test/trace-index.test.ts:26` — the `backdate` helper's comment still describes the pre-PR two-mtime gate

> "which is precisely the 'old dir changed without moving any gated mtime' scenario"

After this change every date dir *is* a gated mtime. The PR updated the in-body comments of the test that uses this helper but not the helper it depends on. The second `backdate(...)` call in the blind-spot test is load-bearing for a reason the comment no longer states: the scenario is now "a changed dir whose mtime was reset to exactly the cached value", not "a dir nobody watches". A maintainer reading the helper will reason with the old gate model and mis-judge which of the two tests still proves a blind spot.

hiyouga:
## Review fixes applied (5e24734)

Latest `main` is merged in and the findings that belong to this change are fixed on the branch. Full chain green under Node 24: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (server **604**, +1 new), `pnpm format:check`.

**Fixed here**

- **#16 — the rewritten blind-spot claim was false.** "The remaining gate blind spot is a backdated write" closed an open-ended caveat into a wrong one: an in-place append moves no directory mtime at all, and that is the *normal resume path* (core pins `dateDir`/`startIndex`, so a resumed Session grows an existing shard and its `size_bytes` goes stale). Both blind spots are now named in the module header, in the changelog, and in `locateAll`'s JSDoc.
- **#13 — stale/again-wrong docs.** `locateAll`'s JSDoc still called "an external write into an old date dir" a blind spot — the exact case this PR closes. Rewritten, and it now states that the force-retry only covers a *whole-session* miss. The changelog's "typically 1–5" is replaced by the honest statement (N is the Agent's lifetime calendar-day count, unpruned), the constant → linear trade is spelled out, and the entry explicitly supersedes the frozen `0.2.1` claim that "the hot path is two stats".
- **#10 / #1 — hot-path ordering.** The gate walked `cached.dates` in readdir order, so the common case (a new shard in today's dir) paid up to N *serial* stats before detecting it. It now stats the newest dir first and short-circuits there, then the remaining dirs **concurrently**. Quiet trees still cost 1+N stats, but they are no longer serial and the write path is back to 2.
- **#14 — the budget is now pinned.** `gateStats` was asserted nowhere, which is how 2 → 1+N landed unnoticed. The steady-state test asserts it, and the rotate test asserts the newest-dir short-circuit. Mutation-checked: reversing the iteration order to oldest-first fails the new assertion (8 stats vs 7).
- **#17 — severity over-stated.** "invisible until the process restarts" → the accurate bounded claim (any later gate trip rescans the dir; in practice midnight).
- **#15 (content half) — the `## Verification` section is removed.** It carried a private session id and raw message counts into a folder that freezes at release, and it was the only such section in the changelog corpus.
- **#19 — the `backdate` helper's comment** now describes the post-change gate.

**Deliberately not fixed here**

- **#2 (future-mtime dir permanently defeats the gate)** is a real regression from widening the watch set, but the mechanism is `cacheable()`'s treatment of a negative age in the *shared* `mtime-gate`, so the fix affects every gate in the server. It wants its own change and a decision on clamping vs. skew tolerance, not a drive-by edit inside this PR.
- **#3–#9, #11, #12, #18** are pre-existing behaviour of `trace-index` that this PR neither introduces nor worsens — transient `stat`/`readdir` failures purging index rows and then caching the tree clean, forced reconciles not coalescing, `locateAll` not forcing on a *partial* session miss, and the date-less `deleteFile`/`upsertFile` pair that lets a shard moved between date dirs be re-registered and deleted in the same pass. Several are more serious than the bug this PR fixes; folding them in would turn a focused fix into a rewrite. They are worth their own issues.
- **#15 (process half)** — the repo rule says changelog entries ride a separate aggregate PR. Recent merged contributions have bundled them anyway, so the entry stays; flagging the divergence rather than acting on it.

PR #277, opened 2026-08-12
feat(web): collapsible agents→evaluation nav group with remembered state

hiyouga:
## Summary

Sidebar batch: collapsible Agents→Evaluation-Center nav group, session pinning + last-active times, and session-list header controls (search / list settings / mode-aware create), refined over several user-feedback rounds.

### Collapsible page-nav group

The run of page entries from **Agents (智能体) through Evaluation Center (评估中心)** — the entire fixed page nav — collapses as one group:

- **Toggle doubling as the nav/session seam**: nav-row-wide **h-4** chevron button under 评估中心 (12px chevron; UP = collapse, DOWN while collapsed = the way back, 200ms flip), soft resting band (`bg-gray-200/70` / `dark:bg-gray-800/70`, hover deepens), no ruled separator at the boundary. `aria-expanded` + 「折叠」/「展开」.
- **Slide animation**: `grid-template-rows` 0fr↔1fr tween (skills/models-page convention), rows mounted but `inert` while collapsed; persistence-restored mounts render instantly; reduced motion inherits the global styles.css override.
- **Remembered state**: `penguin.sidebarNavGroupCollapsed` (global, default expanded). Pure module `lib/nav-group-collapse.ts`. The pinned 「新建对话」 block sits outside the range and always stays visible; the collapsed rail is untouched.

### Session pinning (置顶) + last-active time

- Row actions (rename / archive / delete) plus 置顶/取消置顶 live in a per-row ellipsis menu. Trigger: three **filled** dots, no hover background (icon color deepens only), swapped in place of the right-aligned relative time on hover/focus/open. Menu per the reference design: `text-xs`, compact rows, w-32 body-portaled panel, each item led by a thin-line glyph (pin / pencil / archive / trash; delete keeps red).
- Pinned rows bubble to the top of their group's active list, show a dim pin glyph, persist per Project in `penguin.pinnedSessions.<projectId>` (frontend-side; no server pin field exists). Delete prunes; foreign stale ids stay inert.
- Row time from **`lastActiveAt`** (#280) via `formatRelativeShort` — zh 「刚刚/5 分钟前/…」, en `now/5m/…`, ≥7 days → month-day, invalid → hidden. A running row shows its run-start stamp (lastActiveAt is written at run start and again at run end), so the time recedes while a run continues — the pulsing status dot beside it is what says "active now"; CLI-adopted and subagent rows are not driven by this server, so theirs equals `createdAt`.

### Session-list header controls

- **搜索 — mac-style in-place expansion**: no extra row; the header's grid columns tween (horizontal 0fr/1fr trick, 200ms) — the 「会话列表」 label collapses/fades while a bordered field grows leftward over its place, magnifier morphing into the field's leading glyph; × / Escape collapse it back. Live title filter over **loaded** rows of every group/folder (drafts included): zero-match groups hide, matches force groups/folders open, caps and "More" rows bypassed/hidden while filtering, quiet 「无匹配会话」 on zero hits. Transient.
- **列表选项** (sliders): shared-Dropdown menu, two checkmark-radio sections — **分组方式** (workspace/agent; the old inline toggle relocated here; the Trace page keeps its own) and **排序方式** (手动排序 / 最近更新, global `penguin.sidebarSortMode`, default 最近更新 — which now genuinely sorts by **last activity**: `orderSessionRows` takes a `recencyOf` accessor reading `lastActiveAt`, ties broken by session id desc, ahead of the pin/manual stages, so it also decides where manual-mode newcomers land); rows share the overflow menus' compact type scale.
- **手动排序**: native HTML5 drag-reorder of each group's active rows with a thin accent drop line; per-Project order `penguin.sessionOrder.<projectId>` (tolerant, prune-on-delete), newcomers on top in recency order, switching back keeps the stored order. Pinned cluster stays first; drags never cross the pin boundary; no drag while searching; folders stay chronological.
- **新建** (rightmost): the mode's entity glyph with a bottom-right plus badge — **folder+** (workspace mode) / **robot+** (agent mode). Behavior follows the mode:
  - **agent mode → Agent**: opens the Agents page's existing create dialog via `{ create: true }` route state (consumed once, then stripped).
  - **workspace mode → Workspace**: opens the **same directory-browse menu the draft's workspace picker uses** (`WorkspaceSelect` `trigger` render prop; panel body-portaled, **w-64** — the app's standard menu width, internal list scroll, entry names truncate with tooltips). The picked directory registers per Project and **appears immediately as an empty workspace group** at the top of the list — zero Sessions needed; chats start there via the group header's "+".

### Manually-added Workspaces: rename & remove

- Registry entries are `{ path, alias? }` in `penguin.sidebarWorkspaces.<projectId>` (loads stay tolerant of the earlier string-only shape; junk dropped; normalized + deduped).
- Each **registry-backed** workspace group header carries an ellipsis overflow to the right of its "+" (an action sibling — never toggles the group's collapse; session-derived groups show none):
  - **重命名工作区**: sets a display alias (rename-chat Modal + Input idiom; blank reverts to the directory basename — blank save is valid). The alias relabels the group even once it is session-backed; the full path stays in the tooltip.
  - **删除工作区** ("Remove workspace" in en): asks for confirmation via the shared ConfirmModal (the 删除对话/删除草稿 idiom), with copy honest about the scope — only the sidebar workspace group is removed, the directory on disk and existing chats are untouched, and it can be re-added anytime. A group that still has Sessions persists as session-derived. This resolves the earlier no-removal-affordance caveat.

## Tests

`packages/web`: 60 files / 703 tests, all green. This branch adds: `nav-group-collapse` (9), `pinned-sessions` (9), `formatRelativeShort` ×4, `session-order` (15), `matchesSessionQuery` ×3, `group-list` (1), `workspace-registry` (12 — normalization, register/unregister with same-ref fast exits, alias set/clear/trim, entry round-trip, old string-shape migration, junk tolerance, throwing storage, empty-group merge order, alias relabelling session-backed groups, temp-path exclusion).

## Review fixes

Applied after code review (16 findings):

**Correctness / data loss**
- Manual drag committed only the display-capped VISIBLE ids, so hidden loaded rows fell out of the stored sequence and came back as "newcomers" at the top (15 rows / cap 10 scrambled after one drag). The drop now commits the group's full loaded partition, and `moveInSequence` returns the input reference for every no-op so an inert drop no longer rewrites and persists.
- All four stores resolved `localStorage` as a DEFAULT PARAMETER, outside the guarded body — a SecurityError (blocked site data, partitioned iframe) escaped, and these run from `useState` initializers, so the sidebar threw on first render. Resolution moved inside the try.
- The drag payload put the raw session id on `text/plain`; the composer is a controlled textarea with no drop guard, so a mis-aimed reorder pasted a session id into the user's message. Now a private MIME type.
- win32 drive roots: normalization turned `C:\` into the drive-RELATIVE `C:`, so the registry and Sessions keyed differently (phantom group, new chats resolving elsewhere). Roots preserved.
- Temp-shaped paths were accepted at registration but never groupable — an unremovable ghost entry. Rejected up front.
- The workspace browse panel is long-lived and loads lazily once, so after a Project switch it showed the previous Project's directories and registered them into the new Project. It now remounts per Project.

**Accessibility**
- `Dropdown` had no focus management at all, which made the relocated row actions keyboard-unreachable. It now focuses the first item on open, walks items with Up/Down, and returns focus to the trigger on Escape — yielding to panels that own their own keyboard model (model-select and the slash/skill pickers) and to text fields. Triggers gained `aria-haspopup`.
- IME `isComposing` guard on the workspace-rename Enter (accepting a Chinese candidate saved raw pinyin).

**Behavior**
- Group/folder toggles are inert while searching (they changed nothing on screen but persisted collapse state and fired category fetches).
- The pin action is hidden on folder rows: it could never reorder them, and the stray id shifted the active list's drag partition.
- Empty registered workspaces sort after session-backed groups instead of taking the front of the 10-group capped window; the cap widens when one is added, so a new Workspace is still revealed immediately.
- 手动排序 is offered only with a drag-capable pointer — HTML5 DnD never fires from touch and the setting is global, so the mobile drawer could freeze a list it had no gesture to fix (a stored "manual" degrades to recency there).
- The manual order is stored per grouping mode: one shared array was read against partitions the mode defines, so switching modes scrambled the other's order.
- The search empty state now says it covers loaded chats, and the pinned-cluster comment no longer claims rows beyond the loaded pages surface (pinning reorders fetched rows only — there is no server-side pin).

Not addressed by design: the PR is not split (the user asked for these features in this PR), and `createdAt`-based times/sorting stay until #280 lands.

## Verification

All at the worktree root with Node 24 (re-run after merging origin/main and the `lastActiveAt` switchover):

- `pnpm -r build` — OK
- `pnpm typecheck` — OK (all packages)
- `pnpm test` — all green: skills 21, docs 32, landing 39, core 792 (+5 skipped), server 603, cli 235, **web 715 (60 files, was 650/55 on main)**, desktop 48
- e2e not run locally (pre-existing local failures on clean main); `chat.spec.mjs` (row menu) and `draft.spec.mjs` (grouping via the settings menu) updated for the new UI.

## Caveats

- Search covers loaded rows only (no server-side session search endpoint).
- **Client sorts by activity, server still pages by creation.** #280 added `lastActiveAt` but deliberately left `listByAgent` ordering at `created_at DESC` (there is no index on `last_active_at`), so a Session created long ago but active today sits on a later page and cannot climb into the sidebar's first page until 「更多」 pulls that page in. Fixing it needs a server-side ordering + index change — out of scope here, and noted in `orderSessionRows`.


packages/web/src/components/layout/sidebar.tsx
```diff
@@ -519,24 +799,87 @@ export function Sidebar({
...
+                      : { id: s.sessionId, after },
+                  );
+                },
+                onDragLeave: () => setDropHint((prev) => (prev?.id === s.sessionId ? null : prev)),
+                onDrop: (e: ReactDragEvent) => {
+                  if (!dragging || !samePartition) return;
+                  e.preventDefault();
+                  const rect = e.currentTarget.getBoundingClientRect();
+                  const after = e.clientY - rect.top > rect.height / 2;
+                  const partitionIds = rows
```
hiyouga: **Manual-sort drop commits only the display-capped ids, so a single drag replaces the visible list with rows the user has never seen.**

`rows` here is `shownActive` = `orderedActive.slice(0, cap)` (line 973), so `partitionIds` holds only the ~10 visible ids. `applyManualReorder` then stores exactly those, and on the next render `applyManualOrder` classifies every other loaded row as an "unlisted newcomer" and fronts them (`return [...unlisted, ...listed]`, session-order.ts:105).

A workspace group spanning 2+ Agents routinely holds more loaded active rows than `cap` (each Agent fetches its own page of 10; `hasMore` explicitly covers `parts.active.length > cap`). Reproduced with the PR's own helpers — 15 loaded rows, cap 10, drag `s1` to just after `s3`:

```
BEFORE drag, visible: s1 s2 s3 s4 s5 s6 s7 s8 s9 s10
AFTER  drag, visible: s11 s12 s13 s14 s15 s2 s3 s1 s4 s5
expected            : s2 s3 s1 s4 s5 s6 s7 s8 s9 s10
```

The five previously-hidden rows jump to the top and half the rows being arranged fall below the cap. The same happens on every later "More" click: the newly loaded *older* rows render above the manually ordered ones. The unit tests never model the cap, so they pass.

Fix direction: commit the **full** loaded partition (`parts.active` filtered by pin), not the capped slice — or stop fronting unlisted rows.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -932,75 +1498,102 @@ export function Sidebar({
...
-              <div key={group.key} className="pt-2.5">
-                {/* Group header: collapse toggle (folder icon + directory basename + count, full
+          // Same search treatment as agent mode: cap bypassed, zero-match groups hidden, the rest forced open.
+          orderedWorkspaceGroups
+            .slice(0, searching ? orderedWorkspaceGroups.length : groupCap)
+            .map((group) => {
+              const groupRows = filterRows(group.sessions);
+              if (searching && groupRows.length === 0) return null;
+              const parts = partitionSessions(groupRows);
+              const collapsed = !searching && collapsedGroups.has(group.key);
```
hiyouga, line 1607: **While a search is live, the group header's collapse toggle silently rewrites persisted state.**

`collapsed` is computed as `!searching && collapsedGroups.has(group.key)` (also line 1444 in agent mode, and 1406 for the drafts group), but `onToggle={() => toggleGroup(group.key)}` is unchanged — and `toggleGroup` both flips `collapsedGroups` and calls `saveGroupSet(collapseStoreKey, next)`.

So during a search: the user clicks a header to fold a noisy group away, nothing moves (search wins), `aria-expanded` still reads `true`, so they click a few more. On clearing the search, groups they never meant to collapse are collapsed and previously-collapsed ones are expanded — persisted to localStorage across reloads. Folders have the same shape (`open={searching || openFolders.has(...)}` at line 926 with a live `onToggle`).

Either make the toggle a no-op while `searching`, or keep the header reading the real stored state.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -609,14 +959,26 @@ export function Sidebar({
...
   ) => {
     const cap = groupCaps.get(groupKey) ?? SIDEBAR_PAGE_SIZE;
-    const shownActive = parts.active.slice(0, cap);
+    // Row order: the pinned cluster always first (never hidden behind "More"); under
+    // manual sort the stored order additionally applies within each pin partition
+    // (lib/session-order.ts). Folder rows keep their chronological order: pinning and
+    // manual order are active-list concerns. While searching, the display cap is
+    // bypassed — every loaded match shows, and "More" hides (it pages the unfiltered
+    // list and would read as "more matches", which the server cannot promise).
+    const orderedActive = orderSessionRows(parts.active, (s) => s.sessionId, {
```
hiyouga, line 1046: **Pinning only reorders already-fetched rows, so a pinned conversation outside the first page simply does not render.**

`orderSessionRows`/`pinnedFirst` partition the input array and nothing else — there is no fetch anywhere in `pinned-sessions.ts`, and no call site requests a Session by id to force a pinned row into the page. The store fetches one page per Agent per reload: `limit: SIDEBAR_PAGE_SIZE + 1` (state/sessions.tsx:168), kept at 10, server-sorted `createdAt DESC`.

Trigger: pin conversation P, start 10+ newer conversations under the same Agent (or in the same Workspace group), reload. P is not in the fetched page, so it is absent from `parts.active` and the pin does nothing — the user must click "More" until P loads.

This directly contradicts the comment six lines above: `// Row order: the pinned cluster always first (never hidden behind "More")`.

A real fix needs a server-side pin field (or a targeted fetch of pinned ids) — worth deciding before this ships, since the localStorage layer would then have to be reconciled.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -1439,52 +2139,156 @@ function SessionRow({
...
+            whole-slot overlay: the slot's width rides the time string (2 分钟前 vs 31 分钟前),
+            and a slot-centered glyph landed at a different x per row, so the ellipses never
+            formed a vertical column (the user saw it shift with the time's character count).
+            Right-anchored, every row's dots share one x and line up with the group-header
+            ellipsis above (a right-flush 28px square at the same 4px inset). The swap stays
+            a pure opacity handoff: time hides on row hover (group-hover), when the trigger
+            holds keyboard focus (peer-focus-visible; the button precedes the time span so
+            the peer combinator can reach it), and while the menu is open — regardless of the
+            time being wider than the square. */}
+        <Dropdown
```
hiyouga, line 2247: **Rename / archive / delete are no longer reachable by keyboard after moving into the body-portaled menu.**

On `main` these were three sibling buttons carrying `focus-visible:opacity-100` (the `actionBtn` class), so Tab walked title → 重命名对话 → 归档 → 删除对话. They now live inside `Dropdown`, and `packages/web/src/components/ui/dropdown.tsx` (217 lines) contains **zero** occurrences of `focus`, `tabIndex`, `role=`, `autoFocus` or `onKeyDown` — no focus-into-panel, no `role="menu"`, no roving tabindex, no focus trap, no focus restore.

Because the panel is `createPortal(..., document.body)`, it is appended at the end of the document: pressing Enter on 对话选项 opens a floating panel, and the next Tab lands on the following session row, not on a menu item. Reaching 删除对话 means tabbing through the rest of the page.

The same regression applies to the grouping toggle, which moved from two inline `aria-pressed` buttons (`GroupModeToggle`, one Tab away) into the list-settings Dropdown at line 1305.

The fix belongs in `Dropdown` (focus first item on open, arrow/Home/End, Escape restores focus to the trigger) rather than in the sidebar.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -519,24 +799,87 @@ export function Sidebar({
...
+          dragging.id !== s.sessionId &&
+          pinnedSessions.has(dragging.id) === pinnedSessions.has(s.sessionId);
+        const drag =
+          dragScope === undefined
+            ? {}
+            : {
+                draggable: true,
+                onDragStart: (e: ReactDragEvent) => {
+                  // Firefox refuses to start a drag without payload data.
+                  e.dataTransfer.setData("text/plain", s.sessionId);
```
hiyouga: **The drag payload is the raw session id on `text/plain`, so a mis-aimed reorder drag pastes a UUID into the user's message.**

`onDragOver`/`onDrop` only `preventDefault()` for same-partition rows, so everywhere else the browser's native text-drop applies. A grep across `packages/web/src` finds no `onDrop`/`onDragOver`/`onDragEnter` outside this file and no window/document-level drag listeners. The composer (`features/chat/chat-input.tsx:2349`) is a controlled `<textarea>` with `value`/`onChange` and no drop handling — a native text drop mutates the DOM value and dispatches `input`, which is exactly what React's `onChange` listens to, so `setText` runs and the draft is silently rewritten with the session id inserted at the drop caret. Sidebar and composer are co-mounted, so the gesture is one screen away (gated on 手动排序 being on).

Firefox only needs *some* payload — a private type such as `application/x-penguin-session` satisfies it without being droppable into text fields. A document-level `dragover` guard while `dragSession` is set would also work.

packages/web/src/lib/pinned-sessions.ts
```diff
@@ -0,0 +1,82 @@
...
+  `penguin.pinnedSessions.${projectId}`;
+
+/**
+ * Reads a Project's persisted pinned ids; no Project yet, nothing stored, or corrupted
+ * storage degrade to the empty set (nothing pinned — the default). A stored array
+ * survives element-level junk: non-strings are dropped, strings kept.
+ */
+export function loadPinnedSessions(
+  projectId: string | null,
+  storage: PinnedSessionsStorage = localStorage,
```
hiyouga: **`storage = localStorage` is a default-parameter expression, so it is evaluated *outside* the body's `try` — the documented "throwing storage" tolerance does not hold.**

In a storage-blocked context (third-party/partitioned iframe, or a profile with site data blocked) reading `window.localStorage` throws `SecurityError` on property **access**, before any `getItem` call. Default parameters are evaluated in the parameter scope, so the throw escapes the function and the `catch` never runs. Verified:

```
$ node defparam.mjs
THREW OUT OF THE FUNCTION (catch never ran): SecurityError: access denied
```

All five loaders are called from `useState` initializers / the project effect during the sidebar's first render — `loadPinnedSessions` (sidebar.tsx:389), `loadSessionOrder` (395), `loadWorkspaceRegistry` (399), `initialNavGroupCollapsed` (376), `initialSessionSortMode` (392) — so the sidebar throws on mount, white screen.

The unit tests inject a storage whose *methods* throw, which is the one shape that already works, so they give false confidence. Same line in `session-order.ts:38/64/78`, `workspace-registry.ts:70/91`, `nav-group-collapse.ts:53/64`. Take `storage?: X` and resolve it inside the `try` (or add a shared `safeLocalStorage()`).

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -1439,52 +2139,156 @@ function SessionRow({
...
+                    menuOpen ? "opacity-0" : ""
+                  }`}
+                >
+                  {lastActive}
+                </span>
+              )}
+            </>
+          }
+        >
+          <button type="button" className={overflowMenuRowClass} onClick={item(onTogglePin)}>
```
hiyouga: **Rows inside the Subagents / Scheduled / Archived folders offer 置顶, but folder rows never go through `orderSessionRows` — pinning them is a persisted no-op with side effects.**

`renderFolder` calls `renderRows(rows, withAgentHint)` with `parts[category]` straight from `partitionSessions` (line 932); only `parts.active` is passed through `orderSessionRows` (line 968). `SessionRow` renders the pin item unconditionally.

So: open 对话选项 on an archived chat, click 置顶 — the id is written to `penguin.pinnedSessions.<projectId>`, the row grows a 已置顶 glyph, and the row does not move. The write is not inert either: `samePartition` in `renderRows` (line 813) compares `pinnedSessions.has(...)`, so the stray pin changes the active list's drag partition boundary, and unarchiving the chat later makes it jump to the top for no reason the user can trace.

Either hide the pin item for folder rows, or apply `orderSessionRows` there too.

packages/web/src/lib/session-order.ts
```diff
@@ -0,0 +1,168 @@
...
+  keyOf: (row: T) => string,
+  order: readonly string[],
+): T[] {
+  if (order.length === 0) return [...rows];
+  const pos = new Map(order.map((id, i) => [id, i]));
+  const unlisted: T[] = [];
+  const listed: T[] = [];
+  for (const row of rows) (pos.has(keyOf(row)) ? listed : unlisted).push(row);
+  listed.sort((a, b) => (pos.get(keyOf(a)) ?? 0) - (pos.get(keyOf(b)) ?? 0));
+  return [...unlisted, ...listed];
```
hiyouga, line 122: **One global sort mode plus one per-Project order array, applied across partitions whose boundaries depend on the grouping mode — switching modes scrambles the list.**

`SESSION_SORT_MODE_KEY` is global and `sessionOrder` is a single per-Project array, but the partition it is read against is `groupKey`-scoped and mode-dependent (`dragScope` is the group key in both modes).

Drag Agent A's rows into place in agent mode (`order = [a3,a1,a2]`), then switch to 按工作区分组 — one click in the same list-settings menu. Workspace group W now holds `[b1,a1,a2,b2]`, where b1/b2 belong to Agent B and were never dragged: `applyManualOrder` marks them unlisted and fronts them, giving `[b1,b2,a1,a2]` — the oldest row in the group hoisted above two newer ones. This happens in every group that mixes the other mode's partitions, and symmetrically on the reverse switch.

The module header's premise ("only the RELATIVE order of ids that render together in one partition ever matters") does not survive the mode switch, because the same ids are partitioned differently. Either scope the storage key by grouping mode, or interleave unlisted rows by recency instead of fronting them.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -1258,6 +1851,36 @@ export function Sidebar({
...
+        <Input
+          label={S.chat.renameWorkspaceLabel}
+          hint={S.chat.renameWorkspaceHint}
+          value={workspaceAliasText}
+          placeholder={renamingWorkspace ? workspaceLabel(renamingWorkspace.path) : ""}
+          autoFocus
+          maxLength={80}
+          onChange={(e) => setWorkspaceAliasText(e.target.value)}
+          onKeyDown={(e) => {
+            if (e.key === "Enter") confirmRenameWorkspace();
```
hiyouga: **The workspace-rename Enter handler has no `isComposing` guard, so accepting a Chinese IME candidate saves the half-composed alias and closes the dialog.**

zh UI: the user types `shengchan` to enter 「生产」, the candidate list is up, and they press Enter to accept it. `keydown` fires first with `e.nativeEvent.isComposing === true`, so `confirmRenameWorkspace()` runs against the current `workspaceAliasText` (the pinyin), persists it, and `setRenamingWorkspace(null)` closes the dialog before composition ends. The workspace ends up labelled `shengchan`.

The repo guards exactly this in six other places, including the sibling picker in this same feature — `workspace-select.tsx:210`, `model-select.tsx:99`, `chat-input.tsx:1790/1808/1836`.

```suggestion
            if (e.key === "Enter" && !e.nativeEvent.isComposing) confirmRenameWorkspace();
```

(The 重命名对话 modal a few lines up shares the gap — pre-existing, but worth fixing in the same pass.)

packages/web/src/lib/workspace-registry.ts
```diff
@@ -0,0 +1,176 @@
...
+  `penguin.sidebarWorkspaces.${projectId}`;
+
+/**
+ * Canonical form of a picked path: trimmed, trailing separators dropped (the server
+ * hands back resolved paths without them, so `/srv/app/` must dedup against
+ * `/srv/app`), the filesystem root kept as-is. Empty in → empty out (never registered).
+ */
+export function normalizeWorkspacePath(path: string): string {
+  let p = path.trim();
+  while (p.length > 1 && (p.endsWith("/") || p.endsWith("\\"))) p = p.slice(0, -1);
```
hiyouga: **The `p.length > 1` guard protects the POSIX root but strips the Windows drive root: `C:\` normalizes to `C:`, a drive-*relative* path.**

Verified: `normalizeWorkspacePath("C:\\") === "C:"`, while `path.win32.resolve("C:\\") === "C:\\"` — core keeps the trailing backslash at a drive root, so a Session created there is keyed `C:\` by `workspaceGroupKey` while the registry entry is keyed `C:`. Two groups for one drive; the alias never lands on the session-backed one; and `registeredPaths.has(group.key)` (sidebar.tsx:1564) puts the rename/remove overflow on the phantom group only.

The path is reachable: `routes/dirs.ts:65-69` returns `fs.realpath(target)` and treats the drive root as a browsable terminus (`path.win32.dirname("C:\\") === "C:\\"` → `parent: null`), handing `C:\` straight to `onChange(dir.path)`. win32 is a supported target (`ci-windows` job, desktop `--win` build, `TEMP_WORKSPACE_RE` matching backslashes).

Worse consequence: sidebar.tsx:1556 feeds `group.fullPath` back as a new Session's workspace, and `path.win32.resolve("C:")` resolves against the server process's current directory on C: — so "new chat in this workspace" would create the Session in a **different directory than the one the user picked**.

The unit test covers `C:\\work\\repo\\` but never a bare drive root.

packages/web/src/lib/workspace-registry.ts
```diff
@@ -0,0 +1,176 @@
...
+  registered: readonly WorkspaceEntry[],
+): WorkspaceGroup<T>[] {
+  const aliasByKey = new Map(
+    registered.filter((e) => e.alias !== undefined).map((e) => [e.path, e.alias as string]),
+  );
+  const existing = new Set(groups.map((g) => g.key));
+  const added: WorkspaceGroup<T>[] = [];
+  for (const { path, alias } of registered) {
+    const key = workspaceGroupKey(path);
+    if (existing.has(key) || key !== path) continue; // already grouped, or a temp-shaped path
```
hiyouga, line 183: **A temp-shaped picked path is accepted by `registerWorkspace` but silently discarded here — the click does nothing and the entry can never be removed.**

`registerWorkspace` only rejects the empty string (line 111), so a path matching `…/workspaces/tmp-<8hex>` is prepended and written to `penguin.sidebarWorkspaces.<projectId>`. Then `key !== path` here drops it, so no group is ever added: from the user's side 使用此目录 did nothing, with no error or toast.

The entry is now stuck: no group exists, so `registeredPaths.has(group.key)` is never true and the rename/remove overflow never appears for it; re-picking the same path hits `registerWorkspace`'s same-reference fast exit and stays a no-op forever. The only way out is clearing localStorage.

Reject temp-shaped paths in `registerWorkspace` (and surface why in the picker) rather than accepting a write the merge will throw away.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -294,8 +451,18 @@ export function Sidebar({
...
+  /** Collapse/expand the page-nav group (same store-then-set convention as setGroupMode). */
+  const toggleNavGroup = () => {
+    const next = !navCollapsed;
+    storeNavGroupCollapsed(next);
+    setNavCollapsed(next);
+  };
+
+  /** Workspace groups (workspace mode): computed from the flat list, temp directories merged last, plus the manually-added Workspaces as empty groups on top (newest registration first). */
+  const workspaceGroups = useMemo(
+    () => mergeRegisteredWorkspaces(groupSessionsByWorkspace(sessions), registeredWorkspaces),
```
hiyouga, line 492: **Registered-but-empty Workspaces are unconditionally prepended ahead of every session-derived group, and the 10-group display cap is applied after that.**

`mergeRegisteredWorkspaces` returns `[...added, ...relabelled]` with `added` in raw registration order and no interleaving by recency, so the recency ordering `groupSessionsByWorkspace` establishes ("groups sorted by their newest Session, temp last") is lost at the top of the list. `orderedWorkspaceGroups.slice(0, groupCap)` (line 1513, `groupCap = SIDEBAR_GROUP_PAGE_SIZE = 10`) then fills the visible window with placeholders.

A user who registers ~10 directories up front to set them up for later gets a sidebar where **every group holding actual conversations** is hidden behind 更多分组. With two or three registrations it is subtler but still wrong: the top of the list is oldest-intent, not most-recent-activity.

An empty registered group never dedups away either (only ones with Sessions do), so this persists until the user explicitly removes each entry.

packages/web/src/lib/session-order.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+
+/** Moves `dragId` next to `targetId` (after it when `after`, else before) in a displayed sequence; no-ops when either id is missing or they are the same. */
+export function moveInSequence(
+  sequence: readonly string[],
+  dragId: string,
+  targetId: string,
+  after: boolean,
+): string[] {
+  if (dragId === targetId || !sequence.includes(dragId)) return [...sequence];
```
hiyouga: **A no-op drop still rewrites and persists the order, because `moveInSequence` returns a fresh array on the no-op paths and `commitManualDrop` has no identity guard.**

Both early returns here (`dragId === targetId`, and `at < 0` on line 143) return `[...sequence]` — a new array with identical contents. `commitManualDrop` (sidebar.tsx:538-544) calls `setSessionOrder` + `saveSessionOrder` unconditionally, unlike every other same-reference fast exit in this PR (`removePinnedSession`/`removeFromSessionOrder` at sidebar.tsx:666/671, `applyRegistryChange` at 747, which all rely on identity).

So picking a row up and dropping it back on its own edge writes the current visible sequence into storage — and in any group with more loaded rows than the display cap that immediately triggers the reshuffle described on the drop handler: the user changed nothing and the list rearranged.

Return the **input** `sequence` on the no-op paths and have `commitManualDrop` skip the state update and the write when the sequence is unchanged.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -804,45 +1168,242 @@ export function Sidebar({
...
+                  navigate("/agents", { state: { create: true } });
+                  onNavigate?.();
+                }}
+                className={headerControlClass(false)}
+              >
+                <AddBadgeIcon base={NAV_ICONS.agents} />
+              </button>
+            ) : (
+              <WorkspaceSelect
+                projectId={currentProjectId ?? ""}
```
hiyouga, line 1469: **This `WorkspaceSelect` is a long-lived mount, but the component only ever loads directories on its *first* expand — so it goes stale across Project switches.**

`workspace-select.tsx:118`: `loadOnFirstOpen` sets `browsedRef.current = true` and never resets it; `loadDir` is memoized on `projectId` but nothing re-fetches when `projectId` changes. The pill and form variants never hit this because they remount per draft / per dialog — this trigger sits in the section header with no `key`, so it survives a Project switch.

Browse to `/srv/projA` in Project A, switch to Project B (a pure state update — the Sidebar is not remounted): reopening the panel shows Project A's listing with no refetch, and 使用此目录 registers that path into **Project B's** `penguin.sidebarWorkspaces.<projectId>`.

Secondary: before `currentProject` resolves, `projectId` is `""` and opening the panel requests `/api/projects//dirs`, which 404s into the error row; `browsedRef` is already latched, so reopening never auto-retries (the in-panel 重试 button does recover, since it closes over the fresh `projectId`).

Reset `browsedRef` on a `projectId` change, or gate the trigger on `currentProjectId !== null`.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -563,17 +906,24 @@ export function Sidebar({
     totals: SessionCategoryCounts | undefined,
   ) => {
     const rows = parts[category];
+    // While searching the folder speaks for its loaded MATCHES only: a match hidden
+    // behind a collapsed folder would look like a missing result (the models page's
+    // search-forces-open rationale), so the folder is forced open, labelled by the
+    // match count, hidden when nothing matches, and never offers "More" (the server
+    // cannot search unloaded rows).
+    if (searching && rows.length === 0) return null;
```
hiyouga, line 989: **While a query is live the UI removes every control that could load more rows, then states flatly that nothing matches.**

The search is client-side over loaded rows only (line 524-525, and `listSessions` in `api/endpoints.ts` takes no query param). On a fresh load the client holds ~10 active rows per Agent and **zero** folder rows: `reload()` fetches `active` plus only folder categories already in `pageState` (state/sessions.tsx:159-161), which is empty on mount, and folders fetch exclusively via `toggleFolder` on first manual expand. `open={searching || ...}` is presentational and triggers no fetch, so an unloaded folder stays unloaded and this line hides it entirely.

Meanwhile "More" is suppressed in both places while searching (`const more = !searching && ...` here, `const hasMore = !searching && ...` at 979), and lines 1592-1596 print a definitive 「无匹配会话」 / "No matching chats".

So a user searching for an archived chat, a scheduled run, or simply the 11th-newest conversation gets a confident "no matches" and may conclude it was deleted. The only escape is to clear the search, manually expand each folder, click More until the target loads, and retype — nothing on screen suggests it.

A `q=` param on the list route is the natural depth (the route already parses `category`/`counts`/`cli` and the rows carry `title`). Short of that, the empty state should say "among loaded conversations" and "More" should stay available.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -804,45 +1168,242 @@ export function Sidebar({
...
+                label={S.chat.groupByAgent}
+                checked={groupMode === "agent"}
+                onSelect={() => {
+                  setGroupMode("agent");
+                  setListSettingsOpen(false);
+                }}
+              />
+              <div className="my-1 border-t border-gray-100 dark:border-gray-800" />
+              <p className={menuSectionClass}>{S.chat.sortModeSection}</p>
+              <MenuRadioRow
```
hiyouga: **手动排序 is offered on touch devices, where the only way to change the order does not work — and the preference is global, so the choice follows the user to their desktop.**

The whole Sidebar is rendered inside the mobile drawer (`app-layout.tsx:310-314`, `<Drawer open={drawerOpen} side="left">` behind the `md:hidden` hamburger), so this menu row is fully reachable on a phone. Reordering is implemented purely as HTML5 drag-and-drop (`draggable: true` + `onDragStart`/`onDragOver`/`onDrop`, lines 814-861) — mobile Safari and Chrome for Android do not synthesize `dragstart` from touch, and the diff adds no pointer-events or long-press fallback.

So on a phone: tap 列表选项 → 手动排序. `storeSessionSortMode` writes the single **global** key `penguin.sidebarSortMode` (session-order.ts:28 — deliberately not per-Project, per its comment), so the desktop session flips too. The phone list is now pinned to whatever order is already stored, `applyManualOrder` fronts every unstored row as a newcomer, and there is no gesture that can change any of it. The radio still reads 手动排序, so nothing signals that the mode is inert here.

Either gate the row on a pointer-capable device, or add a touch-capable reorder affordance (pointer events / long-press handle).

PR #280, opened 2026-08-13
feat(server): track session lastActiveAt with migration and backfill

hiyouga:
## What / why

Sessions persisted only `createdAt`; rows now carry a server-maintained `lastActiveAt` so session lists can reflect real last activity (user-approved design). The sidebar display/sorting switch is deliberately NOT in this PR — that lands later via PR #277.

## Design

- **Schema**: new `last_active_at TEXT` column on `sessions` (ISO 8601, same convention as `created_at`), in both the fresh-DB `CREATE TABLE` and `openDatabase`'s idempotent `ensureColumn` ALTER-on-open list — the repo's existing migration mechanism. The value is **monotonic**: both writes clamp with `MAX(COALESCE(last_active_at, ''), ?)`, so a backwards clock step (NTP, resume from suspend) can never regress a row below its previous value or its `created_at`.
- **Write chokepoint**: `SessionManager.drive()` — the one place every driven run passes through (the same spot that already flipped `has_trace`). A run stamps exactly twice: once at start, once at end. A "run" is one Task, one compaction, or one **whole goal loop** (`startGoal` hands all rounds to a single `drive`), so an N-round goal stamps twice, not 2N times. Mid-run activity (steering, approvals, queued follow-ups) rides on the run-end stamp; nothing writes per streamed message. The run-start write is folded into `SessionsRepo.markDriven` (`has_trace` + stamp in one statement: one WAL commit, not two), and both writes are guarded — they sit outside `drive`'s covering try (one before it, one in its `finally`), where an `ERR_INVALID_STATE` from a handle closed by shutdown would otherwise pin the entry at `running` or skip the idle broadcast, title generation, and the auto-start of queued follow-ups.
- **Creation**: `SessionRow.lastActiveAt` is required and set explicitly at every insert site, so no layer computes its own default independently of what was stored. The inserts additionally default it **in SQL** (`COALESCE(?, ?)` over the row's own `created_at`, bound twice — SQLite forbids sibling-column references inside `VALUES`, and a column added by `ALTER TABLE` cannot be given a `DEFAULT`); that statement is the only place able to guarantee the cell is never written NULL.
- **Backfill choice**: one-time, inside the same transaction as the ALTER, and **gated on the ALTER actually happening** (so it never re-scans `sessions` on subsequent opens). Legacy rows take `MAX(usage_records.ts)` for their session — the most recent request timestamp, the closest persisted proxy for last Trace activity — falling back to `created_at`. True Trace-tail timestamps were deliberately not read: neither `trace_files` nor `trace_sessions` stores a last-event timestamp (only date-granularity shard dates / `first_ts`), so deriving them would mean scanning every session's shard files at startup.
- **Index**: `idx_usage_session (session_id)` is reshaped to `idx_usage_session_ts (session_id, ts)` so the backfill's correlated `MAX(ts)` becomes a covering-index scan (`EXPLAIN QUERY PLAN`: `SEARCH usage_records USING COVERING INDEX idx_usage_session_ts`) instead of a random table lookup per usage row — this runs synchronously before the HTTP listener binds, so an unindexed scan would show up as a dead server on the first launch after upgrade. The superseded index is dropped on open, since `CREATE INDEX IF NOT EXISTS` never reshapes an existing one; `(session_id)` is a strict prefix of the new shape, so every query it served is served identically.
- **API**: `SessionInfo.lastActiveAt: string` (additive) flows through create / list / single-GET / PATCH. PATCH re-reads the row after its writes — it previously answered from a snapshot taken before two awaits, so a run finishing in that window returned a regressed value that the web store adopts wholesale. The web app consumes the type from `@prismshadow/penguin-server/api`, so no web source changes were needed (one web test factory gained the field). List ordering is unchanged (`createdAt` desc); switching it to `lastActiveAt` is PR #277's job.
- **Known limitation (documented in the type)**: rows this server never drives keep `createdAt` — Sessions adopted from a CLI Trace, and subagent rows (their work is driven through the parent Session's entry). Making those track real activity means consulting the Trace tree; it changes what the field means, so it is left as a separate decision.

## Backward compatibility

- **Old shape handled**: `web.db` files whose `sessions` table predates the `last_active_at` column, and whose usage index is the older `idx_usage_session (session_id)`.
- **What the migration does**: on the first open after upgrading, `openDatabase` creates `idx_usage_session_ts` and drops the superseded `idx_usage_session`, then ALTERs the column in (`ensureColumn`, guarded by a `PRAGMA table_info` check) and backfills every existing row once — from the session's most recent request timestamp in `usage_records`, else its `created_at`. ALTER and backfill share one transaction, and the backfill only runs in the open that actually adds the column, so the whole path is idempotent: a fresh DB and an already-migrated DB do no work, later opens never rewrite live values, and a crash mid-upgrade rolls back to "no column" and is redone on the next open (never a half-migrated table). An index is derived data, so the reshape costs at most one rebuild, never information.
- **User action**: none — the migration is automatic and silent. Large existing databases pay a one-time index build plus a single indexed pass over `sessions` on that first open.
- **How long the migration code must live**: the `upgradeLastActiveAt` helper (ALTER + gated backfill) and the `DROP INDEX IF EXISTS idx_usage_session` line in `packages/server/src/db/database.ts` must stay until a DB-schema floor is declared — a release explicitly allowed to break pre-floor `web.db` files, per the standing note in `database.ts`. Their removal needs an explicit future decision; flagged here for the changelog summary PR.

## Verification

All commands run at the worktree root with Node 24:

```
pnpm install --frozen-lockfile   # OK
pnpm -r build                    # all packages built
pnpm typecheck                   # all packages clean
pnpm test                        # exit 0: core 792 passed (5 skipped), server 62 files / 603 tests,
                                 # cli 235, web 650, desktop 48, skills 21, docs 32, landing 39 — all
                                 # green (Playwright e2e is a separate script, not part of pnpm test)
pnpm format && pnpm format:check # all matched files use Prettier code style
```

Tests (server):

- `db-upgrade.test.ts` — backfill from last request ts vs `created_at` (asserted on the **raw** column, not the coalesced read); the gate, proven by blanking a cell and reopening (it stays NULL, and `mapRow`'s coalesce is what keeps that invisible); the index reshape (new one present, old one gone); both insert paths storing a non-NULL cell including when a caller supplies nothing; the monotonic clamp under a backwards clock; and per-row isolation. The old-shape fixture is derived from `SCHEMA_SQL` (+ `ALTER TABLE … DROP COLUMN`) instead of a hand-copied DDL replica.
- `session-manager.test.ts` — the two stamps pinned **independently** under a stubbed clock: sampled mid-run at the approval pause (run-start stamp alone) and again after idle with a strict `>` (run-end stamp); compaction stamps its own pair; unrelated rows untouched. Plus a guarded-finally test: with a repo whose stamp writes always throw, both runs still reach idle, the queued follow-up still auto-starts, and the failure is logged and recorded as `session_touch_failed`.
- `session-index.test.ts` — `lastActiveAt` equals `createdAt` at creation (ISO 8601) through create, list, and single GET, with the stored cell read straight from SQLite; and PATCH answering with the post-write row rather than its pre-PATCH snapshot.

Mutation-checked (each of these now fails the suite, and each passed it before this round): deleting the run-end stamp; removing the SQL-side insert default; ungating the backfill.

### DIFF
--- a/packages/server/src/api/types.ts
+++ b/packages/server/src/api/types.ts
@@ -747,10 +747,19 @@ export interface SessionInfo {
   source?: SessionSource;
   createdAt: string;
   /**
-   * Last meaningful activity (ISO 8601, same convention as createdAt): initialized to
-   * createdAt at creation, stamped by the server at each driven run's start and end
-   * (task, goal round, compaction). Legacy rows are backfilled once at startup from the
-   * session's most recent request timestamp, falling back to createdAt.
+   * Last activity the server drove for this Session (ISO 8601, same convention as
+   * createdAt), monotonic — it never moves backwards. Set to createdAt at creation, then
+   * stamped when a run starts and again when it ends; a run here is one Task, one
+   * compaction, or one whole goal loop (every round of a goal belongs to a single run, so
+   * an N-round goal stamps twice, not 2N times).
+   *
+   * Two kinds of row therefore stay at createdAt no matter how busy they look: Sessions
+   * adopted from a CLI Trace (this server drives none of their runs) and subagent rows
+   * (their work is driven through the parent Session's entry). Reading real activity for
+   * those would mean consulting the Trace tree, which this field deliberately does not do.
+   *
+   * Rows that predate the field are backfilled once at startup from the Session's most
+   * recent request timestamp, falling back to createdAt.
    */
   lastActiveAt: string;
   status: SessionStatus;
--- a/packages/server/src/api/types.ts
+++ b/packages/server/src/api/types.ts
@@ -746,6 +746,13 @@ export interface SessionInfo {
   /** Session source (for list badges/folders), derived from core session_meta — the single source of truth (not stored in the DB); unset for user-created sessions. */
   source?: SessionSource;
   createdAt: string;
+  /**
+   * Last meaningful activity (ISO 8601, same convention as createdAt): initialized to
+   * createdAt at creation, stamped by the server at each driven run's start and end
+   * (task, goal round, compaction). Legacy rows are backfilled once at startup from the
+   * session's most recent request timestamp, falling back to createdAt.
+   */
+  lastActiveAt: string;
   status: SessionStatus;
   /** Number of approvals awaiting human decision (a persisted count outside server events, for list badges). */
   pendingApprovalCount: number;
--- a/packages/server/src/db/database.ts
+++ b/packages/server/src/db/database.ts
@@ -29,8 +29,20 @@ export function openDatabase(dbPath: string): DatabaseSync {
   // schema.ts; drop entries only in a release allowed to break existing web.db files.
   ensureColumn(db, "sessions", "client", "TEXT");
   ensureColumn(db, "sessions", "has_trace", "INTEGER NOT NULL DEFAULT 0");
+  ensureColumn(db, "sessions", "last_active_at", "TEXT");
   ensureColumn(db, "auth_sessions", "via", "TEXT");
   ensureColumn(db, "trace_files", "page_stats", "TEXT");
+  // last_active_at backfill for rows formed before the column existed: the session's most
+  // recent request timestamp (usage_records, served by idx_usage_session — the closest
+  // persisted proxy for "last Trace activity"), else created_at. Guarded by IS NULL so it
+  // is idempotent — a backfilled database matches 0 rows, and a crash between the ALTER
+  // above and this UPDATE heals on the next open.
+  db.exec(
+    `UPDATE sessions SET last_active_at = COALESCE(
+       (SELECT MAX(ts) FROM usage_records WHERE usage_records.session_id = sessions.session_id),
+       created_at
+     ) WHERE last_active_at IS NULL`,
+  );
   return db;
 }
 
--- a/packages/server/src/db/database.ts
+++ b/packages/server/src/db/database.ts
@@ -29,26 +29,52 @@ export function openDatabase(dbPath: string): DatabaseSync {
   // schema.ts; drop entries only in a release allowed to break existing web.db files.
   ensureColumn(db, "sessions", "client", "TEXT");
   ensureColumn(db, "sessions", "has_trace", "INTEGER NOT NULL DEFAULT 0");
-  ensureColumn(db, "sessions", "last_active_at", "TEXT");
   ensureColumn(db, "auth_sessions", "via", "TEXT");
   ensureColumn(db, "trace_files", "page_stats", "TEXT");
-  // last_active_at backfill for rows formed before the column existed: the session's most
-  // recent request timestamp (usage_records, served by idx_usage_session — the closest
-  // persisted proxy for "last Trace activity"), else created_at. Guarded by IS NULL so it
-  // is idempotent — a backfilled database matches 0 rows, and a crash between the ALTER
-  // above and this UPDATE heals on the next open.
-  db.exec(
-    `UPDATE sessions SET last_active_at = COALESCE(
-       (SELECT MAX(ts) FROM usage_records WHERE usage_records.session_id = sessions.session_id),
-       created_at
-     ) WHERE last_active_at IS NULL`,
-  );
+  // Superseded by idx_usage_session_ts (session_id, ts), which SCHEMA_SQL just created on
+  // this database: the old index is a strict prefix of it, so every query it served is
+  // served identically. Dropping is safe — an index is derived, never data.
+  db.exec("DROP INDEX IF EXISTS idx_usage_session");
+  upgradeLastActiveAt(db);
   return db;
 }
 
-/** Idempotent per-column upgrade for databases formed before the column existed. */
-function ensureColumn(db: DatabaseSync, table: string, column: string, ddl: string): void {
+/**
+ * One-time `sessions.last_active_at` upgrade, ALTER + backfill in a single transaction.
+ *
+ * The backfill runs **only in the open that actually adds the column** (SQLite's ALTER
+ * TABLE ADD COLUMN is transactional, so a crash mid-upgrade rolls back to "no column" and
+ * the next open redoes both halves — never a half-migrated table, and never a full
+ * `sessions` scan on the millions of opens that follow). Legacy rows take the session's
+ * most recent request timestamp — usage_records, covered by idx_usage_session_ts, the
+ * closest persisted proxy for "last Trace activity" — else their own created_at.
+ */
+function upgradeLastActiveAt(db: DatabaseSync): void {
+  db.exec("BEGIN");
+  try {
+    if (ensureColumn(db, "sessions", "last_active_at", "TEXT")) {
+      db.exec(
+        `UPDATE sessions SET last_active_at = COALESCE(
+           (SELECT MAX(ts) FROM usage_records WHERE usage_records.session_id = sessions.session_id),
+           created_at
+         )`,
+      );
+    }
+    db.exec("COMMIT");
+  } catch (err) {
+    db.exec("ROLLBACK");
+    throw err;
+  }
+}
+
+/**
+ * Idempotent per-column upgrade for databases formed before the column existed.
+ * Returns whether this call actually ALTERed the table (false = the column was already
+ * there), so a caller can gate one-time backfill work on it.
+ */
+function ensureColumn(db: DatabaseSync, table: string, column: string, ddl: string): boolean {
   const cols = db.prepare(`PRAGMA table_info(${table})`).all() as { name: string }[];
-  if (cols.some((c) => c.name === column)) return;
+  if (cols.some((c) => c.name === column)) return false;
   db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${ddl}`);
+  return true;
 }
--- a/packages/server/src/db/repos/sessions.ts
+++ b/packages/server/src/db/repos/sessions.ts
@@ -30,11 +30,14 @@ export interface SessionRow {
   /** Cache: a Trace record exists (set at task start / adoption / subagent registration; backfilled by list hydration). */
   hasTrace?: boolean;
   /**
-   * Last meaningful activity, ISO (bumped at each driven run's start and end — task, goal
-   * round, compaction; see SessionManager.drive). Always present on rows read from the DB
-   * (openDatabase backfills legacy rows once); omitting on insert defaults to createdAt.
+   * Last activity this server drove for the session, ISO: stamped once when a run starts
+   * and once when it ends (see SessionManager.drive — a goal run is ONE drive over all its
+   * rounds, so it stamps twice, not per round). Set to createdAt at insert, and only ever
+   * moves forward (markDriven/touchLastActive clamp with MAX, so a backwards clock step
+   * cannot regress it). Rows this server never drives — CLI-adopted Sessions, subagent
+   * rows (their activity is driven through the parent's entry) — keep createdAt.
    */
-  lastActiveAt?: string;
+  lastActiveAt: string;
   createdAt: string;
 }
 
@@ -62,33 +65,27 @@ export class SessionsRepo {
   constructor(private readonly db: DatabaseSync) {}
 
   insert(row: SessionRow): void {
-    this.db
-      .prepare(
-        `INSERT INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, last_active_at, created_at)
-         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
-      )
-      .run(
-        row.sessionId,
-        row.projectId,
-        row.agentId,
-        row.provider,
-        row.modelId,
-        row.workspace,
-        row.approvalMode,
-        row.title,
-        row.client ?? null,
-        row.hasTrace ? 1 : 0,
-        row.lastActiveAt ?? row.createdAt,
-        row.createdAt,
-      );
+    this.runInsert("INSERT", row);
   }
 
   /** Idempotent insert: used when Trace directory discovery backfills a row (concurrent listing discovering the same Session no longer triggers a UNIQUE violation). */
   insertOrIgnore(row: SessionRow): void {
+    this.runInsert("INSERT OR IGNORE", row);
+  }
+
+  /**
+   * The two inserts' shared column list and binding order (they differ only in conflict
+   * handling). `last_active_at` defaults **in SQL** (`COALESCE(?, ?)` over the row's own
+   * created_at, bound twice — SQLite forbids referencing a sibling column inside VALUES,
+   * and a column added by ALTER TABLE cannot be given a DEFAULT): the column is nullable
+   * for legacy reasons, so this is the one place that can guarantee it is never written
+   * NULL, whatever a caller passes.
+   */
+  private runInsert(verb: "INSERT" | "INSERT OR IGNORE", row: SessionRow): void {
     this.db
       .prepare(
-        `INSERT OR IGNORE INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, last_active_at, created_at)
-         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+        `${verb} INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, last_active_at, created_at)
+         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, COALESCE(?, ?), ?)`,
       )
       .run(
         row.sessionId,
@@ -101,20 +98,39 @@ export class SessionsRepo {
         row.title,
         row.client ?? null,
         row.hasTrace ? 1 : 0,
-        row.lastActiveAt ?? row.createdAt,
+        row.lastActiveAt ?? null,
+        row.createdAt,
         row.createdAt,
       );
   }
 
-  /** Flip the has_trace cache once a Trace record exists (task start / discovery hydration); idempotent. */
+  /** Flip the has_trace cache once a Trace record exists (discovery hydration); idempotent. */
   markHasTrace(sessionId: string): void {
     this.db.prepare("UPDATE sessions SET has_trace = 1 WHERE session_id = ?").run(sessionId);
   }
 
-  /** Bump last_active_at (drive's run start/end — one write per run boundary, never per streamed message). */
+  /**
+   * A driven run touched this Session: flip the has_trace cache and stamp last_active_at in
+   * ONE statement (drive's run start — one WAL commit instead of two).
+   * `MAX(COALESCE(...))` keeps the stamp monotonic: a backwards clock step (NTP, resume
+   * from suspend) leaves the stored value alone rather than regressing the row.
+   */
+  markDriven(sessionId: string, at: string): void {
+    this.db
+      .prepare(
+        `UPDATE sessions SET has_trace = 1, last_active_at = MAX(COALESCE(last_active_at, ''), ?)
+         WHERE session_id = ?`,
+      )
+      .run(at, sessionId);
+  }
+
+  /** Stamp last_active_at alone (drive's run end); monotonic like markDriven. */
   touchLastActive(sessionId: string, at: string): void {
     this.db
-      .prepare("UPDATE sessions SET last_active_at = ? WHERE session_id = ?")
+      .prepare(
+        `UPDATE sessions SET last_active_at = MAX(COALESCE(last_active_at, ''), ?)
+         WHERE session_id = ?`,
+      )
       .run(at, sessionId);
   }
 
--- a/packages/server/src/db/repos/sessions.ts
+++ b/packages/server/src/db/repos/sessions.ts
@@ -29,6 +29,12 @@ export interface SessionRow {
   client?: "web" | "cli" | null;
   /** Cache: a Trace record exists (set at task start / adoption / subagent registration; backfilled by list hydration). */
   hasTrace?: boolean;
+  /**
+   * Last meaningful activity, ISO (bumped at each driven run's start and end — task, goal
+   * round, compaction; see SessionManager.drive). Always present on rows read from the DB
+   * (openDatabase backfills legacy rows once); omitting on insert defaults to createdAt.
+   */
+  lastActiveAt?: string;
   createdAt: string;
 }
 
@@ -45,6 +51,9 @@ function mapRow(r: Record<string, unknown>): SessionRow {
     archivedAt: (r.archived_at as string | null) ?? null,
     client: (r.client as "web" | "cli" | null) ?? null,
     hasTrace: (r.has_trace as number) === 1,
+    // The open-time backfill leaves no NULLs; the coalesce only hardens against a row
+    // somehow inserted as NULL (degrades to createdAt instead of surfacing undefined).
+    lastActiveAt: (r.last_active_at as string | null) ?? (r.created_at as string),
     createdAt: r.created_at as string,
   };
 }
@@ -55,8 +64,8 @@ export class SessionsRepo {
   insert(row: SessionRow): void {
     this.db
       .prepare(
-        `INSERT INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, created_at)
-         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+        `INSERT INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, last_active_at, created_at)
+         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
       )
       .run(
         row.sessionId,
@@ -69,6 +78,7 @@ export class SessionsRepo {
         row.title,
         row.client ?? null,
         row.hasTrace ? 1 : 0,
+        row.lastActiveAt ?? row.createdAt,
         row.createdAt,
       );
   }
@@ -77,8 +87,8 @@ export class SessionsRepo {
   insertOrIgnore(row: SessionRow): void {
     this.db
       .prepare(
-        `INSERT OR IGNORE INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, created_at)
-         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+        `INSERT OR IGNORE INTO sessions (session_id, project_id, agent_id, provider, model_id, workspace, approval_mode, title, client, has_trace, last_active_at, created_at)
+         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
       )
       .run(
         row.sessionId,
@@ -91,6 +101,7 @@ export class SessionsRepo {
         row.title,
         row.client ?? null,
         row.hasTrace ? 1 : 0,
+        row.lastActiveAt ?? row.createdAt,
         row.createdAt,
       );
   }
@@ -98,6 +109,13 @@ export class SessionsRepo {
   /** Flip the has_trace cache once a Trace record exists (task start / discovery hydration); idempotent. */
   markHasTrace(sessionId: string): void {
     this.db.prepare("UPDATE sessions SET has_trace = 1 WHERE session_id = ?").run(sessionId);
+  }
+
+  /** Bump last_active_at (drive's run start/end — one write per run boundary, never per streamed message). */
+  touchLastActive(sessionId: string, at: string): void {
+    this.db
+      .prepare("UPDATE sessions SET last_active_at = ? WHERE session_id = ?")
+      .run(at, sessionId);
   }
 
   findById(sessionId: string): SessionRow | null {
--- a/packages/server/src/db/schema.ts
+++ b/packages/server/src/db/schema.ts
@@ -7,7 +7,9 @@
  * Product not yet released: no migration branches — everything is CREATE IF NOT EXISTS, formed
  * once. The one exception: columns added to an existing table after release of a web.db are
  * ALTERed in by the idempotent per-column guard in database.ts (ensureColumn), since CREATE
- * TABLE IF NOT EXISTS never touches an existing table.
+ * TABLE IF NOT EXISTS never touches an existing table. A *reshaped* index follows the same
+ * rule under a new name, with the superseded one dropped on open (idx_usage_session →
+ * idx_usage_session_ts): CREATE INDEX IF NOT EXISTS never rebuilds an existing index either.
  */
 
 export const SCHEMA_SQL = `
@@ -54,7 +56,7 @@ CREATE TABLE IF NOT EXISTS sessions (
   archived_at   TEXT,                                -- archive time; NULL=not archived (shown by default; archived ones move under "Archived")
   client        TEXT,                                -- creating client: 'web' (created via the Web App) | 'cli' (adopted from a CLI Trace); NULL = legacy row, treated as web
   has_trace     INTEGER NOT NULL DEFAULT 0,          -- cache: a Trace record exists (set at task start / adoption / subagent registration; lazily backfilled by list hydration)
-  last_active_at TEXT,                               -- last meaningful activity (ISO; bumped at each driven run's start and end, initialized to created_at); NULL only before openDatabase's one-time backfill
+  last_active_at TEXT,                               -- last activity this server drove for the session (ISO; stamped once when a run starts and once when it ends, initialized to created_at); monotonic, never moves backwards; NULL only before openDatabase's one-time backfill
   created_at    TEXT NOT NULL
 );                                            -- the schedule/subagent SOURCE is NOT stored: session_meta in the Trace is the single source of truth (see runtime/session-sources.ts); "client" is a different, DB-only axis (who created the row)
 CREATE INDEX IF NOT EXISTS idx_sessions_agent_created ON sessions(project_id, agent_id, created_at DESC);
@@ -75,7 +77,7 @@ CREATE TABLE IF NOT EXISTS usage_records (
   status            TEXT NOT NULL DEFAULT 'completed'  -- request outcome: completed=success (with tokens); others=failure (0 tokens, for success rate)
 );                                            -- cost is not stored: computed at query time from current pricing
 CREATE INDEX IF NOT EXISTS idx_usage_project_date ON usage_records(project_id, date);
-CREATE INDEX IF NOT EXISTS idx_usage_session ON usage_records(session_id);
+CREATE INDEX IF NOT EXISTS idx_usage_session_ts ON usage_records(session_id, ts);  -- (session_id) alone was idx_usage_session (dropped on open): ts makes the last_active_at backfill's MAX(ts) a covering index scan instead of a random row lookup per usage record
 CREATE TABLE IF NOT EXISTS error_records (     -- server-side error capture (the Costs page's "Errors" tab)
   id         INTEGER PRIMARY KEY AUTOINCREMENT,
   ts         TEXT NOT NULL,
--- a/packages/server/src/db/schema.ts
+++ b/packages/server/src/db/schema.ts
@@ -54,6 +54,7 @@ CREATE TABLE IF NOT EXISTS sessions (
   archived_at   TEXT,                                -- archive time; NULL=not archived (shown by default; archived ones move under "Archived")
   client        TEXT,                                -- creating client: 'web' (created via the Web App) | 'cli' (adopted from a CLI Trace); NULL = legacy row, treated as web
   has_trace     INTEGER NOT NULL DEFAULT 0,          -- cache: a Trace record exists (set at task start / adoption / subagent registration; lazily backfilled by list hydration)
+  last_active_at TEXT,                               -- last meaningful activity (ISO; bumped at each driven run's start and end, initialized to created_at); NULL only before openDatabase's one-time backfill
   created_at    TEXT NOT NULL
 );                                            -- the schedule/subagent SOURCE is NOT stored: session_meta in the Trace is the single source of truth (see runtime/session-sources.ts); "client" is a different, DB-only axis (who created the row)
 CREATE INDEX IF NOT EXISTS idx_sessions_agent_created ON sessions(project_id, agent_id, created_at DESC);
--- a/packages/server/src/runtime/session-manager.ts
+++ b/packages/server/src/runtime/session-manager.ts
@@ -1194,8 +1194,12 @@ export class SessionManager {
   ): Promise<void> {
     // Every driven run (task, goal round, compaction) writes Trace lines: flip the row's
     // has_trace cache here, the single choke point, so listing can serve it from the DB
-    // without a directory walk (see SessionService.listSessions).
+    // without a directory walk (see SessionService.listSessions). The same choke point
+    // stamps last_active_at — once per run boundary (here and in the finally below), never
+    // per streamed message, so activity during the run (steering, approvals, follow-up
+    // queueing) is covered by the run-end stamp without any hot-loop writes.
     this.deps.sessions.markHasTrace(entry.sessionId);
+    this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
     let earlyTitleFired = false;
     let mainBodyChars = 0;
     const ctx: UsageContext = {
@@ -1371,6 +1375,9 @@ export class SessionManager {
       // now-empty state.
       entry.pendingSteering = [];
       entry.lastActivityMs = Date.now();
+      // Run-end stamp (see the run-start counterpart at the top of drive). A no-op UPDATE
+      // when the row is mid-deletion — the DELETE already won.
+      this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
       this.publishState(entry, "idle");
       if (titleSource && titleSource.userExcerpt.trim()) {
         // Attempt generation whenever there's user material; whether generation is
--- a/packages/server/src/runtime/session-manager.ts
+++ b/packages/server/src/runtime/session-manager.ts
@@ -222,6 +222,12 @@ export interface SessionManagerDeps {
   log?: (line: string) => void;
   /** Goal run-state persistence (optional like `titles`: without it, goals run but leave no restorable record). */
   goals?: GoalsRepo;
+  /**
+   * Clock for persisted timestamps (last_active_at). Injected like the other services' so a
+   * stubbed clock moves this and `usage_records.ts` together — the pairing the legacy
+   * backfill assumes when it reads MAX(ts) as a session's last activity.
+   */
+  now?: () => Date;
 }
 
 /** One queued follow-up task (`queueIfBusy`): the task input plus the per-turn thinking level it was posted with. */
@@ -396,9 +402,12 @@ export class SessionManager {
   /** Open streaming fragments of running sessions (fed by drive, served to GET /messages; see live-tail.ts). */
   private readonly liveTail = new LiveTailTracker();
   private readonly sweepTimer: NodeJS.Timeout;
+  /** Clock for persisted timestamps (see SessionManagerDeps.now); wall clock unless injected. */
+  private readonly now: () => Date;
 
   constructor(private readonly deps: SessionManagerDeps) {
     this.log = deps.log ?? ((line) => console.error(line));
+    this.now = deps.now ?? (() => new Date());
     this.sweepTimer = setInterval(() => this.sweepIdle(), ENTRY_SWEEP_INTERVAL_MS);
     this.sweepTimer.unref?.();
   }
@@ -1192,14 +1201,6 @@ export class SessionManager {
     gen: AsyncGenerator<OmniMessage>,
     titleSource?: { userExcerpt: string },
   ): Promise<void> {
-    // Every driven run (task, goal round, compaction) writes Trace lines: flip the row's
-    // has_trace cache here, the single choke point, so listing can serve it from the DB
-    // without a directory walk (see SessionService.listSessions). The same choke point
-    // stamps last_active_at — once per run boundary (here and in the finally below), never
-    // per streamed message, so activity during the run (steering, approvals, follow-up
-    // queueing) is covered by the run-end stamp without any hot-loop writes.
-    this.deps.sessions.markHasTrace(entry.sessionId);
-    this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
     let earlyTitleFired = false;
     let mainBodyChars = 0;
     const ctx: UsageContext = {
@@ -1209,6 +1210,17 @@ export class SessionManager {
       provider: entry.provider,
       modelId: entry.modelId,
     };
+    // Every driven run (a Task, a compaction, or a whole goal loop — startGoal hands all of
+    // its rounds to ONE drive) writes Trace lines: flip the row's has_trace cache here, the
+    // single choke point, so listing can serve it from the DB without a directory walk (see
+    // SessionService.listSessions). The same statement stamps last_active_at, so a run costs
+    // one row write here and one more when it ends (see the finally) — never one per streamed
+    // message; activity during the run (steering, approvals, queued follow-ups) rides on the
+    // run-end stamp. Guarded: this runs BEFORE the try below, so an unexpected DB failure
+    // (a closed handle after shutdown) would otherwise reject drive() with no finally to
+    // reach — leaving the entry pinned "running" and every later Task 409ing for the
+    // process's lifetime.
+    this.touchRow(ctx, (repo, at) => repo.markDriven(ctx.sessionId, at));
     // LLM request failures and tool execution failures aren't expressed via throw (core
     // converges them into the message stream), so the try/catch below can't catch them:
     // the watcher inspects messages one by one and fishes them out for persistence
@@ -1375,9 +1387,14 @@ export class SessionManager {
       // now-empty state.
       entry.pendingSteering = [];
       entry.lastActivityMs = Date.now();
-      // Run-end stamp (see the run-start counterpart at the top of drive). A no-op UPDATE
-      // when the row is mid-deletion — the DELETE already won.
-      this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
+      // Run-end stamp (see the run-start counterpart at the top of drive). Guarded like
+      // every other write in this finally: what follows — the idle broadcast, title
+      // generation, and the auto-start of queued follow-ups (this is its only call site) —
+      // must not be skippable by a DB failure, or SSE clients would sit on "running"
+      // forever with their queue stranded. A no-op UPDATE when the row is already gone:
+      // deletion normally awaits the drive first, so that only happens once its 5s wait
+      // times out.
+      this.touchRow(ctx, (repo, at) => repo.touchLastActive(ctx.sessionId, at));
       this.publishState(entry, "idle");
       if (titleSource && titleSource.userExcerpt.trim()) {
         // Attempt generation whenever there's user material; whether generation is
@@ -1445,6 +1462,7 @@ export class SessionManager {
     // row deliberately stores no source column.
     const source = asSessionSource(p.source) ?? "subagent";
     this.deps.sources.set(childSid, source);
+    const createdAt = this.now().toISOString();
     this.deps.sessions.insertOrIgnore({
       sessionId: childSid,
       projectId: entry.projectId,
@@ -1458,7 +1476,10 @@ export class SessionManager {
       title: null,
       // Spawned by this server's run (client NULL = web); its Trace exists by construction.
       hasTrace: true,
-      createdAt: new Date().toISOString(),
+      // A subagent's own runs are driven through the PARENT entry's drive, so nothing ever
+      // stamps this row: it stays at its registration time (see SessionRow.lastActiveAt).
+      lastActiveAt: createdAt,
+      createdAt,
     });
     // Make the subagent appear immediately in the sidebar: notify via the parent
     // Session's channel (a frontend currently watching the parent run refreshes its list in place).
@@ -1478,6 +1499,25 @@ export class SessionManager {
     };
     children.set(childSid, child);
     return child;
+  }
+
+  /**
+   * Runs one `sessions` row write for a driven run's bookkeeping (has_trace /
+   * last_active_at), swallowing failures the same way the per-message `recorder.record`
+   * call does: bookkeeping must never take down a run's lifecycle. The two call sites both
+   * sit outside a covering try — the run-start write precedes drive's try block, the
+   * run-end one lives in its finally — and the realistic failure (the DatabaseSync handle
+   * closed by shutdown while a run outlived its drain window) throws ERR_INVALID_STATE.
+   */
+  private touchRow(ctx: UsageContext, write: (repo: SessionsRepo, at: string) => void): void {
+    try {
+      write(this.deps.sessions, this.now().toISOString());
+    } catch (err) {
+      this.log(
+        `[session] last-active bookkeeping failed: ${err instanceof Error ? err.message : String(err)}`,
+      );
+      this.deps.errors?.record({ source: "session", err, ctx, code: "session_touch_failed" });
+    }
   }
 
   private publishState(entry: RuntimeEntry, state: SessionStatus): void {
--- a/packages/server/src/services/session-service.ts
+++ b/packages/server/src/services/session-service.ts
@@ -82,7 +82,7 @@ export class SessionService {
       ...(row.title !== null ? { title: row.title } : {}),
       ...(source !== undefined ? { source } : {}),
       createdAt: row.createdAt,
-      lastActiveAt: row.lastActiveAt ?? row.createdAt,
+      lastActiveAt: row.lastActiveAt,
       status: this.deps.manager.statusOf(row.sessionId),
       pendingApprovalCount: this.deps.manager.pendingApprovalCount(row.sessionId),
       pendingFollowUpCount: this.deps.manager.pendingFollowUpCount(row.sessionId),
@@ -357,6 +357,7 @@ export class SessionService {
       session.sessionId,
       isSessionMeta(metaMsg) ? (asSessionSource(metaMsg.payload.source) ?? null) : null,
     );
+    const createdAt = new Date().toISOString();
     const row: SessionRow = {
       sessionId: session.sessionId,
       projectId: args.projectId,
@@ -369,7 +370,9 @@ export class SessionService {
       // Everything created through this service is the Web App's (schedule runs included);
       // "cli" is stamped only by Trace adoption. NULL means a legacy row, treated as web.
       client: "web",
-      createdAt: new Date().toISOString(),
+      // Creation is the first activity; the first driven run advances it (see SessionManager.drive).
+      lastActiveAt: createdAt,
+      createdAt,
     };
     this.deps.sessions.insert(row);
     this.deps.manager.adopt(row, session);
@@ -438,6 +441,7 @@ export class SessionService {
     if (facts.provider === null || facts.modelId === null) return null;
     // Registration already narrowed the origin; record it in the registry (single source of truth).
     this.deps.sources.set(sessionId, facts.source);
+    const createdAt = sessionIdCreatedAt(sessionId) ?? facts.firstTs ?? new Date().toISOString();
     const row: SessionRow = {
       sessionId,
       projectId,
@@ -452,7 +456,11 @@ export class SessionService {
       // (web-only) excludes these rows; the "show CLI sessions" preference includes them.
       client: "cli",
       hasTrace: true,
-      createdAt: sessionIdCreatedAt(sessionId) ?? facts.firstTs ?? new Date().toISOString(),
+      createdAt,
+      // The CLI's own activity leaves no mark on this row (this server drives none of its
+      // runs), so an adopted Session reads as last-active at its creation time until it is
+      // resumed here; the Trace tail is not consulted (see SessionRow.lastActiveAt).
+      lastActiveAt: createdAt,
     };
     // Idempotent backfill: concurrent list calls may discover the same Session for the first time simultaneously (consistent with AgentsRepo's convention).
     this.deps.sessions.insertOrIgnore(row);
--- a/packages/server/src/services/session-service.ts
+++ b/packages/server/src/services/session-service.ts
@@ -82,6 +82,7 @@ export class SessionService {
       ...(row.title !== null ? { title: row.title } : {}),
       ...(source !== undefined ? { source } : {}),
       createdAt: row.createdAt,
+      lastActiveAt: row.lastActiveAt ?? row.createdAt,
       status: this.deps.manager.statusOf(row.sessionId),
       pendingApprovalCount: this.deps.manager.pendingApprovalCount(row.sessionId),
       pendingFollowUpCount: this.deps.manager.pendingFollowUpCount(row.sessionId),
--- a/packages/server/test/db-upgrade.test.ts
+++ b/packages/server/test/db-upgrade.test.ts
@@ -8,8 +8,11 @@ import { afterEach, beforeEach, describe, expect, it } from "vitest";
 import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
+import type { DatabaseSync } from "node:sqlite";
 import { openDatabase } from "../src/db/database.js";
+import { SCHEMA_SQL } from "../src/db/schema.js";
 import { SessionsRepo } from "../src/db/repos/sessions.js";
+import type { SessionRow } from "../src/db/repos/sessions.js";
 
 const sqlite = process.getBuiltinModule("node:sqlite");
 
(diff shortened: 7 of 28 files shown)


packages/server/src/runtime/session-manager.ts
```diff
@@ -1371,6 +1375,9 @@ export class SessionManager {
       // now-empty state.
       entry.pendingSteering = [];
       entry.lastActivityMs = Date.now();
+      // Run-end stamp (see the run-start counterpart at the top of drive). A no-op UPDATE
+      // when the row is mid-deletion — the DELETE already won.
+      this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
```
hiyouga: **Unprotected DB write in `drive()`'s `finally`, ahead of the idle publish.**

Every other DB write reachable from `drive()` is exception-proofed: `recorder.record` is wrapped in try/catch (lines 1345-1350) and `errors.record` swallows internally (`error-recorder.ts:118-143`, "never rethrow"). This new call is the first one that can throw out of the `finally`, and it sits *before* `publishState(entry, "idle")`, `titles?.maybeGenerate`, and `startQueuedFollowUp`.

If it throws, `entry.status` has already been set to `"idle"` in memory but the state is never broadcast, the title is never generated, and every queued follow-up is silently dropped — the UI shows the Session running forever until a manual reload.

Reachable triggers:
- `index.ts:179-186` does `await deps.manager.shutdown(5000)` (which races a 5s timeout) and then `deps.db.close()`. A run that outlives the cap reaches this line against a closed handle. Verified on Node v24.18.0: `db.prepare(...)` after `db.close()` throws `ERR_INVALID_STATE: database is not open`.
- `openDatabase` sets only `journal_mode` and `foreign_keys` (`database.ts:24-25`) — no `busy_timeout` — so a contended write surfaces `SQLITE_BUSY` directly instead of retrying.

```suggestion
      try {
        this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
      } catch (err) {
        this.log(
          `[session] last_active_at stamp failed: ${err instanceof Error ? err.message : String(err)}`,
        );
      }
```

packages/server/src/runtime/session-manager.ts
```diff
@@ -1194,8 +1194,12 @@ export class SessionManager {
   ): Promise<void> {
     // Every driven run (task, goal round, compaction) writes Trace lines: flip the row's
     // has_trace cache here, the single choke point, so listing can serve it from the DB
-    // without a directory walk (see SessionService.listSessions).
+    // without a directory walk (see SessionService.listSessions). The same choke point
+    // stamps last_active_at — once per run boundary (here and in the finally below), never
+    // per streamed message, so activity during the run (steering, approvals, follow-up
+    // queueing) is covered by the run-end stamp without any hot-loop writes.
     this.deps.sessions.markHasTrace(entry.sessionId);
+    this.deps.sessions.touchLastActive(entry.sessionId, new Date().toISOString());
```
hiyouga: **Run-start stamp is outside the `try`, so a throw here wedges the Session permanently.**

`drive()`'s `try` begins several lines below. A throw on this line rejects the promise stored in `entry.running` before the `finally` ever runs, so `entry.status` is never reset from `"running"`/`"compacting"` and `entry.abort` is never cleared. From then on every `startTask`/`startCompact` on that Session 409s via `assertIdle`, and only an idle-sweep eviction or a restart recovers it. The same closed-DB / `SQLITE_BUSY` triggers as the run-end stamp apply.

`markHasTrace` on line 1201 already has this shape, so this isn't a new class of bug — but the PR doubles the exposure without guarding either. Wrapping both writes in one try/catch (or moving them inside the `try`) fixes both at once.

packages/server/src/db/repos/sessions.ts
```diff
@@ -91,6 +101,7 @@ export class SessionsRepo {
         row.title,
         row.client ?? null,
         row.hasTrace ? 1 : 0,
+        row.lastActiveAt ?? row.createdAt,
```
hiyouga: **CLI-adopted rows get a `lastActiveAt` that is pinned to `createdAt` forever.**

`adoptTraceSession` (`session-service.ts:441-459`) inserts `client: "cli"` rows with `createdAt: sessionIdCreatedAt(sessionId) ?? facts.firstTs ?? new Date().toISOString()` and no `lastActiveAt`, so this default applies. Those Sessions are never driven by this server, so `SessionManager.drive()` — the only writer — never fires for them, and CLI runs write no `usage_records` into `web.db`, so the open-time backfill also lands on `created_at`.

Net effect: a CLI Session that the user works in daily reports a `lastActiveAt` equal to its creation time, permanently. When PR #277 switches the sidebar to sort by this field, the most actively used CLI Sessions sort to the very bottom.

The data to fix it is already loaded on that path: `discoverTraces`/`traceIndex.repo.listFilesByAgent` returns a per-Session shard `date` (used by `sessionStats` at `session-service.ts:263-265`). Adoption/hydration could refresh `last_active_at` from the newest shard date the same way it already backfills `has_trace` (`session-service.ts:190-195`).

packages/server/src/runtime/session-manager.ts
```diff
@@ -1194,8 +1194,12 @@ export class SessionManager {
   ): Promise<void> {
     // Every driven run (task, goal round, compaction) writes Trace lines: flip the row's
     // has_trace cache here, the single choke point, so listing can serve it from the DB
-    // without a directory walk (see SessionService.listSessions).
+    // without a directory walk (see SessionService.listSessions). The same choke point
```
hiyouga: **"goal round" is not a run boundary — a whole goal is a single `drive()` call.**

`startGoal` builds one `goalStream` and passes it to one `drive()` (lines 646-655), and its own docstring says so: *"one `session.run(input, { goal })` call loops rounds until a terminal state, and the Session stays `running` for the whole goal (every round)"* (lines 575-583).

So the two stamps bracket the entire goal, not each round. A goal that runs for hours across dozens of rounds keeps `last_active_at` frozen at its start instant the whole time — exactly the case where a "last activity" column is most expected to move. The comment (inherited from `markHasTrace`, where per-run granularity doesn't matter because the flag is monotonic) now overstates what the new field guarantees.

Either stamp on the `goal_round` boundary the drive loop already observes, or reword the comment and the `SessionInfo.lastActiveAt` doc so #277 doesn't build on the wrong contract.

packages/server/src/db/database.ts
```diff
@@ -29,8 +29,20 @@ export function openDatabase(dbPath: string): DatabaseSync {
   // schema.ts; drop entries only in a release allowed to break existing web.db files.
   ensureColumn(db, "sessions", "client", "TEXT");
   ensureColumn(db, "sessions", "has_trace", "INTEGER NOT NULL DEFAULT 0");
+  ensureColumn(db, "sessions", "last_active_at", "TEXT");
   ensureColumn(db, "auth_sessions", "via", "TEXT");
   ensureColumn(db, "trace_files", "page_stats", "TEXT");
+  // last_active_at backfill for rows formed before the column existed: the session's most
+  // recent request timestamp (usage_records, served by idx_usage_session — the closest
+  // persisted proxy for "last Trace activity"), else created_at. Guarded by IS NULL so it
```
hiyouga: **The backfill can never see a subagent Session's own activity.**

`usage-recorder.ts:79-88` writes `sessionId: ctx.sessionId` — the **parent** — for every request, putting the child only in `origin_session_id`:

```ts
const originSessionId =
  msg.origin && msg.origin.length > 0 ? msg.origin[msg.origin.length - 1]! : null;
const base = { ts: now.toISOString(), ..., sessionId: ctx.sessionId, originSessionId, ... };
```

So `WHERE usage_records.session_id = sessions.session_id` matches zero rows for every subagent row, and they all silently fall through to `created_at`, while the parent absorbs all of its children's timestamps. It degrades gracefully (a subagent's `created_at` is its registration instant, roughly when it started), but the value is systematically the subagent's *start*, never its end — and the same skew persists after the migration, since `drive()` only stamps the parent entry.

packages/server/src/db/repos/sessions.ts
```diff
@@ -100,6 +111,13 @@ export class SessionsRepo {
     this.db.prepare("UPDATE sessions SET has_trace = 1 WHERE session_id = ?").run(sessionId);
   }
 
+  /** Bump last_active_at (drive's run start/end — one write per run boundary, never per streamed message). */
+  touchLastActive(sessionId: string, at: string): void {
+    this.db
+      .prepare("UPDATE sessions SET last_active_at = ? WHERE session_id = ?")
```
hiyouga: **Two UPDATEs against the same row at the same instant.**

`drive()` calls `markHasTrace(sessionId)` and then `touchLastActive(sessionId, now)` back to back (lines 1201-1202) — two `prepare()` calls, two statement executions, two WAL frames, on one row, with no transaction around them. A single statement does the same job and removes the window where one lands and the other doesn't:

```ts
/** Run start: flip the has_trace cache and stamp last_active_at in one write. */
markRunStart(sessionId: string, at: string): void {
  this.db
    .prepare("UPDATE sessions SET has_trace = 1, last_active_at = ? WHERE session_id = ?")
    .run(at, sessionId);
}
```

(`touchLastActive` is still needed on its own for the run-end stamp.)

packages/server/src/services/session-service.ts
```diff
@@ -82,6 +82,7 @@ export class SessionService {
       ...(row.title !== null ? { title: row.title } : {}),
       ...(source !== undefined ? { source } : {}),
       createdAt: row.createdAt,
+      lastActiveAt: row.lastActiveAt ?? row.createdAt,
```
hiyouga: **Dead fallback — `mapRow` already coalesced.**

`row` here always comes from `SessionsRepo`, and `mapRow` sets `lastActiveAt: (r.last_active_at as string | null) ?? (r.created_at as string)` (`sessions.ts:56`), so `row.lastActiveAt` is never `undefined` and the `?? row.createdAt` branch is unreachable.

Counting them up, the PR now has four independent "fall back to createdAt" layers for one column: the SQL `COALESCE(..., created_at)` in the backfill, `row.lastActiveAt ?? row.createdAt` in `insert`, the same in `insertOrIgnore`, the coalesce in `mapRow` — and this one. Resolving it once at the repo boundary and letting this read be `lastActiveAt: row.lastActiveAt` keeps a single place to change if the semantics ever move off `createdAt`.

packages/server/src/db/repos/sessions.ts
```diff
@@ -29,6 +29,12 @@ export interface SessionRow {
   client?: "web" | "cli" | null;
   /** Cache: a Trace record exists (set at task start / adoption / subagent registration; backfilled by list hydration). */
   hasTrace?: boolean;
+  /**
+   * Last meaningful activity, ISO (bumped at each driven run's start and end — task, goal
+   * round, compaction; see SessionManager.drive). Always present on rows read from the DB
+   * (openDatabase backfills legacy rows once); omitting on insert defaults to createdAt.
+   */
+  lastActiveAt?: string;
```
hiyouga: **`lastActiveAt?: string` is optional even though it is always populated on read.**

The doc comment right above says it — *"Always present on rows read from the DB"* — but the type says otherwise, so every read site has to re-launder it: `session-service.ts:85` uses `??`, the new tests use `!` (`session-manager.test.ts:172`, `:180`). The optionality exists only to let `insert`/`insertOrIgnore` omit it.

That's `SessionRow` doing double duty as both the read shape and the insert shape. Making it required and giving the two insert methods a `Omit<SessionRow, "lastActiveAt"> & { lastActiveAt?: string }` input (or just having callers pass `createdAt`) removes the `!`/`??` noise at every consumer. Same latent issue as `hasTrace?: boolean`, so it's consistent with the file — just worth not deepening.

packages/server/src/db/database.ts
```diff
@@ -29,8 +29,20 @@ export function openDatabase(dbPath: string): DatabaseSync {
...
+  // is idempotent — a backfilled database matches 0 rows, and a crash between the ALTER
+  // above and this UPDATE heals on the next open.
+  db.exec(
```
hiyouga: **DDL and a data backfill, inline in the open path, with no version marker.**

`openDatabase` now runs `PRAGMA`s, `CREATE TABLE`s, five `ensureColumn` ALTERs, and a DML backfill in sequence, with `IS NULL` standing in for "have I already run?". That predicate happens to be sound here — but it only works because the column is new and nothing else writes NULL. The next backfill (recomputing an existing column, splitting one column into two, normalizing a format) will have no such natural marker, and the file has no `PRAGMA user_version` or migrations table to fall back on. Every one added this way becomes a statement that runs on every open forever.

Cost is not the issue today — measured on a synthetic 2,000-Session / 200,000-usage-row database (Node v24.18.0), the one-time backfill is ~80 ms and the steady-state no-op re-scan is ~0.19 ms. This is maintenance altitude: gating the whole ALTER+backfill block behind `PRAGMA user_version` would give the list a real "already applied" marker and a defined removal point, which is also what the standing note on line 29 ("drop entries only in a release allowed to break existing web.db files") is asking for.

packages/server/test/session-manager.test.ts
```diff
@@ -160,6 +160,28 @@ describe("session-manager", () => {
     });
   });
 
+  it("drive stamps last_active_at: created rows start at createdAt, a run advances it, unrelated sessions untouched", async () => {
```
hiyouga: **Test covers only the happy path; the two claims most likely to break are untested.**

This exercises `startTask` → idle and `startCompact` → idle. Neither of the run-end stamp's stated properties is covered:

1. The **abort/error** path. `drive()`'s `finally` runs for aborted and failed runs too, and that's where an unguarded write does the most damage (it precedes `publishState`). A test that aborts mid-run and asserts the Session still reaches `idle` *and* gets stamped would pin both.
2. The comment on lines 1378-1379 asserts *"A no-op UPDATE when the row is mid-deletion — the DELETE already won."* That's true today (`sessions` carries no foreign keys, so the UPDATE matches 0 rows), but nothing tests it, and the delete route only awaits the in-flight run with a 5 s cap (`http/routes/sessions.ts:477-491`) — so the interleaving really does happen.

Also note `session-index.test.ts:106-122` asserts `lastActiveAt === createdAt` at creation only; no test anywhere asserts the field *advances* over an HTTP surface.

packages/server/src/db/schema.ts
```diff
@@ -54,6 +54,7 @@ CREATE TABLE IF NOT EXISTS sessions (
   archived_at   TEXT,                                -- archive time; NULL=not archived (shown by default; archived ones move under "Archived")
   client        TEXT,                                -- creating client: 'web' (created via the Web App) | 'cli' (adopted from a CLI Trace); NULL = legacy row, treated as web
   has_trace     INTEGER NOT NULL DEFAULT 0,          -- cache: a Trace record exists (set at task start / adoption / subagent registration; lazily backfilled by list hydration)
+  last_active_at TEXT,                               -- last meaningful activity (ISO; bumped at each driven run's start and end, initialized to created_at); NULL only before openDatabase's one-time backfill
```
hiyouga: **Design doc drift: the architecture spec still says the Session index row carries *two* DB-owned fields.**

`specs/05-ARCHITECTURE.md:1161` in the design repo reads: 「Session 索引行带两个 DB 专属字段：`client` 标记创建客户端…；`has_trace` 缓存该 Session 是否已有 Trace…」. `last_active_at` is a third field of exactly that kind — DB-owned and surfaced on `SessionInfo` — unlike `source`, which the same sentence explicitly excludes as living in `session_meta`. The count is now wrong.

Precedent: that sentence was added to the spec when `client`/`has_trace` landed (#139), so pairing a new DB-owned column with a spec update is the established practice.

Per `AGENTS.md`: 「修改 AGENTS.md / CLAUDE.md / specs 实际改的都是设计仓库的文件，要在设计仓库（而非实现仓库）提交；实现仓库的 PR 永远不包含这些文件」 — so the fix does **not** belong in this PR. It needs its own `docs/…` branch and PR in `penguin-harness-design`, which doesn't exist yet.

packages/server/src/db/database.ts
```diff
@@ -29,8 +29,20 @@ export function openDatabase(dbPath: string): DatabaseSync {
   // schema.ts; drop entries only in a release allowed to break existing web.db files.
   ensureColumn(db, "sessions", "client", "TEXT");
   ensureColumn(db, "sessions", "has_trace", "INTEGER NOT NULL DEFAULT 0");
+  ensureColumn(db, "sessions", "last_active_at", "TEXT");
```
hiyouga: **Compat code added with no cleanup owner or date.**

`AGENTS.md` (= repo-root `CLAUDE.md`): 「兼容包袱是长期成本，**每次涉及都要提醒用户**：说明该兼容代码要留多久、由谁在何时清理，避免默认永久保留。」 The rule asks for three things; the PR body supplies the first ("must stay until a DB-schema floor is declared") and explicitly defers the other two ("Their removal needs an explicit future decision").

Concretely this line becomes the fifth never-pruned entry on the `ensureColumn` list, alongside `client`, `has_trace`, `via`, and `page_stats`, and its only retirement trigger is a schema floor that nobody has been assigned to declare. The strategy choice itself looks fine and well-documented — this is just the missing who/when half of the reminder.

(To be clear on what is *not* a violation: the absent changelog entry is correct here — 「changelog 用单独的汇总 PR 提交，不塞进各功能 PR」 — though `changelog/unreleased/` is still empty, so the required `YYYY-MM-DD-backward-compatibility.md` for this batch still has to be written in that summary PR.)

packages/server/src/services/session-service.ts
hiyouga: **`PATCH /api/sessions/:id` can return a `lastActiveAt` older than what is in the DB.**

`resolveSession` (`http/routes/sessions.ts:383-403`) does a fresh synchronous `findById` at handler entry, but the handler then awaits twice before reaching this line — `await readJson(c)` (`:420`) and `await deps.sessionService.hasTrace(updated)` (`:462`) — and never re-reads. `updated` is just `{ ...row }` spread from the pre-await snapshot (`:446`), so `row.lastActiveAt` here is whatever was on disk *before* the request body was parsed.

`touchLastActive` is a synchronous SQLite write issued from an async continuation in the same process (`session-manager.ts:1202`, `:1380`), so it lands squarely inside those await windows. A run that starts or finishes while the user toggles archive or renames the Session makes the PATCH response hand back a stale timestamp — and since the web store swaps the whole object (`sessions.tsx:397`, fed by `sidebar.tsx:411`/`:425` and `chat-page.tsx:1060`), a client that trusts the response now holds a value that moved backwards.

Nothing in `packages/web/src` renders the field yet, so this is API-response correctness rather than a visible regression today. Note `title` (a concurrent `TitleGenerator` persist) and `hasTrace` have the same staleness in this handler, so the clean fix is one re-read after the mutations rather than spreading the pre-await row:

```ts
const updated = deps.sessionsRepo.findById(row.sessionId) ?? row;
```

packages/server/src/db/database.ts
```diff
@@ -29,8 +29,20 @@ export function openDatabase(dbPath: string): DatabaseSync {
...
   ensureColumn(db, "auth_sessions", "via", "TEXT");
   ensureColumn(db, "trace_files", "page_stats", "TEXT");
+  // last_active_at backfill for rows formed before the column existed: the session's most
+  // recent request timestamp (usage_records, served by idx_usage_session — the closest
+  // persisted proxy for "last Trace activity"), else created_at. Guarded by IS NULL so it
+  // is idempotent — a backfilled database matches 0 rows, and a crash between the ALTER
+  // above and this UPDATE heals on the next open.
+  db.exec(
+    `UPDATE sessions SET last_active_at = COALESCE(
+       (SELECT MAX(ts) FROM usage_records WHERE usage_records.session_id = sessions.session_id),
```
hiyouga: **A self-healed Session is backfilled from an id that owns no usage rows — silently, and permanently.**

`SessionsRepo.replaceId` rewrites the primary key when the loader produces a new session id (`session-manager.ts:1156-1161`):

```ts
if (session.sessionId !== row.sessionId) {
  this.deps.sessions.replaceId(row.sessionId, session.sessionId);
```

and `replaceId` touches only the `sessions` table (`repos/sessions.ts:178-182`). Grepping `packages/server/src` for `UPDATE usage_records` returns nothing, so the Session's entire request history keeps the **pre-heal** `session_id`.

So for any legacy row that was ever self-healed, this correlated subquery matches zero rows, `COALESCE` falls through to `created_at`, and a Session with months of recorded requests is stamped with its creation time. The `IS NULL` guard then makes that wrong value permanent — no later open re-evaluates it.

Worth either correlating through `origin_session_id`/`trace_sessions` as well, or noting the limitation next to the "closest persisted proxy" claim in the comment above so the gap is at least documented.

packages/server/test/session-index.test.ts
```diff
@@ -103,6 +103,24 @@ describe("session-index", () => {
     expect(list.sessions.map((s) => s.sessionId)).toContain(session.sessionId);
   });
 
+  it("SessionInfo carries lastActiveAt (ISO, = createdAt at creation) through create, list, and single GET", async () => {
+    await configureModels();
+    const res = await api.post(base(), {});
+    expect(res.status).toBe(201);
+    const { session } = (await res.json()) as SessionCreateResponse;
+    expect(session.lastActiveAt).toBe(session.createdAt);
```
hiyouga, line 111: **The new tests would still pass if the feature were deleted.**

Both of the assertions here reduce to `createdAt === createdAt`. Strip the entire server-side feature — the schema column, the `ensureColumn` ALTER, the `openDatabase` backfill, `last_active_at` from both INSERT column lists, `mapRow`'s read of `r.last_active_at`, and both `touchLastActive` calls — and leave only `lastActiveAt: row.lastActiveAt ?? row.createdAt` in `toInfo`. `SELECT *` then returns no `last_active_at` key, the coalesce falls to `created_at`, and every assertion in this test still passes. (That dead fallback, flagged separately, is precisely what makes the test vacuous.) No test anywhere asserts the field *advances* over an HTTP surface.

The unit test has the same problem in a subtler form. `session-manager.test.ts:180`:

```ts
expect(sessions.findById("session-1")!.lastActiveAt! >= afterTask).toBe(true);
```

If `startCompact` never stamped at all, the value is still exactly `afterTask`, and `afterTask >= afterTask` is true — so the comment's claim that "compaction is a driven run too" is unverified. `>=` rather than `>` is forced here because `new Date().toISOString()` is millisecond-resolution and the fake generator completes synchronously. The same test also cannot tell the run-start stamp (`session-manager.ts:1202`) from the run-end stamp (`:1380`): deleting either one leaves every assertion green, even though the design comment turns on having both.

The root cause of the resolution problem is worth fixing directly: `app.ts` threads `overrides.now` into `UsageService` (`:211`), `UpdateCheckService` (`:213`), `UsageRecorder` (`:224`), `ErrorRecorder` (`:225`), `AuthService` (`:273`), `AdminService` (`:281`) and `Scheduler` (`:306`) — but `SessionManagerDeps` has no `now` field, so these two stamps are the only per-run persisted timestamps on the raw clock. Their own backfill proxy (`MAX(usage_records.ts)`) *does* honor the injected clock, so under a stubbed clock the two halves of the same column disagree. Adding `now` to `SessionManagerDeps` would make the stamps injectable and let these tests assert strict advancement.

packages/server/src/runtime/session-manager.ts
hiyouga: **The run-end `touchLastActive` is the first throw-capable DB write ever placed in drive()'s `finally`, and it sits before the idle publish, queued-follow-up auto-start and title generation, so a SQLite error there silently cancels the whole run wrap-up.**

index.ts:179 `await deps.manager.shutdown(5000)` gives up after 5s on a still-streaming run, then `server.close(cb)` runs `deps.db.close()` (index.ts:183). The in-flight drive reaches line 1380 against a closed handle and node:sqlite throws `ERR_INVALID_STATE: database is not open` (reproduced). The throw escapes the finally, so `publishState(entry, "idle")` (1381) never fires (every SSE client stays pinned on 'running'), `titles.maybeGenerate` (1387) never runs, and `startQueuedFollowUp` (1394) — its ONLY call site — never runs, stranding the user's queued follow-ups forever. Same shape for SQLITE_BUSY/FULL/IOERR at any time (no `busy_timeout` is set). The throw also replaces whatever error the catch above just recorded. Note the codebase's own convention: `recorder.record` 30 lines up (1345-1350) is wrapped in try/catch, and `ErrorRecorder.record` swallows internally and never rethrows — pre-PR this finally could not throw at all. Fix: wrap in try/catch, or move it after `publishState`.

packages/server/src/runtime/session-manager.ts
hiyouga: **The run-start stamp sits OUTSIDE drive's `try` (which opens at line 1244), so a throw there skips the `finally` entirely and wedges the session as permanently 'running'.**

A transient DB error at 1202 (closed handle, SQLITE_BUSY/FULL/IOERR) makes `drive()` return a rejected promise before the generator is ever iterated. `launchTask` has already set `entry.status = "running"` and `entry.abort = ac` (763-764) and assigned the rejected promise to `entry.running` (803), and the route has already answered 202. The finally never runs, so status/abort/running are never reset: `assertIdle` rejects every later startTask/startGoal/startCompact on that session with 409 for the life of the process, and `sweepIdle` (1055) never evicts it because it requires `status === "idle" && entry.running === null`. Only deleting the session or restarting the server recovers it; core's `gen` is never iterated nor closed. `markHasTrace` (1201) already had this shape, but the PR doubles the number of unprotected writes in that window instead of moving either inside the try.

packages/server/src/services/session-service.ts
hiyouga: **CLI-adopted rows get `lastActiveAt = createdAt` at adoption and can never advance — the server never drives them, they have no usage_records, and the IS NULL-guarded backfill skips them.**

_Anchored here because the code this is about — session-service.ts:441-458 (adoptTraceSession) — is not part of this diff._

`adoptTraceSession` inserts with no `lastActiveAt`, so `insertOrIgnore` defaults it to `createdAt = sessionIdCreatedAt(sessionId) ?? facts.firstTs` — the minute baked into the session id. `drive()` is only ever entered with a server-driven entry, so `touchLastActive` is never called for `client:'cli'` rows; `usage-recorder` only runs inside this server's drive, so their `MAX(usage_records.ts)` matches nothing; and the open-time backfill's `WHERE last_active_at IS NULL` never revisits a row inserted non-NULL. A CLI session created in January and used from the CLI daily through August reports `lastActiveAt` = January, permanently. Note the asymmetry this creates: the same session gets a real activity timestamp if it happens to be in web.db at upgrade time and was ever driven from the web, but its creation stamp if it is discovered a day later via `includeCli`. The trace index already stores per-shard `date` in `trace_files` and is walked on this very code path (the sibling `has_trace` cache is backfilled there at session-service.ts:191-193), so the correcting signal is already in hand.

packages/server/src/runtime/session-manager.ts
```diff
@@ -1194,8 +1194,12 @@ export class SessionManager {
...
+    // stamps last_active_at — once per run boundary (here and in the finally below), never
+    // per streamed message, so activity during the run (steering, approvals, follow-up
+    // queueing) is covered by the run-end stamp without any hot-loop writes.
```
hiyouga: **Subagent rows have the same hole: `registerChildSession` inserts with no `lastActiveAt` (defaults to registration time) and a subagent is driven inside the PARENT's stream, so `touchLastActive` is never called with the child's id.**

_Anchored here because the code this is about — session-manager.ts:1448-1462 (registerChildSession) — is not part of this diff._

session-manager.ts:1448-1462 does `insertOrIgnore({... createdAt: new Date().toISOString()})` with no `lastActiveAt`. `drive()` is entered once per top-level run with `entry.sessionId` = the parent, so a subagent that runs 20 minutes reports `lastActiveAt` = its spawn second and never advances. The migration cannot repair legacy subagent rows either: `usage-recorder.ts:84` writes `sessionId: ctx.sessionId` (the TOP-LEVEL session) with the child only in `origin_session_id` — schema.ts:67 documents exactly this — so `MAX(ts) WHERE session_id = <child>` matches nothing and the backfill falls to `created_at`. `subagent` is a first-class sidebar category with its own folder and counts, so any activity sort over that bucket degenerates to creation order, and the failure is invisible in review because the column is populated and non-NULL — it just never moves.

packages/server/src/db/database.ts
hiyouga: **`idx_usage_session` is `ON usage_records(session_id)` only, not `(session_id, ts)`, so the correlated `MAX(ts)` is an index SEARCH plus one random table lookup per usage row — the one-time migration stalls startup before the HTTP listener binds.**

EXPLAIN QUERY PLAN gives `SCAN sessions` / `CORRELATED SCALAR SUBQUERY` / `SEARCH usage_records USING INDEX idx_usage_session (session_id=?)` — a SEARCH, not `USING COVERING INDEX`, so each candidate usage row costs a B-tree lookup into the table to read `ts`. Measured on Node 24 / SQLite 3.53.1: 500 sessions / 200k usage rows = 148ms; 2000 sessions / 1M usage rows = 1531-3652ms. `openDatabase` is called synchronously from `buildAppDeps` (app.ts:153), which completes before `serve()` (index.ts:132), so on the first launch after upgrade a heavy user's server accepts no connections for 1.5-3.7s with no log line explaining why. Extending the index to `(session_id, ts)` turns the plan into `USING COVERING INDEX` and the same workloads into 9ms and 24ms. A `GROUP BY`/`UPDATE ... FROM` rewrite does NOT help (measured 1695ms at 1M rows) — the covering index is the fix. The comment at line 36 says the subquery is 'served by idx_usage_session', which is true but reads as if it were cheap.

packages/server/src/db/database.ts
```diff
@@ -29,8 +29,20 @@ export function openDatabase(dbPath: string): DatabaseSync {
...
+       created_at
+     ) WHERE last_active_at IS NULL`,
```
hiyouga: **The backfill UPDATE runs unconditionally on EVERY open, not just when the ALTER actually added the column, so a full scan of `sessions` is paid forever to match zero rows — and `db.exec` discards the `changes` count that would let the migration report what it did.**

On an already-migrated DB the statement matches 0 rows but SQLite still full-scans `sessions` to prove it (`SCAN sessions`; no index covers `last_active_at`). Measured openDatabase with vs without the statement, best of 7: 500 sessions 0.61 vs 0.52ms; 5000 sessions 1.37 vs 0.82ms; 50000 sessions 5.01 vs 0.68ms — at 50k rows the useless scan is ~7x the entire rest of openDatabase, forever. The table grows from more than user chats: every discovered CLI trace (session-service.ts:458) and every subagent spawn (session-manager.ts:1448) inserts a row and nothing prunes them. `ensureColumn` already knows whether it ALTERed — returning that boolean and gating the UPDATE on it makes the migration genuinely one-time, and wrapping the ALTER+UPDATE in a single BEGIN/COMMIT preserves the comment's stated crash-safety for free (I verified `ALTER TABLE ADD COLUMN` rolls back with its transaction). Separately, `prepare(...).run()` returns `{changes}` and is otherwise identical here (no params either way), so moving off `db.exec` would allow a 'backfilled N legacy rows' log line — exactly the diagnostic you want given how many row classes the MAX(ts) proxy silently misses. (Checked and NOT claiming a read-only/full-disk boot regression: `PRAGMA journal_mode = WAL` at line 24 already throws on a read-only file, so that path fails before this statement either way.)

packages/server/src/db/repos/sessions.ts
hiyouga: **`touchLastActive` assigns unconditionally, so `last_active_at` is not monotone: a backward clock step can move it below its previous value or below `created_at`, and nothing ever repairs it.**

Both stamps are raw `new Date().toISOString()` reads of a non-monotonic wall clock (session-manager.ts:1202, :1380) and no clamp exists anywhere (grep for `MAX(last_active` / `Math.max` returns nothing). An NTP correction, a laptop resuming from suspend, or a VM snapshot restore between two runs writes an earlier ISO string than the row already holds — and a backward step between 1202 and 1380 makes a single run's END stamp earlier than its own START stamp. Because the value is a plain string compared lexically by any future sort, the session jumps backwards in a 'recently active' ordering with no way to detect it, and the IS NULL-guarded backfill will never touch it again. `UPDATE sessions SET last_active_at = MAX(COALESCE(last_active_at, ''), ?)` makes the invariant structural at no cost.

packages/server/src/db/schema.ts
hiyouga: **`last_active_at TEXT` is nullable with no DEFAULT and mapRow's `?? created_at` silently masks any NULL, so a row that acquires a NULL after startup loses its real activity time permanently instead of failing loudly.**

The only thing preventing NULLs is JS-side (`row.lastActiveAt ?? row.createdAt` in insert/insertOrIgnore). Concretely: a user downgrades to a pre-PR server binary against the already-upgraded web.db — its INSERT omits `last_active_at`, producing NULL rows. On the next start of the new binary the one-shot backfill fills them with `MAX(usage ts)`/`created_at`, permanently overwriting whatever the real activity time was, with no log line and no error. The same happens for any future INSERT path that forgets the column. SQLite cannot add a DEFAULT to an existing column, so the guarantee has to live in code — `COALESCE(?, created_at)` inside the INSERT (rather than a JS `??`) would at least put it where the DB enforces it once.

packages/server/src/services/session-service.ts
```diff
@@ -82,6 +82,7 @@ export class SessionService {
...
       status: this.deps.manager.statusOf(row.sessionId),
```
hiyouga, line 86: **PATCH /api/sessions/:id answers from a row snapshot taken before two awaits, so it can return a REGRESSED `lastActiveAt` — the first field in that response path the server mutates concurrently.**

_Anchored here because the code this is about — http/routes/sessions.ts:419-464 (PATCH handler) — is not part of this diff._

routes/sessions.ts:419 `const row = resolveSession(c)` is a fresh synchronous `findById`, but `await readJson(c)` (420) and `await deps.sessionService.hasTrace(updated)` (462) both yield, and the response is built from `updated = { ...row }` (446), which never re-reads `last_active_at`. A rename/archive issued while the session is running lets drive's start or end stamp land inside those await windows, and `toInfo` then emits the pre-stamp value. On the client, `packages/web/src/state/sessions.tsx:397` `replace()` swaps the WHOLE SessionInfo object into the store (called from sidebar.tsx:411/425 and chat-page.tsx:1060), so the cached value rolls backwards. The clean fix is one re-read (`deps.sessionsRepo.findById(row.sessionId)`) after the mutations instead of spreading the pre-await row — that also fixes the same latent staleness for `title` (a concurrent TitleGenerator persist) and `hasTrace`.

packages/server/src/api/types.ts
```diff
@@ -746,6 +746,13 @@ export interface SessionInfo {
   /** Session source (for list badges/folders), derived from core session_meta — the single source of truth (not stored in the DB); unset for user-created sessions. */
   source?: SessionSource;
   createdAt: string;
+  /**
+   * Last meaningful activity (ISO 8601, same convention as createdAt): initialized to
+   * createdAt at creation, stamped by the server at each driven run's start and end
```
hiyouga: **The doc promises stamping at 'each driven run's start and end (task, goal round, compaction)', but a goal run is ONE `drive()` covering every round, so an actively working session reports a stale timestamp for the whole run.**

`goalStream` (session-manager.ts:667-683) is a single async generator wrapping one `entry.session.run(..., {goal:{budget}})`, and `startGoal` hands it to a single `drive()` at :655. A 40-round, 3-hour goal therefore produces exactly two stamps: T0 and T+3h. Between them `lastActiveAt` reads 3 hours old while the session is streaming, so any sort-by-recency consumer ranks a still-running session below one that finished 5 minutes ago; the same holds for any long single Task. If the process is SIGKILLed at round 39 the run-end stamp never lands and the value stays frozen at T0. The same wording is mirrored at repos/sessions.ts:33-35, repos/sessions.ts:114, schema.ts:57 and session-manager.ts:1197-1200 — five copies of a policy statement that is wrong the same way in all five, which is itself a maintenance trap.

packages/server/src/runtime/session-manager.ts
```diff
@@ -1194,8 +1194,12 @@ export class SessionManager {
...
     this.deps.sessions.markHasTrace(entry.sessionId);
```
hiyouga: **`markHasTrace` and `touchLastActive` are two separate prepared statements updating the same row of the same table at the same instant, where `UPDATE sessions SET has_trace = 1, last_active_at = ? WHERE session_id = ?` would do.**

node:sqlite autocommits per `.run()`, so this is two write transactions, two WAL commit records, two fsyncs and two `prepare()` compiles per run start. Measured with synchronous=FULL, median of 3 rounds x 1500 iterations: 12.4us for the pair vs 5.9us for the single equivalent UPDATE. The absolute cost is noise at any plausible drive rate (goal rounds do NOT multiply it — they share one drive), but merging is a one-line change that also halves the number of throw sites in the unprotected run-start prefix flagged above, and it matters on a network-backed filesystem where an fsync is milliseconds. More structurally, 'the single choke point' is being expressed as a growing pile of independent repo calls rather than one run-boundary persistence operation — the next per-run row fact will be a third.

packages/server/src/db/repos/sessions.ts
```diff
@@ -45,6 +51,9 @@ function mapRow(r: Record<string, unknown>): SessionRow {
     archivedAt: (r.archived_at as string | null) ?? null,
     client: (r.client as "web" | "cli" | null) ?? null,
     hasTrace: (r.has_trace as number) === 1,
+    // The open-time backfill leaves no NULLs; the coalesce only hardens against a row
+    // somehow inserted as NULL (degrades to createdAt instead of surfacing undefined).
+    lastActiveAt: (r.last_active_at as string | null) ?? (r.created_at as string),
```
hiyouga, line 59: **'Default to createdAt' is now encoded four times (SQL COALESCE, mapRow, both insert paths, session-service.ts:85), and `SessionRow.lastActiveAt?: string` is optional on a read model where mapRow always populates it — so every reader re-supplies a fallback that can never fire.**

Given the unconditional backfill, mapRow's coalesce (line 56) is unreachable and its own comment says so; given mapRow, `session-service.ts:85`'s `?? row.createdAt` is dead for DB-read rows. It stays load-bearing only because `createSession` (session-service.ts:360-373) and `adoptTraceSession` (:441-456) build SessionRow literals WITHOUT the field and hand them straight to `toInfo` — so the create/adopt API response computes its own default independently of what `insert` actually stored. Today both defaults are `createdAt` so they agree by coincidence; the moment the insert default changes (e.g. `now()` for adopted rows) the response silently disagrees with the DB and no test catches it, because every test asserts `lastActiveAt === createdAt` on both sides. Making the field required on the read type (with a separate insert-input type) collapses all read-side fallbacks to zero. Contrast the sibling `hasTrace?: boolean`, where `undefined` and `false` are interchangeable — here `undefined` and `createdAt` are different answers.

packages/server/test/db-upgrade.test.ts
```diff
@@ -92,8 +94,95 @@ describe("openDatabase column upgrade", () => {
...
+    );
+    insertSession.run("session-used", "2026-01-01T00:00:00.000Z");
+    insertSession.run("session-quiet", "2026-01-02T00:00:00.000Z");
+    const insertUsage = old.prepare(
+      `INSERT INTO usage_records (ts, date, project_id, agent_id, session_id, provider, model_id, cache_read, cache_write, output, total)
+       VALUES (?, ?, 'p1', 'a1', 'session-used', 'custom', 'm1', 0, 0, 1, 1)`,
+    );
+    insertUsage.run("2026-02-01T10:00:00.000Z", "2026-02-01");
+    insertUsage.run("2026-02-03T09:30:00.000Z", "2026-02-03");
+    old.close();
```
hiyouga: **The new test adds three more real-file WAL database opens to a file whose `afterEach` uses a plain `rm` with no `maxRetries` — the exact ci-windows EBUSY cascade the repo's own test helper documents and guards against.**

packages/server/test/helpers.ts:86-90 spells it out: 'A plain rm was the top cause of ci-windows cascades (cleanup throws -> hook timeout -> "database is not open" in later files)' and uses `{ recursive: true, force: true, maxRetries: 10, retryDelay: 100 }`. db-upgrade.test.ts:21-23 uses `rm(dir, { recursive: true, force: true })` — `force` swallows only ENOENT, and Node's EBUSY/EPERM/ENOTEMPTY retry count defaults to 0. `openDatabase` sets `PRAGMA journal_mode = WAL` (database.ts:24), so each of the new test's opens leaves web.db-wal/-shm to be released; on ci-windows (ci.yml runs `pnpm test` over packages/server/test) a zero-retry rm can lose that race even on a fully passing run. Compounding it, the seed handle is closed by a bare `old.close();` at line 150 rather than a `finally`, so any throw in the setup leaks an open handle into the same teardown. Fix: add `maxRetries: 10, retryDelay: 100` to the afterEach.

packages/server/src/db/repos/sessions.ts
```diff
@@ -100,6 +111,13 @@ export class SessionsRepo {
     this.db.prepare("UPDATE sessions SET has_trace = 1 WHERE session_id = ?").run(sessionId);
   }
 
+  /** Bump last_active_at (drive's run start/end — one write per run boundary, never per streamed message). */
```
hiyouga: **Both new stamps hardcode `new Date().toISOString()`, bypassing the app-wide injectable clock that every other timestamp writer in the server receives.**

`BuildDepsOverrides.now?: () => Date` (app.ts:148) is threaded into UsageService (:211), UpdateCheckService (:213-214), UsageRecorder (:224), ErrorRecorder (:225), AuthService (:273), AdminService (:281) and Scheduler (:306); `new SessionManager({...})` (app.ts:229-239) passes none and `SessionManagerDeps` has no such field. So `last_active_at` is the only per-run PERSISTED timestamp on the real wall clock — while its own backfill proxy (`MAX(usage_records.ts)`, written through the injected clock at usage-recorder.ts:80) presupposes the two share one. Under a stubbed test clock the same run writes a fake `usage_records.ts` and a real `sessions.last_active_at`. It is also why the new test can only assert a range (`expect(afterTask >= before).toBe(true)`, session-manager.test.ts:174) rather than an exact instant — a clock-dependent assertion with no injectable seam available to fix it.

packages/server/src/db/schema.ts
hiyouga: **Design-spec drift: specs/05-ARCHITECTURE.md:1161 states the Session index row carries 'two DB-only fields' (client, has_trace); this PR makes it three, and it is also the first such column to carry a data backfill rather than a bare ALTER.**

The stale sentence is specs/05-ARCHITECTURE.md:1161: 'Session 索引行带两个 DB 专属字段：`client` 标记创建客户端…；`has_trace` 缓存该 Session 是否已有 Trace…'. Precedent confirms the pairing is established practice — the design repo added that exact sentence when client/has_trace landed in #139. Per the repo-root CLAUDE.md (-> design-repo AGENTS.md:16) the fix must NOT ride this PR ('实现仓库的 PR 永远不包含这些文件'); it needs a separate design-repo PR on a `docs/…` branch, and the design repo currently has no such branch. Left as-is, the doc that is 实现的依据 misdescribes the sessions index. Related, lower confidence: AGENTS.md:40 requires the compat reminder to state 留多久 / 由谁在何时清理 — the PR body answers the first ('until a DB-schema floor is declared') but explicitly defers the other two. (The absence of a changelog entry is NOT a violation: AGENTS.md:53 puts changelog in a separate aggregate PR, matching actual practice on main.)

packages/server/src/runtime/session-manager.ts
hiyouga: **The run-end `touchLastActive` is the first throw-capable DB write ever placed in drive()'s `finally`, and it sits before the idle publish, queued-follow-up auto-start and title generation, so a SQLite error there silently cancels the whole run wrap-up.**

index.ts:179 `await deps.manager.shutdown(5000)` gives up after 5s on a still-streaming run, then `server.close(cb)` runs `deps.db.close()` (index.ts:183). The in-flight drive reaches line 1380 against a closed handle and node:sqlite throws `ERR_INVALID_STATE: database is not open` (reproduced). The throw escapes the finally, so `publishState(entry, "idle")` (1381) never fires (every SSE client stays pinned on 'running'), `titles.maybeGenerate` (1387) never runs, and `startQueuedFollowUp` (1394) — its ONLY call site — never runs, stranding the user's queued follow-ups forever. Same shape for SQLITE_BUSY/FULL/IOERR at any time (no `busy_timeout` is set). The throw also replaces whatever error the catch above just recorded. Note the codebase's own convention: `recorder.record` 30 lines up (1345-1350) is wrapped in try/catch, and `ErrorRecorder.record` swallows internally and never rethrows — pre-PR this finally could not throw at all. The trigger also exists inside the test suite: `session-manager.test.ts:127-130` does `afterEach(() => { channels.dispose(); db.close(); })` with no `manager.shutdown()` await, so any test that leaves a drive settling now has a new way to throw from its finally. Fix: wrap in try/catch, or move it after `publishState`.

packages/server/src/runtime/session-manager.ts
hiyouga: **The run-start stamp sits OUTSIDE drive's `try` (which opens at line 1244), so a throw there skips the `finally` entirely and wedges the session as permanently 'running'.**

A transient DB error at 1202 (closed handle, SQLITE_BUSY/FULL/IOERR) makes `drive()` return a rejected promise before the generator is ever iterated. `launchTask` has already set `entry.status = "running"` and `entry.abort = ac` (763-764) and assigned the rejected promise to `entry.running` (803), and the route has already answered 202. The finally never runs, so status/abort/running are never reset: `assertIdle` rejects every later startTask/startGoal/startCompact on that session with 409 for the life of the process, and `sweepIdle` (1055) never evicts it because it requires `status === "idle" && entry.running === null`. Only deleting the session or restarting the server recovers it; core's `gen` is never iterated nor closed. `markHasTrace` (1201) already had this shape, but the PR doubles the number of unprotected writes in that window instead of moving either inside the try.

packages/server/src/services/session-service.ts
hiyouga: **CLI-adopted rows get `lastActiveAt = createdAt` at adoption and can never advance — the server never drives them, they have no usage_records, and the IS NULL-guarded backfill skips them.**

_Anchored here because the code this is about — session-service.ts:441-458 (adoptTraceSession) — is not part of this diff._

`adoptTraceSession` inserts with no `lastActiveAt`, so `insertOrIgnore` defaults it to `createdAt = sessionIdCreatedAt(sessionId) ?? facts.firstTs` — the minute baked into the session id. `drive()` is only ever entered with a server-driven entry, so `touchLastActive` is never called for `client:'cli'` rows; `usage-recorder` only runs inside this server's drive, so their `MAX(usage_records.ts)` matches nothing; and the open-time backfill's `WHERE last_active_at IS NULL` never revisits a row inserted non-NULL. A CLI session created in January and used from the CLI daily through August reports `lastActiveAt` = January, permanently. Note the asymmetry this creates: the same session gets a real activity timestamp if it happens to be in web.db at upgrade time and was ever driven from the web, but its creation stamp if it is discovered a day later via `includeCli`. The trace index already stores per-shard `date` in `trace_files` and is walked on this very code path (the sibling `has_trace` cache is backfilled there at session-service.ts:191-193), so the correcting signal is already in hand.

packages/server/src/runtime/session-manager.ts
hiyouga: **Subagent rows have the same hole: `registerChildSession` inserts with no `lastActiveAt` (defaults to registration time) and a subagent is driven inside the PARENT's stream, so `touchLastActive` is never called with the child's id.**

_Anchored here because the code this is about — session-manager.ts:1448-1462 (registerChildSession) — is not part of this diff._

session-manager.ts:1448-1462 does `insertOrIgnore({... createdAt: new Date().toISOString()})` with no `lastActiveAt`. `drive()` is entered once per top-level run with `entry.sessionId` = the parent, so a subagent that runs 20 minutes reports `lastActiveAt` = its spawn second and never advances. The migration cannot repair legacy subagent rows either: `usage-recorder.ts:84` writes `sessionId: ctx.sessionId` (the TOP-LEVEL session) with the child only in `origin_session_id` — schema.ts:67 documents exactly this — so `MAX(ts) WHERE session_id = <child>` matches nothing and the backfill falls to `created_at`. `subagent` is a first-class sidebar category with its own folder and counts, so any activity sort over that bucket degenerates to creation order, and the failure is invisible in review because the column is populated and non-NULL — it just never moves.

packages/server/src/db/database.ts
hiyouga: **`idx_usage_session` is `ON usage_records(session_id)` only, not `(session_id, ts)`, so the correlated `MAX(ts)` is an index SEARCH plus one random table lookup per usage row — the one-time migration stalls startup before the HTTP listener binds.**

EXPLAIN QUERY PLAN gives `SCAN sessions` / `CORRELATED SCALAR SUBQUERY` / `SEARCH usage_records USING INDEX idx_usage_session (session_id=?)` — a SEARCH, not `USING COVERING INDEX`, so each candidate usage row costs a B-tree lookup into the table to read `ts`. Measured on Node 24 / SQLite 3.53.1: 500 sessions / 200k usage rows = 148ms; 2000 sessions / 1M usage rows = 1531-3652ms. `openDatabase` is called synchronously from `buildAppDeps` (app.ts:153), which completes before `serve()` (index.ts:132), so on the first launch after upgrade a heavy user's server accepts no connections for 1.5-3.7s with no log line explaining why. Extending the index to `(session_id, ts)` turns the plan into `USING COVERING INDEX` and the same workloads into 9ms and 24ms. A `GROUP BY`/`UPDATE ... FROM` rewrite does NOT help (measured 1695ms at 1M rows) — the covering index is the fix. The comment at line 36 says the subquery is 'served by idx_usage_session', which is true but reads as if it were cheap.

packages/server/src/db/database.ts
hiyouga: **The backfill UPDATE runs unconditionally on EVERY open, not just when the ALTER actually added the column, so a full scan of `sessions` is paid forever to match zero rows — and `db.exec` discards the `changes` count that would let the migration report what it did.**

On an already-migrated DB the statement matches 0 rows but SQLite still full-scans `sessions` to prove it (`SCAN sessions`; no index covers `last_active_at`). Measured openDatabase with vs without the statement, best of 7: 500 sessions 0.61 vs 0.52ms; 5000 sessions 1.37 vs 0.82ms; 50000 sessions 5.01 vs 0.68ms — at 50k rows the useless scan is ~7x the entire rest of openDatabase, forever. The table grows from more than user chats: every discovered CLI trace (session-service.ts:458) and every subagent spawn (session-manager.ts:1448) inserts a row and nothing prunes them. `ensureColumn` already knows whether it ALTERed — returning that boolean and gating the UPDATE on it makes the migration genuinely one-time, and wrapping the ALTER+UPDATE in a single BEGIN/COMMIT preserves the comment's stated crash-safety for free (I verified `ALTER TABLE ADD COLUMN` rolls back with its transaction). Separately, `prepare(...).run()` returns `{changes}` and is otherwise identical here (no params either way), so moving off `db.exec` would allow a 'backfilled N legacy rows' log line — exactly the diagnostic you want given how many row classes the MAX(ts) proxy silently misses. (Checked and NOT claiming a read-only/full-disk boot regression: `PRAGMA journal_mode = WAL` at line 24 already throws on a read-only file, so that path fails before this statement either way.)

packages/server/src/db/repos/sessions.ts
hiyouga: **`touchLastActive` assigns unconditionally, so `last_active_at` is not monotone: a backward clock step can move it below its previous value or below `created_at`, and nothing ever repairs it.**

Both stamps are raw `new Date().toISOString()` reads of a non-monotonic wall clock (session-manager.ts:1202, :1380) and no clamp exists anywhere (grep for `MAX(last_active` / `Math.max` returns nothing). An NTP correction, a laptop resuming from suspend, or a VM snapshot restore between two runs writes an earlier ISO string than the row already holds — and a backward step between 1202 and 1380 makes a single run's END stamp earlier than its own START stamp. Because the value is a plain string compared lexically by any future sort, the session jumps backwards in a 'recently active' ordering with no way to detect it, and the IS NULL-guarded backfill will never touch it again. `UPDATE sessions SET last_active_at = MAX(COALESCE(last_active_at, ''), ?)` makes the invariant structural at no cost.

packages/server/src/db/schema.ts
hiyouga: **`last_active_at TEXT` is nullable with no DEFAULT and mapRow's `?? created_at` silently masks any NULL, so a row that acquires a NULL after startup loses its real activity time permanently instead of failing loudly.**

The only thing preventing NULLs is JS-side (`row.lastActiveAt ?? row.createdAt` in insert/insertOrIgnore). Concretely: a user downgrades to a pre-PR server binary against the already-upgraded web.db — its INSERT omits `last_active_at`, producing NULL rows. On the next start of the new binary the one-shot backfill fills them with `MAX(usage ts)`/`created_at`, permanently overwriting whatever the real activity time was, with no log line and no error. The same happens for any future INSERT path that forgets the column. SQLite cannot add a DEFAULT to an existing column, so the guarantee has to live in code — `COALESCE(?, created_at)` inside the INSERT (rather than a JS `??`) would at least put it where the DB enforces it once.

packages/server/src/services/session-service.ts
hiyouga, line 86: **PATCH /api/sessions/:id answers from a row snapshot taken before two awaits, so it can return a REGRESSED `lastActiveAt` — the first field in that response path the server mutates concurrently.**

_Anchored here because the code this is about — http/routes/sessions.ts:419-464 (PATCH handler) — is not part of this diff._

routes/sessions.ts:419 `const row = resolveSession(c)` is a fresh synchronous `findById`, but `await readJson(c)` (420) and `await deps.sessionService.hasTrace(updated)` (462) both yield, and the response is built from `updated = { ...row }` (446), which never re-reads `last_active_at`. A rename/archive issued while the session is running lets drive's start or end stamp land inside those await windows, and `toInfo` then emits the pre-stamp value. On the client, `packages/web/src/state/sessions.tsx:397` `replace()` swaps the WHOLE SessionInfo object into the store (called from sidebar.tsx:411/425 and chat-page.tsx:1060), so the cached value rolls backwards. The clean fix is one re-read (`deps.sessionsRepo.findById(row.sessionId)`) after the mutations instead of spreading the pre-await row — that also fixes the same latent staleness for `title` (a concurrent TitleGenerator persist) and `hasTrace`.

packages/server/src/api/types.ts
hiyouga: **The doc promises stamping at 'each driven run's start and end (task, goal round, compaction)', but a goal run is ONE `drive()` covering every round, so an actively working session reports a stale timestamp for the whole run.**

`goalStream` (session-manager.ts:667-683) is a single async generator wrapping one `entry.session.run(..., {goal:{budget}})`, and `startGoal` hands it to a single `drive()` at :655. A 40-round, 3-hour goal therefore produces exactly two stamps: T0 and T+3h. Between them `lastActiveAt` reads 3 hours old while the session is streaming, so any sort-by-recency consumer ranks a still-running session below one that finished 5 minutes ago; the same holds for any long single Task. If the process is SIGKILLed at round 39 the run-end stamp never lands and the value stays frozen at T0. The same wording is mirrored at repos/sessions.ts:33-35, repos/sessions.ts:114, schema.ts:57 and session-manager.ts:1197-1200 — five copies of a policy statement that is wrong the same way in all five, which is itself a maintenance trap.

packages/server/src/runtime/session-manager.ts
hiyouga: **`markHasTrace` and `touchLastActive` are two separate prepared statements updating the same row of the same table at the same instant, where `UPDATE sessions SET has_trace = 1, last_active_at = ? WHERE session_id = ?` would do.**

node:sqlite autocommits per `.run()`, so this is two write transactions, two WAL commit records, two fsyncs and two `prepare()` compiles per run start. Measured with synchronous=FULL, median of 3 rounds x 1500 iterations: 12.4us for the pair vs 5.9us for the single equivalent UPDATE. The absolute cost is noise at any plausible drive rate (goal rounds do NOT multiply it — they share one drive), but merging is a one-line change that also halves the number of throw sites in the unprotected run-start prefix flagged above, and it matters on a network-backed filesystem where an fsync is milliseconds. More structurally, 'the single choke point' is being expressed as a growing pile of independent repo calls rather than one run-boundary persistence operation — the next per-run row fact will be a third.

packages/server/src/db/repos/sessions.ts
hiyouga, line 59: **'Default to createdAt' is now encoded four times (SQL COALESCE, mapRow, both insert paths, session-service.ts:85), and `SessionRow.lastActiveAt?: string` is optional on a read model where mapRow always populates it — so every reader re-supplies a fallback that can never fire.**

Given the unconditional backfill, mapRow's coalesce (line 56) is unreachable and its own comment says so; given mapRow, `session-service.ts:85`'s `?? row.createdAt` is dead for DB-read rows. It stays load-bearing only because `createSession` (session-service.ts:360-373) and `adoptTraceSession` (:441-456) build SessionRow literals WITHOUT the field and hand them straight to `toInfo` — so the create/adopt API response computes its own default independently of what `insert` actually stored. Today both defaults are `createdAt` so they agree by coincidence; the moment the insert default changes (e.g. `now()` for adopted rows) the response silently disagrees with the DB and no test catches it, because every test asserts `lastActiveAt === createdAt` on both sides. Making the field required on the read type (with a separate insert-input type) collapses all read-side fallbacks to zero. Contrast the sibling `hasTrace?: boolean`, where `undefined` and `false` are interchangeable — here `undefined` and `createdAt` are different answers.

packages/server/test/db-upgrade.test.ts
hiyouga: **The new test adds three more real-file WAL database opens to a file whose `afterEach` uses a plain `rm` with no `maxRetries` — the exact ci-windows EBUSY cascade the repo's own test helper documents and guards against.**

packages/server/test/helpers.ts:86-90 spells it out: 'A plain rm was the top cause of ci-windows cascades (cleanup throws -> hook timeout -> "database is not open" in later files)' and uses `{ recursive: true, force: true, maxRetries: 10, retryDelay: 100 }`. db-upgrade.test.ts:21-23 uses `rm(dir, { recursive: true, force: true })` — `force` swallows only ENOENT, and Node's EBUSY/EPERM/ENOTEMPTY retry count defaults to 0. `openDatabase` sets `PRAGMA journal_mode = WAL` (database.ts:24), so each of the new test's opens leaves web.db-wal/-shm to be released; on ci-windows (ci.yml runs `pnpm test` over packages/server/test) a zero-retry rm can lose that race even on a fully passing run. Compounding it, the seed handle is closed by a bare `old.close();` at line 150 rather than a `finally`, so any throw in the setup leaks an open handle into the same teardown. Fix: add `maxRetries: 10, retryDelay: 100` to the afterEach.

packages/server/src/db/repos/sessions.ts
hiyouga: **Both new stamps hardcode `new Date().toISOString()`, bypassing the app-wide injectable clock that every other timestamp writer in the server receives.**

`BuildDepsOverrides.now?: () => Date` (app.ts:148) is threaded into UsageService (:211), UpdateCheckService (:213-214), UsageRecorder (:224), ErrorRecorder (:225), AuthService (:273), AdminService (:281) and Scheduler (:306); `new SessionManager({...})` (app.ts:229-239) passes none and `SessionManagerDeps` has no such field. So `last_active_at` is the only per-run PERSISTED timestamp on the real wall clock — while its own backfill proxy (`MAX(usage_records.ts)`, written through the injected clock at usage-recorder.ts:80) presupposes the two share one. Under a stubbed test clock the same run writes a fake `usage_records.ts` and a real `sessions.last_active_at`. It is also why the new test can only assert a range (`expect(afterTask >= before).toBe(true)`, session-manager.test.ts:174) rather than an exact instant — a clock-dependent assertion with no injectable seam available to fix it.

packages/server/src/db/schema.ts
hiyouga: **Design-spec drift: specs/05-ARCHITECTURE.md:1161 states the Session index row carries 'two DB-only fields' (client, has_trace); this PR makes it three, and it is also the first such column to carry a data backfill rather than a bare ALTER.**

The stale sentence is specs/05-ARCHITECTURE.md:1161: 'Session 索引行带两个 DB 专属字段：`client` 标记创建客户端…；`has_trace` 缓存该 Session 是否已有 Trace…'. Precedent confirms the pairing is established practice — the design repo added that exact sentence when client/has_trace landed in #139. Per the repo-root CLAUDE.md (-> design-repo AGENTS.md:16) the fix must NOT ride this PR ('实现仓库的 PR 永远不包含这些文件'); it needs a separate design-repo PR on a `docs/…` branch, and the design repo currently has no such branch. Left as-is, the doc that is 实现的依据 misdescribes the sessions index. Related, lower confidence: AGENTS.md:40 requires the compat reminder to state 留多久 / 由谁在何时清理 — the PR body answers the first ('until a DB-schema floor is declared') but explicitly defers the other two. (The absence of a changelog entry is NOT a violation: AGENTS.md:53 puts changelog in a separate aggregate PR, matching actual practice on main.)

packages/server/test/session-manager.test.ts
```diff
@@ -160,6 +160,28 @@ describe("session-manager", () => {
...
+    await waitFor(() => manager.statusOf("session-1") === "idle");
+    const afterTask = sessions.findById("session-1")!.lastActiveAt!;
+    // Stamped with the server clock in the same ISO shape as created_at, past the pre-run instant.
+    expect(afterTask).toMatch(/^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{3}Z$/);
+    expect(afterTask >= before).toBe(true);
+    expect(afterTask > ROW.createdAt).toBe(true);
+    // Compaction is a driven run too: the same choke point advances the stamp again.
+    await manager.startCompact("session-1");
+    await waitFor(() => manager.statusOf("session-1") === "idle");
+    expect(sessions.findById("session-1")!.lastActiveAt! >= afterTask).toBe(true);
```
hiyouga: **No assertion anywhere distinguishes the run-start stamp from the run-end stamp — I independently mutation-proved that DELETING EITHER ONE leaves all 62 files / 597 server tests green.**

I reran the full server suite twice on the PR head: with `session-manager.ts:1380` (the run-end stamp) deleted -> 597/597 pass; the sweep separately found deleting `:1202` (the run-start stamp) is equally invisible. The cause is that `afterTask` is sampled only after the run has already ended, and both checks use `>=`, which an unchanged value satisfies — so line 180's inline comment ('compaction advances the stamp again') is a tautology. The suite proves only 'at least one of the two stamps fires'. Concretely: a later refactor moves the stamp out of drive's `finally` into `launchTask` — a very natural cleanup for the finding that the unguarded write in `finally` can skip `publishState`/`startQueuedFollowUp` — and CI stays green while every session in production starts reporting its run START time. A 40-minute task, a goal loop or a compaction then shows `lastActiveAt` receding as it works, and a session that runs hourly looks progressively less recently active than one that ran once and stopped. Fix: capture the row's value BETWEEN the two stamps (or stub the clock) and assert strict `>` on the run-end transition.

packages/server/test/db-upgrade.test.ts
```diff
@@ -92,8 +94,95 @@ describe("openDatabase column upgrade", () => {
...
+        projectId: "p1",
+        agentId: "a1",
+        provider: "custom",
+        modelId: "m1",
+        workspace: "/w",
+        approvalMode: "allow-all",
+        title: null,
+        createdAt: "2026-04-01T00:00:00.000Z",
+      });
+      expect(repo.findById("session-fresh")!.lastActiveAt).toBe("2026-04-01T00:00:00.000Z");
```
hiyouga: **No test reads the raw `last_active_at` column — every read goes through mapRow's `?? created_at` and toInfo's second `?? row.createdAt` — so the insert-time default this PR adds is unverifiable. Mutation-proved: nulling it in both INSERT paths leaves all 597 tests green.**

I replaced `row.lastActiveAt ?? row.createdAt` with `row.lastActiveAt ?? null` in BOTH `insert` and `insertOrIgnore` and reran the full server suite: 597/597 pass — including this very line, whose comment reads 'Insert without lastActiveAt on an upgraded database: defaults to createdAt', plus session-manager.test.ts:167 and session-index.test.ts:111/116/121. The two dead coalesces flagged separately on repos/sessions.ts:56 are not merely redundant; they are what blinds the suite. So someone dropping or mis-binding the new parameter (easy while fixing the insert/insertOrIgnore duplication, or adding the missing `archived_at` to the same statement) ships NULLs with a green CI. Note this got WORSE with the gating fix now in the working tree: because the backfill is now gated to the open that actually ALTERs the column, those NULL rows are no longer swept up at the next restart — they persist permanently, and mapRow silently renders them as `created_at`. Fix: assert on the raw column (`db.prepare('SELECT last_active_at FROM sessions WHERE session_id = ?')`) at least once.

packages/server/test/db-upgrade.test.ts
```diff
@@ -92,8 +94,95 @@ describe("openDatabase column upgrade", () => {
...
+      // makes the backfill one-time, never a reopen-time reset).
+      repo.touchLastActive("session-quiet", "2026-03-01T00:00:00.000Z");
+    } finally {
+      db.close();
+    }
+
+    const reopened = openDatabase(dbPath);
+    try {
+      const repo = new SessionsRepo(reopened);
+      expect(repo.findById("session-used")!.lastActiveAt).toBe("2026-02-03T09:30:00.000Z");
```
hiyouga, line 180: **Low severity: in the reopened block this assertion is tautological — an unguarded backfill recomputes the identical `MAX(ts)` for `session-used`, so it passes whether or not the `IS NULL` guard exists (deleting the guard fails only line 170, the `session-quiet` row).**

The `touchLastActive` write that actually exercises 'a live value survives a reopen' is applied only to `session-quiet`, whose backfilled value came from the `created_at` FALLBACK branch. So the production-shaped case — a legacy session that HAD run, was backfilled from `MAX(ts)`, then stamped live by a new run, then restarted — is never covered. A future 'heal stale values' variant of the guard (e.g. `WHERE last_active_at IS NULL OR last_active_at < (SELECT MAX(ts) ...)`, an easy-looking way to repair the CLI/subagent rows the two stamps miss) would clobber live run stamps back to the last request timestamp on exactly the rows this test never builds, while `session-quiet` keeps passing because its subquery is NULL. Fix: give `session-used` the live `touchLastActive` write too, or swap which row gets it.

PR #281, opened 2026-08-13
fix(desktop): match the macOS app icon size to other apps

### LINKED ISSUE #278: macos 上图片问题，好大。这个做产品不该犯的错误
Image

Image

moyi-dong:
PR title:

fix(desktop): match the macOS app icon size to other apps

PR body:

## Summary

- Add a dedicated macOS icon asset with the artwork centered in an 824×824 safe area on a 1024×1024 canvas.
- Configure electron-builder to use the macOS-specific icon.
- Keep the existing Windows and Linux icon assets unchanged.
- Add regression coverage for the macOS icon configuration.

## Root cause

The desktop build previously used the same edge-to-edge icon asset across all platforms. Unlike other platforms, macOS does not automatically add enough visual padding around third-party app icons. As a result, the PenguinHarness icon appeared about 25% larger than neighboring icons in the Dock and Launchpad.

The new macOS-specific asset adds the appropriate padding so that the icon matches the visual size of other apps.

### Before

Before

### After

The updated icon was verified using a locally packaged macOS application alongside other apps in the system Applications folder.

After

## Validation

- All 49 desktop unit tests passed.
- Desktop TypeScript type checking passed.
- The full workspace build passed.
- The macOS application was packaged successfully.
- The packaged `.icns` was verified to contain the new 1024×1024 macOS icon asset.
- The Windows icon asset and Linux icon set remain byte-for-byte unchanged.

Fixes #278


pr-assets/issue-278/after-applications.png
Myriad-Dreamin, line 1: Please remove this large file.

moyi-dong: removed.

hiyouga (approved):

PR #285, opened 2026-08-14
feat: add terminal tab support

Myriad-Dreamin:
## Summary

Server-side terminal for PenguinHarness with a Codex-style surface in the web app: shells run in the server and survive everything the browser does — reloads, pane moves, detached windows, multiple viewers.

### Server (`packages/server/src/terminal/`)

- **Pty sessions owned by the server**: every byte feeds a headless xterm, so the daemon always knows the full screen and browser tabs are disposable views.
- **Control plane**: JSON over HTTP (`/api/terminals` — list / create / capture with line ranges / key tokens / kill; create accepts `cwd`, `name`, `cols`/`rows`, `shell`). Unnamed terminals auto-increment per user ("tmp", "tmp 2", …) and carry a stable per-user `seq` for display numbering.
- **Data plane**: binary WebSocket at `/api/terminals/:id/stream` (`[opcode, slot, payload]`); output coalesced (~5 ms window, leading flush) so keystroke echo pays no latency.
- **Attach = ANSI restore stream** (scrollback, colours, cursor, DEC input modes): a reload rebuilds the exact screen while the shell keeps running. Multi-client attach is supported; size ownership goes to the latest claiming connection.
- **Auth**: per-user ownership (foreign ids answer 404), Origin check on the upgrade, exited sessions readable during a grace period.
- Fixes found by the tests: `dispose()` used to no-op its own `kill()` (shells outlived server shutdown), and SIGHUP escalates to SIGKILL (`nohup`-started servers inherit ignored HUP).

### Web — dock (multi-pane)

- **Panes**: up to one per edge (top/bottom/left/right). `` Ctrl+` `` hides/restores the whole arrangement; panes, sizes, tab order, assignments and each pane's shown terminal all persist.
- **Tabs**: one strip per pane with stable `N:` numbering and live OSC titles (`user@host` stripped for display). Drag sideways to reorder; drag OUT of the strip to move the terminal to another edge — the pane there is created on demand (edge bands at 45 %, drop-target widget, and a preview showing the exact final region). Tabs squeeze browser-style when crowded (number stays readable) and the strip wheel-scrolls. Kill × is a hover overlay; killing the last tab closes the pane, never respawns.
- **No reconnects, ever, for layout changes**: xterm views live in a pool keyed by terminal id (portal + DOM handoff into pane bodies). Moving a terminal between panes, moving a pane to another edge, or splitting a tab out keeps the same xterm DOM node and WebSocket — pinned by e2e.
- **Resize**: the pane/main boundary is a drag handle (ratio-persisted per edge, same-orientation moves carry the ratio, double-click resets, ergonomic clamps).
- **Detach**: per-terminal handoff to a `/terminal?id=…` window; the dock stays open on the remaining tabs.
- **Persistence rules**: opening the dock lands on the last-opened live shell, else the newest live one (created anywhere), else a fresh shell.

### Web — standalone `/terminal` page

Full-window terminal, deep-linkable: `?id=` attaches an existing terminal (the detach handoff), `?cwd=`/`?name=` parameterize a new one; the attached id is written back into the URL so reload/copy-paste reattaches.

### Web — chat toolbar panel switcher

Icon-only triggers for the pinned panels (agents + workspace by default) plus a 创建 menu that opens on hover, listing 智能体面板 / 终端 / 工作区 with per-panel pin toggles. The terminal entry shows a live shell-count badge (on its own icon when pinned, on 创建 otherwise; optimistic on create/kill), and hovering the pinned terminal icon lists the live terminals for one-click focus.

### Terminal UX details

- Clipboard: copy via Ctrl+Shift+C / Ctrl+Insert / Ctrl+C-with-selection; paste rides the browser's native paste event (Ctrl+V / Ctrl+Shift+V / Shift+Insert — no clipboard-read permission). Right-click copies the selection or pastes (page menu suppressed); middle-click pastes; both defer to apps that own the mouse (vim, htop). Click anywhere in the view to focus.
- xterm loads lazily on the first terminal render — its UMD bundles touch `self` at import time, which would crash Node contexts reaching the module through the import graph.

## Test plan

- [x] Server: 643 tests green, incl. 49 terminal tests (restore fidelity, WS handshake auth, reattach + input-mode replay, multi-client, size ownership, hostile frames, burst, exit/grace, manager lifecycle, seq/name assignment). Pty-backed suites skip on win32 (headless CI has no console for ConPTY); pure suites run everywhere.
- [x] Web unit: 650 tests green.
- [x] Web e2e (Playwright + mock LLM): 16 terminal specs — reload fidelity, deep links, dock toggle/persistence, detach, tab reorder/titles/numbering, split-into-pane (with same-xterm-node assertion), layout drag with preview-equals-landing check, boundary resize with ratio persistence, clipboard/right-click/focus, badges and hover menus, tab squeeze + wheel scroll.
- [x] Full e2e suite compared against a `git stash` baseline: the 13 failing chat/compaction/llm specs fail identically without this change (pre-existing in this environment); no new failures introduced.
- [x] Prettier and typecheck clean.

## Notes for review

- Terminals run as the OS account hosting the server and are not Project-scoped; the README documents the trust model (per-user visibility, Origin check, "don't expose to accounts you wouldn't give SSH").
- Closing a pane/dock never kills shells — the only kill paths are the tab ×, the API DELETE, and server shutdown.


hiyouga:
页面刷新后，这个按钮上有个奇怪的框

image

本地测试 macos 上没有 work，启动命令是 `pnpm dev`
image

hiyouga:
mac能用了，但是前端有点小问题，另外可以加个亮色
image

hiyouga:
image
这几个应该共享侧边栏区域，而不是同时打开，否则对话区域太小了

hiyouga:
这个好像没有跟着对话走，应该是对话专属的终端

image

hiyouga (approved):

PR #296, opened 2026-08-14
fix(models): prevent group header overlap at narrow widths

### LINKED ISSUE #294: bug: the model library display problem
Image

模型库的页面宽度减少以后，文字会重叠，需要修复。

moyi-dong:
## Summary

- make model-provider header actions respond to the header's available width instead of the browser viewport
- progressively reveal wide action labels as the content column grows
- add a desktop-width regression test for overflow, provider visibility, and text overlap

The sidebar can reduce the model content column while the viewport still matches desktop breakpoints. The existing viewport-based `sm:` rules therefore kept every action label visible and squeezed the provider name into adjacent controls. This change uses the group header as a size container and replaces only the three affected responsive variants.

## Before

Original issue screenshot (1130px):

Before: model library text overlaps at a reduced width

### 540px — narrow mobile layout

Provider name/count and `新增模型` remain visible. Header/page overflow: 0; text overlaps: 0.

After at 540px: narrow model group headers without overlapping text

### 720px — narrow layout

The mobile shell has more room while the group actions stay compact. Header/page overflow: 0; text overlaps: 0.

After at 720px: model group headers without overlapping text

### 900px — compact desktop header

The expanded desktop sidebar narrows the actual content column; wider actions collapse. Header/page overflow: 0; text overlaps: 0.

After at 900px: compact model group headers without overlapping text

### 1280px — intermediate header

`统一配置 API key` and `测速` progressively appear. Header/page overflow: 0; text overlaps: 0.

After at 1280px: intermediate model group headers without overlapping text

### 1440px — full header

All actions, including `获取 API key`, are visible. Header/page overflow: 0; text overlaps: 0.

After at 1440px: full model group headers without overlapping text

## Validation

- targeted Playwright regression: passed at 900px, 1280px, and 1440px
- additional visual checks: passed at 540px and 720px
- header horizontal overflow: 0 at all five widths
- page horizontal overflow: 0 at all five widths
- visible text overlaps: 0 at all five widths
- provider label remains visible at all five widths
- web typecheck: passed
- web unit tests: 60 files / 720 tests passed
- skills, core, server, and web builds: passed

Fixes #294


hiyouga:
Follow-up per review: narrow group-header rows no longer hide any action — the add-model, bulk-API-key, and speed-test buttons and the get-API-key link now collapse to icon-only controls (text hidden, icon kept, aria-label + title tooltip preserved), with labels reappearing at the @3xl/@4xl row widths, and an unreleased changelog entry was added. The layout e2e now asserts every action stays reachable at all widths with its text hidden on narrow rows (expected regimes are derived from the measured row width, since the container thresholds are rem-based and scale with the environment's root font size). The two other layout.spec failures seen locally ("mobile models" overlap and "mobile chat dropdowns") reproduce identically with origin/main web sources in that environment, so they are pre-existing and unrelated to this PR.

hiyouga:
Brought the branch up to date with `main` (merged, not rebased, so the published commits are untouched) — it was 20 commits behind and conflicting. The only real conflict was `changelog/unreleased/README.md`, where I took main's deletion (#316 dropped the per-release index on purpose) and converted this branch's entry to the new bilingual metadata-block pair (`.md` + `.zh.md`, with `PR`/`Issue` as full URLs).

`models-page.tsx` auto-merged: #325's changes there are client-type semantics (`openai` → `openai-chat`, `canonicalClientType`) and do not touch the group-header actions, so both sides survive — the merged file differs from `main` by exactly this PR's responsive delta. #325 added no new header actions, so nothing else needed the icon-only treatment.

Verified after the merge: `packages/web` unit tests 790/790 pass, prettier clean, and the `layout.spec.mjs` sweep for this PR passes (every group action reachable at 390–1920px, icon-only with an accessible name on narrow rows, no header overlap). One unrelated failure is pre-existing on `main`: `tsc` reports `setTurnThinkingLevel` is not defined in `chat-page.tsx` (from #321) — it reproduces with an unmodified `main` tree, so it is not from this PR, but it will need a fix on `main`.

PR #297, opened 2026-08-14
feat(web): fork sessions from completed replies

### LINKED ISSUE #293: todo: fork session
支持从某个消息开始分叉，放到模型消息下面的一个fork按钮

moyi-dong:
Closes #293.

## Summary

- add a **Fork** action beside Copy for completed root assistant replies
- submit an immutable Trace coordinate and validate that it is a final, completed, resumable Task boundary
- clone retained Trace shards and recursively snapshot the source scratchpad under a new Session id
- rewrite only recognized local `[attached image: ...]` / `[attached file: ...]` paths; keep HTTP images, user-shaped marker text, and inline vision data unchanged
- reject running/compacting sources, preserve access checks, and create an ordinary active Session with the same Agent/model/Workspace/approval mode
- name direct forks with the language-neutral suffix `Source title (1)`, `Source title (2)`, and so on, using one persistent atomic counter shared across every reply position in the source Session
- do not reuse fork numbers after deletion; concurrent forks receive distinct numbers and legacy databases upgrade automatically
- document the API, UI behavior, review checklist, evidence, and remaining verification

Behavior and boundary checklist: [`packages/docs/decisions/issue-293-session-fork.md`](https://github.com/moyi-dong/penguin-harness/blob/d647995/packages/docs/decisions/issue-293-session-fork.md)

## Validation

- `pnpm typecheck`
- `pnpm test` — all workspace packages passed
- final server suite: 63 files / 616 tests passed
- focused fork integration suite: 11 scenarios passed, including cross-position numbering, deletion, no-title fallback, four concurrent forks, and committed-row cleanup after a late response failure
- focused stream-model suite: 96 tests passed
- docs: 32 tests and typecheck passed
- browser flow: forked a deterministic two-turn Session from the first reply, then the second reply, then the first reply again; observed `(1)`, `(2)`, `(3)` with no locale-specific title text
- attachment browser flow: confirmed the later turn was excluded when expected, then deleted the source and confirmed the copied image still rendered

## Browser evidence

Source Session: the short title and both source turns are visible, with the selected reply's fork action.

Source Session fork action

Fork result: `Fork Test (1)` contains only the retained turn and renders the copied file/image.

Forked transcript

After deleting the source: the sidebar contains only `Fork Test (1)` and its copied image still renders.

Fork survives source deletion

Repeated forks from different reply positions share the same source counter; all three suffixes are visible in the sidebar.

Numbered forks

## Review notes

- Forking intentionally duplicates the retained scratchpad and therefore its disk usage.
- The UI does not add parent/child lineage; both entries are ordinary independent Sessions.
- The checklist keeps deeper provider-backed follow-up and injected filesystem-failure scenarios visible as non-blocking follow-up verification items.


hiyouga:
Applied the review feedback in 66d4cd3b:

- **Button order** — the fork action now sits immediately to the **right of the copy button** in the reply footer.
- **Confirmation popup** — clicking fork opens the shared `ConfirmModal` first (it explains that the conversation is duplicated into a new chat up to that reply, Confirm / Cancel); the fork request only fires on Confirm. The plain copy button stays an unconfirmed one-click action, and the new dialog strings are in both locale catalogs.
- **`session_not_found` on fork deletion** — fixed on the requesting side. Root cause: `navigate()` runs inside `React.startTransition`, so the sidebar's sync-lane `remove()` committed one render before the route change, and in that intermediate commit the chat page's deep-link probe re-fetched the just-deleted session id (hence the 404 on every delete-while-open, which is the standard try-then-discard fork flow). The removal is now scheduled in the same transition lane, so the list and the route update in one commit and no request fires.

Verification: new Playwright spec `packages/web/e2e/session-fork.spec.mjs` asserts the button order, the confirm/cancel gate (cancel fires no request), the copied transcript, and that deleting the open fork produces **no** follow-up request for the deleted id and no 4xx — the server request log stays free of `session_not_found` for the whole fork/delete flow (reproduced against the previous build, gone after the fix). `penguin-web` typecheck + 722 unit tests and `penguin-docs` tests pass; changelog entry and web-app docs (en/zh) updated.

Note: `e2e/chat.spec.mjs` currently fails at the sidebar group-collapse assertion (line ~330) on this branch **unchanged** — verified pre-existing at the PR head via `git stash` on a fresh data root, unrelated to this follow-up.

hiyouga:
Fixed in b6fe3e85 — there was a second requester, and my previous fix had also stopped working.

**Why the first fix died.** It sequenced the sidebar's list removal against the router transition using `startTransition`. Main's zustand migration (#295) then landed underneath it: a zustand store update is not subject to React's transition lanes, so the list pruned one commit before the route changed again, and the chat page's deep-link lookup re-fetched the just-deleted Session. Scheduling was never a sound mechanism here, so it is no longer the mechanism.

**The fix.** The Session list tombstones the ids it deletes, and the chat page's lookup consults the tombstones before issuing anything — an id this client deleted can only 404, so no request goes out at all. It is treated as an already-failed probe, so the redirect to another conversation still fires. The tombstone set is bounded and deliberately survives a Project switch (session ids are never reused). The inert `startTransition` and its now-false comment are reverted.

**Why it showed up in the Cost Center.** `/api/sessions/:sessionId` carries no `:projectId`, so the recorder attributes these errors to no Project, and unattributed records are admin-only. That is also why my earlier member-scoped check read zero while you saw a new row after every fork deletion — the Cost Center only *renders* the records, it does not cause them.

**Reproduced, not reasoned.** Driving the real app: `DELETE /api/sessions/<fork>` → 204, then `GET /api/sessions/<fork>` → 404 ~19ms later, one `[http] session_not_found` row.

**Test.** `packages/web/e2e/session-fork.spec.mjs` now asserts your criterion directly — the error-record count taken before and after the delete must be unchanged (admin context, so unattributed records are actually visible). Verified non-vacuous: it fails with `before=0, after=1` against the unfixed build and passes with the fix, with zero 404s in the server log.

Also merged current `main` in (only conflict was the deleted `changelog/unreleased/README.md`, accepted) and converted this branch's changelog entry to the bilingual metadata-block format with a `.zh.md` pair. web 761 + server 658 + docs tests, both typechecks, and prettier all pass.

hiyouga (approved):

PR #302, opened 2026-08-16
fix(web): distinguish running and completed chats

### LINKED ISSUE #291: more chat status
运行中和运行完毕应该有不同的前端提示，现在不是很明显

moyi-dong:
## Summary

- replace the ambiguous pulsing green running dot with a neutral gray spinner
- preserve a visible completion outcome with a transient green dot in both the chat header and sidebar
- keep compacting amber and leave ordinary, acknowledged idle Sessions unmarked

Closes #291.

## Design and state model

| State | Indicator | Lifecycle |
| --- | --- | --- |
| Running | Gray spinning ring + `Running` in the header | Driven by the authoritative `task_state` stream |
| Compacting | Amber dot in the sidebar + the existing compaction banner | Existing treatment retained |
| Recently completed | Solid green dot + `Done` in the header | Added only when an observed running/compacting Session returns to idle |
| Seen / ordinary idle | No indicator | Completion is acknowledged when the Session is reopened; it is also cleared by a new run or Project switch |

The green marker is deliberately not derived from `idle`. Treating every idle Session as completed would fill historical conversations with permanent success dots after a reload. Instead, the client keeps a small in-memory set of observed active-to-idle transitions. This gives completion enough persistence to be noticed without turning the sidebar into a status archive.

The shared glyph exposes localized accessible status labels. Step-level tool and thinking icons are unchanged; this visual language applies only to Session-level attention state.

### Running — neutral gray spinner

Running state

### Compacting — existing amber treatment

Compacting state

### Recently completed — solid green dot

Completed state

### Seen / ordinary idle — no indicator

Seen idle state

## Validation

- `pnpm --dir packages/web exec vitest run test/session-activity.test.ts`
- `pnpm --filter @prismshadow/penguin-web test` — 61 files / 724 tests passed
- `pnpm --filter @prismshadow/penguin-web typecheck`
- `pnpm --filter @prismshadow/penguin-web build`
- `pnpm format:check`
- manually verified all four states against the local mock model with no browser console errors


hiyouga:
Applied the review feedback: session states are now told apart by icon shape, not color alone.

- **Running** — a pulsing gray hourglass (your suggested shape).
- **Completed** — a green checkmark inside a circle (your suggested shape), still the transient marker that clears when the Session is reopened.
- **Compacting** — two amber chevrons collapsing inward (sidebar only; the chat page keeps compaction in its stream banner).

The hourglass and circled check reuse the step-level `StatusIcon` path vocabulary, so the app keeps a single glyph language; the light/dark color pairs also follow `StatusIcon`, with color demoted to a secondary cue. `SessionActivityIcon` now owns the localized label per state (svg `<title>` tooltip + `aria-label`; `role="status"` for live states, `role="img"` for the settled check), shared by the sidebar rows and the chat header — no new strings, the existing Running / Compacting / Done labels cover all three states in both locales.

Tests: `session-activity.test.ts` gained static-markup contract tests — every state renders its own distinct path and carries its label. Web suite 727/727 green, typecheck clean. The unreleased changelog entry was updated to describe the icon shapes.

hiyouga:
Applied in `f6bcfa7c` (plus merge `52d666a7`). This round is the maintainer feedback on the glyphs, and one decision below needs your sign-off.

### The glyph set now

| State | Glyph | Tone |
| --- | --- | --- |
| Running | hourglass, turning | gray |
| Compacting | **same** hourglass, turning | amber |
| Finished, unread | circled check | emerald |
| Finished, read | circled check | muted gray |
| Never ran | *nothing* | — |

- **Rotation.** A 180° flip twice per 3s cycle with a long hold between turns, rather than a continuous spin — an hourglass gets turned over, and at the 12px the sidebar renders it at a spin just smears. Pure CSS `@keyframes`, no JS.
- **`prefers-reduced-motion`.** Followed the convention #328 established: the keyframes only ever set `transform`, so the repo's global `animation: none !important` is a free no-op here — the hourglass stands upright and fully visible rather than disappearing (the login traces in the same stylesheet need an explicit override precisely because they hide themselves in their base state; this one does not). A test pins that the keyframes contain nothing but `transform`.
- **Compacting vs running.** Same shape per your instruction, distinguished by amber — the tone compaction already carries elsewhere in the app — and nothing else. **If you would rather compaction be entirely indistinguishable from running, say so and I will drop the amber**; I kept it because it is the tone compaction already uses and it invents nothing new.
- **Chevron glyph deleted** — nothing else referenced it.
- **Empty sessions** show no glyph, keyed off the server's existing `SessionInfo.hasTrace` ("a Task has been started"), which was already on every list row and unused by the web client. The slot keeps its 12px width so the row does not re-flow when the first run starts.

### Read/unread mechanism — please object if you disagree

I searched the whole repo first: **no notion of read/seen/unread exists anywhere** — no notifications table, no read column on `sessions`, no `SessionInfo` field, nothing client-side. The nearest neighbour is `completion-notify.ts`, which fires an ephemeral desktop notification and persists nothing. So this had to be introduced.

**What I chose:** per-Session seen markers in `localStorage`, keyed per Project as `penguin.sessionSeen.<projectId>`. This follows the precedent `pinned-sessions.ts` already set, which documents client-side persistence for exactly this reason (the API models no such field, and a server route would need schema + API changes).

- A marker records when the user last had a Session open. Unread ⇔ `lastActiveAt` is newer than the marker. Stamped when a row is opened, and again when a run finishes while the user is watching it.
- The marker takes `max(now, lastActiveAt)`, so opening a Session marks it read even if the browser clock lags the server's — a stuck unread badge is the failure people actually notice.
- The first write per Project records a **baseline**, so enabling this does not light up every historical conversation at once.
- Capped at 500 markers per Project (evicting least-recently-seen); a deleted Session's marker is pruned.

**Limits, stated plainly:** it is **per browser, not per user** — the same account on another device starts fresh — and there is **no cross-tab sync**, so a second tab keeps its own view until it reloads (the repo has no `storage` listener anywhere and says that is deliberate).

**The alternative** is the server's existing per-user `ui_prefs` blob (`GET`/`PUT /api/me/prefs`, shallow-merged), which would make read state follow the user across devices. I did not use it: it is a single JSON document per user, so every session open becomes a network write of a growing map, and two tabs writing the same key would clobber each other. **Happy to switch if cross-device read state matters more to you than those costs.**

### Accessibility

Every glyph names its exact state in its accessible name and tooltip in both locales, read/unread included (`Done` / `Done, unread`; `运行完毕` / `运行完毕，未读`), so nothing is reachable only by seeing a colour. Busy states announce as `role="status"`, settled ones as `role="img"`.

Since read/unread is carried by colour alone, I picked the pair by measurement rather than by eye — the two tones differ in **lightness** as well as hue, so the distinction survives colour-blind vision:

| | light | dark |
| --- | --- | --- |
| unread vs background | 5.48:1 | 10.47:1 |
| read vs background | 2.54:1 | 2.66:1 (recessive on purpose) |
| **read vs unread** | **2.16:1** | **3.93:1** |
| read vs unread, simulated deuteranopia | 2.35:1 | 3.55:1 |

Worth knowing: the obvious pairing — the previous `emerald-500` against a `gray-400` — has a **1.00:1** luminance ratio, i.e. identical lightness, and would have been invisible as a distinction to a red-green colour-blind user. That is why unread is `emerald-700` in light mode rather than the codebase's more common `emerald-500`.

### Housekeeping

- Merged `origin/main` (the branch was 16 commits behind). Two conflicts, both resolved on the merits: main moved the Sessions state into a zustand store, and deleted the changelog index. `sessions.tsx` now matches main **exactly** — the transient in-memory completion set is gone, since a durable marker answers the same question and survives a reload.
- Changelog converted to the new bilingual format (`.md` + `.zh.md`, metadata block, no index line).

### Verification

`packages/web`: typecheck clean, **787 tests across 64 files pass**, prettier applied to touched files. Production build checked too — confirming the Tailwind tones survive purging, since they live inside a string literal. Tests updated for the new glyph set, plus a new `session-seen.test.ts` (17 cases) covering the clock-skew path, the baseline seed, the 500-marker cap, and malformed-storage degradation. Per instruction I did not run the e2e suite. No docs changes: the docs never described the status dot.

hiyouga:
## Live status for every row, not just the open one

Applied in `a2b796a`.

### Cause

`setStatus` had exactly one caller: `chat-page.tsx`, for the Session whose stream that tab is subscribed to. So a row's status was live only while it *was* the conversation on screen. Every other row kept whatever the last list fetch returned — jump away from a running Session and its row claims `running` until something reloads the list, and a Session that starts running in the background (a scheduled task, a subagent, another tab, or the one you just left finishing) never flips at all. `sessionActivity()` was fine; it was being fed stale input.

### The shape I chose, and why

A **distinct user-channel event** rather than an optional `sessionId` on `task_state`:

```ts
| { type: "session_state"; sessionId: string; state: SessionStatus; lastActiveAt: string }
```

- The per-Session `task_state` contract is then untouched **by construction** — no consumer has to learn to distinguish "with id" from "without id", and `sse-stream.test.ts` still pins that frame with a strict `toEqual`.
- `task_state` also carries `queued` and the `pendingSteering` mirror, which includes user message text. Those belong to the composer of the conversation being watched, not to a list row, and there is no reason to put them on a broader channel. The new event carries neither (there is a test for that).
- It joins the `session_*` family already used for facts that update a list in place (`session_title`, `session_created`).

`lastActiveAt` rides along because it is what makes a background completion legible. The read/unread split compares the row's `lastActiveAt` against the seen marker, so without a fresh stamp a Session that finished while you were elsewhere would flip to `idle` and settle straight into the *muted, already-read* glyph — the exact case you need to notice. It is read back from the row rather than reconstructed client-side: both flips publish immediately after the write that stamps them (`markDriven` at run start, `touchLastActive` in `drive`'s finally), so it is exactly what a list fetch would return, and no browser clock has to agree with the server's.

### Scoping

`app.ts` resolves the audience and binds the channel, the same split the scheduler's `notify` uses — `SessionManager` holds no membership repos. Audience is the Project's **owner plus its members**, i.e. precisely what `ProjectsRepo.listAccessible` grants. `channels.peek()`, not `get()`: a user who has never opened an event stream has no channel, and conjuring one to buffer badge updates nobody is listening to is waste — their next connection fetches the list anyway. Tests cover a member receiving it, a logged-in stranger with a live channel of their own receiving nothing, and a removed member going quiet.

### Requirements

- **Starts running → hourglass; finishes → completed-unread**, without opening it. Covered.
- **The tab that did not start the task, and the Session just navigated away from** — both are just "not the subscribed Session", which is the whole point of moving this to the user channel.
- **No polling.** One addition in that spirit: a reconnect landing outside the channel's replay buffer already reports `resync_required`, and the list now refetches once *on that event*, so a delivery gap cannot strand a row on an hourglass that never stops.
- **Unknown ids are ignored**, never turned into a row — which is also what filters another Project's Sessions, since the store only ever holds the current Project's rows.

### The open Session now has two writers

You asked me to check this, and there were two real flicker paths:

1. The chat page's sync effect depended on the whole `selected` **row object**. The user channel's `session_state` for that same Session replaces that object, which re-ran the effect and re-asserted whatever `stream.taskState` still held — on two independent EventSource connections with no ordering guarantee, a run that had just ended could be pushed back to `running` until this tab's Session stream caught up. It now keys on `selectedSessionId`, so only a real state change writes.
2. The sidebar drew the open conversation as unread for one tick between the row stamp advancing and the chat page's `noteSessionSeen` effect running. It now never draws the open conversation as unread — the user is looking at it — which is both true and deterministic regardless of effect ordering. (This path existed before this PR, via the `reloadSessions()` on the active→idle edge; it just became easier to hit.)

### Verification

`packages/web` 802 tests / `packages/server` 650 tests / `packages/docs` 53 tests all pass; typecheck clean on both packages; `prettier --check .` clean. New: `packages/web/test/session-status-events.test.ts` (15) for the store's handling of the event, and `packages/server/test/session-state-events.test.ts` (4) for publish scope. No e2e run. No conflict with `main` (upstream's only new commit touches `packages/desktop` and quickstart docs), so nothing was merged in.

### Not changed

`design/specs/` — the spec repo is read-only from here, so proposing the wording instead: the SSE section should gain `session_state` as the user channel's counterpart to the session-scoped `task_state`, noting that the id and the row stamp are what a Session list needs and that `queued` / `pendingSteering` deliberately stay session-scoped. `packages/docs/content/server-api.{en,zh}.md` is updated in this commit; `web-app.{en,zh}.md` has no paragraph describing the sidebar's status glyph at all, so there was nothing there to keep in step — the bilingual changelog entry pair carries the behaviour.

hiyouga:
## Blank row after the hourglass — fixed in `3ad4c2d`

Confirmed and reproduced. `sessionActivity` checks status first, so `running` draws the hourglass whatever `hasTrace` says, and then falls to `if (!hasTrace) return null` once the run settles. `hasTrace` is the server's "a Task has been started" flag carried on the **list row**, and `session_state` was updating only `status` and `lastActiveAt` — so a Session running its *first* task still held the `false` its row was fetched with, and the moment the hourglass stopped the row rendered nothing. My fault, from the live-status work: before it the status never flipped client-side at all.

Fixed at the source, not by weakening the "never ran" rule — a fresh Session must still stay blank. The glyph is drawn from three fields, so the flip now carries all three.

**One rule, both sides.** They are not redundant, and building it turned up why:

- `session_state` gains `hasTrace`, read from the row lookup `publishState` already does — free. It reports **true whenever the state is running or compacting**, rather than the raw column: `markDriven` (which sets `has_trace`) runs inside `drive()`, and `startTask` has already published the first `running` by then. My first attempt sent the raw flag and the test caught it emitting `[false, true]` — i.e. telling a client the Session has never run at the exact moment it visibly is running, which is the same bug for any other API consumer.
- The store treats `hasTrace` as one-way and takes a live status as proof the Session has run. This covers the caller that carries no row fields at all — chat-page's two-argument call, fed by the Session channel's `task_state` — and it also covers a narrow race the event alone cannot: if a list fetch that returned `hasTrace: false` lands *after* the run started, `reload()` replaces the array wholesale and only the settle event is left to fix it.

**Also consolidated** the row glyph into `sessionRowActivity(row, seen, activeSessionId)` — activity + read marker + "the open conversation is never unread" — so the sidebar and the tests share one rule instead of the JSX holding its own copy. That is what makes the sequence testable without a DOM.

**Sequence verified end to end** (`packages/web/test/session-status-events.test.ts`): never-run → blank; starts → hourglass; finishes → completed **unread**; opened → completed read. The same sequence in a tab with a different conversation open. Compaction → its own glyph → settles. A run finishing while the user is viewing it → completed and **read**, no unread flash (that test uses a deliberately stale seen marker, so only the "on screen" rule can keep it out of unread). Six of these fail against the previous behaviour — including the specific `running → idle on a never-run row` regression.

**Housekeeping.** Merged `main` in (`462e8df`, kept separate from the fix): one conflict, an import line in `chat-page.tsx` where #321's toast imports met this branch's activity-icon import — both kept. `packages/web` typechecks clean on the merged tree.

**Verification:** web 840 tests / server 658 tests / docs 53 tests pass, both typechecks clean, `prettier --check .` clean. No e2e run. `server-api.{en,zh}.md` and the bilingual changelog pair updated for the new field.

hiyouga:
## Three states, in `ace2371`

- **Running** — the hourglass, unchanged, still turning.
- **Finished, unread** — the green dot from before the icon round, recovered from this branch's own history (`cf8d08cc^`) rather than redrawn.
- **Finished, read** — nothing at all.

The circled checkmark is gone, along with the `completed` activity state and the `statusCompleted` string in both locales. The chat header now carries the hourglass alone — the conversation on screen is by definition read, so it has no dot to show.

**Compacting keeps the hourglass and keeps its own tone** (amber vs the running hourglass's gray), per the earlier "压缩也用漏斗". The two live states still differ by colour alone and still name themselves in tooltip and accessible name.

**`hasTrace` is still load-bearing — and now there is a test saying so, not a comment.** Read and never-ran render identically, so the field looks redundant. It is not: unread is a *timestamp comparison*, and a conversation created **after** this browser first saw the Project has no read marker of its own and falls back to the baseline — its own creation time is later, so it tests as unread. Without the flag every brand-new row would wear the "go look" dot before it had ever run. The test asserts `isSessionUnread(...) === true` on exactly such a row and then that the glyph is still `null`, so the guard cannot be silently deleted later.

**Colour.** The dot is `emerald-600` light / `emerald-400` dark: **3.77:1** and **10.37:1**, clearing the 3:1 bar for non-text graphics. It is one shade darker in light mode than the `emerald-500` it was originally drawn in, which reaches only **2.57:1** on white — as the only marker left in the list it now has to carry itself. Dark mode keeps the brighter tone. Say the word if you'd rather have the literal original shade back and accept the light-mode contrast.

**Layout and a11y.** Every glyph — hourglass or dot — occupies the same reserved 12px box, so a row never shifts as one appears, changes shape, or disappears; there is a test asserting the rendered box size for all three states. The dot has an accessible name for the unread state; the read state renders nothing and announces nothing, which is correct.

**Transitions re-pinned to the new scheme**, including the settle case from last round: never-run row, running → idle now lands on the **dot** (not the check), and once opened it lands on **nothing**. Same for the tab that did not start the task, compaction, and a run settling under the user's eyes (still read, still no flash).

**Housekeeping.** Merged `main` (`eb8638d`, no conflicts). web 876 tests / server 693 tests pass, both typechecks clean, `prettier --check .` clean, i18n parity green after the string deletion. No doc sentence describes the markers, so only the bilingual changelog pair needed rewriting.

hiyouga (approved):

hiyouga:
Dot colour changed to main's in `d57c3e7`: `bg-emerald-500`, one tone in both themes, taken from main's own `StatusDot` in `packages/web/src/components/layout/sidebar.tsx` (commented "Session status dot") and its twin in the chat header — the same per-row Session marker this replaces, so it keeps the same green. web 876 tests + typecheck + prettier green.

PR #304, opened 2026-08-17
feat: recall queued steering and follow-up messages into the composer

### LINKED ISSUE #287: Support recall the steering message
用户发的插话和排队消息要支持撤回到输入框，允许用户编辑后再重新发

Myriad-Dreamin:
Fixes #287

A mid-run message that is still waiting — an undelivered steering message, or a follow-up queued behind the current run — can now be recalled back into the input box, edited, and resent.

## User-visible changes

- Every queued hint line above the composer (undelivered steering, and a new per-entry list of queued follow-ups) carries a **Recall** (撤回) button. Clicking it withdraws the message server-side and restores its original content into the draft: text lands in front of anything currently typed, images and file attachments come back as composer chips, and a recalled follow-up also restores the per-turn thinking level it was queued with.
- A steering message that already reached the model — or a follow-up that already auto-started — answers `409 not_pending`, surfaced as a localized toast; the queued hint retires on its own via the re-broadcast `task_state`.

## Implementation

- **Core**: `ContextEngine`/`Session.unsteer(input)` withdraws a queued steering entry (matched by input-list identity) and refuses once the queue drained into a request — what makes recall race-safe against delivery.
- **Server**: queued entries get server-assigned ids (`pendingSteering` entries gained `id`; new `pendingFollowUps` field on `task_state` lists each queued follow-up's content next to the existing `queued` count). New endpoints `DELETE /api/sessions/:id/steer/:steerId` and `DELETE /api/sessions/:id/follow-ups/:followUpId` return the withdrawn message's original content `{text, images, files, thinkingLevel?}`. File attachments are not held in memory while queued: they are read back from the Session scratchpad at recall time and their disk copies deleted.
- **Web**: recall buttons on the queued hint lines, draft restoration in the composer, `pendingFollowUps` wiring through the stream controller, new strings (en/zh) including the `not_pending` error mapping. The bare "N queued" count stays as the fallback against an older server.

## Tests & docs

- Core: `unsteer` semantics (withdraw before delivery, refuse after drain/idle).
- Server: recall round-trip for both endpoints — content restoration including scratchpad file read-back and cleanup, double-recall 409, recall-after-delivery leaving the mirror for the stream shift, and the follow-up queue only auto-starting what was not recalled.
- Web: stream-controller coverage for the new `pendingFollowUps` field (both buffering and live paths).
- Docs: server API tables and the web-app page updated in both languages; changelog entry added.


hiyouga:
Heads-up: #321 builds directly on this branch (your commit kept as-is) and adds one review/hardening commit on top — a lost-keystroke fix in the recall merge, releasing a staged goal chip when file attachments are restored, and race/SSE regression tests. Reviewed the implementation against #287; verdict and details are in #321's description. Feel free to either fold that commit into this PR (cherry-pick a3fff24a) or let #321 supersede — whichever you prefer.

PR #317, opened 2026-08-18
docs(skills): remote-claude-code relay, key-stepping, and switch rules

### LINKED ISSUE #307: some problems for remote claude code
现在的有几个问题

- 连接成功后，用户发的信息原话转发给 claude code，本地不需要做任何事情，只负责消息的传递和转发
- 用户如果说“切换到 fable5 max”，意思是指模型切换到 fable 5，思考等级切换到 max
- Up 和 s 一起发时，TUI 还没处理完方向键，s 就作用在原来的 Sonnet 上了。改为分开发送、每步确认。
- claude code 运行结束后出现在会话框里面的是 ai 建议的输入，而不是真实输入的未提交文字，通过 tmux 查看的可能分辨不出来。如果需要输入，直接输入新文本就可以了，这个推荐自然会消失。

hiyouga:
## What & why

Issue #307 reports four problems when the `remote-claude-code` skill drives a Claude Code TUI on the user's behalf: the local agent answering instead of relaying, "model + level" requests like "switch to fable5 max" read as a single unknown model name, batched key sequences (`Up` + `s`) racing the TUI so keys land on the previous selection, and post-run composer text (an AI-rendered suggestion) mistaken for real uncommitted user input.

This PR rewrites/extends the skill's instructions to close all four, bumps the skill to `version: 2` / `updated: 2026-08-18T00:00:00Z`, and pins the new rules with a content test.

### `@prismshadow/penguin-skills`

`skills/remote-claude-code/SKILL.md` (v1 → v2):

- **New section 4 "Relaying a conversation — the user's words go through verbatim"**: once the session is up, the local agent is a pure message pipe — every user message is forwarded to Claude Code word-for-word (no answering, rephrasing, translating, summarizing, or doing any part of the task locally), the turn is waited out, and Claude Code's reply is reported back faithfully. Only session-control requests (model/thinking switches, interrupt, attach/detach/close, connection repair) are executed against the TUI directly; when in doubt, relay verbatim.
- **New §3.4 '"switch to fable5 max" — model and thinking level are two settings'**: a model name plus a level word means switch the model (Fable 5) AND the thinking level (max) — never one unknown model name, never chat text to relay. Executed via `/model` and the build's thinking control, one keystroke at a time, with a final capture confirming both values.
- **New §3.3 "One keystroke at a time — capture between keys"**: batched sequences (`tmux send-keys Up s`, or back-to-back calls with no check) race the TUI, so `s` acts on the previous selection. Rule for every menu/picker/dialog: one key per `send-keys` call, `capture-pane` after it, confirm the expected state change, only then the next key; never fire the next key on faith. The section-3 conversation examples now also split message text and Enter into separate calls with a capture in between.
- **Strengthened §3.2**: input-line text after a completed run is usually a Claude-generated suggested next message, and a `capture-pane` dump cannot reliably tell it from real uncommitted input — never submit it, never report it as the user's unsent draft; typing the new message makes the suggestion disappear on its own.
- Troubleshooting table: three rows added/updated (menu-key race, "fable5 max"-style requests, post-run suggestion). Verification checklist: relay/menu/switch items added. Frontmatter `description` extended with the relay + keystroke contract; `short_description` / `short_description_zh` unchanged.

`test/skills.test.ts`: new content test pinning the four rules and `version >= 2` (relay section wording, two-settings interpretation, one-key-per-call rule, suggestion warning, and §3.3-before-§4 ordering), following the neighboring content-pinning test conventions.

### `changelog/`

- New `changelog/unreleased/2026-08-18-remote-claude-code-driving.md` + one line at the top of `changelog/unreleased/README.md`.

### Docs

No docs pages describe these skill internals — `packages/docs/content/skills.{en,zh}.md` only carry the library table's one-line summary, which stays accurate — so no docs changes.

## Tests

- `pnpm --filter @prismshadow/penguin-skills test`: 22 passed (frontmatter validation — version natural number, `updated` timestamp format, short descriptions — plus the new content test).
- `pnpm --filter @prismshadow/penguin-skills typecheck`: clean.
- Consumer sanity checks (packages not touched, but they read the skill files at runtime): `packages/core/test/builtin-agents.test.ts` 6 passed, `packages/server/test/skills.test.ts` 15 passed (after building `skills` and `core` dists, needed in a fresh worktree regardless of this change).
- Touched files formatted with prettier.

Fixes #307


hiyouga:
## Review

The four rules from #307 are all present and each one is worded as an instruction an agent can actually follow, which is the main thing. Reviewed as an operational procedure rather than prose, though, the new relay contract raises the bar for the rest of the file and several pre-v1 passages have not been brought up to it: one leftover instruction now directly contradicts §4, and the only documented way to deliver a message (`send-keys -l`) provably cannot deliver an arbitrary user message verbatim — I reproduced two corruptions on tmux 3.3a. The other findings are missing branches: what to do when a capture shows the *wrong* state, when the TUI is busy, or when the session is gone. Nothing here is a reason to hold the PR's direction; it is a reason to finish the sections it touches.

### Findings

1. **Leftover v1 instruction contradicts the relay contract** (`packages/skills/skills/remote-claude-code/SKILL.md:69`) — "Tool-permission prompts: answer them through `send-keys` as they appear" tells the local agent to decide, on the user's behalf, whether Claude Code may run a tool. §4 says the agent is "a message pipe — nothing more" and may act alone only on session control. A permission prompt is Claude Code asking *the user*, and it is the highest-stakes thing in the file to answer unilaterally (an approved `rm -rf`, a push, a deploy). This is exactly the leftover-contradiction failure mode: v1's text was fine when the agent was the driver, and is wrong now. Fix: relay the prompt to the user, send the key they pick (one key + capture, §3.3); `--dangerously-skip-permissions` stays the explicit unattended opt-in. Same gap in the other direction: nothing says the agent must not answer a clarifying question Claude Code asks mid-turn.

2. **`send-keys -l` cannot deliver an arbitrary message verbatim** (`SKILL.md:80-86`, and the §3 example at `:61`) — the relay contract promises word-for-word delivery, but the only documented delivery path silently corrupts two common cases. Verified on tmux 3.3a against a pane running `cat`:
   - a newline inside the literal argument is delivered as Return, so `send-keys -l "line one\nline two"` submits `line one` as its own message and leaves `line two` as a new draft. Any pasted or multi-line user message is split into several Claude Code turns.
   - a `;` that ends an argument is consumed by tmux's own command-sequence parser: `send-keys -l "ends with semi;"` arrived as `ends with semi`. The relay drops characters and neither the agent nor the user can see it happened.

   `capture-pane` right after would show the damage, but the skill tells the agent to check that the text "sits on the input line", not that it matches the user's message exactly, so the check does not catch it. Fix: document the paste path for anything that is not a single simple line — `tmux load-buffer -` fed by a quoted heredoc (no shell or tmux re-parsing) then `tmux paste-buffer -d -p -t <sess>`, then capture, then Enter. Verified the same way: semicolons, `"`, `'`, `$VAR`, backticks and CJK all arrived intact, and `-d` drops the buffer so the message does not linger in tmux.

3. **§3.3 has no branch for "the expected state was not reached"** (`SKILL.md:103`) — "If the capture doesn't show the change yet, wait a second and capture again — never fire the next key on faith" covers only *late*, never *wrong* (highlight moved the other way, key swallowed, picker never opened) and has no bound, so a stuck picker leaves the agent with no instruction — in practice it improvises more keys into a menu whose state it does not know. Fix: bound the re-capture, then split the two cases (nothing happened → re-send the same key once; wrong row → correct with the opposite key and re-verify), and after that stop and show the user the captured screen. Also worth stating that the agent should capture the picker *before* stepping, locate the current highlight and the target row, and count the moves instead of pressing `Up` blindly.

4. **§3.4 is pinned to its own example** (`SKILL.md:105-109`) — the rule names "switch to fable5 max" and "use opus high" and nothing else, but #307 was filed in Chinese and the phrasing that triggered it ("切换到 fable5 max") does not appear. An agent matching on the examples misses: Chinese forms (`切换到…`, `换成…`, `用 opus 高`), level-only requests (`改成 max`, "set thinking to high"), and model-only requests. There is also no list of what counts as a level word, so the agent has to guess whether `max` names a level or is part of a model name — the exact confusion the issue reports. Fix: generalize to "any message naming a model, a thinking level, or both, in any language", enumerate the level vocabulary, and say plainly that a level-only request changes only the level and a model-only request only the model.

5. **A message arriving mid-run is not covered** (§4) — the contract is deliver → wait → report, with nothing about the user sending something while a turn is running, which is when it happens most. The agent either types into a busy TUI (the text is queued into or mixed with the running turn) or interrupts a run the user wanted to keep. Fix: if the working footer (`esc to interrupt`) is up, hold the message until the turn ends (§3.1) and deliver it then; if the user clearly wants to stop or redirect now ("stop", "停", "别做了"), Escape as its own key, capture to confirm the footer is gone, then deliver. Also worth a warning that Escape twice in a row is not a stronger interrupt — it opens Claude Code's rewind-to-a-previous-message UI.

6. **A dead tmux session or server is never detected** (whole file) — every command assumes the session is alive. When it is not (host rebooted, server killed, someone ran `tmux kill-session`), `capture-pane` fails with `can't find pane: <sess>` or `no server running on <socket>`, exit 1 (verified). The likeliest wrong move is to `tmux new-session` a fresh `claude` and keep relaying into a conversation that has silently lost all context. Fix: treat a failed capture as a dead session — probe with `tmux has-session -t <sess>`, tell the user, and rebuild with `claude --continue` rather than pretending continuity. Deserves a troubleshooting row.

7. **`| tail -30` truncates the answer being relayed back** (`SKILL.md:65`, `:88`, `:98`) — §4 requires reporting what Claude Code said faithfully, but every capture in the file pipes through `tail -N`, so a long answer that scrolled past the tail is cut and the agent reports the fragment as the whole reply. Fix: use `capture-pane -p -S -200` (scrollback) for the report capture, and keep the small `tail -N` for key-state checks.

8. **The "it landed" check is not exact enough to guarantee verbatim** (`SKILL.md:80-88`) — the capture check is "the text sits on the input line". Under §4 the requirement is stronger: the input line must contain *exactly* the user's message, or what gets submitted is not verbatim (finding 2, or a suggestion the new text did not fully displace). Fix: state the check as "your message and nothing else", and give the recovery — do not submit a merged line; clear it (`send-keys C-u`) and send again; if it still merges, tell the user instead of submitting something the user did not write. That also closes §3.2's open end, where "don't clear or edit" leaves no move when the line really is dirty.

9. **The ambiguous middle is unresolved** (`SKILL.md:118`) — "when in doubt, relay verbatim" is the right default but it misroutes the common case of a question aimed at the local agent about the session itself ("is it still running?", "什么情况了", "did it finish?"), which gets relayed to Claude Code instead of answered from a capture. Fix: draw the boundary by subject — questions about the *session* (state, screen, connection, whether the turn is done) are answered locally from a capture; anything about the *work* goes through unchanged; a message that does both is relayed in full and the session half is also answered locally.

10. **Trust-dialog line skips its own capture rule** (`SKILL.md:68`) — "Answer it once via `tmux send-keys -t <sess> Enter`" while §3.3 names the trust dialog as a place where every key needs a capture check. Polish, but it is the kind of inconsistency an agent resolves by picking the more convenient rule.

11. **Not a defect, for the record** — the new test's exact-prose assertions match the house style in this file (`packages/skills/test/skills.test.ts:245`, `:318`), so I am keeping that shape; it does mean any rewording has to move with the test. Frontmatter conventions check out: integer `version` bump 1 → 2 with `updated` moved, matching the other revised skills, and the 471-char `description` is well inside the range already in the library.

- None was supplied for this PR. I will apply findings 1-10 to the skill (1, 2, 3, 5, 6, 8 as behavior-relevant fixes; 4, 7, 9, 10 as scoped wording/coverage fixes), extend the branch's existing changelog entry and its index line, and keep the new test in sync with the reworded sections. Finding 11 needs no change.

hiyouga:
Applied in `0fa02d4b`:

- **Permission prompts go back to the user** (finding 1) — replaced "answer them through `send-keys` as they appear", which contradicted §4, with: carry the prompt and its options to the user, send the key they choose, never approve a tool call on their behalf. Same rule for clarifying questions and plan approvals, as a §4 bullet.
- **Verbatim delivery survives the transport** (finding 2) — §3.2 now sends multi-line / long / punctuation-heavy messages via `load-buffer` + `paste-buffer -d -p` from a quoted heredoc, and documents the two `send-keys -l` corruptions with the tmux 3.3a evidence. The landed check became exact ("your message and nothing else") with named recoveries for a merged line (`C-u`, resend, escalate) and a paste that submitted itself.
- **§3.3 failure branch** (finding 3) — bounded re-capture, then split by what the screen shows: swallowed key → resend once; wrong move → opposite key; still wrong after two corrections → stop and show the user the screen. Plus "capture the picker, count the moves" before stepping.
- **§3.4 generalized** (finding 4) — fires on any message naming a model, a level, or both, in any language; level vocabulary listed (en + zh); level-only, model-only and unmatched-name cases spelled out; the final capture must also show the untouched setting unchanged.
- **§4 gaps** (findings 5, 7, 9) — mid-turn messages are held until the turn ends, or `Escape` once (with the "twice is not stronger, it rewinds" warning); replies are read with `capture-pane -p -S -200` instead of a truncating `tail`; the task/control boundary is drawn by subject, with the both-at-once case handled.
- **Dead session** (finding 6) — a `capture-pane` failing with `can't find pane` / `no server running` is recognized as a dead session, confirmed with `has-session`, and rebuilt with `claude --continue`.
- Finding 10 (trust-dialog capture) folded in. Troubleshooting gained 4 rows, the checklist 1 item and 2 rewrites, and the skills test pins each new rule.

Also updated: the branch's existing changelog entry + its index line, and the `remote-claude-code` row in `skills.en.md` / `skills.zh.md`.

Deliberately not changed:

- **Finding 11** — no change needed; the new test's exact-prose assertions match the house style in `skills.test.ts`.
- **`version` stays 2** — this is the same unreleased change, not a second revision.
- **The Chinese examples in §3.4 / §4** are quoted user input, not prose: the rule has to match on `改成 max` and `停` to fire at all, which an English paraphrase cannot teach. Same reasoning as `humanizer/reference/zh-cues.md`. The changelog entry stays English-only.

Tests: `packages/skills` 22/22, `packages/docs` 32/32 (incl. `skills-sync`), `packages/server` `skills.test.ts` 15/15, skills typecheck clean, prettier applied.

PR #318, opened 2026-08-18
fix(desktop): let a dev run coexist with an installed release build

### LINKED ISSUE #292: dev: multiple desktop
开发时候应该也能运行desktop，和发布版本的运行中的不冲突

hiyouga:
## What & why

Issue #292: during development it should be possible to run the desktop app while a RELEASED/installed build is already running, without the two conflicting.

The unpackaged (source) shell shared the release build's entire identity. Audit of the actual collision points in this codebase:

| Collision point | Status before | Fix |
| --- | --- | --- |
| Electron single-instance lock | Real — the lock is derived from userData, which is keyed on the shared app name `PenguinHarness`; a dev launch while the installed app ran **quit immediately into the release window** (and vice versa) | Dev runs take the app name `PenguinHarness-Dev` → own userData → own lock |
| userData contents (Chromium profile, `preferred-port` / `server-port` / `cli-install-offered`) | Real — shared directory, dev clobbered the release's port memory and profile | Same fix: split userData via the dev app name |
| Data root | Real — launched bare (`pnpm --dir packages/desktop start`, no `PENGUIN_HOME`), the dev shell found the release install's `server.lock` on `~/.penguin/data` and **attached its window to the running RELEASE server** (or, with the release idle, spawned its own server over the release's data) | Unpackaged runs default to `~/.penguin/dev-data` inside the shell itself (the repo's existing dev convention); explicit `PENGUIN_HOME` still wins in both forms |
| Windows AppUserModelID | Real (minor) — dev stamped the installed app's toast/taskbar identity | Dev stamps `com.prismshadow.penguinharness.dev` |
| Embedded server port | Not a real conflict — `PORT=0` allocator + per-userData stickiness + free-port probe (design § "桌面端原型 · 端口与就绪" calls the allocator the only port source); split userData gives each instance its own stable port | No knob added, per the design |
| Protocol handlers / deep links, tray | Do not exist in this app | Nothing |
| Auto-update, CLI-install offer | Already stand down when unpackaged (`updateSupport` → `reason: "dev"`, `cliInstallKind` → `null`) | Nothing |

Dev/release detection uses `app.isPackaged`, the same flag the shell already branches on (`serverEntryPath`, `updateSupport`, `cliInstallKind`). **Release builds are behaviorally unchanged**: same name, AppUserModelID, userData, and shared CLI data root.

## Key changes by package

- `packages/desktop`
  - New `src/app-identity.ts` (pure, unit-testable): `appIdentity(isPackaged)` → name + AppUserModelID per form; `devDataRoot(homedir)` → `~/.penguin/dev-data`.
  - `src/main.ts`: sets the identity from `appIdentity(app.isPackaged)`; `boot()` resolves the data root as `PENGUIN_HOME ?? (packaged ? resolveRoot() : devDataRoot())` and logs a one-line dev-instance banner on unpackaged runs.
  - New `test/app-identity.test.ts`: pins the release identity to `electron-builder.yml`'s `productName`/`appId` (that sync was previously enforced by comment only), asserts the dev identity differs, and covers `devDataRoot`.
- `CONTRIBUTING.md`: `pnpm desktop` added to the daily commands; the data-root paragraph now documents the desktop dev isolation (dev identity, own userData/lock/sticky port, dev-data default without the env var).
- `packages/docs/content/installation.{en,zh}.md`: `pnpm desktop` added to the dev entry point list, plus one sentence on the dev app identity coexisting with an installed build.
- `changelog/unreleased/2026-08-18-desktop-dev-instance.md` + index line in `changelog/unreleased/README.md`.

## Tests & verification

- `pnpm --filter @prismshadow/penguin-desktop typecheck` — pass.
- `pnpm --filter @prismshadow/penguin-desktop test` — 7 files, 52 tests pass (3 new).
- `pnpm --filter @prismshadow/penguin-docs typecheck && … test` — pass.
- `npx prettier --check` on all touched files — pass.
- Live end-to-end under `xvfb-run` (headless Linux) with a sandboxed `$HOME`, simulating the issue's exact scenario — a real `penguin-server` owning the release data root (`server.lock`, port 35911) plus a minimal Electron process holding the release single-instance lock (name `PenguinHarness`):
  - **Base commit**: with the release lock held, the dev shell exited immediately with no output (the #292 bug); without the lock holder, a bare launch attached to the release server (`attaching to the running server at http://localhost:35911`, smoke result `embedded: false`, release login page).
  - **This branch**: the dev shell boots regardless — `[shell] dev instance 'PenguinHarness-Dev' on data root …/.penguin/dev-data`, spawns its own embedded server (port 43849, smoke result `embedded: true`, full web app loaded via desktop-login), leaves the release server and its userData untouched, writes its own `…/.config/PenguinHarness-Dev/preferred-port`, and reclaims that sticky port on relaunch. Graceful shutdown path exercised by the smoke hook.
  - Actual window rendering was exercised through Xvfb (page title `Chat · PenguinHarness` captured in the smoke result); no interactive display was available on this machine.

## Design docs

Not strictly required — the spec (design/specs/06-PROTOTYPE.md § 桌面端原型) describes the released product, whose behavior is unchanged; the dev workflow lives in CONTRIBUTING.md. If a spec note is nevertheless wanted, the proposed exact edit is one bullet appended to § "桌面端原型 · 数据根与实例互斥":

> - 开发运行（未打包）使用独立应用标识 `PenguinHarness-Dev`（独立 userData 与单实例锁）并默认 `~/.penguin/dev-data` 数据根，可与已安装的发布版并行运行；发布版行为不变（#292）。

Fixes #292


hiyouga:
## Review

The mechanism is right, and I verified it rather than read it. On Linux with this repo's own Electron 43.2.0 I reproduced the premise of #292 (two shells under one app name: the second is refused the single-instance lock), confirmed the fix (`PenguinHarness` and `PenguinHarness-Dev` both acquire it concurrently), and booted the real dev shell bare with `PENGUIN_HOME` unset — it lands on `~/.config/PenguinHarness-Dev` and `~/.penguin/dev-data`, starts its own embedded server on an OS-assigned port, logs `[updater] disabled (dev)`, and creates no `~/.config/PenguinHarness` at all. `app.setName()` also buys more than the PR claims: `logs`, `sessionData` and `crashDumps` all hang off userData, so they split too. Release identity is provably untouched. Findings are one test-quality gap, one migration/doc gap, one attribution bug, and a set of things I am reporting rather than fixing because they sit outside #292.

### Findings

1. **The data-root precedence rule is the behavior #292 is about, and nothing tests it** (`packages/desktop/src/main.ts:195`) — `PENGUIN_HOME ?? (isPackaged ? resolveRoot() : devDataRoot())` sits inline in `main.ts`, which imports `electron` and so cannot be unit-tested. `app-identity.test.ts` pins the *constants* (`appIdentity`, `devDataRoot`) but not the *rule*: a later edit that flips the ternary, drops the `PENGUIN_HOME` precedence, or lets a packaged build take `devDataRoot()` keeps the whole suite green. Fix: extract a pure `desktopDataRoot({ envHome, isPackaged, homedir, releaseRoot })` beside `appIdentity`, keeping `releaseRoot` as a thunk so core's `resolveRoot` stays the single definition point of `~/.penguin/data`, and pin all four cases — including the #292 regression guard that an unpackaged run with no env never resolves to the release root.

2. **`fatal()` hard-codes the release name** (`packages/desktop/src/main.ts:72`, `:257`) — `dialog.showErrorBox("PenguinHarness", …)` and `"PenguinHarness failed to start."`. This is the only dialog a dev run can actually raise (the CLI-install and updater dialogs are unreachable when unpackaged), and with a release build running beside it an error box titled "PenguinHarness" cannot be attributed to an instance — exactly the confusion this PR exists to remove. `app.name` is already the marker elsewhere (`menu.ts:35`). One-line fix.

3. **The migration is real and undocumented** (`CONTRIBUTING.md:48`) — two groups move state and neither is told. (a) Anyone running the shell bare (`pnpm --dir packages/desktop start`) was on `~/.penguin/data`, i.e. the release/CLI root; after this lands their sessions are simply not in the window, and nothing states the one-line way back (`PENGUIN_HOME=~/.penguin/data`). (b) *Every* dev run's userData moves, so the Chromium profile goes with it: the window's origin-scoped `localStorage` (theme, language, layout) and the remembered `preferred-port` start fresh once. I do not think a one-time in-app notice is warranted — this is a dev-only path, and the shell already prints `[shell] dev instance '<name>' on data root <root>` on every unpackaged launch, which is the discoverable signal. A sentence in CONTRIBUTING plus a changelog line is proportionate; that is what I will add. Worth recording the other direction as an unadvertised win: dev runs were sharing the *installed app's* Chromium profile and stomping its `preferred-port`, and now stop.

### Reported, deliberately not fixed

4. **A dev run's identity ignores `PENGUIN_HOME`, so two checkouts still collide** (`app-identity.ts:38`) — `appIdentity(false)` is constant, so every unpackaged shell on the machine is `PenguinHarness-Dev` and shares one userData and one instance lock. Two worktrees each running `pnpm desktop` still fight: the second is refused the lock and quits, focusing the *first checkout's* window (`main.ts:214`). A distinct `PENGUIN_HOME` per checkout does not help — it moves the data root but not the identity, so the second shell dies before it reads the env var. Outside #292 (dev-vs-release, not dev-vs-dev) and the fix has design surface a maintainer should choose; the smallest shape would be an opt-in `PENGUIN_DESKTOP_INSTANCE=<label>` → `PenguinHarness-Dev-<label>`.

5. **An empty `PENGUIN_HOME` resolves to an empty data root** (`main.ts:195`, `packages/core/src/state/paths.ts:24`) — `??` treats `PENGUIN_HOME=""` as set, so the root becomes `""` and every `path.join("", …)` is relative: `server.lock`, `web.db` and the projects land in the CWD (the repo). `scripts/run-with-env.mjs:40` implements the opposite, documented `${VAR:-value}` rule, so `pnpm desktop` and `pnpm penguin` normalize empty to the default and only the bare `start` path — the one this PR newly blesses — is exposed. Pre-existing and repo-wide; the owner is core's `resolveRoot`, shared by CLI and server. Patching only the desktop copy would make desktop and core disagree about one variable, so I am leaving it.

6. **The `penguin_session` cookie is port-blind** (`packages/server/src/auth/middleware.ts:17`, `routes/auth.ts:19-21` — name `penguin_session`, `path: "/"`). Cookies are not port-scoped, so `localhost:<dev>` and `localhost:<release>` share one jar. Inside the shells this is now fine (separate Chromium profiles), but a normal browser pointed at both instances still collides: logging into one signs you out of the other. Genuinely outside this PR — the fix is server-side cookie naming/scoping.

7. **Smaller identity leftovers.** The utilityProcess label is `serviceName: "penguin-server"` for both forms (`server-process.ts:124`), so `ps`/Task Manager cannot tell the two servers apart. electron-updater's download cache is keyed on the packaged `appId`, not `app.name`, so `setName` would not have split it — moot only because `initUpdater` returns before touching `autoUpdater` on an unpackaged run (`update-support.ts:18`). And on macOS an unpackaged run has no bundle identifier of its own (it inherits the Electron dev binary's), so anything macOS keys on bundle id is not release identity but is also not dev-specific.

8. **Note, not a defect:** `~/.penguin/dev-data` is also `pnpm dev:server`'s root (`packages/server/package.json:44`, port 7368) and `pnpm penguin`'s. Since `server.lock` is per data root, `pnpm dev:server` followed by `pnpm desktop` now puts the shell into attach mode against the dev backend instead of embedding its own — pre-existing for `pnpm desktop`, and consistent with the documented attach behavior, but the bare `start` form inherits it as of this PR.

### Verified, no action needed

- **The `setName`-before-anything ordering survives the import graph.** ESM evaluates `main.ts`'s imports before its first statement, and `updater.ts:25` eagerly destructures electron-updater's `autoUpdater` *getter*, constructing a platform updater at import time. Probed directly: reading `app.getPath("userData")` before `setName` pins it permanently to `~/.config/@prismshadow/penguin-desktop`, but the updater's construction does not read it, so the relocation still works. This is a live tripwire for anyone later adding a module-scope `app.getPath()` to an imported module, and `app-identity.ts`'s header is the place to say so.
- **The new pinning test is genuinely two-sided.** I mutated each side alone — `appId` in `electron-builder.yml`, `productName` in `electron-builder.yml`, and the release name in `app-identity.ts`. All three fail it. Not a shape assertion.
- **Auto-update and the CLI-install offer stand down.** `updateSupport` returns `{supported:false, reason:"dev"}` on `!isPackaged` (`update-support.ts:18`) before any updater state exists, and `cliInstallKind` returns `null` (`launcher.ts:175`), which also suppresses the menu entry and makes `maybeOfferCliInstall` return before it writes its flag — so no dev run can install a `penguin` shim shadowing the released one.
- **Ports and the server lock separate for free.** `server-port`/`preferred-port` live under userData (`main.ts:151`) and `server.lock` under the data root; the identity and root split already moved both.
- **No other global identity exists to collide.** Swept the repo: no keychain use at all (no keytar, no `safeStorage`, no `security`/libsecret/wincred), no `setAsDefaultProtocolClient`, no `Tray`, no `crashReporter`, no jump list, no login item, no named pipe, no unix socket, no pid file beyond `server.lock`.

### Cross-platform

Everything above was verified on Linux only, on this box. I am not asserting the rest:

- **Windows** — unverified here. The AUMID split is right in principle (Windows routes toasts and groups taskbar buttons by AUMID), and Chromium's `ProcessSingleton` keys its mutex on the user-data path, so the lock should follow userData. `app-identity.ts:24` is honest that dev toasts may not render without an installed shortcut carrying the dev AUMID.
- **macOS** — unverified here. `SingletonLock` is a userData-relative symlink there too, so the same reasoning applies, and the absent dev bundle id (finding 7) only reinforces separation. `menu.ts:35` uses `app.name`, so the app menu really does read `PenguinHarness-Dev` as the module comment claims.

### Design docs

`design/specs/06-PROTOTYPE.md` § "数据根与实例互斥" (line 220) still states flatly that the desktop app defaults to the CLI's data root (`PENGUIN_HOME` or `~/.penguin/data`); this PR adds an exception the spec does not carry. `design/` is a shared read-only checkout here, so I am proposing wording rather than editing — suggested addition under that bullet:

> 未打包（源码）运行取 dev 化身份：应用名 `PenguinHarness-Dev`（独立 userData，因而独立单实例锁与粘滞端口）、Windows AUMID 加 `.dev` 后缀，数据根缺省 `~/.penguin/dev-data`；显式 `PENGUIN_HOME` 仍优先。发布版形态不变，二者可并行运行。

None was given for this PR. I will apply findings 1–3 and leave 4–8 as reported, each noted with its reason in the follow-up comment.

hiyouga:
Applied in 824144e8:

- **Finding 1** — extracted a pure `desktopDataRoot({ envHome, isPackaged, homedir, releaseRoot })` into `app-identity.ts` and wired `main.ts` through it. `releaseRoot` is a thunk (core's `resolveRoot`) so `~/.penguin/data` keeps one definition point. Three new tests pin the rule: explicit `PENGUIN_HOME` wins in both forms and short-circuits core's resolver, a packaged build with no env *delegates* to it, and an unpackaged run with no env takes the dev root and never the release one. Behavior is unchanged — this only moves the rule somewhere it can be tested.
- **Finding 2** — `fatal()` now titles its box with `app.name` and the startup-failure message interpolates it, so a failing dev run is attributable when both instances are up.
- **Finding 3** — CONTRIBUTING.md documents the one-time move (bare `start` was on `~/.penguin/data`; `PENGUIN_HOME=~/.penguin/data` is the way back; the dev userData move resets the window's origin-scoped preferences and remembered port once) and states that the dev identity is one fixed name, so two checkouts still share an instance lock. The changelog entry and its index line carry the same, plus the `logs`/`sessionData`/`crashDumps` split that also falls out of `app.setName`.
- Added a tripwire note to `app-identity.ts`'s header: Electron caches the userData path on first read, and ESM evaluates main.ts's imports before its first statement, so a module-scope `app.getPath(…)` anywhere in the import graph would silently undo the split.

Deliberately not changed, each with its reason:

- **Finding 4** (a dev identity that ignores `PENGUIN_HOME`, so two checkouts collide) — outside #292, which asks for dev-vs-release. The fix needs a naming scheme someone should choose rather than have me invent here.
- **Finding 5** (`PENGUIN_HOME=""` → empty data root) — pre-existing and repo-wide; the owner is core's `resolveRoot`, shared with CLI and server. Patching only the desktop copy would make the two disagree about one variable.
- **Finding 6** (`penguin_session` is port-blind, so an external browser pointed at both instances shares one jar) — server-side cookie naming/scoping, well outside this PR.
- **Finding 7** (`serviceName: "penguin-server"` identical for both; updater cache keyed on `appId` not `app.name`; no dev-specific macOS bundle id) — diagnostic-only, unreachable in dev, and not settable for an unpackaged run respectively.
- **Design docs** — `design/` is read-only from here, so the proposed sentence for `06-PROTOTYPE.md` § "数据根与实例互斥" stands in the review comment above rather than in the diff.

Verification: 55/55 desktop unit tests, `tsc --noEmit` clean, prettier clean on every touched file. Each new test was mutation-checked — flipping the packaged branch, dropping the `PENGUIN_HOME` precedence, and re-deriving the release root instead of delegating each fail it. End-to-end under xvfb on Electron 43.2.0 with a fake `HOME`: the rebuilt shell boots bare onto `~/.penguin/dev-data`, honours an explicit `PENGUIN_HOME`, logs `[updater] disabled (dev)`, and creates no `~/.config/PenguinHarness` in either case. Windows and macOS remain reasoned about, not verified — I have no way to run either here.

PR #319, opened 2026-08-18
feat(web): drag-and-drop file upload onto the chat window

### LINKED ISSUE #311: support image drag upload
支持附件和图片拖到窗口上方上传

hiyouga:
## What & why

Attachments had two entry points — the "+" menu pickers and image paste — and dragging a file onto the page did the worst possible third thing: the browser navigated to the file, replacing the running app. This PR makes drag-and-drop a first-class third entry into the **same** attachment pipeline: same accepted types, same 10MB per-file validation with the same rejection toast, same attachment chips in the composer, same send paths (normal, steering/queue mid-run, handoff/model-switch).

Behavior:

- Dragging OS files anywhere over the window shows a full-window "Drop files to attach" overlay (both locales), on the chat page and the draft page alike — mounted with the composer, so dropping works exactly when there is a composer to attach to, including while a Task is running (attachments then follow the mid-run send mode as usual).
- Releasing attaches the whole batch (multiple files supported): images join the pasted-image pipeline (`image_url` parts; vision-less models keep the scratchpad-path fallback and hint), everything else becomes a file attachment (scratchpad + `[attached file: <path>]` line). Oversize files are refused before being read, with the existing `attachmentTooLarge` toast.
- The overlay hides when the drag leaves the window: a depth counter absorbs the dragenter/dragleave churn of internal element boundaries, and drop/dragend reset it outright so a missed dragleave can never strand a stale overlay.
- Non-file drags never trigger it: text selections, links, dragged page images, and the sidebar's session-reorder drag (its private MIME exists for exactly this reason) all lack the `Files` type.
- Goal mode takes dropped images (they fold into the objective) but refuses files with a toast — the same reason the "+" menu grays its file entry out there; a drop has no disabled affordance, so the refusal is said out loud instead of silently swallowing the files.
- App-shell guard: a file dropped anywhere no drop zone claimed (other pages, loading states) cancels the browser's default navigate-to-file and shows a no-drop cursor — while text/link drags keep their native behavior, so dropping selected text into the composer textarea still inserts it.

Paste-to-attach is untouched (`onPaste`/`addFiles` unchanged) and keeps working exactly as before.

## Key changes

**packages/web**

- `src/lib/file-drop.ts` (new): DOM-free decision logic — `isFileDrag` (drag carries OS files), `nextDragDepth` (the overlay's enter/leave/drop/end depth machine, clamped and self-recovering), `splitDroppedFiles` (images vs. file attachments, drop order preserved), and the app-shell guards `guardWindowDragOver` / `guardWindowDrop` (structural event type, so tests pass plain objects).
- `src/features/chat/drop-zone.tsx` (new): `FileDropZone` — window-level drag listeners + the full-window overlay (portaled to `<body>` like Modal, overlay z tier, `pointer-events-none`, `aria-hidden`; the accessible entry point stays the "+" menu). Claims active file drags (preventDefault + copy cursor) so the shell guard stands aside; the claim and the guard converge on the same outcome in either listener registration order.
- `src/features/chat/chat-input.tsx`: `addDroppedFiles` routes a dropped batch into the existing `addFiles` (images) / `addAttachments` (other files, 10MB cap + toast) intakes, with the goal-mode file refusal toast; renders `<FileDropZone>`; header doc updated.
- `src/app.tsx`: installs the window-level guard (`dragover`/`drop`) for the whole app shell.
- `src/lib/strings.ts` + `src/lib/strings-en.ts`: `dropFilesTitle`, `dropFilesDesc`, `dropFilesGoalHint` (zh + en).
- `test/file-drop.test.ts` (new): 12 tests over the pure logic — file-drag detection (incl. the sidebar's private MIME and null dataTransfer), depth pairing/clamping/reset, MIME-based batch splitting (incl. files with no MIME type), and both guards' claimed/unclaimed/non-file cases.

**packages/docs**

- `content/web-app.en.md` / `content/web-app.zh.md`: new bullet in "Input and Shortcuts" describing drag-and-drop, its validation parity, mid-run and goal-mode behavior, and the navigate-away prevention.

**changelog**

- `changelog/unreleased/2026-08-18-web-drag-drop-upload.md` + index line in `changelog/unreleased/README.md`.

## Tests

- `pnpm --filter @prismshadow/penguin-web test`: 61 files, 733 tests passed (includes the new `file-drop.test.ts`).
- `pnpm --filter @prismshadow/penguin-web typecheck`: clean.
- `pnpm --filter @prismshadow/penguin-docs test` / `typecheck`: clean.
- Note: the fresh worktree needed `pnpm -r build` once so vitest could resolve the `@prismshadow/penguin-core` workspace dist — pre-existing setup, unrelated to this change.

## Design docs

`design/specs/06-PROTOTYPE.md` covers the composer's paste/"+"-menu attachment pipeline (输入区 bullet) but does not mention drag-and-drop; the only drag interactions specified are the sidebar's session reorder and panel-width dragging. Proposed spec addition (not applied — `design/` is read-only for this task):

- **File**: `design/specs/06-PROTOTYPE.md`, section「对话页」→「输入区」bullet — insert after the sentence ending "…模型按路径读取（见 05「读图工具（多模态）」）；" and before "每次发送发起一个 Task；":

  > 支持把文件**拖拽到窗口**添加附件：文件拖入窗口期间全窗显示放置提示浮层（拖离窗口即隐藏；文本选区、页面内拖动等非文件拖动不触发），松开即把整批文件按既有管线附加到输入框——图片等同粘贴图片、其余文件等同上传附件，类型与大小校验及报错口径一致；任务运行中照常附加（随运行中发送模式送出）；目标模式下仅图片可附，拖入文件给出提示、不添加；未被放置区接管的文件拖放在应用任意位置一律取消浏览器默认的「打开文件」导航（文本拖放保持原生行为）；

Fixes #311


hiyouga:
## Review

The feature is well built and the decision logic is properly factored into a DOM-free, unit-tested module — validation parity with the "+" menu, the goal-mode refusal, multi-file batching and the paste path all check out (details in finding 4 below, which is a "verified, no action" entry). The problem is the one the maintainer named: the drop target is the whole window, so a file released on the sidebar, the mobile top bar, the chat toolbar or the docked Files/Subagents panels is attached to the composer as if it had been dropped on the conversation. Reviewing around that, I found a second, independent correctness bug in the same component: the drag-depth counter is an accumulator whose only unconditional resets never fire for the drag that is most likely to desync it, so a single missed `dragleave` strands the overlay on screen for the rest of the page's life. Both are fixed by the same change — anchoring the zone to the chat area lets `active` be derived from the continuously-firing `dragover`, which cannot desync.

### Findings

1. **Drop is claimed window-wide, so non-chat regions attach files** (`packages/web/src/features/chat/drop-zone.tsx:58-67`, listeners at `:72-76`) — the `drop` handler runs on a `window` listener and gates only on `wasActive`, i.e. "a file drag was somewhere over the window". Nothing tests *where* the pointer was released. Dropping a file on a sidebar session row, the mobile hamburger bar, the chat title toolbar or the docked Files panel therefore attaches it to the composer. I confirmed the sidebar's own reorder handlers (`packages/web/src/components/layout/sidebar.tsx:892-915`) early-return on a non-session drag *without* `stopPropagation`, so file drags bubble straight through to the window listener — the sidebar cannot opt out. Same root cause makes the `fixed inset-0` overlay (`drop-zone.tsx:88-90`) light up over the sidebar and every non-chat page that happens to have a composer mounted. Fix: anchor the zone to the chat-area container and hit-test the event target against it.

2. **The drag-depth counter can strand the overlay permanently** (`packages/web/src/lib/file-drop.ts:303-313`, consumed at `drop-zone.tsx:44-84`) — `dragenter` increments and `dragleave` decrements, and the only unconditional resets are `drop` and `dragend`. But `dragend` does not fire for an OS-originated file drag (the component's own comment at `drop-zone.tsx:68-70` states this), and a drag that ends *outside* the browser produces no `drop` either. So the two events that could rescue a desynced counter are exactly the two that do not happen in the failure case. Worse, the counter is sticky rather than self-correcting: once it is stuck at 1, every subsequent drag adds an enter and a leave and returns it to 1, so the overlay stays up until the user completes a real in-window drop. A depth counter is the right tool when you must infer presence from enter/leave alone; it is the wrong tool here, because `dragover` fires continuously (spec: at least every 350ms) and carries the pointer's current target, which is a direct, stateless answer to "is the drag over the region right now". Fix: derive `active` from `dragover`'s hit test, use `dragleave` only for the leaving-the-window signal (`relatedTarget` outside the region), and drop the accumulator entirely.

3. **Full-viewport `backdrop-blur-sm` for the duration of every drag** (`drop-zone.tsx:90`) — a window-sized backdrop filter forces a compositing layer over the whole page while the pointer is moving. Cosmetic consequence of finding 1 rather than a separate bug; bounding the overlay to the chat column shrinks it and it stops covering chrome that was never a drop target.

4. **Verified, no action needed** — I checked the things most likely to be wrong in a second attachment entry point and they are all correct. Size/type parity: images route to `addFiles` (`chat-input.tsx:1846`), which has no size cap — and neither does the "+" menu's image picker, so the two agree; every other file routes to `addAttachments` (`:1891`), which applies the 10MB `MAX_ATTACHMENT_BYTES` check with the existing `S.chat.attachmentTooLarge` toast *before* reading, identical to `onPickAttachments`. The 20-file count cap and 12MB total are server-side (`packages/server/src/services/task-attachments.ts:49-50`), so both entry points meet the same 400. Multi-file: `Array.from(e.dataTransfer.files)` takes the whole batch and `splitDroppedFiles` preserves drop order within each intake. Goal mode: `addAttachments` already early-returns on `goalOn`, and the drop path adds a toast so the refusal is spoken rather than silent — correct, and images are still accepted, matching the "+" menu. Non-file drags: `isFileDrag` keys on the `"Files"` type, and the sidebar's reorder deliberately uses a private MIME (`sidebar.tsx:170`), so neither text selections nor session reorders can trigger the zone or the shell guard. Paste-to-attach: `onPaste` (`:1859`) is untouched by this PR.

5. **Minor, out of scope** — dropping a *folder* yields a `File` with an empty MIME type that `FileReader` cannot read; `readDataUrl` resolves `null` on error (`chat-input.tsx:810-817`) and the batch silently drops it, so dropping a folder does nothing at all with no feedback. Drag-and-drop is the first entry point that can reach this, since the "+" picker cannot select folders. Fixing it properly means either a per-file read-failure toast or `webkitGetAsEntry().isDirectory` detection, and the first would change the picker path too — noting it rather than widening this PR.

6. **Nit** (`drop-zone.tsx:42`) — `onFilesRef.current = onFiles` is a write during render. Harmless today, but it is the pattern React's concurrent rendering guidance calls out; an effect is the sanctioned form.

- **"It should only take effect in the chat area."** — I will scope the drop target to the conversation + composer column. The anchor is `packages/web/src/features/chat/chat-page.tsx:1580`, the one wrapper that contains both the draft branch (`DraftView`) and the conversation branch, and excludes the sidebar, the mobile top bar, the chat title toolbar (a sibling at `:1261`) and the docked Subagents/Files panels (siblings at `:1688`/`:1699`). That element becomes a `ChatDropRegion` which publishes itself through context; `FileDropZone` stays mounted by `ChatInput` — so "there is a composer to attach to" remains true by construction on the chat page, on the draft page and while a Task is running, and pages with no composer get no zone — but it now hit-tests every drag event against the region element and portals a bounded `absolute inset-0` overlay *into* that element instead of a `fixed inset-0` portal to `<body>`. Precedent for the bounded form already exists in this codebase (`conversation-outline.tsx:299` inside `message-stream.tsx:367`).
- **The app-shell guard stays, and the distinction is deliberate.** There are two separate behaviors here and only one of them is being scoped down. The *drop zone* — overlay, copy cursor, attach — is confined to the chat area. The *guard* in `app.tsx` remains app-wide and stays completely silent: no overlay, no attach, no toast. Its only job is to cancel the browser's default action for a file drop that no zone claimed, because that default is to navigate the tab to the dropped file, discarding the running app and any unsent draft. Without it, dropping a PDF on the sidebar would replace AgentHub with a PDF viewer. With it, dropping a PDF on the sidebar does nothing whatsoever — which is what "only take effect in the chat area" should mean in practice: outside the chat area the gesture is inert, not destructive. The guard also shows a no-drop cursor while such a drag hovers, so the inertness is legible before the user releases. Non-file drags are untouched by the guard, so dropping selected text into the composer still inserts it natively.
- **Fallout I will carry through**: the unit tests gain the region dimension (a drop outside the region produces no attachments, and the shell guard still fires there), the depth-counter tests are replaced by tests for the stateless decision, and the changelog entry plus the en/zh `web-app` docs sentences change from "the window" to the chat area.

hiyouga:
Applied in `5c4f012e`:

- **Scoped the drop target to the chat area.** `ChatDropRegion` (new, in `drop-zone.tsx`) wraps the chat column at `chat-page.tsx:1583` — the one wrapper covering the conversation and the composer in both the draft and the session branch — and publishes its DOM node through context. `FileDropZone` stays mounted by `ChatInput`, so it is still present exactly when there is a composer (chat page, draft page, mid-Task), but every drag event is now hit-tested against the region with `contains`. The sidebar, the mobile top bar, the conversation toolbar, the docked Files/Subagents panels and every page without a composer are outside it.
- **Overlay is bounded to that region**: `absolute inset-0` portaled into the region element instead of `fixed inset-0` portaled to `<body>`, with `relative` on the region. It no longer covers chrome that was never a drop target, and the full-viewport `backdrop-blur` during every drag is gone with it.
- **The app-shell guard is unchanged and stays app-wide, deliberately.** Two different behaviors: the *zone* (overlay, copy cursor, attach) is chat-area only; the *guard* is silent everywhere else — no overlay, no attach, no toast. It exists solely so a file dropped on the sidebar does **nothing**, rather than the browser's default of navigating the tab to that file and discarding the running app with any unsent draft. A no-drop cursor makes that inertness legible before the release. Non-file drags stay untouched, so dropping selected text into the composer still inserts it natively.
- **Replaced the drag-depth counter** (finding 2 in the review). It could strand the overlay permanently: its only unconditional resets were `drop` and `dragend`, and a file drag that ends outside the browser fires neither, so a single missed `dragleave` left it off by one for the rest of the page's life. `dropRegionAction` is now a pure function of the current event — `dragover` fires continuously and names the element under the pointer, so the answer is re-derived rather than accumulated and any missed event is corrected by the next one. `dragleave` is kept only for the leaving-the-window signal, and tested on the *related* target so internal element boundaries do not disturb it.
- **Tests**: `test/file-drop.test.ts` gains the region dimension — a drop outside the chat area produces no attachments while the guard still cancels it, non-file drags are never claimed in any signal, a `dragleave` into a child keeps the overlay up, and a replay of out-of-order signals pins the statelessness. 16 tests; full web suite 737 passing, typecheck and `vite build` clean.
- **Docs/changelog/strings**: changelog entry + index line, `web-app.en.md` and `web-app.zh.md` all say the chat area instead of the window, and spell out that a drop elsewhere is silently ignored. The overlay strings themselves were already region-neutral; only their comments changed.
- Also fixed while rewriting: `onFilesRef` was written during render (review finding 6) — now an effect.

Deliberately not changed:
- **Dropped folders** (review finding 5) still do nothing silently: a folder arrives as a `File` with an empty MIME type, `readDataUrl` resolves `null` on the read failure, and the batch drops it without feedback. Fixing it properly means per-file read-failure toasts or `webkitGetAsEntry().isDirectory` detection, and the former would change the "+" picker path too — out of scope for this PR.
- **The docked Files panel is outside the region.** A file dropped on the workspace file tree plausibly means "upload into the workspace", which is a different feature with a different destination; treating it as a composer attachment would be a guess. It is inert for now.

PR #320, opened 2026-08-18
feat(web): confirm mid-chat thinking-level switches to protect the prefix cache

### LINKED ISSUE #310: changing thinking level in middle chat breaks prefix caching
> Image
> DeepSeek 的思考深度控制就是在 chat template 最前面注入提示词实现的。关闭思考和 low 没有提示词，high 喝 max 提示词有差异。所以如果你们能测的话你会发现，如果发你好两个字，关闭思考或 low 时是输入 5 个 token，high 是 84 个 token，max 是 97 个 token，就是这两段提示词的差异

在切换前弹出，提示用户建议先手动 compact 以后再切换，而不是直接切换，这样可以减少成本。但是也允许用户强行切换

hiyouga:
## What & why

Some providers (DeepSeek is the reported example) implement the thinking level by injecting a different prompt prefix at the very FRONT of the chat template — off/low means no prefix, high and max each mean a different long one. Switching the level in the middle of a conversation therefore invalidates the provider's prefix cache for the entire history, and the next request re-bills all of it at the uncached input rate.

The active-session composer's thinking-level picker no longer switches silently once the conversation has history. A pick that would change the level is staged behind the app's standard confirmation card (neutral tone), whose body explains the prefix-cache cost, advises running `/compact` first, and offers **Switch anyway** to force the change; Cancel keeps the current level. The dialog is skipped whenever the switch cannot hurt:

- the conversation has no messages yet — a brand-new session's initial selection stays free (and the draft picker, which has no session at all, is untouched);
- the pick equals the level the picker already displays — it merely pins the level for the session, nothing provider-visible changes;
- the transcript ends in a successful compaction — the user who followed the "compact first, then switch" advice is not warned a second time (a failed or still-running trailing compaction leaves the old context in effect, so those still confirm).

An unknown displayed level ("" — the Agent config is unset or its fetch failed) confirms conservatively whenever history exists: a needless dialog costs one click, a silently invalidated prefix cache costs real money.

Fixes #310

## Key changes by package

**packages/web**
- `src/features/chat/thinking-level.ts`: new pure guard logic — `prefixCacheAtRisk(items)` (empty transcript free; trailing successful compaction free; anything else at risk) and `needsThinkingSwitchConfirm(items, currentLevel, nextLevel)` (re-pick of the displayed level never confirms). Operates on a structural subset of the stream's ChatItems, so the live tail window is passed as-is.
- `src/features/chat/chat-page.tsx`: the session picker's pick handler now routes through the guard — staged picks land in `pendingThinkingLevel` and render a `ConfirmModal` (tone `primary`, confirm label "Switch anyway"); confirm applies the staged level, cancel discards it, and a session switch resets the staged state. The composer only mounts once history load settles, so the tail items are authoritative at pick time.
- `src/lib/strings.ts` / `src/lib/strings-en.ts`: `chat.thinkingSwitchTitle`, `chat.thinkingSwitchBody(to)`, `chat.thinkingSwitchConfirm` in both locales.
- `test/thinking-level.test.ts`: 11 new unit tests covering the guard — dialog only when history exists, no dialog on empty transcripts / re-picks / after a successful trailing compaction, failed or running trailing compactions still confirm, new turns after a compaction re-arm the guard, unknown current level confirms.

**packages/docs**
- `content/web-app.en.md` / `content/web-app.zh.md`: the per-turn thinking-level paragraph now documents the confirmation flow and its skip conditions.

**changelog**
- `changelog/unreleased/2026-08-18-thinking-switch-guard.md` + index line in `changelog/unreleased/README.md`.

## Design decisions

- **Two buttons, not three.** The unified confirm card (design/specs/06-PROTOTYPE.md, confirmation policy) is deliberately a fixed Cancel/Confirm pair — "one compact, identical card" across the app — so no "compact first" button was added. The body's advice names `/compact`, and the trailing-compaction skip makes the advised flow friction-free: cancel, run `/compact`, re-pick — no second dialog.
- **Neutral tone (`primary`).** Nothing is destroyed; the two spec'd tones are red-triangle (deletion) and neutral pencil (overwrite/confirm), and the latter fits a settings change with a cost caveat.
- **Guard compares against the DISPLAYED level** (session pick, else Agent config) — the same value the user sees on the trigger — rather than re-resolving the server-side fallback chain, keeping the dialog's trigger predictable from what is on screen.

## Design docs

`design/` is read-only for this PR; proposed edits to `design/specs/06-PROTOTYPE.md`:

1. In the confirmation-policy bullet (the「确认口径」bullet), the quick-control exemption gains a carve-out — replace

   「……确认后执行并以顶部浮层报结果；对话内的即改即存快捷控件（审批模式、思考等级）除外。」

   with

   「……确认后执行并以顶部浮层报结果；对话内的即改即存快捷控件（审批模式、思考等级）除外——唯会话中途切换思考等级需按「输入区」一节的前缀缓存守卫先行确认。」

2. In the input-area bullet (the「输入区」bullet), append to the thinking-level selector clause, after「（初始即显示 Agent 配置等级、未选择时随配置生效，选定后随后各 Task 自动携带、不写回 Agent 设置；不设「跟随 Agent 配置」选项）」:

   「会话已有历史时改选其他档位不静默生效：部分模型服务商按思考等级在提示词最前端注入不同前缀，此时切换会使服务商的前缀缓存整体失效、下一次请求按未缓存价格重新计入全部历史，故先弹统一样式的确认框（中性色调，正文说明该代价并建议先 /compact 压缩再切换，确认按钮为「仍要切换」，取消保持原档位）；会话尚无消息、重选当前显示档位、或紧跟一次成功压缩之后的改选直接生效、不弹框；当前显示档位未知（Agent 配置未设或未取得）而会话已有历史时按需确认处理。草稿态选择器不受影响（尚无会话、亦无服务商缓存）。」

## Tests

- `pnpm --filter @prismshadow/penguin-web test` — 60 files / 732 tests passed (includes the 11 new guard tests; `test/thinking-level.test.ts` alone: 18 passed).
- `pnpm --filter @prismshadow/penguin-docs test` — 32 passed; `typecheck` clean.
- `npx prettier --check` on all touched files — clean.
- Note: the workspace packages `skills`/`core`/`server` had to be built once in the fresh worktree before the web typecheck could resolve `@prismshadow/penguin-core` (environmental, not caused by this change).


hiyouga:
## Review

The guard itself is in good shape: `prefixCacheAtRisk` / `needsThinkingSwitchConfirm` are pure, well-documented and pinned by real behavioral tests, the wiring only touches the active-session picker (the draft picker keeps writing through to the Agent config untouched), staged state resets on session switch, and the repo conventions are all met — both locales, en+zh docs, one changelog entry extended in place. What the PR does not yet do is give the user a way to *act* on the advice it prints, which is the maintainer's point below; that same gap also has a concrete correctness edge (the advice is literally unreachable mid-run). Everything else I found is small.

### Findings

1. **The dialog advises `/compact`, but mid-run that command cannot be run at all** (`packages/web/src/features/chat/chat-input.tsx:1390`, `packages/server/src/runtime/session-manager.ts:846`) — the picker is only disabled by the composer's local `busy` (send-in-flight) flag (`chat-input.tsx:2518`), so a level can be picked while the session is `running`/`compacting`. The dialog then says "run `/compact` first", while the slash menu is suppressed exactly in those states (`slashTok = null` when `running || compacting`) and `startCompact` asserts idle server-side (409 `task_in_progress` / `compacting`). So the recommended path is dead copy precisely when the user is most likely to hit it. The three-choice dialog fixes this by owning the compaction itself and disabling option 1 with a stated reason while a task runs.

2. **Compaction failures surface untranslated in the zh UI** (`packages/web/src/lib/api-error.ts:16`, `packages/web/src/lib/strings.ts` `errors.byCode`) — `apiErrorText` maps `ApiError.code` through `S.errors.byCode`, which has `task_in_progress` but neither `nothing_to_compact` nor `compacting`; both 409s therefore render the raw English server sentence even under 中文. Today that is only reachable by typing `/compact`; once the dialog has a "Compact, then switch" button it becomes a first-class path (e.g. picking it on a session whose first turn was interrupted → `nothing_to_compact`). Adding both codes to both catalogs.

3. **Requirement gap, deliberately not fixed here** — the guard covers only the composer picker. While `turnThinkingLevel` is untouched the picker auto-follows the Agent config, so editing `model.thinking_level` on the Agent settings page silently changes the level of every existing session on its next send: the same prefix-cache invalidation, no dialog. #310 is about the in-chat switch, so widening the guard to the settings surface is a separate change; flagging it so it isn't assumed covered.

4. **Tail-window caveat (accepted, not a bug in practice)** — `prefixCacheAtRisk` sees only `stream.model.items` (the live tail); backfilled `prefixItems` are not passed. A tail that consisted *only* of compaction rows would read as "no history" and skip the dialog, but back-to-back compactions are rejected by `compactability()` (`just_compacted`), so it takes an unrealistic run of consecutive failed compactions to construct. Worth keeping in the function's doc comment, which it already is.

Explicitly checked and found sound (no change needed):

- **Trailing successful compaction = safe** is correct *and* robust to item ordering: `task_stats` is deliberately inserted **before** the trailing run of compaction banners (`packages/web/src/lib/omni/stream-model.ts:829-837`), so a manual `/compact` really does leave `compaction` as the last item; `status` values match core's `StopReason` (`completed` / `failed` / `aborted`).
- **Unknown current level ("") confirms conservatively** — right trade: one wasted click vs. re-billing the whole history.
- **Brand-new session skips the guard** — `items.length === 0` → `false`, and an empty `items` can never mean "history not loaded yet": the composer only mounts on the `!stream.loading && !stream.error` branch (`chat-page.tsx:1644-1650`).
- **Session switch resets staged state** — `chat-page.tsx:633-641`, keyed on `routeSessionId`. A compaction that self-heals to a new session id re-routes and therefore also drops the staged pick, which is consistent with `turnThinkingLevel` itself resetting there.
- **Draft picker never confirms** — the guarded prop is the session-only `onChangeTurnThinkingLevel` (`chat-input.tsx:2514`); the draft's `onChangeThinkingLevel` (`draft-view.tsx:427`) still writes straight through to the Agent config.
- Reading `stream.model.items` inside `onPickTurnThinkingLevel` is correct even though the model is mutated in place — the closure resolves `.items` at click time.

> The dialog must also offer the option to compact and then switch. Recommend the user pick from: 1. compact manually and switch, 2. switch directly, 3. cancel.

- **Three choices.** `ConfirmModal` grows one optional secondary action (`secondaryLabel` + `onSecondary`) plus `confirmDisabled`; both default to today's exact markup, so the other 20-odd call sites are untouched. No bespoke modal, no raw `Modal`+footer fork. Buttons become: `Cancel` — `Switch anyway` (neutral) — **`Compact, then switch`** (accent primary, with a "recommended" hint line in the body).
- **Option 1 reuses the existing path**: the page's own `onCompact` (`chat-page.tsx:1078` → `api.postCompact` → `POST /api/sessions/:id/compact`), refactored into a `runCompaction(): Promise<boolean>` that `onCompact` still delegates to, so the composer's `/compact` behaves identically.
- **Sequencing.** `POST /compact` returns **202** and resolves before the compaction runs, so the switch cannot be applied on the promise. The pick is staged as `{ phase: "compacting", level, baseline }` and applied only when the stream shows the compaction *completed*: a new pure function counts settled vs. completed compaction items in the tail against the baseline captured at click time (robust to items appended afterwards, e.g. a queued follow-up starting).
- **On failure, do not switch** (stated choice): a compaction that settles `failed`/`aborted`, or a POST that 409s, clears the staged pick, leaves the level unchanged, and says so — an error toast for the run-time failure (the transcript's red compaction banner alone would leave the user assuming the switch landed) on top of the existing `apiErrorText` toast for the rejected POST.
- **Mid-run**: option 1 is **disabled with a visible reason** while `stream.taskState !== "idle"` — compaction is not queueable server-side (`queueIfBusy` exists only for `startTask`), so queueing it would be inventing a second code path. Options 2 and 3 stay available.
- Both locales, the branch's changelog entry + its index line, en/zh docs, and the guard tests extended to the three paths (including compaction-fails-so-do-not-switch).

hiyouga:
Applied in 373ae053:

- **Three choices in the dialog** (maintainer feedback): **Compact, then switch** (primary/recommended, with a hint line naming it as the recommendation) — **Switch anyway** — **Cancel**.
- **Option 1 reuses the existing compaction path**: `onCompact` was split into `runCompaction(): Promise<boolean>` (the single call to `POST /api/sessions/:id/compact`) with `onCompact` delegating to it, so the composer's `/compact` command is byte-for-byte the same behavior.
- **Sequencing**: the endpoint answers 202, so the pick is held as `{ phase: "compacting", level, baseline }` and released only when the transcript shows a compaction *completed*, via two new pure functions (`compactionTally`, `thinkingSwitchAfterCompaction`). Counting settled/completed compactions against a baseline taken at click time (rather than matching item ids) keeps it correct when the stream reconnects and rebuilds history, and when items land after the compaction (a queued follow-up starting).
- **Compaction fails / is aborted / is refused → the level does not change**, and the user is told: an error toast for a run-time failure (the transcript's red banner alone would let the user assume the switch landed), the existing `apiErrorText` toast for a rejected POST. An info toast on start says the switch is waiting for the compaction.
- **Mid-run**: option 1 is disabled with the reason shown in the body — `startCompact` asserts idle and compaction is not queueable server-side (`queueIfBusy` exists only for `startTask`), so queueing it would have meant a second code path. Options 2 and 3 stay available.
- **Shared `ConfirmModal`** gained an optional third action (`secondaryLabel` + `onSecondary`) and `confirmDisabled`; both default to the previous two-button card, so all other call sites render and behave exactly as before. Its button row now wraps (`flex-wrap`) so three long labels don't overflow the bottom-sheet width on a phone; a two-button card never reaches the wrap point.
- **Localized the `compacting` server error code** (both catalogs) — now reachable from a button rather than only from a typed command.
- Tests extended: `pending` / `apply` / `failed` paths, the failure-already-on-record case (a failed compaction that was in the transcript when the dialog opened must not settle the wait), and the failed-retry-after-success case. 25 tests in that file, full web suite 739 passing, `tsc --noEmit` clean for web and docs, prettier applied.
- Docs (en + zh) and the branch's changelog entry + index line rewritten for the three-choice flow.

Deliberately not changed:

- **`nothing_to_compact` is still NOT in `S.errors.byCode`**, contrary to what my review said I would do: `apiErrorText` prefers the mapped text over the server message, and the server deliberately sends three *different* reasons under that one code (compaction not configured / no completed turn yet / just compacted). One localized sentence would flatten them, and would be factually wrong for the "not configured" case — the specific English message is the better outcome. Only `compacting`, which has a single meaning, was mapped.
- **The Agent-settings surface is still unguarded**: while the session picker is untouched it auto-follows `model.thinking_level`, so editing that on the Agent settings page silently changes the level of existing sessions on their next send — same prefix-cache cost, no dialog. Out of scope for #310's in-chat switch; worth its own issue if you want it covered.
- A compaction that **self-heals to a new session id** re-routes and therefore drops the held pick (the per-session reset also clears `turnThinkingLevel` today, so the level would have been reset regardless). Documented in the reset effect's comment rather than special-cased.

hiyouga:
Applied in 3f540ddc (on top of a clean `origin/main` merge, 8045d288):

**1. Switch even when the compaction fails.** "Compact, then switch" now applies the picked level whichever way the compaction ends. A failed/aborted one still reports that it did not finish (the context was not rewritten, so the switch does cost cache hits) but no longer swallows the switch; a compaction the server refuses (409 `nothing_to_compact` and friends) releases the pick immediately, since nothing will ever settle. The pick is still *held* until the compaction ends — that is about order, not veto: a level applied mid-compaction would ride on anything sent in between and invalidate the cache the compaction is there to shrink. The release decision moved into a pure `heldThinkingSwitch(staged, items)`, so "failed compaction ⇒ still apply" is pinned by a test rather than living in an effect.

**2. Simpler copy.** The dialog body is one sentence in both locales: switching mid-conversation lowers the prompt-cache hit rate and raises cost, compacting first is cheaper. The prefix-injection/re-billing explanation and the separate "recommended" line are deleted (`thinkingSwitchRecommend` is gone); the busy hint is trimmed to why compaction has to wait.

**3. The level survives a reload** — it was React state only, so a refresh silently reverted to the Agent config. It is now a Session field:

- nullable `sessions.thinking_level`, added through the existing `ensureColumn` upgrade path (no migration), behind a new `SessionInfo.thinkingLevel`, written with `PATCH /api/sessions/:sessionId { thinkingLevel }` (enum-validated, 400 otherwise; added to the `no_update` guard).
- `SessionManager` resolves it at launch for every run that carries no level of its own — tasks, queued follow-ups and goal runs — so the chain is: request's own level → Session pin → Agent config. That keeps the pin honest for non-web clients too, instead of making it a browser-only illusion.
- Unset (a new Session) behaves exactly as before: display and runs auto-follow the Agent config. Once pinned, an Agent-config edit no longer moves that session.

`session_meta` was considered first and rejected: it is an immutable Trace record of per-session invariants, and core deliberately removed the thinking level from it when it became a per-turn parameter.

**Verification**: web 769 tests, server 648 tests, docs 53 tests, `tsc --noEmit` clean for web/server/docs/cli, prettier on all touched files. New coverage: PATCH pins and reads back (plus invalid → 400 and empty body → 400), a pinned level applies to later tasks while a request's own level still wins and a re-pin takes effect from the next task, a failed/aborted compaction still applies the pick, a refused compaction applies it too, and a pinned level beats a changed Agent config while an unpinned session follows it. No e2e run.

Docs: Web App (en+zh) rewritten for the three choices and the durable level; Server API (en+zh) documents the PATCH field, `SessionInfo.thinkingLevel` and the new fallback chain. The bilingual changelog pair was rewritten to match (the entry now also carries `server` in its scope).

PR #321, opened 2026-08-19
feat: recall queued steering and follow-up messages into the composer

### LINKED ISSUE #287: Support recall the steering message

hiyouga:
Builds on draft PR #304 by @Myriad-Dreamin (branch `cancel-steer`): this branch keeps that PR's commit as-is and adds one review/hardening commit on top.

Fixes #287

## What & why

A message the user sent mid-run but that has not actually gone out yet — an undelivered steering message (插话), or a follow-up queued behind the current run — can now be **recalled back into the composer**, edited, and resent. Every queued hint line above the input box gains a Recall button; the server withdraws the entry with exactly-one-of delivery/recall semantics and returns the original content (text, images, file attachments, and — for follow-ups — the per-turn thinking level it was queued with), which the composer restores into the draft.

## Key changes by package

- **core**: `ContextEngine.unsteer(input)` / `Session.unsteer(input)` withdraw a queued steering input before delivery, matched by input-list identity; they refuse once the queue drained (delivered, or the run exited), which is what makes recall race-safe against delivery — the drain and the withdrawal are both synchronous on the engine loop, so exactly one of them gets the entry.
- **server**:
  - `DELETE /api/sessions/:id/steer/:steerId` and `DELETE /api/sessions/:id/follow-ups/:followUpId` → 200 with the withdrawn message's original content `{text, images, files, thinkingLevel?}`; 409 `not_pending` once the steering was delivered to the model / the follow-up auto-started (or the id is unknown).
  - `pendingSteering` mirror entries gain a server-assigned `id`; a new `pendingFollowUps` list (id + text + image/file counts per entry) rides `task_state` events and the SSE subscribe snapshot, so the per-entry hint lines and their recall handles survive reloads.
  - A `RecallStore` (text, image URLs, attachment file names/paths/mime) is kept per queued entry; on recall, file attachments are read back from the Session scratchpad as the same data URLs the composer submitted (mime now captured by `parseAttachmentPart`), and the disk copies are deleted — nothing references them once the message is withdrawn.
  - Steering recall defers to core's `unsteer` as the authority (the display mirror can lag delivery until the `[user_steering]` message is observed on the stream: a recall in that window is a clean 409, and the mirror entry retires through the normal stream shift). Follow-up recall splices the queue synchronously; the auto-start drain revalidates under the session lock, so a recall landing in the idle gap wins and nothing starts.
  - Recall re-broadcasts `task_state`, so other open tabs see the queued entry disappear immediately.
- **web**: per-entry hint lines for both channels with a Recall button (one recall in flight at a time; suppressed while a send is in flight, and absent for old servers that send no ids/per-entry list — the bare count hint stays as the fallback). Restored content merges into the draft: text in front of whatever is typed, images/files back as composer chips, thinking level restored for follow-ups; a 409 surfaces as a localized toast while the hint line retires via the re-broadcast state. New strings in both locales (`strings.ts` zh / `strings-en.ts` en), including the `not_pending` error text.
- **docs**: `server-api.{en,zh}.md` route table rows for both DELETE endpoints; `web-app.{en,zh}.md` input-area bullet describing the Recall affordance.
- **changelog**: `changelog/unreleased/2026-08-16-recall-queued-messages.md` + index line.

## Review verdict on #304

**Already solid** — the implementation was complete and correct against the issue and the specs:

- Both channels (steering and queued follow-ups) recallable end-to-end, attachments included, with the composer-shaped response the draft needs.
- Race safety done right: core's identity-matched `unsteer` is the single authority for steering (mirror lag handled, with a test); follow-up recall vs auto-start is exactly-once by construction (synchronous splice vs under-lock revalidating dequeue); double recall / cross-tab recall → clean 409 `not_pending` (tested); scratchpad hygiene on recall and on the 409-steer fallback.
- Only the tasks route feeds the follow-up queue, so every real entry carries a recall store (the scheduler keeps its own queue and never enqueues here); production wiring is sound — the loader hands the manager core's `Session`, which now carries `unsteer`.
- Multi-tab convergence via `task_state` re-broadcasts; reload survival via the SSE snapshot; old-server/new-web degradation handled; strings in both locales; docs and changelog in sync. All three suites and typechecks were green as submitted.

**Changed on top** (commit `fix(web,server): harden queued-message recall (#287)`):

1. **Lost-keystroke fix (web)**: `applyRecalled` merged the recalled text with the draft text captured when the Recall button was clicked, so anything typed during the DELETE round-trip was overwritten. The merge now reads the live value off the controlled textarea, through a small extracted module `recall-draft.ts` (`mergeRecalledDraft`, unit-tested).
2. **Goal-chip invariant (web)**: recalling a message with file attachments while a goal chip was staged parked the files behind a Send that can never enable (a goal draft cannot carry files — engaging the chip clears them and the server refuses them), with their scratchpad copies already deleted server-side. The recall now releases the staged goal chip when files come back, mirroring `toggleGoal`'s own invariant.
3. **Race-condition test coverage (server)**: a deterministic test for the recall vs auto-start race — a recall enqueued as a microtask from inside the idle `task_state` broadcast lands exactly in the gap after the drain is scheduled and before its locked dequeue; the recall must win exactly once and the drain must start nothing (also asserts the re-broadcast another tab would see). Plus a test that the SSE subscribe snapshot carries `pendingFollowUps` (what rebuilds the recall lines after a reload).
4. **Changelog**: the goal-chip release noted in the existing feature entry.

Considered and left as-is: a queued follow-up composed with selected skills recalls its literal `[use_skills]` block as editable text — an unedited resend delivers exactly the original message, so the restore is lossless; re-parsing the block back into skill chips was judged not worth the edge cases (e.g. a skill uninstalled meanwhile).

## Tests

- `@prismshadow/penguin-core`: 797 passed (5 skipped) — includes the new `unsteer` engine test from #304.
- `@prismshadow/penguin-server`: 608 passed — includes the two new tests (idle-gap race, SSE snapshot) and #304's recall/steer/follow-up tests.
- `@prismshadow/penguin-web`: 725 passed — includes the new `recall-draft` tests and #304's stream-controller tests.
- `typecheck` clean for core, server, and web. Touched files formatted with Prettier.

## Design docs

The specs describe the steering mirror and follow-up queue but predate recall. Proposed edits (not applied — `design/` is shared):

1. `design/specs/05-ARCHITECTURE.md`, 「运行中 steering」 bullet (§ Task 运行与流式通道, ~line 1195), append after 「……送达后由消息流中的用户插话小块接替。」:

   > 未送达的插话可**撤回**（#287）：`DELETE /api/sessions/:sessionId/steer/:steerId`（id 随镜像逐条下发）经核心 `Session.unsteer` 按输入列表同一性从队列撤出该条并同步移除镜像条目，返回其原始内容 `{text, images, files}`（文件从会话 scratchpad 读回为 data URL，磁盘副本随之删除），供输入框恢复编辑后重发；核心队列已排空（已送达或运行结束）即拒绝，故送达与撤回严格二选一——已送达返回 409 `not_pending`，镜像条目留待流上出现对应 `[user_steering]` 后照常退场。

2. `design/specs/05-ARCHITECTURE.md`, 「follow-up 排队」 bullet (~line 1196), append:

   > 排队中的跟进消息同样可撤回：`DELETE /api/sessions/:sessionId/follow-ups/:followUpId`（id 随 `task_state` 的 `pendingFollowUps` 列表逐条下发，含文本与图片/文件计数）在自动发出前把该条移出队列并返回原始内容（含随队暂存的 thinkingLevel）；自动发出在会话锁下重校验队列，与撤回严格二选一——已发出（或运行时条目已被重启清空）返回 409 `not_pending`。

3. `design/specs/05-ARCHITECTURE.md`, Session 级服务端事件 (~line 1201), change 「`task_state`（任务状态快照与变更，携带 follow-up 排队数与未送达 steering 镜像）」 to:

   > `task_state`（任务状态快照与变更，携带 follow-up 排队数、未送达 steering 镜像与排队中的跟进消息列表 `pendingFollowUps`——两者逐条含撤回用 id）

4. `design/specs/05-ARCHITECTURE.md`, API 原型 (~line 1265), add after the `POST /steer` line:

   > `DELETE /api/sessions/:sessionId/steer/:steerId` # 撤回未送达的插话：返回原始内容供输入框恢复编辑；已送达 409 not_pending
   > `DELETE /api/sessions/:sessionId/follow-ups/:followUpId` # 撤回排队中的跟进消息：自动发出前移除并返回原始内容（含 thinkingLevel）；已发出 409 not_pending

5. `design/specs/06-PROTOTYPE.md`, 「输入区」 bullet (~line 75), append after 「……排队数显示在输入区旁（不藏进弹出分组）。」:

   > 两类排队提示行（未送达插话、排队中跟进消息——后者逐条显示内容，取代裸计数）都带**撤回**按钮：点击即在服务端撤出该条并把原始内容恢复进输入框编辑重发——文字置于现有草稿之前，图片与文件附件恢复为输入区小卡片（文件从 scratchpad 读回、磁盘副本随之删除），跟进消息还恢复随队的思考等级；恢复文件附件时若正暂存 goal chip 则释放之（goal 草稿不携带文件附件）。已送达/已发出的条目以 409 提示无法撤回，提示行随事件自行退场；另一打开的页面经 task_state 即时看到被撤回的条目消失。


hiyouga:
## Review

The recall semantics hold up. I went after the exactly-once question specifically — whether a recall racing delivery can produce a duplicate or a silently lost message — and both queues are safe for a structural reason, not by luck: core's `deliverSteering` snapshots and clears the whole steering queue synchronously before its first `await` (`context-engine.ts:519-522`), and `unsteer` is a synchronous `indexOf` + `splice` called synchronously from the route, so a recall either wins outright (the entry never reaches the model) or loses with 409 (it is already on its way) — there is no window where both happen. The follow-up queue is the same shape one level up: `startQueuedFollowUp` re-checks `followUps.length` and `shift()`s with no `await` in between, under the session lock (`session-manager.ts:881-889`), and `recallFollowUp` is fully synchronous, so it cannot interleave inside that critical section. The test at `followup.test.ts` that fires a recall from a microtask inside the idle broadcast pins exactly the gap where this could have gone wrong. Ordering survives too: `splice` keeps the remaining entries FIFO, delivery `shift()`s the head, and the display mirror stays index-aligned with core's queue even when a middle entry is withdrawn. Findings below are one convergence bug the PR introduces, one durability limitation worth stating out loud, and a nit.

### Findings

1. **A recall from another tab leaves a phantom "steering queued" hint behind** (`packages/web/src/features/chat/chat-input.tsx:1343-1350`, rendered at `:2206-2212`) — `steerPending` is the local bridge between the steer 202 and the first `task_state` carrying the mirror, and it is cleared only on delivery (`steeringDeliveredCount` increases) or when the run stops. `recallQueued` clears it explicitly, but only in the tab that clicked. In any other tab that posted a steer, the recall empties `pendingSteering` *without* a delivery and without ending the run, so the render falls through to the bare `steerPending && <hint>` branch and that tab advertises a queued steering message that no longer exists — for the rest of the run. This state is new: before this PR the mirror could only empty via delivery (`session-manager.ts:1391`) or run end (`:1509`, published as `idle`), and both of those also clear the bridge. Fix is one condition: the bridge has done its job the moment the mirror has been seen non-empty, so clear it then.

2. **After a recall, the composer holds the only copy of the images and files** (`packages/web/src/features/chat/draft-cache.ts:17-45`) — the draft cache persists `text` and the staged chips, never `images` or `attachments`; those live in `ChatInput` state alone. A successful recall reads the attachments back and then deletes their scratchpad copies (`http/routes/sessions.ts:345-357`) and drops the queue entry, so from that moment the browser tab is the sole holder. A reload, a session switch, or an unmount while the DELETE is still in flight (there is no `AbortController`, so `applyRecalled` just setStates into an unmounted tree) loses them permanently — the text comes back from the cache, the pictures and files do not. Before recall existed, a lost composer attachment was always re-pickable from the user's own disk; this is the first flow where it isn't. I am not fixing it here: persisting up to 12MB of base64 needs a store `localStorage` cannot be (IndexedDB), which is its own change. Worth stating in the docs or accepting knowingly.

3. **The recall control stays enabled while a send is in flight** (`chat-input.tsx:1396-1398` vs `:2199`/`:2224`) — `recallQueued` correctly refuses to run when `busy` (a send is about to clear the draft), but `disabled` is only `recallingId !== null`, so during a send the button looks live and a click does nothing at all, with no feedback. Folding `busy` into `disabled` makes the refusal visible.

4. *(Noted, not changing)* The 409 copy — "This message already went out and can no longer be recalled" — is also what you see when another tab won the race, where the message did not go out at all. Distinguishing the two would need a second error code for a case the user cannot act on differently either way.

Verified and found correct, for the record: the attachment round-trip (the `parseAttachmentPart` regex now captures the media type, `attachFilesToInput` writes sequentially so the `written` ↔ `attachments` zip in `recallStore` is order-safe, files are read *before* they are deleted, a vanished file is skipped instead of failing the recall, and both `readRecalledFiles` and `removeAttachments` swallow their own I/O errors so a recall can never 500 *after* withdrawing the entry — which would have been the one real way to lose a message); multi-tab convergence on both queues (every recall and every auto-start re-broadcasts `task_state`); the 409 path leaving the steering mirror for the stream shift to retire; goal tasks never entering the follow-up queue, so there is no goal-shaped recall to get wrong; and the two fixes in `a3fff24a` — reading the live textarea value is right because the textarea is controlled (its DOM value is always the latest committed state, keystrokes during the round-trip included), and releasing the goal chip is safe because `toggleGoal(false)` clears attachments only on the `true` branch, so the prepend that follows survives. Conventions check out: both locales, `.en.md` + `.zh.md`, one changelog entry plus its index line, English-only code.

- **Change the recall control to a curved-back arrow icon — no text.** I'll replace the textual button with an icon-only control using a curved-back (undo-style) arrow drawn in the repo's own idiom: a new `RECALL_ICON` path constant next to the other chat glyphs, rendered through `GlyphIcon` (24×24 viewBox, `strokeWidth="1.7"`, round caps/joins) at the size the neighbouring composer controls use. The two existing localized strings stay and become the accessible name: `aria-label={S.chat.recallQueued}` ("Recall" / "撤回") and `title={S.chat.recallQueuedTitle}` (the full "edit and resend" tooltip), matching how every other icon-only button in the composer is named. Hit area, hover and dark-mode colors follow the hint-row idiom already used for the copy button on the per-turn stats line (`h-5 w-5`, `rounded`, `hover:bg-gray-100 dark:hover:bg-gray-800`), with the icon inheriting the hint line's own gray so it reads as part of that row; focus comes from the global `button:focus-visible` outline. The line stays `min-w-0 truncate` with a `shrink-0` button, so the control survives narrow widths while the text truncates. No test changes are needed for this: `packages/web` runs vitest in a node environment with no DOM tests, and no e2e spec clicks the recall control today — the `aria-label` is what one would query by when it does.

hiyouga:
Applied in `b0669765`:

- **Maintainer feedback — the recall control is now an icon, no text.** A curved-back (undo) arrow: a new `RECALL_ICON` path rendered through `GlyphIcon` in the repo's own idiom (24×24, `strokeWidth="1.7"`, round caps), at the size the other icon controls on a hint row use. The two existing localized strings became its accessible name rather than its body — `aria-label={S.chat.recallQueued}` ("Recall" / "撤回"), `title={S.chat.recallQueuedTitle}` for the tooltip — so nothing was deleted from either catalog. It carries the icon set's gray (`text-gray-500 dark:text-gray-400`) instead of the hint text's: `gray-400` on white is 2.6:1, under the 3:1 an interactive control owes, and this one is interactive. Verified the glyph reads as a return arrow at its actual 13px in both themes, and the line stays `min-w-0 truncate` + `shrink-0` so the control survives narrow widths.
- **Finding 1 (cross-tab phantom hint).** The local "steering queued" bridge now also retires once the server's mirror has arrived — the moment it has nothing left to bridge — so a recall performed in another tab no longer strands the sending tab with a hint for a message that no longer exists.
- **Finding 3 (silent refusal).** `busy` folded into the buttons' `disabled`, so the refusal `recallQueued` already performed is visible instead of a dead click.
- **One thing I missed in the review and found while fixing:** `packages/web/e2e/reloaded-followup.spec.mjs:104` asserted on the bare `"1 条跟进消息已排队，本轮结束后自动发送"` count hint, which this branch replaced with the per-entry line — that spec has been failing since the feature commit, not just now. I confirmed it (ran it on the old assertion: `element(s) not found`), then updated it to pin the per-entry hint *and* the recall control's accessible name, which is now the only test anywhere that holds the icon to having a name. Both affected specs pass (`reloaded-followup`, `steer-reload`), run on a private port.
- Restored `appendAttachmentParts`' doc comment, which the feature commit orphaned onto `dataUrlBytes` by inserting that helper between the comment and its function.
- Docs (en+zh), the changelog entry, and the string catalog comments describe the icon control and the cross-tab behavior.

Deliberately not changed:

- **Finding 2 (recalled attachments live only in composer state).** Real, and I stand by flagging it, but the fix is a storage change, not a patch: the draft cache is `localStorage` and the attachment budget is 12MB, so persisting recalled images/files needs IndexedDB and belongs in its own PR. Nothing in this branch makes it worse than the moment the recall was designed; it just deserves to be a known limitation rather than a surprise.
- **Finding 4 (409 copy).** Left alone — a second error code for "another tab took it back" buys the user nothing they can act on.

Tests: `packages/web` 725/725, `packages/server` steer + followup + session-manager 44/44, web typecheck clean, prettier clean. I did not run the full e2e suite.

PR #322, opened 2026-08-19
feat(cli): runtime thinking-level control and collapsed tool output

### LINKED ISSUE #305: cli模式下 最好可以提供 思考强度修改 还有就是 隐藏一些输入输出 满屏都是exec_command的返回

hiyouga:
## What & why

Issue #305 asks for two things in CLI mode: (a) a way to change the thinking level, and (b) less screen noise — long `exec_command` returns flood the chat. This PR adds both, reusing the semantics the product already has elsewhere (core's tri-state `thinkingLevel` options and the web active-session picker) so the CLI doesn't invent a third model.

**Thinking level**
- `penguin run --thinking <low|medium|high|xhigh>` and `penguin chat --thinking <level>` pin the Session's thinking level at creation (`createSession.thinkingLevel`), so spawned subagent sessions follow it. Omitted, the existing configured chain applies (Agent `model.thinking_level` > Project `default_chat.thinking_level` > `medium`).
- In the chat REPL, `/thinking` shows the level the next turn will run at; `/thinking <level>` sets a per-turn override carried on every subsequent run (`RunOptions.thinkingLevel`) — the same "per-turn override, never written back to the Agent config" semantics as the web active-session picker. Goal mode reuses the override for every round (core behavior); `/compact` is unaffected (per spec).
- Under `--resume`, `--thinking` is accepted and becomes the initial override (the thinking level is a per-turn run parameter, unlike workspace/model, which stay rejected); `/clear` keeps both the flag default and the override.
- Selectable tiers reuse core's exported `DEFAULT_CHAT_THINKING_LEVELS` — no `none`, matching the web picker; a stored legacy `none` still displays as-is. Invalid values exit 1 before any Session exists (same contract as `--approve`).

**Collapsed tool output**
- Chat collapses long tool outputs by default: the first 4 lines stream live (progress stays visible), then a dim elision marker `… (+N lines, /verbose for full output)` and the last 4 lines print when the stream ends — the tail is where an exec result carries its verdict. Outputs of up to 9 lines render whole; the marker never hides fewer than 2 lines. Buffering is bounded (ring of tail+1 lines).
- `/verbose` toggles full output mid-chat; `penguin chat --verbose` starts with it. Resumed history (`--resume`) collapses by the same head/tail rule. A stream cut off by Ctrl-C still settles its held-back tail at task end, so nothing silently vanishes.
- Display-only truncation: the model, the Trace, and the Web App always receive the complete output. `penguin run` never collapses — its stdout feeds pipes and nested CLIs (`exec_command`-driven `penguin run`, #102 territory), where dropping lines would lose real data.

## Key changes by package

**packages/cli**
- `src/thinking-command.ts` (new): `parseThinkingLevel` / `parseThinkingCommand` (the `/thinking` line parser, `goal-command.ts` pattern), `resolveThinkingLevel` (flag validation, `resolveApprovalMode` contract), and `configuredThinkingLevel` (display mirror of core's resolution chain, structurally typed so test doubles stay small).
- `src/output-collapse.ts` (new): `ToolOutputCollapser` (streaming head/tail state machine with bounded memory) and `collapseLines` (complete outputs, for history rendering), plus the head/tail constants.
- `src/render.ts`: `StreamRenderer` takes `{ collapseToolOutput }` (default off, so `run` and every existing caller are unchanged) with a `setCollapseToolOutput` toggle; `handlePartialToolOutput` routes deltas through the collapser and settles marker+tail at `stop` (image placeholders settle held-back text first to preserve order); `endTask` settles leftover collapsers of aborted streams; `renderHistory` gains the same option for `--resume`.
- `src/task-loop.ts`: `RunTaskOptions.thinkingLevel` forwarded into `session.run` (omitted when unset, so core's construction default stays authoritative).
- `src/commands/chat.ts`: `--thinking` / `--verbose` options; `/thinking` and `/verbose` handled as instant idle-state commands (mid-run they remain steering text, like every slash command); flag → `createSession` for new and `/clear` Sessions, override → per-run option.
- `src/commands/run.ts`: `--thinking` → `createSession`.
- `src/i18n.ts`: new en/zh strings (`thinkingCurrent/Set/Invalid`, `verboseOn/Off`, `toolOutputElided`, option descriptions) and both `chatHints` lines now teach `/thinking` and `/verbose`.

**packages/docs**
- `content/cli.en.md` + `content/cli.zh.md`: `--thinking` in the run table, `--verbose` in the chat table, the `--resume` interaction, `/thinking` + `/verbose` rows in the REPL command table, and a paragraph on collapsed output (defaults, display-only guarantee, `run` exemption).

**changelog**
- `changelog/unreleased/2026-08-18-cli-thinking-and-output-collapse.md` + index line in `changelog/unreleased/README.md`.

## Tests

All run from the worktree; every suite green.

- `pnpm --filter @prismshadow/penguin-cli test` — **279 passed** (245 pre-existing + 34 new):
  - `test/thinking-command.test.ts` (new): level/command parsing (case, trailing tokens, `none` rejected) and the configured-chain display fallback.
  - `test/output-collapse.test.ts` (new): live head passthrough, chunk-straddled lines, the ≤ head+tail+1 whole-render rule, hidden ≥ 2 invariant, unterminated final line, bounded-ring drop counting, flush reset.
  - `test/render.test.ts` (extended): streaming collapse (exact output), short-output untouched, default renderer never collapses, `/verbose` toggle, endTask settling of aborted streams, diff coloring surviving collapse, `renderHistory` collapse on/off.
  - `test/chat-thinking.test.ts` (new, chat-clear harness): drives the real REPL — bare `/thinking` display via the config chain, override applied to subsequent `run` options (invalid input changes nothing), `--thinking` → `createSession` with no per-run override, `/clear` persistence, and collapsed→full switching across `/verbose` with a real 12-line streamed tool output (both flag polarities).
  - `test/task-loop.test.ts` (extended): `thinkingLevel` forwarded to `session.run`, omitted when unset.
  - `test/i18n.test.ts` (extended): en/zh parity of the new strings; both hint lines teach the new commands.
- `pnpm --filter @prismshadow/penguin-cli typecheck` — clean.
- `pnpm --filter @prismshadow/penguin-docs test` (32 passed) and `typecheck` — clean.
- Built-binary smoke: `chat --help` / `run --help` list the new flags; invalid `--thinking` exits 1 with the localized message in en and zh.

Fixes #305

## Design docs

`design/` is shared, so the spec edits are proposed here instead of committed. File: `design/specs/05-ARCHITECTURE.md`, § CLI.

1. Command prototypes (the ```bash block): extend the three lines —

   - `penguin chat …` (interactive line): append `[--thinking <low|medium|high|xhigh>] [--verbose]`
   - `penguin chat --resume …` line: append `[--thinking <low|medium|high|xhigh>] [--verbose]`
   - `penguin run …` line: append `[--thinking <low|medium|high|xhigh>]`

2. REPL commands paragraph (the one beginning `REPL 内 \`/goal[:<budget>] <objective>\``): before `` `/exit`（或 `/quit`）退出。`` insert:

   > `/thinking` 显示下一轮将使用的思考等级，`/thinking <low|medium|high|xhigh>` 覆盖本次对话后续轮次的思考等级（逐轮运行参数 `RunOptions.thinkingLevel`，口径同 Web 活跃会话选择器：不写回 Agent 配置，可选项不含 `none`；压缩请求不受覆盖影响）；`/verbose` 在折叠与完整工具输出之间切换；

3. New paragraph after the REPL commands paragraph:

   > `--thinking` 在创建 Session 时钉定其缺省思考等级（`createSession` 的 `thinkingLevel`，派生子会话随之）；未给出时按配置链取值（Agent `model.thinking_level` > Project `default_chat.thinking_level` > `medium`）。`--resume` 下思考等级作为逐轮参数仍可指定：`--thinking` 作为初始的 `/thinking` 覆盖生效。`penguin chat` 缺省折叠过长的工具输出：前 4 行照常流式打印，输出结束时打印暗色省略标记（另有 N 行）与最后 4 行；不超过 9 行的输出完整显示，省略标记至少隐藏 2 行；被中断的输出流在 Task 收尾时补印被扣留的尾部。这只是显示层截断——模型、Trace 与 Web App 收到完整输出。`/verbose` 与 `--verbose` 切换为完整输出；`--resume` 渲染历史按同一规则折叠；`penguin run` 从不折叠（其输出供管道与嵌套 CLI 消费）。

4. `--resume` paragraph: after `审批模式属运行期选项，可重新指定` append: `；思考等级同为逐轮参数，--thinking 作为初始覆盖生效`.


hiyouga:
## Review

Solid PR: both halves of #305 are delivered, the collapse logic is factored into a pure,
well-tested state machine, `--thinking` respects core's tri-state `createSession.thinkingLevel`
contract, and the display-only claim holds (the Trace, the model and the Web App still get the
full text — only `partial_tool_call_output` rendering is touched, and complete model messages are
never rendered by the CLI anyway). I stress-tested the collapser against the failure modes that
worried me and could not break it: one-character-per-delta chunking, a line split across the
head/hold boundary, an unterminated final line, two interleaved tool-output streams (the existing
holder lock serializes them and the collapsers are keyed by `tool_call_id`, so no cross-talk), an
abort with no `stop` (settled by `endTask`), mixed text/image deltas, and CJK. Terminal colour
state is also fine: `writeToolOutput` re-emits `dim label ->` + `reset` at every line start, so an
SGR opened in a head line and reset inside a hidden line can tint at most the marker's own gutter,
never leak past it. Errors never get swallowed either — Environment appends `[tool error]` /
`[tool timeout]` / `[tool aborted]` / exit-status notes as the *last* deltas of the stream
(environment.ts `noteSuffix`), so they always land inside the 4-line tail. The findings below are
one small correctness wart in the resumed-history path plus semantics/wording gaps around the
feature's central distinction.

### Findings

1. **Resumed history counts the output's trailing newline as a hidden line** (`packages/cli/src/render.ts:288`) —
   `(p.output ?? "").split("\n")` on a real tool output (exec_command yields the process bytes
   verbatim, so it almost always ends in `\n`) produces a trailing `""` element. The streaming
   collapser deliberately does *not* count that (`flush` only buffers `partial` when non-empty),
   so the same output renders as `… (+12 lines)` live and `… (+13 lines)` after `--resume`, and
   one of the four precious tail rows is spent on a blank gutter line. Fix: drop a single trailing
   empty segment in the collapse branch (leaving the uncollapsed path byte-identical).

2. **`/thinking` never says whether the level it prints is the Session default or an active
   per-turn override** (`packages/cli/src/commands/chat.ts:410`, `i18n.ts:532`) — that distinction
   is the whole subtlety of this feature (a `--thinking` on a fresh chat pins the Session and
   propagates to subagents; a `/thinking`, or `--thinking` under `--resume`, is a per-turn
   `RunOptions.thinkingLevel` that does neither), and it is explained in the docs but not where
   the user actually meets it. `[thinking] level: high — change with /thinking <…>` reads
   identically in both states. Fix: two messages — the default form, and an override form that
   also names the Session default it is overriding.

3. **The `/verbose` toggle lines read as if they act on what is already on screen**
   (`i18n.ts:536-537`, `792-798`) — "showing full tool output" / "long tool output is collapsed"
   are both timeless statements, but the toggle only affects output rendered from that point on
   (already-collapsed text cannot be re-expanded — the renderer no longer holds it). Fix: say
   "from here on" in both locales. Same file: zh translates the thinking label (`[思考]`) but
   leaves `[verbose]`; make the pair consistent.

4. **Test gap: the `--resume` + `--thinking` branch is untested** (`chat.ts:150-153`) — this is the
   trickiest line in the PR (the flag silently changes meaning from "creation default" to "initial
   per-turn override"), and `chat-thinking.test.ts` covers fresh-session, `/clear` and override
   paths but never resume. The fake agent's `resumeSession` currently returns `undefined`.

5. **Docs overstate subagent propagation by omission** (`packages/docs/content/cli.*.md`) — the
   `--thinking` row correctly says the level is pinned at Session creation "so spawned subagent
   sessions follow it", but the `/thinking` row is silent about the flip side: core passes the
   *construction-time* level into `createSession` at subagent spawn (`core/src/agent.ts:659`), so a
   `/thinking` override (and `--thinking` under `--resume`) applies to this Session's own turns
   only — subagents keep the creation-time level. One clause in each locale.

### Reported, not fixed

- **A single very long line is never collapsed.** The head allowance is spent by newlines only, so
  a 5 MB minified-JSON blob on one line is genuinely "1 line" and streams through in full — the
  one flood shape this PR does not cover. Not a bug (the collapse is line-based by design, and the
  memory story is fine: a long line landing in the hidden region is dropped, and core already holds
  the whole output for the model anyway), but a byte cap per rendered line — the repo already does
  this at `tool-render.ts:359`, `MAX_APPROVAL_PAYLOAD_LINE = 200` — would be a worthwhile
  follow-up. Out of scope here: truncating content mid-line needs its own marker wording and
  i18n, and it is a separate product decision.
- **`configuredThinkingLevel` is duplicated** in `cli/src/thinking-command.ts` from core's private
  `Agent.configuredThinkingLevel` (`core/src/agent.ts:176`); the copy is correct today and the
  comment names the sites to keep in sync. Promoting core's method to public would remove the drift
  risk, but that is a public SDK-surface change on `Agent` and would need a `design/specs` update,
  which this round cannot make (shared read-only repo). Proposing it here instead.
- `ToolOutputCollapser`/`collapseLines` with `tail = 0` would hit `slice(-0)` and print the whole
  buffer. Unreachable — production only ever constructs with the defaults — so left alone rather
  than adding a branch for a caller that does not exist.
- Pre-existing, untouched: the approval-payload elision note at `tool-render.ts:361`
  (`[… N more lines not shown]`) is hardcoded English, unlike the new `toolOutputElided`. Worth a
  separate cleanup.
- `setCollapseToolOutput` flipping mid-stream would interleave live and held lines. The REPL only
  accepts commands while idle, so it cannot happen; the existing comment says so and I left it as
  documentation rather than adding a defensive settle for an unreachable state.

### Verified, no change needed

- The override is process-local and is deliberately not persisted: a later `--resume` starts from
  the configured chain again, matching core, which stopped recording `thinking_level` in
  `session_meta` precisely because it became a per-turn parameter (`agent.ts:435-441`). The CLI's
  display fallback resolves to the identical chain, so `/thinking` on a resumed session tells the
  truth.
- An invalid level in the REPL prints an error and leaves the previous override intact (no crash,
  no session loss); an invalid `--thinking` exits 1 on stderr, matching `resolveApprovalMode`.
- The marker's `/verbose` advice never appears where the command does not exist: collapsing is off
  by default in `StreamRenderer`/`renderHistory`, and `penguin run` never enables it.
- Collapsing reaches tool output only — assistant text, approval payload previews, approval
  prompts, abort/steering/stats lines and subagent (`origin`-tagged) rendering are all untouched.

None was given for this PR; I am applying findings 1-5 above.

hiyouga:
Applied in 96e2bd68:

- **`/thinking` now names the source of the level it prints** — this Session's own default, or an active per-turn override (which also names the default it replaces). New `thinkingCurrentDefault` / `thinkingCurrentOverride` messages in both locales; `thinkingSet` also states that the Agent config is untouched.
- **Resumed history stops counting the trailing newline as a hidden line** (`render.ts:286`): a newline-terminated output now reports the same hidden count on `--resume` as it does live, and no longer spends a tail row on a blank gutter line. The uncollapsed path is byte-identical to before.
- **`/verbose` on/off lines say "from here on"** — the renderer cannot re-expand output that is already printed, so a timeless "showing full tool output" was misleading; the zh label is now translated (`[详细输出]`), matching `[思考]`.
- **Tests**: `--resume` with and without `--thinking` (the branch where the flag silently becomes the initial per-turn override — previously uncovered, the fake agent's `resumeSession` returned `undefined`), the override form of the `/thinking` display, and a newline-terminated resumed output cross-checked against the live stream.
- **Docs (en+zh) + changelog entry** updated for the display change and to record that a `/thinking` override reaches this Session's turns only — subagents are spawned with the level the Session was created with (`core/src/agent.ts:659`).

Deliberately not changed:

- **The elision marker text** stays `… (+N lines, /verbose for full output)`. It cannot be made literally retroactive, but it only ever renders where `/verbose` exists, and the toggle's own reply now states the scope — reworking the marker to hedge would cost length on a line that has to stay scannable.
- **The long-single-line gap** (a 5 MB blob on one line is genuinely "1 line" and streams in full): a per-line byte cap is a separate product decision needing its own marker wording and i18n. Filed in the review as a follow-up.
- **The duplicated `configuredThinkingLevel`**: de-duplicating means promoting core's private `Agent` method to public SDK surface, which needs a `design/specs` change this round cannot make.
- `tail = 0` `slice(-0)` in the collapser and the pre-existing hardcoded-English elision note at `tool-render.ts:361`: both unreachable/out of scope here, noted in the review.

CLI: 282 tests pass, `tsc --noEmit` clean; docs package: 32 tests pass. Prettier clean (`*.md` is prettier-ignored in this repo).

Issue #171, opened 2026-08-04
Goal mode aborts instead of continuing after a Task reaches max_turns


zzzbitz:
## Problem

In Goal mode, if the current Task reaches `max_turns`, the Goal immediately ends as `aborted` and no new Task is started, even when the Goal has no token budget limit.

Current behavior:

```text
Task 1 reaches max_turns
→ final message has stop_reason: failed
→ goal_finished(outcome: aborted)
```

This prevents Goal mode from helping the long-running tasks that most need continuation.

## Expected behavior

Keep `max_turns` as the limit for one Task, but treat reaching it as a round boundary:

```text
Task 1 reaches max_turns
→ preserve workspace and progress
→ start Task 2 for the same Goal
```

The Goal can still be bounded by its token budget, `maxRounds`, user aborts, and no-progress protection.

## Relevant code

`packages/core/src/goal/goal-loop.ts` currently treats a max-turn cutoff as terminal via `roundFailed` and returns `goal_finished("aborted")`.

PR #323, opened 2026-08-19
feat(core): optional `thinking_level` argument on run_subagent

### LINKED ISSUE #306: support specify subagent thinking level
在启动subagent相关工具的时候，给模型一个设置思考强度的选项

hiyouga:
## What & why

`run_subagent` gains an optional `thinking_level` argument so the calling model can pick the thinking level a subagent runs at — lower for cheap mechanical subtasks, higher for hard analysis — instead of every child Session always running at the parent's level. Omitting the argument keeps today's behavior bit-for-bit: the child inherits the parent Session's effective level (including "no level" when the parent has none, the tri-state `null`).

Fixes #306

### packages/core

- `interfaces.ts`: new exported constant `SUBAGENT_THINKING_LEVELS` (`low` / `medium` / `high` / `xhigh`) — the four selectable tiers, mirroring the user-facing pickers (`none` stays a valid stored/wire value but is not offered, because many models cannot disable thinking; a parent genuinely running without a level still passes that down via inheritance). `SubagentRunner.spawn` gains an optional `thinkingLevel?: ThinkingLevelName`.
- `state/default-config.ts`: the default `run_subagent` tool entry declares `thinking_level` as a string enum with guidance ("Omit to inherit the current session's level. Lower it for cheap mechanical subtasks; raise it for hard analysis or design subtasks."), placed before `yield_time_ms`; the tool description now names the option alongside `agent_id` / `model_id`.
- `environment/tools/run-subagent.ts`: validates the argument — an unknown value fails loudly, listing the valid tiers, rather than silently running the child at an inherited level the caller did not ask for (a silent fallback would defeat the override); JSON `null` counts as omitted. A valid level is forwarded to `runner.spawn`.
- `agent.ts`: the spawn closure resolves `spawnThinkingLevel ?? parentEffectiveLevel ?? null` into the child's `createSession`, so an explicit level also applies when the parent itself has no level; comments updated at the three sites that describe the inheritance.
- `state/kernel-history.ts`: the default `run_subagent` entry changed, so `KERNEL_VERSION` advances to `2026-08-18` with a new pinned-hash generation (only the `tools.builtin.run_subagent` leaf differs from `2026-08-11`). Existing Agents whose entry still matches an older default pick the new schema up through the normal kernel update; customized entries are kept, as usual.
- `test/kernel-version.test.ts`: the pre-toggles reconstruction proof (`it.runIf(KERNEL_VERSION === TOGGLES_GENERATION)`) self-retires now that the defaults evolved past the toggles generation — exactly as its comment planned ("Once the defaults evolve past it this proof self-retires — the pinned hashes remain the frozen source of truth on their own"); the test is removed, the `preTogglesDefaultConfig` helper stays (still seeds the smart-merge scenarios) with its doc updated to say the tool leaves now carry current defaults.

### packages/docs

- `tools.{en,zh}.md` § Subagents: the `run_subagent` argument block and the follows-the-parent bullet document `thinking_level` (with the lower/higher guidance).
- `interfaces.{en,zh}.md` § Subagent interfaces: the `SubagentRunner.spawn` snippet gains `thinkingLevel` — and catches up on the `provider` half of the model pair it had been missing since #171-era pairing rules.

### changelog

- `changelog/unreleased/2026-08-18-subagent-thinking-level.md` + index line at the top of `changelog/unreleased/README.md`.

### Not included (deliberate)

- No web UI surface: the thinking level is a per-run parameter, deliberately absent from `session_meta` (invariants only), and the agents-panel subagent nodes show agent name + call description with no model/level row today — there is no obvious existing place to show it, so per the task this nice-to-have is skipped.
- The four-tier enum intentionally excludes `none`, matching `SELECTABLE_THINKING_LEVELS` (web picker) and `DEFAULT_CHAT_THINKING_LEVELS` (project defaults): pickers stopped offering `none` because many models cannot disable thinking; a spawn override is the same kind of surface.

## Tests

- `core/test/subagent.test.ts` (tool layer): pass-through of a valid `thinking_level` to `spawn` (extended spawn-args test); loud rejection of unknown values (`"hgih"`, `"none"`, `""`, `3`) with no spawn; omitted/`null` forwards no level (inheritance left to the runner).
- `core/test/agent.test.ts` (runner layer): an explicit spawn level pins the child over the parent's effective `"high"`; an explicit level also applies when the parent has no level at all (tri-state `null` parent); existing inheritance tests unchanged and green.
- `core/test/kernel-version.test.ts`: pinned-hash guard green for the new generation.

Results:

- `pnpm --filter @prismshadow/penguin-core test`: 799 passed, 5 skipped (e2e, needs keys), 0 failed.
- `pnpm --filter @prismshadow/penguin-server test`: 638 passed, 0 failed (a first run under parallel-agent machine load flaked 6 tests in 5 files; a clean rerun is fully green — none touch this change).
- `pnpm --filter @prismshadow/penguin-docs test`: 32 passed.
- CLI suite: 222/222 tests pass; one test *file* fails to collect because the workspace `@prismshadow/penguin-server` dist wasn't built in a fresh worktree (import of `@prismshadow/penguin-server/...` subpaths) — resolved by building server first; unrelated to this change.
- Typecheck: core clean except pre-existing `test/builtin-agents.test.ts` errors (`InstalledSkill`-related; identical on the base commit, verified via `git stash`); server and docs clean; CLI clean once the server dist exists.

## Design docs

`design/` is shared and read-only for this task; proposed edits:

1. `design/specs/05-ARCHITECTURE.md`, § Subagent, the derivation paragraph (currently ending "……未指定时跟随父会话的 (provider, model_id)，Project 默认 Model 不再参与派生回退。子会话同时跟随父会话的 thinking level 与 Workspace。") — replace the last sentence with:

   > 子会话同时跟随父会话的 thinking level 与 Workspace；thinking level 亦可在派生时经 `thinking_level` 参数显式指定（仅 `low` / `medium` / `high` / `xhigh` 四个可选档，与思考等级选择器同口径、不提供 `none`；非法值即拒绝，缺省即跟随父会话——含父会话无等级时的「无等级」）。

2. `design/specs/08-CHANGELOG.md` — new entry at the top:

   > - 2026-08-18 run_subagent 增补可选 `thinking_level` 参数：模型可为子会话显式指定思考等级（仅 low / medium / high / xhigh 四档，与选择器口径一致、不提供 none；非法值拒绝），缺省仍跟随父会话的有效等级（含无等级）；内置默认工具条目随之变化，配置内核代次推进至 2026-08-18（仅 tools.builtin.run_subagent 叶子变化）。见 05「Subagent」（#306）。


hiyouga:
## Review

The feature itself is right and small: the override sits above the inherit-or-null resolution, the enum mirrors the pickers' selectable tiers exactly as `design/specs/06-PROTOTYPE.md` describes them (`可选项仅 低 / 中 / 高 / 极高`, `none` 仍是合法的存量配置值、仅不再作为可选项), fail-loud validation matches how this codebase treats every other semantics-selecting argument, and the tests read the real captured LLM config rather than asserting call shapes. I re-verified the kernel bump end-to-end against a real on-disk config rather than by reading: **the migration is safe in all four scenarios**, and the newly pinned `run_subagent` hash is genuinely the new default while the frozen `2026-08-11` hash is genuinely the old one. Three things to fix: a design-spec disagreement, a deleted test whose retirement condition was keyed to the wrong thing, and a code path that this PR makes reachable for the first time and that nothing tests.

### Findings

1. **The design spec pins the behavior this PR changes, and was not updated** (`design/specs/05-ARCHITECTURE.md:533`). The Subagent section states the rule unconditionally: *"派生时可显式指定完整的 `(provider, modelId)` 模型对……**子会话同时跟随父会话的 thinking level 与 Workspace**"*, and `createSession`'s `thinkingLevel` is described at `:95` as existing *"供子会话派生跟随父会话"*. `CLAUDE.md` makes the design docs ground truth ("when code and design disagree, the design docs win"), so this needs a spec change, not just code. `design/` is a shared read-only symlink here, so proposing the wording instead — suggested amendment to the second half of `:533`:

   > 派生时另可显式指定子会话的 thinking level（`run_subagent` 的 `thinking_level` 参数，可选值与思考等级选择器同口径：低 / 中 / 高 / 极高，`none` 不作为可选项）：显式值优先，未指定时子会话跟随父会话的 thinking level（父会话无等级时同样跟随「无等级」，不回退子 Agent 配置）；非法值即拒绝该次工具调用，不静默回退继承。Workspace 始终跟随父会话。

   `:546` ("`run_subagent` 工具定义保持稳定，只描述参数、审批和权限约束") stays true as written and needs no change.

2. **The retired reconstruction proof still holds — its guard was keyed to the wrong thing** (`packages/core/test/kernel-version.test.ts:95-99`, deleted block). The original guard was `it.runIf(KERNEL_VERSION === TOGGLES_GENERATION)`, which retires the proof as soon as *any* default leaf changes. But `preTogglesDefaultConfig()` only rewrites `system_prompt` (LEGACY_* sections swapped back in, `{{SCHEDULES}}` dropped) and removes the three toggle sections — every other leaf it emits is the *current* default verbatim. This PR changed only `tools.builtin.run_subagent`, so the reconstruction is still byte-exact for `system_prompt`, which is the only leaf the recipe is actually about. The removal follows the letter of the old comment but drops coverage that is still live today, and will stay live through every future generation that does not touch the prompt template. Fix: restore the proof, guarded by the condition it actually depends on (current `system_prompt` hash still equals the toggles generation's) and asserted over the leaves that are still reconstructible, rather than by kernel version.

3. **The old-generation *advance* path for a tool entry is untested — and this PR is the first time it can ever fire** (`packages/core/src/state/kernel-update.ts:170-175`). In the per-tool merge branch, `historicalHashes(path).has(storedHash) → mergedTools[index] = defaultValue` was unreachable dead code until now: every recorded generation pinned byte-identical hashes for all ten tool leaves, so a stored tool entry could only ever be *equal to current* (skip) or *unrecognized* (kept). `2026-08-18` makes `tools.builtin.run_subagent` the first leaf that differs between generations, so that line is now the exact path every existing user's config takes to pick up the new schema — the PR's headline migration promise — and no test exercises it. The existing "appends a tool only when it is new in the latest generation (history seam)" test covers only the *absent-entry* branches. Worth adding, via the same `options.history` seam so it cannot rot.

4. **Migration safety — verified, no action.** I seeded real configs on disk and ran `applyKernelUpdate` for each case:
   - *(a) untouched*, stamped `2026-08-11`: `advanced: ["tools.builtin.run_subagent"]`, `kept: []`; the entry becomes the new default verbatim and the file is re-stamped `2026-08-18`. Idempotent on a second run.
   - *(b) user edited the entry*: `kept: ["tools.builtin.run_subagent"]`, `advanced: []`; the custom wording survives untouched and the file is still re-stamped. The user simply does not get `thinking_level` — correct and conservative.
   - *(b') user disabled the tool*: there is no `enabled` flag on `ToolDefinitionConfig`; disabling is deleting the entry. `newInLatestGeneration("tools.builtin.run_subagent")` is `false` (the path is recorded in all three generations), so the removal is preserved and the tool is **not** silently re-added. This is the case that would have been a real bug if the leaf had been treated as new, and it is handled.
   - *(c) a config several generations back* (no stamp at all, pre-`2026-08-10` shape): the old entry still hash-matches, so it advances, alongside the `vault`/`skills`/`schedules` sections being materialized.

   I also confirmed the frozen `2026-08-11` hash `c3f93add…` is exactly `sha256(canonical(current entry − thinking_level property, with the old description sentence))` — i.e. the old pin is the genuine previous default and the only two changes to the entry are the ones the diff shows. That is the assertion that matters most here: a wrong old hash would silently turn every user's untouched entry into a "customization" and freeze it forever. The new pin is proven by the guard test itself (it recomputes all 26 leaves), and the 25 unchanged leaves in the new generation are byte-identical copies of `2026-08-11`'s, as they must be.

5. **Fail-loud validation — verified consistent, no action.** I surveyed every builtin tool's argument handling. The codebase draws a clean line: soft knobs fall back silently (`yield_time_ms` via `clampYield`, `describe_image`'s `prompt`, `edit_file`'s `replace_all`, a non-string `agent_id`/`workdir`), while anything that *selects semantics* fails loudly — required strings, valid-shaped-but-unknown ids (`unknown subagent_id`, `unknown process_id`, `agent "x" does not exist`), the `model_id`/`provider` pair, `input_command`'s ETX-mixing check, `read_image`'s unsupported-mime rejection (structurally an enum rejection: echoes the bad value, enumerates the legal set), and `read_file`'s `Invalid "offset": expected a number (got …)`. A thinking level is semantics selection, and there is **no** schema-validation layer between the LLM and `execute` (`environment.ts:253-262` is a bare `JSON.parse`; no ajv/zod anywhere in core) — so if the tool does not check the enum, nothing does. Falling back to inheritance would run the child at a level the caller did not ask for and report success, which is the failure mode `input_command`'s comment explicitly rejects. The message reaches the model correctly: the `fail` delta is accumulated by Environment and finalized into a real `tool_call_output` (`environment.ts:492-498`) with `stop_reason: "failed"`, it echoes the offending value via `JSON.stringify`, lists the four legal values, and names the omit-to-inherit escape hatch — a self-correctable error.

6. **Inheritance semantics — verified, no action.**
   - *Parent at `"none"`*: reachable, since `system_config.yaml`'s `model.thinking_level` is typed `ThinkingLevelName` (includes `"none"`) even though the pickers no longer offer it. With no override it is inherited verbatim; with an override the child is pinned. Correct — and note the asymmetry is deliberate: the model can raise a `"none"` parent's child but cannot ask for `"none"` itself, matching the picker rule.
   - *Parent with no level* (tri-state `null`): `spawnThinkingLevel ?? thinkingLevel ?? null` puts the override above the resolution, so it applies. Locked by the second new test.
   - *Nested subagents*: moot today and correctly so — `MAX_SUBAGENT_DEPTH = 1`, and `agent.ts:719` filters `run_subagent` out of a depth-1 child's toolset entirely, so a subagent never even sees the tool. If the cap were ever raised, the override becomes the child's own effective level and would propagate onward from there, which is the right semantic (one hop of override, then normal inheritance resumes).
   - *Per-model thinking support*: there is no such notion anywhere in the codebase to interact with — `ModelEntry` has `vision`, `max_tokens`, `context_window`, `pricing`, but no thinking-capability flag, and `mapThinkingLevel` translates unconditionally. So the override is exactly as safe (or unsafe) against a non-thinking model as the existing per-session picker and the Agent config already are. Nothing to gate here; gating would be a separate issue about `ModelEntry`.

7. **Observability — verified, no action needed.** Skipping the UI is right, and the trace does record it, exactly: the `run_subagent` `tool_call` with its arguments is written to the **parent's** Trace, so the chosen level is auditable at the one place the decision was made. `session_meta` is deliberately the wrong home — `agent.ts:323-326` and `design/specs/05-ARCHITECTURE.md:647` both state it holds session invariants only and that the thinking level became a per-turn run parameter (`resumeSession` even ignores a legacy `thinking_level` in old meta JSON on purpose). Adding it back for subagents would reintroduce exactly the field that was removed. If anything is ever wanted in the web subagent panel, `features/chat/agent-topology.ts` already parses `agent_id` and `description` out of the same tool-call arguments, so it is a pure frontend follow-up with no core change — worth a separate issue at most.

8. **Polish: the selectable-tier list now exists three times** (`interfaces.ts:248` `SUBAGENT_THINKING_LEVELS`, `state/project-config.ts:120` `DEFAULT_CHAT_THINKING_LEVELS`, `web/features/chat/thinking-level.ts:30` `SELECTABLE_THINKING_LEVELS`) with nothing checking they agree; the web one is pinned by a test, the two core ones are not. Deduplicating properly would mean moving the canonical list and renaming a public export, which is more churn than this PR deserves — but the invariant the doc comment claims ("mirroring the pickers", never `"none"`) should at least be pinned by a test so the lists cannot drift apart silently. Also minor: `SUBAGENT_THINKING_LEVELS` is typed `readonly ThinkingLevelName[]`, one notch wider than the values it holds, which is why `run-subagent.ts:113` needs an `as` cast after the check.

9. **Nit, not fixing:** `interfaces.ts` was a pure type-contract module before this PR (every import is `import type`, zero runtime exports); `SUBAGENT_THINKING_LEVELS` makes it emit a runtime module. There is no cycle risk (it imports nothing at runtime) and no better home without the rename in (8), so this is only worth knowing about.

Docs (en+zh both updated, including the `provider` line the `SubagentRunner` snippet had been missing), the changelog entry and its index line, and English-only code all check out.

None was given for this PR. Applying findings 1-3 and 8: restoring the reconstruction proof under a guard keyed to the prompt template, adding coverage for the old-generation tool advance path and for the tier-list invariant, and updating the changelog sentence that claims the proof self-retired. Finding 1 is a design-doc change I cannot make from here (read-only shared symlink) — the proposed wording is above for a maintainer to land in `penguin-harness-design`.

hiyouga:
Applied in 481aae15:

- **Restored the pre-toggles reconstruction proof** (`packages/core/test/kernel-version.test.ts`) under the condition it actually depends on. The old `it.runIf(KERNEL_VERSION === TOGGLES_GENERATION)` guard retired the proof whenever *any* default leaf moved, but the recipe only rewrites `system_prompt` and strips the three toggle sections — so retirement belongs per leaf, not per generation. It now proves 20 of the 21 pre-toggles leaves, dropping only `tools.builtin.run_subagent`; `system_prompt`, the leaf the frozen `LEGACY_*` sections are actually about, stays proven, and an explicit `expect(reconstructible).toContain("system_prompt")` stops it from silently going vacuous if the template itself ever moves.
- **Covered the tool-entry advance path** (`kernel-version.test.ts`). `kernel-update.ts:172-175` — a stored tool entry that hash-matches an older generation being replaced by the new default — was unreachable until this PR, because every recorded generation pinned byte-identical hashes for all ten tool leaves. It is now the path every existing Agent takes to pick up `thinking_level`, so it is tested, driven through the existing `options.history` seam so it cannot rot when the real generations move on.
- **Pinned the tier-list invariant** (`packages/core/test/state.test.ts`): `SUBAGENT_THINKING_LEVELS` equals `DEFAULT_CHAT_THINKING_LEVELS`, never contains `"none"`, and is the set the shipped `run_subagent` schema declares. Three parallel copies of that list now exist (the third is the web picker's, pinned by its own test) and nothing checked the two core ones against each other.
- **Changelog** updated to match: the self-retirement sentence was wrong as written, and the entry now also states the removal case (an entry deleted to switch the tool off stays deleted) and the newly reachable merge branch.

Deliberately not changed:

- **No production-code change.** The migration is correct as written — I verified it against real on-disk configs rather than by reading, in all four scenarios (untouched → advances and re-stamps; edited → kept verbatim; deleted → stays deleted; unstamped/older → advances), and confirmed the frozen `2026-08-11` hash is genuinely the previous default entry. Fail-loud validation, the `spawnThinkingLevel ?? thinkingLevel ?? null` ordering, and the choice to skip the UI surface are all right as they stand — details in the review above.
- **`design/specs/05-ARCHITECTURE.md:533`** still says 子会话同时跟随父会话的 thinking level 与 Workspace unconditionally, which this PR makes untrue. `design/` is a shared read-only symlink from here, so the proposed replacement wording is in the review comment for a maintainer to land in the design repo before this merges.
- **Deduplicating the three tier lists** for real would mean moving the canonical constant out of `interfaces.ts` and renaming a public export — more churn than this PR's subject deserves. Pinned by test instead; worth a follow-up issue.
- **`interfaces.ts` becoming a runtime module** (it was pure `import type` before `SUBAGENT_THINKING_LEVELS`) — no cycle risk, and no better home without the rename above.

Verification: `packages/core` typecheck clean; full core unit suite 803 passed / 3 skipped (40 files); `packages/server` `agent-kernel-update.test.ts` green; prettier clean on touched files.

hiyouga:
One addendum from a last pass over the config read/write paths — **not a defect in this PR, and not fixed here**, but this PR is what makes it reachable, so it should not go unrecorded.

I checked whether the new JSON-schema `enum` key (the first one any builtin tool entry has ever carried) survives every path. It does: `validateToolsBuiltin` (`packages/server/src/services/agent-config-service.ts:465-494`) only inspects `name`/`description`/`permission`/`forModel`/`call_description`/`timeoutMs`/`maxOutputLength` and never touches `parameters`; `applyCallDescriptionToggle` (`packages/core/src/state/agent-state.ts:775-788`) shallow-spreads, keeping each property object by reference; `toolDefinitionsToSchemas` (`packages/core/src/llm/generative-model.ts:1393-1399`) passes `parameters` through opaquely; and the YAML round trip through `parseDocument`/`setIn` serializes the nested string array correctly. The trace viewer even renders it already (`trace-event-row.tsx:238-243`).

The one path that can lose it is the Web Tools tab. `agent-settings-page.tsx:914-920` snapshots the whole tool entries into `useState` at mount and never re-seeds while mounted, and Save submits each entry whole (`:943-945`, `{ ...row.base }`). So: open an Agent's settings on the Tools tab → the kernel update runs (from the agents list, another tab, the CLI, a hand edit) → flip a permission or timeout on the still-open stale tab → Save writes the pre-`2026-08-18` `run_subagent` entry back over the migrated one.

What makes that sticky rather than self-healing: the config PUT does not rewrite `kernel_version` (`applyConfigUpdate` sets only the fields it was given), so the file ends up stamped `2026-08-18` while carrying the old entry. `isKernelOutdated` is then false, the "Kernel update available" badge (`agents-page.tsx:263`) never reappears, and that Agent silently never gets `thinking_level` — even though the merge logic would happily advance it if it were ever asked again.

The hazard is pre-existing and generic (it applies to any out-of-band edit to `parameters`, and to `description` for that matter), but every kernel generation so far only changed `system_prompt` and the toggle sections — nothing the Tools tab round-trips. `2026-08-18` is the first one that changes a tool entry, which is why it starts to bite now.

Deliberately not fixing it in this PR: it lives in `packages/web` (and arguably in the PUT contract — whether a client's copy of `parameters` should be trusted over what is on disk), it is out of this PR's subject, and the sensible fix is a judgement call between re-seeding the form on external change and having the form submit only the fields it actually owns. Worth its own issue.

PR #324, opened 2026-08-19
feat(models): auto-detect custom model protocol from the base URL

hiyouga:
## What & why

AgentHub 0.4.2 ships three generic protocol clients — `openai-responses` (OpenAI Responses API), `ant-messages` (Anthropic Messages API), and `openai-chat` (Chat Completions; renamed from the old generic `openai` client, which remains an alias). Until now, custom and user-defined groups were pinned to `client_type = "openai"` with no choice exposed, so an Anthropic-protocol or Responses-only endpoint could not be added as a custom model at all.

This PR makes the protocol AUTO-DETECTED from the base URL when adding or editing a custom model, in the required order **openai-responses → ant-messages → openai-chat** (first served protocol wins), surfaces the result in the dialog, and stores it as the entry's `client_type` — with a manual protocol selector as override. The probed URLs and auth headers are exactly what the AgentHub clients construct (verified against the wrapped OpenAI / Anthropic SDKs: `POST {base}/responses` and `POST {base}/chat/completions` with `Authorization: Bearer`; `POST {base}/v1/messages` with `x-api-key` + `Authorization: Bearer` + `anthropic-version: 2023-06-01`), so a detected protocol is one that will actually work after saving.

## Key changes by package

**server**
- New `packages/server/src/services/protocol-detect.ts`: sequential probes in the required order, stop at first hit, 5s timeout each, injectable `fetchImpl`. Each probe POSTs a minimal invalid `{}` body — zero tokens billed, no model id needed, and keyless probing works. Classification: 404/405 → `route_missing` by status alone (OpenAI-style servers wrap unknown paths in protocol-shaped 404 JSON, and with no model in the body a 404 can never mean model-not-found); other 4xx with a structured API-error JSON (tolerant of OpenAI / Anthropic / vLLM `object:"error"` / FastAPI `detail` dialects — the probed PATH is the protocol discriminator) → `served`, auth failures included; 2xx needs protocol-specific success markers (guards catch-all gateways); 5xx → `server_error` (proves nothing about the path); HTML / non-JSON → `junk`; fetch failures → `timeout` / `network_error`. The API key travels only in headers — never in URLs, results, or logs.
- New route `POST /api/projects/:projectId/models/detect` (owner) in `routes/models.ts`, following the `/test` conventions: hand-rolled validation (`baseUrl` required, absolute http(s), max 2000 chars; optional `apiKey` / `clearApiKey` / paired `provider`+`modelId`). `ProjectConfigService.detectProtocol` resolves the probe credential like `testModel` does: body key wins, else the stored entry's key when the paired reference names one and "clear" isn't checked.
- DTOs `ModelProtocolDetectRequest` / `ModelProtocolProbeDto` / `ModelProtocolDetectResponse` (`detected?` + per-probe outcomes for debugging) in `src/api/types.ts`.

**web**
- `models-page.tsx`: custom / user-defined groups get a protocol row below the base URL — a dropdown with the three protocols plus an "Auto-detect" button in the label row (same link-next-to-label idiom as "get API key"). Detection runs automatically when the base URL field blurs, but only for a probeable URL the user actually changed in this dialog (a click-through must never rewrite a working entry's protocol); results show inline ("Detected OpenAI Responses; applied" / none-matched / failure notices), and a monotonic run counter makes manual picks supersede in-flight detections. Legacy `openai` values display as Chat Completions without being rewritten until the user picks or detection applies; vendor-pinned client types outside the generic family keep the existing read-only note. `clientTypeAfterProviderChange` now keeps generic protocol types when moving a model to Custom. New exported pure helpers: `PROTOCOL_CLIENT_TYPES`, `isGenericProtocolClientType`, `protocolSelectorValue`, `detectableBaseUrl`.
- `protocol-path.ts`: the in-field grey suffix maps `openai-responses` → `/responses` and `ant-messages` → `/v1/messages` (matched before the bare `openai` substring, mirroring AutoLLMClient's routing order), so the suffix tracks the selector live.
- `endpoints.ts`: `detectProtocol()`. Strings in both locales (`strings.ts` / `strings-en.ts`): protocol label + names, detect button/progress/result/failure texts, custom add-dialog hint and title, updated user-group description.

**core**
- `resolveModelEnv` routes `ant-messages` to `ANTHROPIC_API_KEY` / `ANTHROPIC_BASE_URL` (placed before the `openai` substring match, mirroring AutoLLMClient; `openai-responses` / `openai-chat` already resolve `OPENAI_*`), so the dialog's env-fallback hint follows the resolved client per PRN-021. `ModelEntry.client_type` doc comment names the generic protocol clients. Custom entries already persisted `client_type` and passed it through `agent.ts` → `GenerativeModel` → `AutoLLMClient` unchanged — verified, no plumbing changes needed.

**docs / changelog**
- `models.{en,zh}.md`: new "Protocol auto-detection for custom models" section, updated `client_type` row and the vLLM recipe; `configuration.{en,zh}.md` `client_type` row; `server-api.{en,zh}.md` route table row. Changelog entry `changelog/unreleased/2026-08-18-custom-model-protocol-detect.md` + index line.

## Tests

- `packages/server/test/protocol-detect.test.ts` (new, 18 tests): classification matrix (route-exists vs 404/405 vs HTML/junk vs 5xx vs shapeless JSON, protocol-shaped 404s stay missing, vLLM/FastAPI dialects count as served), probe ordering and stop-at-first, probe failures not aborting the sequence (timeout / network-error fall through), per-protocol auth header shapes and `{}` bodies with the key never in URLs, keyless probing, URL joining; route integration against local `node:http` endpoints — validation 400s, ant-messages detection with per-probe outcomes, key never echoed in the response, stored-key fallback via the paired reference and `clearApiKey` suppressing it.
- `packages/web/test/protocol-detect.test.ts` (new): detection-order constant, generic-family membership, selector value mapping, probeable-URL gating, `clientTypeAfterProviderChange` keeping the trio, protocol-path additions with legacy branches intact.
- `packages/core/test/model-catalog.test.ts`: `resolveModelEnv` cases for the three generic client types.

Results: server suite ✓ (all files, including models.test.ts), web suite ✓ (61 files / 729 tests), core suite ✓ (full run), docs test ✓; `typecheck` ✓ for core, web, server, docs; prettier run on all touched files.

## Design docs

The current spec pins custom / user-defined groups to one protocol; this PR amends that. Proposed edits (not applied — `design/` is shared):

- `specs/01-PRINCIPLES.md` · PRN-021, third bullet — replace the tail "显式 `client_type` 优先于 id 路由——网关分组（OpenRouter / Fireworks AI / SiliconFlow / Qwen Token Plan / Qwen Pay-As-You-Go）与 custom 的模型 `client_type = "openai"`，一律走 OpenAI 客户端、读取 `OPENAI_API_KEY` / `OPENAI_BASE_URL`。" with:
  "显式 `client_type` 优先于 id 路由——网关分组（OpenRouter / Fireworks AI / SiliconFlow / Qwen Token Plan / Qwen Pay-As-You-Go）固定 `client_type = "openai"`、走 OpenAI 客户端；custom 与自建分组的模型使用通用协议客户端（`openai-responses` / `ant-messages` / `openai-chat`，旧值 `openai` 仍为 `openai-chat` 的别名），环境变量按所选客户端读取：OpenAI 系协议读 `OPENAI_API_KEY` / `OPENAI_BASE_URL`，`ant-messages` 读 `ANTHROPIC_API_KEY` / `ANTHROPIC_BASE_URL`。"
- `specs/01-PRINCIPLES.md` · PRN-021, fourth bullet — replace "自建分组与 custom 同语义（OpenAI 协议缺省、base URL 必填）" with "自建分组与 custom 同语义（通用协议客户端、协议按 base URL 自动检测或手动选择、base URL 必填）".
- `specs/06-PROTOTYPE.md` · 模型配置页 · 「新增模型入口」bullet — replace "custom / 自建分组 / 网关固定 OpenAI chat completion（client_type=openai）" with:
  "网关固定 OpenAI chat completion（client_type=openai）；custom / 自建分组使用通用协议客户端，接口协议按 base URL 自动检测——base URL 失焦（仅当该 URL 是本次弹窗内修改过的）或点协议行的「自动检测」触发 `POST /api/projects/:p/models/detect`，按 openai-responses → ant-messages → openai-chat 顺序探测、取第一个真实提供的协议写入 client_type；协议下拉框即手动覆盖，检测结果/未识别/失败提示内联在协议行下方".
- `specs/06-PROTOTYPE.md` · 模型配置页 · 连通性测试 bullet 之后追加一条：
  "协议检测（`POST /api/projects/:p/models/detect`，owner，可带尚未保存的 apiKey；带成对引用时回落既存 key）：逐个发最小非法请求（`{}` 请求体、每个 5s 超时、零 Token），404/405 = 路径不存在，其余 4xx 且响应为结构化 API 错误（含鉴权失败）= 路由存在，2xx 需协议成功形态，HTML / 非 JSON / 5xx 一律不算数；命中即止并回报各协议探测结果，key 只进请求头、绝不回显。"
- `specs/06-PROTOTYPE.md` · 模型配置页 · 「自定义模型与自建分组」bullet — replace "二者均仅支持添加 OpenAI 协议模型（client_type 固定 openai，不再暴露选择）" with "二者支持三种通用协议（openai-responses / ant-messages / openai-chat）：client_type 由自动检测写入或经协议下拉框手动选择，旧条目的 `openai` 值继续等效 `openai-chat`、不被静默改写".


hiyouga:
## Review

Solid, well-argued PR: the probe design (minimal invalid `{}` bodies, exactly the paths and auth headers the wrapped SDKs construct, stop-at-first-hit) is the right shape, the classification reasoning is written down where the next reader will need it, and the tests pin real behavior rather than implementation shape — the route integration tests run against real `node:http` endpoints, and the ordering tests assert that later probes never run. Conventions are all met: changelog entry plus its index line, strings in both locales, en+zh docs, English-only code, `design/` untouched with the proposed spec wording in the PR body. I verified the `resolveModelEnv` placement claim (no earlier branch substring-matches `ant-messages`) and the "key never leaks" claim end to end. One classification finding is a genuine mis-detection against a real, common gateway; the rest is hardening and UI polish.

### Findings

1. **A catch-all gateway that answers every path with a bare `{message}` error is detected as `openai-responses`** (`packages/server/src/services/protocol-detect.ts:135-143`, used at `:164`) — correctness. `isApiErrorShape` returns true for any non-empty `body.message`, so an endpoint that answers an unmatched route with `403 {"message":"Forbidden"}` — which is exactly what AWS API Gateway does — makes the **first** probe (`/responses`) classify as `served`. Detection then confidently reports `openai-responses` for an endpoint that only serves `/chat/completions`. The header doc's premise, "the probed PATH is what discriminates protocols", holds only when the endpoint answers *differently* for a path it does not serve; a uniform 4xx across all three paths carries no discriminating information at all, and the current code cannot tell the two apart because it never observes a path it knows should fail. Fix: send one control probe to a path no protocol serves (e.g. `POST {base}/__agenthub_protocol_probe__`) and, if that also classifies `served`, return `detected: undefined`. That costs one request and converts a confidently wrong answer into an honest "none matched", which the UI already handles. (Alternative, cheaper but blunter: require the strong shapes — `error` object, `type:"error"`, `object:"error"` — for 401/403, keeping the loose `message`/`detail` shapes only for 400/422.)

2. **`detail: null` counts as a structured API error** (`protocol-detect.ts:142`) — correctness, small. `if (body.detail !== undefined) return true;` accepts `{"detail": null}` and `{"detail": ""}`, so a JSON body carrying a null field and nothing else is read as proof the route exists. Require a non-empty string, object, or array instead.

3. **The probe response body is read unbounded** (`protocol-detect.ts:191`, `await res.text()`) — robustness. The 5s `AbortSignal` caps elapsed time, not bytes: an endpoint that streams fast for five seconds gets fully buffered into server memory before classification, even though only the first few hundred bytes are ever inspected. This is the only raw `fetch(` in `packages/server/src`, so no shared helper caps it. Fix: read the stream with a byte budget (64 KB is generous for an API error) and classify anything larger as `junk`.

4. **One `/detect` call can take 15s wall clock, with no overall deadline** (`protocol-detect.ts:206-212`) — noted, not a bug. Three sequential probes at `PROBE_TIMEOUT_MS = 5_000`. I checked the failure mode carefully and it is bounded and leak-free: `AbortSignal.timeout` is absolute from signal creation, and on this repo's Node 24 / undici it does abort an in-flight `res.text()`, so a server that sends headers fast and then dribbles the body cannot hold a probe past 5s. Nothing wedges — a pending probe is an idle promise plus one socket — and 15s is in line with `testModel`'s existing `requestTimeoutMs: 20_000`. The user-facing cost is that the dialog can sit on "Detecting protocol…" for 15s with no cancel. An overall budget (~12s) would be worth considering; I have not changed it.

5. **Redirects are followed** (`protocol-detect.ts:184`, `fetch` default `redirect: "follow"`) — hardening. A base URL that 302s elsewhere is followed, so the probe does not necessarily hit the URL the real client would. The blast radius is small (the reply is reduced to an enum plus a status and is never echoed), but `redirect: "manual"` costs nothing and makes the probe honest about what it measured.

6. **Stale detection notice survives a base URL edit** (`packages/web/src/features/models/models-page.tsx:1236`) — UI polish. `detectResult` is cleared only when a new run *starts*. Change the base URL to something else without triggering a run and the green "Detected …; applied" line keeps asserting a result for the previous URL. Clear it in the `baseUrl` `onChange`.

7. **Blur-to-Save race** (`models-page.tsx:1733-1739`) — noted, not fixed. Tabbing or clicking from the base URL field straight to Save fires the blur detection and submits with the pre-detection `clientType`; the in-flight result then lands on an unmounted dialog and is dropped. Detection is assistive and the user can reopen and re-run, so I do not think blocking submit on an in-flight probe is the right trade — but it is worth knowing the first save after a paste may carry the old protocol.

**Auto-detect-on-blur redundancy** (asked for explicitly): the guard is correct for the three cases that matter. Typing fires nothing (the trigger is blur, not change); re-opening the dialog on an existing working model and clicking through the field fires nothing (`bu === form.originalBaseUrl`); a second blur with no edit fires nothing (`bu === lastDetectedUrl.current`); and gateway groups never reach the code at all (`showProtocolSelector` requires a custom-like group). A failed run still records `lastDetectedUrl`, so a broken endpoint does not retry on every blur — correct, and the explicit re-run affordance covers it. One consistency note: renaming `modelId` in the dialog makes the paired reference miss the stored entry, so the probe silently falls back to keyless. That matches `/test`'s behavior and keyless probing works, so it is fine.

**SSRF verdict** (asked for explicitly): acceptable in this product's threat model, and **not a new capability**. `POST /models/test` already hands a user-supplied `baseUrl` to `GenerativeModel` and streams from it (`packages/server/src/services/project-config-service.ts:399-433`) behind the same owner-only guard (`requireProjectOwner`, `packages/server/src/http/routes/models.ts:199`). `/detect` is in fact **stricter** than the existing route — it rejects anything that is not an absolute http(s) URL (`routes/models.ts:242`), which `/test` does not — and it **leaks less**: `/test` surfaces the upstream message in `ModelTestResponse.message`, while `/detect` reduces every answer to an outcome enum plus an HTTP status. In a self-hosted product where the project owner is configuring their own inference endpoints, pointing the server at an internal host is the feature. It is worth stating that explicitly in the route doc rather than leaving it implicit, which I will do.

**API key leakage** (asked for explicitly): clean on every path I traced. The key is placed in request headers only (`bearerHeaders` / `anthropicHeaders`, `protocol-detect.ts:82-104`) and never in `probeUrl`; the response DTOs carry no key; the access log is `method path status ms` only (`packages/server/src/app.ts:375`); `app.onError` records the `Error` plus a projectId, never the request body (`app.ts:359-366`); and `detectModelProtocol` never throws — every fetch failure is caught into a probe outcome, so no undici error (which does carry the URL) ever reaches the logger or the error recorder. The stored-key fallback reads an entry from the caller's own project after `requireProjectOwner`, so no cross-tenant exposure. Only nit: `apiKey` has no `maxLen` in the route, matching `/test`'s existing convention — left alone.

**On the 5xx "proves nothing" decision**: I agree with the call. A 5xx genuinely does not separate "route exists but the handler choked on `{}`" from "route missing behind a gateway that 500s". The cost is a false negative for a naive server that 500s on a missing `model` field: its real protocol gets skipped. Because `openai-chat` is probed last and is by far the most widely served path, the realistic outcome is "none matched" — the honest answer, with the manual override right there — rather than a wrong pick. Keeping it.

- **"The protocol detection and selection can reuse the component to the right of the custom base URL — don't put it on its own new row."** The component being pointed at is the in-field grey protocol-path suffix (`models-page.tsx:1750-1752`), which already existed on `main` and already displays `/responses` / `/v1/messages` / `/chat/completions` — i.e. it is already showing the protocol, passively. I will make that suffix the control: it becomes a compact borderless trigger inside the base URL field that opens a portal menu (built on the existing `Dropdown` primitive, so it inherits viewport clamping, direction flip, Escape layering inside the modal, and arrow-key navigation) listing "Auto-detect" as the first row followed by the three protocols, each row showing the protocol name as its title and the path it appends as its description. The whole `protocol` row — label, `Select`, `Auto-detect` button and the three notice paragraphs — is deleted.
  - Detected protocol stays visible: the trigger shows the path live, and the open menu marks the current protocol by name with a checkmark.
  - Still manually overridable: picking a row is the override, exactly as the dropdown was.
  - Detection still runs on base URL change and can be re-run: the `onBlur` trigger is unchanged, and the menu's first row re-runs it on demand.
  - Failure and no-match stay communicated without a row: a single conditional line in the base URL field's own message slot (where its field error already renders), plus the trigger going amber. Nothing renders there until a run has actually happened.
  - No layout shift between states: the "detecting" state swaps the chevron for a spinner of the same box size and the warning state is colour-only, so the reserved in-field padding never changes.
  - The wrapping `<label>` has to become a `<div>` with the same `FieldLabel` span, because a `<label>` may not contain a second labelable element besides its control — a nested `<button>` would both fire and re-focus the input. The API key block directly above already uses exactly this shape for the same reason; the base URL `Input` gains an `aria-label` so it stays named.
- Also applying findings 2, 3, 5 and 6, and the explicit SSRF note on the route. Finding 1 is reported with a concrete fix but **not** implemented — a control probe changes the documented probe protocol and the spec wording proposed in this PR, so it wants an owner decision rather than a drive-by change in a review round. Findings 4 and 7 are reported only, for the reasons given above.

hiyouga:
Applied in `ded1ef4a`:

- **Maintainer feedback — the protocol control now lives in the base URL field.** The grey in-field suffix that field already had (`/responses`, `/v1/messages`, `/chat/completions`) became the control, since that path is one-to-one with the three protocol clients: it is a compact borderless trigger inside the field, and clicking it opens a menu whose first row re-runs auto-detection and whose remaining rows are the three protocols, each labelled with its name and the path it appends, checkmark on the current one. The standalone protocol row — label, `Select`, "Auto-detect" button, three notice paragraphs — is deleted. Built on the shared `Dropdown` primitive in portal mode, so it inherits viewport clamping, up/down flip, the modal Escape-layer stack and arrow-key navigation.
- Detection still fires on base URL blur under the same guards, and the picker's first row is the explicit re-run. The verdict now renders in the base URL field's own message slot and only after a run has happened, and the trigger turns amber for no-match / failure. Neither the busy state (chevron swaps for a same-size spinner) nor the warning state (colour only) changes the trigger's width, so the input's reserved padding never shifts.
- The field wrapper became a `<div>` — a `<label>` may not contain a second labelable element besides its control, and the nested `<button>` would have re-focused the input on every click. Same shape as the API key block above it; the input gained an `aria-label` so it stays named. Pure protocol helpers moved to `protocol-types.ts` so the control can share them without importing the page back.
- Finding 2 (`detail: null` counted as a structured API error) and finding 3 (probe body read unbounded — now streamed with a 64 KiB budget, abandoning and classifying as junk past the cap) are fixed, with tests. Finding 6 (stale "Detected …" line surviving a base URL edit) is fixed in the field's `onChange`.
- The route doc now states the SSRF position explicitly: owner-only, no new capability over `/models/test`, and stricter than it.

Deliberately not changed:

- **Finding 1 (catch-all gateways answering every path with a bare `{message}`, e.g. AWS API Gateway's `403 {"message":"Forbidden"}`, detect as `openai-responses`).** The fix I proposed — a control probe against a path no protocol serves, reporting "none matched" when that also looks served — changes the documented probe protocol and the spec wording this PR proposes, so it wants an owner decision rather than a drive-by change in a review round. It remains the one finding I would prioritise for a follow-up.
- **Finding 5, `redirect: "manual"`.** I withdraw this one: I had the reasoning backwards. The SDKs the real clients wrap follow redirects, so pinning `manual` would report "no protocol" for the ordinary case of a base URL that 301s http → https, breaking the invariant that a detected protocol is one that works after saving. The marginal hardening is not worth that, especially as the probe reply is an outcome enum plus a status and is never echoed. The code now carries that reasoning in a comment instead.
- **Finding 4 (15s worst case, no overall deadline)** and **finding 7 (blur-to-Save race)**: reported only, for the reasons in the review.

Verification: server 39 tests (protocol-detect + models), core 15, web 729 / 61 files, docs 32; typecheck clean for core / server / web / docs; prettier on every touched file. I also drove the real app (own port, cleaned up afterwards) at 390px and 1280px in both themes: no horizontal overflow, all four menu rows on screen at phone width, the trigger geometrically inside the input with the reserved padding clearing it, the full detection round trip on blur, and the amber no-match state resolving to amber-600 / amber-500. The only text overlaps the check reported are the dialog's own content scrolling under its sticky footer — pre-existing, unrelated to this change.

hiyouga:
Applied in `6757b2f9` (on top of merge `04343f76` with `main`).

**Maintainer feedback**

- **Detection is gated on the API key.** New pure helper `protocolDetectBlocker(hasApiKey, baseUrl)` states the preconditions in order — API key first (hard gate), then a probeable absolute http(s) URL. Key availability reuses the dialog's existing `hasKey` notion, so an **already-stored** key counts and editing an existing model never forces re-entry; ticking "clear API key" takes it away again. `runDetect` re-checks the preconditions itself, so no entry point can bypass them, and **base-URL blur no longer starts a run while a precondition is unmet**.
- **Detect moved to the field's top-right**, next to the base URL label — the same idiom the API key field already uses for its "get API key" link. Disabled state stays visible with the reason on one muted line under the field ("Fill in the API key first…" / "Fill in an absolute http(s) base URL first").
- **Keeping both would have been redundant, so I preferred your placement**: the suffix picker's "Auto-detect" row is gone and the in-field suffix is now *selection only*. There is exactly one detect affordance. The suffix still shows the live path, spins during a run, and turns amber on an inconclusive one, since the value being determined is the one it displays. Manual selection needs no API key, so a keyless endpoint (local vLLM/llama.cpp) is still fully configurable by hand.

Preserved: probe order `openai-responses → ant-messages → openai-chat` (first hit wins), manual override, detected/none/failed feedback, both locales, narrow-width + dark-mode classes (reused from the API key row and the existing muted-hint tokens).

**Merge with `main`** (#325/#327/#328): 8 conflicts, both sides kept. `openai` → `openai-chat` adopted throughout (incl. `clientTypeAfterProviderChange`, which still preserves a generic protocol type when moving to Custom); `main` had independently added generic-protocol routing to `protocol-path.ts`, so that reconciled to one implementation; the legacy `client_type` note now compares via `canonicalClientType` **and** yields to the protocol selector.

**Not changed:** the `/models/detect` endpoint still works without a key — the gate is UI-only, so the server contract and its tests are untouched. Detection on base-URL blur is kept (now gated) rather than removed; say the word if you want the button to be the only trigger.

Verified: typecheck + unit tests green for web (754), core (801), server (666), cli (245); prettier clean on touched files. No e2e run.

hiyouga:
Applied in `437d17ad` — this round revises the API-key gate from the previous one.

1. **Detect is always clickable.** `protocolDetectBlocker` is gone; `detecting` now only guards re-entrancy.
2. **Env-var fallback.** The server resolves the probe credential in three layers: the key typed in the dialog → the entry's stored key → the environment variable for the protocol **each probe** speaks. Resolution is *per probe* precisely because the protocol is the thing being determined: `ant-messages` reads `ANTHROPIC_API_KEY`, the two OpenAI protocols read `OPENAI_API_KEY`, through the same `resolveModelEnv` mapping the saved model reads. Env values are read server-side only and never reach the browser, the response, or the logs.
3. **Failures pop up a dialog.** New `AlertModal` — the existing `ConfirmModal` card with a single dismiss button — stacked over the config dialog the way the save confirmation already is. It names the case: every probe failed to connect, versus the endpoint answered but serves none of the three paths. A hit still reports as one green line under the field.
4. **Save detects first.** Confirming while the protocol is unset runs detection, then continues the save with the result; the confirm button locks and reads "Detecting…". A failed probe **aborts the save** and leaves the dialog open over the popup, so you can pick manually or fix the URL. A single-flight ref means the button and the save path share one run rather than double-firing.
5. **Default protocol is empty**, and `rowToEntry` guarantees a custom-like entry never persists an empty `client_type` — AutoLLMClient resolves an unmatched type by *throwing*, not by defaulting, so that would be a model that cannot start. The non-probing paths (set-default / set-vision-proxy / remove) rely on that net rather than probing as a side effect. Preset and vendor rows still persist empty so AgentHub keeps inferring from the model id.
6. **The "fill in the API key first" copy is deleted** from both locales, along with the group-description clause that said a key was needed.

Kept: button at the field's top-right, suffix picker as selection-only, probe order with first hit winning, manual override, both locales, narrow-width + dark-mode.

Worth flagging: adding the env fallback made an existing test read the developer's **real** `ANTHROPIC_API_KEY` into a probe. The environment is now pinned wherever detection is exercised. I also settled a semantic the feedback did not cover — "clear API key" suppresses the *stored* key but the env var still applies, since that is what the saved model would run on; that is asserted explicitly now.

Verified: typecheck + tests green for web (763) and server (672); prettier clean. Core untouched this round. No e2e.

hiyouga:
Applied in `d68818cd`.

1. **One plain error for every failure.** Unreachable, timeout, gateway junk, none-of-the-three served, unusable URL, server error — all now show the same short message: *"Could not detect the protocol. Please check the API key and the base URL."* No protocol names, no description of what the endpoint did, no "go pick one manually". `detectNone`, `detectUnreachable`, `detectNeedsUrl`, `detectFailed(msg)` and `classifyDetectFailure` are deleted, not just unused — a test asserts the distinguishing keys are absent from both catalogs. The per-protocol probe outcomes are untouched and still returned by `/models/detect` for debugging.

2. **Success pops the same `AlertModal`**, naming the protocol applied.

   **Judgement call, flagged for you to object to:** the success dialog fires only for the **manual Detect button**. On the save path a hit proceeds silently — you already committed to saving, and a dialog there would make you dismiss something to complete an action in progress. Failures still pop and still abort the save. If you'd rather see the confirmation on save too, that's a one-line change.

   **Second call, also please object if wrong:** I removed the implicit detect-on-blur. Now that feedback is a dialog, a passive trigger either throws a modal at you mid-form (tabbing out of the base URL field) or fails silently leaving an amber suffix with no explanation. Detection is now the button plus the save-time backstop — both explicit.

3. **The empty default was NOT real in the UI** — thank you for pushing on this. `protocolSelectorValue("")` returned `"openai-chat"`, so a fresh custom model rendered `/chat/completions` in the field and a checkmark on OpenAI Chat Completions in the menu: a default nobody chose, indistinguishable from a real pick. It now returns null. Untouched add-custom-model state renders **"Select protocol" / 「选择协议」** as the suffix, **no menu row checked**, and `client_type` stays empty until a pick or a detection. The field's reserved padding now measures display width, so the CJK placeholder isn't under-reserved by half.

   Left alone deliberately: moving an *existing* model into Custom still yields `openai-chat` (`clientTypeAfterProviderChange`) — that's #325's behaviour with its own test, and it converts an already-working entry rather than starting a blank one. Say the word if you want that blank too.

Verified: web typecheck + 766 tests green, prettier clean, en/zh docs and the changelog pair updated. Server untouched this round. PR still MERGEABLE.

hiyouga:
Applied in `0425e5ed`. Understood — 弹窗 meant a toast, and the result must not take up room in the form either. Both points are in this commit.

- **Both outcomes are toasts now**, via the app's existing helper — the same channel the connectivity test in this very dialog already uses, so tone and duration follow its conventions (error 6s, success 2.5s). Failure keeps the one-sentence copy; success names the protocol applied.
- **Checked the save-abort case specifically**, since a toast hidden behind the dialog was the obvious failure mode: `Toaster` portals to `document.body` at `z-[100]` while `Modal` is `z-50`, and it mounts at the app root outside the dialog tree — so it renders above the open dialog and cannot be clipped by it. On a failed save-triggered probe `onSubmit` is never reached, so the dialog stays open with the toast carrying the reason. Manual Detect toasts on both outcomes and blocks nothing.
- **`AlertModal` deleted** — models-page was its only consumer — along with the two accessible-name strings it needed, in both locales.
- **The inline verdict under the base URL field is gone.** Nothing conditional renders below the input any more, so the idle and post-detection layouts are identical: no reserved height, no shift. The state behind it collapsed from a result object to a single `detectFailed` flag whose only remaining job is tinting the suffix amber — nothing dangling. Where the protocol ended up is shown by the suffix picker, which is the thing that actually holds it; the button's "Detecting…" / disabled state stays, as that is the control's own state.

Verified: web typecheck + 768 tests green (two new ones pin that no dialog-era title strings survive and that both toast messages stay to one short line), prettier clean, en/zh docs and the changelog pair updated. Server untouched this round. PR still MERGEABLE.

hiyouga:
Applied in `0c5ccfb7`.

## 1. Custom groups infer nothing from the model id

The concrete inference was the dialog's API-key hint: with no protocol chosen it routed by model id, so typing `claude-sonnet-5` into a custom group claimed the entry reads `ANTHROPIC_API_KEY`. Custom and user-defined groups now always resolve against the compatible client. Vendor groups keep id-based routing, which is genuinely how they work.

**End-to-end behaviour for a custom model, so you can correct me if I read you wrong:**

1. The protocol selector starts **empty** — suffix reads "Select protocol", no menu row checked.
2. **Detect** is always clickable; a hit applies the protocol and toasts, a failure toasts "check the API key and the base URL" and changes nothing.
3. Pressing **Confirm** with the protocol still unset runs detection first.
4. If that detection **fails, the model now saves anyway on `openai-chat`**, with a toast saying "Protocol not detected; saved as OpenAI Chat Completions". It no longer refuses to save.
5. Nothing is ever persisted without a protocol, so the AutoLLMClient-throws concern is gone — the fallback is a real client.

That is your 默认都走兼容类型 as I read it. The one thing I chose myself: the save-path failure gets its own toast rather than the plain failure message, because "check your API key" followed by a successful save reads as a contradiction.

## 2. Vision detection

**Detect** now sits beside the vision switch. It sends one 1x1 PNG (67 bytes) with a one-word prompt, capped at 8 output tokens. Three outcomes, because "the probe failed" and "the model says no" are different facts:

- answered → vision **on**;
- an image-specific rejection → vision **off** (a real answer, not an error);
- anything else (auth, network, unrelated 400) → **setting untouched**, same one-line copy.

The rejection patterns only match text naming an image/modality concept — never a bare "invalid request" — because a false positive would switch vision off on a working model. Always clickable, single-flight, toast-only, like the protocol control. New custom models start with vision off; vendor/gateway adds keep their catalog-known capability.

**Cost, please note:** unlike the protocol probes, this one **spends real money on the user's key** — an image request cannot be shaped to cost nothing the way a deliberately invalid `{}` body can. So it runs only on an explicit press: never implicitly, and never as part of save.

**One thing I deliberately did not do:** `ModelEntry.vision` stays a plain boolean. An absent value already means *supported* throughout core, the CLI and the web app (everything tests `vision !== false`), so making it tri-state would have silently downgraded every existing un-annotated entry to non-vision — images would start folding to scratchpad paths for models that work fine today. "Starts unset" is therefore implemented as the switch starting **off** for new custom models, with detection turning it on. Tell me if you wanted the storage tri-state instead and I'll do the migration properly.

Heads-up: #326 moves the fast-mode and vision switches onto one row, so that area will conflict at merge; I kept this change scoped and did not implement that layout.

Verified: web typecheck + 771 tests, server typecheck + 681 tests (9 new for the vision verdict), prettier clean, en/zh models + server-api docs and the changelog pair updated. PR still MERGEABLE.

PR #325, opened 2026-08-19
feat(core,web,server,cli,skills,docs): refresh model catalog for AgentHub 0.4.2

### LINKED ISSUE #313: deepseek price is increased
need to align with the latest pricing: https://api-docs.deepseek.com/quick_start/pricing/
we display the lower one

hiyouga:
## What & why

Refreshes the built-in model catalog for the AgentHub 0.4.2 dependency: new Gemini 3.7 / GLM-5.3 / OpenRouter presets, the delisted Ling free tier removed, DeepSeek re-priced to the current official list (off-peak tier), and the gateway `client_type` migrated to AgentHub's canonical `openai-chat` spelling with full backward compatibility for existing user configs. The `agenthub-models` skill and the model docs are brought in line with the 0.4.2 registry and client line-up.

Fixes #313

### core

- `model-catalog.ts`
  - New direct presets: `gemini-3.7-flash` (Google group; official list price $0.15/$1.50/$7.50 per MTok — Google's launch discount halving the rates through 2026-12-31 is noted in a comment but not stored, matching the catalog's no-promotions policy) and `glm-5.3` (Z.AI group; 1M context, text-only, Z.AI's published $0.26/$1.40/$4.40 — same as GLM-5.2; served by 0.4.2's unified GLM client).
  - New OpenRouter rows (prices re-read 2026-08-18 from the model pages and the per-model endpoints API, per the block's re-read convention): `google/gemini-3.7-flash` ($0.075/$0.75/$3.75 — the halved launch-discount rates OpenRouter actually bills, comment says to re-read when the discount ends), `x-ai/grok-4.6` ($0.50/$2/$6, 500K ctx, vision), `deepseek/deepseek-v4-pro-0813` ($0.022/$0.66/$1.98, text-only per the endpoints API, 1,048,576 ctx from the default DeepSeek endpoint).
  - Removed `inclusionai/ling-3.0-flash:free` (delisted from OpenRouter); scope comment records the removal.
  - **#313**: direct DeepSeek group re-priced from https://api-docs.deepseek.com/quick_start/pricing/ storing the **off-peak tier** (the lower published price, as the issue requested): flash ¥0.05/¥1.5/¥4.5, pro ¥0.15/¥4.5/¥13.5 (cache hit/input/output; peak hours — Beijing 9:00–12:00, 14:00–18:00 — bill double, noted as a 2x-underestimate in the group comment). Stale OpenRouter DeepSeek rows re-read the same day: flash $0.0168/$0.0679/$0.168, flash-0731 $0.0157192/$0.078596/$0.157192, pro $0.022/$0.66/$1.98.
  - All gateway rows now pin `client_type: "openai-chat"`; new exported `canonicalClientType()` normalizes the deprecated bare `openai` alias; `resolveModelEnv` mirrors 0.4.2's AutoLLMClient (gpt-5.6 generation, unified GLM/Kimi branches, `ant-messages` → `ANTHROPIC_*`, `openai-responses`/`openai-chat` → `OPENAI_*`).
- `project-config.ts`: `client_type` normalized on read (`assertModelEntry`, copy-not-mutate so cached parses stay clean) and on write (`addModel`) — configs saved before the rename keep working and converge on the canonical spelling at their next save.
- `generative-model.ts`: comment updated for the rename.

### server

- `project-config-service.ts`: `canonicalClientType` applied in `getModels` (legacy configs report `openai-chat` without a disk rewrite), `updateModels` (PUTs store the canonical spelling), and `testModel`.

### web

- `models-page.tsx`: add-model default and move-to-Custom now produce `openai-chat`; the legacy-protocol read-only hint compares canonically so old `openai` entries are not flagged.
- `protocol-path.ts`: 0.4.2's generic protocol clients map to their real endpoint shapes — `openai-responses` → `/responses`, `ant-messages` → `/v1/messages` — checked ahead of the generic `openai` substring; comments re-verified against the vendored 0.4.2 clients.
- Model library sanity: no new provider groups were needed (google/zhipu/openrouter groups already exist; grouping is provider-keyed) — verified by the real-catalog sync test and the full web suite.

### cli

- `config.ts`: `model add` group default for custom/self-hosted/gateway entries is now `openai-chat` (explicit `--client-type` passes through; core normalizes the alias).

### skills

- `agenthub-models/SKILL.md` bumped to v12 (updated 2026-08-18): model-id table synced to the 0.4.2 registry (Gemini 3.7 row, GLM 5.3, GPT-5.6 variants with OpenRouter ids, MiniMax M3, Kimi K2.7 Code, `deepseek-v4-pro-0813`, `anthropic/claude-opus-5`), routing section rewritten for the 0.4.2 AutoLLMClient order and the three generic protocol clients (`openai-chat` rename + alias note, `openai-responses`, `ant-messages`), env list gains `MINIMAX_API_KEY`, `fast_mode` documented in the config list and with its per-client support matrix, and the parameter notes reflect 0.4.2 (Claude 4.6 now rejects `temperature`; GLM-5.3/Gemini-3.7 thinking clamps).

### docs

- `models.{en,zh}.md`: `client_type` examples say `openai-chat` (with the deprecated-alias note), gateway paragraph names the generic Chat Completions client, free-tier list drops Ling, preset examples refreshed (gemini-3.7-flash, glm-5.3, claude-sonnet-5, kimi-k3), and the DeepSeek off-peak pricing convention is stated.
- `configuration.{en,zh}.md`: the `client_type` field row updated the same way.

### changelog

- `changelog/unreleased/2026-08-18-model-catalog-refresh.md` + index line.

## Tests

- `@prismshadow/penguin-core`: 799 passed (40 files; includes new tests for the new rows/ordering, DeepSeek off-peak prices, `canonicalClientType`, resolveModelEnv's new client types, and a config read/write normalization round-trip).
- `@prismshadow/penguin-server`: full suite 639 passed / 67 files when run alone (includes a new legacy-on-disk `client_type = "openai"` test and PUT-normalization assertions). An earlier concurrent run showed 6 failures that were pure resource-contention flakes — the clean solo run and the unchanged base behavior confirm no real regressions.
- `@prismshadow/penguin-web`: 722 passed (60 files; protocol-path tests extended for the new protocol clients).
- `@prismshadow/penguin-cli`: 245 passed (17 files; group-default expectations updated to `openai-chat`).
- `@prismshadow/penguin-skills`: 21 passed.
- `pnpm -r typecheck`: all 8 packages clean; touched files pass `prettier --check`.
- Live-dependency sanity: constructed `AutoLLMClient` from the installed AgentHub 0.4.2 with every clientType the catalog now stores (`openai-chat`, deprecated `openai`, `minimax-m3`, `openai-responses`, `ant-messages`) and auto-routed `gemini-3.7-flash` / `glm-5.3` / `gpt-5.6-luna` — all construct; registry pricing for `gemini-3.7-flash` matches the new direct row.

## Design docs

The specs still name the old `openai` client type; proposed edits (not applied — `design/` is shared):

- `design/specs/01-PRINCIPLES.md` line 170: `client_type = "<客户端协议，如 openai>"  # 缺省由请求 id 推断；OpenAI 兼容接口填 openai` → `client_type = "<客户端协议，如 openai-chat>"  # 缺省由请求 id 推断；OpenAI 兼容接口填 openai-chat（AgentHub 0.4.2 起的规范名；旧写法 openai 为已废弃别名，读取时归一化）`
- `design/specs/01-PRINCIPLES.md` line 325: 末句 `网关分组（…）与 custom 的模型 client_type = "openai"，一律走 OpenAI 客户端、读取 OPENAI_API_KEY / OPENAI_BASE_URL。` → `网关分组（…）与 custom 的模型 client_type = "openai-chat"（AgentHub 0.4.2 起的规范名，旧写法 openai 为已废弃别名、读取时归一化），一律走通用 OpenAI Chat Completions 客户端、读取 OPENAI_API_KEY / OPENAI_BASE_URL。`
- `design/specs/06-PROTOTYPE.md` line 117: `网关模型（client_type: openai）预置 base URL` → `网关模型（client_type: openai-chat）预置 base URL`
- `design/specs/06-PROTOTYPE.md` line 124: `custom / 自建分组 / 网关固定 OpenAI chat completion（client_type=openai）` → `custom / 自建分组 / 网关固定 OpenAI chat completion（client_type=openai-chat）`
- `design/specs/06-PROTOTYPE.md` line 128: `（client_type 固定 openai，不再暴露选择）` → `（client_type 固定 openai-chat，不再暴露选择）`
- `design/specs/05-ARCHITECTURE.md` line 1047: `custom / 自建分组 / 网关为 openai` → `custom / 自建分组 / 网关为 openai-chat`


hiyouga:
## Review

Solid, well-documented catalog refresh: the `openai` → `openai-chat` migration is genuinely airtight on every path that can reach `AutoLLMClient`, the DeepSeek re-pricing matches the official page and #313's "display the lower one", and the tests pin real values rather than implementation shape. What I cannot sign off on is the pricing data itself: I re-read OpenRouter's models and per-model endpoints APIs today (2026-08-18) and **four rows are off by exactly 2x**, including one of the rows this PR adds. Details below, plus the OpenRouter/OpenAI coverage gaps the maintainer's feedback points at, and the removed-preset migration question.

### Findings

1. **`google/gemini-3.7-flash` is priced 2x what OpenRouter bills** (`packages/core/src/state/model-catalog.ts:358`) — the row stores `usd(0.075, 0.75, 3.75)` and its comment says it stores "what OpenRouter currently bills", assuming a 50% launch discount. `GET /api/v1/models/google/gemini-3.7-flash/endpoints` (read 2026-08-18) reports `discount: 0.75` on the default Google endpoint: `prompt` $0.375, `completion` $1.875, `input_cache_read` $0.0375 per MTok. Every cost estimate for this model is therefore double the real spend. Fix: `usd(0.0375, 0.375, 1.875)` — cache_write keeps the input price, since Gemini's `input_cache_write` is an hourly storage rate, per the block's own convention.

2. **The three pre-existing `openai/gpt-5.6-*` rows are stale, two of them understated** (`model-catalog.ts:448`, `:458`, `:468`). Re-read today from the same API:
   - `openai/gpt-5.6-luna` stores `usd(0.01, 0.125, 0.6)`; actual $0.02 / $0.25 / $1.20 — the "50% off" banner the row's own comment tells us to re-read on has ended. **2x understated.**
   - `openai/gpt-5.6-terra` stores `usd(0.1, 1.25, 6)`; actual $0.20 / $2.50 / $12. **2x understated.**
   - `openai/gpt-5.6-sol` stores `usd(0.5, 6.25, 30)`; actual $0.25 / $3.125 / $15, the endpoint now carrying `discount: 0.5`. **2x overstated.**

   A PR that stamps "re-read on 2026-08-18" on the OpenRouter block should not leave same-gateway rows sitting on 2026-08-07 data — especially the two that make a model look cheaper than it is.

3. **`z-ai/glm-5.3` exists on OpenRouter but only the direct GLM-5.3 row was added** (`model-catalog.ts:1040`) — inconsistent with how this same PR treats Gemini 3.7, which it adds to both the direct group and OpenRouter. OpenRouter lists `z-ai/glm-5.3`: 1,048,576 context, text-only, $0.26 / $1.40 / $4.40, `discount: 0`. That incidentally corroborates the new direct row's price exactly, which is a nice independent check on it.

4. **The direct OpenAI group stops at GPT-5.5** (`model-catalog.ts:986`) — AgentHub 0.4.2 ships a native `gpt-5.6` client and registry rows for `gpt-5.6` / `gpt-5.6-terra` / `gpt-5.6-luna` against `api.openai.com`, and this PR already teaches `resolveModelEnv` about `gpt-5.6` (`:1153`) and documents the generation in the skill table — but a user with their own OpenAI key still cannot pick GPT-5.6 from the presets. A requirement gap for a "refresh for AgentHub 0.4.2" PR.

5. **OpenRouter parity gap for the direct OpenAI models** (the maintainer's item, see below) — the direct group carries six ids and only `openai/gpt-5.5` has an OpenRouter counterpart. `openai/gpt-5.5-pro`, `openai/gpt-5.4`, `openai/gpt-5.4-mini`, `openai/gpt-5.4-nano` and `openai/gpt-5.4-pro` all exist on OpenRouter today and are absent from the catalog.

6. **Removing the `ling-3.0-flash:free` preset degrades an existing user's row silently, and no test pins the behavior.** I traced it end to end: nothing crashes, nothing is pruned, no credential is lost. `presetModelEntries` had persisted everything except `displayName` into the user's TOML, so `context_window`, `pricing`, `vision`, `client_type`, `base_url` and `api_key` all survive; `ensurePresetModels` (`project-config-service.ts:255`) early-returns whenever the table is non-empty and never diffs against the catalog; `syncRowsWithCatalog` (`catalog-sync.ts:53`) iterates over presets rather than rows, so an ex-preset row is left untouched and in place; and the runtime never consults the catalog at all (`agent.ts` does not import `model-catalog`), so the model stays selectable and remains valid as the default. What *does* change, with no signal to the user:
   - `displayName` was the one field living only in the catalog, so the card silently switches from "Ling 3.0 Flash (free)" to the raw `inclusionai/ling-3.0-flash:free` (`project-config-service.ts:517`, `models-page.tsx:992`).
   - `isPreset` flips to false (`models-page.tsx:270`), so the edit dialog switches to custom-model semantics: `baseUrlRequired` becomes true, the vision toggle becomes user-editable instead of catalog-locked, and a cleared context window would auto-fill to `CUSTOM_CONTEXT_DEFAULT` (128000) rather than meaning "unknown".
   - `modelHomepageUrl` is pattern-based for OpenRouter, so "Model page" still links to a now-delisted page.

   That is acceptable graceful degradation and I would not redesign it here — but this PR is the one that removes a preset, and **no test anywhere constructs "row exists in config, gone from `MODEL_CATALOG`"**. The nearest tests (`server/test/models.test.ts:148`, `web/test/catalog-sync.test.ts:138`) cover *never-in-catalog* and *user-added* rows, which is a different intent. I will add a regression test pinning that such a row still loads, keeps its pricing/context/credential, and is still reported by GET.

7. **Stale `openai` spelling left in two doc comments and both CLI locales** — `packages/cli/src/i18n.ts:331` (`"AgentHub client type (e.g. openai)"`) and `:590` (`（如 openai）`) still advertise the deprecated alias while the CLI's own `model add` default is now `openai-chat`; `packages/core/src/interfaces.ts:115` and `packages/server/src/api/types.ts:292` were missed while the equivalent comment in `generative-model.ts` was updated. Cheap, and the i18n one is user-visible in both locales.

### Verified clean (stating explicitly, since these were the risky parts)

- **The `openai` → `openai-chat` normalization really does cover every read/write boundary.** There are exactly two `GenerativeModel` construction sites outside tests: `packages/core/src/agent.ts` (fed only from `loadProjectConfig` → `projectConfigFromTable` → `assertModelEntry`, which normalizes at `project-config.ts:222` and returns a copy rather than mutating a possibly-cached parse) and `project-config-service.ts:418` (`testModel`), which *does* bypass `assertModelEntry` via the raw mtime-cached TOML read but normalizes explicitly at `:408`. Server GET normalizes at `:506`, PUT at `:630`. So an old TOML cannot reach `AutoLLMClient` with a stale string — and even if it did, AgentHub still routes bare `openai` to `OpenaiChatClient` as an alias, so this is belt-and-braces rather than a load-bearing migration.
  - Two residuals, both low severity and worth knowing rather than fixing here: (a) the server's non-model writers (`setName`, `setDefaultModelRef`, `setChatDefaults`, `ensurePresetModels`) spread the raw cached table, so a stale `client_type = "openai"` survives verbatim *on disk* until a models PUT or a core mutator round-trips it — "migrates on next save" holds only for model-scoped saves; (b) `assertModelEntry` normalizes only when `typeof client_type === "string"`, so a non-string value in hand-edited TOML still falls through untyped (pre-existing).

- **Every gateway row pins an explicit `client_type`** — I checked all five gateway groups programmatically: 0 rows rely on auto-routing. That matters more than it looks. `AutoLLMClient._createClientForModel` (agenthub `src_ts/src/autoClient.ts:74`) resolves `clientType = (client_type || CLIENT_TYPE || model).toLowerCase()` and then runs *raw substring tests in a fixed order*; `base_url` never influences routing and the registry is never consulted. An unpinned `openai/gpt-5.6-sol` would match the `gpt-5.6` branch and get the native OpenAI client aimed at a gateway; an unpinned `anthropic/claude-opus-4.8` would **throw**, because dotted "4.8" matches neither `4-8` nor `-5`. The vendor prefix in a gateway id gives no protection. Keep the invariant: gateway rows always pin, and always with hyphens — `openai_responses` with an underscore is the one misspelling that fails *silently* into `OpenaiChatClient` instead of throwing.

- **DeepSeek re-pricing (#313) is correct.** Verified against api-docs.deepseek.com today: off-peak `deepseek-v4-flash` ¥0.05 / ¥1.5 / ¥4.5 and `deepseek-v4-pro` ¥0.15 / ¥4.5 / ¥13.5 (the page renders these as $0.007 / $0.22 / $0.66 and $0.022 / $0.66 / $1.98), peak exactly 2x during UTC 01:00–04:00 and 06:00–10:00 = Beijing 09:00–12:00 and 14:00–18:00, precisely as the new group comment says. The OpenRouter `deepseek/*` rows agree to the cent, and their API records carry the same `utc_start`/`utc_end` doubling windows. Storing the lower tier is what #313 asked for.

### On the DeepSeek off-peak price and what the user actually sees

Worth being straight about this, since it is a "the number is half the truth half the day" situation. **The peak-hour doubling is invisible in the product.** The user sees three bare numbers — the card meta line joins `cacheRead / cacheWrite / output` with slashes and no labels (`models-page.tsx:938`), the dialog shows three inputs with a currency adornment and a `/M tok` suffix (`:1690`), and the CLI prints one `price=a/b/c` column (`cli/src/commands/config.ts:86`). Cost-centre and per-turn costs are computed from that single rate. The doubling is stated in the new group comment and in `models.{en,zh}.md:79` — in source and on the docs page, never next to the price.

I checked whether a caveat could be attached cheaply: it cannot. There is no note/tooltip/footnote field anywhere on the path — not on `ModelCatalogEntry`, not on `ModelPricing` (which is what round-trips through TOML and what `getModels` actually reads), and not on the server's `ModelInfo` DTO. A per-model caveat would need a new field in all three plus new bilingual strings, which is a feature, not a fix, and outside this PR. The one existing caveat mechanism is the cost centre's global `*` footnote for *unpriced* models, which has no hook for rate accuracy.

So, honestly: the PR is consistent with the precedent already in the file — OpenAI's >272K surcharge, Gemini 3.1 Pro's >200K, MiniMax's >512K doubling and the Gemini launch discount are all equally invisible in the UI and equally documented only in comments and the docs page. #313 explicitly asked for the lower number, so the choice is right; it is the *disclosure* that is thin, and it is thin for every model, not just DeepSeek. If you want that fixed it deserves its own issue: a `pricingNote` on the catalog entry surfaced as a `?` next to the price triple would cover DeepSeek's peak hours, the long-context tiers and the promotional discounts in one go. I have not built it here.

> "OpenAI's new models should be listed on OpenRouter too. OpenRouter can switch to the responses interface."

- **(A) OpenRouter coverage for OpenAI models.** Worth flagging that the premise is already half-satisfied: `openai/gpt-5.6-sol` / `-terra` / `-luna` and `openai/gpt-5.5` have been on OpenRouter since before this PR. The gap that actually exists runs the other way, and in the 5.4 family, so I will close both directions: add the five missing OpenRouter rows (`openai/gpt-5.5-pro`, `openai/gpt-5.4`, `openai/gpt-5.4-mini`, `openai/gpt-5.4-nano`, `openai/gpt-5.4-pro` — every id verified present in OpenRouter's live models API, prices and context windows taken from the per-model endpoints API today), and add the missing direct `gpt-5.6` / `gpt-5.6-terra` / `gpt-5.6-luna` rows so "OpenAI's new models" are listed on *both* sides. After that the direct group and the OpenRouter group carry the same OpenAI line-up. If you would rather keep the direct additions out of this PR they are a self-contained three-row block and easy to drop.

- **(B) Responses protocol for the OpenRouter OpenAI rows.** Confirmed end to end before changing anything: agenthub's `OpenaiResponsesClient` builds `new OpenAI({ baseURL })` + `responses.create()` (`src_ts/src/openai_responses/client.ts:59,431`), i.e. `POST {base}/responses`; with the catalog's existing `https://openrouter.ai/api/v1` that is `https://openrouter.ai/api/v1/responses`, which answers `401` unauthenticated while a bogus path under the same prefix returns a `404` HTML page — the endpoint is real, not a guess. Agenthub's own e2e maps `openrouter → openai-responses → https://openrouter.ai/api/v1` (`src_ts/tests/client.test.ts:147`) and runs `openai/gpt-5.6-luna` through all three protocol clients, and the 0.4.2 changelog lists OpenRouter among the servers `openai_responses` covers.

  Per-row decision, and why it is not "switch the whole gateway":
  - **`openai/*` rows → `client_type: "openai-responses"`.** These are the rows whose upstream genuinely *is* an OpenAI Responses server, so the client's heavy reasoning-item round-tripping (replaying `encrypted_content` / `signature` / `summary`, the `phase` field, `store: false`) has real semantics at the other end. This is also the exact combination agenthub verified live.
  - **Everything else on OpenRouter stays `openai-chat`** — `anthropic/*`, `google/*`, `deepseek/*`, `x-ai/*`, `z-ai/*`, `moonshotai/*`, `minimax/*`, `qwen/*`, the `:free` rows and `openrouter/free`. OpenRouter will accept `/responses` for them by translating, but nothing has verified that the reasoning-replay semantics survive that translation for a non-OpenAI upstream, and a wrong protocol pin breaks the model for every user who has it configured. Not worth guessing on. (`minimax/minimax-m3` is the one arguable case, since MiniMax serves a Responses API directly — but *via OpenRouter* it is unverified, so it stays put.)
  - The other four gateways (Fireworks / SiliconFlow / Qwen Token Plan / Qwen Pay-As-You-Go) carry no OpenAI-family ids at all, so they are untouched.

  I will also update the group comment in `model-catalog.ts` and the `MODEL_PROVIDERS` gateway note, which currently claim the gateway groups uniformly pin `client_type: "openai"` / `openai-chat` — that stops being true with this change — plus the corresponding paragraph in `models.{en,zh}.md`.

### Skipped

Nothing material. `packages/core/test/agent.test.ts` and `packages/cli/test/config-format.test.ts` still use the literal `client_type: "openai"` in fixtures; that is correct as-is — they exercise the alias path and the CLI's display of a hand-built config, and rewriting them would reduce coverage of exactly the backward-compatibility case this PR adds. The `pricingNote` idea above is deliberately not built here.

hiyouga:
Applied in `091292d4`:

**Maintainer feedback**

- **OpenAI line-up now mirrored on both sides.** Added the direct `gpt-5.6` / `gpt-5.6-terra` / `gpt-5.6-luna` rows (AgentHub 0.4.2's native client, OpenAI list prices) and the five OpenRouter counterparts that were missing: `openai/gpt-5.5-pro`, `openai/gpt-5.4`, `openai/gpt-5.4-mini`, `openai/gpt-5.4-nano`, `openai/gpt-5.4-pro`. Every direct OpenAI model now has an `openai/<id>` gateway row and vice versa, pinned by a test so the two groups cannot drift apart.
- **The nine OpenRouter `openai/*` rows pin `client_type: "openai-responses"`.** Non-OpenAI OpenRouter rows and the other four gateways stay on `openai-chat` — reasoning-item round-tripping is only guaranteed when the upstream is genuinely OpenAI, and a wrong pin would break the model for everyone who has it configured. Both protocols read the same `OPENAI_*` pair, so credentials and the env-fallback hint are unchanged.
- Updated the comments that claimed gateway groups uniformly pin `openai-chat` (file header, `MODEL_PROVIDERS`, the OpenRouter block, `presetModelEntries`) plus the gateway paragraph in `models.{en,zh}.md`, and recorded *why* every gateway row must pin explicitly.

**Review findings**

- Re-read prices on 2026-08-18: `google/gemini-3.7-flash` → `0.0375 / 0.375 / 1.875` (the endpoint runs `discount: 0.75`, not 0.5), `openai/gpt-5.6-luna` and `-terra` doubled to full rate after their promotion lapsed, `openai/gpt-5.6-sol` halved now that it runs one. Rows on a promotion now name the rate to restore.
- Added the missing `z-ai/glm-5.3` OpenRouter row.
- Added the catalog-removal regression test (a configured row dropped from the catalog keeps its pricing / context / vision / base URL / API key and stays valid as the default; only the display name is lost).
- Refreshed the stale bare `openai` in the CLI `--client-type` help (both locales) and two doc comments.

**Deliberately not changed**

- **Per-model pricing caveats** (DeepSeek's peak-hour doubling, the long-context tiers, promotional rates) are still invisible in the UI. There is nowhere for such a note to live — `ModelCatalogEntry`, `ModelPricing` and `ModelInfo` all lack the field — so it is a feature, not a fix. Suggested as its own issue in the review.
- **`minimax/minimax-m3` on OpenRouter** stays `openai-chat`: MiniMax serves a Responses API directly, but the gateway path for it is unverified.
- The `client_type: "openai"` fixtures in `core/test/agent.test.ts` and `cli/test/config-format.test.ts` stay as they are — they cover the deprecated-alias path this PR adds.

Tests: core 800 passed / 5 skipped, server 640, web 722, cli 245, skills 21; `pnpm -r typecheck` clean across all 8 packages; touched files prettier-clean.

**Design docs** — one addition to the proposals already in the PR description, since `design/` is read-only here. `design/specs/01-PRINCIPLES.md` line 325 and `design/specs/06-PROTOTYPE.md` lines 117/124 say gateway models uniformly use the OpenAI client; they now need the exception, e.g. after the `openai-chat` wording: `；其中 OpenRouter 的 openai/* 预置固定 client_type = "openai-responses"（OpenRouter 在同一 base URL 上提供 Responses API），仍读取 OPENAI_API_KEY / OPENAI_BASE_URL。`

PR #326, opened 2026-08-19
feat(models): per-model fast mode (default off)

hiyouga:
## What & why

AgentHub 0.4.2 added `fast_mode` to its `UniConfig`: some models offer a faster-output serving tier at premium pricing (OpenAI-protocol clients map it to `service_tier: "priority"`; Anthropic-protocol clients to `speed: "fast"` plus the `fast-mode-2026-02-01` beta header). This PR exposes it as an optional per-model setting across core, server, web, and CLI.

Contract: **default off, only `true` is ever persisted** (`fast_mode = true` on the model entry; absent = off). Existing configs are byte-identical in behavior — `buildUniConfig` sets the key only when enabled, so models that would reject the parameter never see it unless the user opts in.

AgentHub's registry exposes **no** per-model fast-tier capability flag (verified against `src_ts/src/registry.ts` in 0.4.2), so the web toggle is offered on every model with helper text, and runtime safety handles the unsupported case (below).

### core

- `ModelEntry.fast_mode?: boolean` (`packages/core/src/state/project-config.ts`), with `addModel` upsert merge (`--no-fast-mode` clears the annotation instead of writing `false`; a hand-edited `false` normalizes to absent).
- `GenerativeModelConfig.fastMode` → `buildUniConfig` sets `UniConfig.fast_mode` only when enabled (`packages/core/src/llm/generative-model.ts`).
- Wiring (`packages/core/src/agent.ts`): rides the session LLM config, and therefore post-compaction rebuilds. The bare/meta LLM (title generation) and the `describe_image` vision describer deliberately do **not** inherit it — their background requests gain nothing user-facing from a premium tier, and skipping keeps them working (and standard-priced) even while the annotation is enabled on a model that rejects it.
- **Runtime safety** (researched against AgentHub 0.4.2 source): clients without a fast tier (`gemini3_7`, `glm5_3`, `kimi_k3`, `deepseek_v4`, `openai_embedding`, and `claude5` on Bedrock or Claude 4.6 ids) throw `UnsupportedParameterError` **before any network I/O** — a deterministic client-side rejection. Under the engine's retry-everything-but-auth policy it would burn the full ~60 s reconnect ladder (5 retries) before surfacing; the policy's stated rationale ("the classifier can't tell a permanent provider 4xx from a transient one") does not apply to AgentHub's own typed error. So:
  - `isFastModeUnsupportedError` detects it (instanceof + name fallback, cause-chain walk, scoped to `parameter === "fast_mode"`, and only when this model object's own config enabled fast mode);
  - the outcome is `failed` with a new optional `LLMOutcome.permanent: true` hint and an actionable message (AgentHub's text + "Fast mode is enabled for this model; turn it off in the model settings to use this model.");
  - the engine (`packages/core/src/engine/context-engine.ts`) aborts immediately on a permanent failed — same terminal shape as `auth`, but hosts do not gate input (the fix is a settings change, not a credential) — and `plannedRetryDelayMs` announces no `retry_in_ms` for it, so the frontend never renders a countdown for a retry that never comes.
  - Every other failure (other `UnsupportedParameterError` parameters included) keeps today's classification and ladder unchanged.
  - Known bounded edge: a compaction request could in principle hit the rejection first only via manual `/compact` as the very first request of a resumed session with fast mode newly enabled; it then retries the compaction ladder (3 attempts) and fails safely with the context kept. Not special-cased — unreachable in the normal turn flow, where the first regular request aborts permanently before any compaction.

### server

- `ModelInfo.fastMode` / `ModelUpdateEntry.fastMode` DTOs (`packages/server/src/api/types.ts`); hand-rolled validation mirrors `vision` (`models[i].fastMode must be a boolean`, `packages/server/src/http/routes/models.ts`).
- `project-config-service.ts`: GET reports `fastMode: true` only when annotated; PUT persists only `true` and clears on omission (added to the full-table-replace delete block, so the annotation is actually clearable).
- `POST /models/test` accepts a `fastMode` override, falling back to the stored annotation — the connectivity probe exercises exactly the serving tier sessions would use, so a fast-mode rejection (or Anthropic's 429 for un-provisioned orgs) surfaces at test time, before saving.

### web

- "Fast mode" toggle in the model config dialog (`packages/web/src/features/models/models-page.tsx`): same inline-switch shape as the vision toggle but shown for **every** model (preset ones included — no catalog capability flag exists), default off, `title` hover hint plus a muted hint line while ON: faster output at premium pricing; models without a fast tier reject requests.
- The connectivity test always sends the form's current toggle (like base URL), so an unsaved toggle-off overrides a stored `true`.
- Catalog sync (`catalog-sync.ts`) treats fast mode as user-owned like `max_tokens` — never clobbered by "sync presets".
- Strings in both locales (`strings.ts` zh / `strings-en.ts` en).

### cli

- `penguin config model add` gains tri-state `--fast-mode` / `--no-fast-mode` (neither given keeps the current value), mirroring `--vision` / `--no-vision`; i18n en + zh (`packages/cli/src/commands/config.ts`, `packages/cli/src/i18n.ts`).

### docs & changelog

- `models.en/zh.md`: `fast_mode` row in the ModelEntry table + a new "Fast mode" section (semantics, premium pricing not reflected in the recorded per-token prices, per-client rejection behavior, background requests excluded, test-before-save).
- `configuration.en/zh.md` (field table row), `cli.en/zh.md` (flag row), `web-app.en/zh.md` (models page), `interfaces.en/zh.md` (`GenerativeModelConfig.fastMode`, `LLMOutcome.permanent`).
- Changelog: `changelog/unreleased/2026-08-18-model-fast-mode.md` + index line.

## Tests

New coverage:

- core `test/llm.test.ts`: `buildUniConfig` fast_mode on/off/absent; `isFastModeUnsupportedError` (typed, cause-chain, name-fallback, scope negatives); `streamGenerate` classification — permanent failed with both message parts, and the config guard (same error without fast_mode enabled stays a plain failed).
- core `test/engine.test.ts`: a permanent failed aborts after exactly 1 attempt with the actionable reason and no announced `retry_in_ms`; a plain failed still takes the ladder.
- server `test/models-fast-mode.test.ts` (new file, mirrors `models-vision.test.ts`): true persists and reads back (PUT response + fresh GET), false/omitted never persisted, omission clears, preset models accept the annotation, non-boolean 400.
- web `test/models-input.test.ts` (`toRow` carry-through), `test/catalog-sync.test.ts` (RowState fixture).
- cli `test/config-model.test.ts`: `--fast-mode` round-trips to `fast_mode = true` in the TOML; upsert without flags keeps it; `--no-fast-mode` removes the key entirely.

Results (all run in this worktree):

- `@prismshadow/penguin-core` tests: **802 passed, 5 skipped** (skips are env-gated e2e, pre-existing)
- `@prismshadow/penguin-server` tests: **641 passed**
- `@prismshadow/penguin-web` tests: **722 passed**
- `@prismshadow/penguin-cli` tests: **246 passed**
- `typecheck`: clean for core, server, web, cli (core initially showed `@prismshadow/penguin-skills` module-resolution errors on the unbuilt base checkout; after `pnpm -r build` the error set is empty, and a before/after diff confirmed this change introduces none).
- `prettier --write` run on every touched file.

## Design docs

The design specs don't mention fast mode yet; per the worktree brief, proposed edits (not applied — `design/` is shared):

1. `design/specs/06-PROTOTYPE.md` § 模型配置页, 配置弹窗 bullet — after 「三项价格（标签自含，无独立价格标题行）与视觉开关」 insert:

   > 与**快速模式开关**（所有模型可配置，含预置模型；默认关闭，只持久化 `true`。开启即让该模型的会话请求进入厂商的溢价快速档位（AgentHub `fast_mode`）；内置目录无该能力标志，故开关对全部模型展示，开启态下方一行小字说明「输出更快、按溢价计费；仅部分模型支持，不支持的模型会拒绝请求」。后台请求（标题生成、describe_image 代读）不携带）

   and in the 连通性测试 bullet, extend 「可带尚未保存的 api_key / base_url / client_type」 to 「可带尚未保存的 api_key / base_url / client_type / fast_mode（前端恒发送弹窗当前开关值，测试即测所配档位，被拒在保存前即可发现）」.

2. `design/specs/05-ARCHITECTURE.md` L421 — amend the parenthetical 「不设配额启发式特判，确定性错误烧完阶梯与直接失败同终态，是接受的取舍」 with the carve-out:

   > 唯一例外是 AgentHub 在发起网络请求前抛出的 fast_mode `UnsupportedParameterError`（在没有 fast 档位的模型上启用快速模式）：这是 SDK 自身的类型化确定性拒绝、非厂商文案，`LLMOutcome` 以 `permanent: true` 标记，引擎立即带可行动的错误消息终止本轮（同 auth 的终止路径，但不触发输入置灰），且其 `request_end` 不携带 `retry_in_ms`。

## Notes for reviewers

- `LLMOutcome.permanent` is engine-internal (not on the wire): `request_end` still carries `status: "failed"` + `error_message`; the run then ends with the abort event, so observability records an honest failure with the real reason.
- Deliberately **not** done, to stay scoped: no fast-mode badge on model cards, no `config model list` column (matching `max_tokens`, which the list also omits), no per-model premium price list (the recorded pricing stays the standard tier — documented in the models page), and no generalization of `permanent` to other `UnsupportedParameterError` parameters (their behavior is unchanged; a follow-up could widen it).
- Merge-conflict note: sibling PRs touch the model catalog and the custom-model add flow; this PR's edits to the shared files (`models-page.tsx`, `strings*.ts`, `config.ts`, `i18n.ts`, docs tables) are small, additive blocks keyed next to the vision field, so conflicts should be trivial.


hiyouga:
快速模式要弹出提示警告用户，而且不支持的模型就不要加这个按钮，以免报错。

hiyouga:
## Review

The plumbing is solid and the repo conventions are all met: `fast_mode` is annotation-only (never `false` on disk), the wire key is set only when enabled so existing configs are bit-identical, both locales ship, en+zh docs and the changelog entry are updated, and the new tests genuinely pin behavior rather than shape (the engine test asserts *one* attempt and *no* announced `retry_in_ms`, not just a flag). I verified the three claims that carry the design: catalog sync really does leave `fastMode` alone (it sits outside `presetFields`, `catalog-sync.ts:16-26`), the dialog's no-op guard really is field-exact (`JSON.stringify(form)` vs a rebuilt baseline, `models-page.tsx:1427-1439`), and the three `GenerativeModel` construction sites in `agent.ts` are wired as described (session at `:842/:848`, describer at `:744`, bare/meta at `:857`).

The central design decision, though, is wrong, and it is the one the maintainer flagged. The PR reasons "AgentHub's registry exposes no per-model fast-tier capability flag, so the toggle is offered on every model". The registry claim is true, but the capability is still fully determined — by the *client* AgentHub routes to, which is a pure function of `(client_type || model_id)` plus the base URL. Offering the toggle everywhere means the UI happily arms a switch that is guaranteed to kill the user's next turn on 15 of the 80 built-in catalog rows. Findings below, then the maintainer items.

### Findings

1. **The fast-mode toggle is offered on models AgentHub will certainly reject** (`packages/web/src/features/models/models-page.tsx:1776-1801`) — the maintainer's point, and a correctness bug in its own right. `AutoLLMClient` (0.4.2, `dist/autoClient.js:55-113`) selects a client from `t = (clientType || CLIENT_TYPE || model).toLowerCase()`, and each client's fast-mode behavior is fixed:

   | branch on `t` | client | `fast_mode` |
   |---|---|---|
   | `=== "minimax-m3"` | MiniMaxM3 | `service_tier: "priority"` |
   | `includes("gemini-3" \| "gemini-embedding")` | Gemini3_7 | **throws** |
   | `includes("claude") && includes("4-6"\|"4-7"\|"4-8"\|"-5")` | Claude5 | `speed: "fast"` + beta — **throws** on `bedrock://` or a model id containing `4-6` |
   | `includes("gpt-5.4"\|"gpt-5.5"\|"gpt-5.6")` | GPT5_6 | `service_tier: "priority"` |
   | `includes("glm-5")` | GLM5_3 | **throws** |
   | `includes("kimi-k3"\|"kimi-k2.5"\|"kimi-k2.6")` | KimiK3 | **throws** |
   | `includes("deepseek-v4")` | DeepSeekV4 | **throws** |
   | `includes("ant-messages")` | AntMessages | `speed: "fast"` + beta |
   | `includes("openai-responses")` | OpenaiResponses | `service_tier: "priority"` |
   | `includes("openai") && includes("embedding")` | OpenaiEmbedding | **throws** |
   | `includes("openai")` | OpenaiChat | `service_tier: "priority"` |
   | else | — | unroutable (AgentHub refuses the model outright) |

   Concretely, with today's catalog the toggle is armed-and-fatal on `deepseek-v4-flash`, `deepseek-v4-pro`, all six `gemini-3.*`, `glm-5.2`/`glm-5.1`/`glm-5`, `kimi-k3`/`kimi-k2.6`/`kimi-k2.5`, and `claude-sonnet-4-6` — 15 rows. Note the last one especially: it routes to Claude5, whose family the client *does* serve, but the client refuses the id by name (`claude5/client.js:203-220`, substring `"4-6"` against the raw model id).

   Two subtleties the rule has to respect, because the catalog's *declared* `client_type` is not always what runs. Resolution is `(client_type || CLIENT_TYPE || model_id).toLowerCase()`, so an entry with no explicit client type self-routes on its **model id**, and that can land somewhere other than the group would suggest: `anthropic/claude-fable-5` self-routes to the native Claude5 client (flipping the wire mapping from `service_tier: "priority"` to `speed: "fast"`), `openai/gpt-5.6-sol` self-routes to GPT5_6 rather than the `openai` catch-all (the `gpt-5.*` branch precedes it), and the dotted `anthropic/claude-opus-4.8` / `4.7` match *neither* `"4-8"` nor `"-5"` and so match nothing at all — AgentHub throws rather than falling back to openai_chat. Today's catalog papers over all three by pinning `client_type: "openai"` on every gateway row, but the helper must resolve the effective client the same way the router does instead of trusting the declared type, or a user-added id in a gateway group gets the wrong verdict.

   The fix must be a **rule**, not a list, because sibling PR #325 adds Gemini 3.7 / GLM-5.3 / OpenRouter rows on another branch; a hand-maintained per-row table would silently give them the wrong answer. Mirroring the branch table above gets them right for free (`gemini-3.7-*` → hidden via `includes("gemini-3")`, `glm-5.3` → hidden via `includes("glm-5")`, OpenRouter rows → shown, since they pin `client_type: "openai"`). There is an exact precedent to model it on: `resolveModelEnv` in `model-catalog.ts:1064-1089` already mirrors this same branch order for env-var fallback and documents itself that way.

2. **Cost is silently under-reported while fast mode is on** (`packages/server/src/services/usage-service.ts:81-89`, `packages/web/src/lib/omni/task-stats.ts:58-69`, `packages/web/src/features/traces/trace-file-view.tsx:236-244`). Cost is never persisted — `usage_records` stores only token buckets keyed by `(provider, model_id)` (`packages/server/src/db/schema.ts:63-78`: *"cost is not stored: computed at query time from current pricing"*), and every cost surface multiplies those buckets by the entry's single price triple. Nothing anywhere carries a serving tier. So with fast mode enabled, the usage cards, the daily-cost trend, the chat header's live cost chip, and the trace viewer all under-report — by whatever the provider's premium markup is (MiniMax documents 1.5x; OpenAI and Anthropic publish separate premium price lists). The PR does document this in `models.en.md:99`, which is honest, but a docs line is not where a user meets the surprise.

   **I am not fixing this in this PR, and I think that is the right call** — not because it is hard to hack in, but because every cheap version is wrong. The one natural chokepoint, `getPricing` (`packages/server/src/services/project-config-service.ts:359-375`), already has `entry.fast_mode` in scope, so a multiplier there is a few lines — but (a) pricing is applied retroactively at query time, so flipping the toggle today would silently re-price *last month's* history for that model, which `packages/server/test/usage.test.ts:166` pins as deliberate behavior; (b) the meta/title LLM and the `describe_image` describer deliberately *skip* fast mode yet record under the same `(provider, model_id)`, so a model-level multiplier would over-charge exactly the requests the PR was careful to keep cheap; and (c) there is no single correct multiplier anyway — 1.5x is MiniMax's number, not OpenAI's or Anthropic's. Doing it properly means a `fast_mode` column on `usage_records` plus splitting the aggregates by it, which is a schema change and its own PR. What this PR *can* do cheaply is stop the surprise being silent, so I am putting the under-reporting into the new confirmation copy and leaving the accounting alone.

3. **The wiring claim that scopes the whole feature has no test** (`packages/core/src/agent.ts:828-836`). The PR's central scoping decision — fast mode rides the session LLM (and post-compaction rebuilds) while `createBareLLM` (title generation) and the `describe_image` describer deliberately do not inherit it — is asserted in three comments, a changelog paragraph, and the docs, and pinned by nothing. A future refactor that hoists `fastMode` into a shared config helper would start billing every session title at premium rates with all suites still green. `packages/core/test/agent.test.ts:651-664` already reaches in for `createBareLLM` to test the max-tokens split, so the seam is there and the test is cheap. Adding it.

4. **Polish: nothing in the model list shows that fast mode is on** (`packages/web/src/features/models/models-page.tsx:999-1008`). The row already badges Free / Vision / Proxy-vision; a switch that moves the model onto a premium price list is at least as worth seeing without opening the dialog — and after finding 2 it is the only place the user is reminded that their cost readout is now low. Adding a badge next to the existing three.

5. **Pre-existing, out of scope, reporting only:** `resolveModelEnv` (`packages/core/src/state/model-catalog.ts:1076`) matches `gpt-5.4` and `gpt-5.5` but not `gpt-5.6`, while AgentHub 0.4.2 routes all three to `GPT5_6Client`. A user adding a `gpt-5.6` id under the `openai` group therefore gets no `OPENAI_API_KEY` fallback hint and a spurious "cannot auto-route" note. It predates this branch and belongs to the env-hint feature, so I am not touching it here — but my new helper does handle `gpt-5.6`, so the two mirrors in that file will disagree on that one id until it is fixed. Flagging it so it does not read as a typo introduced by this PR.

No other correctness issues found. Specifically checked and clean: the tri-state CLI flags really do default to `undefined` (Commander only flips the default when `--no-x` is declared alone; the test at `config-model.test.ts:306` pins the keep-current case), the `permanent` branch sits in the right place in the turn loop (after `aborted`/`auth`, before `completed`), `plannedRetryDelayMs` refuses to announce a countdown for it, `isFastModeUnsupportedError` is correctly narrowed to `parameter === "fast_mode"` *and* guarded on the model's own `uniConfig.fast_mode` so a foreign error cannot be mislabeled, and the full-table-replace clear path (`delete next.fast_mode`) is present so the annotation is actually clearable.

> 快速模式要弹出提示警告用户，而且不支持的模型就不要加这个按钮，以免报错。

- **(A) Do not offer the toggle where the model cannot serve fast mode.** New helper `fastModeProtocol(modelId, clientType?, baseUrl?)` in `packages/core/src/state/model-catalog.ts`, next to `resolveModelEnv` and mirroring `AutoLLMClient`'s branch order exactly, returning `"openai"` (→ `service_tier: "priority"`), `"anthropic"` (→ `speed: "fast"` + `fast-mode-2026-02-01`), or `undefined` for everything else. Rule, stated plainly: **the toggle appears only for models whose AgentHub client actually puts the parameter on the wire** — which is protocol-determined, with two carve-outs on the Anthropic client (a `bedrock://` base URL, and any model id containing `4-6`). Table-driven tests, including a sweep asserting a verdict for every one of the 80 catalog rows so a new row can never quietly acquire the wrong answer. Deliberate decisions inside the rule:
  - *Custom models behind an arbitrary base URL keep the toggle.* They route through `openai_chat` (the web pins `client_type: "openai"` for custom and user-defined groups), and the endpoint may well be OpenAI itself. Hiding it there would be wrong more often than showing it; the confirmation carries the caveat that some OpenAI-compatible servers accept the parameter and quietly serve standard tier.
  - *A row whose stored value is already `true` keeps its toggle even when the rule says unsupported.* Otherwise a config that got there another way (hand-edited TOML, `--fast-mode`, a model id renamed after the fact) would produce a turn-killing error whose message says "turn it off in the model settings" while the setting is nowhere on screen. Escape hatch first.
  - *No server-side rejection of `fastMode: true`.* Tempting, but `PUT /models` is a whole-table replace: one pre-existing annotation on a now-unsupported row would fail the entire save, and the user could no longer edit anything on the page. The web hides it, the runtime catches it, the API stays permissive.
- **(B) Warn before enabling.** Flipping the switch ON opens the app's standard `ConfirmModal` (same stacked-on-the-dialog pattern the save/default/remove confirmations already use, `models-page.tsx:1805`): premium billing, plus the note that the recorded per-token prices are not adjusted so the Cost center will under-report (finding 2), plus — only for `"anthropic"` — that fast mode is a limited research preview and requests return 429 until the organization is granted access. Turning it OFF is immediate, no confirmation. Both locales.
- **Keeping the runtime `UnsupportedParameterError` catch**, per the note in the task, and it is genuinely still load-bearing after (A) rather than belt-and-braces theater. Four ways a rejection still reaches the wire: a custom endpoint we deliberately allowed (above); `ANTHROPIC_BASE_URL=bedrock://…` or `CLIENT_TYPE=…` in the *server's* environment, which override the entry and are invisible to any pure function of the config; a stale `fast_mode = true` from a hand-edited TOML or a CLI flag; and AgentHub version drift, where a client we currently classify as supporting could start refusing. The one thing I am changing is the message — it must name where the switch lives, and after (A) that pointer has to stay true even for a row whose toggle the rule would otherwise hide, which is exactly why the stored-`true` escape hatch above exists.

hiyouga:
Applied in `eba64688`:

- **Fast mode is only offered where it can be served.** New `fastModeProtocol(modelId, clientType?, baseUrl?)` in `packages/core/src/state/model-catalog.ts`, next to `resolveModelEnv` and mirroring `AutoLLMClient`'s branch order rather than listing models — so #325's Gemini 3.7 / GLM-5.3 / OpenRouter rows get the right answer with no edit here. It returns the protocol that would carry the parameter (`"openai"` → `service_tier`, `"anthropic"` → `speed` + beta) or `undefined`, and reads routing from `client_type` else the model id, checking the two claude5 carve-outs (`bedrock://` base URL, any id containing `4-6`) against the raw entry the way the client does. Ids AgentHub cannot route return `undefined` too — no client, no fast tier.
- **The dialog's switch is gated on that rule**, recomputed from the live form so editing the id, protocol or base URL updates it as it is typed. Fifteen built-in rows lose the switch: `deepseek-v4-*`, the six `gemini-3.*`, `glm-5*`, `kimi-k3`/`k2.6`/`k2.5`, and `claude-sonnet-4-6`.
- **A row that already stores `fast_mode = true` keeps its switch** even where the rule would not have offered one, with an amber "not supported, turn it off" line. Without this, a value from a hand-edited TOML or `--fast-mode` would strand the user: the runtime error says "turn it off in the model settings" and the setting would not be there.
- **Enabling raises a warning** (shared `ConfirmModal`, stacked like the existing save/remove ones, alert-triangle tone): premium billing — MiniMax 1.5x, separate OpenAI/Anthropic premium lists — plus the note that the recorded per-token prices do not follow, so the Cost center under-reports; Anthropic-protocol models add the research-preview 429. Disabling stays one click. Both locales.
- **Fast mode is badged in the model list** — a standing premium-billing choice should not require opening the dialog to notice.
- **CLI**: `--fast-mode` on a model whose client rejects it still writes the entry but warns on stderr naming `--no-fast-mode`. A warning, not a refusal: an entry may point at an endpoint whose capability is not visible from the config.
- **Tests**: the rule against every catalog row plus the routing traps (self-routing vs. the declared client type, dotted OpenRouter ids that match no branch, Bedrock, Claude 4.6 by name); the dialog's gating and its escape hatch; the CLI warning; and — from finding 3 — the previously untested wiring claim that fast mode rides the session LLM and is withheld from the meta one.
- Docs (en+zh): the models page now carries a per-client support table and the routing rule behind it, the two things the toggle cannot check (Anthropic's 429, and `CLIENT_TYPE` / `ANTHROPIC_BASE_URL` overriding the entry from the server's environment), and the confirmation; CLI/web-app/configuration pages updated; changelog entry and index line extended.

**Why the runtime `UnsupportedParameterError` catch stays** (unchanged, per review): a rejection can still reach the wire four ways — a custom endpoint we deliberately keep the toggle on, `CLIENT_TYPE`/`ANTHROPIC_BASE_URL` in the server's environment overriding the entry invisibly, a stale annotation from a hand-edited config or the CLI flag, and AgentHub version drift. Its message already names where the switch lives, and the stored-`true` escape hatch above is what keeps that pointer honest.

**Not changed, deliberately:**

- **Cost accounting for fast-mode requests stays standard-tier** (finding 2). `usage_records` stores tokens only, keyed by `(provider, model_id)`, and cost is computed at query time — so the cheap fix (a multiplier in `getPricing`) would retroactively re-price history the moment the toggle flips, and would over-charge exactly the title/`describe_image` requests the PR keeps on the standard tier. There is also no single correct multiplier across providers. Doing it right needs a per-request tier on the usage records plus split aggregates: its own PR. The confirmation copy and the docs now state the under-reporting instead of leaving it to `models.md`.
- **No server-side rejection of `fastMode: true`.** `PUT /models` is a whole-table replace; one pre-existing annotation on a now-unsupported row would fail the entire save and lock the user out of editing the page.
- **`resolveModelEnv`'s missing `gpt-5.6` branch** (finding 5) — pre-existing, belongs to the env-hint feature, needs its own test and changelog line. Noted so the two routing mirrors in that file are not mistaken for a typo introduced here.

Tests: core 810 passed / 5 skipped (env-gated e2e), web 726, cli 247, server 641. Typecheck clean for all four; prettier run on every touched file.

hiyouga:
Design-spec wording, revised — `design/` is shared and read-only, so proposing here. This **supersedes** the 06-PROTOTYPE wording in the PR description, which described the toggle as available on every model; the 05-ARCHITECTURE carve-out proposed there is unaffected and still stands as written.

**1. `design/specs/06-PROTOTYPE.md` L120** (配置弹窗 bullet) — after 「三项价格（标签自含，无独立价格标题行）与视觉开关」 insert:

> 与**快速模式开关**（默认关闭，只持久化 `true`；开启即让该模型的会话请求进入厂商的溢价快速档位，AgentHub `fast_mode`）。该开关只对「AgentHub 为其选用的 client 确实会发送该参数」的模型显示——按 client_type、缺省时按 model_id 复刻 AgentHub 的路由规则判定，Gemini / GLM / Kimi / DeepSeek、Bedrock 上的 Claude 与 Claude 4.6 系列 id 均不显示；已存有 `fast_mode = true` 的条目即便判定为不支持也保留开关（并标注不支持），以便随时关闭。开启态下方一行小字说明「输出更快，按厂商的溢价档位计费；成本中心仍按条目记录的标准单价统计」。后台请求（标题生成、describe_image 代读）不携带。

**2. `design/specs/06-PROTOTYPE.md` L122** (提示与确认 bullet) — the existing sentence lists the config-writing actions that need confirmation; fast mode is a different case (it confirms on toggle, before anything is written), so append:

> 此外，快速模式开关**打开**时先弹二次确认，说明溢价计费（MiniMax 为标准价 1.5 倍，OpenAI 与 Anthropic 另有溢价价目表）、条目记录的单价不随之调整因而成本中心会低估，Anthropic 协议的模型另说明其为限量 research preview、未获授权时请求返回 429；关闭不需确认（这是不支持该参数的模型的退出路径）。

**3. `design/specs/06-PROTOTYPE.md` L125** (连通性测试 bullet) — extend 「可带尚未保存的 api_key / base_url / client_type」 to:

> 可带尚未保存的 api_key / base_url / client_type / fast_mode（前端恒发送弹窗当前开关值，测试即测所配档位，被拒在保存前即可发现）

One spec-level note worth a maintainer decision, not written into the wording above: the model card's badge row (视觉 / 免费 / 视觉代理) now also carries a 快速 badge when fast mode is on. It is implemented, but the badge set is spec'd in 06-PROTOTYPE and should be listed there if you want it kept.

hiyouga:
Applied in 741f8632 (plus merge 24ec4a77):

- **Fast mode and vision support now share one row.** Reused the dialog's existing two-up grid (`grid grid-cols-2 items-start gap-2` — the same one carrying context window / max output tokens), so no new layout primitive. Labels, hover titles, hint lines, aria-labels and dark-mode styling are unchanged.
- **The row degrades in both directions.** Neither switch is guaranteed: vision is read-only catalog metadata on preset models, and fast mode is withheld wherever the routed client rejects the parameter — so the pair is a coincidence, not an invariant. New `capabilityRow` helper drops the row when neither applies (an empty grid would still draw the parent's `space-y` gap) and gives a lone switch the full width instead of a half-width cell beside dead space. Unit tests pin all four combinations, including the common vision-only case.
- **No overflow at phone width.** That grid already holds two full text inputs at this width; two inline switches are strictly narrower (~130px of label + switch in a ~160px cell on a 360px viewport), and the hint lines wrap inside their cell.
- **Merged `origin/main`** (#325, #327, #328). Two conflicts, both resolved keeping each side: `testModel` now normalizes the client type through `canonicalClientType` *and* resolves `fastMode` from the form draft; `models-page.tsx` took main's `openai` → `openai-chat` rename alongside the `fastMode` row field. The rename does not affect the fast-mode gate — `fastModeProtocol` mirrors `AutoLLMClient`, where `openai-chat` and the deprecated bare `openai` alias hit the same OpenAI-protocol branch.
- Changelog entry pair extended to describe the shared row; docs unchanged, since they document the toggle's gating and confirmation rather than its placement.

Fast-mode behavior is untouched: the switch still appears only where the client can serve it, and enabling it still confirms first.

Verified: typecheck + unit tests for core (814), server (649), cli (247) and web (751), all passing; prettier clean on touched files. Per instructions the full e2e suite was not run.

PR #327, opened 2026-08-19
fix(web,server): removable exited processes, icon-only copy feedback, single-line trace file row

### LINKED ISSUE #312: exited session can be removed
会话进程

export PATH="$HOME/.local/bin:$PATH" && pnpm dev
2026-08-18 20:04 · pid 93426
已退出

会话右上角的进程列表里面。已退出的应该支持从列表里面删除掉

另外所有复制按钮，复制后从复制按钮切换成对勾按钮就行，不需要已复制三个字

轨迹文件只写单行文件名就行了，支持复制完整路径和点击进入轨迹查看页面。

hiyouga:
Implements all three asks of issue #312 (session details-card UX):

1. **Exited session processes can be removed from the list.** The details card's process
   list keeps entries whose command already exited; those rows now carry a **Remove**
   button (styled like the existing Stop button, next to the "已退出/exited" label) that
   deletes the entry. Running rows keep Stop only — a bare removal must never
   surprise-signal a live process group.
2. **All copy buttons confirm with the check icon alone.** After a successful copy the
   button's copy icon swaps to a check for 1.5s and the tooltip flips to "已复制/Copied";
   the transient visible "已复制/Copied" **text label** is gone everywhere. The three
   toast-based outliers ("Copy prompt" in the memory tab ×2 and the skill import dialog)
   move onto the shared convention: their glyph flips, their label never changes, no toast.
3. **Trace file row shows the single-line file NAME.** No more full path wrapped over
   several lines: the name is the click target (deep-link into the Trace browser focused
   on the session, unchanged), the full path lives in the tooltip and in a new
   copy-full-path button beside the name.

Fixes #312

### server
- `src/runtime/session-manager.ts`: new `removeProcess(sessionId, processId)` →
  `"removed" | "running" | "not_found"`. Refuses running processes; the removal of an
  exited entry reuses the kill path (core's registry removal signals the dead group
  defensively and drops the row, exactly like `input_command`'s post-exit reap), and the
  only race — an entry seen running that exits before a retry — is benign (caller
  refreshes and sees the exited row).
- `src/http/routes/sessions.ts`: new `DELETE /api/sessions/:sessionId/processes/:processId`
  — 204 on removal, 409 `process_running` while it runs (stop it instead), 404
  `process_not_found` when the id is unknown or the runtime is gone. Same
  `resolveSession` auth semantics as the sibling routes.
- `test/session-processes.test.ts` (new): first coverage of the processes routes at all —
  GET serialization, DELETE 204/409/404 (including double-delete and unloaded-runtime),
  kill-still-works, and foreign/unknown-session 404s. 6 tests, modeled on
  `retry-now.test.ts`'s fake-RuntimeSession injection.

### web
- `src/components/ui/copy-button.tsx`: `showCopiedText` prop removed (feedback is the
  icon swap + tooltip flip only); new exports `CopyCheckGlyph` (the copy→check glyph
  pair) and `ROW_COPY_CLASS` (the shared look of the copy button beside mono value rows,
  previously duplicated inline at two sites).
- `src/features/chat/chat-page.tsx`: exited process rows gain the Remove button +
  `onRemoveProcess` handler (mirrors `onKillProcess`; 404 and 409 are swallowed — the
  follow-up refresh shows the truth). Trace file row shows `pathFileName(tracePath)` on
  one line (tooltip = full path) with a copy-full-path `CopyButton`; the name keeps the
  existing `/traces?agentId=&sessionId=` deep-link. Session id row loses
  `showCopiedText`.
- `src/features/agents/agent-settings-page.tsx`: State path row loses `showCopiedText`.
- `src/features/agents/memory-tab.tsx`, `skills-tab.tsx`: the three "Copy prompt"
  buttons switch from `navigator.clipboard` + success toast to the shared `useCopied` +
  `CopyCheckGlyph` (label constant, glyph flips).
- `src/api/endpoints.ts`: `removeSessionProcess` (DELETE).
- `src/lib/file-path.ts`: new `pathFileName()` — final path segment for display, splits
  on both separators (Windows Trace paths arrive with `\`), never returns blank.
- `src/lib/strings.ts` + `strings-en.ts`: new `chat.processRemove` (移除/Remove) and
  `chat.copyTracePath` (复制完整路径/Copy full path); removed now-unused
  `memory.editCopied` and `skills.importCopied`. `common.copied` stays (tooltip flip).
- `test/file-path.test.ts`: 6 new `pathFileName` cases.
- e2e: `processes.spec.mjs` — trace-row assertions tightened to the name-only display,
  and a new test drives the new mock branch ("background exit test", `sleep 0.5` past a
  300ms yield) through exited-row Remove → empty server-side list, then the trace row's
  copy-full-path (clipboard = absolute path ending in the shown name) and the deep-link.
  `chat-tweaks.spec.mjs` — the Session-id copy assertions now expect the icon-only
  feedback (tooltip flips to 已复制 and back, no text content, clipboard checked).
  `mock-llm.mjs` — the new branch.

### docs
- `content/server-api.{en,zh}.md`: the session endpoint table now lists all three
  `/processes` routes (the existing GET and kill POST had never been documented) plus
  the new DELETE.
- `content/web-app.{en,zh}.md`: new "Header Stats and the Details Card" subsection
  (processes list with Stop/Remove, trace file row, the site-wide copy feedback).

### changelog
- `changelog/unreleased/2026-08-18-session-details-and-copy-ux.md` + index line.

## Tests run

- `pnpm --filter @prismshadow/penguin-server test` — full suite green: 68 files /
  644 tests (including the new `session-processes.test.ts`, 6/6). One earlier run under
  heavy parallel-agent machine load hung in the untouched
  `test/reset-admin-password.test.ts`; it passes in isolation and in the quiet full run.
- `pnpm --filter @prismshadow/penguin-server typecheck` — clean.
- `pnpm --filter @prismshadow/penguin-web test` — 60 files / 727 tests, all pass.
- `pnpm --filter @prismshadow/penguin-docs test` / `typecheck` — clean.
- Playwright e2e against a freshly built server + web bundle: `processes.spec.mjs`
  (both tests, including the new exited-removal + trace-row-actions one) and
  `chat-tweaks.spec.mjs` (updated copy-feedback assertions) **pass**. `chat.spec.mjs`
  fails at line 330 (sidebar group-collapse: the session title is still visible after
  collapsing the first group) — verified **pre-existing on the base commit** via
  `git stash` + base rebuild: identical failure with this PR's changes fully removed;
  not chased per the worktree brief.
- Prettier run on all touched files.

## Design docs

Issue #312 changes behavior specified in `design/specs/06-PROTOTYPE.md` (the chat-page
details-card paragraph, currently line 56). Proposed edits — three sentence-level
replacements inside that paragraph, plus a changelog line:

1. **Copy feedback** — replace
   「旁附复制按钮——复制反馈落在按钮处：短暂显示对勾 + 「已复制」，「Session id」标签不变；全站复制按钮统一此口径，反馈在按钮上而非替换标题」
   with
   「旁附复制按钮——复制反馈落在按钮处：图标短暂切换为对勾、tooltip 转「已复制」，不再附加可见的「已复制」文字，「Session id」标签不变；全站复制按钮统一此口径，反馈在按钮上而非替换标题；带文字标签的复制按钮（如「复制 Prompt」）标签保持不变、仅切换图标，不再弹出提示」

2. **Session processes** — replace
   「运行中的行带「停止」按钮——杀停整个进程组并从列表移除（如对话起的 localhost 开发服务器可在此关闭）」
   with
   「运行中的行带「停止」按钮——杀停整个进程组并从列表移除（如对话起的 localhost 开发服务器可在此关闭）；已退出的行保留「已退出」标注并带「移除」按钮——仅把该条目从列表删除（`DELETE /api/sessions/:id/processes/:processId`，仅接受已退出的条目：仍在运行返回 409 `process_running`，运行中的进程只能经「停止」结束）」

3. **Trace file row** — replace
   「标签下以等宽字展示当前 Trace 文件完整路径，路径本身即链接（下划线样式，点击深链到轨迹观测；尚无 Trace 时隐藏）」
   with
   「标签下以等宽字**单行**展示当前 Trace 文件的**文件名**（完整路径见悬停提示，旁附「复制完整路径」按钮），文件名本身即链接（下划线样式，点击深链到轨迹观测；尚无 Trace 时隐藏）」

4. **08-CHANGELOG.md** — new dated entry (top of the list):
   「2026-08-18 会话详情卡三处体验修正（issue #312）：**会话进程**列表已退出的行带「移除」按钮，可把条目从列表删除（新增 `DELETE /api/sessions/:id/processes/:processId`，仅接受已退出条目、运行中 409；processes 三个端点补进 server-api 文档）；**全站复制按钮**反馈收敛为图标转对勾 + tooltip 转「已复制」，不再显示「已复制」文字（记忆页与技能导入的「复制 Prompt」从 toast 迁到同一口径）；**轨迹文件**行改为单行仅显示文件名，完整路径入悬停提示与新增的「复制完整路径」按钮，点击文件名仍深链轨迹观测（impl #<PR>）。」


hiyouga:
## Review

All three asks in #312 land: exited rows get a Remove button backed by a new `DELETE /api/sessions/:sessionId/processes/:processId`, every copy affordance confirms with the check glyph alone, and the Trace row is a single-line file name with copy-full-path plus the deep link. The server route reuses the existing `resolveSession` + `requireProjectAccess` semantics (so a foreign session is a 404 exactly like kill/abort), refuses a running process with 409 rather than surprise-signalling a live group, and comes with a real integration test. Both locales, both doc languages and the changelog are updated, and the clipboard audit is clean — `navigator.clipboard.writeText` now appears exactly once in `packages/web/src`, and `editCopied`/`importCopied` are gone from both catalogs with no remaining references. What I found is one state bug in the shared copy hook, one accessibility gap that the icon-only feedback makes load-bearing, one data-loss surprise behind the new Remove button, and two nits.

### Findings

1. **`useCopied` never clears its timeout** (`packages/web/src/components/ui/copy-button.tsx:32`) — correctness, two ways it bites. (a) Repeated clicks on the same button: the first click's timer still fires `COPIED_MS` after the *first* click, so a second copy at t=1.4s shows the check for 100ms. Now that the check is the only feedback, a check that vanishes right after the click reads as "the copy didn't take". (b) Unmount inside the window: the Trace-row click calls `setInfoOpen(false)` itself, closing the popover that hosts the copy buttons; the memory / skill-import modals close the same way. The pending timer then calls `setCopied` on a torn-down component. Fix: hold the handle in a ref, clear it at the top of `flash` and on unmount.

2. **The copied confirmation is silent for screen readers** (`packages/web/src/components/ui/copy-button.tsx:73`) — the accessible name is a static `aria-label={label}`, so the `title` flip is never announced (`title` only contributes when there is no `aria-label`), and the glyph is `aria-hidden`. This is not a regression — with `aria-label` present, the old visible "已复制" text was not announced either — but the PR makes the icon the entire message, so the gap is now the whole story. Fix: a visually hidden `role="status" aria-live="polite"` region beside the control, carrying `S.common.copied` while copied; same for the three `CopyCheckGlyph` text buttons, whose label deliberately does not change.

3. **Remove silently discards the process's captured output** (`packages/server/src/runtime/session-manager.ts:960`) — the implementation is right (it is exactly `input_command`'s post-exit reap: `killBackgroundCommand` → `BackgroundRegistry.remove` → `ManagedSession.kill`), but the user-facing consequence is stated nowhere. The list entry is the only handle on that process's buffered output: after a removal, `input_command <process_id>` answers `unknown process_id`, so the agent can no longer be asked to read it. The likely sequence is the one in the issue's own screenshot — `pnpm dev` dies, the user tidies the "已退出" row away, then asks why it died, and the output is already gone. One click, no confirmation, no undo. I do not think a confirm modal belongs here (#312 asked for plain removal), but the affordance and the docs should say what leaves with the row; I will add a hint `title` on the button (both locales) and a docs clause.

4. **`pathFileName`'s comment claims a guarantee the code does not make** (`packages/web/src/lib/file-path.ts:107`) — "A backslash inside a POSIX directory name cannot leak in — only the part after the last `/` is inspected" is not what `Math.max(lastIndexOf("/"), lastIndexOf("\\"))` does: a backslash *after* the last slash does cut, so `/home/u/we\ird.jsonl` → `ird.jsonl`. The unit test named after that claim passes for a different reason (the later `/` wins the max). Harmless for the server-generated `<sessionId>_NNN.jsonl` names this is used on, but the comment and the test name should describe the rule the code actually implements.

5. **Stale-session write after the request settles** (`packages/web/src/features/chat/chat-page.tsx:739` and `:761`) — both `onKillProcess` (pre-existing) and the new `onRemoveProcess` fire their follow-up `getSessionProcesses(selected.sessionId)` and `setProcesses` unconditionally in `finally`. Switch sessions while the request is in flight and the previous session's list paints the new session's card; with the popover closed and nothing running, `processesCanChange` is false, so no poll ever corrects it. Cheap to guard on the still-selected session, and the two handlers are otherwise identical.

### Observations (verified, no change proposed)

- **The 409 is unreachable from the UI, and that is fine.** `running` is `!exited` and `exited` is monotone, so a row the client renders as "exited" can never be running server-side. The guard is for direct API callers; the race the doc comment describes ("observed running, exits before the retry") is the benign direction.
- **Nothing to persist.** The list is the runtime registry only — `listProcesses` returns `[]` for an unloaded entry, and a resumed session starts with a fresh environment — so a server restart, an eviction or a reload already shows an empty list. There is no on-disk view to diverge from; a second tab converges on its next poll or on reopening the card.
- **Insecure origins still lie.** `writeClipboard` is `navigator.clipboard?.writeText(...)`, and `navigator.clipboard` is *undefined* on a non-localhost `http://` origin — the LAN-access scenario `configuration.en.md` documents for `PENGUIN_PREVIEW_ORIGIN`. The check then flashes for a copy that never happened. Pre-existing and equally true of the old "已复制" text, so out of this PR's scope, but worth a follow-up (an `execCommand("copy")` fallback, or withholding the check when there is no clipboard API).
- **The trace name can truncate on a phone.** It is `<sessionId>_NNN.jsonl` (~45 chars); the card is `w-96` shrinking to `calc(100vw-1.5rem)`, so tail truncation eats the `_NNN` index — the only part the Session id row above does not already show. The `title` and the copy button recover it, so I am leaving it; a middle ellipsis would be the nicer fix.
- **Auth and CSRF are consistent.** DELETE goes through the same `resolveSession` + `requireProjectAccess` as kill/abort, so a project member can remove another member's session process exactly as they can already stop it. The bodyless DELETE carries no `Content-Type`, which `jsonOnlyWrites` deliberately allows, and SameSite=Lax keeps it off cross-site requests.
- **Double-click is safe.** The UI's `procBusy` guard blocks the second click, and a second DELETE from another tab is a benign 404 the handler swallows before refreshing.

None was given for this PR. From my own findings I will apply: the `useCopied` timer fix (1), the screen-reader live region (2), the "output goes with the row" hint plus docs wording (3), the `pathFileName` comment/test wording (4), and the stale-session guard shared by both process handlers (5). I will not add a confirm dialog for Remove, and I will not touch the insecure-origin clipboard fallback or the trace-name middle-ellipsis — both are pre-existing and outside this PR's subject.

hiyouga:
Applied in d6d4f47c:

- **`useCopied` holds its reset timer** — a repeat click restarts the full flash window instead of inheriting the previous click's countdown, and the timeout is cleared on unmount (the Trace row closes the details card itself, so this was reachable).
- **Copy confirmation is no longer silent** — every affordance renders a visually hidden polite live region beside the control; the accessible name still stays the copy action. Deliberately a bare `aria-live` rather than `role="status"`: that role is this app's running-spinner marker (`status-icon.tsx`), and `malformed.spec.mjs:90` asserts no `[role="status"]` survives a finished turn — using it here would have broken that spec. `chat-tweaks.spec` now pins the region appearing once and clearing with the flash.
- **Remove says what it discards** — the entry owns the ManagedSession holding that process's captured output, so after a removal `input_command` on the same `process_id` fails. New `processRemoveHint` tooltip (both locales), a clause in `web-app.{en,zh}.md` and the `server-api.{en,zh}.md` DELETE row, a note on `SessionManager.removeProcess`, and a changelog paragraph. **No confirm dialog** — #312 asked for a plain one-click removal of a dead row, and a modal on every tidy-up would be worse than the honest tooltip; happy to add one if you disagree.
- **Stale-session refresh guarded** — `onKillProcess` (pre-existing) and `onRemoveProcess` now share one body, and its post-request `getSessionProcesses` is applied only while its own session is still selected. Switching sessions mid-request used to paint the old session's process list into the new session's card, with `processesCanChange` false and therefore no poll to correct it.
- **`pathFileName` comment corrected** — it claimed "only the part after the last `/` is inspected"; the code cuts at the last separator of *either* kind. Reworded, and the real rule is pinned by an extra test (the existing test passed for a different reason than its name claimed).

Deliberately not changed:

- **Insecure-origin clipboard.** `navigator.clipboard` is undefined on a non-localhost `http://` origin — the LAN access `configuration.en.md` documents — so the check flashes for a copy that never happened. Pre-existing and equally true of the old "已复制" text; the fix is an `execCommand` fallback with its own browser quirks, which does not belong in a feedback PR.
- **Trace-name truncation.** `<sessionId>_NNN.jsonl` fits `w-96` on desktop and truncates on a phone, eating the `_NNN` index. The `title` and the copy button recover it; a middle ellipsis would be the nicer fix, separately.
- **No unit test for the timer restart.** `packages/web`'s vitest is deliberately `environment: "node"`, pure logic only — the hook cannot be rendered there. The e2e covers the flash and the flip-back; the restart-on-repeat-click path is verified by reading.

Verified: `packages/web` typecheck + 728 unit tests, `packages/server` typecheck + 644 unit tests (all green, run after the change), prettier clean on the touched files. E2E not run per the round's ground rules — the two specs this touches (`chat-tweaks`, `processes`) were checked by reading, and the `role="status"` collision above is the one thing that would have broken.

PR #328, opened 2026-08-19
fix(web): scroll truncated sidebar titles into view on hover / keyboard focus

### LINKED ISSUE #309: Sidebar should reveal the full title for truncated conversations
## Description

Long conversation titles are truncated with an ellipsis in the sidebar. When the pointer is placed over a truncated title, the complete title is still not shown, so conversations with the same prefix are difficult to distinguish.

## Steps to reproduce

1. Create a Session whose title is wider than the available sidebar row.
2. Open the chat sidebar.
3. Hover over the truncated Session title.

## Current behavior

The title remains truncated and no visible tooltip or popover reveals the complete title.

## Expected behavior

Hovering a truncated title should reveal the full title. The same information should also be available on keyboard focus; touch users should have an equivalent accessible path.

## Screenshot

Cropped sidebar showing truncated conversation titles

_Cropped from a local desktop test environment._

## Possible approach

Add an accessible tooltip for ellipsized titles and show it on both hover and keyboard focus.

hiyouga:
## What & why

Long conversation titles in the chat sidebar are ellipsis-truncated, and hovering revealed nothing beyond the easily-missed native tooltip — conversations sharing a prefix were indistinguishable. Per the maintainer's direction on the issue (可以用滚动的方式来显示标题全文), a truncated title now reveals its full text by scrolling (marquee-style) rather than by a tooltip: while the row is hovered or holds keyboard focus, the title slides left at a constant reading speed until its tail is in view, holds there, and snaps back the instant the hover/focus ends.

Fixes #309

## How it works

- **Measured, not assumed**: `Truncated` (the existing measure-overflow-then-tooltip helper) now records the overflow *distance* (`scrollWidth - clientWidth`, same 1px subpixel tolerance as before). Titles that fit get neither tooltip nor scroll; the measurement re-runs via the existing `ResizeObserver` + `[text, className]` effect (sidebar resize, rename, active-row font-weight swap).
- **Pure CSS animation, JS only sets numbers**: when a title overflows (and motion is allowed), `Truncated` adds a `title-scroll` class and two custom properties — `--title-scroll-shift` (the measured overflow) and `--title-scroll-ms` (duration proportional to it, 60px/s clamped to 0.35–4s, computed by the new pure helpers in `lib/title-reveal.ts`). New `styles.css` rules trigger on the row: `[data-title-reveal]:is(:hover, :has(:focus-visible))` swaps the ellipsis for `text-overflow: clip` and transitions the inner span's transform after a 0.3s start delay (so a mouse merely passing across the sidebar doesn't set every long title in motion). No timers exist, so none can leak.
- **Instant, artifact-free reset**: at rest the inner span is a plain inline element — not transformable — so leaving the row removes the transform and restores the ellipsis in the same frame; there is no return transition for the ellipsis to fight, and idle rendering is pixel-identical to the previous markup (verified, see below).
- **Reduced motion / touch / screen readers**: under `prefers-reduced-motion` the scroll class is withheld (live `matchMedia` listener, the `sheet.tsx` pattern) and the reveal falls back to the existing conditional `title` tooltip — which also serves touch long-press; a CSS `transform: none !important` override in the existing reduced-motion block covers the instant before the listener fires. The full title always remains in the DOM, so screen readers announce it regardless of the visual clipping.
- **No interaction or layout regressions**: the scroll is a transform clipped inside the title's existing `overflow: hidden` box — zero layout shift, the trailing time/menu slot is never painted over, and row click, HTML5 drag-reorder, and the hover action buttons are untouched (drag suppresses `:hover`, so nothing scrolls mid-drag). `:focus-visible` (not `:focus-within`) scopes the keyboard trigger to actual keyboard focus; inside `:is()` the `:has()` alternative is forgiving, so engines without it keep the hover trigger.

## Key changes by package

- **web**
  - `src/lib/title-reveal.ts` (new): pure helpers `revealDistancePx` (overflow with the 1px tolerance) and `revealDurationMs` (proportional duration with floor/ceiling clamps), unit-testable in the node-only vitest setup.
  - `src/components/ui/truncated.tsx`: opt-in `scrollReveal` prop; measures distance instead of a boolean, sets the CSS variables, withholds the animation under reduced motion. Default rendering for all other callers is byte-identical to before (bare text node, conditional `title`).
  - `src/components/layout/sidebar.tsx`: `data-title-reveal` on the session-row and parked-draft-row containers, `scrollReveal` on their `Truncated`s; comments updated.
  - `src/styles.css`: the `[data-title-reveal] .title-scroll` trigger rules plus a reduced-motion override in the existing block.
  - `test/title-reveal.test.ts` (new): distance tolerance/fit/overflow cases; duration proportionality, clamps, rounding, and zero-distance.
- **changelog**: `changelog/unreleased/2026-08-18-sidebar-title-scroll-reveal.md` + index line.
- **docs**: none — the web-app guide does not describe the sidebar session-list title behavior.

## Tests run

- `pnpm --filter @prismshadow/penguin-web test` — 61 files / 729 tests passed (includes the new `title-reveal.test.ts`).
- `pnpm --filter @prismshadow/penguin-web build` — clean; verified the emitted CSS keeps the `:is(:hover,:has(:focus-visible))` selectors intact through Tailwind 4/LightningCSS.
- Headless-Chromium behavioral verification of the exact CSS + DOM contract (scratch script, not committed): idle rendering pixel-identical to the previous markup (ellipsis intact); nothing moves during the 0.3s delay; scroll animates and holds at exactly the measured overflow; a click mid-scroll lands on the row button; unhover snaps back instantly and repaints pixel-identical to idle; Tab focus starts the scroll and blur resets it; with reduced motion emulated the text never moves.
- Playwright e2e (mock-LLM stack on dedicated ports; the CI machine was concurrently running several other agents' suites, so full-suite runs were load-flaky): the sidebar-relevant specs were run on this branch and every failure was re-run on the base commit (`git stash` + web rebuild) for attribution.
  - Passing on this branch: `draft-park.spec.mjs` (exercises the parked-draft row this PR touches), `chat-tweaks.spec.mjs`, `compact-abort.spec.mjs:27`, `layout.spec.mjs` login + document-growth tests, `paging`/`layout`/`draft-user-scope` partials per below.
  - Pre-existing (identical failures on base, unrelated surfaces): `compact-abort.spec.mjs:95`, `compaction.spec.mjs:26/:162` (compaction marker `[压缩]` never renders), `layout.spec.mjs:83` (models-page `@390` text-overlap count 14 — same 14 on base), `layout.spec.mjs:373` (mobile approval dropdown), `paging.spec.mjs:26` (20 rows expected, 10 render — same on base), `draft-user-scope.spec.mjs:18` (login-page 90s timeout under load). Branch and base fail sets are identical.
  - `chat.spec.mjs` (the 12-minute mega-spec) timed out under the concurrent-suite machine load and was not base-compared for cost reasons; its failure mode (mid-stream expect timeouts) is unrelated to sidebar title rendering.

## Design docs

`design/specs/06-PROTOTYPE.md` currently says (Session list section): 「每条 Session 支持重命名、归档与删除；标题被截断时悬停显示全文。」 The mechanism is now a scroll reveal rather than a tooltip, so I propose updating that sentence to:

> 每条 Session 支持重命名、归档与删除；标题被截断时以滚动方式显示全文——悬停该行或键盘聚焦时，标题以匀速横向滚动到末尾并停住（时长与被截断宽度成正比），移开后立即复位；系统开启「减少动态效果」时退回原生 title 提示（触屏长按同样经 title 可达，屏幕阅读器始终读到 DOM 中的完整标题）。


hiyouga:
## Review

The approach is right and the implementation is unusually careful: the overflow is measured rather than assumed, the animation is pure CSS with no timers, idle rendering really is byte-identical to the pre-PR markup, and the measurement re-runs on every input that can change it. I re-verified the behavioral claims in headless Chromium against the exact CSS and DOM this branch emits, and most of them hold exactly as advertised. Two things do not: the advertised 0.3s "don't react to a passing mouse" hold does not actually delay the visible reaction, and the touch story in the comments is not true of real touch browsers. There is also a shape issue — the reduced-motion guarantee and a per-row `matchMedia` subscription both exist in JS to do something the repo's existing global CSS block would do for free if the reveal were an `animation` instead of a `transition`.

### Findings

1. **The 0.3s hold does not delay the visible reaction — the ellipsis drops instantly on hover** (`packages/web/src/styles.css:313-320`). The delay is attached only to the `transition`; `text-overflow: clip` and `display: inline-block` are plain declarations that take effect the moment `:hover` matches. Measured in headless Chromium 120ms into a hover (i.e. 180ms before anything is meant to move): computed style is already `text-overflow: clip` / `display: inline-block`, `transform` is still identity — and the row's pixels already differ from idle. The reason is `display: inline-block`: Blink does not paint an ellipsis for an overflowing *atomic inline*, so the "…" disappears on contact regardless of the `text-overflow` declaration (confirmed separately — with `text-overflow` left at `ellipsis`, an `inline-block` child simply hangs 340px past the box with no ellipsis drawn). Net effect: a mouse merely crossing the sidebar makes every long title flick its ellipsis off and back on, which is precisely the flicker the delay was introduced to prevent. **Fix**: drive the reveal with a delayed `animation` instead of a delayed `transition`, and carry `display: inline-block` (inner span) and `text-overflow: clip` (outer span) inside the keyframes with `forwards` fill only. During the delay there is no fill, so the row stays byte-identical to idle; motion and clip begin in the same frame. Verified: pixels unchanged through the delay, then scrolls and settles flush (0.09px from the right edge).

2. **The reduced-motion guarantee is carried by JS state where CSS would carry it for free** (`packages/web/src/components/ui/truncated.tsx:28-39,79`). `usePrefersReducedMotion` gates the class, and `styles.css:391` adds a `transform: none !important` override for "the window before its media listener fires" — a window that does not exist, since the `useState` initializer reads `matchMedia` synchronously on first render. Once finding 1 turns the reveal into an `animation`, the repo's existing global `@media (prefers-reduced-motion: reduce) { *, ::before, ::after { animation: none !important } }` (`styles.css:372-379`) neutralises the whole thing by itself. Verified with reduced motion emulated: the text never moves, `display` stays `inline`, `text-overflow` stays `ellipsis`, the ellipsis is intact, and the `title` tooltip remains the fallback. That makes the hook *and* the extra override redundant, and moves the guarantee from "JS state that could be stale" to "CSS that cannot be".

3. **One `matchMedia` subscription per `Truncated` instance** (`packages/web/src/components/ui/truncated.tsx:28-39`). The hook runs unconditionally, so it also fires for the three callers that never opt into `scrollReveal` (`chat-page.tsx:1264`, `traces-page.tsx:423`, `benchmark-page.tsx:122`), and it constructs two `MediaQueryList` objects per mount (initializer + effect). A sidebar listing 200 sessions registers 200 `change` listeners for one global bit. Nothing leaks — the effect cleans up correctly — but with finding 2 the whole hook disappears rather than needing to be optimised.

4. **`title` does not give touch browsers a "long-press peek"** (`truncated.tsx:21`, `sidebar.tsx:2216`, and the changelog entry). Mobile browsers do not surface `title` on long-press as a general rule (Android Chrome never does; iOS shows it only in some link contexts), so the sentence overstates what ships. What I could verify: on an emulated touch device (`hover: none` matching), Chromium applies sticky `:hover` on tap and the reveal *does* run to completion — but the row's own tap handler calls `onNavigate?.()` (`sidebar.tsx:740,764`), which closes the mobile drawer, so the scroll plays as the sidebar leaves. So the issue's "touch users should have an equivalent accessible path" is not really met. This PR does not *regress* it (the `title` attribute is exactly what was there before), and a genuine touch affordance is a bigger design question than this PR should settle — but the comments and changelog should stop claiming it is solved.

5. **The CSS↔JS contract is pinned nowhere** (`packages/web/test/title-reveal.test.ts`). The unit tests cover only the pure math, which was never the fragile part. Nothing fails if `title-scroll`, `title-scroll-text`, `--title-scroll-shift`, `--title-scroll-ms` or `data-title-reveal` is renamed on one side of the JS/CSS boundary, or if someone converts the reveal back to a `transition` — which, per finding 2, would silently un-do the reduced-motion guarantee with every test still green. vitest here is `environment: "node"` with no jsdom/happy-dom dependency, so a real DOM test would mean adding a dependency for one feature — not worth it. A file-level contract test that reads `styles.css` and `truncated.tsx` and asserts the four names match and the reveal is animation-driven costs ~30 lines and catches exactly those regressions.

### Verified fine (checked, no change needed)

- **Measurement freshness.** Re-runs on rename / auto-title (`text` dep), on sidebar and window resize (`ResizeObserver`), and on the active-row `font-medium` swap (`className` dep). No webfont is loaded — `styles.css:104-113` is a pure system stack — so late font loading is not a factor here.
- **Hover controls do not shrink the title.** The session row's `⋯` trigger is `absolute right-0` (out of flow) and swaps opacity only; the draft row's delete button keeps its box and swaps opacity only. So the available width does not change on hover, and no hover-time re-measure is triggered.
- **Re-measuring mid-reveal is safe.** With the transform applied, `scrollWidth` read 548 mid-scroll and at rest — identical to idle. A `ResizeObserver` or title-change re-measure landing mid-hover computes the same distance, not a collapsed one.
- **Non-truncated rows never react.** The gate is in JS (`title-scroll` is attached only past the tolerance), so the CSS matches nothing for a title that fits — confirmed: a short-title row is pixel-identical after a 2.5s hover. The 1px tolerance is byte-for-byte the pre-PR `scrollWidth > clientWidth + 1`, so the `title` tooltip's threshold is unchanged.
- **Keyboard.** Tab reaches the row button, `[data-title-reveal]` matches `:has(:focus-visible)`, and the reveal runs. The full title is a plain text node inside the button, so it is in the accessible name whether or not it is visually clipped; the `:is()` wrapper does keep the hover trigger alive on engines without `:has()`.
- **Interaction safety.** Idle is pixel-identical to the pre-PR bare text node; unhover restores it pixel-identically; a click at the row's centre mid-scroll still lands on the row button; the transform is clipped by the existing `overflow: hidden` and settles 0.09px from the right edge, so nothing bleeds into the time/menu slot.
- **Not applicable.** No collapsed rail renders these rows, and the sidebar has no drag-resize handle.
- **Docs.** Correctly untouched — no `packages/docs/content/*.md` describes the sidebar session-list title behavior.

None was given for this PR, so I am applying findings 1-5: replace the delayed transition with a delayed keyframe animation (1), drop `usePrefersReducedMotion` and the now-redundant reduced-motion override in favour of the existing global `animation: none !important` (2, 3), correct the touch wording in the comments and changelog (4), and add a file-level CSS-contract test (5). The touch affordance itself I am leaving alone and reporting, as noted in finding 4.

hiyouga:
Applied in `ce639bfa`:

- **The 0.3s hold now actually holds** (finding 1). The reveal is one delayed `animation` whose keyframes carry `display: inline-block` *and* the transform, with a `forwards`-only fill — during the delay nothing is filled, so the row paints exactly as at rest. Verified against the built stylesheet: pixels unchanged through the hold, then it scrolls and settles flush.
- **Reduced motion is now a CSS guarantee** (finding 2). The global `animation: none !important` block disables the reveal outright; the added `transform: none !important` override is gone. Verified with reduced motion emulated: nothing moves, ellipsis intact, `title` tooltip serves.
- **`usePrefersReducedMotion` deleted** (finding 3) — no `matchMedia` subscription per `Truncated` any more, so the three non-`scrollReveal` callers stop paying for it too.
- **`text-overflow: clip` dropped.** With `display: inline-block` in play Blink paints no ellipsis anyway, so the rule was a no-op there; leaving it out means an engine that ignores `display` in keyframes degrades to a clean no-op rather than a clipped-but-motionless title.
- **Touch wording corrected** (finding 4) in `truncated.tsx`, `sidebar.tsx` and the changelog — `title` is the pointer/reduced-motion fallback, not a touch path.
- **CSS contract pinned** (finding 5): 6 tests in `test/title-reveal.test.ts` assert the four names agree across `truncated.tsx` / `styles.css` / `sidebar.tsx`, that the reveal is an animation (a transition would silently escape reduced motion), and that `display` lives in the keyframes rather than on the trigger. Mutation-checked — each of those three regressions fails the suite.

Deliberately not changed:

- **A real touch affordance.** Tapping a row calls `onNavigate?.()` and closes the mobile drawer, so the sticky-hover reveal plays as the sidebar leaves; a proper touch path (long-press peek, or a details affordance in the row) is a design decision beyond this PR. Reported rather than invented.
- **The design-spec sentence.** `design/` is read-only from here; the proposed wording in the PR description still stands, with "悬停后短暂停顿再开始滚动，停顿期间标题保持原样" worth adding now that the hold is real.

`pnpm --filter @prismshadow/penguin-web test` 61 files / 735 tests pass, `typecheck` clean, `build` clean.

PR #329, opened 2026-08-19
fix(core,server,web,cli): compaction dangling-span healing, interruption-flow failures, and streamed summaries (#288, #290)

### LINKED ISSUE #288: Agent loop stops after compaction
agent loop 在生成 tool call 以后压缩，压缩后不会正常继续执行任务。
应该是生成 tool call 以后，先补一个空的回复，然后让他压缩，压缩后继续执行任务

压缩后有时候前端也会出错，后续的消息发了以后刷新就消失了。打断后还会出现未知工具

### LINKED ISSUE #290: Compact does not show texts
前端的压缩可以流式输出压缩的输出。可以适当加一个流式信号

hiyouga:
## What & why

Two compaction problems, one event-flow surgery — done without adding anything to the protocol or the LLM interface.

**#288 — "Agent loop stops after compaction."** Two concrete bugs behind the report:

1. **Post-compaction persistence.** A process death mid-compaction (crash, kill, a shutdown that outran the drive, a torn tail line) leaves a `compaction_begin` with no matching end. The session resumes fine (core's replay is dangle-tolerant) and appends the follow-up conversation to the same shard — but every stateless reader (the Web reducer, the server's message-window scanner) treats messages between the pair as compaction-internal, so **everything sent after the compaction vanished on refresh**, forever.
2. **"Unknown tool" after interruption.** The same wedge: once a later compaction's `compaction_end` un-wedged the swallowed state mid-task, tool outputs rendered without their (still swallowed) calls — the reducer creates a nameless card that displays "(unknown tool)".

Beyond the wedge, an abandoned mid-Task compaction left the loop in an ill-defined state: the run either ended empty-handed right after the compaction, or continued on bare `[tool error]` repair outputs the model had no instruction to act on. It now ends through the **ordinary interruption flow** and the compaction is **made up at the next trigger** — see below.

**#290 — "Compact does not show texts."** The compaction request's streamed output was fully dark to the frontend. The summary being generated now rides the output stream as the compaction span's **own** `partial_text` (or the complete `text`, for LLM implementations that stream nothing), positioned between the paired compaction events. The Web banner shows it being written live and keeps the full text in an expandable body once settled, with live and reloaded views showing the same text.

## Design constraints this follows

Two things were deliberately **not** added:

- **No new protocol type.** An earlier revision of this PR introduced a stream-only `compaction_delta` event. It is gone: readers already treat model messages inside a compaction span as compaction-internal, and the span's complete assistant text is already what history rebuild reads back — so the live path and the rebuild path converge on one channel instead of two parallel ones. The Trace needs no structural exception in the Writer, and the protocol keeps its invariant that every `event_msg` is recordable. A bonus falls out: the server's live tail understands `partial_text`, so joining or refreshing **mid-compaction** seeds the banner with the summary prefix generated so far rather than starting blank.
- **Nothing new on the LLM side.** An earlier revision added an optional `LLMInterface.appendExchange` so a mid-Task compaction could close the pending tool exchange before compacting. It is gone: the compaction request keeps folding the turn's outputs in (append-only, exactly as before), and `LLMInterface` is back to `streamGenerate` alone.

## Key changes by package

**core**
- `ContextEngine`: a failed or aborted **mid-Task** compaction now ends the run the way any interruption does — the turn's pending state is held as carry-over under the existing committed/not-committed binary (#85), an `abort` event closes the run (`compaction failed` for the failure case), and the next message resends that carry-over merged with the user's input, where the still-standing threshold triggers the compaction again. A boundary-compaction failure keeps the original context and ends the run as before. No synthetic continuation input is invented to keep the loop running on a context that was supposed to shrink.
- `runCompactionRequest` forwards the summary's own text to the output stream between the paired compaction events: `partial_text` fragments verbatim (all three phases, so the server's live tail opens and closes its fragment), or the complete `text` when no partial carried content — never both, so consumers see each character once. The request's other raw messages (thinking, the compaction Prompt, request events) stay Trace-only as before, and the Trace's contents are unchanged.
- Dangling-span healing: `resumeTrace` reports an unmatched `compaction_begin` (`ResumeResult.danglingCompaction`), and `resumeSession` appends a synthetic aborted `compaction_end` before continuing the shard — idempotent, best-effort, skipped for healthy traces.

**server**
- `message-window` scanner mirrors the reducer's stale-span closure: a `compaction_begin` while a span is open, an `abort` event, or a rotation's `session_meta` closes it (the engine always closes the pair before aborting, and spans never nest, so all three signals are unambiguous). `CACHE_VERSION` bumped to 2 so cached page stats recompute.
- Trace analysis (`trace-service.analyze`) clears a dangling `compactionActive` on abort events, so later turns aren't misfiled as compaction turns.
- No changes were needed for the streamed summary: the server reads Traces, and the Trace is byte-identical to before.

**web**
- Stream reducer: closes stale compaction spans on the same three signals; accumulates the span's `partial_text` onto the running banner (`CompactionItem.summaryText`), and on history replay rebuilds the identical text from the span's complete assistant output (live and reloaded views agree).
- `CompactionBanner`: shows the streamed summary under the header while running (tail-clamped to 600 chars, summary tags stripped with core's lenient `extractSummary`), and folds the full text into an expandable body once settled. `StepBanner` gains an always-visible `liveBody` slot (the settled expandable `children` behavior is unchanged; the MCP connect row is unaffected). No new user-visible strings.

**cli**
- `handlePartialText` stays quiet inside a compaction span (the renderer already tracks `compactionActive` for its token accounting), so the terminal keeps its one-line compaction progress and prints none of the streamed summary. Nested child-session views never render text at all, so a subagent's compaction cannot leak either.

**docs**
- `agent-loop` (§ Compaction), `message-flow` (compaction row), `omni-message` (event catalogue) updated in both languages.

**changelog**
- `changelog/unreleased/2026-08-18-compaction-continuity-and-streaming.{md,zh.md}`.

## Tests

- `@prismshadow/penguin-core` — 800 passed, 5 skipped. New/updated: the streamed summary appears between the paired events as `partial_text` and the complete text does not re-emit; an LLM yielding only complete messages streams that text instead; the Trace still records the compaction dialogue and never records partials; a mid-Task compaction failure ends the run with an abort and holds the right carry-over on all three absorption paths (committed-empty rejection, committed tool-calling rejection, all-transport failure); `resumeTrace` dangling-span reporting and `resumeSession` healing (idempotent).
- `@prismshadow/penguin-server` — 640 passed. A healed dangling span keeps the post-crash conversation visible in windowed reads (and windows still tile the transcript exactly); an unhealed span closes at an abort event.
- `@prismshadow/penguin-web` — 727 passed. Span `partial_text` accumulates onto the banner and renders no transcript item of its own; text after the span is ordinary output again; replay parity of the summary text; dangling-span vanish regression (healed trace renders everything; no "(unknown tool)" card); stale-span closure at a nested begin / abort / rotation session_meta.
- `@prismshadow/penguin-cli` — 246 passed, including a new case pinning that the summary streamed inside a span stays off the terminal while text after it renders.
- Typecheck clean for core / server / web / docs; `prettier --check` clean.

## Design docs

This PR changes behavior the spec pins down; proposed edits (file + wording), not applied since `design/` is shared:

1. `design/specs/05-ARCHITECTURE.md` § 上下文管理 › 压缩方式（`summarize` 条目）— **不改**：任务中途折叠本轮工具结果的现行条文与实现一致。
2. 同节压缩失败处理（「summarize 失败……下一轮触发时重试」附近）追加：
   「Task 进行中压缩失败或被用户中断时按**打断流程**收场：本轮待发状态按同一 committed 二分转入 carry-over（已提交则只剩配对修复，未提交则工具输出原样保留），产出 abort 事件（失败一路 reason 为 `compaction failed`）结束本次运行；下一条消息把 carry-over 与用户输入合并重发，阈值依然超标则在那里补做压缩。引擎不合成任何续跑注记去让运行在本该缩小的上下文上继续。」
3. 同文件「过程可见性与失败处理」— 现有例外条款（压缩请求的 `token_usage` 推入输出流、位于成对事件之间）追加同款第二条例外：
   「压缩请求**正在生成的摘要文本**同样推入输出流（位于成对压缩事件之间）——以普通 `partial_text` 分片上行（无流式产出的 LLM 实现则为完整 `text`；两者互斥，同一内容只上行一次），不新增事件类型、Trace 内容不变；压缩请求的其余原始消息（thinking、压缩 Prompt、请求事件）一如既往只写旧 Trace。前端在压缩横幅内实时展示、收束后折叠为可展开正文；历史重建从压缩区间已记录的 assistant 文本读回同一段内容，刷新前后一致。渲染对话的消费方本就把压缩区间内的模型消息视为压缩内部消息。」
4. 同文件「消息格式」压缩事件条目（`compaction_begin` / `compaction_end` 段）追加一句：成对事件之间除各次尝试的 `token_usage` 外，还会出现承载摘要文本的普通 `partial_text` / `text` 消息（**不**新增事件类型）。
5. 同文件「Session 恢复」追加：
   「重放发现文件中存在未配对的 `compaction_begin`（进程在压缩期间死亡）时，恢复方在续写该文件前先补写一条 `status: "aborted"` 的 `compaction_end`（幂等；健康文件不写）——否则无状态读取方会把其后的全部对话当作压缩内部消息隐藏。读取方（Web 归约器、服务端窗口扫描、Trace 分析）自身也在明确信号（嵌套的 `compaction_begin`、`abort` 事件、轮转 `session_meta`）处闭合悬空区间，保证治愈前受损的存量 Trace 依然可读。」
6. `design/specs/06-PROTOTYPE.md` 对话页压缩横幅一句（`compaction_begin` 与 `compaction_end` 之间显示压缩进行中横幅……）追加：
   「进行中横幅内实时滚动显示正在生成的摘要文本（尾部截断、剥离 `[summary]` 标记），完成后全文折叠为横幅的可展开正文；刷新后由压缩区间重建同一文本，压缩过程中加入亦由服务端实时尾巴补齐已生成的前缀。」

Fixes #288
Fixes #290

https://claude.ai/code/session_01NB2WpZpKW5n8EVkMV3ZNwo


hiyouga:
压缩和思考一样，默认折叠，用户点开以后才显示文本流式。而且现在流式前端有点问题

hiyouga:
Round addressing the three review points (merged `main` first — #323/#325/#327/#328 came in clean, no conflicts).

**1. Compaction after a tool call waits for the tools.** This turned out to already hold after ac36d0a1: the compaction checkpoint sits *after* `runTurn`, which returns only once every tool call of the turn has completed, so the results ride the compaction request ahead of the Prompt in their original call order — and no synthetic assistant text exists any more (`appendExchange` went with that commit). What was missing was proof, so the behavior is now pinned: parallel calls in one turn (all three results ride, in order), a gated slow tool (compaction has not started and no clock runs while the tool is outstanding — it starts only once the result lands), a failing tool (`[tool error]` output is what the summary is written from), a denied tool, and manual `/compact` (no pending turn, Prompt only). The loop still resumes on the summary after a successful mid-task compaction.

**2. The compaction row is collapsed by default, like thinking.** The summary now streams *inside* the collapsed body — same `md-body` treatment and streaming renderer `ThinkingBlock` uses — with the chevron present from the start of a summarize compaction so the row does not shift when the first token lands. `StepBanner`'s always-visible `liveBody` slot (the thing that forced the old expanded view) had no other consumer and is deleted. The body text moved into a small pure helper so it can be unit-tested; the repo has no component-test setup (node env, no jsdom), so the collapse itself is structural — `StepBanner` now has exactly one body slot, gated by `useState(false)`.

**3. An interrupted compaction is just a failed compaction.** Resume closes the unmatched span with a `failed` `compaction_end` before appending anything else, and the half-written draft is discarded — no `[context_summary]` is reconstructed, the original context stands, and the standing threshold makes the compaction up at the next trigger. The reducer drops the partial text from any compaction that did not complete, so a truncated draft is never shown as if it had been adopted. The reader-side stale-span heuristics (web reducer, server window scanner, Trace analysis) and the `CACHE_VERSION` bump are removed — closing the span at load is enough. Both regressions stay pinned: messages sent after a compaction survive a refresh, and no `(unknown tool)` cards appear after an interrupt.

Docs (`agent-loop` en/zh) and the bilingual changelog entry were rewritten for the new behavior. Verified: core 816, web 752, server 648, docs 32, cli 246 passing; typecheck clean for core/web/server; touched files prettier-formatted. No e2e run — note `packages/web/e2e/compaction.spec.mjs` still asserts the pre-`StepBanner` `[压缩]` bracket wording and was already stale before this PR.

One thing worth flagging: `packages/web/e2e/compaction.spec.mjs` aside, the compaction row remains a standalone card rather than a row *inside* the reasoning-and-tools group. Making it a group member would change task segmentation and the stats-row splice (compaction is deliberately ordered after a turn's ledger as housekeeping), so I matched thinking's *affordance and default state* rather than relocating the row. Happy to move it into the group if that is what you meant.

hiyouga:
Follow-up: `packages/web/e2e/compaction.spec.mjs` is fixed in this PR rather than left stale (4df49964).

It was broken on `main` in two ways, both owned by this PR's rendering rework:

- It matched the row as `[压缩]` — a bracketed wording that predates the `StepBanner` row (which reads `压缩` plus its outcome line).
- Its trigger no longer fired at all: the threshold came from the model's 240-token context window, but the engine ignores windows under 4096 as implausible and falls back to the 128000 default, so the mock's ~190-token usage never crossed it. The threshold is now set explicitly on the Agent (`compaction.max_context_length`), which expresses the 180 the spec always intended and is independent of the window derivation.

New coverage for the reworked rendering: the row is collapsed by default (`aria-expanded="false"`), expanding reveals the generated summary in the body, and collapsing removes it from the DOM again — plus the same collapsed default on the manual `/compact` path. The mock answers the compaction request with the same prose it uses for the reply, so the summary is counted (1 occurrence collapsed → 2 expanded) rather than matched by text alone.

Nothing was dropped as obsolete: the "row shows no token counts" check was re-expressed against the new row, and the file header was corrected — it claimed `reply -> compaction -> stats` while the assertions below it (and the shipped order) are `reply -> stats -> compaction`.

Ran this spec alone against a freshly built stack (own ports, processes cleaned up): **2 passed**. The rest of the e2e suite was not run.

PR #340, opened 2026-08-19
fix(web): split thinking-level labels by surface, drop the summarize notice

hiyouga:
Two display corrections requested by the maintainer, both scoped to `packages/web`.

## 1. The summarize compaction row drops its outcome notice

**Kept the `discard` variant, removed the `summarize` one.** `compactionDone` now returns `undefined` for `summarize`. `StepBanner` already guards its detail slot with `detail !== undefined` and lays the header out with a flex `gap-2`, so the slot disappears cleanly — no stray separator, bullet, parenthesis or double space where the text used to be.

What each row renders once settled:

| Mode | Before | After |
| --- | --- | --- |
| `summarize` | ✓ 压缩 · 已切换到摘要后的新上下文 · 1.4s · ⌄ | ✓ 压缩 · 1.4s · ⌄ |
| `discard` | ✓ 压缩 · 已丢弃旧上下文 · 1.4s | unchanged |

The `discard` notice stays because it is the one outcome the row cannot show any other way: that mode writes no summary, so it has no expandable body. A `summarize` row, by contrast, carries the adopted summary behind its chevron — the notice only restated the row.

Failure and running states are untouched.

## 2. Thinking-level labels split by surface

`S.chat.thinkingLevelNames` became the **plain** tier name (zh: 低 / 中 / 高 / 极高 / 最高, plus 无 for a stored legacy `none`). A new `S.chat.thinkingLevelMenuName(name, level)` composes the **annotated** dropdown-row variant.

| Surface | Variant | Change |
| --- | --- | --- |
| Composer picker trigger + tooltip — draft **and** active-session mounts (one `ThinkingLevelSelect`, mounted twice) | plain | via the table |
| Composer picker **dropdown rows** | annotated | call site updated |
| Mid-chat switch dialog body | plain | via `thinkingLevelLabel` |
| Mid-chat switch success toast | plain | via `thinkingLevelLabel` |
| Project chat-defaults `<select>` options | plain | see below |
| Project chat-defaults read-only row | plain | via the table |
| Agent settings runtime rows | wire value | deliberately unchanged, see below |

**English keeps no split.** Its tier name already *is* the wire value, so annotating would only repeat itself. Rather than duplicate a six-row table, `en`'s formatter takes just the name and returns it — TypeScript accepts the shorter signature against `Strings`, so the shared shape expresses "no split" without forcing anything awkward on `en`.

**Why the Project chat-defaults control stays plain.** It is a native `<select>`, which paints the picked option's own text onto the collapsed control. Annotating its rows would put `(xhigh)` straight back onto what is, once closed, a trigger — the exact thing being removed. The maintainer's phrasing is imperative for the trigger and permissive for the dropdown ("the dropdown *can* carry the annotation"), so the trigger requirement wins where one element serves both roles. It also matches the approval-mode select directly above it, whose options are plain localized names.

**Why the Agent settings rows were left alone.** They never read `thinkingLevelNames` — `thinkingLevelOptionsFor` sets both `triggerLabel` and `label` from the raw wire value, in both locales. That is coherent with the form around it: every field there is labelled with its literal config path (`model.thinking_level`, `model.max_tokens`, `max_context_length`) and shows values as stored. Localizing only this one value would make it the sole translated value in a raw-config editor, and it never exhibited the `极高 (xhigh)` noise the maintainer objected to. Happy to change it if you disagree.

## Drift guard

The guard described in the brief turned out to live in **`packages/cli`** (`thinking-command.test.ts`, over the CLI's printed messages) — it is unaffected by this change. The web side had no such guard: `thinking-level.test.ts` worked against a hand-written `NAMES` stub and never imported the real dictionaries, so a new tier could ship unlabelled.

Added one, following the `placeholders-parity.test.ts` precedent — it walks core's `DEFAULT_CHAT_THINKING_LEVELS` against both real catalogs and asserts:

- the picker offers exactly the tiers core accepts;
- every stored level (including `none`) is named in both locales;
- no zh name carries latin letters or parentheses — this is what pins the trigger requirement;
- the menu variant carries the wire value, and `en`'s adds nothing.

Verified it is not vacuous: restoring `xhigh: "极高 (xhigh)"` fails it with `zh xhigh must not carry the wire value`.

## Also updated

- `draft.spec.mjs` and `reloaded-followup.spec.mjs` asserted the old trigger text (`toContainText("中 (medium)")`). They now assert the plain name **and** `not.toContainText("medium")`, so a trigger that regains the annotation fails; the menu-row assertions (`低 (low)`, `高 (high)`, `最高 (max)`) are unchanged, since rows keep it. Not run — e2e is excluded from this round.
- Changelog entry pair `changelog/unreleased/2026-08-19-thinking-label-and-compaction-notice.{md,zh.md}`.

## Not changed

- **`packages/cli`'s own compaction line** (`[compaction] done; continuing with the summarized context`, `i18n.ts:531/800`). Different surface, different wording, and the CLI has no expandable summary body, so the text is not redundant there. Say the word if it should go too.
- **Docs.** No `packages/docs/content/*.md` quotes either the label format or the compaction notice; `web-app.*.md` describes the shared step-row anatomy generically, which still reads true.

## Verification

`packages/web` typecheck clean; 901 tests across 68 files pass. Prettier clean on all touched files. No other package touched.


hiyouga:
Applied the follow-up in `d0693a9b`: the discard row is no longer labelled compaction at all — the change moved up from the detail slot into the row's **title**.

`compactionTitle` now takes the mode. Settled rows, both locales:

| Mode | zh | en |
| --- | --- | --- |
| `summarize` | `✓ 压缩 · 1.4s · ⌄` | `✓ Compaction · 1.4s · ⌄` |
| `discard` | `✓ 清空 · 1.4s` | `✓ Clear · 1.4s` |

Picked **Clear** as the English counterpart to 清空: same short label register as `Compaction`, and it says what the step did rather than what it kept.

## Deleted

- `已丢弃旧上下文` and `old context discarded` from both catalogs.
- `compactionDone` itself. With both branches gone it had no caller and nothing to return, so it was removed rather than left as an empty function. `compactionFailed` is the only outcome string left, carrying the one thing a title cannot say.

## Also changed

- **The running row** stopped rendering `detail={item.mode}` — the raw, untranslated `summarize` / `discard`. The title carries the mode now, so that slot was duplicating it in wire form. A live row reads `压缩 · 3.2s ⌄` / `清空 · 1.1s`.
- **`e2e/compaction.spec.mjs:73`** still asserted `toContainText("已切换到摘要后的新上下文")` — the string removed in the previous commit. My sweep there only covered `.ts`/`.tsx`/`.md` and missed `.mjs`; caught it this round. It now asserts the succeeded row writes no outcome text. The `hasText: "压缩"` locators are unaffected: no e2e spec exercises discard mode.
- **Tests**: `compaction-summary.test.ts` gained locale-parity cases pinning summarize → 压缩 / Compaction, discard → 清空 / Clear, that a discard is never titled the same as a summarize in either locale, and that an unknown mode falls back to the compaction title.
- **Changelog pair** rewritten — it described only the detail-slot removal.

## Found but not changed — needs your call

**The Trace page still badges a compaction round "压缩" / "Compaction"** (`trace-file-view.tsx:483`, `S.traces.compactionRound`). It is a different string from a different namespace, and it cannot be made mode-aware here: the badge renders off `st?.compaction === true`, a **boolean** on the round stats, with no mode anywhere in that payload. Teaching it 清空 means adding the mode to the trace-index round stats — a server/core change, not a copy fix. Say the word and I'll open a separate PR.

Arguably it should stay as-is regardless: there it labels a *round type* ("this round isn't answering the user"), and it groups with the `压缩次数` count on the same page, which also counts both modes.

**Unrelated observation:** `e2e/compact-abort.spec.mjs` locates the banner with `getByText(/\[压缩\]/)` — a *bracketed* `[压缩]`. Nothing in `packages/web/src` renders brackets around that title, and nothing in this PR ever did, so that locator looks stale independently of these changes. Flagging it rather than touching it, since it is outside this PR's subject and I did not run e2e.

## Verification

`packages/web` typecheck clean; 904 tests across 68 files pass (3 new). Prettier clean on all touched files. No other package touched — `packages/cli`'s own `[压缩] 完成，已切换到摘要后的新上下文` line is still deliberately untouched, as noted in the PR description.

hiyouga:
Closed the Trace-page gap in `1ad092f6`. One correction to my earlier report first, because it changes what the fix actually was.

## The Trace page was never mislabelling a discard

A `discard` compaction **issues no model request** — core emits `compaction_begin` and `compaction_end` back to back (`context-engine.ts:1281-1288`, `discardContext`). The server only ever set `TraceTaskStats.compaction` inside its `request_begin` branch, so a discarded round never carried the flag and the badge never rendered for it.

So it was not saying 压缩 for a discard. It was showing a **bare, unlabelled round card with no badge and no stats** — which is arguably the worse failure, since nothing on the card explains what that round was. That is what this commit fixes.

## DTO shape

Additive, not a replacement:

```ts
compaction?: boolean;              // unchanged — still the sole gate
compactionMode?: CompactionMode;   // new, beside it
```

**An old client still reads it correctly.** `compaction` keeps its exact meaning and population, so a consumer that only knows the boolean sees the same rounds flagged as before and simply ignores the new field. An absent `compactionMode` — an old server, or a `compaction_begin` with no readable mode — is defined to read as `summarize`, which is what the badge always displayed. Replacing the boolean with a mode-bearing shape would have broken that, so I kept them side by side.

## Population

Flagged from the round's own `compaction_begin` in the message-attribution pass rather than from the compaction request, since a discard has no request. For a `summarize` this resolves to the same turn the request-side path already flagged — the pre-existing summarize test passes unchanged, which is the check that matters. Subagent messages are skipped, matching the main loop.

## Badge

`trace-file-view.tsx` now calls the chat row's `S.chat.compactionTitle(mode)` through a small exported `compactionBadgeLabel` helper, so the Trace page and the conversation cannot drift. `S.traces.compactionRound` had no other consumer and was deleted from both catalogs rather than left as a second copy of the word.

## Tests

- **Server** (`trace-service.test.ts`): the existing summarize round now also asserts `compactionMode === "summarize"`; a new discard fixture asserts the round is flagged, carries `"discard"`, and really did run no request (`startTs === ""`, `llmMs === 0`). 32/32 pass.
- **Web** (`trace-compaction-badge.test.ts`, new): the flag alone gates the badge (a mode without the flag renders nothing), summarize/discard read 压缩 / 清空 and Compaction / Clear, the label is literally `compactionTitle`'s output, and an absent mode falls back to the compaction title.

## Verification

`packages/server`: typecheck clean, 711 tests across 73 files pass. `packages/web`: typecheck clean, 908 tests across 69 files pass. Prettier clean on all touched files.

No docs change: `server-api.{en,zh}.md` lists the analysis endpoint as a single route row and never enumerates `TraceTaskStats` fields, so there was nothing to keep in sync.

Left the raw trace **event** rows alone as agreed — that view shows wire values next to `compaction_begin` deliberately.

PR #341, opened 2026-08-19
feat(web): restore the row's archive+delete hover pair and move the full menu to right-click

hiyouga:
## What & why

Maintainer request: put the Session row's hover menu back to the previously released **archive and delete**, make today's richer popup a **right-click** behavior instead, and give the **list options** (grouping mode and its siblings) icons.

## 1. The hover affordance — which vintage I matched, and how today's differed

I went looking for the released behavior rather than inventing one, and the history does not say quite what the request assumes — please correct me if I picked the wrong vintage.

**No released tag ever had an ellipsis menu on the row.** At every tag from v0.1.4 through v0.2.2 (the newest tag in the repo) the row carried three inline icon buttons, `opacity-0` → `group-hover:opacity-100`, under the comment `{/* Action group: rename + archive/unarchive + delete */}` — identical in all five tags:

| | v0.1.4 – v0.2.2 (released) | main today |
| --- | --- | --- |
| Affordance | 3 inline icon buttons on hover | 1 ellipsis button opening a dropdown |
| Actions | rename, archive/unarchive, delete | pin/unpin, rename, archive/unarchive, delete |

The ellipsis dropdown **and** the pin item both arrived after v0.2.2, together, in `c31cd3be` (*collapsible agents→evaluation nav group*, #277). The **archive + delete** pair you remember does exist in the history, but earlier than any tag: `design/specs/08-CHANGELOG.md` line 121 records it on 2026-07-08 — "Session 行操作组合并归档 + 删除" — and manual rename joined the row the next day (line 114, 2026-07-09), before v0.0.1 shipped.

**What I did**: restored the released *pattern* (direct icon buttons in the row's trailing slot, no dropdown) carrying the *pair you named* (archive/unarchive + delete). Rename is the one released action that is not in the hover pair — it moved to the right-click menu, which also keeps `design/specs/06-PROTOTYPE.md`'s "每条 Session 支持重命名、归档与删除" satisfied.

The slot now reserves the pair's width (`min-w-12`), so rows with no last-active time no longer let the icons overhang the title; the icons still form one column down the list. The last-active-time ↔ actions swap is unchanged in kind.

## 2. What moved to right-click

Right-clicking a row opens **pin/unpin · rename · archive/unarchive · delete** at the pointer — the whole popup, not just the two items the hover pair dropped. Pin is still gated on `canPin` (active-list rows only), exactly as the dropdown gated it.

There was **no context-menu or long-press prior art anywhere in the repo** (`grep -rn onContextMenu packages/` was empty), so rather than hand-roll a popup I taught the existing `Dropdown` primitive a **virtual anchor**:

- New optional `anchorRect` places the panel against a given viewport box instead of the container's own. Everything else is inherited, not re-implemented: dismissal via the shared esc-layer stack (Escape closes only the topmost layer), outside click, scroll; focus handling (opening focuses the first item, Escape hands focus back — via a new `returnFocus` hook, since a context menu has no trigger button of its own); and the existing edge clamping + flip, so the panel stays on screen at any viewport edge, sidebar collapsed or expanded, at phone width, in both themes.
- A pointer-anchored panel **dismisses** on scroll instead of re-placing itself — the point it was anchored to has moved and there is no trigger left to follow.
- Native-menu suppression is scoped by construction: `preventDefault()` lives inside the row's own `onContextMenu`, and nothing is bound at `document`/`window` level. Right-clicking anywhere else in the app is untouched. A test pins this against the source text (no global `contextmenu` listener in `packages/web/src`; exactly one file defines an `onContextMenu`).

## 3. Accessibility path — and why this one

A right-click-only menu strands keyboard and touch users, and this repo has already written that rule down: `design/specs/06-PROTOTYPE.md` requires touch to keep a non-hover path because 触屏没有 hover. Tailwind v4 scopes `hover:` behind `@media (hover: hover)`, so the *old* ellipsis was in fact already unreachable on a touch screen — this change closes that gap rather than widening it.

I did **not** rely on "keep the ellipsis as a second entry point", because the pared-back hover pair no longer contains pin or rename. Instead, three gestures open the same menu:

- **Mouse** — secondary click, anchored at the pointer.
- **Keyboard** — `Shift+F10` on the focused row, anchored to the row's own box rather than the viewport corner. Handled explicitly rather than left to the browser: Windows/Linux browsers synthesize a `contextmenu` event from that chord (and from the Menu key, which is also handled, since the same handler receives it), but macOS keyboards have no Menu key and Safari synthesizes nothing.
- **Touch** — press-and-hold (500ms; abandoned if the finger travels >10px so it never eats a scroll). Two follow-on hazards are handled: the click a touch screen replays when the finger lifts no longer opens the conversation, and the compatibility `mousedown` it also replays — which lands on the row, i.e. *outside* the panel — no longer dismisses the menu the gesture just opened (a 350ms settle window).

Archive and delete remain reachable by every route: hover/keyboard-focus as direct buttons, and inside the context menu.

## 4. List option icons

Hand-rolled inline SVG in the existing `GlyphIcon` vocabulary — same `viewBox="0 0 24 24"`, `stroke-width 1.7`, round caps, rendered through the shared `Icon` at the menus' 13px, `aria-hidden`:

| Row | Glyph |
| --- | --- |
| Group by workspace | folder (`FOLDER_ICON`) |
| Group by agent | agent (`AGENT_GROUP_ICON`) |
| Manual order | up/down reorder arrows (new `REORDER_ICON`) |
| Most recent | clock (new `CLOCK_ICON`) |

The two grouping glyphs are the ones the section header's grouping toggle *already* shows; both call sites now read one shared `GROUP_MODE_ICONS` map so a mode cannot wear different icons in the toggle and the menu. The two sort glyphs name what decides the order rather than decorating the row. Every row keeps its text label, so accessible names are unchanged.

## Design specs

**No spec is contradicted.** `design/specs/06-PROTOTYPE.md` names the action set (重命名、归档与删除) but never the surface — no 省略号 / 右键 / 上下文菜单 anywhere in `specs/` — and all three stay reachable. Session-level pin and the sort-mode menu are not specced at all. If you want the new surface written down, the additive wording would be, in the `- Session 列表：` bullet:

> 行尾悬停显示归档与删除；右键点击该行（触屏长按、键盘 Shift+F10）在指针处打开完整菜单：置顶、重命名、归档、删除。

I did not edit `design/` — it is the read-only shared symlink.

## Tests

New/extended, following the package's node-only convention (no jsdom; `title-reveal.test.ts`'s source-text precedent for what a DOM-less suite cannot otherwise reach):

- `packages/web/test/session-row-menu.test.ts` — the hover set is exactly `["archive","delete"]` and contains neither rename nor pin; the context set is the full four, drops only pin when `canPin` is false, and is a superset of the hover pair; label/glyph flips for archived and pinned; distinct glyphs per action; both locales name every action (the hover buttons are icon-only, so the label *is* their accessible name).
- `packages/web/test/context-menu.test.ts` — pointer vs keyboard anchoring (including the `(0,0)` coordinates some browsers report for the Menu key); `Shift+F10` recognition and that a bare F10 is ignored; touch/pen-only press-and-hold, its slop tolerance and settle window; and the suppression-scope scan described above.
- `packages/web/test/group-list.test.ts` — extended with the icon-per-option mapping: one glyph per grouping/sort option, all four distinct, every row still labelled in both locales.

Verification: `packages/web` 874 tests pass, `packages/docs` 53 pass, both typecheck clean, prettier clean on every touched file. E2E not run (shared box).

## Also

- `S.chat.sessionMenu` ("对话选项" / "Chat options") was the ellipsis trigger's only consumer and is removed from both catalogs.
- Docs: `packages/docs/content/web-app.{en,zh}.md` had no conversation-list section at all, so a new one documents both surfaces and all three gestures — a right-click gesture nobody wrote down is a gesture nobody finds.
- Changelog: `changelog/unreleased/2026-08-19-session-row-actions.{md,zh.md}`.


hiyouga:
## Self-review round — applied in `d95851bb`

I put the diff through a critical review before asking for yours. It found three real defects, all in the press-and-hold lifecycle, which was the one part of this change with **no test of its own** (the pure helpers were well covered; the hook composing them was covered only by grep). That lifecycle is now a pure reducer in `lib/context-menu.ts`, so the node-only suite can drive whole touch sequences.

Each fix is pinned by a test I verified fails against the old behavior (mutation-checked one at a time — each mutation failed exactly one test, no more):

1. **The grace window was measured from the wrong moment.** It started when the menu opened, but the compatibility mouse events a touch screen replays arrive when the finger **lifts**. So anyone who actually read the menu before lifting — longer than 350ms, i.e. the normal case — had it dismissed by their own finger. Now measured from the lift.
2. **Android's own mid-hold `contextmenu` was not recognized as a hold.** Chrome raises it and can beat our 500ms timer; that path never set the held flag, so the click the lift replayed fell through and opened the conversation on top of the menu. On the primary touch platform.
3. **Portal events bubbled back into the row.** React propagates through its own tree, and the Dropdown's body portal is a React child of the row — so right-clicking a menu item re-entered the row's handler and jumped the open menu under the cursor. The handlers now ignore anything failing a DOM containment check, which only portal content can fail.

Also applied:

- Menu items hand focus back to the row before acting. This mattered more than it looks: the context menu is the **only** keyboard route to pin and rename, so `Shift+F10 → Enter` was dropping the user onto `<body>`.
- A resize dismisses a pointer-anchored panel, for the same reason a scroll does.
- The hover buttons reveal on plain `focus` rather than `focus-visible`, so they agree with the time span they swap with instead of leaving the slot blank after a mouse click.
- **Docs corrected** — they claimed focusing the row reveals both buttons. It doesn't; tabbing onto either reveals that one. Both locales and both changelog files fixed.

Deliberately not changed:

- **Escape inside the grace window** still moves focus before a close the hook can veto. Reachable only on a touch device with a hardware keyboard, within 350ms of a lift; the fix would mean teaching `Dropdown` that a close can be refused, which is more surface than the case warrants.
- **Near the right edge the panel clamps rather than flipping to open leftward**, so it can cover the row it acts on. Pre-existing `Dropdown` behavior, shared with every other menu; changing it belongs in its own PR.

Verification: 883 web tests (up from 875), 53 docs tests, both typechecks, `vite build`, prettier — all clean.

PR #353, opened 2026-08-20
feat: introduce extension

Myriad-Dreamin:
An extension mechanism for the harness: one entry point, two typed events, and a deployment-owned list of which extensions to load.

An extension is a module exporting **one function** — `activate(ctx: ExtensionContext)` — run once per process at load time, and awaited when async. Everything after that is typed events:

- **`ctx.on("initialize", h)`** — `PenguinInterface`, the *definition* view, once per App creation: `iface.workflow` (a name → factory Map that refuses a duplicate name) and `iface.tool` (reserved).
- **`ctx.on("create", h)`** — `PenguinContext`, a harness *instance*: platform members flatten onto it (`context.terminals`), workflow instances live at `context.workflows`.
- **`ExtensionEvents`** — the one typed registry mapping each event name to its payload; adding an entry types the platform's emit and every handler together.
- **`ctx.disposables`** — cleanup for what `activate` allocates. Disposers may be async; they run concurrently and the graceful shutdown awaits them (≤5s), so they must be mutually independent — there is no teardown order.
- **`extensions.json`** — which extensions a deployment runs, resolved against the installation.

Four commits: the seam, then extensions-as-configuration (loader, startup step, SDK subpath), then the docs page for the boot order and the extension lifecycle, then the changelog entry.

## Where the contract lives

The contract is declared in the SDK, at `@prismshadow/penguin-core/extension`. `PenguinContext` and `PenguinInterface` are **open**: the harness contributes the members it owns — `terminals` — by augmenting that module, and re-exports both halves from `@prismshadow/penguin-server/extension` so an extension package has one import site. Core names nothing it cannot reach, this layer names what it owns, and an extension sees one `PenguinContext` either way.

Both subpaths emit types only — 33 bytes of source map and nothing else — so an extension package carries no runtime dependency on either and stays a self-contained library that happens to satisfy an interface.

## What a workflow is here

Registration plus a plain JS call — nothing more:

```ts
iface.workflow.set("name", () => ({ run: (input) => output }))   // definition view
context.workflows.run("name", input)                             // instance view
```

No Session, no approval, no streaming; a workflow does not reach agents. Instances are built once per App, after registration closes, so an extension always sees its own workflow instantiated in the App it registered into, and a stateful factory never carries the previous App's state across a swap. This slot is where fuller workflow support lands later; `feat/workflow-hmr` builds on it.

## Why it is shaped this way

**Extensions are configuration, not platform capability.** They are not compiled into the platform bundle and a hot push does not deliver one: `<root>/extensions.json` lists package specifiers and the runtime resolves each against the *installation*. That is why this lands before sandbox support — a sandbox backend is a package a deployment installs, and the harness must not depend on any of them.

Resolution is anchored at `process.argv[1]` rather than the module's own location, for the same reason the packaged bundle's own resolver is: a pushed bundle running from `hmr/store` has no `node_modules` of its own, so anchoring at the bundle would find nothing.

**Loading is the runtime's job; delivery is the platform's.** `extensions.json` is read once per process at startup step ④ — after the instance-lock pre-check, so a process about to exit with code 3 never imports third-party modules — and the resulting host is published into the resource registry. Each App then *claims* that host rather than importing one. This is load-bearing, not tidiness: a pushed bundle is compiled standalone, so a module-level singleton on the platform side would be that bundle's own empty host after the first push, and every configured extension would silently vanish. A runtime that published none yields an empty host — the honest reading of "this runtime knows nothing about extensions".

**Per-App redelivery is a structural fix.** Both events fire at *every* App creation — the packaged boot and every hot-swap boot. An extension's registrations therefore cannot end up "seeded in the old instance, empty in the new", which is exactly the bug class that bit workflows.

**The subscription window is `activate`, sealed once it settles.** Handlers run per App, so a subscription made inside one would accumulate one copy per hot swap — `on(...)` after `activate` settles throws instead, at the packaged boot, loudly. Disposables seal with the same window for the same reason. This is also why this shape beats an event emitter: subscription is data fixed at activation, so the double-subscribe-per-swap bug class is unrepresentable, and there is no on/off/once lifecycle surface nothing uses.

**Two failure classes, deliberately different.** A *per-entry* failure — an unresolvable specifier, a module with no `activate` export, a throwing or rejecting `activate` — is reported and skipped, and whatever that `activate` had already pushed into `ctx.disposables` runs before the entry is dropped. A *config-level* failure is fatal: an absent `extensions.json` means no extensions (the default deployment shape), but one that exists and cannot be read, or is malformed, fails the boot rather than presenting as a healthy server that silently dropped every capability the config asked for.

**Event handlers stay synchronous.** An App is assembled synchronously around the emit, so a handler that returns a promise could not be awaited and its rejection would escape unhandled — possibly killing a process that has not installed its handlers yet. Returning a thenable is refused at delivery, and the refusal claims the promise so the rejection cannot escape either way. Async work belongs in `activate`, which is awaited.

## Deliberately not done

`iface.tool` lands as an honest placeholder — an empty factory map. Members grow when a concrete need names them. The open decision it reserves is written where it will be made: whether extension tools reach agent Environments, which would need a platform→core seam along the `proxyEnv` precedent.

`"initialize"` and `"create"` are the only events so far; the registry is an open set.

## Verification

`pnpm build`, `pnpm typecheck` (0 errors), `pnpm format:check`, and the affected package tests: server 87 files / 925 tests, docs 53.

New coverage is 25 cases. Sixteen on the seam — activate-once with typed delivery in activation order, an awaited async `activate`, the sealed window, a failed `activate` rolling back what it had registered, an async handler refused without leaking its rejection, concurrent disposal with failures isolated and idempotent, a refused duplicate workflow name, per-App redelivery across the two boots a swap performs, workflow registration/instantiation/calling, and the empty-host fallback. Nine on the loader — missing config, malformed config, a config that exists but cannot be read, per-entry failure isolation, a module with no `activate` export, and an import-time throw.

## Not in this PR

Sandbox support builds on this seam and is #354, based on this branch.


packages/docs/content/skills.en.md
```diff
@@ -72,7 +72,7 @@ The built-in Skills, by group (the group manifest is `SKILL_GROUPS` in `packages
 | | `vllm` | Deploy and serve LLMs with vLLM behind an OpenAI-compatible endpoint, with tool calling enabled for agent workloads |
 | | `ollama` | Deploy and serve local models with Ollama: pull and run them, then expose the OpenAI-compatible endpoint to apps and agents |
 | | `llamafactory` | Fine-tune LLMs with LlamaFactory: register datasets, train via YAML configs, merge LoRA adapters and serve the result |
-| | `skill-porting` | Port skills from external sources — plugin marketplaces, skills.sh registries, GitHub repos or local folders — into the agent after review and normalization |
+| | `skill-porting` | Port skills from external sources — extension marketplaces, skills.sh registries, GitHub repos or local folders — into the agent after review and normalization |
```
hiyouga: 这个和 Skill 的 plugin 不是一个东西，不应该改这里

hiyouga (approved):

PR #354, opened 2026-08-20
feat: add confinement interface, with backends as plugins

Myriad-Dreamin:
> **Stack:** `main` ← **this** ← #543 ← #546 ← #656 ← #383 ← #526 ← #551 ← #544 ← #448 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #627 ← #628 ← #630 ← #634. Review the diff against `main`.

Shell commands an agent runs can now be confined. The plugin mechanism this rides on (#353) has merged, so this sits directly on `main`.

## What this adds

1. **A spawn seam in core** — `confineSpawn`, threaded along the existing `proxyEnv` line (`CreateAgentOptions → Agent → Environment → CommandSessionManager`) and applied in `ManagedSession` right before spawn. It receives the exact argv (shell included) plus `cwd`/`workspaceDir` and returns the argv to spawn instead. Mechanism only: it never interprets the argv.
2. **The harness's own sandbox interface and service** (`packages/server/src/sandbox/`) — the vocabulary, and routing from a policy to a backend that implements it.
3. **Four backend plugin packages** — `plugins/sandbox-{dsh,bwrap,seatbelt,mxc}`, in the top-level `plugins/` directory that #542 already gave the plugin library (the workspace's existing `plugins/*` glob reaches them).
4. **A closed plugin contract** that owns the sandbox vocabulary (see below).
5. **The rename**: what #353 shipped as the *extension* mechanism is the **plugin** mechanism from here on — config file, SDK subpaths, type names, package names and directory (see "The rename", below).

## Why it is shaped this way

**The interface is ours; DSH is one adaptor behind it.** The design is modeled on the DeepSeek Harness sandbox (argv rewriting, honest enforcement reporting, fail-closed) but owned here, which is what lets it name dimensions DSH's vocabulary explicitly excludes:

| dimension | who implements it |
| --- | --- |
| `fs-write` | every confining backend |
| `network` | optional |
| `mask-paths` | optional |

A backend *declares* the subset it implements — declaring nothing means filesystem-only, which is exactly what a stock DSH backend is — and the service **routes** each policy to a backend that implements what it requires. A filesystem-only policy still takes the portable adaptor on every OS; a policy needing no-network or hidden paths takes a backend that implements them; a requirement nothing covers **fails closed**, naming what each mounted backend does cover. An unimplemented dimension can never be silently ignored.

**Fail-closed is the contract, not a mode.** `confine` must return enforcing argv or throw. A throw aborts the spawn and surfaces as the command's spawn error — the command never runs unconfined. This holds end to end: a broken provider makes `exec_command` return `failed` with the reason, and the file the command would have written does not appear.

**Backends are packages, not platform code.** Each compiles against the type-only plugin SDK and carries no runtime dependency on the harness. The DSH dependencies (cordis, `dsh-sandbox*`, and through them koffi and the landlock-run launcher) live inside the DSH package — the harness depends on none of them. A consequence worth noting: with no backend in the platform bundle, the bundle no longer inlines any native-adjacent module, and `deploy.mjs`'s runtime-resolved-natives shim is deleted.

They live under the top-level `plugins/`, beside the plugin library's data packages, and nothing else in this repository ships them — a deployment installs them by name. Nothing under `packages/` depends on one, the platform bundle carries none, and all four are `private`. The two kinds of package do not meet at runtime: the library loader enumerates core's `@penguinharness/` dependencies only, and these are `@prismshadow/penguin-plugin-*`. `plugins/README.md` now has a "Runtime plugins" section stating the distinction, since its opening line ("data, not code — nothing to build") is true of the library alone.

**Settings survive hot swaps.** They ride the parked platform context, so a push cannot silently un-confine a deployment; they are omitted while pristine, so a default deployment's document stays compatible with sandbox-ignorant bundles.

## The rename: extension → plugin

One word now covers the whole mechanism, so a reader meets the same term wherever the concept appears.

| was | is |
| --- | --- |
| `<root>/extensions.json`, key `extensions` | `<root>/plugins.json`, key `plugins` |
| `@prismshadow/penguin-core/extension` | `@prismshadow/penguin-core/plugin` |
| `@prismshadow/penguin-server/extension` | `@prismshadow/penguin-server/plugin` |
| `Extension` / `ExtensionContext` / `ExtensionEvents` | `Plugin` / `PluginContext` / `PluginEvents` |
| `ExtensionHost`, `extensionHostFrom` | `PluginHost`, `pluginHostFrom` |
| `extensions/sandbox-*` | `plugins/sandbox-*` |
| `@prismshadow/penguin-extension-sandbox-*` | `@prismshadow/penguin-plugin-sandbox-*` |

`PenguinContext`, `PenguinInterface`, `HarnessContext` and the entire sandbox vocabulary keep their names.

**`extensions.json` on disk is a clean break.** It is not read and not migrated. A deployment that configured backends under the old name keeps running with none of them loaded — for the sandbox that means commands spawn unconfined. Rename the file and its key; the specifiers inside change only where they name this repo's backends. The sandbox entry's `Breaking` line names it and its `## Compatibility` section carries the migration — one entry for this PR, not two.

**Two collisions the word creates, resolved at their sites.** The DSH adaptor mounts DSH's chain through cordis's own `Context.plugin` — the two vocabularies share nothing but the word, and the local is named `cordisCtx` to say so. And the shipped `skill-porting` skill opened with "Penguin has no plugin mechanism and needs none", which this makes false; it now states what a Penguin plugin is before distinguishing it from the ecosystem plugins that skill ports from.

**What was deliberately not renamed:** file extensions throughout, micromark/remark syntax extensions, the composer's "+" extension menu, the docs' SDK "Extension seams" (interface extension points, a different concept), and the frozen `changelog/0.2.8/` entry that shipped the mechanism under its former name — the unreleased entry links to it as history.

## The contract is closed now — the one breaking change

`PenguinContext` and `PenguinInterface` named nothing and were reopened by each embedder through `declare module`. Any augmentation anywhere in a program became part of the contract everywhere, so a plugin type-checking against "the contract" was really compiling against members only one embedder happens to supply. They name every member now, and the server declares `HarnessContext extends PenguinContext` for what is its own.

The sandbox vocabulary moved with it: `SandboxPolicy`, `SandboxProvider`, `SandboxProviderRegistry`, `ConfinedArgv` and the mode/dimension unions are in `@prismshadow/penguin-core/plugin` (`plugin/sandbox.ts`), because a backend is written against exactly those names. The helpers that *read* them — `SANDBOX_DIMENSIONS`, `providerDimensions`, `requestedDimensions` — stayed in the server package, which keeps both plugin subpaths emitting types and nothing else.

**What breaks:** `terminals` is a harness class and cannot be in the contract. A plugin reading `context.terminals` now writes `ctx as HarnessContext` — stating at the use site that it depends on this embedder. The emitted value still satisfies the contract, and this change alters no wire or on-disk format (the config file's rename is the section above).

## Platform coverage

| plugin | platform | fs-write | network | mask-paths |
| `penguin-plugin-sandbox-bwrap` | Linux | yes | yes | yes |
| `penguin-plugin-sandbox-seatbelt` | macOS | yes | yes | yes |
| `penguin-plugin-sandbox-mxc` | Windows | yes | yes | yes |
| `penguin-plugin-sandbox-dsh` | all three | yes | — | — |

Each of the four ships its own `README.md` — listed in `files`, so it travels with the package — documenting what that backend actually enforces: bwrap's mount ordering and unprivileged-userns requirement, Seatbelt's last-rule-wins SBPL and path canonicalization, MXC's Windows-only mapping and its optional peer SDK, the DSH adaptor's platform chain. They also declare `Apache-2.0`, which they had not while presenting themselves as installable. #383's marketplace inlines those same files rather than keeping a copy.

Windows goes through Microsoft MXC (`processcontainer`), which is the one Windows mechanism expressing all three dimensions — a search of the DSH ecosystem turned up no third-party Windows backend beyond DSH's own restricted-token one, which is file-writes-only and already reachable through the adaptor.

## Verification, and what is NOT verified

`pnpm install`, `pnpm build`, `pnpm typecheck` (0 errors) and `pnpm format:check` clean; core **1024**, server **1507**, web **1533**, cli **384**, docs **53**, landing **69**, skills **27**, plus the four plugin packages (**39 passing, 11 host-gated skips**). One local-only failure, not from this diff: `packages/desktop` `test/cli-link.test.ts > repairs a link left dangling by a moved or updated app` fails identically on a clean `main` checkout of this machine and passes in CI's `test (rest)` on all three platforms; the diff touches nothing under `packages/desktop`. **CI is green on all 21 checks** across Linux, macOS and Windows. After the rebase onto #576, the `style` job additionally runs oxlint (#600); it flagged three files importing the same module twice, merged in the last commit, and the server suite (**1754**) passes with it. One transient failure is worth recording for anyone reading the run history: the first push failed `runtime (windows-latest, --win)` at preflight with `Downloading Electron binary... TypeError: fetch failed`. Moving the four packages rotated the lockfile hash and with it setup-node's pnpm store cache key, so that runner had to fetch Electron's binary instead of reusing a warm store, and the fetch failed. `pnpm install --frozen-lockfile` itself succeeded there, and the lockfile diff is only the four importer keys — `settings:`, `overrides:` and every resolved dependency are byte-identical — so nothing in the diff changes which build scripts run. It passed on the next run.

The typecheck was confirmed by **injection**, not by absence of errors: five deliberate type violations across the package boundary were each reported (e.g. `Type 'SandboxPolicy' is not assignable to type 'number'`), which is what proves cross-package resolution is real rather than silently degraded. The rename re-ran the same chain: the four plugin packages resolve `@prismshadow/penguin-core/plugin` from their new depth, and both renamed subpaths emit into `dist/plugin/`.

**Verified against real kernels on Linux:** writes confined to the workspace with outside writes denied (background children included), an empty network namespace (only `lo` in the confined process's `/proc/net/dev`, nothing resolves), a masked directory reading empty with its secret unreachable, and all three at once. The DSH adaptor's live suite passes here too.

**Verified by CI on macOS:** the Seatbelt live suite runs and passes on the macOS runner, so that enforcement path is covered by the same real-kernel standard as Linux.

**Not verified: Windows live enforcement.** The MXC path has only its config generation unit-tested, asserted against the *installed* SDK so schema drift fails in CI rather than on a Windows box. The DSH adaptor's live suite is `skipIf(win32)` — not a capability gate: CI shows the ACL chain loads there and the kernel denial fires (UTF-16 "Access is denied" from the runner). What is missing is a cmd-dialect probe set with UTF-16-tolerant denial matching, marked `TODO(win32)` at the skip.

## Risks worth a reviewer's attention

- **MXC is Public Preview**; its schema may move before 1.0. The coupling surface is stated on the module (the runner path and the `--config-base64` contract), and both facts were verified against the real 0.7.0 package rather than assumed.
- **`--unshare-net` leaves `/sys` readable.** Network *access* is enforced, but `/sys/class/net` still lists the host's interfaces through the read-only bind of `/` — host network topology remains readable as stale metadata. Documented on the provider; mask it explicitly if a deployment cares.
- **Windows has no live-enforcement coverage** (see above): the mechanism is demonstrably working in CI, but nothing asserts the confined behavior end to end there.
- **Default deployments have no backend.** With no `plugins.json`, nothing is confined — the direct consequence of "plugins are configuration". The changelog says so.
- **The rename lands on a stack.** #543 (`feat/module-tree`) and #383 (`feat/plugin-list`) are based on this branch and will conflict on rebase; #383 already uses the `plugin` word, #526 (`feat/extension-languages`) does not yet.

## Not in this PR

Stderr classification (denial vs runner failure), the escalation flow, and any UI for the config surface.


hiyouga (changes requested):
可能需要rebase一下最新的plugins

Myriad-Dreamin:
Rebased onto the current `main` (b5b21977, #576) on 2026-09-05; the branch now contains #542's `plugins/` and the four backends sit beside it. The oxlint failure the rebase picked up from #600 is fixed in 5e63dd15, and `plugins/README.md` says which packages there are runtime plugins. Ready for another look.

hiyouga (approved):

PR #358, opened 2026-08-20
refactor(server): hot reload the entire server

Myriad-Dreamin:
Moves the whole business surface — every service and every route — out of the runtime and onto the hot-swappable platform, so a push replaces it wholesale. The runtime keeps only what `hmr/README.md` calls mechanism: transport, auth, the HMR channel, the resource registry, static hosting.

Stacked below `clarify-task` (#285) and `feat/plugin-mvp` (#353); base is `feat/terminals`, which names the terminal block this work sits on.

## What moves, and why

`bootAppDeps` builds the runtime core, publishes its capabilities into the resource registry, and boots the platform; `platformImpl.create` then claims those capabilities and assembles the business graph plus **one** Hono app carrying every platform route — the terminal group beside the business groups. Each swap publishes `{deps, app, shutdown}` as a single registry write (`platform:current`), so the deps pointer and the route table riding on it flip together and no reader can observe a half-swapped pair.

Swap semantics for state that is not parked are a **hard stop**: pending approvals deny, active runs abort, the scheduler dies with its App, and the next App rebuilds from the claimed capabilities. Only resources that implement park/adopt ride across.

## Agreeing on live objects

A parked context document is strict-parsed by the kernel. A live resource cannot be — a claim is a cast — so the two sides of a seam agree through a descriptor instead: a `family` plus, per interface name, the **member set** a consumer depends on.

It is structural rather than a version number for the same reason the kernel's iface is (see `boot()`'s method-set check). A number says nothing about what is actually there, has to be remembered and bumped by hand, and every bump is global — widening `db` would decline a bundle that only ever touches `terminal`. A member set narrows that automatically, and is checkable from all three directions:

- **against the type, at compile time** — `MembersOf<T>` is `readonly (keyof T & string)[]`, so a renamed or removed member fails the build at the declaration rather than drifting into a descriptor that describes a shape nothing has;
- **against the live object, at claim time** — `lacksMembers` runs the same check `boot()` runs on an impl, so an honest-but-wrong runtime is caught at the claim rather than at the first call site;
- **between the two sides** — Go semantics, `required ⊆ offered`: the offering side may carry more, never less, so adding a member cannot decline a consumer that never named it.

A differing `family` short-circuits all of it: the names belong to another vocabulary, so a platform that does not want to inherit penguin's interfaces changes its family rather than arguing with each entry. Signature changes stay out of reach of both checks — the wire form is strings — and that is the documented limit.

The same descriptor drives resource reconciliation across a swap: each App's `create()` leaves its declaration in the registry and its successor integrates a group only when that declaration offers every member this build names, hard-stopping the rest in reverse registration order before it adopts anything. This rides the registry, not the kernel iface, so the kernel's park/validate/swap mechanism is untouched and the reconciliation policy itself ships by push. The one kernel change is +8 lines on `Resources`: an optional, ordered `disposeGroup`.

## Layout

`src/platform/` is dissolved: its HMR machinery joins `src/hmr/`, its terminal tree joins `src/terminal/`, and the business builders return to `app.ts` beside the runtime shell they were split from. That makes `app.ts ↔ hmr/platform.ts` a module cycle, which is safe here and documented at both ends: both are hoisted function declarations, called only at App creation, long after either module body has been evaluated.

## Verification

Per-commit green. At the branch tip: repo-wide `typecheck` + `format:check`, server 718, web 730, cli 245, desktop 49.

The compile-time constraint was checked by breaking it three ways on purpose — a typo'd `AuthService` member, a renamed `ServerConfig` field, a `TerminalSession` member that does not exist — each of which fails `tsc` at the declaration.

Beyond the suites, the load-bearing properties were exercised against a real server built from `dist/`:

- a real pty opened before a real `POST /api/hmr/upgrade` survives the swap, while the business surface is rebuilt and the web dist swaps atomically;
- a bundle compiled by the real `deploy.mjs` options serves the business routes after the push, proving the surface travels with the platform version rather than this build;
- the capability handshake accepts the real published objects (no declines in the log), and graceful shutdown drains the manager through the pointer.


Myriad-Dreamin:
Superseded by #359: the branch is now `feat/app-hmr`, stacked on `clarify-task` so the terminal feature stays contiguous. `feat/terminals` — which only ever named a cut through clarify-task's own history — is gone.

Issue #352, opened 2026-08-19
opus 5 is missing in the model catalog


hiyouga:
https://platform.claude.com/docs/en/about-claude/pricing

PR #360, opened 2026-08-20
fix(core,cli,web): upgrade AgentHub to 0.4.5 for unrecognized stream output

hiyouga:
## What this fixes

A relay in front of a Responses-compatible API injected `{"type":"ping","cost":"@"}` into the SSE stream. `ping` is the Anthropic Messages heartbeat spelling, so the Responses clients' unknown-event guard raised `Unknown output: {"type":"ping","cost":"@"}`; `context_engine` classifies that throw as a retryable request failure, and each of the five retries met the next heartbeat. The turn ended as:

```
llm request failed after 5 retries: Unknown output: {"type":"ping","cost":"@"}
```

[#286](https://github.com/Prism-Shadow/penguin-harness/issues/286) and the 0.4.3 bump closed the same class of failure for the spelling each family had already seen — `keepalive` on Responses, `ping` on Messages — but either spelling crossing into the other protocol still threw.

## The upgrade

`@prismshadow/agenthub` 0.4.4 → 0.4.5 (agenthub [#181](https://github.com/Prism-Shadow/agenthub/pull/181)):

- Every streaming client marks an unrecognized event `unused`, and the Gemini client skips an unrecognized content part. `AGENTHUB_DEBUG` set to anything other than `0`/`false`/`no`/`off` restores the guards with the same error text, for when you want to know what an endpoint is really sending.
- `AutoLLMClient` also gained `list_models()`, `default_headers`, and an `UnsupportedOperationError` class. The harness calls none of them yet.
- 0.4.5 aligned the `openai-responses` and `minimax-m3` message transforms with the GPT client, which flushes buffered assistant text at the end of a message rather than before each top-level item. A replayed turn of `[text, tool_call]` now reaches those two servers as `[function_call, message]` — the order the GPT client has always sent, and the only behavior change that touches what the harness puts on the wire.

## Verification

- The reported payload, through the **installed** 0.4.5 in this branch's `node_modules`: `openai_responses` → `unused`, `claude5` → `unused`; with `AGENTHUB_DEBUG=1` the Responses client raises `Unknown output: {"type":"ping","cost":"@"}` again.
- `pnpm -r build` clean; `pnpm -r typecheck` clean; `packages/core` → 839 passed, 5 skipped.
- Routing is untouched: no model catalog or protocol-path change, and `protocol-path.ts`'s verification note now names 0.4.5 as the version it was re-checked against.
- `pnpm-workspace.yaml`'s `minimumReleaseAgeExclude` names 0.4.5 and drops 0.4.4, as that file's own note instructed; the note above it now describes what the exemption protects today.

https://claude.ai/code/session_013STCF9fADwqT8jJH9C3g5h


hiyouga:
The conflict was the lockfile alone and it is resolved, so the PR is mergeable again and CI is green on the merged head (8/8). The bump itself is sound: the fix is real in the published package, the harness needs no new handling to absorb it (and now has a test saying so), and nothing crossed the AgentHub seam. Two things want a decision before merge — 0.4.5 reorders what the `openai-responses` and `minimax-m3` clients put on the wire when replaying history, which is a behaviour change travelling inside a fix release that nothing in this repo's tests exercises; and 0.4.5's supported-model registry is what makes a sentence #363 adds to a **shipped** skill stale, so these two PRs have a merge-order dependency.

## Fixed

- **`pnpm-lock.yaml` conflict** — `4bac905f`. Merged `origin/main` (moved on by #385, #381, #378), took main's lockfile whole, then re-ran `pnpm install` so pnpm produced the result rather than a hand-merge; `pnpm install --frozen-lockfile` passes afterwards. Only the lockfile conflicted — `pnpm-workspace.yaml`'s two edits (main's `node-pty: true` under `allowBuilds`, this branch's `minimumReleaseAgeExclude`) merged cleanly. One incidental delta against main: pnpm restored `engines: {node: '>=24'}` on the `packages/core` and `packages/skills` importer entries, which main's lockfile was missing. That is pnpm's own output, so I left it in.
- **Nothing tested the thing the bump changes** — `f4f3cbab`. 0.4.5's entire effect on this repo is that an unrecognized frame now reaches `EventTranslator` as an `unused` event carrying no content items instead of throwing, and no test said the translator ignores it. Added one case to `packages/core/test/llm.test.ts` feeding an `unused` event between two text deltas and asserting the message sequence, the concatenated text and the usage snapshot are unchanged.
- **Two claims in the changelog entry do not check out at HEAD** — `1ae99d7d`.
  - "Every streaming client now marks an unrecognized event `unused`": the four Chat Completions clients (`openai_chat`, `deepseek_v4`, `glm5_3`, `kimi_k3`) contain no `Unknown output` throw in either version and were untouched — only six clients had a guard to convert. Reworded to name the families that did.
  - `list_models()` / `default_headers` are the Python spellings; the package this repo installs declares `listModels()` and `defaultHeaders` (`dist/autoClient.d.ts`). Also replaced "The harness calls none of them yet" — a standing claim about the codebase, which `changelog/README.md` rules out — with what the bump did.

## For you to decide

- **Merge-order dependency with #363.** #363 adds this to `packages/skills/skills/agenthub-models/SKILL.md`, a file that installs into real users' agents: *"…and AgentHub 0.4.4 lists neither `claude-opus-5` nor `anthropic/claude-opus-5` although both route. For an id the registry omits, take the context window and price from the vendor's own page."* True against `main`; falsified by this branch — I diffed `dist/registry.js` and 0.4.5 adds both ids (56 → 59 entries). The sentence stays literally true because it names 0.4.4, but it reads as live advice. I did **not** touch the file: #363 owns it and my diff contains no skill file. Whoever lands second should apply this, dropping the version-pinned example so it survives the next bump too:

  > The registry is the curated current line-up, so prefer it when picking a model or estimating cost. It is not the routing table, and it lags in both directions: older ids in the table above (`gpt-5.4`, `claude-opus-4-7`, `gemini-3.1-pro-preview`, `gemini-3.1-flash-lite`) still route fine without appearing in it, and a newly launched id can route before the registry carries it. For an id the registry omits, take the context window and price from the vendor's own page.

  …plus a `version:` bump and a fresh `updated:` in the frontmatter (#363 already takes it to v13, so this is v14 if #363 lands first). #363's changelog entry carries the same clause; that one names 0.4.4 as a record of what was true when written, so I read it as fine to leave.
- **The replay-order change is the one wire-visible behaviour change, and it rides in a fix bump.** 0.4.5 rewrote `transformUniMessageToModelInput` in `openai_responses` and `minimax_m3`: the flush before each top-level item is gone, so buffered assistant text is emitted once at the end of the message (confirmed by diffing the published 0.4.4 and 0.4.5 tarballs). A replayed assistant turn of `[text, tool_call]` now leaves as `[function_call, message]`. The entry discloses it and carries no `Breaking` field, which reads right by this repo's definition — nothing on disk or in config breaks — but it is a contract with third-party relays, the same class of software that produced the bug this PR fixes. Options: take it as-is (that order is what the GPT client has always sent, so it is the best-exercised of the three); hold the bump until a Responses-protocol gateway has run a multi-turn tool-calling session end to end; or ask agenthub to land the transform refactor separately from the fix. I did not act because whether to take the upgrade at all is yours.
- **A genuine gateway error frame is now dropped silently.** The new branch is a catch-all: any event a client cannot name becomes `unused` with empty content, and none of the six clients' known-event lists mention `error` or `response.failed`, so a gateway's error frame goes the same way as its heartbeats — agenthub [#181](https://github.com/Prism-Shadow/agenthub/pull/181) says that is deliberate. Downstream here the stream then ends with nothing, AgentHub raises `EmptyResponseError`, `isMalformedJsonParseError` (`packages/core/src/llm/generative-model.ts:765`) labels it `malformed`, and the engine burns the retry ladder before the user sees a message that no longer carries the provider's text — where before 0.4.5 the payload at least appeared inside `Unknown output: …`. `AGENTHUB_DEBUG` is the escape hatch, and outside the changelog entry it is documented nowhere in this repo; `packages/docs/content/configuration.en.md:8` holds the environment-variable table it would go in. Options: accept the trade (heartbeats are far more common than error frames); document the switch there in both languages; or ask agenthub to keep named error frames loud.
- **The registry now disagrees with a deliberate catalog decision.** Every OpenRouter row in `dist/registry.js` that read `client: "openai-chat"` now reads `"openai-responses"`. `packages/core/src/state/model-catalog.ts:265-272` deliberately keeps the non-OpenAI OpenRouter rows on `openai-chat`, because the Responses client leans on OpenAI-specific reasoning-item round-tripping that is only guaranteed when the upstream really is OpenAI. Nothing enforces the registry's column here (the catalog imports nothing from agenthub), so this is advice drift, not a defect — but the reasoning in that comment is now contradicted by upstream and is worth re-checking on its own rather than inside a bump PR.
- **The shipped `agenthub-models` skill is also behind the library in a second way.** Beyond the registry sentence above, it documents nothing about `listModels()`, `defaultHeaders`, `UnsupportedOperationError` or the unknown-event policy. Nothing in it became false, so this is a scope call: updating it means another version bump plus the docs skills-sync test, in a PR that otherwise touches no skills.

## Checked and clean

- **Version consistency.** `@prismshadow/agenthub` is declared in exactly two manifests — `packages/cli/package.json:27` and `packages/core/package.json:49` — both `^0.4.5`; nothing else in the tree (root, `server`, `web`, `desktop`) references it. After the resolution the lockfile's two importer entries and all three peer-variant snapshot keys resolve 0.4.5.
- **The fix is real, and the absent code change is the correct absence.** Diffing the published tarballs: six clients carried an unknown-event guard (`claude5`, `ant_messages`, `openai_responses`, `gpt5_6`, `minimax_m3` on events, `gemini3_7` on content parts) and all six now fall through to `unused` / skip unless `isDebugEnabled()`. Every one of those branches returns `content_items: []`, `usage_metadata: null`, `finish_reason: null`, so `EventTranslator.pushEvent` (`packages/core/src/llm/generative-model.ts:335`) yields nothing for it — which is why there is no handling change to make. `cli` and `web` have no stream-event path at all.
- **Registry pass (56 → 59 entries), and it costs the catalog nothing.** `dist/registry.js` gained `claude-opus-5`, `anthropic/claude-opus-5` and `z-ai/glm-5.3`; renamed the direct `gpt-5.6` entry to `gpt-5.6-sol` (the bare alias still routes there); gave the direct GLM-5.3 row the price it previously omitted; and moved prices on `openai/gpt-5.6-{sol,terra,luna}` and `z-ai/glm-5.2`. None of it reaches this repo at runtime — `model-catalog.ts` imports nothing from agenthub and never calls `listSupportedModels()` — and none of it makes a catalog value wrong: the 2026-08-18 re-read already carries the registry's new numbers for `openai/gpt-5.6-sol` (cache_read 0.25 / output 15), `terra` (0.2 / 12), `luna` (0.02 / 1.2) and direct `glm-5.3` (0.26 / 1.4 / 4.4); `z-ai/glm-5.2` differs only by the documented promotion policy (OpenRouter rows store what OpenRouter bills, not list price); and `z-ai/glm-5.3` and `anthropic/claude-opus-5` are already rows. The direct `claude-opus-5` gap is issue #352, owned by #363.
- **Type compatibility.** `types.d.ts` is byte-identical between the two versions, so `EventType` (`"start" | "delta" | "stop" | "unused"`) neither narrowed nor widened. The additions are `UnsupportedOperationError` on the index plus `listModels()` / `defaultHeaders` on the clients. `LLMClient` did gain an **abstract** `listModels()`, which would break a subclass — but it is not on the index and the package's `exports` map allows only `.`, `./integration/tracer` and `./integration/playground`, so no deep import can reach it. `pnpm -r typecheck` clean across all 8 projects.
- **`protocol-path.ts`'s re-verification note holds at 0.4.5.** No base URL or endpoint construction changed. The one constructor worth checking is Gemini's, where `httpOptions` went from `url ? { baseUrl: url } : undefined` to an always-present object: `@google/genai@1.52.0`'s `patchHttpOptions` (`dist/index.cjs:7860`) iterates the patch's own entries, so an empty object is a no-op over the defaults, and `getDefaultHeaders()` (`:8061`) returns a plain record that survives its JSON deep copy. Same request URL either way.
- **The seam.** No `resolveModelEnv` or client-selection change, and `protocolPathForModel` still keys off `(provider, clientType)` only.
- **Changelog.** Metadata order is Date / Type / Scope / PR; no `Issue` field is owed — #286 is the earlier `keepalive` report, closed by the 0.4.3 bump, and no open issue tracks this one. No `## Why` / `## Verification` / `## Risks`; the pair mirrors section for section (`## Details` / `## 细节`, four bullets each); the cross-repo reference names its repo.
- **No design change owed.** `design/specs/08-CHANGELOG.md` stopped recording agenthub bumps after 0.4.1 — 0.4.2, 0.4.3 and 0.4.4 have no entry — and no spec pins the stream-event contract.
- **No new user-facing string**, so nothing is owed in `strings-en.ts` / `strings.ts` / `packages/cli/src/i18n.ts`.
- **Ran here:** `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm -r typecheck`, `core` 840 passed / 5 skipped, `cli` 292 passed, `web` 1014 passed, `pnpm format` → `format:check` clean, `git diff --check` clean. CI on the merged head: all 8 checks green.

PR #362, opened 2026-08-20
feat(web): pack same-name non-overlapping tool calls into shared timeline rows

hiyouga:
## What this changes

The trace observation timeline rendered every tool-call execution in its own row, so a long serial sequence of calls to one tool (e.g. ten `exec_command` runs) cost ten rows. Tool rows are now packed **per tool name**:

- Same-name calls whose time ranges don't overlap share one row; ten serial `exec_command` calls now take one row instead of ten.
- A different tool name, or same-name calls that overlap in time (parallel execution), still open extra rows — greedy first-fit, so an overlapping burst uses exactly as many rows as its maximum concurrency.
- Touching endpoints (one call ends exactly when the next starts) don't count as overlap; a still-running call blocks its row through the task end, so later same-name calls never draw over it.
- Rows of one name stay adjacent and names are ordered by their earliest call; the row label still shows the tool name once.

The packing lives in a new pure module `packages/web/src/features/traces/lane-packing.ts`; `timeline-chart.tsx` builds its tool rows from it (replacing the write-only `toolNames` field). Bar rendering, hover/click highlighting, event-list linkage, legend fade-out, and zoom/pan behavior are unchanged — only the row assignment changed.

## Verification

- `pnpm --filter @prismshadow/penguin-web test` — 76 files, 1000 tests passed, including 11 new `lane-packing.test.ts` cases (sequential sharing, touching endpoints, zero-duration spans, overlap/nesting splits, name isolation, lane ordering, first-fit reuse, unsorted input, open spans).


hiyouga:
The packing algorithm itself is correct — I hammered it with 400 randomised cases and it never overlapped, dropped or duplicated a span, never mixed names in a row, and (when no span is zero-length) always uses exactly as many rows as the name's maximum concurrency, so greedy first-fit is optimal here rather than merely adequate. It runs once per data change inside `useMemo`, not per render. I pushed one real fix (the packing footprint did not cover everything the row actually draws), two follow-ons from the same read, and a Chinese-punctuation correction. **Safe to merge after you have looked at the first "decide" item** — the min-width floor means a dense burst of short calls, which is exactly the case this PR optimises for, now draws its bars on top of each other in the shared row. That is a density trade-off rather than an algorithm bug, so I did not touch it.

## Fixed

All in 59f79f84.

- **`timeline-chart.tsx:239` — the footprint did not cover what the bar draws.** Packing claimed `callMs → outputMs`, but the row draws an approval-wait bar from the call to the decision *and* an execution bar starting at the decision (`execStart = s.approvalMs ?? s.callMs`, line 576). If a clock step records a decision before its own call, the execution bar starts left of the claimed footprint; if it records an output before its own decision, the approval-wait bar runs past the claimed end. Either way the next same-name call packs into the same row and draws over it. Simulating the first shape (`call=200, approval=150, output=400` after a call ending at 200) gave one row with the two bars overlapping across 30px; it is now two rows. New `toolSpanBounds` in `lane-packing.ts:28` returns the union of both bars and never an end before its start; 7 tests cover it.
- **`timeline-chart.tsx:340` — the bar index no longer walked bars in render order.** `firstBarKeyByTs` documents itself as "timestamp → the **first** bar's key (in render order)", and it decides which single bar lights up when the event list hovers a row carrying only a timestamp. It iterated the flat `g.spans` list, which *was* render order before this PR and is not after packing reorders the rows. It now walks `g.toolRows`.
- **`TaskGroup.spans` removed** (`timeline-chart.tsx:136`). Once the index above stopped reading it, the field was write-only — the same shape as the `toolNames` field this branch removed. `buildGroups` still keeps the list as a local.
- **`changelog/…-trace-lane-packing.zh.md` used halfwidth punctuation throughout** (25 halfwidth, 0 fullwidth). It is the only Chinese entry in `changelog/` with that ratio; every other one is predominantly fullwidth. Converted, and mirrored the new footprint bullet into both files.

One shape note on the first fix: I exported `toolSpanBounds` from `lane-packing.ts` rather than leaving the `min`/`max` inline in `buildGroups`, because the web vitest config is `environment: "node"` with "no DOM tests", so a rule living inside `buildGroups` cannot be tested at all. If you would rather keep it inline, the revert is mechanical — the seven tests go with it.

## For you to decide

- **The `min-w-[2px]` floor makes a packed row overdraw itself, and this is the PR's own target case.** Every bar carries `min-w-[2px]` (`timeline-chart.tsx:633`), so any call shorter than ~2px of track renders as a fixed 2px stub. Twelve serial `exec_command` calls of 40ms each, 50ms apart, inside a 60s task on a 600px track: all twelve pack into one row (correct — they do not overlap in time), and 30 pairs of bars physically overlap, each earlier stub keeping about 0.5px exposed. Before this PR those twelve calls had twelve rows and twelve full hover targets. The sharpest form is already pinned by a test: `lane-packing.test.ts` "shares a lane with zero-duration spans" puts `[5,5]` and `[5,9]` in one row, and since both are `absolute` with no z-index the later sibling paints the zero-duration bar away entirely. Mitigation that already exists: hovering the *event list* still highlights the buried bar by key, and clicking the event row still jumps — it is the timeline→list direction that degrades. Options as I see them: (a) accept it, density is the point; (b) require a minimum gap in the packer, e.g. `total/300` ≈ 2px at 1× zoom, which costs extra rows at high zoom where they are not needed; (c) keep packing purely temporal and enforce separation at render, which needs the measured track width; (d) fix only the exact-same-instant case, which is the one where a bar disappears completely. Picking between these is a visual-density call, so I left it.
- **No design-repo counterpart.** `design/specs/08-CHANGELOG.md:125` (2026-07-08) records the timeline rework as "每个工具独立成行（不再合并「工具1/2」，展示异步重叠）". Read literally that is *per tool*, which is what this PR implements — the pre-PR code was per *call* — and the normative line, `design/specs/06-PROTOTYPE.md:144`, only says "执行时间线按泳道展示各类区段", so nothing in the specs contradicts the change. But `CLAUDE.md` asks design and code to move together and this branch ships no design-side entry. Whether that repo owes a dated entry is yours; I did not write to `design/`, it is a shared read-only checkout here.

## Checked and clean

- **Randomised properties, 400 seeds**, up to 40 spans over 3 names, durations 0–5: no two spans in a row overlap under half-open semantics, no span dropped or duplicated, one name per row, same-name rows adjacent, names ordered by earliest start, the input array not mutated. With no zero-length spans the row count per name equals that name's maximum concurrency exactly. Zero-length spans legitimately fold into a row that a strict concurrency count would call full — they are empty intervals — and the packer still refuses to fold one *inside* a longer span (`[0,10]` + `[5,5]` → two rows), which is the visually right answer.
- **Edge cases exercised directly:** touching endpoints (`end == start`) share a row; zero-duration spans; unsorted input (`packToolLanes` sorts by `(startMs, endMs)`, and `Array.prototype.sort` is stable so equal pairs keep call order); identical timestamps — two calls both at `[8,9]` correctly take two rows; still-running calls, including two open calls of one name, which correctly take two rows since both draw to the task end; an output recorded before its own call, which clamps to a zero-width bar and cannot corrupt the row (`lastEnd` is a running `Math.max`). NaN timestamps cannot reach the packer at all: `msOf` (`timeline-chart.tsx:108`) returns null for anything non-finite and `buildGroups` drops those spans.
- **"Same name" cannot collide across subagents or sessions.** `trace-service.ts:1179` only pushes a `TraceToolSpan` when `!hasOrigin`, so a subagent's tool calls never enter the timeline; rows are built per `taskIndex` bucket, so two tasks never share one. An empty tool name falls back to `toolCallId` (`timeline-chart.tsx:190`), unique per call, so unnamed calls keep individual rows instead of all collapsing into one.
- **Performance.** `buildGroups` sits in `useMemo` on `[segments, toolSpans, otherSpans]` (`timeline-chart.tsx:314`) and the only call site passes memoised arrays (`trace-file-view.tsx:271` builds `tasks` in a `useMemo`; `t.spans` at :567), so packing runs once per data change, not per render. 20,000 serial same-name calls pack into one row in 2–4 ms. The `lanes.find` scan at `lane-packing.ts:59` is O(rows) per span, so the worst case is quadratic — 5,000 mutually overlapping same-name calls took ~45–115 ms and produced 5,000 rows. That needs 5,000 concurrent calls to one tool, which the chart could not display anyway; not worth a heap.
- **Against `origin/main` incl. #378.** `git merge-tree` reports no conflict, and #378 never touched `timeline-chart.tsx`. No new palette or icon-size classes: the only palette strings on `+` lines (`text-gray-500 dark:text-gray-400` on the row label, `ring-1 ring-red-500` on a failed bar) are moved verbatim, and `src/lib/tone.ts`'s header explicitly excludes both "the timeline phase bars" and secondary body text from the tone vocabulary. The diff contains no `InfoPopover` or `HelpFold`, so `test/disclosure-anchor.test.ts` has nothing to catch, and no icons, so `icon-scale.ts` does not apply.
- **No i18n gap.** The set of `S.*` keys is identical before and after — the row label is a tool name and every tooltip reuses an existing `S.traces.*` key — so `strings-en.ts` / `strings.ts` need no entry.
- `toolNames` was genuinely write-only before the change: declared and assigned at 25dc53e5, never read.
- Nothing became unreachable from packing itself: the per-bar `title` tooltip, the hover key (`w-`/`p-`/`e-` + `toolCallId`), click-to-jump and legend fade-out are all still per call, and the row label still names the tool.
- **Verification at 59f79f84:** `pnpm --filter @prismshadow/penguin-web test` → 76 files, 1007 tests; `tsc --noEmit` clean; `pnpm format` then `format:check` clean. Unrelated but worth knowing: a fresh worktree needs `pnpm -r build` before the web tests, or 23 files fail to collect on a missing `@prismshadow/penguin-core/omnimessage` — that happens on `main` too.

PR #40, opened 2026-07-22
feat: session source in session_meta and an Automated sidebar folder

hiyouga:
## What

Sessions created by automation carry their origin end to end: core `session_meta` gains a `source` field (`"subagent" | "schedule"`; **absent = user-created**) and is the **single source of truth** — the DB stores no source column. The chat sidebar files those sessions into collapsed per-origin folders (**Subagents / 子智能体** and **Scheduled / 定时任务**) parallel to **Archived**, inside every group in both grouping modes, and every group is display-capped at 20 sessions with a **More / 更多** row backed by server paging.

## Meta field and plumbing map

- **Core** (`packages/core`): `SessionMetaPayload.source?` (omnimessage/types.ts) and `CreateSessionOptions.source?` (agent.ts). `createSession` records the option into the meta (key omitted when unset); the run_subagent spawn site passes `source: "subagent"`.
- **Resume preservation**: `resumeSession` rebuilds the meta from the stored trace `session_meta` and carries `source` over — a resumed scheduled/subagent session stays marked, and a compaction-rotated trace file re-records it. On-disk values are narrowed: only exactly `subagent`/`schedule` pass; junk written by third parties is dropped, never cast through.
- **Server: session_meta as the single source of truth** (`packages/server`):
  - The `sessions` table has **no `source` column** (schema is CREATE-IF-NOT-EXISTS with no migration branches pre-release, so the column is removed outright; a stale dev DB keeps a dead nullable column that nothing reads or writes).
  - `SessionInfo.source` is **derived**: an in-process registry (`runtime/session-sources.ts`) is populated wherever the server has the meta in hand — session creation (reads the just-created core Session's meta), subagent registration (forwarded child meta, with the old "subagent" inference as fallback for pre-field metas), and Trace adoption. For sessions indexed by a **previous process**, the list path lazily reads the earliest Trace shard's `session_meta` once and caches it (this extends the existing session-index philosophy: DB rows are an index, Traces are the truth, and adoption re-reads Traces). Trace-less unknown sessions are not cached negatively — their meta may appear with the first run.
  - `SessionService.createSession` forwards the scheduler's `source` into core so the Trace records `"schedule"`; the self-heal rebuild re-records a registry-known origin into the fresh meta; the `session_created` event still carries `source` (now from the forwarded meta). All untrusted reads are narrowed via `asSessionSource`.
[...]

PR #364, opened 2026-08-20
feat(core,web,docs): let an MCP server's permission level be set by hand

hiyouga:
## What

An MCP server entry gained an optional `permission` key with three states — **`auto` (default) / `r` / `rw`** — that fixes the approval level of every tool that server exposes.

```yaml
tools:
  mcpServers:
    - name: linear
      config:
        transport: http
        url: https://mcp.linear.app/mcp
        permission: r        # auto (default) | r | rw
```

- `auto` is today's behaviour: each tool takes the level its own `readOnlyHint` annotation implies (`true` → `r`, anything else → `rw`).
- `r` / `rw` override the annotation for **every** tool of that server, in both directions.

## Why

`readOnlyHint` is optional in the MCP spec and a lot of servers never set it, so their whole toolset lands on `rw` and stops for approval on every call. There was no way to say otherwise short of switching the whole Session's approval mode.

Per **server**, not per tool: that is where the user makes the decision — at "add a server", before any tool has been discovered. A per-tool surface would make the stored config a function of discovery results, which are themselves a Session-lifetime snapshot.

## What the flag does and does not do

`permission` decides which of a server's tool calls PenguinHarness stops for human approval. That is the whole of it. It is **not a sandbox**: it does not restrict what the server's tools do once they run, it is never sent to or verified against the server, and the server keeps whatever capabilities its transport gives it. Marking a server `r` that can in fact write removes the prompt that would have caught the write. The changelog entry and `/docs/tools` both say this in as many words.

## Changes

**core** — `packages/core/src/environment/mcp/config.ts` validates the key next to `connectTimeoutMs` / `timeoutMs` / `maxOutputLength` and carries it on `ResolvedMCPServer` (unset = `auto`); an unrecognized value makes the entry invalid, which in this module means a human-readable warning and a skipped server, never a broken Agent. `provider.ts` applies it: `conn.server.permission ?? (readOnlyHint === true ? "r" : "rw")`.

**web** — `permission` joined `KNOWN_CONFIG_KEYS` (it no longer rides along in `extras`) and is threaded through `McpServerFormState` / `serverToForm` / `formToServer`. Saving `auto` writes no key, following the file's existing rule that a value equal to the default is not stored, so the entry keeps tracking the default. The Add/Edit modal carries the same three-way `OptionMenu` the builtin tool table uses, plus the `auto` row; the server table gained a `permission` column showing each entry's effective level. en + zh strings both added.

**cli** — nothing to do: `packages/cli` exposes no MCP server configuration at all (`config` covers models / vault / lang only), and its only MCP surface is echoing the `mcp_connect_begin` / `mcp_connect_end` events. No permission level is ever printed there.

**docs** — `/docs/tools § "MCP servers"` (en + zh): the key in the shared-options paragraph, the split permission-mapping bullet, a new bullet stating the boundary, and the YAML example. Configuration reference `tools.mcpServers` row and the Web App Tools-tab row, both languages.

**design** — `Prism-Shadow/penguin-harness-design` [#40](https://github.com/Prism-Shadow/penguin-harness-design/pull/40) updates `05-ARCHITECTURE.md` (the MCP config paragraph and the permission mapping), `06-PROTOTYPE.md` (the Tools tab) and `08-CHANGELOG.md`.

## Verification

```
pnpm --filter @prismshadow/penguin-core exec vitest run test/mcp-config.test.ts test/mcp-tools.test.ts   # 49 passed
pnpm --filter @prismshadow/penguin-web test                                                              # 992 passed (75 files)
pnpm --filter @prismshadow/penguin-server exec vitest run test/agent-config-mcp-servers.test.ts           # 8 passed
pnpm --filter @prismshadow/penguin-docs test                                                             # 53 passed
pnpm --filter {core,web,server} typecheck && pnpm format:check
```

New cases cover all three states and, most importantly, an explicit level beating a contradicting annotation in both directions — the stdio fixture's `echo` advertises `readOnlyHint: true` and `fail` advertises nothing, so one server exercises `rw` over a true hint and `r` over an absent one — plus an invalid value producing a warning and a skipped entry.

https://claude.ai/code/session_017Z4mn56LTt4qJR3m9GjHka


hiyouga:
Mechanically this is sound and I'd merge it: the override lands at the single place an MCP tool's permission is decided, an unrecognized value costs one server and a warning rather than the Agent, and the form round-trips without leaking the key into `extras`. The one thing I'd want settled first is a claim, not a bug — `permission` is consulted **only** when the session's approval mode is `read-only`, and the Web App's default mode is `allow-all`, so in the default mode `r` removes no prompt and `rw` adds none. The "What the flag governs" section exists precisely to state the boundary exactly, and that qualifier is the piece it's missing. It's copy in two languages across both repos, so I left it to you rather than rewriting it.

## Fixed

- **1f694887** — `packages/core/src/environment/environment.ts:212`: `toolPermission`'s JSDoc still said "MCP tools answer from their server's `readOnlyHint` annotation after discovery." It was the one `readOnlyHint` claim in the tree the PR didn't update.

## For you to decide

**1. "Stops for approval" is only true in `read-only` mode.**

`toolPermission(...)` is read in exactly one branch, in both frontends: `packages/server/src/runtime/approvals.ts:107-110` and `packages/cli/src/approval.ts:59-62`, `case "read-only"`. `allow-all` returns allow, `deny-all` returns deny, `always-ask` routes to manual — none of them look at the level. And the default is `allow-all`: `packages/web/src/features/chat/draft-view.tsx:179-181` for a new chat, and `--approve` for the CLI (`tools.en.md:229`).

So `r` doesn't remove a prompt and `rw` doesn't add one unless the session is running read-only. Note the direction — this is *not* a security overclaim, it makes `r` sound more consequential than it is. But `rw` reads as "make this ask", which it does not do outside read-only mode.

Affected: `packages/docs/content/tools.en.md:286` / `tools.zh.md:283`, `changelog/unreleased/2026-08-20-mcp-server-permission.md:32-33` (+ zh), `packages/web/src/lib/strings-en.ts:423` / `strings.ts:385`, and `design/specs/05-ARCHITECTURE.md:211`.

The precise phrasing already exists in the repo, two lines from the new strings: `strings-en.ts:381` — *"Only reads. Auto-approved when the approval mode is read-only."* And `tools.en.md:285` keeps "(auto-approved by the read-only approval mode)". It's only the new boundary bullet that drops it. Options: (a) add the qualifier in all six places, which means moving design [#40](https://github.com/Prism-Shadow/penguin-harness-design/pull/40) in step; (b) tighten only `mcpPermissionHint`, the string a user reads at the moment of choosing; (c) accept "approval" as shorthand and leave it.

**2. The new hint is permanently-visible semantics, which main's frontend skill now says to disclose.**

`.agents/skills/penguin-harness-frontend/SKILL.md` landed on main with #378 and splits explanatory text: formatting stays in the field's `hint`, semantics is *disclosed on request* — `InfoPopover` where a title exists, `HelpFold` where none does. `mcpPermissionHint` at `mcp-servers-section.tsx:506-508` is pure semantics under a labelled control, so by that rule it wants the "?" beside the field label.

It can't have one as things stand: #378 gave `Field` an `info` / `infoLabel` / `controlId` layout and `Input`/`Textarea` forward it, but `OptionMenu` doesn't, and that layout needs the control to carry an id. Counter-argument for leaving it: the same modal's `transportHints` and `mcpBudgetsHint` are still plain `<p>`s on main with the same classes, so converting only the new one makes the modal less internally consistent. Either leave it and sweep the modal later, or teach `OptionMenu` `info` as part of the sweep. `test/disclosure-anchor.test.ts` passes either way — it catches a misplaced `InfoPopover`, not a missing disclosure.

**3. The table column shows the configured mode; the prose calls it the effective level.**

`permissionOf` (`mcp-servers-form.ts:144-147`) returns `auto` both for an entry with no key and for one whose stored value is garbage — which core refuses to load at all (`config.ts:204-207` skips the whole server). A hand-edited `permission: readonly` therefore reads "auto" in the Tools tab while the Agent runs without that server. That's how the form treats every other bad known value, so it may be the price you want; the wording is the part that isn't quite right — "each entry's effective level" (`changelog…md:27`, `web-app.en.md:94`, design 06) doesn't hold for an `auto` row, whose tools each have their own level.

**4. Nits, none acted on.** `mcpPermissionAuto` is a dictionary key holding the untranslatable literal `"auto"` while its siblings `"r"` / `"rw"` are hardcoded fifteen lines below it (`mcp-servers-section.tsx:141-158`). `readPermission` (`config.ts:108-114`) substitutes the default into a local and then re-tests it — `if (value === undefined || value === DEFAULT_MCP_PERMISSION) return undefined;` says the same thing without the `unknown`. The servers table kept `min-w-[520px]` after gaining a fifth column.

## Checked and clean

- **Single derivation site.** `provider.ts:409-410` is the only place an MCP tool's permission is decided; `Environment.toolPermission` → `McpToolProvider.toolPermission` → the wrapper's `definition.permission` is the only read path, and nothing under `packages/*/src` re-derives a level from an annotation. `permission` sits on `ResolvedMCPServer`, never on the resolved transport, so it never reaches the wire — "never sent to or verified against the server" holds exactly.
- **Failure mode.** `readPermission` returns `null` for anything outside `auto`/`r`/`rw`, including a bare `permission:` in YAML (parses to `null`), matching how `readPositiveInt` treats `timeoutMs:`. `resolveMCPServer` throws, `resolveMCPServers` warns and drops that one entry. Covered at the resolver (unknown value, non-string value, and the list-level skip keeping the valid neighbour) and end-to-end ("skips a server whose permission value is not a level, keeping the Agent up"). The override tests are real, not tautological: the fixture's `echo` carries `readOnlyHint: true` and `fail` carries no annotation, so one server exercises `rw` over a true hint and `r` over an absent one.
- **Round-trip.** `r`/`rw` survive load→save unchanged, `auto` normalises to no key like the budgets do, and `permission` being in `KNOWN_CONFIG_KEYS` keeps even an unrecognized stored value out of `extras` (asserted). Server-side the key rides through verbatim — `validateMcpServers` validates via `resolveMCPServer` and returns the entry as-is, so a bad value is a 400 rather than a silently stored bad config.
- **Conventions.** Every new key is in both dictionaries (typecheck enforces it, since `Strings` comes from the zh side); no Chinese outside the zh dictionary, zh docs and zh changelog; no design-doc citations in comments; the changelog pair mirrors section for section, metadata block in order, `Issue`/`Breaking` correctly omitted; docs bilingual on all three pages; design [#40](https://github.com/Prism-Shadow/penguin-harness-design/pull/40) covers 05/06/08 and matches the code. CLI genuinely has no MCP configuration surface — its only MCP references are the two connect events in `render.ts`.
- **Against current main.** Merges clean (#378 touched this file's header and import block). On the merged tree the web suite is 80 files / 1017 passed, `disclosure-anchor` and `tone` included. On the branch: core `mcp-config` + `mcp-tools` 49 passed, web 992 passed, docs 53 passed, core/web/server typecheck clean, `prettier --check` clean on every touched file.

PR #365, opened 2026-08-20
fix(skills): state the penguin-sdk trigger and CLI-first tool wiring

hiyouga:
## What this changes

The `penguin-sdk` skill's trigger was implicit — its `description` only listed capabilities, so requests phrased as "build an agent" / "构建智能体" could fail to select it. The description now opens with the trigger sentence: use the skill whenever the user wants to build an agent, AI app, or any agentic application.

The body gains a **"Wiring in the user's tools"** section: integrate the user's existing tools as CLI commands the embedded agent invokes through the built-in `exec_command` tool (nothing to register, approval still gates every call, testable by hand), reserving MCP servers (`tools.mcpServers` in `system_config.yaml`) for integrations a CLI wrapper cannot express — long-lived authenticated connections or typed schemas.

Also:

- `short_description` / `short_description_zh` now name agent building; the bilingual docs skill tables (`packages/docs/content/skills.{en,zh}.md`) were updated to match.
- Skill `version` 19 → 20, `updated` stamped 2026-08-20.
- Bilingual changelog pair under `changelog/unreleased/`.

Note: shipped-skill updates reach existing agents through the usual skill update flow; nothing migrates automatically.

## Verification

- `pnpm --filter @prismshadow/penguin-skills test` — 22 passed.
- `pnpm --filter @prismshadow/penguin-docs exec vitest run test/skills-sync.test.ts` — 3 passed.
- `pnpm format` + `pnpm format:check` clean; `git diff --check` clean.

### DIFF
--- a/changelog/unreleased/2026-08-20-penguin-sdk-skill-trigger.md
+++ b/changelog/unreleased/2026-08-20-penguin-sdk-skill-trigger.md
@@ -13,10 +13,13 @@ user's own tools.
 ## Details
 
 - The skill's `description` now opens with the trigger: use the skill whenever the user
-  wants to build an agent, AI app, or any agentic application.
+  wants to build an agent application — their own program with an embedded agent, such as
+  an AI, agentic or RAG app. It also states the boundary: application code written on the
+  SDK, not configuring an Agent State inside PenguinHarness.
 - Added a "Wiring in the user's tools" section: integrate the user's existing tools as
   CLI commands the embedded agent invokes through the built-in `exec_command` tool,
   reserving MCP servers (`tools.mcpServers` in `system_config.yaml`) for integrations a
   CLI wrapper cannot express.
 - Extended `short_description` / `short_description_zh` and the bilingual docs skill
-  tables to name agent building, and bumped the skill version to 20.
+  tables to name agent applications and the same boundary, and bumped the skill version
+  to 21.
--- a/changelog/unreleased/2026-08-20-penguin-sdk-skill-trigger.md
+++ b/changelog/unreleased/2026-08-20-penguin-sdk-skill-trigger.md
@@ -1,0 +1,21 @@
+# penguin-sdk skill: explicit trigger and CLI-first tool wiring
+
+- **Date:** 2026-08-20
+- **Type:** process
+- **Scope:** `skills`, `docs`
+
+[中文版](2026-08-20-penguin-sdk-skill-trigger.zh.md)
+
+Sharpened when the `penguin-sdk` skill fires and what it recommends for wiring in the
+user's own tools.
+
+## Details
+
+- The skill's `description` now opens with the trigger: use the skill whenever the user
+  wants to build an agent, AI app, or any agentic application.
+- Added a "Wiring in the user's tools" section: integrate the user's existing tools as
+  CLI commands the embedded agent invokes through the built-in `exec_command` tool,
+  reserving MCP servers (`tools.mcpServers` in `system_config.yaml`) for integrations a
+  CLI wrapper cannot express.
+- Extended `short_description` / `short_description_zh` and the bilingual docs skill
+  tables to name agent building, and bumped the skill version to 20.
--- a/changelog/unreleased/2026-08-20-penguin-sdk-skill-trigger.md
+++ b/changelog/unreleased/2026-08-20-penguin-sdk-skill-trigger.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-20
 - **Type:** process
 - **Scope:** `skills`, `docs`
+- **PR:** [#365](https://github.com/Prism-Shadow/penguin-harness/pull/365)
 
 [中文版](2026-08-20-penguin-sdk-skill-trigger.zh.md)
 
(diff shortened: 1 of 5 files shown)


hiyouga:
Reviewed at `35f5f03`, diffed against `origin/main` (5 files, matching `gh pr view`). Every factual claim the diff introduces holds against the shipped code, the version bump and both bilingual surfaces are in place, the changelog pair is complete and correctly shaped, and the required tests pass. I found nothing to fix, so I pushed nothing. **Safe to merge as it stands** — with one thing I would want you to look at first, because it is a one-line change either way and it ships verbatim into the system prompt of every agent that has the skill installed: the new trigger claims the phrase `agent-creation` already owns.

## For you to decide

**1. "build an agent" / "构建智能体" collides with the preinstalled `agent-creation` skill.**

Neither skill sets `preinstall: false`, so both descriptions land verbatim in the same `{{SKILL_METADATA}}` list — one line per skill, sorted by name, full `description` (`skillMetadataSection`, `packages/core/src/state/agent-state.ts:574`):

```
- `agent-creation` — Create or configure an Agent State from a user requirement by writing AGENTS.md, setting identity metadata, and installing only needed Skills.
- `penguin-sdk` — Use whenever the user wants to build an agent, AI app, or any agentic application. Covers the Penguin Harness SDK — …
```

`agent-creation` is purely descriptive; `penguin-sdk` now opens with an unconditional imperative over the broadest available noun. On "build me a research agent" — and especially on 构建智能体, the exact phrase the PR description names as the motivating case — the in-product Agent reading is at least as likely as the standalone-app reading, and that case is `agent-creation`'s whole job. The group manifest already draws the line the description now blurs (Agent Tuning vs AI App Development, `packages/skills/src/index.ts:195`), and the skill's own body draws it too: `packages/skills/skills/penguin-sdk/SKILL.md:18` — "the SDK is for building applications, not for invoking agents".

Options as I see them: (a) leave it, on the view that a mis-fire is cheap because `## Before you start` recovers; (b) narrow the noun so it fires on the app case only — "an agent **application**", "an app with an embedded agent", "a standalone agentic app"; (c) keep it and give `agent-creation` a matching "use when" opening, so the two triggers separate each other instead of competing. Wording and scope are your call, so I left it alone.

**2. The new section promises approval gating, but the skill's only worked example denies everything.**

`SKILL.md:125` says "the `approve` callback still gates every invocation" — true, and I verified it: every complete `tool_call` goes through `approve` in `packages/core/src/engine/context-engine.ts:1132`, and a throw collapses to deny. What the section never says is that the app then has to return `"allow"` for the wrapped command. Its only full server example hardcodes `approve: async () => "deny"` (`SKILL.md:296`), correctly, because pure-retrieval QA wants no tools. A reader who follows the RAG recipe and then adds a `tools/` script per the new section gets a silently denied CLI. One sentence, or an allowlist-by-`cmd` snippet, would close it — but that is a scope decision about how much the section carries.

**3. A hand-edited `tools.mcpServers` does not survive "restore defaults".**

The MCP path at `SKILL.md:127` is correct and hand-editing the YAML really is the route — the penguin CLI has no `mcp` command, only `config model|vault|lang`. Worth knowing anyway: `resetSystemConfigToDefaults` replaces the entire `tools` block, `mcpServers` included (`packages/core/src/state/agent-state.ts:245-253`). Whether that caveat belongs in a section whose point is "prefer CLI, MCP is the exception" is a judgement call.

## Checked and clean

- **Every command, flag and path the diff introduces.** `exec_command` is a real built-in and is in the default tool list with no opt-in (`packages/core/src/environment/tools/exec-command.ts:30`, `packages/core/src/state/default-config.ts:643`). Its `workdir` defaults to the session's `workspaceDir`, which in this skill's own recipe is the project root (`SKILL.md:287`, `workspaceDir: ROOT`), so a relative `tools/…` resolves as the section assumes. `tools.mcpServers` in `system_config.yaml` matches the tree (`packages/core/src/interfaces.ts:72`, default `[]` at `default-config.ts:912`), as do `AGENTS.md` and `agent_state/`. The diff introduces no `penguin` CLI invocation, so there was nothing new to check in `packages/cli`.
- **Version bump.** 19 → 20, with `updated` moved 2026-08-06T00:00:00Z → 2026-08-20T00:00:00Z (`SKILL.md:6-7`).
- **Bilingual surface.** `short_description_zh` updated alongside `short_description`; both docs tables changed, and only the `penguin-sdk` row (`packages/docs/content/skills.en.md:69`, `skills.zh.md:69`). No other surface carries a per-skill description: the three README group tables are name-only, and `packages/landing/src/lib/strings{,-en}.ts` plus `packages/web/src/features/chat/example-tasks.ts` reference the id only. No test hardcodes the old description — core `markers`/`agent-skills`/`builtin-agents`, server `skills` and web `skill-slash` all read through `librarySkill()` or synthetic fixtures.
- **Changelog pair.** Both files present and mirroring section for section (lead paragraph + `## Details` / `## 细节`, three bullets each). Metadata in the fixed order Date / Type / Scope / PR, with `Issue` and `Breaking` correctly omitted rather than placeholdered. `Type: process` is right for a skills change and `Scope: skills, docs` matches the diff. Counterpart links in both directions; the PR written as a full link, added in a follow-up commit as the contract prescribes. No `## Why` / `## Verification` / `## Risks`.
- **Repo conventions.** Skill body and both docs tables English-only; Chinese confined to `short_description_zh`, the `.zh.md` entry and the zh docs page. No design-doc citations. No change narration or session vantage in the skill body — the two "now"s are in the changelog entry, which is the one place narration is correct (125 released entries of precedent). Every reference in the new prose resolves at HEAD.
- **Tests.** `@prismshadow/penguin-skills test` — 22 passed. Docs `skills-sync.test.ts` — 3 passed. `pnpm format:check` clean, `git diff --check` clean.

PR #366, opened 2026-08-20
feat(core,docs): raise the seeded compaction threshold to 256000

hiyouga:
Newly created Agents are seeded with `compaction.max_context_length: 256000` instead of `128000`.

## Why 256000 and not more

The vendors' long-context pricing boundaries. The catalog records base-tier rates only and says so (`packages/core/src/state/model-catalog.ts:16-18`): OpenAI charges extra above 272K input, Gemini 3.1 Pro above 200K, MiniMax M3 doubles every rate above 512K. At the old default of 128000 a default-configured Agent never reached any of them. 256000 clears all three while still doubling the context an Agent carries before compaction folds it; 512000 would have parked default Agents above two of the three boundaries, where the cost center under-reports every long Session.

## Why it is safe

The model's own `context_window` is the backstop. `effectiveMaxContextLength` takes the smaller of the configured value and `context_window − COMPACTION_HEADROOM` (2048) at **every use**, so a larger configured number can never push compaction past a window it would overflow:

| model window | fires at |
| --- | --- |
| 32768 (local vLLM) | 30720 — unchanged |
| 200000 | 197952 |
| no usable `context_window` | 125952 (from the assumed `DEFAULT_CONTEXT_WINDOW` of 128000) |
| above 258048 | 256000 — the seeded threshold itself |

That last row is not the rare case: 89 of the 93 built-in catalog entries that carry a `context_window` sit above 258048, so on nearly the whole catalog the seeded number is what fires. The four that do not are one 128000, two 200000 and one 256000 entry, where the window decides. The docs, the constant's doc comment and the changelog entry all state it that way.

## The two 128000s

`DEFAULT_CONTEXT_WINDOW` (core `llm/context-limits.ts`, mirrored in web `lib/context.ts`) is **also** 128000 and is a different fact: the assumed window for a model entry that configures none. It is deliberately left alone, so the two constants no longer share a value. Both doc comments name the other one, and the backstop test asserts the unconfigured-window path against `DEFAULT_CONTEXT_WINDOW` explicitly rather than a bare literal.

## Details

- The seeded value moved into `DEFAULT_MAX_CONTEXT_LENGTH` (`packages/core/src/state/default-config.ts`), read both by `defaultSystemConfig()` and by the Agent composition layer's fallback for a config with no `compaction` section — previously two independent `128000` literals.
- `COMPACTION_HEADROOM`'s doc comment argued its case with numbers that named the old default. Rewritten so every claim in it is verifiable at HEAD.
- Bilingual `configuration` and `agent-loop` docs record the new default and what actually decides the trigger point.
- New test `backstops the seeded default threshold with the model's context window` in `core/test/context-limits.test.ts` covers the shipped constant across all four window cases. It is written against the constant, so it did not need renumbering when the value changed.

## Kernel version

Changing a built-in default trips the pinned-hash guard, so `KERNEL_VERSION` advanced to generation `2026-08-20` with `compaction.max_context_length` as its only changed leaf (`b412afcc…` = sha256 of `256000`). The `2026-08-19` generation shipped in v0.2.3 and is frozen. `2026-08-20` has never shipped, so when the value moved from 512000 to 256000 that generation was revised in place rather than stacked on — the header comment sanctions exactly this for an entry no later generation has frozen.

## Compatibility — no migration, by decision

Existing Agents keep whatever their `system_config.yaml` stores (128000 for anything created before this). An Agent runs from its config file verbatim and nothing rewrites it. Only Projects and Agents created after this change are seeded with 256000.

Three explicit ways to adopt it: edit `compaction.max_context_length` on the **Runtime** tab of the Agent settings page (or in `system_config.yaml`), **Update kernel**, or **Restore default configuration**. The kernel bump means existing Agents show the kernel-update hint; under Update kernel a stored value still equal to a recorded generation's default advances to 256000, a user-edited value is kept and reported.

No compatibility shim, no migration code, no `backward-compatibility.md` entry.

## Verification

```
pnpm format && pnpm format:check                                  # clean
pnpm --filter @prismshadow/penguin-core typecheck                 # clean
core: context-limits + kernel-version + state + agent + compaction  # 170 passed
server: config + agent-kernel-update + agent-config-reset           # 14 passed
web: context.test.ts                                                # 4 passed
docs: whole package                                                 # 53 passed
```

## Left alone deliberately

- `DEFAULT_CONTEXT_WINDOW` in core and web, `CUSTOM_CONTEXT_DEFAULT` in the models page, the donut/ring fallbacks, `openrouter/free`'s catalog `contextWindow` — all context-window facts, not this default.
- Published blog posts (`simple-harness-is-all-you-need`) and released changelog folders quoting 128,000 — dated artifacts.

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/41

https://claude.ai/code/session_017Z4mn56LTt4qJR3m9GjHka


hiyouga:
Reviewed at `d3342397` against `origin/main` (12 files, matching `gh pr view`). **Safe to merge.** The arithmetic backstop holds on every path I could find, the kernel bump is correct in every detail I checked, and nothing auto-migrates. The one real defect was a factual claim repeated in three places — that 512000 sits above every current model window — which is false for 74 of the 93 built-in catalog entries; that is fixed and pushed. Everything else below is a judgement call I deliberately did not make for you.

## Fixed

**`67888a83` — the seeded threshold is not above every model window.**

`packages/core/src/state/default-config.ts:101` ("Set well above every current model window on purpose"), `packages/core/test/context-limits.test.ts:198` ("deliberately larger than any current model window") and `packages/docs/content/agent-loop.{en,zh}.md` ("so it is the model's window that decides", unqualified) all asserted that the model window is what decides the trigger point. `packages/core/src/state/model-catalog.ts` carries 74 entries with `context_window` above 514048 — 40× `1000000`, 19× `1048576`, 14× `1050000`, 1× `524288` — out of 93 entries with a window at all. On every one of those the seeded 512000 is exactly what fires, so the claim is wrong for ~80% of the shipped catalog, and it contradicts this PR's own description ("> 512000 | 512000 — the only case where the configured threshold itself fires") and the changelog's `## Details` bullet 2.

All three sites now state the actual rule — the effective threshold is the smaller of the seeded value and `context_window − COMPACTION_HEADROOM` — and the changelog's `## Docs` bullet follows in both languages so the pair stays mirrored. `packages/docs/content/configuration.{en,zh}.md` already had the qualifier and is untouched.

Verified: `prettier --check .` clean, core `tsc --noEmit` clean, core `context-limits` / `kernel-version` / `state` / `agent` (129 passed), docs package (53 passed), server `config` + `agent-kernel-update` + `agent-config-reset` (14 passed).

## For you to decide

**1. The default now sits above three vendors' long-context pricing boundaries.** `packages/core/src/state/model-catalog.ts:16-18` records base-tier rates and says so explicitly: "OpenAI charges extra for >272K input, Gemini 3.1 Pro for >200K input, and MiniMax M3 doubles every rate above 512K input; this catalog records their base tier (the cost center uses a single rate, so long-context usage will be underestimated)." At the old default of 128000, a default-configured Agent never reached any of those boundaries. At 512000 it routinely will, so the Usage / cost figures become systematically low for default-configured Agents on those models — a new consequence of this change, not a pre-existing one. This is a pricing / product call: accept it, note it in the entry, or seed lower (272000 would clear OpenAI and Gemini). I did not touch it.

**2. The design PR carries the same claim I just corrected here.** `specs/05-ARCHITECTURE.md` on `docs/compaction-512k` (design PR #41) says 缺省阈值 512000 **刻意高于当前所有模型窗口**, and 故实际触发点**一般**由窗口决定. The first is false by the catalog count above; the second is hedged but points the wrong way when 80% of entries are on the other side. `CLAUDE.md` says to confirm with you before changing a design doc, so I left `design/` alone — but the design and this branch now disagree on that sentence, and design is meant to win. It needs the same edit in #41 before or with the merge.

**3. "No migration" has one uncovered case.** `packages/core/src/agent.ts:913-916` falls back to `DEFAULT_MAX_CONTEXT_LENGTH` when the stored config has no `compaction.max_context_length`. Sharing the constant is the right call (that is the whole point of extracting it), but it means an existing Agent whose `system_config.yaml` lacks the key — hand-edited, or a config predating the field — moves from 128000 to 512000 with no user action. `defaultSystemConfig()` always writes the key, so this is narrow, but the changelog's flat "every Agent already on disk keeps the `compaction.max_context_length` it stores" does not cover it. Options: leave it, or add a clause to the `## Compatibility` bullet in both languages.

**4. Nit, not acted on.** `packages/core/src/llm/context-limits.ts:79` names both the symbol and the literal — ``(`DEFAULT_MAX_CONTEXT_LENGTH`, 512000)``. Two representations of one fact; it goes stale if the constant moves again. I left the literal because the magnitude is what carries the "far outside such a window" argument, but dropping it costs nothing.

## Checked and clean

- **The cap arithmetic, traced by hand.** `effectiveMaxContextLength` (`packages/core/src/llm/context-limits.ts:193`) is `configured <= 0 ? configured : min(configured, (resolveContextWindow(w) ?? DEFAULT_CONTEXT_WINDOW) - 2048)`. With 512000: 32768 → 30720; 200000 → 197952; `undefined` / `2048` / `"unknown"` → 125952; 1000000 → 512000. Break-even is a window of 514048. `<=0` still short-circuits before the cap, so the "compaction disabled" contract is intact.
- **No path escapes the cap.** `packages/core/src/agent.ts:913` is the only construction site of `CompactionSettings` in production, and `composeRuntime` is shared by `createSession`, `resumeSession` and the subagent path (`childAgent.createSession`, `agent.ts:684`), so nested sessions go through the same derivation. `packages/core/src/engine/context-engine.ts:1300` only reads the already-capped value. The server exposes the raw stored number in the config DTO (`agent-config-service.ts:175`) — correct, that is the editor's value — and its PUT validator has no upper bound that 512000 would trip.
- **The kernel hash.** `sha256("512000")` = `aa4e55a97c93d61fbad2fb4733b5421c94c5f280785cc52ad406a6e9c59069a4`, matching `kernel-history.ts:302`. Diffing the `2026-08-19` and `2026-08-20` maps leaf by leaf: 27 leaves each, exactly one differs (`compaction.max_context_length`, from `8eda794b…` = sha256("128000")). The `2026-08-19` generation is byte-identical apart from the doc-comment reflow. `KERNEL_VERSION` equals the newest history key, so the pinned-hash test enforces both directions.
- **No auto-migration.** `packages/core/src/state/agent-state.ts:170` reads an existing `system_config.yaml` verbatim with no defaults merge; the `defaultSystemConfig()` spread is only on the init branch. `applyKernelUpdate` has exactly one production caller, `packages/server/src/services/agent-config-service.ts:293`, behind the explicit kernel-update endpoint; `resetSystemConfigToDefaults` likewise. Nothing runs either on load or startup.
- **The two 128000s are not conflated anywhere.** `DEFAULT_CONTEXT_WINDOW` (core `llm/context-limits.ts:41`, mirrored in `packages/web/src/lib/context.ts:9`) is untouched and still 128000. Every remaining 128000 in the tree is a context-window fact, a frozen released changelog folder, or a dated blog post. `packages/docs/content/models.{en,zh}.md`'s "compaction derives from an assumed 128000" is still true. The new test's `expect(DEFAULT_MAX_CONTEXT_LENGTH).toBeGreaterThan(DEFAULT_CONTEXT_WINDOW)` is a real guard against re-conflation, not decoration.
- **Conventions.** Changelog pair complete and mirrored section for section (`Details`/`细节`, `Compatibility`/`兼容性`, `Docs`/`文档`, same bullet counts); metadata block in `Date`/`Type`/`Scope`/`PR` order with `Issue`/`Breaking` correctly omitted; counterpart link directly below it; no `## Why` / `## Verification` / `## Risks`; entry in `unreleased/`, no index file. Docs changes bilingual. All new code comments English, no design-doc citations. The UI labels the entry names — **Runtime** tab, **Update kernel**, **Restore default configuration** — match `packages/web/src/lib/strings-en.ts` (`agent.tabRuntime`, `agent.kernelUpdateAction`, `agent.resetConfigAction`).
- **Interaction with the three commits `main` picked up after this branch.** `git merge-tree` against `origin/main` is conflict-free. #378 (web tone tokens, `InfoPopover`/`HelpFold`, its AST guard) touches only `packages/web`, which this PR does not; #381 and #385 are server/CI test plumbing. Nothing in web or server hard-codes the compaction default — `agent-settings-page.tsx` renders whatever the config holds, `web/e2e/compaction.spec.mjs` pins its own `maxContextLength`, and `web/test/kernel-labels.test.ts` asserts label coverage, not values. The leaf set is unchanged, so no new kernel label is needed.
- **Prose.** The `COMPACTION_HEADROOM` rewrite is accurate ("fire at ≈ 30720 instead of never" — without the cap a 512000 threshold on a 32768 window is genuinely unreachable), the change narration in the old test comment is gone, and the `kernel-history.ts` generation notes now read as a frozen list with only the newest marked current.

hiyouga:
Follow-up to my review above — two of the three open items are now closed, and the number in that comment is stale. Reading it after this: everywhere it says 512000, read 256000.

**The default is now 256000** (`2f223191`). The reason is the pricing item I raised: at 512000 a default-configured Agent would sit above two of the three long-context boundaries the catalog records at base tier only (OpenAI >272K, Gemini 3.1 Pro >200K, MiniMax M3 >512K), so the cost center would under-report every long Session. 256000 clears all three and still doubles the context an Agent carries before compaction folds it.

**The correction I made in `67888a83` gets sharper, not softer.** The break-even moved from a 514048 window to 258048, so the seeded number is what fires on **89 of the 93** catalog entries that carry a `context_window` — the four exceptions are one 128000, two 200000 and one 256000 entry. The docs, the constant's doc comment and the entry's worked examples all say it that way now.

Worked examples re-verified against the shipped function rather than by hand — 32768 → 30720, 200000 → 197952, no usable window → 125952, 1000000 → 256000. The backstop test needed no renumbering: it was written against `DEFAULT_MAX_CONTEXT_LENGTH`, not literals, which is exactly what that buys.

**Kernel version: kept at `2026-08-20`, revised in place.** `KERNEL_VERSION` and the generation key both stay `2026-08-20`; only the `compaction.max_context_length` hash changed, to `b412afcc…` (sha256 of `256000`). I verified the 512000 hash is gone from the file entirely and that the generation still differs from `2026-08-19` by exactly one leaf. Reasoning for not re-dating to today: `kernel-history.ts`'s own header sanctions revising an entry no later generation has frozen, and this one has never shipped — v0.2.3 shipped `2026-08-19`, which is untouched. The key is an ordering token whose only hard constraints are `KERNEL_VERSION` equalling the newest key and sorting after `2026-08-19`; both hold. Re-dating would change nothing observable — no config on any disk carries either key — while desynchronizing the stamp from the changelog entry's `Date:` and filename prefix, which are the one place this change's date is a record rather than a token.

**Design PR [#41](https://github.com/Prism-Shadow/penguin-harness-design/pull/41) is fixed** (`0c4894d`), with the user's go-ahead. Both problems: 刻意高于当前所有模型窗口 is gone (it was wrong for 89 of 93 entries), and `05-ARCHITECTURE.md` now states the real boundary — window decides below 258048, threshold above — plus the pricing reason for choosing 256000, which belongs in the design repo since 08-CHANGELOG records 关键决策. The spec YAML, its prose and the 2026-08-20 changelog line all carry 256000. Title and body updated on both PRs.

**Still open, unchanged by this:** the `agent.ts:913` fallback (a config with no stored `compaction.max_context_length` moves to the new default with no user action — same shape at 256000, just a smaller move), and the `context-limits.ts:79` symbol-plus-literal nit.

Re-verified: `prettier --check .` clean, core typecheck clean, core context-limits + kernel-version + state + agent + compaction (170), server config + kernel-update + config-reset (14), web `context.test.ts` (4), docs package (53). Still a clean merge with `origin/main`.

PR #363, opened 2026-08-20
fix(model-catalog): add the direct claude-opus-5 preset

hiyouga:
Closes #352.

## What was missing

`packages/core/src/state/model-catalog.ts` carried the OpenRouter row `anthropic/claude-opus-5` but the direct Anthropic group went `claude-fable-5` → `claude-opus-4-8` → `claude-opus-4-7` → `claude-sonnet-5` → `claude-sonnet-4-6`. A Project holding only an `ANTHROPIC_API_KEY` therefore had no preset for Opus 5 and had to add the model by hand. (The design changelog recorded the direct row as deliberately deferred on 2026-07-24, pending AgentHub registry coverage; AgentHub routes `claude-opus-5` today via the `claude` + `-5` substring branch, so the deferral is over.)

## The row

```ts
{
  modelId: "claude-opus-5",
  displayName: "Claude Opus 5",
  provider: "anthropic",
  contextWindow: 1000000,
  pricing: usd(0.5, 6.25, 25),
  supportsVision: true,
},
```

Placed between `claude-fable-5` and `claude-opus-4-8` — the catalog's dictionary order by model id with newer versions of a series first, hand-precomputed, matching the order the OpenRouter Claude rows already use.

Pricing read on 2026-08-20 from <https://platform.claude.com/docs/en/about-claude/pricing>: Claude Opus 5 is `$5` base input / `$6.25` 5m cache write / `$0.50` cache hit / `$25` output per MTok, and Claude 4.6-and-later models bill the full 1M window at one rate. That is exactly `usd(cache_read=0.5, cache_write=6.25, output=25)` under the group's `cache write = 1.25 x input` convention, and it agrees with the OpenRouter twin to the cent.

Uniqueness in the catalog is the `(provider, model_id)` pair, so the direct row and the gateway row coexist by design.

## Sibling-row audit

Every other Anthropic row was checked against the same page. All five match, so nothing was re-priced:

| Row | Catalog | Page (cache hit / 5m write / output) |
| --- | --- | --- |
| `claude-fable-5` | `usd(1, 12.5, 50)` | $1 / $12.50 / $50 |
| `claude-opus-4-8` | `usd(0.5, 6.25, 25)` | $0.50 / $6.25 / $25 |
| `claude-opus-4-7` | `usd(0.5, 6.25, 25)` | $0.50 / $6.25 / $25 |
| `claude-sonnet-5` | `usd(0.2, 2.5, 10)` | $0.20 / $2.50 / $10 |
[...]

PR #368, opened 2026-08-20
feat(web,server,core): import a provider's whole model listing when adding a group

hiyouga:
#360 is merged; this branch has origin/main merged in and its base retargeted to main.

## What this adds

The models page's "add group" dialog now has two modes, chosen under the name field:

- **Create only** — the light path: a valid name hands off to that group's add-model dialog.
- **Import models** — fills the brand-new group from its endpoint, in the add-model dialog's field rhythm: **API key** first, then **Base URL** with the **Detect** action at its top-right and the in-field protocol picker (`ProtocolSuffixMenu`) as manual override. Once a protocol is determined — detected or hand-picked — **Import all models** appears: it fetches every model id the endpoint serves and appends them all as entries of the new group in one table PUT, each carrying base URL, protocol, and typed key inline. A failed detection turns the suffix amber and blocks nothing (pick by hand, or switch back to create-only); listing failures render inside the dialog and persist nothing.

Also in this PR:

- **Whole-group delete** for user-defined groups: a trash action on the group header, one ConfirmModal naming the group and its model count, then one table write removes every model in it — default/vision pointers into the deleted group are dropped exactly as single-model delete does. Built-in groups never offer it.
- **Probes no longer disable thinking**: the connectivity test and the vision probe send `thinkingLevel: "low"` (the lowest real level) instead of `"none"` — several reasoning endpoints reject a request that turns thinking off outright, so the probe failed on the knob it sent rather than on the endpoint. (Same fact the session picker already documents: most models can't switch thinking off.)

## How it works

- **Core** exposes `listEndpointModels({clientType, apiKey?, baseUrl?})`, a thin wrapper over AgentHub 0.4.5's `AutoLLMClient.listModels()` (agenthub [#183](https://github.com/Prism-Shadow/agenthub/pull/183)): the server keeps no direct AgentHub dependency. Routed through a generic protocol client the listing is the endpoint's whole catalog; errors propagate for callers to collapse.
- **Server** adds owner-only `POST /api/projects/:p/models/list` (`{baseUrl, clientType, apiKey?}` → `{ok, models?, unsupported?, message?}`): absolute-http(s) URL validation like `/detect`, DTO-reduced replies, keys only in upstream request headers. Failures collapse into `ok:false` (`unsupported` flags AgentHub's `UnsupportedOperationError`), listings bounded at 20s. An omitted key follows the SDK's per-protocol environment fallback.
- **Web**: `AddGroupDialog` owns the two-mode flow (Segmented mode switch; detect supersession via a run counter, same convention as the model dialog); `buildImportedRows` keeps endpoint order, dedupes against existing `(provider, modelId)` pairs and within the listing, and counts skips for the toast. Group delete filters by provider and clears matching pointers before the PUT.
- Docs: the custom-protocol section of `models` and the `server-api` route table updated for the two-mode flow, both languages; the bilingual changelog entry covers all three changes.

## Verification

- `pnpm -r build`, `pnpm typecheck`: clean.
- `packages/core` `test/list-endpoint-models.test.ts`: 3 passed (constructor routing, env-fallback omission, error passthrough — AgentHub mocked).
- `packages/server` model routes/services (`models`, `models-list`, `protocol-detect`, `model-probe`, `models-vision`, `models-fast-mode`): 75 passed — service outcome collapsing incl. `unsupported` flag and timeout bound, route validation, owner-only guard, stubbed end-to-end DTO shape; all green with the probes on `thinkingLevel: "low"`.
- `packages/web` full suite: 973 passed (incl. `group-import.test.ts`: order, dedup counts, `rowToEntry` persistence shape with and without a key).
- `packages/docs` tests (search index over the edited pages) passed; `pnpm format:check`, `git diff --check`: clean.


hiyouga:
Reviewed at `52607ab2`, diffed against `origin/main` after merging it in.

**Verdict: yes, once you're happy with the two behavioural corrections below.** The import gets the thing that matters most right: it takes *only* the model id from the endpoint and leaves pricing, context window and display name empty, so no provider-reported number can land in a field the hand-curated catalog owns, and `(provider, model_id)` uniqueness holds for free because the group name must be new (so an imported id can never collide with a catalog preset). Two things were not right and are fixed here: every imported row claimed vision support — the *opposite* of what the same model gets when added to that group by hand — and a single id the config cannot hold (blank, over 200 chars, control characters) 400'd the entire table PUT, after which `persist` left the phantom group on screen and the dialog's own name check refused the retry as "name taken". The four conflicts are resolved, the PR is `MERGEABLE` again with every check green (`ci`, `ci-macos`, both `ci-windows` shards, all three `runtime` jobs, `build`), and it shrank from 26 files to 19 because the AgentHub 0.4.5 bump it was carrying is now identical to what #360 landed on main.

## Fixed

**The merge — `9170e893`.** Exactly the four files, resolved as follows:

- `changelog/unreleased/2026-08-20-agenthub-0.4.5{,.zh}.md` (add/add) — took main's and dropped this branch's duplicate. I read both: main's is the reviewed text from #360 (`listModels()` / `defaultHeaders` spelled as the API spells them, not `list_models()` / `default_headers`; the client families named explicitly in the first bullet). The branch copy documented nothing main's did not, so nothing to merge across.
- `pnpm-lock.yaml` — took main's file whole, then re-ran `pnpm install`: "Lockfile is up to date, resolution step is skipped", no diff produced, and `pnpm install --frozen-lockfile` passes. Both sides had already moved agenthub to `^0.4.5`, which is also why `pnpm-workspace.yaml`, both `package.json`s and `protocol-path.ts`'s version note auto-merged and then vanished from the diff.
- `packages/web/src/features/models/models-page.tsx` — a real content conflict, and only from #378 (#363 never touched this file). Kept this branch's `AddGroupDialog` call site, and ported #378's change into the new component: the group-name `Input` now carries `S.models.addGroupDesc` behind its "?" (`info` / `infoLabel`, models-page.tsx:1254-1255) instead of the standing `<p>` the branch had re-added, and its wrapper is a `<div>` because `Field` splits the label out under `info`. That is the `.agents/skills/penguin-harness-frontend/SKILL.md` rule — semantics disclose — and `test/disclosure-anchor.test.ts` passes with it.

**`934f0e87` — imported models no longer claim vision, and unusable ids are skipped instead of sinking the import.**

- `group-import.ts:50` built every row with `vision: true`, commented as "custom-model semantics". It is the reverse: models-page.tsx:1549 starts a hand-added model as `vision: !isCustomLikeGroup(addProvider)`, i.e. **false** for a user-defined group, with an explicit "A new custom model claims no vision support until it is detected or switched on by hand (per maintainer)". `rowToEntry` omits `vision` when true, so every imported row persisted as vision-capable and a text-only model would be handed images directly instead of via the vision proxy. Now `vision: false`, with the comment corrected.
- `group-import.ts:22-38` — the listing is provider-supplied text and every string in it became a config key unchecked. `models[i].modelId` is `minLen 1, maxLen 200` at routes/models.ts:66-70, so one over-long id 400'd the **whole** PUT: the entire import lost behind `models[57].modelId must be at most 200 characters`. Entries that are blank after trimming, over 200 characters, or carry C0/DEL characters are now skipped and counted; ids are trimmed before keying.
- Because `skipped` no longer means "duplicates" only, the toast says "skipped N entries" in both dictionaries, and `groupImportEmpty` was reworded ("No models to import from this endpoint" / "该端点没有可导入的模型") since `added === 0` now also covers a listing whose every entry was filtered.
- Docs and the changelog entry state both rules and the "only the id is taken" invariant. The docs paragraph also moved *below* "Nothing is inferred from the model id in these groups" — it was inserted with a leading blank line but no trailing one, so that sentence had been swallowed out of the manual-override paragraph and into the import one.

**`8601c973` — a failed import save no longer wedges the dialog.** `persist` applies the attempted rows on failure (models-page.tsx:556 — right for a field edit the user is still holding). For an import that left the whole group rendered as if it had saved, and `AddGroupDialog.validName()` then read it out of `rows` and answered `groupNameExists` on the retry the dialog stays open for — no way forward but a reload. `importGroup` now captures the table it read and restores it when `persist` returns false (models-page.tsx:641-651).

**`8601c973` — `label="API key"` and `Base URL` were English literals in JSX.** A zh reader saw "Base URL" where the add-model dialog two hundred lines down says "自定义 base URL". Both now read `S.models.apiKey` / `S.models.baseUrl` (models-page.tsx:1285, 1300, 1324) — the same keys the add-model dialog uses at :2329 and :2723.

**`835ba6dc` — dropped a duplicate string pair.** `groupImportNoProtocol` was a second sentence for the condition the model dialog already reports as `detectFailedBody`: same trigger (a detection run that returned no protocol), same advice (check the API key and the base URL), two translations to keep in step. The dialog uses the existing pair; `models.en.md` already promises every detection failure "says the same thing".

**`3d03e42d`, `52607ab2` — prose that would not survive the session.** Editing "thinking disabled" → "thinking at the lowest level" in place left two doc comments wrapping mid-clause, and "a reasoning-heavy model **may still** burn the whole tiny output cap" narrated an edit relative to a sentence that is no longer there. And "the abandoned promise is dropped — **the SDK's own timeout bounds the socket**" is a claim about AgentHub internals nobody reading this repo can check; it now states what is checkable: the race stops waiting, nothing cancels the request behind it, and the loser's later rejection is already handled by the race.

## For you to decide

- **Nothing bounds how many models an import writes.** The service returns whatever AgentHub hands back (project-config-service.ts:614-643), the browser turns all of it into rows, and one PUT writes them into `.project_config.toml`. The only ceiling today is the global `bodyLimit` at app.ts:471 (the admin-settable attachment limit), so a gateway serving 5,000 ids produces a 5,000-row group and a 5,000-row models page, with no count shown before the write. I did not pick a number because it is a product call — OpenRouter alone serves several hundred, so anything under ~1,000 rejects a legitimate endpoint. Options: a hard cap in the service; or a confirm step naming the count (and possibly a selection list) before `onImport`; or leave it and rely on the body limit.
- **The plaintext key is stored once per imported row.** `importedRow` inlines `config.apiKey` on every row, so a 400-model import writes the same key 400 times into `.project_config.toml`. That is how the bulk-key dialog already writes a group, so the import did not invent it — but it is the first thing that turns a handful of copies into hundreds. A group-level credential is a config-shape change, so I only flagged it.
- **`persist`'s failure contract, for the other four callers.** I fixed only `importGroup`. Whole-group delete (models-page.tsx:958) still removes the group from the page when its PUT fails while the server keeps it — but that is exactly what single-model delete has always done (:1033, :1050), so making them all roll back is a decision about the contract rather than a local fix.
- **Two injection styles for one route.** `listEndpointModels(req, listImpl = coreListEndpointModels, timeoutMs = LIST_MODELS_TIMEOUT_MS)` puts a test seam in the public signature, where the neighbouring `detectProtocol` just calls its module function; the route test then *also* monkeypatches the method (models-list.test.ts:122-125). Core's own test mocks the module with `vi.mock` instead. Worth settling on one.
- **Imported entries carry no context window**, so every one of them falls back to `DEFAULT_CONTEXT_WINDOW` = 128000 (core `context-limits.ts:41`) for compaction. Correct in that no number is invented, but a 32K gateway model will compact far too late. Taking `context_length` from listings that publish one (OpenRouter does; the plain OpenAI `/models` shape does not) is an API-shape change and reopens the provider-reported-numbers question, so I left it — the docs now say the field stays empty.
- **`unsupported` is a string compare across the AgentHub boundary** — `err.name === "UnsupportedOperationError"` (project-config-service.ts:634, and `list-models.ts`'s doc says the same). It works, but the harness does not own that name; 0.4.5 exports the class, so core could `instanceof` it and hand the server a typed flag. Left alone because it crosses the core/server seam this PR deliberately keeps thin.
- **The mode `Segmented` has no label** — it is the one control in the dialog without one, sitting under the name field. Both options are self-describing, so this is taste, not a bug.
- **No cancel for a slow listing.** The 20s bound is server-side only; the dialog shows "Fetching model list…" with Cancel disabled and no way to abandon it (`core listEndpointModels` takes no `AbortSignal`). Adding one is an API change.

## Checked and clean

- **Catalog conventions.** Nothing but the id crosses from the endpoint: `rowToEntry` receives empty `contextWindow` / `cacheRead` / `cacheWrite` / `output` / `displayName`, so the persisted entry is exactly `{provider, modelId, clientType, vision:false, baseUrl, apiKey?}` — asserted in `group-import.test.ts`. No provider-reported price can end up in the three-bucket USD fields, and no tiered-pricing question arises. A collision with a catalog preset is structurally impossible: `validName()` rejects any name in `MODEL_PROVIDERS` or already used as a `provider` on an existing row, so the new group's `(provider, model_id)` pairs are all fresh, and `syncRowsWithCatalog` never touches a user-defined provider.
- **Failure and partial states.** 401/timeout/unreachable all collapse to `{ok:false, message}` truncated at 300 chars (verified in `models-list.test.ts`); `UnsupportedOperationError` additionally sets `unsupported` and the dialog points at the manual path; the 20s bound is real and matches the connectivity test's own `requestTimeoutMs: 20_000` at :529. Nothing is written before the listing succeeds — the only write is the single PUT — and with `934f0e87` a partially-junk listing imports its usable half instead of failing whole.
- **Route surface.** Owner-only via `requireProjectOwner`, `requireValidId` on the project, absolute-http(s) `baseUrl` (2000 max), `clientType` (64 max), `apiKey` type-checked, reply reduced to a DTO, key only in the upstream request headers — the same shape and the same server-fetches-a-caller-URL exposure `/detect` and `/test` already have; no private-IP or redirect filtering, but that is unchanged from `/detect`, not new here.
- **Whole-group delete.** Offered only when the provider is not in `MODEL_PROVIDERS`, so `custom` and every vendor/gateway group are excluded; the confirm names the group and the *unfiltered* row count (so it stays honest while a search filters the visible rows); default/vision pointers into the group are dropped before the PUT, mirroring the bulk-key handler at :935 line for line.
- **#378 conventions.** No new palette class: the amber suffix goes through `ProtocolSuffixMenu`'s existing `tone="warn"`, the required-asterisk red and the brand-coloured Detect link are copied verbatim from the add-model dialog's own base-URL block (neither is a status colour). The new `TRASH_ICON` renders through `GlyphIcon` at `13`, matching the four sibling group-header glyphs at :775/:792/:810/:848 — that whole file still uses bare numbers rather than `ICON_SIZE`, so migrating it is a separate pass, not this PR's. `disclosure-anchor.test.ts` and `icon-scale.test.ts` both pass.
- **i18n.** Every new key exists in both dictionaries with matching placeholders (`placeholders-parity.test.ts` passes); no Chinese outside the zh dictionary; comments, tests and docs English.
- **Changelog.** Bilingual pair, mirrored section for section, metadata in `Date` / `Type` / `Scope` / `PR` order with `Issue` and `Breaking` correctly omitted, counterpart links both ways, lead paragraph plus a single bespoke `## Details` / `## 细节` — no `## Why` / `## Verification` / `## Risks`, no index touched. `#368` and agenthub `#183` are both real PRs.
- **Verification run here** (Node 24.18, pnpm 11.18): `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck` (8 packages), `pnpm format` → `format:check`, `git diff --check`; web 1066/1066, server 816/816, core 852 passed + 5 skipped, docs 53/53. Server model routes and `list-endpoint-models.test.ts` included.

PR #370, opened 2026-08-20
fix(desktop): import the login shell environment on GUI launches

### LINKED ISSUE #351: desktop can read env vars for model authing
桌面版应该也能读取环境变量来使用模型 api，无需重复添加。包括模型连接和 bash tool

hiyouga:
## What this fixes

Fixes #351. Two halves:

1. **Import the login shell environment on GUI launches.** A desktop app started from the macOS Dock or a Linux desktop session inherits the session manager's environment, not the user's login shell — so API keys exported in `.zshrc` / `.profile` never reached core's model env fallback, and the agent shell ran on the bare system `PATH`. Users had to re-enter keys that already worked in their terminal.
2. **Make the fallback visible — for first-party official entries only.** Even with the variables present, the Models page said only "No key" for a model without a stored key; whether the env fallback would apply could not be seen anywhere.

### Login-shell import (desktop)

On macOS/Linux GUI launches (no `TERM`), before anything reads `process.env`, the shell runs the user's `$SHELL` once as an interactive login shell (`-i -l -c`) around a sentinel-framed `env -0` dump, then merges the result **fill-missing-only** into `process.env`. The embedded server — and with it the bash tool — inherits the merged environment at fork time; core is untouched.

- **`env -0` + sentinel frame** rather than a Node re-entry probe: NUL separation keeps newline-containing values intact, the frame cuts away rc-file noise (the close is the *next* sentinel, so `zlogout` output after it is ignored), and nothing depends on the `RunAsNode` fuse staying enabled if packaging hardens later.
- **Fill-missing-only**: an explicitly provided variable always wins. `PATH` is the one merged key — login-shell entries first, launch-only entries appended, deduplicated.
- Never imported: `_`, `SHLVL`, `PWD`, `OLDPWD`, `TERM`, `ELECTRON_RUN_AS_NODE`.
- Best-effort: 5s hard timeout with an immediate settle on kill, 4MB output cap, exit code ignored. Any failure logs one line and changes nothing. Escape hatch `PENGUIN_NO_LOGIN_SHELL_ENV`.

### Env-fallback visibility (server + web)

**First-party gate** (`envFallbackFirstParty`, exported for tests): the preview is offered only when the entry points at the provider's own official endpoint — provider is a built-in **vendor** group (not `custom`, no `gatewayBaseUrl`), and the entry keeps the catalog shape (a preset's `client_type`/`base_url` match the catalog's own values; an off-catalog vendor id qualifies only as a bare auto-routed entry). Gateway rows resell through generic OpenAI-protocol clients whose fallback is `OPENAI_API_KEY` — the *official OpenAI* variable — and the preview must not encourage steering it to a reseller, so they are excluded along with custom and user-defined groups. `envKey` itself is still reported as before for every routable entry; the gate governs only the preview.

**API**: `envKeyPresent` (earlier revision of this PR) is replaced by `envKeyMasked?: string` — the variable's current value masked by the **same `maskApiKey` rule as stored keys**; presence is implied by the field and the plaintext is never serialized. Read from the server's own `process.env` (on the desktop, the post-import environment).

**Web**: the card keeps its original dot-separated meta text — no new badge visuals: stored key → mask; detected first-party fallback → the variable name (e.g. `OPENAI_API_KEY`); otherwise → "No key". The model dialog renders a detected fallback exactly like a stored key: the masked value in the same slot, "Read from environment variable NAME" in the created-at position, **no clear control** (an env variable cannot be cleared from here); typing and saving a key follows the existing stored-key-over-env priority and the display switches to the stored form.

State matrix (dialog credential slot): stored key → mask + created-at + clear checkbox · env detected (first-party, no stored) → mask + "Read from environment variable" · neither → nothing (placeholder hints unchanged).

## Verification

- `packages/server` `models.test.ts`: 22 pass, including the reworked case — first-party preset masked (exact `maskApiKey` output), empty variable → no preview, off-catalog vendor auto-route entry masked, vendor row re-pointed at `openai` protocol excluded, gateway and custom rows excluded even with `OPENAI_API_KEY` set, and the raw response text never contains either plaintext value.
- `packages/desktop`: 75 tests pass (login-shell import, unchanged this round). `packages/web`: full suite passes (toRow passthrough case updated); `packages/docs`: passes.
- Branch rebased onto current `main`; full `pnpm build` + `pnpm typecheck` clean; `pnpm format` + `format:check` clean; `git diff --check` clean. No `envKeyPresent` references remain.
- Not testable here: a real macOS/Linux GUI launch (headless box); the launch gate, probe, parse, and merge paths are unit-tested individually.


hiyouga:
## Review

The desktop half is sound: the probe cannot be injected into, cannot be hung past 5s, cannot clobber a variable the launch already set, and never logs a value. I found one real correctness bug in it (a multi-byte value straddling a pipe read came back as replacement characters) and one missing exclusion class (`PENGUIN_WEB_DIST` / `PENGUIN_CLI_ENTRY` pick *which code* the embedded server serves and self-updates through, and the fork overrides neither), both fixed below. The server half never serializes the plaintext and masks by the same `maskApiKey` rule as stored keys, with the first-party gate doing what its comment claims. **The code is merge-ready; the PR is not, on the repo's own rule** — this changes three behaviours that `design/specs/06-PROTOTYPE.md` describes (the card's key-status slot, the credential hint, and the desktop shell's startup) and ships no design update. `design/` is a shared read-only symlink in my worktree, so that part is yours: land the design lines, then this merges.

## Changed on request

All three are in `packages/web/src/features/models/models-page.tsx`, pushed as `f5814b06`.

**(a) The card-slot line inside the dialog drops the variable name.** `readFromEnv` is a plain string in both dictionaries now, not a function (`strings-en.ts:691`, `strings.ts:676`); nothing else called it. Rendered at `models-page.tsx:2003` next to `form.envKeyMasked`, which is what identifies the key.

| | before | after |
| --- | --- | --- |
| en | `Read from environment variable ANTHROPIC_API_KEY` | `Read from environment variable` |
| zh | `读取自环境变量 ANTHROPIC_API_KEY` | `读取自环境变量` |

**(b) The list row shows the abbreviation, not the variable name** (`models-page.tsx:1082`). The key-status column now falls through to `row.envKeyMasked`, so an env-backed row reads exactly like a configured one; `S.models.noKey` still covers the genuinely unconfigured case. Same text in both languages, since it is a masked key:

| | before | after |
| --- | --- | --- |
| en / zh | `ANTHROPIC_API_KEY` | `sk-a…3456` |

**(c) The hint no longer promises a variable that is not there.** Both sites were gated on *a variable name being known*, which says nothing about whether it is set. The signal is `envKeyMasked`, which the server emits only when the variable actually holds a value; because an environment is process-wide, one row carrying it settles the question for every entry reading the same variable. `detectedEnvKeys()` (`models-page.tsx:360`, exported and unit-tested) collects them, memoised once per load at `:582` and handed to both dialogs as a prop.

- Model dialog (`:1771`–`:1786`): `envHintKey` = no stored key **and** the live-resolved variable is one of the detected set. Stored key → keep-hint as before. Detected → env hint. Neither → **no placeholder**, rather than the previous fall-through to "Leave empty to keep the current key", which is not true of a field that has no current key.
- Group-wide key dialog (`:2441`): it does have access to the signal — the call site already derives `count` from `rows`, so the same set goes in as a prop. Placeholder only when detected, otherwise none (the keep-hint is meaningless in a bulk-set dialog).

| | before (always) | after (variable unset) |
| --- | --- | --- |
| en | `Leave empty to use the ANTHROPIC_API_KEY env var` | no placeholder |
| zh | `留空则使用环境变量 ANTHROPIC_API_KEY` | 无 placeholder |

The default-endpoint note below (`:1788`) is documented as appearing only when the env hint appears, and my re-gating would have desynchronised it — it was keyed off `liveEnvKey` directly. It now keys off `envHintKey`, so the two are in lockstep by construction rather than by two conditions that have to be kept equal.

`packages/web/test/models-input.test.ts` gained a `detectedEnvKeys` case (a variable with a masked preview, a variable known by name only, and a row with neither); the PR's `toRow` passthrough case is unchanged.

## Fixed

**`eaaa4ce1` — a value whose UTF-8 bytes straddle a read boundary was corrupted.** `resolveLoginShellEnv` accumulated `stdout += String(chunk)`, decoding each Buffer independently, so any sequence split by a pipe-sized read became replacement characters. This is not theoretical: a fake shell writing 150 KB in one `write` reproduces it every run (3 replacement characters, value length off by 2). Any environment large enough to need a second read corrupts whichever value spans the cut — a CJK path, a prompt string, `LS_COLORS`-sized junk followed by anything non-ASCII. Fixed with `setEncoding("utf8")` (`login-shell-env.ts:198`), which carries the partial sequence across; the regression test at `test/login-shell-env.test.ts` fails without the one-line change.

Same commit: `settle` now drops the stdout pipe (`:183`). The comment above it already explains that a killed shell's children inherit that pipe and hold `close` open — but the `data` listener stayed attached after the promise resolved, so those children could keep growing a string nobody would ever read.

**`1256a578` — the import could retarget the app's own code.** The merge is fill-missing-only, so `HOME`, `USER`, `TMPDIR` and friends are safe by construction (a GUI session always sets them). Two variables are not: `PENGUIN_WEB_DIST` and `PENGUIN_CLI_ENTRY` are normally unset in a GUI session, so they *are* filled — and `server-process.ts:117` spreads `process.env` into the utilityProcess fork while overriding only `PENGUIN_HOME` / `HOST` / `PORT` / `PENGUIN_DESKTOP_TOKEN` / `PENGUIN_PORT_FILE`. `config.ts:134` then reads `PENGUIN_WEB_DIST` for the Web App it serves, and `version.ts:138` reads `PENGUIN_CLI_ENTRY` for the entry it self-updates through. A line in a developer's `.zshrc` written for a repo checkout would silently reach into the installed app they only double-clicked. Both added to `EXCLUDED_KEYS` with a test. `PENGUIN_HOME` deliberately still travels — that selects *data*, which is a choice a user can mean to make once for both the CLI and the app, and it is the same code/data line `core`'s `STRIPPED_ENV_KEYS` already draws (`session-manager.ts:67-78`).

**`a78dbb04` — stale comment.** The four lines above `...(envKey ? { envKey } : {})` in `project-config-service.ts` survived the `envKeyPresent` → `envKeyMasked` rework: they describe a presence flag ("presence only, never the value") while sitting on `envKey`, which is the variable's *name*, and they repeat what the comment on the `envKeyMasked` computation says 10 lines above. Deleted.

Docs and the changelog pair were updated in step with (a)–(c) and the new exclusions — `quickstart-desktop.{en,zh}.md` said the card shows the variable name, and the changelog entry described the old card text and the old exclusion list.

## For you to decide

1. **Design docs (blocking, by `CLAUDE.md`'s sync rule).** Three places in `design/specs/06-PROTOTYPE.md` now disagree with the code, and I did not touch the shared symlink:
   - `:119` — the card meta line is specified as "API key 状态：已配置显示掩码、未配置写明未配置". It now has a third state (detected env fallback → that variable's value, masked).
   - `:127` — "api_key 留空则按 PRN-021 解析出的客户端环境变量兜底（弹窗提示随解析结果实时变化）；页面只显示掩码后的 key 与创建时间". The hint no longer follows the resolution alone but the resolution *plus a detected value*, and the dialog now shows the variable's masked value with "read from environment variable" in the created-at slot.
   - `:166`+ (桌面端原型) — the shell's startup sequence is specified there (port announcement, desktop login, …) and gains a step: the login-shell import, its gate, and `PENGUIN_NO_LOGIN_SHELL_ENV`.
2. **`NODE_OPTIONS` is imported.** I left it, and want a ruling. Against excluding: a terminal launch already hands it to the forked server today, so importing it only makes a GUI launch match, which is the whole point of the feature — and agent commands running `node` arguably *should* see the user's flags. For excluding: `packages/server/src/platform/terminal/session.ts:126` already deletes it before handing an environment to the pty, on the reasoning that "a pty is not that server and these would point the shell (and anything it starts) at the wrong runtime" — the utilityProcess fork is not that server either.
3. **Two chat surfaces still treat an env-backed model as keyless**, which now reads oddly next to a Models page that shows it as configured. `model-grouping.ts:113` `hasConfiguredKey` is `!!m.credential?.apiKeyMasked`; it drives the dropdown's key filter (`model-select.tsx:183`) and, separately, `chat-page.tsx:934` pops the once-ever credential guide when the default model has no *stored* key. A desktop user with `ANTHROPIC_API_KEY` exported — the exact #351 user — is still told to configure a credential. The rule was recorded with its reason in `design/specs/08-CHANGELOG.md:35` ("envKey 仅是变量名不算"); this PR is what changes that premise, so it is a design decision, not a cleanup. Note the gate is first-party-only, so a gateway model would stay keyless either way.
4. **The group-key dialog can still name `OPENAI_API_KEY` for a gateway or user-defined group** when that variable is set (`userProviderInfo` synthesizes `envKey: "OPENAI_API_KEY"`). Pre-existing, and my change only narrows it — but it is the same "don't steer the official OpenAI variable at a reseller" tension the server's first-party gate is built around. Suppressing it for non-vendor groups is a one-line product call.
5. **A 5s startup stall is possible.** `main.ts:257` awaits the import before `boot()`, so on a slow interactive rc file (oh-my-zsh + nvm is routinely 1–3s) nothing is on screen until it returns. Options: accept; drop the timeout; or create the window first and only make the server fork wait.
6. **`envFallbackFirstParty` is exported and nothing imports it** — the PR description says "exported for tests", and no test does. Either drop the `export` or pin the predicate directly; `model-probe.test.ts:18` shows the repo already imports pure helpers straight out of this service.
7. **`TERM` as the GUI gate** is a heuristic worth stating out loud: where it produces a false negative (a Linux session that exports `TERM`, e.g. `startx` from a tty) the feature silently does nothing and logs nothing. That case usually *did* come from a login shell, so the heuristic is defensible; `process.stdout.isTTY` would be the more direct signal if you ever see a report.

## Checked and clean

**The shell-env import**, exercised path by path: the shell is `$SHELL` when absolute, else `/bin/zsh` on darwin and `/bin/bash` elsewhere (`login-shell-env.ts:85`); a hostile `$SHELL` buys nothing, because `spawn` is called with an argv array and no `shell: true`, and `PROBE_COMMAND` (`:32`) is a static string with the sentinel interpolated at module load, not user data — a non-shell binary just fails the frame and returns null. A shell that prompts gets EOF (`stdio: ["ignore", "pipe", "ignore"]`, `:170`) and its stderr noise is discarded; a shell that prints a banner or a `zlogout` line is cut away by the frame, whose close is the *next* sentinel rather than the last; a shell that hangs is SIGKILLed at 5s and the promise settles immediately rather than waiting on `close` that leaked grandchildren would hold open; a shell that exits non-zero is still believed, deliberately; output above 4 MB is dropped. Every one of those has a test against a fake shell, and every failure path logs one line naming only the shell path and returns 0 with the environment untouched.

**Secrets.** Nothing logs a value: the only two log lines are a count plus the shell path (`:232`, `:239`), and nothing in the tree serializes `process.env` into a log or a response — all six sites that read it wholesale (`...process.env` / `Object.entries(process.env)`) are building a child process's environment. On the server, the plaintext never leaves: `envKeyMasked` is computed through the same `maskApiKey` (`project-config-service.ts:80`, `***` under 13 chars) that stored keys go through, an empty variable yields no field, and the PR's test asserts the raw response text contains neither plaintext. It is one code path (`getModels`, `:633`) that both `GET` and the `PUT` response go through (`updateModels` ends in `return this.getModels(...)`, `:889`). The first-party gate at `:101` does match its comment: `providerInfo` must resolve, `custom` and any `gatewayBaseUrl` group are out, a catalog preset must still match the catalog's own `client_type` and `base_url`, and an off-catalog vendor entry qualifies only when it pins neither.

**Conventions.** No CJK outside the zh dictionary and the `.zh.md` files; no design-doc citations in comments; the changelog pair mirrors section for section with the metadata block in order and no `## Why` / `## Verification` / `## Risks`; `#351` verified as an issue, not a PR; docs bilingual. The PR also dropped a blank line above `TONE_CLASS` for no reason — restored.

**Ran** (all green, on this branch at `1256a578`): `pnpm --filter @prismshadow/penguin-desktop test` (77), `pnpm --filter @prismshadow/penguin-web test` (83 files / 1062), server `models.test.ts` (22) plus `isolation.test.ts`, `project-config-cache.test.ts`, `model-probe.test.ts`, `packages/docs` (53), `tsc --noEmit` for web / server / desktop / docs, `pnpm format` → `format:check`, `git diff --check`. `git merge-tree` against current `origin/main` (which now carries #378's tone tokens, #381's `isolate: false` and #385) is conflict-free, and the new markup adds no `InfoPopover`, so `disclosure-anchor.test.ts` is unaffected.

Not verifiable here: a real macOS/Linux GUI launch.

PR #371, opened 2026-08-20
feat(web,server): usage charts as smooth time series with range and precision controls

hiyouga:
## What this changes

Rebuilds the cost center around one shared time axis: every chart is a time series over a user-selected range and precision, nothing scrolls, and the errors panel is bounded by pagination alone. (Second revision — merges the calls and success-rate charts into one combo, adds trailing windows with minute buckets, and reverts smoothing except for the hit-rate curve.)

**Controls** — a date-range preset (trailing **last hour** / **last 24 hours**, calendar last 7 / 30 / 90 days — **7 days is the default** — or custom from/to date inputs) plus a precision selector. The preset drives the offered precisions and an invalidated selection snaps to the preset's default:

| Preset | Precisions (default bold) |
|---|---|
| Last hour | **minute** |
| Last 24 hours | **hour** |
| Last 7 days | hour, **day** |
| Last 30 / 90 days | **day**, week |
| Custom | by range length: hour (≤14d), day (≤366d), week (≥14d), month (≥60d) |

**Requests + success rate combo** (full-width, replaces the separate calls and success-rate charts, styled after the OpenAI usage panel): request bars from the bottom on the left axis; the success-rate polyline on top against its own right-hand 0–100% axis, with a dot at the line's end; the legend underneath. The card header holds a **by-Agent / by-Model dimension toggle** and an **entity dropdown**:

- **All**: bars stack by entity — top 4 entities keep their own CVD-checked `SERIES_COLORS` hue, the tail folds into a neutral gray "other" — under the overall success-rate line.
- **One entity**: bars are that entity's requests (sky), the line is that entity's own success rate — which restores the labeled per-model success rate.
- The tooltip reports the bucket's per-entity request counts **and** the success rate together.

**Token chart**: keeps the three-segment stacked bars; the cache hit rate is a **dashed smooth curve in front of the bars** on its own right 0–100% axis, now **continuous** — a bucket with no cache traffic counts as 0, never a gap. Hovering anywhere in a column (the curve included) reports the whole bucket: all three Token counts plus the hit rate; hovering a segment additionally highlights it. Bars fit the card (`fitBarWidth`), no horizontal scroll.

**Cost chart**: back to a straight polyline with a dot on every point (area fill kept), per the original design; dots collapse to hover-only past daily density. Empty or unpriced buckets display **0**, never a dash — the zero-not-dash rule applies across every chart tooltip.

The hit-rate curve is the only remaining consumer of monotone-cubic smoothing (Fritsch–Carlson, overshoot-proof); the unused smooth-area path and the null-gap splitting were deleted with their callers.

## Server API (additive)

`GET /api/projects/:p/usage`:

- `granularity` gains `minute`; `fromTs`/`toTs` (ISO timestamps, given together, normalized server-side) bound a trailing window down to instants — required for `minute`, and they refine every range-scoped aggregate (summary total, groups, series, entity series, status rows; the calendar-fixed today/last-7d cards stay date-only). Minute/hour buckets are enumerated between the instants (DST-safe), other precisions stay date-driven.
- The response's `byAgentSeries` entries gain per-bucket `completed`/`denominator` arrays, and `byModelSeries` joins them (same shape, `(provider, modelId)` identity; each ignores its own dimension's filter so all entities stay offered). `series`, `trend`, and all pre-existing fields are unchanged.
- The 500-bucket cap still rejects oversized range × precision combinations with 400, with early-bail enumeration.

## Notes for review

- The dual-axis combo and token charts are the explicitly requested composition (reference: OpenAI usage panel); the right-axis machinery stays opt-in per chart.
- Idle buckets keep the site-wide `successRate(0,0)=1` convention (a flat 100% line, tooltip shows the counts), so trailing windows with sparse traffic don't render as failures.
- Branch was rebased onto latest main (picking up the #378 status-tone/icon refactor in `errors-panel.tsx`) and force-pushed with `--force-with-lease`.
- No screenshots: rendering is validated through path-math and folding unit tests; happy to add screenshots on request.

## Verification

- `pnpm --filter @prismshadow/penguin-server exec vitest run test/usage.test.ts` — 24 passed (series zero-fill, week/month keying, local-clock hour buckets, **minute window bounds + range refinement**, **byModelSeries semantics**, per-entity success counts, defaults, oversize/missing-bounds rejection).
- `pnpm --filter @prismshadow/penguin-web test` — 81 files / 1028 tests passed (`usage-controls.test.ts` covers the preset×precision matrix, trailing windows, entity folding, zero-continuous hit rate; `usage-charts.test.ts` covers monotone no-overshoot, `fitBarWidth`, shared-x right-axis geoms).
- `pnpm --filter @prismshadow/penguin-docs test` — 53 passed (bilingual server-api / web-app docs updated).
- `pnpm -r build`, `pnpm typecheck`, `pnpm format` + `pnpm format:check`, `git diff --check` — clean.


hiyouga:
Reviewed at `b08b4514`, then reworked the requests chart as asked and pushed three commits. The time-series work underneath is sound — bucket keys agree byte for byte between `BUCKET_EXPRS` and the enumerators, week/month boundary keys line up with a mid-period `from`, zero-fill is index-exact, the 500-bucket cap stops enumeration one key past the limit rather than materializing first, and `ts` is compared as one fixed-width ISO spelling on both sides — with one real hole: `enumerateTsBuckets` only collapsed *adjacent* duplicate keys, so a minute-precision window crossing a DST fall-back emitted the replayed local hour twice. Fixed with a regression test. The other things I changed are one duplicated query per request, an axis that labelled whole requests `0 / 1 / 3 / 4 / 5`, a dot-density guard that counted points instead of measuring them, and three helpers this PR left without a consumer.

## Changed on request

Two cards side by side (`usage-page.tsx:381` and `:388`), `Requests & success rate by agent` and `… by model`. Each stacks a bucket's requests one segment per entity on the left count axis and draws every entity's success rate as its own line on the right-hand 0–100% axis in that entity's hue. `RequestsControls` (dimension toggle + entity dropdown) is gone, along with the `reqDim` / `reqEntity` page state and the `""` = aggregate default. Both dimensions are on screen at once, without scrolling, at `lg` and up.

**How many entities before it turns to mush — four, and the chart says so.** `MAX_NAMED_SERIES` (`usage-controls.ts:149`) is now literally `SERIES_COLORS.length`, so the cap is the palette rather than a taste setting: past it a fifth entity would repeat a hue in both the stack and the lines. The tail folds into one neutral series labelled `Other (N)` — `legendOther: (n) => …` in both catalogs — so the fold is visible in the legend, in the stack, and in the hover bubble. A tail of exactly one keeps its own name (a mystery bucket of one is worse than a fifth label), and the fold sums `completed` / `denominator` too, so the tail's line is the combined rate rather than one entity's.

**Colour.** `seriesColor(i)` by rank within each card, plus `NEUTRAL_SERIES` for the tail — no new palette. Agent ids and model names are disjoint namespaces, so there is no entity that appears in both cards to keep stable; what is stable is that rank drives hue and the server sorts by total requests descending, so a given entity keeps its colour as long as its rank holds. Identity never rests on colour alone: the legend and the hover bubble both name every entity, which is the backstop `SERIES_COLORS`' own doc requires for the sky/amber pair.

**Idle buckets break the line instead of scoring 100%.** `rateSeries` (`usage-controls.ts:190`) now returns `null` where `denominator` is 0. With one aggregate line, the site's `successRate(0,0)=1` convention was defensible. With five, it drew five confident flat 100% lines across every interval those entities never ran in. A gap says "no rated request here"; the bubble shows `—` in the same buckets, so the dash and the hole say the same thing. A run of one bucket strokes nothing, so it renders as a dot (`usage-charts.tsx:261`).

**Two scales in one card.** Unchanged from your combo: counts on the left, `rightAxis` 0/50/100% on the right, both geoms sharing `n` / `w` / `PAD_R_AXIS` so `x()` is identical. The rate lines are **dashed**, matching the Token chart's hit-rate curve — I tried solid first and the lines read as bar edges, because here a line shares its entity's bar colour and shape is the only separator left.

**Isolating one entity.** The dropdown is gone, so the legend took over: hovering an item leaves that entity's bars and line at full strength and fades the rest — the same `highlight = fade out the rest` convention `TokenLegend` already uses. State is local to `RequestsChart`, not lifted, since the legend sits under the chart rather than in the card header.

**Hover.** One row per drawn entity — swatch, label, request count, rate — then a muted `Total` row computed from the drawn series, so the number matches the stack height you are looking at rather than `series.requests` (which honours both filters while the entity series do not).

I drove the built server against a scratch `PENGUIN_HOME` with 160 seeded records over six agents and six models, and checked 7d/daily, 24h/hourly, 30d/daily, last-hour-empty, light and dark. Server claim verified rather than assumed: `byAgentSeries` / `byModelSeries` are index-for-index with `series` and unaffected by their own dimension's filter, so this needed nothing from the server.

## Fixed

**`20ccd2c2`** (web)

- `enumerateTsBuckets` aside, the loudest visual defect: the requests axis is quartered by `ChartFrame`, so a max of 5 labelled its gridlines `0 / 1 / 3 / 4 / 5` — skipping 2 and implying even spacing that isn't there. `niceCountMax` (`usage-charts.tsx:69`) rounds the top up to a multiple of 4. The 24-hour view now reads `0 / 2 / 4 / 6 / 8`.
- `trend-chart.tsx:44`: `everyDot = series.length <= 92` was written for daily density, but 61 minute buckets also pass it, and in a half-width card they sit ~7px apart — the cost line rendered as a rope of touching dots. Now gated on `geom.step >= MIN_DOT_STEP`, which is what the comment already claimed.
- `chart-geom.ts` `successRate`, `chart-svg.tsx` `scrollToEnd`, and `category-colors.ts` `CATEGORY_COLORS` / `categoryColor` / `CategoryColor` had no production consumer left once the pie and the success-rate bars went — `successRate` was reachable only from its own test, `scrollToEnd` from nothing at all. Removed, with `category-colors.ts`'s header (which still described the deleted pie chart) rewritten to say what the file is now: the place a categorical hue is spelled out, explicitly not `tone.ts`.
- `lineSegments` was about to exist twice — `benchmark-metrics.ts` had it, and the rate lines need it. Moved to `chart-geom.ts` (which `benchmark-page.tsx` already imports `makeRangeGeom` from) with a `segmentPath` companion; benchmark's import and its test block moved with it. Same for the `text-gray-400 dark:text-gray-500` "the rest" swatch, which was written inline in two features and is now `NEUTRAL_SERIES`.
- Strings: `chartRequests` / `dimLabel` / `dimByAgent` / `dimByModel` / `entityLabel` / `callsOther` out, `chartRequestsByAgent` / `chartRequestsByModel` / `legendOther(n)` / `bucketTotal` in — both catalogs, same order, 49 keys each. Changelog pair and both `web-app` docs rewritten for the two-card shape.

**`4217189f`** (server)

- `enumerateTsBuckets` (`dates.ts:67`) collapsed only adjacent duplicate keys. During a DST fall-back the replay of `01:00`–`01:59` comes an hour later, never adjacent, so a minute-precision window across that boundary emitted every one of those keys twice. `strftime('%Y-%m-%dT%H:%M', ts, 'localtime')` folds both instants into a single key, so `foldSeries`'s `keys.map(k => byKey.get(k)!)` handed the same bucket object to two chart points, and `foldAgentSeries`'s `Map` kept only the later index — one of the two points would have drawn empty. Both enumerators dedupe by key now, and the doc comment that claimed the collapse already happened is corrected. The new test fails on the old code and passes on the new one; it pins spring-forward too (23 hour keys, no invented `02:00`).
- `byModelSeries` dropped only the model filter, so with no model filter set it re-ran exactly the query `series` had already run — the widest statement on the route, `GROUP BY bucket, provider, model_id`. `usage-service.ts:207` reuses those rows instead. That is the common case: the Web App sends a model filter only when the user picks one.
- `foldSeries`'s comment listed `day/week/month` and `hour` but not `minute`, which the same defensive drop now applies to.

**`7dacaa9c`** (server test) — `byAgentSeries` alignment was pinned at day granularity only; the minute window is where the skeleton comes from a different enumerator, so that is the case worth pinning. Added to the existing minute test.

## For you to decide

1. **Five response fields now have no reader.** `groupBy`, `groups`, `trend`, `byAgent` and `success` all return 0 hits across `packages/web/src` — `trend`, `byAgent` and `success` became dead in *this* PR when the pie and the success-rate bars were replaced. Behind them, `usage-service.ts:156`, `:198` and `:217` run three repo queries per request for nobody. Options: leave them (public API, and `/usage` has non-browser callers you may know about); drop fields and queries in a follow-up with a `Breaking` changelog entry; or gate them behind a query parameter so the Web App stops paying for them. I did not touch this — it is an API-shape call.
2. **The own-dimension filter exemption reads differently without the dropdown.** `byAgentSeries` ignores `agentId` and `byModelSeries` ignores the model pair (`types.ts`, documented). That existed so the dropdown could offer every entity. Now, filtering the page to one agent narrows the Token and cost charts while "Requests by agent" still shows all six. Options: keep and say so in the card when a filter is active; honour the filter (the card then collapses to a single-entity stack, which is what you asked me to move away from); or leave it, on the grounds that a *breakdown* chart's job is the breakdown.
3. **`granularity=hour` with no `from`/`to` is a 400.** The series window defaults to the trend's 30 days, which is 720 hour buckets against the 500 cap (`usage-service.ts:177`, `:180`). The Web App never reaches it (its `7d` preset offers hour = 168 buckets), but a direct API caller asking for hourly without a range gets rejected rather than served a shorter window. Either is defensible; only one is documented.
4. **Legend isolation is pointer-only.** `onMouseEnter` / `onMouseLeave`, same as `TokenLegend`. Nothing isolates an entity by keyboard or on touch. Click-to-pin would fix both and is a small addition, but it changes an established interaction across two legends, so I left it.
5. **The rate axis is fixed 0–100%.** Real success rates cluster at 85–100%, so the lines pile against the top gridline and several overlap exactly at 100%. A zoomed rate axis would separate them and would also make a 2-point difference look like a cliff. I kept the honest scale; flag it if you would rather trade.
6. **Errors first page went 20 → 10** (`ERROR_RECENT_N`). It is in the changelog, so it is deliberate, but it is an unrelated behaviour change riding in a charts PR — worth a conscious yes rather than a silent one.
7. **Cost trend prints `$0.0000` on four gridlines** when nothing in the range is priced. Visible in every screenshot I took. Pre-existing to the cost chart and outside the two-card scope, but it looks broken.

## Checked and clean

- **Bucketing.** `BUCKET_EXPRS` and `enumerateBuckets` / `enumerateTsBuckets` produce the same strings: `date(date,'-6 days','weekday 1')` vs `localWeekStart(from)` agree when `from` is mid-week; `substr(date,1,7)` vs `localMonthKey` agree when `from` is mid-month; the hour path steps real timestamps rather than assuming 24 per day. Empty buckets exist as zero points with `cost: null`. A range with no data yields an all-zero series and, in the app, both requests cards render the empty state (confirmed on the last-hour preset).
- **Timezone.** SQLite's `'localtime'` and JS `getHours()` resolve in the same process, and `date` was already recorded local, so the daily key and the hour key cannot disagree by zone. DST is now pinned in both directions.
- **`ts` comparison.** Rows write `now.toISOString()` (`usage-recorder.ts:80`); the route normalizes through `new Date(ms).toISOString()`, so `ts >= :fromTs` compares one fixed-width spelling against the same spelling. `fromTs`/`toTs` are validated as a pair and ordered, and `minute` without them is rejected in the service, not only at the route.
- **Boundedness.** Bucket count is capped at 500 with enumeration stopping one key past it, so an oversized range never materializes an array before the 400. Row volume is bounded by traffic in the range rather than by the cap — a 500-day daily range still scans every record in it — but that is the same shape as the pre-existing `trend`/`groups` queries and `idx_usage_project_date` covers the scan. The trailing presets always send the one or two dates their instants span, so the `ts` refinement filters rows the date index already narrowed.
- **Conventions.** Every new string is in both `strings-en.ts` and `strings.ts`, same keys in the same order (49 apiece, checked mechanically — note there is no test enforcing this). No Chinese outside the zh catalog, the `.zh.md` pair and the pre-existing `中文` label. No design-doc citations. Changelog is a mirrored bilingual pair in `unreleased/` with the metadata block in `Date`/`Type`/`Scope`/`PR` order, no index file, no `## Why` / `## Verification` / `## Risks`. Docs updated in both languages.
- **Prose.** Grepped the added lines for stack/review vantage, change narration, dead citations and hedged planning residue — clean. The two comments that no longer matched their code (`dates.ts`'s "collapse to one key", `trend-chart.tsx`'s rope claim) are fixed above.
- **Verification.** `penguin-web` 1032 tests / 81 files, `penguin-server` 801 tests / 78 files (full suite, solo), `penguin-docs` 53, `pnpm typecheck` across the workspace, `pnpm format` → `format:check`, `git diff --check`. Screenshots taken against a built server on a scratch data root at 7d/daily, 24h/hourly, 30d/daily and the empty last-hour case, light and dark.

hiyouga:
Acting on the ruling for the first open item — 没有前端消费的应该整理一下. Pushed as `b58550c5`. Three of the five candidates were dead and are gone with their queries; two were not dead and stay.

## Kept — `groupBy` and `groups`

Not dead. `chat-page.tsx:789` fetches `/usage?groupBy=session&agentId=…` and reads `res.groups.find(g => g.key === sessionId)` for a session's running cost and token buckets; `header-stats.ts:40` documents the row subset it consumes. My earlier sweep matched `data.groups` and missed `res.groups`, which is exactly the reason to grep the field name rather than the variable. `groupBy` is also a documented query parameter in both `server-api` docs. `UsageGroupRow`, `UsageGroupBy`, `groupsByModel` and `foldGroups` all stay.

## Removed — `trend`, `byAgent`, `success`

Evidence, swept across `packages/*/src`, `packages/*/test`, `packages/*/e2e`, `packages/docs/content/**`, `examples/` and `scripts/`, on the field names and on the wire strings:

| Field | Only readers found | Query behind it |
| --- | --- | --- |
| `trend` | `usage.test.ts:219–220` (server's own tests) | `groupsByModel(projectId, "date", win(trendFrom))` — a fixed 30-day aggregate |
| `byAgent` | none at all, tests included | `groupsByModel(projectId, "agent", agentFree)` |
| `success` | `usage.test.ts:233, 252, 261, 272, 276` (server's own tests) | `statusByModel` |

The CLI has no usage client of its own — `packages/cli` does not reference any of them. `server-api.en.md` / `.zh.md` document `/usage` by **query parameter only** and never enumerate the response shape, so no docs row named these fields; the only prose that described the charts they fed is in `web-app.*.md` and was already rewritten for the two-card layout. Desktop runs the unchanged server and web app.

Each removal took the whole chain: the `UsageResponse` field, its API type (`UsageTrendPoint`, `UsageAgentCount`, `UsageSuccessRate`), the service code that filled it (`foldTrend`, the `byAgentMap` fold, the `statusRows` sort), and the repo method behind it (`statusByModel` and its `UsageStatusCount`). The two doc comments on `byAgentSeries` / `byModelSeries` that explained themselves as "like `byAgent`" / "like `success`" were rewritten, since those references died with the fields. `S.usage.successAborted` — the "Aborted (excluded)" label the removed success-rate bars used — went from both catalogs.

**Response shape now:** `summary`, `groupBy`, `groups`, `granularity`, `series`, `byAgentSeries`, `byModelSeries`, `errors`, `agentIds`, `models`. Confirmed on the wire against a running server, not just in the types.

**One thing genuinely lost:** `success` carried a per-status failure breakdown — `aborted`, `failed`, `timeout`, `malformed` — that `byModelSeries` does not reproduce, since it keeps only `completed` and the denominator. Nothing displayed it, and failure diagnosis is what the `errors` panel is for (it records source and code per failure), so I removed it rather than inventing somewhere to put it. It is named in the Compatibility section so the loss is on the record rather than discovered later.

## Breaking: yes

The repo's precedent decides this. `changelog/0.2.3/2026-08-19-compact-after-restart.md` marks `Breaking: yes` for an HTTP-only change with no on-disk component — two `POST /compact` refusals moving between error codes. Removing three response fields is strictly larger than relabelling an error code, so treating this as non-breaking would contradict the repo's own bar. `changelog/README.md` defines the field as "only when breaking / `yes — <one line>`" and does not scope it to on-disk state; the SKILL's on-disk framing is about *which strategy* a compatibility break needs, not about what counts as one.

Both entries now carry the field after `PR` (Date / Type / Scope / PR / Breaking order held) and a `## Compatibility` / `## 兼容性` section stating where each number moved: `trend` → `series` (same points, and it defaults to exactly the last 30 days at day precision when no range is given); `byAgent` request counts → sum `byAgentSeries[].requests`, per-Agent token totals → `?groupBy=agent`; `success` → sum `byModelSeries[].completed` and `[].denominator`, with the per-status counts noted as having no replacement. The Details bullet claiming "Existing fields were left unchanged" was false once this landed and is corrected.

## Survivors verified, not assumed

Two things could have moved a number:

- **`trendFrom` was load-bearing beyond `trend`** — `seriesFrom = q.from ?? trendFrom` used it as the default series window. It is now inlined as `localDateMinusDays(this.now(), 29)`, same value; the existing "no from/to serves the last 30 days" test still passes, and the live smoke shows `series.length === 30` for an unranged query.
- **`trendRows` fed the pricing-lookup set.** Dropping it means a Model priced only inside the 30-day window but outside the requested range is no longer looked up. Rather than argue it, I pinned it: a new test inserts such a Model, asserts the lookup callback is called for exactly `["m1"]`, and asserts `summary.today`, `summary.total`, `groups[0]` and `series[0]` costs are all unchanged. It would have failed before this commit.

Tests were rewritten, not dropped. The aborted-is-not-a-model-failure regression — the one guarding a real past bug — now reads `series.completed` / `series.denominator` and `byAgentSeries`, and additionally pins something the old version did not: an interrupt still counts as a `request` while staying out of the denominator. The "any non-completed, non-aborted status counts against the rate" case keeps its three failure statuses and asserts through `series` and `byModelSeries`. The filter-scoping test was reshaped to assert the contrast the series suite did not cover — a model filter narrows `series` but not `byModelSeries`, an agent filter narrows both.

## Verification

`penguin-server` 802 tests / 78 files (full suite, solo — 801 before, +1 for the pricing-scope test), `penguin-web` 1032 / 81, `penguin-docs` 53, `pnpm -r build`, `pnpm typecheck` across the workspace, `pnpm format` → `format:check`, `git diff --check`. Plus a live check against a running server on a scratch data root: the ten remaining top-level keys, no `trend` / `byAgent` / `success`, and `?groupBy=session` still answering with `groupBy` and `groups` for the chat page's path.

Your other open items are untouched.

hiyouga:
Five more decided changes, pushed as `fdbb1e43`. All of them are visual, so all of them were checked in the running app rather than only in tests.

## Same hovers on every chart

The requests charts now take the affordances `TokenBarChart` already had, and the Token chart got the one it was missing.

Rather than write a parallel version, the machinery moved down. `barSegments` was Token-specific — three fixed keys, hit-band water-filling baked in — and is now a thin wrapper over `stackSegments(geom, values)` in `chart-geom.ts`, which does the stacking and the `MIN_HIT_H` allocation for any number of segments. The requests bars stack one segment per entity through the same function, so a one-request slab is reachable for exactly the reason a sub-pixel `output` segment is. Line hovering is a new shared `LineHits` component in `chart-svg.tsx`: one transparent band per bucket, `LINE_HIT_H` tall, centred on that bucket's point. Per bucket rather than one band along the path, so entering a line reports *which bucket* as well as which series.

All three ways of singling out a series in a requests chart resolve to one `RequestsMark { entity, bucket }` — a bar segment sets both, a rate line or a legend item sets `bucket: null`. One shape means they cannot disagree about what is lit: hovering a segment leaves that rect and its own rate line up, hovering the line or the legend leaves the whole series up. Verified by reading the DOM under the pointer, not by eye alone — on a segment hover exactly one of the 28 bar rects is at opacity 1 and the other 27 are at 0.15; on a Token line hover every bar segment drops to 0.2 while the line stays at 1.

## A missing point is 0

`rateSeries` returns 0 where the denominator is 0 — reversing the gap I introduced, and **not** returning to `successRate(0, 0) = 1`. A bucket an entity did nothing in sits on the axis.

The thing the gap was protecting is still real: 0 now means either "made no request" or "made requests and failed all of them", and those should not read alike. So `idleBuckets` marks the first case, the bubble greys those percentages, and a footnote says `0% where nothing was counted toward the rate` — shown only in buckets that actually have one. The legend needs no change: it names entities and the line shape, never a value.

The cache hit rate already plotted 0 for a bucket with no cache traffic (`hitRateValues` ends in `?? 0`), so it needed nothing; I checked every line series on the page and those two are the only ones with a can-be-absent value — cost is always a number.

What it looks like: entities that run intermittently now dive to the floor and climb back rather than going quiet. Busier than the gaps were, and unambiguous in a way the gaps were not.

## Lines above bars

They already drew after the bars, so the paint order was right; what was missing is that the pointer did not agree. The line hit bands are appended **after** the segment hits in `hitLayer`, so a line answers the pointer where it crosses a tall segment instead of the bar underneath winning. Both charts, same ordering: column hits, then segment hits, then line hits.

## No precision control

Gone from the page. `presetDefaultGranularity(preset, rangeDays(from, to))` is now the single place the choice is made, called where the request is built. `presetGranularities`, `granularityOptions` and `coerceGranularity` went with the control — with nothing to choose there is nothing to offer and nothing to snap — and so did the six `granularity*` strings from both catalogs.

**Every preset was checked against a live server**, driving the real derivation helpers (imported from `usage-controls.ts`, not reimplemented) and asserting the response echoes the granularity asked for:

| range | derived | HTTP | buckets |
| --- | --- | --- | --- |
| last hour | `minute` | 200 | 61 |
| last 24 hours | `hour` | 200 | 25 |
| last 7 days | `day` | 200 | 7 |
| last 30 days | `day` | 200 | 30 |
| last 90 days | `day` | 200 | 90 |
| custom 21d | `day` | 200 | 21 |
| custom 143d | `week` | 200 | 21 |
| custom 233d | `month` | 200 | 8 |
| custom 2425d | `month` | 200 | 80 |
| custom 13382d | `month` | 200 | 440 |

Nothing the UI can produce is rejected. `usage-controls.test.ts` keeps the coverage as a derivation matrix — one precision per preset, the custom tiers at their boundaries (92/93, 190/191), the bucket counts against the 500 cap, and that exactly the two trailing presets derive a precision needing `fromTs`/`toTs`, which are exactly the two that send them.

`minute` stays reachable: the last-hour preset still derives it and still sends window bounds, so neither the minute path nor `enumerateTsBuckets` is orphaned.

I did find the real edge, by pushing until it broke: a custom range past ~41 years derives `month` and exceeds 500 buckets — `1960-01-01 → 2026-08-21` is a 400, `1990-01-01` (440 buckets) is fine. That is unchanged by this PR (month was already the coarsest precision offered, so the old selector could not escape it either), and it is listed below rather than fixed.

## No smoothing

The cache hit rate was the last smoothed line; it draws with `linePath` now, like cost and like the success-rate lines already did (confirmed at the call site rather than assumed). `monotonePath`, `monotoneTangents` and the header paragraph arguing that Fritsch–Carlson tangents keep a 0–100% rate from arcing past its endpoints are all gone — no comment defends a technique the file no longer uses. Proof they were dead first: after the switch, `monotonePath` matched only its own test file across `packages/*/src` and `packages/*/test`.

**Dots:** I did not put them on the rate lines. The cost chart's dots earn their place on one line with room between points; five lines that mostly sit between 85% and 100% would smear dots exactly where the lines already overlap, and the bubble is where per-bucket values get read. The Token hit-rate line keeps its hovered-point dot, which is the same thing the cost chart falls back to when its cells get tight.

**`lineSegments` survives** — `benchmark-page.tsx:188` still splits score series on gaps, a real production consumer. `segmentPath` would have been orphaned by the switch, so instead of deleting it I pointed benchmark's hand-rolled `M`/`L` join at it: same output, one representation, and the helper keeps a consumer.

## Plus the dead-API cleanup (`b58550c5`)

Covered in [the previous comment](https://github.com/Prism-Shadow/penguin-harness/pull/371#issuecomment-5372233829): `trend`, `byAgent` and `success` removed with their three queries; `groupBy`/`groups` kept because `chat-page.tsx:789` reads them; `Breaking: yes` with a `## Compatibility` section, on the precedent of `changelog/0.2.3/2026-08-19-compact-after-restart.md` marking an HTTP-only error-code change breaking.

## Still open, unchanged

Your earlier queue stands, minus the one you ruled on. New from this round:

- **A custom range beyond ~41 years is a 400.** Month is the coarsest bucket, and 500 of them is 41 years. Options: clamp the date inputs, add a coarser bucket (year), or leave it — an unreachable range in practice, and it fails loudly rather than silently.
- **Line hover is pointer-only**, like the legend hover before it. Keyboard and touch reach neither.

## Verification

`penguin-web` 1028 tests / 81 files, `penguin-server` usage suite 28, `penguin-docs` 53, `pnpm typecheck` across the workspace, `pnpm -r build`, `pnpm format` → `format:check`, `git diff --check`. In the app on a scratch data root with 161 seeded records over six agents and six models: the 2x2 grid with three controls and no precision selector, segment hover, line hover on both chart kinds, the greyed 0% cells with their footnote, and the straight unsmoothed lines above the bars.

hiyouga:
Four decided changes, pushed as `3dcf8571`. All four are visual, so all four were checked in the running app on seeded data with genuinely empty buckets, not only in tests.

## One stroke width

`DATA_STROKE_W = 2` lives in `chart-svg.tsx` and is now the only weight any data line is drawn at. The three were the cost line at 2 (your reference), the cache-hit-rate curve at 2, and the success-rate lines at 1.5. Making the odd one out match would have left three literals that happen to agree, so the constant is what they all read now — a fourth line added later inherits the weight instead of picking one.

`chart-svg.tsx:237` / `:273` are untouched, and the new break marks join them at 1: the constant's comment says chrome stays at 1 and is deliberately not this.

Read off the live DOM: all six data paths report `stroke-width="2"` — five dashed (two rate lines per requests chart, plus the hit-rate curve) and one solid (cost).

## A null rate plots 100, and the table prints a dash

`rateSeries` and `hitRateValues` both return `(number | null)[]` now. Null means the bucket has no rate at all: the entity made no request in it, or there was no cache traffic. Two readings come off that one array.

- **Drawing.** `plotRates` lifts nulls to `NO_RATE_PLOT = 100` for the path and for the hover bands, so the stroke crosses the interval at the one height that invents no failure.
- **Reading.** `rateCell` prints `formatPercent(null)`, which is the em dash the rest of the app already uses for an undefined ratio — `formatMoney`, `formatTps` and the Trace page's hit rate all render it, so no new string was needed. A real 0% still prints `0%`.

Live, on a bucket where one agent was idle and the other failed every request:

```
2026-08-18
default_agent      0    —
analytics_agent    2    0%
Total              2    0%
```

and the same rule on the `bucketTotal` row, which reads `—` when the whole column had nothing to rate. The Token chart's bubble on a bucket with no cache traffic reads `Cache hit rate —` while its curve rides 100% through that bucket.

The path data shows the two cases are told apart by shape as well as by text. `default_agent` (67% on 08-15, idle 08-16 through 08-19, a real 100% on 08-20, 67% on 08-21) draws `M…,66 L…,10 L…,10 L…,10 L…,10 L…,10 L…,66`; `analytics_agent`, which failed every request on 08-18, drops to `y=178` there and nowhere else. On this geometry `y=10` is 100% and `y=178` is 0%.

## What is left of the idle machinery

Nothing. All three pieces existed to disambiguate a `0%` that no longer appears, and a dash is the distinction the footnote was describing.

| Removed | Why |
| --- | --- |
| `idleBuckets` | It marked `denominator === 0`, which is exactly what `rateSeries` now returns null for. Two representations of one fact, and the second one could drift. |
| the grey class on an idle cell | Colour was standing in for "this number is not what it looks like". The cell no longer holds a number. |
| `S.usage.rateIdleNote` (both catalogs) | It explained an ambiguity that is gone. Left in place it would caption a table full of dashes with a sentence about `0%`. |

`NO_RATE_PLOT` and `plotRates` replace them, and they are about drawing only. The catalogs gained `skippedBuckets` for the caption below, so both dictionaries stay key-for-key identical.

## Cost and statistics skip empty buckets

`compactSeries` (in `usage-controls.ts`) drops the buckets that recorded nothing, and the survivors go into `makeGeom` **as the whole series**. Nothing in the geometry changed: `step = innerW / n` with `x(i)` at the cell's midpoint already puts one point in the middle of the card, two at a quarter and three quarters, and subdivides from there. That is the "从中间开始，横刻度逐渐变多" behaviour, and it is the same code `origin/main` ran.

I checked what "as before" meant instead of guessing. On `main` the cost and Token charts drew `trend`, and `foldTrend` built a map keyed by the rows it actually had, so a date with no usage rows simply had no point. `requests` is `COUNT(*)` in the series query, so `requests > 0` is exactly "a row existed". The predicate is `requests === 0 && total === 0`; the `total` half keeps a bucket whose only requests spent nothing, which `main` kept as well.

Labels follow the plotted buckets, so a slot always names the bucket in it. Both charts keep the label logic they had on `main`: the Token chart labels via `autoLabelIdx`, the cost chart via `ChartFrame`'s sparse first/middle/last. Live over a 7-day range with three empty days, the Token chart labels `08-15 08-18 08-20 08-21` and the cost chart `08-15 08-18 08-21`.

`TrendChart` gained the `Empty` state the other three charts already had. A range that compacts to nothing used to draw an empty grid, which reads as a flat zero cost.

## The two x axes, and how they are told apart

The requests charts keep every bucket in the range. Compressing them is not even well defined — a bucket `default_agent` skipped is a bucket `analytics_agent` worked in, so there is no single "empty" to drop, and dropping the union would erase real data. The mismatch is real, and it had to be handled instead of designed away. Three things handle it:

1. **A break mark on the axis.** `ChartFrame` takes `axisBreaks` and draws the standard `//` at the baseline between two points that are not neighbours in time. It suppresses itself below an 8px cell, where successive marks would sit closer together than they are wide and hatch the whole baseline.
2. **A caption under the chart.** `Skips 3 empty intervals — this axis is not aligned with the charts above` / `已跳过 3 个空区间，横轴与上方图表不一致`. It names the count, so "3 skipped" against a 7-day range is itself readable.
3. **One compaction for both.** `compactSeries` runs once in `usage-page.tsx` and feeds both charts. Had each compacted itself, the Token chart's `total === 0` guard and the cost chart's null-cost handling could have parted ways and produced two axes that disagree with each other as well as with the row above.

The caption lives inside the chart component rather than beside it, which matters in one case I hit: a window whose only non-empty bucket spent no tokens leaves the Token chart with nothing to draw, and a chart showing "No usage records" must not caption an axis it never drew. Confirmed live — Token card empty with no caption, cost card one `$0` point with the caption.

I did not try to make the two axes match. Any scheme that does (padding the compressed charts back out, or compressing the requests charts by some union rule) would undo the change you asked for or invent data. Saying plainly that they differ is the honest option.

## Live checks

Server on a scratch data root, 15 seeded records over two agents and two models across 2026-08-15…21, with 08-16, 08-17 and 08-19 recording nothing.

| Range | Cost/Token points | Skipped | Break marks | What it shows |
| last 7 days | 4 | 3 | 2 per chart | requests charts label all 7 buckets, the other two label only what they plot |
| last 30 days | 4 | 26 | 2 per chart | 26 empty days vanish instead of flattening the line |
| last 24 hours | 7 | 18 | 2 per chart | labels jump `16:00 → 08:00` across the overnight gap, and the mark is there |
| last hour | 1 | 60 | none | the single point sits in the middle of the card; one point has no gap to mark |
| 08-16…19 | 1 (cost) | 3 | none | Token card empty with no caption, cost card one `$0` point with the caption |
| 2026-01-01…10 | 0 | — | none | all four cards show the empty state, no captions |

Also read off the live DOM: two captions on the page and eight break lines (two marks × two skips × two charts), never on the requests charts.

## Verification

`penguin-web` 1035 tests / 81 files (1028 before; the seven new ones cover null rates, `plotRates`, and `compactSeries`'s points, breaks, skips and the leading/trailing cases), `penguin-server` usage suite 28, `penguin-docs` 53, `tsc --noEmit` on `packages/web`, `prettier --check` on every touched file, `git diff --check`. The `usage-controls` rate tests were rewritten rather than dropped: the case that asserted `0` for an unrated bucket now asserts `null`, and a new case pins that a rated `0` survives as `0`.

Docs and the changelog pair moved with it. `web-app.en.md` / `.zh.md` gained a bullet for the two things a chart cannot draw, and the entries gained a `## Buckets with nothing in them` / `## 空桶怎么处理` section, mirrored section for section.

## Still open, unchanged

Your earlier queue stands. New from this round:

- **The cost chart labels three of its four points**, because `ChartFrame`'s default sparse labelling is what `main` gave it. On a compressed axis an unlabelled slot is slightly harder to place than it used to be. Switching it to `autoLabelIdx` (what the Token chart uses) would label all four and make the two compressed cards consistent, but it also thickens the axis on a 90-day range, so I left `main`'s behaviour alone instead of changing a fifth thing.
- **The break mark is dropped below an 8px cell.** At that density the axis carries only the caption. Reachable on a 90-day range that has values in most buckets.

## Design docs

`design/specs/06-PROTOTYPE.md` 「成本与统计中心」 (lines 133–138) still describes the pre-PR page: the Agent pie chart, the per-Model progress bars, 逐日 Token and 逐日成本, and no date-range control. It is a read-only symlink from this worktree, so I have not touched it. It needs the whole section rewritten for this PR, and this round adds three lines to that rewrite: the range control and its derived precision, the two requests charts on every bucket versus the cost and Token charts on non-empty buckets only (with the break mark and caption), and the rule that a bucket with no rate is drawn at the top of the axis and read as a dash.

hiyouga:
Both changes, pushed as `ff4a6b53`.

## Every chart skips the empty buckets

You were right that the objection was about the wrong thing. "Empty" for a *bucket* and "idle" for an *entity in a bucket* are separate questions, and `compactSeries` only ever asked the first one — so the requests charts can compact through the same call. A bucket stays if anything at all ran in it, and the entities that did not run sit at zero in that column with a dash for their rate, exactly as before.

`compactSeries` runs once in `usage-page.tsx` and feeds all four charts. It now returns `kept` (where each surviving bucket sat in the original series) next to `points`, and `compactCounts` re-indexes each per-entity series through that same list:

```tsx
const plotted = compactSeries(data?.series ?? []);
const agentEntities = (data?.byAgentSeries ?? []).map((s) => ({
  label: s.agentId,
  ...compactCounts(s, plotted.kept),
}));
```

One `kept` list for the series and for every entity is the whole safety argument: there is no second filter that could disagree with the first.

## Confirming the alignment survived

Two ways, because this is the invariant the per-entity series are built on.

**By bucket key, in a unit test.** `compactCounts (per-entity alignment across the compaction)` compacts a five-bucket series with two empty buckets, then asserts by key rather than by position — for every drawn point, the entity's value equals what the server sent for *that bucket*, found by `findIndex` on the bucket name. A positional-only assertion would pass under an off-by-one; this one cannot. Two more cases pin the parts that matter: an entity idle in a kept bucket keeps its zero there (`rateSeries` → `[null, null, 100]`, which is what makes it a dash and not a missing column), and a short array reads as zero.

**Against the live server.** I hovered every column of the by-agent chart and compared the bubble to the raw `/usage` response at the *original* index for that bucket key:

| column | bubble | server at that bucket |
| --- | --- | --- |
| 0 · 2026-08-15 | `default_agent 3 67%` · `analytics_agent 0 —` | `series[0]`: default 3 req 2/3, analytics 0 req, denominator 0 |
| 1 · 2026-08-18 | `default_agent 0 —` · `analytics_agent 2 0%` | `series[3]`: default 0 req, analytics 2 req 0/2 |
| 2 · 2026-08-20 | `default_agent 4 100%` · `analytics_agent 3 100%` | `series[5]`: default 4 req 4/4, analytics 3 req 2/2 (one aborted, counted as a request, out of the denominator) |
| 3 · 2026-08-21 | `default_agent 3 67%` · `analytics_agent 0 —` | `series[6]`: default 3 req 2/3, analytics 0 req |

Original indices 0, 3, 5, 6 against drawn columns 0–3: every value stayed with its own bucket across a filter that moved three of the four.

## The caption is gone

`S.usage.skippedBuckets` is deleted from both catalogs (42 keys each, verified equal), the `AxisSkipNote` component and the `skipped` prop on both charts are gone, and `compactSeries` no longer computes a `skipped` count that nothing would read.

**The break marks stay.** They were never really carrying the mismatch; they say that time was skipped, which is still true and is now the axis's own shape rather than a discrepancy between two axes. Everything that justified them as a mismatch signal was rewritten: the `ChartFrame` header, the `axisBreaks` prop doc, the `usage-charts.tsx` header ("Two kinds of x axis on one page" became "One x axis for the whole page"), the `usage-page.tsx` header, `compactSeries`' own doc, and both docs files. The changelog pair's `## Buckets with nothing in them` / `## 空桶怎么处理` section was rewritten the same way and still mirrors section for section.

## In the app

Same seeded data — 15 records over two agents and two models across 2026-08-15…21, with 08-16, 08-17 and 08-19 recording nothing.

```
x-labels [Requests & success rate by agent]: 08-15 08-18 08-20 08-21
x-labels [Requests & success rate by model]: 08-15 08-18 08-20 08-21
x-labels [Token trend]:                      08-15 08-18 08-20 08-21
x-labels [Cost trend]:                       08-15 08-18 08-21
captions on page: 0 | break lines: 16
```

Four charts, the same four columns, sixteen break lines (two marks × two skips × four charts). The cost chart plots the same four points and labels three of them, since `ChartFrame`'s default sparse labelling is what it had on `main`.

Across the other ranges: last 30 days gives the same four columns in all four charts; last 24 hours gives the same six in all four, with the overnight jump (`16:00 → 08:00`) now marked on the requests charts too; a range with no records leaves all four cards showing their empty state. No console errors.

One thing I did not touch, since it predates this change and nobody asked: the cost card has no right-hand axis, so its `padR` is 8 where the other three use 34. The three right-axis charts line up column-for-column across cards; the cost chart plots the same buckets but its points sit a few pixels wider apart. Giving it `PAD_R_AXIS` would align it exactly at the cost of an empty 34px right margin.

## Verification

`penguin-web` 1039 tests / 81 files (1035 before; the `compactSeries` cases now assert `kept`, and the three new `compactCounts` cases cover the alignment), `tsc --noEmit` on `packages/web`, `prettier --check` on every touched file, `git diff --check`, plus the live pass above.

## For the design PR

The rule to write into [penguin-harness-design#51](https://github.com/Prism-Shadow/penguin-harness-design/pull/51) is one sentence, and it replaces the two-axis paragraph entirely:

> 所有图表都跳过空桶——「空」按桶判断（该时间桶内没有任何请求、也没有任何 Token），绝不按单个对象判断：只要有任何对象在该区间跑过，桶就保留，当时空闲的对象在该列显示 0 与「—」。四张图因此共用同一条横轴，跳过之处画断轴标记，不再有任何文字注解。

Supporting detail if it helps: the per-entity series (`byAgentSeries` / `byModelSeries`) are re-indexed onto the surviving buckets by the same list as `series`, so the index-for-index alignment the server guarantees holds after filtering. And the rate rule is unchanged from the previous round — a bucket with no rate is drawn at the top of the axis and read as a dash.

PR #373, opened 2026-08-20
feat(web): link a live background process to the service URL it announced

hiyouga:
## What this does

The session details card's background-process list now shows the service URL a live process announced, as a clickable link next to its pid — a dev server the conversation started is one click away instead of a copy-paste from the transcript.

## How it works

Detection is entirely web-side and pure — no server or core change, no new API field, no extra polling cost:

- The transcript the page already holds carries the binding: exec_command's promotion note (`[process running with process_id …]`, and input_command's `[process still running with process_id …]` spelling) names the process an output belongs to, and every input_command call carries its `process_id` argument.
- `process-service-url.ts` walks the loaded items (nested subagent conversations included), strips ANSI CSI/OSC wrapping, and keeps the **last** local URL each process's outputs printed. Hosts that mean "this machine" only (`localhost`, `127.0.0.1`, `[::1]`, and the listen-side wildcards `0.0.0.0` / `[::]`, rewritten to `localhost` so the link is clickable); remote hosts in log lines are ignored. Sentence punctuation is trimmed; a poll that prints no URL leaves the previous URL standing.
- The row links only while the process is running — an exited service's URL is dead. The link opens in a new tab (`rel="noopener noreferrer"`), shows the scheme-less form at 11px, and carries the full URL in its tooltip. No new i18n strings were needed.

Placing detection web-side was deliberate: the command session's output buffer in core is a private, drain-on-read structure with no transcript accessor, and the server would have to re-scan stored messages on every processes poll — while the web page already has the full transcript in memory. It also keeps this change fully clear of the concurrent background-execution work in core's `environment/tools/` and the server session-manager (no shared files beyond `chat-page.tsx`'s process list markup).

## Verification

- New `packages/web/test/process-service-url.test.ts`: 16 tests — promotion-note binding, input_command args binding (complete and partially streamed JSON), the still-running spelling, last-URL-wins with URL-less polls keeping the previous value, ANSI CSI + OSC stripping, wildcard-host normalization, loopback hosts kept, punctuation trim, https/port-less origins, remote hosts ignored, nested subagent walking (bound card and standalone card).
- `pnpm --filter @prismshadow/penguin-web test`: 76 files / 1005 tests passed.
- `pnpm typecheck` across the workspace: clean. `pnpm format` + `format:check`, `git diff --check`: clean.


hiyouga:
Closing per maintainer request — superseded by #376. The transcript-scanning approach here misses backgrounded processes: their tool output carries only the `[process running with process_id …]` note, so the service's own URL never reaches the transcript. The feature moves into #376, where the command-session layer sees the full output stream (write-time URL scanning) and can fall back to an OS listening-port probe on the process group.

PR #374, opened 2026-08-20
feat(core,server,web): add project sandbox command policy

hiyouga:
## What this adds

A **Project-level sandbox command policy** plus a **tabbed Project settings dialog**. The policy is a deny-rule list over shell text, enforced ahead of the approval mode: a hit is refused outright — even under allow-all — with the fixed `aborted` line `Tool call denied by policy.`, so the model changes course instead of reading a user cancellation. Both tools that reach a shell pass the same gate: the command `exec_command` launches (`cmd`) and what `input_command` types into an already-running one (`chars`; a lone Ctrl-C is exempt — the tool turns it into SIGINT, so it is not typed text). Placement follows the settings-ownership principle: security policy is Project-owned config, deliberately not Agent State the Agent itself can rewrite.

## The boundary

Enforcement sits at the Human boundary, not in the engine. `Session.run` wraps the injected `approve` callback with the policy (`withCommandPolicy`, core-internal): a vetoed call answers `"forbidden"` before the host is ever consulted, which is what makes the policy outrank every approval mode — for every host, and for subagents, since a child Session wraps its own forwarded callback through the same `Session.run` path. `context_engine` holds no policy and keeps doing exactly what the spec assigns it: call the approval boundary once per complete `tool_call`, transcribe the answer into an `approval_decision` event, feed a denied call its terminal output.

The engine's change shrank to its floor: the only diff against the merge base is the denied-output wording. `ApprovalDecision` is now three-valued — `"allow" | "deny" | "forbidden"` — and the engine picks "Tool call denied by user." or "Tool call denied by policy." from the decision, `aborted` either way. No union type, no narrowing helper, no extra event field: `"forbidden"` IS the audit marker, carried by `approval_decision.decision` itself, and the web tool card's decision label ("Denied · policy"), the CLI decision line and the Trace viewer row render straight from it — unlike the earlier `policy_rule` draft field, dropped because nothing consumed it. The denial does not name the matched rule; the rule list lives in Project Settings.

## Source of truth

`.project_config.toml` is the **only** truth: the `[command_policy]` block holds the rule list; an absent block or absent `rules` list means the factory set, resolved through **one** function (`effectiveCommandPolicyRules`) shared by enforcement, the server's GET, and the seeding path — while a stored empty list means "no rules", a distinct state. The policy is parsed once when the Agent loads its config and travels **by reference** — no re-parse, no clone: `ProjectConfig.command_policy` → `SessionConfig.commandPolicy` → `Session.run`'s wrapper. That one parse is the snapshot: the running Session's effective copy is an in-memory object a disk edit cannot reach, and applies from the next config load. The UI reads the file through GET and every Save is a whole-list PUT that materializes the list back into the file, rejecting uncompilable patterns up front.

## Rules are seeded, editable data

No built-in tier. Every rule is `{ name, pattern, description?, enabled? }`; the factory set (`state/command-policy-defaults.ts`) is **seeded like the model presets** — copied into each new project at `writeInitialConfig`, never auto-rewritten; a pre-seeding project behaves as the factory set until its first saved edit materializes the list. "Restore defaults" re-buffers the factory set (served as `defaultRules`) and Save writes it back. Factory set, POSIX plus Windows counterparts: `rm` with both a recursive and a force flag · `mkfs` · `dd` onto a block device · the fork bomb · `>` redirection onto a block device (`/dev/null` stays legal) · `Remove-Item -Recurse -Force` / `rd /s /q` · `format C:` / `Format-Volume` · raw `\\.\PhysicalDriveN` writes / `Clear-Disk` · `%0|%0`. Matching normalizes whitespace and re-tests each rule against a quote-stripped variant of the command, so a leading path (`/bin/rm`), a wrapper (`sudo`, `env`, `xargs`), a quoted or escaped command word (`"rm"`, `r''m`, `\rm`) and a literal `sh -c '<payload>'` all hit. Disabling just the `rm` rule is the escape hatch for `rm -rf node_modules` workflows.

## What it is deliberately not

An accident guardrail, not a security boundary. A command assembled at run time — `$IFS`, `X=rm; $X`, a substitution, `eval`, base64 into a shell, an interpreter fed through a pipe — is deliberately uncovered, and a test pins nine such spellings as non-matches so a future "win" there gets re-examined rather than celebrated: more patterns in that direction buy the appearance of coverage, at permanent maintenance cost. MCP tools are out of scope (the per-server `permission` level from #364 fixes what its tools report to the approval mode; it confines nothing). The real boundary is process confinement (#354); the docs position this policy as complementary to it.

## Tabbed Project settings

Left tab rail (icon + label; horizontal strip on narrow screens), row-styled pages: **General** (display name, Project ID, delete), **Members** (permission table; hidden in the single-user desktop app), **Defaults** (the `[default_chat]` section), **Security policy** (intro copy → master switch → rule list with per-row switch / inline edit / delete / add / Restore defaults → one explicit Save, blocked while a row edit is open). Server routes: `GET|PUT /api/projects/:projectId/command-policy` (member read / owner write; the GET also serves `defaultRules` for restore).

## Compatibility notes

- **Behavior change, flagged `Breaking` in the changelog:** every project starts denying the factory set under every approval mode until a rule (or the policy) is disabled in Project Settings → Security policy.
- **Protocol:** `approval_decision.decision` gains the third value `"forbidden"` (the policy veto); `allow` / `deny` and the human path's wire format are unchanged, and traces predating the value never contain it. The denial output reuses the existing `aborted` tool-output shape — one fixed line per decider.
- **SDK surface:** `ApprovalDecision` widened from two to three values; `ApproveFn`'s shape (`Promise<ApprovalDecision>`) is what it was before this PR, and existing `"allow"` / `"deny"` callbacks need no change. `"forbidden"` is reserved for the policy wrapper — hosts keep answering the two human values.
- **History:** the branch was squash-rebased onto `main` once early on (single commit, `--force-with-lease`, all commits this branch's own); everything since is ordinary follow-up commits, plus one merge of `origin/main` to resolve import-list conflicts with #376 (background execution).

## Verification

Full suites at `88d573b9`: core 884 passed / 5 skipped, server 816, web 1060, cli 292, docs 53; `tsc --noEmit` clean on all seven packages; `prettier --check .` clean; CI green (build, ci, ci-macos, ci-windows, runtime ×3). After the convergence to the three-valued decision and the merge of current `main` (`d2c52b39`, `f0a8dff1`): full core suite 926 passed / 5 skipped, the six approval-touching server suites plus the command-policy API suite 65, cli, web (1060) and docs (53) package suites green; `pnpm -r build`, `pnpm typecheck` and `pnpm format:check` clean. Behavior contracts pinned in tests: a policy hit is refused with `failed` and the host's callback is never invoked, allow-all included; `input_command`'s typed text is checked with the lone Ctrl-C exempt; absent block ⇒ factory set and stored `[]` ⇒ no rules; the seeded copy round-trips through the config file; nine run-time-assembly spellings are pinned as deliberate non-matches; an uncompilable pattern is 400 at write and skipped at match.


hiyouga:
Both objections hold, and I acted on both: enforcement moved out of `context_engine` to the approval boundary in `Session.run`, and the factory rules moved from `environment/` to `state/`. The feature itself is sound — one parse, one fallback, rules as editable data, owner-only writes, and a config layer that is genuinely tolerant of a hand-broken file. What is not sound is what the PR says about it. "Outside anything the Agent's own tools can rewrite" is false (there is no path jail; `write_file` resolves an absolute path), "in any spelling" is false (`/bin/rm -rf /` walked straight past), and "Restore defaults writes the factory set back" describes a button that only buffers. I fixed those, closed four cheap matcher holes, and fixed three bugs in the settings page. What remains open is listed at the bottom, including every bypass I found and left alone — and the design docs, which this PR contradicts in five places and does not update.

Pushed on top of `1bead4d7`: `87b07df0`, `78a9583d`, `fcf69f85`, `02803487`.

## The boundary question

Your reading is right, and the spec text is unambiguous. `design/specs/05-ARCHITECTURE.md`:

- **:13** — `context_engine` 负责维护模型上下文，并处理 Human、LLM 和 Environment 三个主要接口之间的事件流动。
- **:21** — 所有与 Human、LLM 和 Environment 相关的能力都由对应接口负责。`context_engine` **只负责**处理 OmniMessage 格式的消息。
- **:17** — `Human`：…并**审批所有工具调用**。
- **:45** — Human 审批并同意工具调用后，`context_engine` 请求 Environment 执行工具。

At `context-engine.ts:1130-1150` the engine imported `vetoForToolCall`, matched a regex against the tool call's arguments, and emitted `approvalDecision("deny", …)`. Two separate breaches: the engine held a capability (:21), and it authored a Human decision nobody made (:17, :45). Building the `approval_decision` message is the engine's job — that part was fine and unchanged — but the *decision* has to arrive through the approval channel.

**Where it went.** Not the Environment: you ruled it out, and `read-file.ts`'s `SECRET_BASENAMES` precedent only argues that a *tool* may refuse its own arguments, which is a different thing from a project-wide policy over another tool. The Human boundary in this codebase is `Session.run`'s injected `approve` — `session.ts:4-8` states exactly that, and there is already precedent for decorating it: `agent.ts:710` wraps the parent's callback with `withOrigin` before handing it to a child. So `Session.run` now wraps `opts.approve` with the policy (`session.ts:310-318`, `internal/command-policy.ts:121`). A vetoed call answers "deny" without the host being asked; the engine is back to ask → record → feed a terminal output, and imports nothing about policy.

The guarantee survives the move intact. It is not "the CLI and the server both remember to check" — it is one wrap inside core, above every host and below none, applied before the goal/task split so a goal round is covered too. Subagents get it twice: the child Session wraps its own forwarded callback, and the parent's wrapper also sees the child's origin-tagged `tool_call` through the relay in `subagent/session.ts`.

**Failed vs. aborted — the obstacle, solved rather than reported.** You were right that this is the real cost. `ApproveFn` returned one bit, so a wrapper could not say *why*. Options were: collapse the two stop reasons (you said not to, and the distinction is real), or widen the channel. I widened it:

```ts
export interface ApprovalRefusal {
  decision: "deny";
  message: string;                            // fed back verbatim as the tool output
  stopReason: Extract<StopReason, "failed" | "aborted">;
}
export type ApproveFn = (tc) => Promise<ApprovalDecision | ApprovalRefusal>;
```

A bare `"deny"` still produces `aborted` / "Tool call denied by user.", so no existing callback changes behavior. Note the shape carries no policy vocabulary — it is a denial with a reason, which is a property of approvals generally, not of this feature.

Two things fell out of it, both worth knowing:

- **`approval_decision.policy_rule` is gone.** Nothing read it — the only reference outside the builder was the engine test's own assertion. Keeping it would have meant threading a `policyRule` field through the generic approval type, which is the same speculative-seam smell round 2 already cut once. The rule name is in the model-visible message, and `failed` vs `aborted` separates the two denials in the Trace. That also means the OmniMessage protocol is now unchanged by this PR — one fewer spec edit.
- **Widening the return has a blast radius**, which is the honest argument against the whole approach: eight call sites broke, all of them consumers that read a decision (the CLI renderer at `task-loop.ts:83,105`, and six server test fakes that mimic the engine). `approvalDecisionOf` (`interfaces.ts:111`) is the single narrowing they all use now. Small, but it is a public SDK contract change and an embedder with a strict `ApproveFn` return type will see it.

Tests: the engine test dropped its policy knowledge and now pins the widened contract (`engine.test.ts:482`); `command-policy.test.ts` gained a `withCommandPolicy` unit block and two Session-level tests built on a fake bootstrap, proving the wiring end to end — approve never called, Environment never reached, `failed` output fed back, and the mirror case where `{ enabled: false }` lets the same command through to execution.

## Where the defaults live

Also right, and the module's own header was the evidence: it existed because `state/project-config.ts` needed the data and importing it from `environment/` was a cycle. A module that exists to work around living in the wrong directory should move, not keep its workaround. `state/` is where this repo already keeps seeded catalogs (`model-catalog.ts`, `default-config.ts`), and `project-config.ts:207` seeds these exactly like the model presets.

Now `packages/core/src/state/command-policy-defaults.ts`, exported through the state barrel. `state/project-config.ts` imports a sibling; the cycle is gone rather than routed around, and the header no longer has to explain a workaround.

On the counter-weight you raised — `effectiveCommandPolicyRules` is the enforcement-time fallback, so whichever side it sits on, the other imports it. It went with the data, because it is one line over that data and the alternative splits "the factory set" across two modules. The matcher (`evaluateCommandPolicy` / `vetoForToolCall` / `withCommandPolicy`) went to `packages/core/src/internal/command-policy.ts` — it is neither seeded state nor an Environment capability any more; it is what `Session` applies to an approval callback, which is what `internal/` holds (`session-support.ts`, `session-title.ts`). It still imports `EXEC_COMMAND_NAME` from the tool it guards rather than re-declaring the name.

## Fixed

**`78a9583d` — four cheap matcher holes.** These were plain spellings, not evasion, and each was one anchor away. Test added per case (`command-policy.test.ts`).

| Got past | Now |
| --- | --- |
| `/bin/rm -rf /`, `sudo /usr/bin/rm -rf /`, `/sbin/mkfs.ext4 /dev/sda1`, `/bin/dd …` | shared `CMD_START` allows the directory part of a path. `mkfs` normally *lives* in `/sbin`, so the rule was missing the common spelling, not a clever one |
| `rm -rF /`, `rm -R -F /` | recursive already accepted `-R`; force was lowercase-only. Now `[fF]` |
| `dd … of=/dev/sda; sync`, `echo x > /dev/sda&`, `(dd … of=/dev/sda)` | the trailing `(?:\s\|$)` was redundant against a greedy device class and only cost matches; dropped |
| `dd … of="/dev/sda"`, `> '/dev/sda'` | optional quote before the operand |

Knock-on I accepted: `mkfs` now also fires on `./mkfs.sh`, consistent with it already firing on `man mkfs` — the word is the risk signal. I also checked the patterns for backtracking: `:(` × 4000, `/` × 4000, 20 KB of junk all run in well under a millisecond, and scaling is linear.

**`fcf69f85` — claims that were not true.**

- `configuration.{en,zh}.md:96`, both changelog files, `server/src/http/routes/command-policy.ts:5` all said the config sits "outside anything the Agent's own tools can rewrite". There is no path jail: `write-file.ts:86` and `edit-file.ts:103` are `path.resolve(ctx.workspaceDir, filePath)`, so an absolute path passes through, and `exec_command` goes anywhere. Only `read_file` refuses those basenames (`read-file.ts:68`). The true claim is the narrow one the core header already made — outside the *Agent State* an Agent edits for itself, plus the per-Session snapshot. Rewritten to that, with the limitation stated.
- "`rm` … in any spelling" (docs ×4, changelog ×2, and the seeded rule `description` itself) → "carrying both a recursive and a force flag", with the path/`sudo` coverage named.
- "'Restore defaults' writes the factory set back" → it buffers; Save writes.
- The docs' one sentence about "deliberate obfuscation" is now an explicit gap list: nested interpreter, `input_command`, quoting/`$IFS`/variables, MCP shells, and everything not on the list. The web intro string says `exec_command` launches rather than "every shell command", and names the three routes around it.
- The zh changelog had `## Details` untranslated (`changelog/README.md:77` mandates `## 细节`) and left the `Breaking` reason in English (README:72 — it is the one prose field that gets translated).

**`fcf69f85` — three bugs in the settings page** (`project-dialogs.tsx`):

1. The load effect never cleared its own state, and the render checks `loadError` before `saved` — so one failed fetch latched the error permanently over every later successful load. `MembersSection:440` already does this correctly; now so does this one.
2. Save was `disabled={busy || !dirty}` while Restore/Add/Edit/Delete all guard on `editing !== null`. Saving mid-edit replaced the list underneath an editor keyed `edit-${i}`, which does not remount — so the stale draft survived and Apply then wrote it over index `i` of the *new* list. Save now takes the same guard.
3. The rule list used `pointer-events-none` as its disabled state while the master switch is off. That is a visual, not a disable: every switch and button stayed keyboard-focusable and fully operable via Tab. Each control now disables on `!enabled` too.

**`02803487`** — the `approvalDecisionOf` narrowing helper plus its call sites, described in the boundary section.

## For you to decide

**The design docs. This PR contradicts them in five places and updates none of them** — `CLAUDE.md` says design and code move together, and `design/` is read-only here, so these are yours to write:

1. `05-ARCHITECTURE.md:97` — pins `ApproveFn` as `(toolCall) => Promise<"allow" | "deny">`. Needs the `ApprovalDecision | ApprovalRefusal` union, and one clause saying `Session.run` wraps the injected callback with the Project command policy, so a hit is refused before the callback is consulted. (`:837` and the payload sample at `:844` need nothing — dropping `policy_rule` kept the message format unchanged.)
2. `01-PRINCIPLES.md:160-179` — the `.project_config.toml` field structure block has no `[command_policy]`. Add `enabled` and `[[command_policy.rules]]` (`name` / `pattern` / `description` / `enabled`), plus the absent-block-means-factory-set rule. `:97`, which enumerates what the file records, needs the same.
3. `05-ARCHITECTURE.md:1240`-block — the route list needs `GET|PUT /api/projects/:projectId/command-policy` (member read / owner write). Precedent is mixed: `PUT …/models/default` and the chat-defaults routes from #191 never made it into this list either, so this may be deliberate.
4. `06-PROTOTYPE.md:52` — describes Project settings as one dialog with a defaults block below Members. It is now a four-tab dialog with a Security policy page.
5. `08-CHANGELOG.md` — no entry for this change.

**Bypasses I found and did not close.** Whether any are worth a rule is your scope call; I would argue the first is the one that matters, because it makes the guardrail exactly one interpreter launch deep.

- **`input_command`, two calls.** `exec_command {cmd:"bash"}` matches no rule; the tool returns `process_id` in its own note; `input_command {process_id, chars:"rm -rf /\n"}` writes straight to stdin (`tools/command/session.ts:250`) with no policy call and no validation. Documented as intentional, and the existing suite pins it (`command-policy.test.ts:166`).
- **MCP tools are entirely uncovered** — the gate is `toolName !== EXEC_COMMAND_NAME`, so any MCP server exposing a shell runs unchecked. Adjacent to #364, which gave MCP servers their own permission level; a policy that reaches MCP tool arguments would be the natural sequel, and the data member is already positioned for it.
- **Nested interpreters and pipes** — `sh -c 'rm -rf /'`, `eval "…"`, `echo … | base64 -d | sh`, `curl … | sh`, `python -c "shutil.rmtree('/')"`. Regex over the launch string cannot reach a quoted payload. (Amusingly the *unquoted* `bash <<EOF\nrm -rf /\nEOF` and `find . -exec rm -rf {} +` do match, purely because a space lands before `rm`.)
- **Shell-level indirection** — `\rm`, `"rm"`, `r''m`, `rm -rf$IFS/`, `X=rm; $X -rf /`, `R=-rf; rm $R /`, `$(echo rm) -rf /`, `alias del='rm -rf'; del /`. Needs tokenization, not patterns.
- **Windows is 100% uncovered.** `shell.ts` resolves pwsh/cmd, but every factory rule is POSIX. `Remove-Item -Recurse -Force C:\`, `rd /s /q C:\`, `format C: /q`, `Clear-Disk` — none match. If Windows is a supported host for `exec_command`, the factory set arguably owes it three rules.
- **Not on the list at all** — `shred`, `wipefs`, `sgdisk`, `mkswap`, `find / -delete`, `rsync --delete`, `chmod -R 000 /`, `git clean -xfd`, `git reset --hard`, `crontab -r`, `docker system prune -af`. Deliberate ("kept small"), and users can add rules; worth saying so in the docs, which now do.
- Separator note, since the module header claims patterns "never cross a `;`, `|` or `&`": they do. `normalizeCommand` folds newlines into spaces and `[\s;&|(]` is a valid match position, so `echo hi; rm -rf /`, a real newline, `&&`, `||`, `for f in *; do rm -rf $f; done` and a whole shell script pasted as one `cmd` are all matched. What `[^;&|]*` limits is only the flag scan — both flags must sit in the `rm`'s own segment. The header oversells the limitation; the behavior is better than advertised.

**Judgment calls I left alone:**

- **False positives are the product decision here.** `rm -rf node_modules`, `rm -rf dist`, `git rm -rf --cached .` are all blocked, and — by design — un-overridable even in allow-all. The escape hatch is "disable the `rm` rule", i.e. all or nothing per rule. A path-aware rule (deny only outside the workspace) would keep the common workflow *and* the protection; that is a bigger design question than a review can settle. Related: the rules also match prose, so `git commit -m "fix rm -rf handling"` is blocked while `git commit -m "docs: warn about rm -rf"` is not, because the boundary depends on the character after the phrase. That looks arbitrary from the outside.
- **"Sandbox".** Nothing is sandboxed — it is a deny list over one tool's `cmd`. The naming is already inconsistent (the EN docs heading is "Command policy", the zh is 沙箱安全策略, the UI tab is "Security policy", the denial string says "sandbox policy"). A rename touches the changelog filename, a user-visible denial message and a zh doc anchor, so I left it; a half-rename would be worse than either end.
- **A rule's regex is owner-supplied and runs against model-supplied text**, compiled fresh per evaluation with no timeout. The factory patterns are linear (measured), but an owner who writes `(a+)+$` gets a hang on the approval path. Owner-only writes make it low severity; a length cap exists (512) but says nothing about complexity.
- **Duplicate rule names** are accepted by both `RuleEditor.apply` and `validateRules`, which makes the denial message ("rule: X") ambiguous about which rule to disable.
- **`parseCommandPolicy` narrowing is silent in one direction that matters**: `rules = "oops"` reads as *absent*, i.e. the factory set, while a rule missing `name` or `pattern` is dropped individually and quietly shrinks the deny set. Both are the documented lenient stance and the write path validates, so this only bites a hand-edited file — but "dropped a rule" and "no rules stored" deserve different outcomes.
- **No backward-compatibility entry.** `SKILL.md:103-105` asks that permanent on-disk tolerance name how long it stays and who removes it; `parseCommandPolicy`'s absent-block fallback is exactly that, and five prior releases have such an entry.
- **`RuleEditor` list keys** fold in the index (`${r.name}-${i}`), so deleting rule 0 remounts the whole list. Cosmetic now that Save is blocked during an edit.

## Checked and clean

- **i18n parity** — 21 new keys, identical key sets in `strings-en.ts` and `strings.ts`, no value left in the wrong language. (Also type-enforced: `strings.ts:1687` `type Strings = typeof zh`.)
- **English-only** — no CJK outside the zh dictionary, `*.zh.md`, and the required `[中文版]` link. Zero in `core/`, `server/`, `project-dialogs.tsx`, or any test.
- **Comment conventions** — no `design/specs` citations. `Docs: /docs/… § "…"` is the established form (33 files on `origin/main` use it) and the referenced heading exists.
- **Changelog structure** — right folder, right metadata order, `Issue` correctly omitted, `Breaking: yes` correct with a `## Compatibility` section, PR number correct. (I also cut "and the model is steered to safer spellings" from that section: `changelog/README.md:58` asks for the migration instruction, not an argument that the rest is safe.)
- **Docs mirroring** — en/zh line-for-line; `content.test.ts` anchor and in-page-link rules satisfied, including the two new cross-page links I added (which have to be `/interfaces`, not `/docs/interfaces` — the test strips only the leading slash).
- **Config layer** — non-object block, non-boolean `enabled`, non-array `rules`, a rule missing name/pattern, an uncompilable pattern (kept at parse, skipped at match, rejected at write), an over-long list (server caps at 64), and the `{rules: []}` ≠ absent distinction all behave as documented.
- **Coverage of the veto's claimed paths** — every host builds Sessions through `agent.createSession` / `resumeSession` (`session-service.ts:335`, `cli/commands/chat.ts:133-135`, `run.ts:80`, `scheduler.ts:390`), and both constructions carry the snapshot, so scheduled tasks, resumed sessions, goal rounds and subagents are all covered. Nothing constructs `ContextEngine` directly outside tests.
- **Verification** — core 875 passed / 5 skipped (full suite, twice); server 816 passed (full suite, solo); web 1060; cli 292; docs 53; `tsc --noEmit` clean on all seven packages; `prettier --check .` clean.

Not merged.

hiyouga:
Follow-up: the three coverage gaps are closed — `input_command` is now checked, plain spellings are normalized before matching, and Windows has the counterparts of the rules it was missing. What I did not do is chase the class of bypass that patterns cannot reach, and the docs, the changelog, the settings copy and the module header now say so in those words rather than gesturing at "deliberate obfuscation". `00abe7d5`, `5dc1b6bf`, `795658ed`, `88d573b9`.

One correction to my own earlier report, found while checking rather than assuming: **leading wrappers were never a gap.** `env rm -rf /`, `nice`, `nohup`, `command`, `time`, `xargs` all matched already — the anchor accepts any separator, so `sudo` was never special. `command-policy.test.ts:123` now pins that so nobody adds a redundant alternation for it later.

## Closed

**1. `input_command`.** The gate was keyed to one tool name; it is now keyed to each tool's shell-text argument — `exec_command`'s `cmd` and `input_command`'s `chars` — via a small map in `internal/command-policy.ts`. `exec_command {cmd:"bash"}` still matches nothing, but the `rm -rf /` typed into it afterwards does. The exemption is a lone `\u0003`: `input-command.ts` special-cases it into SIGINT rather than writing it to stdin, so it is a Ctrl-C, not a command — the constant is now exported and shared rather than re-declared.

- `command-policy.test.ts:254` — both tools hit on their own argument; `write_file` and an `mcp__server__shell` call with an identical `cmd` argument still pass.
- `command-policy.test.ts:271` — the lone Ctrl-C is exempt.
- `command-policy.test.ts:327` — through `withCommandPolicy`: a typed `rm -rf /` is refused with `failed` and the host is never asked (`asked === 0`), while `make build\n` reaches it.

**2. Plain-spelling normalization.** Rather than adding alternations to every rule, each rule is now tested against the command as written *and* against the same text with quote marks removed (`'` and `"` dropped, `\x` reduced to `x`). Both are tested, so unquoting can only add matches. That one change covers `"rm" -rf /`, `'rm' -rf /`, `r''m -rf /`, `\rm -rf /`, `rm -r''f /`, `rm "-rf" /` and the literal `sh -c '<payload>'` form — including into other rules, so `sh -c 'mkfs.ext4 /dev/sda1'` hits `mkfs` and `zsh -c "dd if=/dev/zero of=/dev/sda"` hits `dd-to-block-device`.

It removes quote marks and nothing else. No expansion, no substitution, no decoding — which is exactly the line between what is covered and what is not, and is stated as such in the function's own comment.

- `command-policy.test.ts:102` (quoting and escapes), `:115` (`sh -c` payloads across four rules), `:123` (wrappers, the non-gap).

**3. Windows.** `shell.ts` resolves pwsh or cmd, and every factory rule was POSIX-shaped, so a Windows host got no protection from rules that were meant to be universal. Four rules, counterparts of the existing five and nothing more, case-insensitive inline (`(?i:…)`, which Node 24 supports) because both shells are:

| New rule | Counterpart of | Catches |
| --- | --- | --- |
| `windows-recursive-delete` | `rm-recursive-force` | `Remove-Item … -Recurse … -Force` (order-free, `-Rec`/`-r`/`-f`), `rd`/`rmdir`/`del`/`erase` with `/s` and `/q` — including run-together `/s/q` |
| `windows-format-volume` | `mkfs` | `format C:`, `Format-Volume` |
| `windows-disk-overwrite` | `dd-to-block-device`, `overwrite-block-device` | a raw write to `\\.\PhysicalDriveN`, `Clear-Disk` |
| `windows-fork-bomb` | `fork-bomb` | `%0|%0` |

Each requires the same pair the POSIX rule requires (`/s` **and** `/q`, mirroring recursive **and** force), so `rmdir empty` and `del one.txt` stay clean, and `format` needs a drive letter so `pnpm format`, `pnpm format:check` and `npm run format && npm test` are not hits. `command-policy.test.ts:159-193`.

Two notes on this. `shell.ts:123-137` prefers bash on Windows when one is present and only falls back to pwsh/powershell, so both rule families are live there — these are additive, not a Windows branch. And `(?i:…)` is an ES2025 inline modifier: `packages/core/package.json` requires Node ≥24, which has it, but on an older engine `compileRule` would return null and skip those four rules silently rather than throw — degraded, not broken.

## Left open, and why

**A command assembled at run time is not covered, and adding rules would not change that.** `rm -rf$IFS/tmp`, `X=rm; $X -rf /`, `R=-rf; rm $R /`, `$(echo rm) -rf /`, `eval $CMD`, `echo … | base64 -d | sh`, `curl … | sh`, `python -c "shutil.rmtree('/')"`. Shell is a programming language. Deciding what a program will do by reading its text before it runs is not a problem you get closer to solving with more patterns — you get closer to *looking* like you solved it, which is worse, because the list of rules is what a reader uses to judge how much protection is there. Every pattern in that direction is also maintained forever and false-positives against ordinary work. Anyone who wants the command to run can get it to run; the point of the list is that nobody runs it by accident.

That is now pinned as a test, not just prose: `command-policy.test.ts:137` asserts those nine spellings **do not match**, with a comment saying that a future rule making one of them match is a claim to re-examine rather than a win.

**MCP tools stay out of scope — but #364 is not the reason I first wrote down, and the docs now say so.** I had written that an MCP server's shell "is governed by" its per-server `permission` level. #364's own page says the opposite in plain words (`tools.en.md:286`): the level fixes what a server's tools report to the approval mode, is never sent to or verified against the server, confines nothing — and `allow-all` never consults it at all. So the honest statement, which is what shipped in `795658ed`, is that `permission` is the knob that exists there rather than protection, and that pushing a shell-text matcher into arbitrary MCP arguments would add a second weak control to a surface that needs a real one. Same class of overclaim I flagged in the first review; I caught this one in my own sentence.

**New command classes are yours to decide, not mine to add.** `shred`, `wipefs`, `sgdisk`, `mkswap`, `find / -delete`, `rsync --delete`, `chmod -R 000 /`, `chown -R`, `git clean -xfd`, `git reset --hard`, `crontab -r`, `docker system prune -af`. Each is a real way to lose data and none matches a factory rule. The module header says the list is deliberately small, and what belongs on it is a product call about what the list is *for* — an accident list, or a hazard list. The docs now tell users to add their own rules for what their project cares about.

**Where the real boundary belongs.** The docs and changelog now point at [#354](https://github.com/Prism-Shadow/penguin-harness/pull/354) — process confinement via bubblewrap and dsh — and position this policy as complementary to it, not a substitute. The framing I used, in both languages: a speed bump is worth having in front of an irreversible command, but a boundary is a process that cannot reach the rest of the filesystem regardless of what it runs, and that is a different mechanism.

## Design PR #50 — claims that changed

From the `08-CHANGELOG.md` sentence beginning **能力边界写在文档里**:

| Claim in #50 | Now |
| --- | --- |
| `` `exec_command` 起 `bash` 再用 `input_command` 往 stdin 写命令即可两步绕过 `` | **Wrong** — `input_command` is checked on `chars`, with a lone `\u0003` exempt as SIGINT |
| 嵌套解释器…不覆盖 | **Partly wrong** — a literal `sh -c '<payload>'` / `bash -c "…"` is covered; an interpreter reached through a pipe, or fed a computed string, is not |
| shell 间接写法…不覆盖 | **Partly wrong** — quoting and backslash escaping of the command word are covered; `$IFS`, a variable, a substitution and an alias are not |
| Windows…不覆盖 | **Wrong** — four counterpart rules added (the four in the table above) |
| MCP 工具…不覆盖 | **Still true.** Worth not attributing it to #364's `permission`, which gates prompting rather than capability — see the MCP note above |
| 它是正则匹配 shell 文本的减速带，不是沙箱 | **Still true.** Worth adding where the boundary does live — #354 |

Suggested replacement for the tail of that sentence, in your voice to keep: 覆盖的是常规写法（带路径、前缀命令、命令词加引号或转义、字面量 `sh -c`），以及 POSIX 与 Windows 两套写法；**运行期才拼出来的命令刻意不覆盖**（`$IFS`、变量、命令替换、`eval`、base64、管道进解释器）——正则读不懂程序，加规则只换来覆盖的表象。MCP 工具不在覆盖内（#364 的 per-server `permission` 只决定其工具向审批模式报告哪个等级，不做隔离）；真正的边界是进程隔离（#354），本策略与之互补。

A third, which the round produced rather than changed: `tools.{en,zh}.md` still printed the old `ApproveFn` signature and described every denial as `aborted` — the one approval page I missed when the union widened. Fixed in `795658ed`, so #50's `05-ARCHITECTURE.md:97` and the shipped docs now agree.

Two more, unrelated to this round:

- **`05-ARCHITECTURE.md:97` is now correct as written** — it names `ApprovalOutcome`, which the code did not export. I added the alias (`5dc1b6bf`) rather than asking you to rename it; the union was spelled out at six call sites anyway. No edit needed there.
- **`06-PROTOTYPE.md` says each rule carries 一句给模型看的驳回说明.** The implementation does not do that: `description` is free text shown in the settings UI, and the denial the model sees is a fixed template naming the rule. Either reading is defensible — a per-rule hint for the model is arguably better — but it is a behavior change to a user-visible string, so I left it for you. Tell me which way and I will make them agree.

## Verification

Core 884 passed / 5 skipped (full suite), server 816, web 1060, cli 292, docs 53. `tsc --noEmit` clean on all seven packages; `prettier --check .` clean. CI green on the branch head — build, ci, ci-macos, ci-windows (packages + server), runtime ×3.

Still not merged.

PR #375, opened 2026-08-20
feat(web): memory changes below the file summary, with a peer Memory side panel

hiyouga:
## What this adds

When a Task changes the Agent's Memory, the conversation shows it in two places:

1. **Memory-changes card** — directly below the existing file-summary card at the end of the Task: one row per memory topic file, scope and change-kind markers (icon + tooltip). Clicking a row opens the Memory panel directly on that memory's **content**.
2. **Memory panel** — a **peer of the subagents, terminal and Workspace panels**: its own entry in the chat toolbar's panel switcher (pinnable, book glyph), the same docked/Sheet shell (drag-to-resize, inert width-collapse) and the same exclusivity choreography (opening any docked panel hard-closes the closed sibling and hands the open one to the sequenced swap; terminal takeover and draft resets close it like the others). Two levels, modeled by pure `memory-nav.ts`:
   - **List** — both scopes' topic lists over the same server memory routes the agent-settings tab reads; topics changed by this conversation carry a marker.
   - **Detail** — the memory's content rendered like the file viewer (Markdown, frontmatter stripped; the header shows title/updated). Back returns to the list.

**Entry routing:** card row → panel open directly on that memory's content (with a one-shot locate flash); panel switcher / card header → list. Session switches reset to the list.

**Deleted memories leave the lists:** a changed file the **loaded** listing no longer carries was deleted in a later turn — it is filtered out of the card and absent from the panel list (`deletedChangeKeys` / `buildMemoryList`); while the listing hasn't loaded, nothing is treated as deleted. The in-detail unavailable note only covers the race where the file disappears between the listing load and the click.

**No diff display.** Earlier revisions rendered per-call and then whole-file diffs; they are fully removed (`lib/memory-replay.ts`, `lib/line-diff.ts`, their tests, the degradation paths and note strings). The change record the stream model keeps is now minimal: scope, file, op, and the latest call's start time. Rationale: the memory contract has the model maintain topic files with ordinary file tools, and in practice it re-writes files wholesale (read → `write_file` full content) nearly every time — under that write pattern any transcript-replay diff degenerates to "everything added" by construction, which is exactly what the diff view kept showing.

**Streaming stability** carries over: aggregated rows keep their identity while content is unchanged (`sameMemoryChanges`), so the listing fetch and detail refetch fire only when a settled Task actually changed memory; the locate flash is driven solely by explicit `openMemory` commands.

## Coordination note

`chat-page.tsx` changes are additive and confined to the panel wrapper block (tri-exclusivity around the existing `swapPanels`), three close-site one-liners, the memory memo/listing block, ctx entries, one toolbar prop pair, and the panel render site. PR #376 (background-execution) touches the chat agents/process panels concurrently — whichever merges second rebases; no shared hunks are expected beyond `chat-page.tsx` proximity.

## Verification

- `memory-nav.test.ts` (9): entry routing, back, list building (markers; deleted topics absent once the listing is loaded; change-derived rows while it is not), deleted-key set (null while loading / exact set / empty).
- `memory-changes.test.ts` (13): classification, merge (write-dominant, latest atMs, cross-scope isolation), aggregation, `sameMemoryChanges` identity stability + content-move detection.
- `stream-model.test.ts` (107): collection, exclusions, Task boundaries.
- Full suite post-rebase (through the AgentHub 0.4.5 upgrade, #360): `pnpm --filter @prismshadow/penguin-web test` — 85 files / **1087 tests passed**; `pnpm typecheck` all packages; `pnpm format` + `format:check`; `git diff --check` — clean. Pushed with `--force-with-lease` after the rebase (own commits only).


hiyouga:
Reviewed the whole diff against `origin/main`, made the two UI changes you asked for, and pushed the review fixes as `e300eec3`. The feature holds up: the stream-model derivation, the pure nav model and the tri-panel exclusivity are all correct, and the edge cases I could reach are either already covered by the PR's own tests or now are. One premise in my brief was wrong and worth correcting up front — **`main` has not moved**: `origin/main` *is* the merge-base `9292d19b`, the branch is 7 ahead / 0 behind, and #362 through #385 (including #372 and #378) are already ancestors of this branch. `git merge-tree --write-tree HEAD origin/main` yields a tree byte-identical to `HEAD^{tree}`, so the merge is a literal no-op and there is nothing to rebase. The one thing standing between this and merge is the design-spec sync, which by `SKILL.md` belongs in its own design-repo PR.

## Changed on request

**(a) The settings link is labelled now** — `packages/web/src/features/chat/memory-view.tsx:186-203`.

Before: an icon-only button — a `GlyphIcon`, a `title` tooltip and an `sr-only` span, both carrying `S.chat.openAgentMemory`. Nothing visible but the glyph.

After: the glyph keeps its place and a visible label sits beside it; the `sr-only` span is gone (the visible text *is* the accessible name now) and `title` repeats that label verbatim rather than expanding on it, so the accessible name and the visible one agree (WCAG "label in name"). The string is also shortened, because the panel title one row above already says "Memory" / "记忆" — repeating "memory" in the link was the longest word doing the least work:

| | before | after |
| --- | --- | --- |
| `strings-en.ts` | `Manage memory in agent settings` | `Manage in agent settings` |
| `strings.ts` | `在 Agent 设置中管理记忆` | `在 Agent 设置中管理` |

Width behaviour, measured in the browser at the shared width's 320px minimum (`use-panel-width.ts:25`): en label 202.6px, zh 157.5px, inside a 320px panel whose content box ends at the `px-3.5` padding — neither truncates, neither overflows. Below that (only reachable by raising the browser text size) the label `truncate`s and the glyph stays `shrink-0`, so the row degrades instead of pushing wider.

**(b) The locate flash is gone** — same file.

Removed the `flashLocate` state, its 1600ms timer, the `transition-colors duration-700`, the `bg-brand-50 dark:bg-brand-900/20` tint, and the `rounded-lg` that existed only to shape it. The routing effect is now three lines and the detail container is a plain `overflow-y-auto` box.

I checked this really does match the files card rather than inventing a third behaviour: `message-files-card.tsx` → `chat-page.tsx`'s `onOpenFile` → `workspace-browser.tsx:366-373`, which does `setPath(dir)` + `previewPath(target)` and nothing else. No highlight, no landing affordance. The two arrivals now agree.

Two things that came out of removing it, both arguing the same way:

- `bg-brand-50 dark:bg-brand-900/20` was the **only** occurrence of that tint in `packages/web/src`.
- It was a static reimplementation of a gesture the app already has: `.outline-flash` (`styles.css:265-291`, applied by `conversation-outline.tsx:169`) animates `--color-brand-50` → transparent in light and **`--color-brand-950`** → transparent in dark. The inline version used `brand-900/20` in dark, so the two "you landed here" flashes would not have matched each other.

## Fixed

All in `e300eec3`.

1. `memory-changes-card.tsx:125` — raw `size={12}` on the show-more `Chevron` → `ICON_SIZE.chevronDense` (#378's scale; the sibling this card mirrors, `message-files-card.tsx:156`, already uses it, and the file already imports `ICON_SIZE`).
2. `lib/omni/stream-model.ts` — the pre-existing `session_meta` comment said *"For the main session it carries nothing the view model needs"*. This PR made that false: the main session's `agentState` comes from exactly there. Rewritten to say what both levels take from it, and the new assignment no longer wedges itself between that comment and the `model.nested` branch the comment documents.
3. Wording left over from earlier revisions of this PR, now describing code that no longer exists:
   - *"located at this row's **diffs**"* (`message-stream.tsx:51`) and *"A located **diff** needs the space"* (`use-memory-panel.ts:64`) — the PR removed diff rendering entirely.
   - *"the card renders those rows **unopenable**"* (`message-stream.tsx:53`, `use-memory-listing.ts:4`, `chat-page.tsx:571`) — the card **filters them out**: `memory-changes-card.tsx:45-46`, and hides itself when nothing survives.
4. `chat-page.tsx` — the aggregation comment claimed *"backfilled windows included"*. They are read, but each backfill builds a fresh model (`stream-controller.ts:510`) that derives rows only once a `session_meta` has told it where the Memory root is — which this PR's own test pins (`stream-model.test.ts`, *"without session_meta no rows are derived"*). The comment now states the condition instead of the headline. See the third open item below for the underlying gap.
5. Two-panel wording left at two by a PR that added a third peer — `chat-page.tsx` (module header, the exclusivity block's *"closes the other … owns both"*, the terminal-takeover and draft-reset effects, both `ctx` callbacks, the panel-switcher render comment), `panels-toolbar.tsx:5-9` (*"Three panels exist"* → four), `use-panel-width.ts:2-4, 15, 174`, `use-files-panel.ts:5, 15`, `use-subagents-panel.ts:6, 10`. The last three files are not in the PR's diff but are wrong *because of* it — all three now have a third `usePanelWidth()` subscriber.
6. `panels-toolbar.tsx:5` carried `"智能体面板"` inside an English-only comment. `CLAUDE.md` says translate residue when you touch the file, and this PR touches it.
7. `changelog/unreleased/2026-08-20-memory-change-list.zh.md` used ASCII `,` `(` `)` `:` `;` throughout; all six neighbouring zh entries use full-width forms. Normalised. No prose changed.
8. New `packages/web/test/frontmatter.test.ts` (8 cases). `lib/frontmatter.ts` is new — extracted verbatim from `memory-tab.tsx` — and shipped with no test at all.

## For you to decide

**1. The design spec is out of sync, and I did not touch it.** The *mechanism* this PR ships is exactly what `design/specs/06-PROTOTYPE.md` prescribes — one shared width, one persisted layout preference, sequenced close→open, open state surviving a Session switch, reset only on new chat. The *cardinality* is not:

- `06-PROTOTYPE.md:56` — "Workspace 文件与子智能体各为一个右侧停靠面板 … **两个面板**互斥打开 … 共用同一个宽度 … 只有新建对话会把**两个面板**都复位为关闭", plus the toolbar clause "工作区与智能体面板开关".
- `06-PROTOTYPE.md:67` — "智能体面板（右侧停靠，**与文件面板互斥**）" and the auto-open rule's "不覆盖已打开的**文件面板**" (the code's guard is now `!filesPanelRaw.open && !memoryPanelRaw.open`).
- `06-PROTOTYPE.md:78` — "工作区与智能体面板按钮只留图标".

And the Memory panel has no prose anywhere in `design/specs/` — `grep -rn 记忆面板 design/specs/` comes back empty. New text is needed for the two levels, the entry routing and the changes card. Worth knowing while you write it: the panel switcher and the terminal dock are *already* undocumented in that section, so this is a widening of an existing gap, not a new one.

**2. The changed-marker dot's colour.** `memory-view.tsx:249` paints the "changed in this conversation" dot `bg-brand-500`. The structurally identical dot in this PR's *other* file, `panels-toolbar.tsx:320` and `:405`, uses `toneDot.attention` from #378's vocabulary — same geometry (`h-1.5 w-1.5 rounded-full`), same "something happened here" job, different colour source. `tone.ts:18-22` enumerates its categorical exemption (chart series, token buckets, phase bars, skill avatars, terminal theme) and `brand` is not on that list. Options: (a) switch to `toneDot.attention` if "changed" is a state rather than an identity; (b) keep brand and add it to `tone.ts`'s exemption with the reason. Point in favour of leaving it: the brand dot carries an `sr-only` label that the toolbar's `aria-hidden` dots don't. I left it — it's a visual-vocabulary call past what Part 1 settles.

**3. Memory rows are missing from windowed history.** `collectTaskMemoryChanges` needs `model.agentState`, and only `session_meta` supplies it. The initial load is a **tail window** (`stream-controller.ts:368`, `TAIL_UNITS = 200`) and each scroll-up backfill builds a fresh model (`:510`). `session_meta` sits at ordinal 0 of each shard (`trace-service.ts:1250`), so a window whose start lands mid-shard carries none and its Tasks show no memory card at all. Under 200 turns the window covers the whole transcript and nobody ever sees this; a long session silently loses cards on older Tasks. Options: (a) accept it and say so in the changelog's limitations bullet, next to the `exec_command` one; (b) have `readMessagesPage` prepend the session's `session_meta` to every window (server-side); (c) carry `agentState` across model rebuilds in `stream-controller.ts` — partial, it only helps when a window happens to cross a shard boundary; (d) put the agent-state path on `SessionInfo` so the model never needs the meta. I did not touch it: (b) and (d) are the API-shape calls the brief reserved for you, and (c) alone would be a fix that looks complete and isn't.

**4. The header row is a lone right-aligned link.** Now that it has a label, the link is the only thing on its own row beneath the panel's "Memory" title, with ~120px of empty space to its left. It reads fine, but it could move up onto the title row or left-align. Pure layout judgement, so I left the row structure exactly as the PR had it.

**5. A pre-existing race the third peer widens.** `chat-page.tsx:626-645` (the subagents auto-open) is the only opening path that skips `cancelPanelSwap()` and calls the raw setter. Inside the 200ms `swapPanels` window every panel's `.open` reads `false`, so a live spawn landing there opens subagents and the pending timer then opens the other panel — two docked panels at once. `main` had the same race with two panels; this PR does not introduce it, and fixing it means reworking the auto-open effect. The PR did correctly add `!memoryPanelRaw.open` to the guard at `:637`.

## Checked and clean

**#372 compatibility (the semantic-conflict check).** `MemoryScopeInfo` (`server/src/api/types.ts:838`) still carries `kind`, `scopeKey`, `workspacePath`; `MemoryFileInfo` (`:854`) still carries `name`, `title`, `description`, `updatedAt`, `modifiedAt` — every field `buildMemoryList` reads. `hasIndex` (`:848`) is import-confirmation-only (its sole consumer is `memory-transfer.ts:102`), so the panel is right to ignore it, and `memory-nav.test.ts:36` already fixtures it. No spurious `MEMORY.md` row can appear: the index is excluded at the source (`memory-service.ts:114`), independently of `classifyMemoryPath`'s own filter — the two agree rather than one masking the other.

**Scope-key agreement**, the thing that would have broken deletion-marking silently: `classifyMemoryPath` derives `scopeKey` from the path segment under `memory/`, `MemoryScopeInfo.scopeKey` is documented as exactly that directory name (`types.ts:839-840`), and `getMemoryFile` hands it straight to `/scopes/:scopeKey/files/:fileName`. They line up, so `deletedChangeKeys` compares like with like.

**Stream-model edge cases, and where each is actually exercised:**

| Edge case | Covered by |
| --- | --- |
| Task fails or is aborted mid-change | `stream-model.test.ts` — `"failed"` and `"aborted"` stop reasons excluded via the `outputComplete`/`outputStopReason` gate |
| Index rewrite, non-memory path, opaque `exec_command` | same test |
| Two Tasks touching the same file | `memory-changes.test.ts:119` — write-dominant, latest `atMs`, first-appearance order |
| A change to a file later deleted | `memory-nav.test.ts:79` (leaves the list) and `:117` (exact deleted-key set), with `:84`/`:113` pinning "not loaded ≠ deleted" |
| Resumed session replaying history | `stream-model.test.ts` — no `session_meta` → no rows, and Task boundaries stop the backward walk. This is the behaviour, not a bug, but see open item 3 |
| Streaming identity stability | `memory-changes.test.ts:136-149`, both directions. Structurally it holds too: per-Task rows are computed once in `finalizeOpenTask`, never per chunk, and the session aggregate is pinned by the ref at `chat-page.tsx:549-553` |

**`lib/frontmatter.ts`.** The regex is a verbatim move out of `memory-tab.tsx`, so no behaviour changed, but the new test pins the cases that were unguarded: no block, unterminated block (body preserved), first-fence-wins when a `---` rule appears later in the body, CRLF, leading BOM, empty string, and block-with-nothing-after.

**Panel peer relationship.** `use-memory-panel.ts:47` calls the same `usePanelWidth()` as its siblings — one module-level zustand store, one `penguin.panelWidth` key (`use-panel-width.ts:26, 65`), with only `resizing`/`panelRef` per instance. No new preference, no third `useState` fighting over the key. Tri-exclusivity is symmetric: each of the three wrappers names the other two and each close branch ORs the other two (`chat-page.tsx:399-434`); terminal takeover (`:449`) and the draft reset (`:604`) both gained the third close. Pinning is generic over `PanelKey` — `panels-toolbar.tsx:55` filters a stored list against `PANEL_ORDER` — so a stored `["agents","workspace"]` loads unchanged and no migration is needed or missing.

**Conventions.** 13 string keys added to both dictionaries, same names, same order, same arity, every zh value translated (`memoryEmptyAll`'s lowercase `agent` matches the existing `S.memory.emptyUserScope`, not a slip). No Chinese outside the zh dict, the `.zh.md` entry, and its `[中文版]` link. No design-doc citations in comments. `git diff --check` clean. Changelog pair: metadata in fixed order, counterpart links both ways, sections mirror one-for-one, none of `## Why` / `## Verification` / `## Risks`.

**Verification run.** `pnpm --filter @prismshadow/penguin-web test` → 86 files / **1095 tests** (1087 before, plus the 8 new frontmatter cases). `pnpm -r typecheck` → all 8 packages clean. `pnpm format` then `format:check` → clean. `disclosure-anchor.test.ts` passes; the PR introduces no `InfoPopover` at all.

**Driven in the real app**, because both changes are things you judge by eye: built server on a scratch `PENGUIN_HOME`, port 9143, Chromium at `deviceScaleFactor: 2`. Seeded three User-scope topics on disk, opened the Memory panel from the switcher dropdown at 480px and at the 320px minimum, in both locales, and clicked through to a detail in each. What I saw: the labelled link reads cleanly at both widths in both languages with no truncation; the list renders the three topics with their frontmatter titles, descriptions and dates; the detail renders the description, the body and real Markdown with the frontmatter stripped, on a plain background — I asserted zero `bg-brand-50` / `bg-brand-900` nodes anywhere in the panel subtree in all four runs.

Not merged, per your instruction.

PR #376, opened 2026-08-20
feat(core,server,web,docs): background execution with completion reports and kill tools

hiyouga:
## What this adds

**`run_in_background` on `exec_command` and `run_subagent`.** The call registers the session and returns its `process_id` / `subagent_id` immediately — no foreground window. When the task settles, its result reaches the conversation **as a user message injected by the harness**, so the model never needs to poll: a `[background_task_done]` marker block (kind, id, status, one-line detail) followed by what ran and the tail of its yet-undelivered output (capped at 4000 chars). The Web App collapses the block into a one-line banner with the report body beneath it.

**Delivery design.** The Session keeps a single notice queue fed by the Environment's completion events (`Environment.setBackgroundTaskListener` → `Session` → engine). Delivery consumes from that one queue exactly once, along three paths:

- *Task running:* the engine drains the queue at every input-assembly boundary (the same points steering uses — run start, post-turn, post-compaction), yielding each notice to the stream and writing it to the Trace. A final reply that already streamed doesn't lose the notice — the Task continues for one more turn to react to it (`!midTask && injected.length === 0` now counts notices).
- *Session idle (server):* core fires `Session.onBackgroundNotice`; the server manager revalidates under the session lock, takes the queue (`takeBackgroundNotices`), and launches an ordinary Task with it. Drive's finally also sweeps notices that raced a run's exit.
- *No host subscriber (SDK):* the next run's start drains the queue into the first request.

**Two new tools.** `kill_command` (SIGTERM the process group, SIGKILL after the grace period; returns undelivered output; removes exited sessions too) and `kill_subagent` (aborts the run, denies pending approvals, frees the concurrency slot). Both disarm the completion report first — the kill's own result already carries the outcome. `kill_subagent` is filtered out for depth-capped leaf agents alongside the other subagent tools.

**`sender` on user texts.** `TextPayload` gains optional `sender?: "user" | "parent_agent" | "harness" | "server"`. The parent agent's `run_subagent` / `input_subagent` prompts record as `sender: "parent_agent"` in the child's Trace; harness-injected completion reports as `sender: "harness"`; the server's scheduled-task triggers as `sender: "server"`. Strictly additive: absent = the human user, which is also the correct reading of every pre-existing Trace — no migration, no compatibility code, and the field never reaches the provider wire (`payloadToContentItem` picks fields explicitly).

**`input_command` default.** The empty-poll default wait rises from 5000ms to 120000ms (clamp ceiling `timeoutMs: 130000` still clears it), so one poll waits out most builds; output still streams as it arrives and the wait ends early on exit.

**Kernel bump.** The default toolset changed substantively (5 `tools.builtin.*` leaves), so `KERNEL_VERSION` advances to `2026-08-20` with a new pinned-hash generation. Existing agents adopt the new tools/arguments through the settings page's kernel update (user customizations kept per leaf); until then they keep their stored toolset — same convention as when the file tools were added. Bilingual docs (`tools`, `interfaces`, `omni-message`) updated, including a new "Background completion reports" section.

**Service links on the process list** (moved here from the closed PR [#373](https://github.com/Prism-Shadow/penguin-harness/pull/373)). #373 scanned the web transcript, which cannot work for background launches — a `run_in_background` command's URL never reaches any tool output. Detection now lives in core with two sources: the command session incrementally scans its own output stream for local URLs (ANSI-stripped across chunk boundaries, wildcard hosts normalized, latest hit wins — foreground and background covered alike), and when nothing was printed, a listen-port probe of the session's process group synthesizes `http://localhost:<port>` (linux `ss`+`ps`, macOS `lsof -g`, Windows `Get-NetTCPConnection` + the CIM process tree — the same process-ownership rules the kill paths use; multiple ports collapse to the smallest for stability). The printed URL outranks the probed origin; probes run when the processes route is fetched, TTL-cached per session with a hard 1.5s timeout, a failure keeping the previous answer and a clean empty probe clearing it. The processes API rows gain an additive `serviceUrl`; the web panel links it next to the pid on running rows (new tab, tooltip with the full URL).

**Notification-visibility audit** (reported: the completion report "did not appear in the frontend"). A new end-to-end test — real core `Session`, real `Environment`, a real OS process, the real `SessionManager` via its loader, scripted LLM only — proves the full chain: a background command finishing after the session went idle lands on the session's event channel (the exact SSE feed) as a `[background_task_done]` user message with `sender: "harness"`, auto-starts a task whose model input contains the report, and broadcasts both `task_state` flips. The chain is sound; the reported repro matches the **kernel-gating precondition**: an agent whose stored `system_config.yaml` predates this branch has no `run_in_background` in its `exec_command` schema, so the model can only use the foreground-promotion path — recognizable by its note still reading `use input_command to send input or poll for output` (the background launch's note says the completion "will arrive as a user message"). Running the settings page's kernel update (or creating a fresh agent) enables the new arguments and tools. The audit still hardened two real loss windows: idle eviction now pins runtime entries whose Session holds undelivered notices, and the notice auto-starter logs instead of silently returning when the entry is gone.

**Fix: completions now reach an open page — the real root cause.** The earlier audit proved the chain through the loader path, but production creates sessions through `adopt` (POST /sessions), which never registered the idle-arrival notice listener — so a fresh session's completion reports queued forever and the chat page sat unrefreshed. Both entry paths now share `registerNoticeListener`, with an adopt-path regression test, and a **Playwright end-to-end test** (`background-notify.spec.mjs` + the mock's "background notify test" branch) drives the real server and browser: send → `run_in_background` launch → task ends → command exits ~6s later → the `[background_task_done]` banner and the model's acknowledgement turn appear in the open page with **no interaction** (it also asserts the launched command shows in the header's running-services count — same registry as promoted processes).

**Background subagents in the subagents panel.** `SubagentHandle` gains a one-shot `takeMeta`; the background branch forwards the child's origin-tagged `session_meta` at launch, so the panel shows the node (and the server registers the child) before any poll. The panel derives a background node's running state from the harness's own notes — the launch/still-running markers, `[background_task_done]` notices, idle polls, kill notes; last mention wins — and completion notices no longer count as panel-scope boundaries, so the node visibly settles instead of vanishing when the report lands. Live message forwarding still starts at the first poll (unchanged trade-off; the child's own Trace is complete throughout).

**Trace page.** After review, a notice-triggered task stays its own round (an earlier merge attempt left an ugly idle gap in the merged timeline and was reverted), and no user-message marks are drawn on the timeline (that experiment is fully removed — lane, rendering, derivation, tests and strings). `timeline-chart.tsx` ends this PR effectively untouched, so no conflict with PR [#362](https://github.com/Prism-Shadow/penguin-harness/pull/362) remains.

**Fix: background subagents were unusable — their tools parked at approval forever.** Reproduced first (a failing e2e spec, then the child's trace: `exec_command` CALL with no decision and no output — the reported instant `aborted` is that park meeting any later abort). Signal lifecycle after the fix: the child's abort signal is its own (`ManagedSubagentSession.abortCtrl`) from birth — the launching call's `ctx.signal` never reaches it, and only `kill_subagent`, Session end, or registry eviction aborts it; what WAS coupled to the launching call was approvals — the child's requests queue for a window sink that only a `run_subagent`/`input_subagent` collect window attaches, and a `run_in_background` launch never opens one. The launch now attaches its own `ctx.approve` as a **standing sink** (window sinks still take precedence while attached), so an allow-all launch runs unattended. Foreground semantics are untouched.

**Live streaming + failure reports.** Background children now stream to the frontend in real time: a forwarding tap replaces the poll buffer at launch (`Environment` → `Session.onBackgroundMessage` → channel publish + usage recording, same origin-tagged shapes a foreground window relays), continuing past the launching turn's end until the child's terminal state; a later `input_subagent` poll no longer re-delivers what already streamed. Every terminal state reports — a child whose run fails (e.g. its provider rejects) still sends the completion notice with `status: failed`, so the panel settles instead of showing "running" forever.

**Notice rendering.** The chat notice is a collapsible one-liner in the work group's own "运行完毕 / Done" summary-bar form: the header's chrome, uppercase title styling and scrollport sticky (`-top-4 z-[5]`) are extracted verbatim into `disclosure-row.tsx` as header-family constants (alongside the row-family ones the thinking block and tool cards use, the tool cards' output-`<pre>` class and the card chrome), the work group references them, and the shared `DisclosureRow` gains a `header` variant the notice renders through — visually indistinguishable from a settled group header apart from label and icon semantics, and it pins while scrolling exactly like one. The thinking block keeps the row variant (byte-identical classes and DOM order, same collapse-scroll behavior). Collapsed the notice shows only the outcome ("后台命令完成" / "后台任务失败" — Background command finished / Background task failed); expanded, the handle, exit detail and output tail render in the tool cards' output styling.

**Real-app evidence (Playwright, real server + mock LLM).** `background-subagent.spec.mjs`: success path — launch, parent task closes, child works on (its trace: approval `allow`, `exec_command` → `completed`), the collapsed notice and the model's acknowledgement appear with zero interaction (2.5s); failure path — the child's requests are rejected as an auth error, its run fails at once, and the **failed** notice still reaches the open page unprompted (0.8s). `background-notify.spec.mjs` (command path) passes against the same build.

## Verification

- `pnpm build` and `pnpm typecheck` clean across the workspace.
- New `packages/core/test/background-execution.test.ts` (17 tests): marker round-trip and title-strip, `userText` sender marking, real-process background completion reports (success / non-zero exit / pre-listener buffering / dispose suppression), `kill_command` semantics (drain + no report + unknown id), fake-runner subagent background launches (per-round reports, `kill_subagent`), engine boundary delivery (pre-run drain rides the first request; a mid-run notice extends the Task by one turn), and Session queue/signal wiring.
- Updated pinned assertions: kernel hash history (new generation), `TITLE_NOISE_TAGS`, `buildToolConfig` tool lists, `exec_command` property list.
- Second `origin/main` merge kept the branch current (no conflicts). E2E: `background-notify.spec.mjs` passes on the built app (6.9s); `processes.spec` and the other `subagent.spec` cases pass — the one `subagent.spec` panel-lifecycle failure reproduces identically on a build **without** this branch's web changes (pre-existing local flake). New unit coverage: adopt-path listener regression (server), upfront-meta forwarding (core), background topology states and notice scope rules, `noticeTaskRemap`/`userMarksOf` (web).
- Merged current `origin/main` into the branch (no conflicts). Service-URL: core `service-url.test.ts` — 20 tests (extractor, incremental scanner across chunk/ANSI boundaries, per-platform probe parsers over fixtures, a live linux probe against a real listener, and the exec_command→list wiring); the server processes-route test asserts the additive `serviceUrl` and that the route refreshes probes before listing.
- Server: `session-manager.test.ts` gains the end-to-end notice test above plus an idle-arrival auto-start test (notice taken exactly once, published to subscribers, empty-signal no-op); full server suite 788/789 — the one failure is `terminal-stream.test.ts`'s PTY replay assertion, which fails identically on a clean `origin/main` checkout under full-suite load and passes standalone, i.e. a pre-existing load flake unrelated to this diff.
- Core suite 856 passed / 5 skipped; web suite 989 passed. `pnpm format` + `format:check` clean.


hiyouga:
## Review — background execution

Reviewed in a dedicated worktree on `feat/background-execution` (`0e3e80cd`). Baseline before my changes: `pnpm -r build`, `pnpm typecheck` clean; core (`background-execution`, `service-url`, `markers`, `state`), server (`session-manager`, `session-processes`) and web (`agent-topology`) suites all green.

The delivery contract holds up under scrutiny — exactly-once across the three paths (engine run-start drain, boundary drain, server idle auto-start plus `drive`'s finally sweep), disarm-before-kill in both kill tools, `Environment.dispose()` suppressing before it kills, and `ServiceUrlScanner`'s chunk/ANSI boundary handling. What follows is what did not.

**I pushed fixes for 4 findings** (commits on this branch, see the bottom). The rest are reported only — each needs a call that is yours to make.

---

### 1. A background subagent's tools are auto-denied under the default approval mode

`run_subagent`'s background branch installs the launching run's `ctx.approve` as the standing sink. On the server that callback closes over `entry.approvals`, and `drive`'s `finally` runs `entry.approvals.denyAll()` at the end of **every** run (`session-manager.ts`, also `interrupt()`). With `approvalMode` = `always-ask` (the server default) or `read-only`, the child's first read-write tool routes to `manual()` → `registry.wait(...)`, and the launching task ends seconds later → the request resolves as **`deny`**.

So the fix in "background subagents were unusable" trades a permanent park for a silent denial: the child's first `exec_command` is refused moments after launch, and any later unrelated task ending denies whatever the child has pending at that moment.

Reproduced with a throwaway server test — not committed — a fake `RuntimeSession` that hands its `opts.approve` to an out-of-band caller during the run, with `approvalMode` left at the `ROW` default `always-ask`:

```
AssertionError: expected 'deny' to be 'allow'
Expected: "allow"
Received: "deny"
```

All three new Playwright cases pin `approvalMode: "allow-all"`, so nothing in the suite covers this. The docs already scope the claim correctly ("so an `allow-all` launch runs unattended") — but the tool is exposed in every mode.

This is the one finding I would want settled before merge. I did **not** fix it: the honest fix changes approval semantics (an approval raised by detached background work should survive a foreground run's end, and only `interrupt`/close should converge it), which means either `ApprovalRegistry` learning ownership or core telling the server which origins are detached. That is your design call, not a review edit.

### 2. A working background subagent no longer strands its Session — fixed

`sweepIdle` pinned entries on a running background **command** and on pending notices, but not on a running background **subagent**. A child outliving `ENTRY_IDLE_MS` (30 min) with the session idle got its runtime entry deleted; from there `forwardBackgroundMessage` returns early (live messages dropped on the floor) and the completion notice queues onto a core `Session` no longer in `this.entries`, so `startBackgroundNoticeTask` takes the `runtime entry not loaded` branch and the report is lost. The next access resumes a *fresh* Session/Environment, orphaning the still-running child.

Fixed by adding `hasRunningBackgroundSubagents()` along the existing `listBackgroundCommands` seam (`SubagentSessionManager.hasRunning` → `Environment` → `EnvironmentInterface` → `Session` → `RuntimeSession`) and pinning on it, plus stamping `entry.lastActivityMs` in `forwardBackgroundMessage` — a child producing messages *is* session activity, and without the stamp the idle clock only ever measured the launching task. Covered by a new `sweepIdle` test and two assertions in the existing core round-report test.

### 3. The header cost chip loses the pre-notice half of a Task — fixed

`taskStartCount` stopped counting `[background_task_done]` texts, but the reducer still calls `startTask()` for them (`stream-model.ts`), which runs `resetTaskCounters(model.stats)`. `header-stats.ts` documents `taskCount` as "1:1 with the model's `startTask` calls" and folds `lastLiveUsd` into `settledUsd` only when it increases or the Task closes. On a **mid-run** notice neither happens — `finalizeOpenTask` + `startTask` are one synchronous step, so `taskOpen` is never observed false — and `hold.lastLiveUsd` is overwritten with the freshly zeroed bucket. The pre-notice live cost disappears from the chip until the next authoritative usage fetch.

Fixed by splitting the two meanings: `taskStartCount` keeps the panel-scope semantics you introduced, and a new `modelTaskStartCount` (every `user_text`/`user_image`, notices included) feeds `advanceCostStat` — restoring exactly the pre-PR cost behaviour. Test extended.

### 4. Hot-path marker parse on every render — fixed

`isScopeStart` ran `parseBackgroundTaskDoneMessage` (two full `[\s\S]*?` scans) on **every** `user_text`, and `taskStartCount(stream.model.items)` is recomputed on every `ChatPage` render — many times a second while streaming. Added an `includes("[background_task_done]")` prefilter so only text that actually carries the tag pays for the parse.

### 5. A background child's failures never reach the error store

`forwardBackgroundMessage` publishes and records usage but bypasses `drive`, so its messages never reach `StreamErrorWatcher` — and the comment there explicitly says "subagent failures flow through this same stream too". A background child whose provider request fails (exactly the path `background-subagent.spec.mjs` now covers) gets no persisted LLM/tool error. Same bypass drops `child.assistantExcerpt`, so background children only ever get prompt-fallback titles.

Not fixed: the watcher is per-run and closed in `drive`'s finally, so covering this needs an entry-scoped watcher — again a design call.

### 6. An `input_subagent` poll on a background child returns an empty tail

`pump`'s `finally` calls `wakeSignal.notify()` and then `fireSettleWatcher()` **synchronously**. The notify only *schedules* `collectWindow`'s continuation, so the settle watcher's `session.drainText()` runs first and `collectWindow`'s "final drain" yields nothing. Most of the answer still streams incrementally, so this is only the tail between the window's last drain and the settle — but the poll's own result reads as if the round produced nothing, with the text arriving in the notice instead. `queueMicrotask(() => this.fireSettleWatcher())` in that `finally` would order it after the awakened window (`onSettled` already uses `queueMicrotask` for the already-terminal case) — for background sessions `drainMessages()` is always empty because of the tap, so the continuation reaches `drainText()` without an intervening `yield`.

### 7. `input_command`'s new default applies to agents that never ran the kernel update

`DEFAULT_EMPTY_POLL_YIELD_MS` is a **code** constant, not part of the stored schema. An existing agent's `system_config.yaml` still describes "empty polls default to 5000" while the code now waits 120 s — the model reads one number and gets a 24× longer block, with no kernel update involved. The PR's compat story ("existing agents keep their stored toolset") holds for the tools and arguments but not for this one. Per the repo's backward-compatibility rule this is the kind of thing to decide explicitly rather than let ride: gate the default on the stored description generation, accept it and say so in the changelog, or leave the constant and bump only the description.

Related and smaller: a 120 s empty poll blocks the turn, so user steering queued during one waits out the full window. Deliberate, worth stating in the docs.

### 8. The Windows service probe will almost always time out

`probeGroupListenPorts`'s win32 branch spawns `powershell -NoProfile -Command` running `Get-CimInstance Win32_Process` (every process on the box) under `PROBE_TIMEOUT_MS = 1_500`. PowerShell cold start alone routinely eats most of that budget. The failure is benign by design (`null` = "don't know"), but in practice Windows would only ever get `serviceUrl` from printed output, never from the probe. Worth either a platform-specific budget or narrowing the query. (I could not measure this from Linux — flagging, not asserting.)

### Nits

- `pumpApprovals`' comment says "window sinks still take precedence while attached", but the `pumpingApprovals` re-entrancy guard means a newly attached `input_subagent` window cannot take over a request already in flight on the standing sink; it only gets a turn once that consultation resolves. The behaviour is defensible — the comment isn't.
- `BackgroundDoneBanner` renders `` `${done.id}\n${body}` ``, and `body` already opens with the head line containing `(process_id proc-…)`. The handle prints twice.
- The PR description says `KERNEL_VERSION` advances to `2026-08-20`; the code (and the changelog entry) say `2026-08-21`.
- `interfaces.{en,zh}.md` listed `backgroundDone` on `EnvironmentServices` but not `backgroundForward` — added in the pushed commit.
- `extractLastLocalUrl` scans all output, so a command that merely *prints* a localhost URL (a `curl`, a README, a test log) gets a service link on its running row, and the printed-URL branch permanently suppresses the port probe. Acceptable heuristic, worth knowing.

---

### Pushed on this branch

- `227ba88d` `fix(core,server): pin a Session's runtime entry while a background subagent works` — finding 2
- `b7c3f7c7` `fix(web): keep the cost chip's Task boundary on the model's own count` — findings 3 and 4
- `ca61974b` `docs(interfaces): list backgroundForward on EnvironmentServices`

Verification after the changes: `pnpm -r build`, `pnpm typecheck`, `pnpm format` + `pnpm format:check` clean; full suites green — core 888 passed / 5 skipped, web 1063 passed, server 813 passed (the `terminal-stream` replay flake the PR body mentions did not reproduce in this run).

PR #383, opened 2026-08-20
feat(plugins): list, install, and manage plugins

Myriad-Dreamin:
> **Stack:** `main` ← #599 ← #546 ← #656 ← **this** ← #674 ← #672 ← #551 ← #718 ← #719 ← #544 ← #448 ← #450 ← #577 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #637 ← #628 ← #642 ← #639 ← #645 ← #641 ← #526 ← #726 ← #727 — one linear chain, in this order since 2026-09-14. Review the diff against `refactor/parked-platform-state`.

> **Reworked on 2026-09-10** after a review of the whole diff. Three things left it: installing from npm (its own PR, #672 — this platform ships builtin plugins only, and the first version's removal did not survive the next `npm install`), the Sandbox settings page (#673 — it drives a service that predates this PR and shares nothing with the platform), and the content-addressed push (down into #656, the layer that pushes). What stayed was fixed in place: the installed-plugins dialog installed through PUT and never went through the install rule; a plugin whose import threw showed as "restart pending" instead of its reason; the collapsed rail listed /plugins twice; the e2e specs still queried the old nav label; saving one plugin's options wiped every other form's draft and a number box snapped empty mid-keystroke; a kernel visibility change had hidden inside the rename commit with no test or mention (reverted); copy, docs and comments still said `plugins.json`, `/api/plugins` and "at the next restart". The section below is updated accordingly.

> **This PR is the plugin platform, in its final form.** On 2026-09-09 the work that used to be spread over later PRs was moved down here as real commits, so a reviewer reads each piece once instead of a first version and its fixes: the plugins a deployment installs (eight commits that were #551's last section), one plugin list per Project applied without a restart (#627, without its machine half — that is in #448), the platform loading its own plugins (#635) with the two push fixes that followed (#636's), and the plugin configuration mechanism split out of #641 — since moved on again, to #674. The tree at the top of the chain is byte-identical to the tree before the move. The sections below describe the platform as it stands; the marketplace section is the original one.

## What a deployment installs, and what a Project asks for

Builtin plugins ship in every build as the npm packages they are — `scripts/build-plugins.mjs` runs each `plugins/*` package's own build, packs it with `pnpm pack` and has `npm install` lay the tarballs out as an npm prefix (dependencies included, nothing rewritten) that rides the push's assets and the desktop build; the loader looks a package up the way Node does and reads the entry its `exports` names, and a readme is the `README.md` npm shipped with the package, read on request — as a *tag*, not an install — the operator still consents from the catalogue row, and a catalogue row says what the plugin is for this server (install / the restart it waits for / running / why it failed to load). Asking for a plugin the build does not ship is refused (`plugin_not_shipped`) rather than listed, so a Project never names a package that is not on the machine; fetching one from npm is #672.

The list then stops being a per-deployment file: **a Project owns its plugin list** — the `[plugins]` table of its `.project_config.toml`, package name → requirement in the shape of Cargo's `[dependencies]` (`"@scope/name" = "*"`, a version string, or `{ version = "…" }` where later fields go; a table rather than `string[]` so an entry can grow without a format change, and the list form the key briefly had is not read — PRFC-0010 on design #124 says the same), the process runs the closure over its Projects, and a write applies without a restart: the App re-assembles itself — `platformImpl` boots the App proper as an inner instance behind a shell the runtime holds throughout, and `Reassembly.reassemble()` runs the kernel's own `upgrade` over that inner instance (a boot that fails is recovered onto the previous document and reported as "did not take"). Nothing in `packages/hmr` changes for this, so it works on every runtime; PRFC-0010 (design #124) describes the same shape. `GET|PUT|POST|DELETE /api/projects/:p/plugins/installed` replace the deployment-wide routes; `plugins.json` in the data root is no longer read (the changelog's Breaking line).

## The platform loads its own plugins

`loadPlugins` used to run in the HMR layer at process start, so a machine whose program predated a loading rule could never learn it from a push. The platform now reads the closure itself; what stays parked is the imported objects, keyed by the resolved file so a push that moves the builtin plugins to a new assets directory imports them again rather than running the previous build's code forever, and a push loads the plugins of the version it is booting with, not the previously committed one. Test-only plugin objects reach the host through their own parked id (`PARKED_TEST_PLUGINS_RESOURCE_ID`), never down the production path.

## What is not here any more

- Installing from npm and `npm uninstall` on removal: #672, based on this branch.
- The sandbox settings: #673, since merged into #674, where they are a settings group on the Plugins page (no route of their own).
- The blob store, `/api/hmr/assets/probe` and the sparse push: #656, section 8.
- Plugin configuration (`package.json#penguin.configuration`, `GET|PUT /api/admin/plugin-config`, the `PluginConfig` mechanism and the Settings Plugins page): #674, inserted into the chain right above this PR because #641 uses it.

## Marketplace (the original PR)

Extension discovery in the Web App: a **registry** concept on the server, a **shared index format**, a **listing** and a **detail page with the readme**. The four sandbox backends from #354 are what it lists, so this is stacked on that PR.

The surface speaks `extension` throughout — route, page, types and folder — the same word #353 gave the mechanism and `extensions.json` gives the config file.

## What

- `packages/server/src/extension/registry.ts` — `ExtensionRegistry = { source; index(); readme(name) }`, two implementations:
  - **builtin registry**: serves the index embedded in the server package (`builtin-index.json` — the four sandbox backends the workspace ships) with readmes from `builtin-readmes.ts`;
  - **http registry**: fetches an `index.json` URL (injectable fetch) and runs it through the same validator, so a remote index is trusted no further than the embedded one.
- **Shared index format**, modeled on typst/packages' `index.json`: a flat array of per-version entries (`name`, `version`, `description`, `authors`, `license` + optional `repository` / `homepage` / `keywords` / `categories` / `updatedAt`), typed as `ExtensionIndexEntry` in `api/types.ts`. An entry's `name` is the specifier an operator writes into `extensions.json`; the target OS rides `keywords`, the capability floor rides `categories`.
- `GET /api/extensions` and `GET /api/extensions/readme?name=…` (skills-route pattern, any logged-in user). The registry list is fixed to the builtin one for now — merging additional sources is the named future step, in this route.
- Web: a single-column listing (`features/extensions/extensions-page.tsx`) where the whole row is a link, a detail page (`extension-detail-page.tsx`) rendering the readme as Markdown, nav key `extensions` after `models` in `NAV_GROUP_KEYS` (+ the hand-maintained collapsed rail), routes `/extensions` and `/extensions/*`, bilingual labels 扩展市场 / Extensions.
- The registry lives in `src/extension/`, beside the host and the loader, rather than a `src/extensions/` of its own: one folder owns the concept, and an `extension` + `extensions` pair would have been a coin flip at every import.

## Design notes

- **Layer verdict** (hmr/README three questions): registry, index format, and both endpoints are business-side (`http/routes/` + `app.ts` mount) — deliverable by a platform push, zero runtime-directory changes. The pages are web-package UI.
- **One column, because the specifier is the identifier.** It is long, scoped and monospace; two columns truncated exactly what an operator came to read.
- **Readmes are fetched per entry, not carried in the index.** The listing is sent in full on every visit; a readme is large and wanted only for the entry someone opened. The endpoint answers only for entries the deployment lists — a 404 otherwise — so it cannot be used to probe what exists. A registry with no readme for an entry answers `null` rather than a guessed URL.
- The detail route is a **splat** (`/extensions/*`): specifiers are scoped (`@scope/name`), so a single path segment would not hold one.
- **Discovery only, by design**: no install action anywhere. Installing stays the operator-side `extensions.json` edit (`extension/loader.ts`), and no registry code ever imports extension code.
- **Strict whole-document validation** for an index, deliberately unlike `extensions.json`'s per-entry tolerance: an index is one publisher's single artifact, so a malformed row means the artifact is broken; `extensions.json` entries are independent operator choices, skipped one by one.
- The builtin index is validated through the same parser at serve time, so a broken embedded index fails loudly in tests rather than serving garbage.

## The readmes are content, and want reading

Each backend's readme is that package's own `README.md`, added with the packages in #354 and written from the backend's module doc and source: bubblewrap's mount ordering and unprivileged-user-namespace requirement, Seatbelt's last-rule-wins SBPL and path canonicalization, MXC's Windows-only mapping onto all three dimensions and its optional ~40MB peer SDK, and the DSH adaptor's platform chain and `fs-write`-only vocabulary. Nothing verifies prose — a reviewer should.

A readme is read from the package on this machine when its entry is opened (the shipped prefix, the data root's own, or the installation), so the harness depends on no backend at build time and the catalogue holds no copy of the prose at all. (The first version inlined the four files into the bundle; that went with the packing rewrite.)

**What the catalogue asserts is now checked against the packages.** A test pins every listed entry's `name`, `version`, `description` and `license` to what that package's `package.json` declares, that its readme is byte-for-byte the package's own file, that the package ships it, and that no readme exists for anything unlisted. It caught the listing claiming `Apache-2.0` for four packages that declared no license at all; they declare it now, in #354 with the readmes. The specifiers are `@prismshadow/penguin-extension-sandbox-*`, which is what the four publish as; that field is the string an operator copies into `extensions.json`, so a wrong one names a package that would not resolve.

## Verification

- `packages/server/test/extension-registry.test.ts`: shared-format validation including rejection cases, the builtin registry's four entries, the http registry with stubbed fetch (HTTP error / non-JSON / malformed document), the readme endpoint's listed-only rule (404 for an unlisted name, 400 with no name), and both endpoints behind the auth gate (401 → 200).
- Server **1523 / 110 files**, web **1533 / 111 files** (includes the updated `nav-group-collapse` manifest pins); `build`, `typecheck` (0 errors) and `format:check` clean across the repo.

## Follow-ups (out of scope)

- Stale comment at `packages/web/src/features/agents/skill-import-source.ts:5` ("PenguinHarness has no plugin mechanism of its own") is now wrong and worth a sweep.
- Multi-source registry configuration is #526.


packages/server/src/hmr/platform.ts
```diff
@@ -224,194 +231,312 @@ export const DECLARED_RESOURCES: ParkedInterfaces = {
...
+            impl: innerImpl,
+            iface: PlatformIface,
+            resources: ctx.resources,
+          });
+          if (result.status === "ok") {
+            inner = result.instance;
+            return true;
+          }
+          if (result.status === "failed") {
+            inner = await boot(innerImpl, PlatformIface, result.doc, ctx.resources);
```
Myriad-Dreamin, line 503: **A deterministic plugin boot failure during re-assembly is not recoverable.**

When the new tree fails to boot (`result.status === "failed"`), this restore `boot()` re-runs the same `createInner` against the same closure that was already written to the TOML by `setPlugins`. It throws identically, only the `finally` guards it, and `inner` is left pointing at the tree the kernel already disposed (`core/kernel/upgrade.ts:166`).

Trigger: POST a shipped plugin whose module `create()` throws, or whose manifest `requires` an interface the tree lacks. `loadPluginHost` only isolates import / table / name-clash failures (`loader.ts:462-501`); `bootModules` throwing `ModuleBootError` is not isolated. Result: the route answers 500, every later `outer.http` forwards to a tree whose instances were cleared, and the next process start fails at `bootAppDeps` for the same reason until the TOML is hand-edited.

There is no test that exercises `reassemble()` at all. Suggest: write the TOML only after the re-assembly succeeds (or restore the previous list before the restore boot), and add a failing-boot test.

packages/server/src/hmr/platform.ts
```diff
@@ -224,194 +231,312 @@ export const DECLARED_RESOURCES: ParkedInterfaces = {
...
+  // effect, so it is exposed as api.drained(), which the KERNEL awaits between dispose
+  // and the successor's boot (kernel/upgrade.ts). Nothing about the handover touches
+  // the registry.
+  let drained: Promise<void> | undefined;
+  ctx.effect(() => {
+    // Forwards to machines are DELIVERED, not suspended: the ssh children are separate
+    // processes that keep forwarding across the swap, and the successor adopts them by the
+    // pid recorded in web.db (machines/service.ts).
+    const drains: Promise<unknown>[] = [];
+    if (manager !== null) drains.push(manager.shutdown(DRAIN_GRACE_MS));
```
Myriad-Dreamin, line 388: **A per-Project plugin install / remove re-assembles the whole App, which runs `manager.shutdown()`.**

`reassemble()` → kernel `upgrade()` → `inner.dispose()` → this effect → `manager.shutdown(DRAIN_GRACE_MS)` (`session-manager.ts:1549-1556`: every pending approval is denied, every drive across all entries is aborted) and `tree.dispose()`.

So an admin clicking Install for Project A aborts a run that a member of Project B has mid-flight, with no confirmation on the page and copy that only says "applies without a restart" (`strings-en.ts:1333`, `docs/server-api.en.md:209`, `endpoints.ts:1348`). The comment at line 344 accepts this for a hot push; this PR extends it to an ordinary list edit.

This is a design call rather than a local fix: either the plugin closure becomes swappable below the App, or the page gets a confirm step and honest copy.

packages/server/src/http/routes/plugins-installed.ts
```diff
@@ -0,0 +1,234 @@
...
+    await deps.apply();
+    return c.json(await view(projectId));
+  });
+
+  app.put("/", async (c) => {
+    requireAdmin(c);
+    const projectId = scope(c);
+    const body = await readJson(c);
+    const list = body.plugins;
+    if (!Array.isArray(list) || list.some((s) => typeof s !== "string" || s.trim() === "")) {
```
Myriad-Dreamin: **PUT skips both gates that POST enforces.**

POST runs the specifier through `specifierOf` (shape check) and the "only a shipped package" check; PUT only rejects empty / non-string entries. So `PUT {plugins:['/tmp/anything.js']}` answers 200, `parsePluginTable` keeps any trimmed key, and on `apply()` `resolvePlugin` (`loader.ts:232-235`) treats an existing absolute specifier as the entry file and `importPlugin` (line 275) imports it into the server process. A relative `../x` falls through to a bare `import(specifier)` at line 277.

DELETE cannot remove it either: its query does go through `specifierOf` and answers 400, so only another PUT or a hand edit removes the row. The header comment says "never a path, a URL or a version range". `plugins-installed.test.ts` covers unshipped *names* via PUT, not paths.

packages/server/src/http/routes/plugins-installed.ts
```diff
@@ -0,0 +1,234 @@
...
+    // that is not on disk is exactly the state this route exists to avoid.
+    const shipped = await discoverBuiltinPlugins(pluginBases(deps.root, deps.assetsDir()));
+    if (!shipped.includes(specifier)) {
+      throw new HttpError(
+        400,
+        "plugin_not_shipped",
+        `'${specifier}' does not ship with this build; only builtin plugins can be installed.`,
+      );
+    }
+    const listed = await deps.projectConfig.getPlugins(projectId);
```
Myriad-Dreamin: **POST / PUT / DELETE read-modify-write the `[plugins]` table with no serialisation between requests.** Only `reassemble()` is queued on `op`; the `getPlugins` → `setPlugins` pair is not.

Two tabs (the page's `pendingSpecifier` gate is per tab): POST `@a/x` and DELETE `@a/y` arrive together. Both read `{y}`; POST writes `{y,x}`; DELETE writes `{}` via `ProjectConfigService.setPlugins` (`project-config-service.ts:561-564`, a bare `readRaw` → `writeRaw` with no lock, queue or version check; `readTable` is only an mtime cache). Both `apply()` calls then queue after both writes landed, the second re-assembly reads `{}`, and POST's client gets the "Installed @a/x" toast while `x` is gone.

`setCommandPolicy` has the same exposure, but these routes add the first admin-facing POST/DELETE pair on the file. Suggest running the read-modify-write inside the same `op` queue as the re-assembly.

packages/server/src/plugin/loader.ts
```diff
@@ -37,65 +56,280 @@ export interface PluginLoadResult {
...
+  const condition = (value: unknown): string | null => {
+    if (typeof value === "string") return value;
+    if (value === null || typeof value !== "object") return null;
+    const v = value as Record<string, unknown>;
+    return condition(v.import) ?? condition(v.default) ?? null;
+  };
+  let rel: string | null = null;
+  if (pkg.exports !== undefined) {
+    const exp = pkg.exports as Record<string, unknown>;
+    rel = condition("." in exp ? exp["."] : exp);
```
Myriad-Dreamin: **`"." in exp` throws a `TypeError` when `exports` is the valid string form** (`"exports": "./dist/index.js"`). Verified on Node 24: `'.' in './x'` → TypeError.

The throw is outside `packageEntry`'s `try`, so `resolvePlugin` rejects. `readPluginDeclaration` (line 299) calls it with no guard, awaited from `view()` (`plugins-installed.ts:91`) → `app.onError` → 500 for the Project's whole list and for every POST / DELETE / PUT (they all end with `view()`). The reuse check at line 457 is also outside the `try`, so a held entry whose package.json now uses the string form makes `loadPlugins` itself reject → `createInner` throws.

Guard with `typeof exp === "object" && exp !== null` before the `in`.

packages/server/src/plugin/loader.ts
```diff
@@ -37,65 +56,280 @@ export interface PluginLoadResult {
...
+
+/** One Project's list, in the order it wrote them; empty when it asks for none. */
+export async function readProjectPluginList(root: string, projectId: string): Promise<string[]> {
   let text: string;
   try {
-    text = await fs.readFile(path.join(root, PLUGINS_FILE), "utf8");
+    text = await fs.readFile(projectConfigPath(root, projectId), "utf8");
   } catch (err) {
     if ((err as NodeJS.ErrnoException).code === "ENOENT") return [];
     throw new Error(
```
Myriad-Dreamin: **One unreadable Project config now fails the whole server boot.** `readProjectPluginList` rethrows any non-ENOENT failure, and it now runs inside the platform's `create()`.

Data root with p1..p9; p7's `.project_config.toml` is chmod 000 or replaced by a directory. On base the server boots and p7's routes 500 on demand. On this branch `loadPluginHost` → `readPluginClosure` (111-119) throws `p7: .project_config.toml could not be read`; nothing catches it in `loadPlugins` (446), `loadPluginHost` (530) or `createInner` (`platform.ts:298-301`). `bootAppDeps` fails at process start, and on a push `hmr/host.ts` recovers the previous platform whose `create()` reads the same closure and throws again, so the deployment can neither serve nor be updated.

The comment two lines above says a broken Project must not take the deployment down. Suggest: warn, treat that Project as `[]`, and surface the reason in the list view.

Myriad-Dreamin (commented):
Review of `feat(plugins): list, install, and manage plugins`, 43 files on base `refactor/parked-platform-state`, at `ee23cf4c`.

Ten inline comments below cover the confirmed correctness findings. The two that need a decision before the rest are the re-assembly failure path (`platform.ts:493`, the process is left serving a disposed App) and the fact that a per-Project list edit runs `manager.shutdown()` for every Project (`platform.ts:388`).

Verified but not anchored inline:

- `scripts/build-plugins.mjs:62` passes unquoted args under `shell: true` on win32; a Windows profile path with a space breaks `pnpm pack` / `npm install`. `scripts/run-with-env.mjs` already has a `quote` helper.
- `packages/web/e2e/skills.spec.mjs` and `project-switch.spec.mjs` now fail silently against the folded, `inert` library list. CI does not run the web Playwright suite, so this will not show up red.
- `loader.ts:412` docstring and `docs/server-boot.en.md:87` still promise "malformed config fails the boot"; the code now warns and boots with `[]`.
- Stale `plugins.json` references remain in four plugin READMEs, docs, CONTRIBUTING, and in two comments this PR adds (`routes/plugins.ts:6`, `registry.ts:67`).
- Changelog `Breaking:` lacks the `yes — ` prefix and its Compatibility section ends in decision rationale rather than what changed.
- No spec under `design/specs` names the new routes or the `[plugins]` table. Design PR #124 still describes `plugins.json` and a deployment-wide `PUT /api/plugins/installed`, so design and implementation have diverged; per the repo rules the design should be reconciled first.

packages/server/src/plugin/loader.ts
```diff
@@ -163,13 +414,53 @@ function asPlugin(module: unknown): Plugin | null {
...
   for (const specifier of specifiers) {
+    // Reused only when the SAME FILE is behind the name. A push writes the builtin plugins to
+    // a new assets directory, so keeping an entry by specifier alone would run the previous
+    // build's plugin code forever — the push would land everywhere except the plugins.
+    const held = reuse.get(specifier);
+    const heldFile = held?.file;
+    if (
+      held !== undefined &&
+      heldFile != null &&
+      heldFile === resolvePlugin(specifier, bases)?.file
```
Myriad-Dreamin: **Reuse across a swap is keyed on the entry file path, and `import()` carries no cache-busting key.** A plugin updated in place at the same path keeps running the previous code after re-apply.

Operator installs a newer version into `<root>/plugins/node_modules/@acme/x`, then POST / PUT to re-apply. Here `heldFile === resolvePlugin(...)?.file` → the old entry is reused (454-461). Even without reuse, `importPlugin` (275) does `import(pathToFileURL(file).href)` and Node's ESM cache returns the old module (verified: rewriting the file and importing the same URL again returns the cached module).

The doc comment at 431-435 only covers the push case (new assets dir → new path). The `<root>/plugins` prefix is exactly the base a registry install would write to, so this will bite. Key the reuse on path + mtime (or package version) and append a query key to the import URL when it changed.

packages/server/src/plugin/loader.ts
```diff
@@ -37,65 +56,280 @@ export interface PluginLoadResult {
...
+  return null;
+}
+
+/**
+ * The file an `import "<name>"` of the package would load: `exports["."]` — a string, or its
+ * `import` / `default` condition — else `main`, else `index.js`. The packages are ESM and
+ * name one entry, which is all this reads; anything richer is the package's own business
+ * once Node imports it.
+ */
+function packageEntry(dir: string, manifest: string): string | null {
```
Myriad-Dreamin, line 229: **The hand-rolled resolver is narrower than `createRequire(entry).resolve(specifier)` in three ways that matter:**

1. A subpath specifier silently imports the package **root** entry. A Project lists `@acme/tools/sandbox` (`parsePluginTable` accepts it; only POST's `specifierOf` rejects slashes). Verified: `findPackageJSON('@acme/x/sandbox', base)` returns `@acme/x/package.json`, so `packageEntry` returns the root entry and the wrong module's default export joins the tree with no error.
2. A `node`-only export condition (`{".": {"node": "./dist/index.js"}}`) yields `null` and falls to `./index.js`.
3. An extensionless `main` (`"main": "dist/index"`) resolves to a path that does not exist; there is no `existsSync` in the package branch (237-240). Both surface as `ERR_MODULE_NOT_FOUND` skip reasons for packages that are on disk and worked on base.

Either go back to `createRequire` for the actual resolution (keeping `findPackageJSON` only to find the manifest), or reject subpaths in `parsePluginTable` and add `node` / `require` to the condition list plus an existence check.

packages/server/src/plugin/loader.ts
```diff
@@ -37,65 +56,280 @@ export interface PluginLoadResult {
...
+  if (root !== undefined && root !== "") {
+    bases.push({ file: path.join(pluginsPrefix(root), "package.json"), builtin: false });
+  }
+  if (assetsDir !== null) {
+    bases.push({ file: path.join(assetsDir, "plugins", "package.json"), builtin: true });
+  }
   const entry = process.argv[1];
   if (typeof entry === "string" && entry.length > 0) {
+    bases.push({
+      file: path.join(path.dirname(entry), "..", "plugins", "package.json"),
```
Myriad-Dreamin, line 176: **`findPackageJSON` walks `node_modules` upward from this base, so a package installed beside the program resolves through the `builtin: true` entry and is mislabelled.**

Plugin at `<install>/node_modules/@acme/x` (the case base 4, `builtin: false`, exists for), listed in a Project. Verified: `findPackageJSON('@acme/x', '<install>/plugins/package.json')` returns `<install>/node_modules/@acme/x/package.json` even when `<install>/plugins/node_modules` does not exist. This base is ordered before the argv[1] entry (160), so `resolvePluginPackage` returns `base.builtin === true` → `readPluginDeclaration` reports `builtin: true` → page shows the "built in" tag. But `discoverBuiltinPlugins` reads only the prefix manifest's `dependencies`, so `shipped[]` (255-263) lacks it → `POST {specifier:'@acme/x'}` → 400 `plugin_not_shipped` for the very row the page marks built in.

The same upward walk lets base 1 (`<root>/plugins`) pick up any `node_modules` above the data root. Decide `builtin` from the resolved path being under the prefix's own `node_modules`, not from which base found it.

packages/web/src/features/plugins/plugins-page.tsx
```diff
@@ -592,15 +705,325 @@ export function PluginsPage() {
...
+  for (const group of groups) {
+    const category = localizedText(locale, group.title, group.titleZh);
+    for (const plugin of group.plugins) rows.push({ kind: "library", plugin, category });
+  }
+  for (const listed of deployment?.plugins ?? []) {
+    rows.push({
+      kind: "module",
+      specifier: listed.specifier,
+      entry: index.find((e) => e.name === listed.specifier),
+      state: listed.active ? "active" : "pending",
```
Myriad-Dreamin: **The page never reads `InstalledPlugin.error`.** Every listed-but-inactive plugin is rendered as "restart to load".

POST a shipped plugin whose entry throws on import: the server answers `{active:false, error:'...', restartPending:false}` (`plugins-installed.ts:112-122`, pinned by `plugins-installed.test.ts:137-145`). This line maps it to `pending`; there is no `.error` access anywhere in the file. `ModuleRow` (~1372) shows `installedRestart` with the attention tone, the "Restart to load" facet counts it, the banner gate at line 559 stays off (`restartPending` is false), and `runDeploymentInstall` (286) toasts "Installed X". The operator restarts, it fails again, and the reason is never visible.

The `stateInactive` string was added to both dictionaries for this case and is referenced nowhere. Add a `failed` state carrying `error`, render it with the danger tone and the reason, and skip the success toast when the response says `error`.

hiyouga (approved):

PR #386, opened 2026-08-21
feat(web,server,desktop): update the desktop client from the account menu

hiyouga:
## What this adds

The account menu of the desktop shell's own window gains a **client-update row** — in exactly the slot where the server self-update row hides in desktop mode. It drives the shell's electron-updater end to end: check for updates, background download with live progress, then a one-click **Restart to install**, with the installed client version as the right-aligned chip. Unsupported forms (dev run, a Linux install that is not an AppImage) render the row disabled with the reason as a tooltip.

## Path chosen: server relay, not a preload bridge

`main.ts` documents two standing rules: the window is a **plain browser** (no preload, no node integration — an earlier preload-bridge proposal is recorded there as rejected), and every capability flows through the server's HTTP API. This PR keeps both:

- **Shell ⇄ server**: the embedded server already runs as an Electron `utilityProcess`; the shell pushes every updater status fold over the existing message port (`child.postMessage` / `process.parentPort`) and receives `check` / `install` command frames back. Under a plain `penguin server|web` run `process.parentPort` does not exist and nothing is wired.
- **Web ⇄ server**: new cookie-authed routes — `GET /api/desktop/update` (latest snapshot), `POST /api/desktop/update/check`, `POST /api/desktop/update/install` (202; 503 `shell_unreachable` when no shell is listening). Wire shapes live in `api/types.ts`, imported type-only by the shell.
- **Session gate**, both in the UI predicate (`offersClientUpdate`) and on the routes: `desktopMode && sessionVia === "desktop"` — the change-password rule inverted. A browser signed into the same desktop-mode server can neither read this machine's updater state nor restart its GUI app; a stale desktop cookie against a plain server finds no routes.

## Status machine

`packages/desktop/src/updater-status.ts` (pure, no Electron import, unit-tested) folds electron-updater events into a snapshot: `idle → checking → up-to-date | downloading(version, percent) → downloaded(version)`, plus `error(message)` and `unsupported(reason)`. One deliberate rule: **`downloaded` outranks every later event except another download or an install** — once a build is on disk, a periodic re-check (or its network failure) must not hide the actionable restart step. The web row maps snapshots through `clientUpdateRow`, polls while the menu is open (1.5s), and a row-initiated check toasts exactly one outcome (found / up to date / failed) via `clientCheckSettle`, the same manual-check contract as the server-update row.

The native menu flow (Check for Updates…, its dialogs, the restart prompt when a download lands) is unchanged; `quitAndInstall` still goes through the normal quit path so the before-quit hook stops the embedded server gracefully.

## Verification

- `pnpm build` and `pnpm typecheck`: clean across the workspace.
- `@prismshadow/penguin-desktop` tests: 66 passed (8 files), including new `updater-status.test.ts` (event walks, `downloaded` precedence, stale-progress guard, command-frame parsing).
- `@prismshadow/penguin-server` `desktop-update.test.ts` + `desktop.test.ts`: 19 passed — snapshot storage/serving, 403 `desktop_shell_only` for password-via sessions, 404 outside desktop mode, command forwarding, 503 without a wired port, status-frame validation, fake-port wiring.
- `@prismshadow/penguin-web` tests: 1001 passed (76 files), including new `desktop-update.test.ts` (predicate all four combinations, row models per state, settle rules).
- `pnpm format` + `pnpm format:check`, `git diff --check`: clean.
- Branch fast-forwarded onto current `origin/main` before the final run.

**Not verifiable in this environment** (no packaged builds, no GUI) — worth a pass on real machines:
1. macOS/Windows/Linux-AppImage packaged build: row appears in the shell window's account menu, check finds a staged release (`PENGUIN_UPDATE_FEED_URL` against a local generic feed works for a dry run), progress ticks, restart installs.
2. A browser signed into the desktop-mode server over loopback: no row, and `GET /api/desktop/update` answers 403.
3. `.deb` install and `pnpm start` dev run: row disabled with the reason tooltip.
4. Server crash/restart under the shell: the fresh child receives the current snapshot (relay re-wires on restart).

## Review fixes (38c6d9f9)

A 15-finding review pass; every finding was fixed, none skipped:

- **Stale install target**: automatic re-checks stand down while a build is pending (a re-download invalidates the package under the install button), and the status fold yields a `downloaded` headline to a *different* version actually replacing it.
- **Download vs. concurrent check**: while `downloading`, a stray `checking`/same-version re-announce no longer wipes the download context (progress kept flowing is pinned by test); a manual menu check during a download answers with a dialog instead of silently colliding.
- **Settle correctness**: the shell stamps `seq` on every fold; a row-initiated check settles on seq movement (snapshot equality cannot signal "checked, still up to date"), adds `ready` (already downloaded — previously mislabeled as "downloading in the background") and `unsupported` outcomes, and times out at 60s. A check the shell cannot run NACKs with a status re-push instead of silence.
- **Menu-close survival**: snapshot, polling and the armed watch moved to module level (`use-desktop-update.ts`, the use-version-info.ts convention) behind one always-mounted Sidebar hook — closing the menu no longer loses the promised toast. Polling shares one in-flight GET (no reorder clobber) and paces by state: fast only while checking/downloading/armed, slow for static snapshots with the menu open, none otherwise.
- **Install consent**: the row now opens a ConfirmModal carrying the native prompt's interruption warning before POSTing; an install failure toasts the real reason (`apiErrorText`) instead of "check failed".
- **First-paint honesty**: before the shell's first push the row renders disabled instead of a clickable stub.
- **Cleanups**: both update rows render through one shared `UpdateMenuRow` shell; desktop-mode test fixtures (`TEST_DESKTOP_TOKEN`, `createDesktopApp`, `desktopLoginCookie`) consolidated into server `test/helpers.ts` (a third drifted copy existed in desktop.test.ts); `action`/`busy` derive from `labelKind` via one table; the updater status initializes from `app.getVersion()` at module level, deleting the web-side placeholder scrub and its test; a dead ternary in the row label collapsed.

Re-verified after the fixes: `pnpm build` + full `pnpm typecheck` clean; desktop 68/68 (new fold-rule and NACK cases included), server desktop suites 19/19 (seq validation case added), web 1001/1001 (settle/table/disabled-row cases rewritten); `pnpm format:check` and `git diff --check` clean.


hiyouga:
## Review

The conflict with `main` is resolved and pushed; the PR is `MERGEABLE` again. I would not merge it yet, and the reason is not the code: `design/specs/06-PROTOTYPE.md:268` still says update interaction runs through native UI only and that "Web App 在桌面模式仍隐藏自更新入口（维持现状）". This PR is the direct negation of that sentence, and `CLAUDE.md` makes the design doc the tiebreaker — the spec has to move first, in its own repo, before this lands. Everything else I found is either fixed below or an open call for you; the mechanism itself holds up. I read the whole verification chain rather than assuming it, and the important thing to say plainly is that this PR **adds no new trust**: it routes to the same `electron-updater` call the native menu item already used, and `git diff origin/main...HEAD -- packages/desktop/electron-builder.yml .github/workflows/` is empty. The one behavioural risk I did find on the install path is a silent dead end, listed below rather than fixed, because undoing it means reversing a fold rule a previous review round deliberately installed.

## Fixed

**The conflict — `packages/web/src/components/layout/sidebar.tsx`** (`15d9f36c`, a merge of `origin/main` so nothing needed a force-push).

Three hunks conflicted, and the region had moved further than the label suggests. #369 did not just relocate the proxy and upload rows: it took the **entire server-update row** out of the menu and into `features/settings/updates-section.tsx`, leaving the menu with a System settings row, a `newVersion !== null` reminder row that opens the dialog on its Updates page, and logout.

- **Kept from `main`:** all of the above verbatim — the `settingsOpen` / `settingsSection` state, the `SettingsDialog` / `SettingsSectionKey` / `ICON_SIZE` imports, `const { update } = useVersionInfo(userOpen)` (the branch's `version` half has no consumer left), and the reminder row's own markup. The `ChangePasswordDialog` / `ProxySettingsDialog` / `UploadLimitsDialog` / `UpdateDialog` imports and JSX that this branch still carried are gone with their rows.
- **Kept from this branch:** `clientUpdateOffered` / `useDesktopUpdate` / `clientInstallOpen`, the `DesktopUpdateRow` mount, and the install `ConfirmModal`. The row sits at `sidebar.tsx:1720`, directly after the reminder row it stands in for — the same slot the PR always described, and still the right one: `lib/settings-sections.ts:56` drops the Updates page entirely in desktop mode, so the menu is the only surface left.
- `sessionVia` had to be restored to the `useAuth()` destructure at `sidebar.tsx:285`; #369 dropped it when `offersChangePassword` moved out.

Two things fell out of the resolution rather than being chosen:

- **`components/account/update-menu-row.tsx` is deleted**, its markup folded back into `desktop-update-row.tsx`. The file existed to keep two rows from drifting apart visually; with the server row gone from the menu, `rg` finds exactly one consumer, so its stated reason for existing is no longer true at HEAD.
- **The unsupported row's tooltip never opened.** `title` was written on the `<button disabled>` itself, and Chromium dispatches no pointer events to a disabled control, so it renders no tooltip (Firefox does). On a `.deb` install or a dev run the row was a greyed-out "Check for updates" explaining nothing — which is also the thing `06-PROTOTYPE.md:267` asks for by name ("禁用更新入口并说明升级方式"). The reason now hangs off a wrapper span, `desktop-update-row.tsx:82-90`.

**A download failure reported as a check failure** (`66360eb9`). `updater.ts:196` folds every `electron-updater` `error` into `state: "error"`, and `clientCheckSettle` flattened all of them to `{ kind: "failed" }` → `S.update.checkFailed`, "Update check failed — try again later". But a download rejected by its sha512, or on Windows by the Authenticode publisher check `NsisUpdater` pins from `app-update.yml`, arrives on that same event. The generic wording made a signature rejection read as a network hiccup, and `DesktopUpdateStatus.message` crossed the wire with no consumer at all. The settle now carries it (`lib/desktop-update.ts:155`) and the toast renders it through a new bilingual `clientUpdateFailed`; the generic string is left to the one case with nothing to report, the watch timing out. Covered by a new case in `web/test/desktop-update.test.ts`.

**No docs change for a user-facing feature** (`66360eb9`). `packages/docs/content/web-app.{en,zh}.md` §Version and Updates described only the server path, and `web-app.en.md:129` asserts the desktop app has no Updates page "because updating is the desktop shell's job" — true, and now incomplete. Added one mirrored paragraph to each: what the row does, that it is the shell window only, the restart confirmation, and the two disabled forms.

**Design-doc citations and CJK in code comments** (`606e41fe`). `updater.ts:2` and `update-support.ts:3` both carried `design § "桌面端原型 · 自动更新"` — a dead citation *and* the only two Chinese-carrying comments left in `packages/desktop/src`. Both sat on comment blocks this branch already edits. Removed; the surrounding text states the constraint on its own.

**The install route had no negative-authorization test** (`b6b13a98`). Only the `GET` was pinned at 403. `POST …/install` is the route that replaces the running application, so it now has password-session 403, unauthenticated 401 and plain-server 404, for both POSTs, asserting nothing reached the shell on any of the four refused attempts.

**Changelog** (`15d9f36c`, `66360eb9`): `Scope` gained `docs`; the "slot where the server self-update row hides" and "manual-check contract of the server-update row" lines were rewritten against the post-#369 reality, in both files.

## For you to decide

**1. The design doc forbids this feature.** `design/specs/06-PROTOTYPE.md:268`:

> 交互（与"不引入私有 IPC"约束的兼容）：壳不向页面注入任何通道，故更新交互一律走**原生 UI**⋯⋯Web App 在桌面模式仍隐藏自更新入口（维持现状，见「桌面登录」）。

The PR's own framing — no preload, no renderer IPC, everything through the server's HTTP API — satisfies the *premise* of that sentence but not its conclusion. `06-PROTOTYPE.md:34` describes the account menu's single update row, but that is the server-side check row, and it is itself stale after #369. Nothing in the spec describes a client-update affordance in the page. Per `CLAUDE.md` this is a design change first, code second. I did not touch `design/` (shared read-only symlink, other reviews are running against it). While you are in there, `:34` also still describes the retired proxy/upload/CLI-sessions menu rows and the old single update row, and `:268`'s companion `:269` records the proxy gap in item 3 below.

**2. A failed install dead-ends silently.** `updater-status.ts:47-50`: once `state === "downloaded"`, every event except `unsupported` or a *different* version is suppressed — including `error`. That rule exists to stop a re-check's network failure from hiding the restart step, and it does that well. But `quitAndInstall()` failing (`updater.ts:115`) surfaces as exactly that suppressed `error`. The user clicks Restart, the ConfirmModal closes, the POST answers 202, and if the shell cannot restart, nothing ever says so — the row keeps reading "Restart to install vX" forever, with no watch armed on the install path to time out. Worth noting that `check()` now stands down entirely while `downloadedVersion !== null` (`updater.ts:226`), so the re-check the suppression was written against can no longer produce an `error` in that state; letting `error` through would cost close to nothing today. Options as I see them: (a) let `error` land in `downloaded` and accept that a straggling in-flight failure could replace the headline; (b) arm a watch on install the way `requestClientCheck` does, so the row reports a shell that did not come back; (c) leave it and accept a silent no-op. I did not pick, because reversing the rule undoes a previous review round's decision.

**3. `PENGUIN_UPDATE_FEED_URL` accepts plain `http:` and any host.** `update-support.ts:39` validates the scheme only, and `updater.ts:151` feeds it to `setFeedURL({ provider: "generic" })`. Everything the updater then trusts — including the sha512 it verifies the artifact against — comes from that origin. On Windows the Authenticode publisher pin (`electron-builder.yml:87-93`, CI-asserted at `desktop-build.yml:341-344`) still gates it, and on macOS Squirrel's own signature check does; on Linux AppImage, which has no signature at all, sha512-from-the-feed is the entire chain. Pre-existing from #193, unchanged here, and nothing documents the variable — but this PR is what puts a one-click update button in front of it on AppImage, so it is worth a decision: drop `http:`, or require the override to be loopback, or leave it as the dry-run hatch it was built to be. This is release-plumbing territory, so I did not touch it.

**4. Nothing configures a proxy for `electron-updater`.** `06-PROTOTYPE.md:269` records that it uses Node http directly, bypasses the Electron session and reads no proxy env vars, so "其出网必须由壳按开关显式配置；注入方式待 spike". Still unspiked: `osProxyEnv()` is applied only to the server child (`server-process.ts:122`), never to the main process. Behind a corporate proxy every check fails — which used to be a background silence and is now a toast the user asked for. Not a regression, but the feature makes the gap visible.

**5. In desktop mode the avatar reminder dot has no row to open.** `sidebar.tsx:1666` shows the dot whenever `updateAvailable`, with no `!desktopMode` guard, while the row that explains it at `:1699` is `!desktopMode`-gated. `useVersionInfo` fetches regardless of mode. So a desktop user can get a dot and a "New version vX available" tooltip on the avatar with nothing in the menu about it — and now, next to a client-update row driven by a *different* source (electron-updater vs. the server's GitHub check) that can disagree with it. `main`'s, not yours, but this PR is what puts the two side by side.

**6. `checking` and `downloading` have no client-side escape.** The row renders `busy`/disabled straight off the snapshot, and the snapshot has no timestamp, so a shell stuck mid-check spins forever in the menu even though the web watch settles at 60s and toasts. `electron-updater`'s own request timeouts should always produce a terminal event, so I rate this bounded rather than broken — but giving the row a way out means putting a stamp on the wire, which is an API-shape change.

## Checked and clean

**The update path, read rather than assumed.** `electron-updater@6.8.9` / `electron-builder@26.15.3` / `electron@43.2.0`. What verifies an artifact before it replaces the app: sha512 from `latest*.yml` on every platform (`AppUpdater.js:562-566` → `DigestTransform`, and the same digest over a reassembled blockmap download); on Windows additionally `NsisUpdater.verifySignature`, which reads `publisherName` from the packaged `app-update.yml` and **skips verification entirely when it is absent** — it is present, pinned to two literal subject strings at `electron-builder.yml:87-93`, `verifyUpdateCodeSignature: true` at `:80`, and `desktop-build.yml:341-344` fails the build if it does not reach `app-update.yml`; on macOS, Squirrel.Mac's native identity match on a hardened, notarized, stapled build. Linux AppImage is sha512-only, unsigned. `latest*.yml` and `.blockmap` do ship as Release assets (`desktop-build.yml:487-502`, `release.yml:410-434`); `SHA256SUMS.desktop` is for the OSS mirror, not the updater. No GPG/minisign/cosign anywhere in the repo. Target is `nsis`, not `nsis-web`, so the web-installer branch is unreachable; `allowDowngrade` / `allowPrerelease` / `channel` are unset. **This PR changes none of that** — the diff against `origin/main` for `electron-builder.yml` and `.github/workflows/` is empty.

**Can the UI move a user to a build it did not offer?** No path I could find. `check()` stands down whenever `downloadedVersion !== null` (`updater.ts:226`), and so do both manual entries (`:98`, `:251`), so no replacement download can start under a pending install; `handleUpdaterCommand("install")` drops a stale frame when nothing is downloaded (`:108`); the row only offers install in `downloaded`. The "yields to a different version" branch of the fold is consequently unreachable in production — belt and braces, not a hole.

**Authorization, exercised.** Routes mount only when `deps.desktop` is set (`app.ts:514-516`) and *after* `app.use("/api/*", auth)` at `:510`, so Hono runs the auth middleware on them — unlike the Bearer-token `/api/desktop/shutdown` mounted at `:494`. `sessionVia === "desktop"` is only obtainable from the one-shot `desktop-login` redemption, which mints an admin session; a password login against the same desktop-mode server, a stale cookie against a plain server, and an unauthenticated caller all now have tests. CSRF holds via the existing model (SameSite=Lax + `jsonOnlyWrites`), so a cross-site POST never carries the cookie. In attach mode (`main.ts` boot, existing lock) the shell wires no relay and the window logs in with a password — no row, and I confirmed the predicate covers it.

**Server with no shell attached.** `shellPortOf` returns null under plain Node (`index.ts:223`), `getUpdateStatus()` stays null → the row renders `unknown`/disabled, and both POSTs answer 503 `shell_unreachable`. Ordering is safe: `index.ts`'s synchronous module body registers the port before `serve()`'s listen callback can fire, and the shell only posts after `waitForHttp` resolves.

**#381 compatibility.** The `test/helpers.ts` addition is a `const` plus two stateless factories — no module-level mutable state, no `vi.mock` anywhere in `desktop-update.test.ts`, and `test/isolation.test.ts` passes in the same run. Nothing to leak under `isolate: false`.

**Suites, on the merged tree:** desktop 68/68, web 1072/1072 (84 files), server 817/817 (79 files, run solo), docs 53/53. `pnpm -r build`, `pnpm -r typecheck`, `pnpm format` → `format:check`, `git diff --check`: all clean.

PR #389, opened 2026-08-21
feat(web): group the conversation list by time, and show the Agent in the session details

hiyouga:
Adds a third grouping mode to the chat sidebar's conversation list and surfaces the Session's Agent in the header's details card. Workspace grouping stays the default and "most recent" stays the default sort — both were already in place; this change extends the mode set.

## By time

A third radio in the list-options menu, next to Workspace and Agent: **past day / past month / earlier**, cut on each conversation's `lastActiveAt`.

`lastActiveAt` rather than `createdAt`, because that is the stamp the row's own compact timestamp and the recency sort already read — bucketing on creation would put a row labelled "3m ago" under "Earlier". Boundaries are rolling 24h / 30d windows; an unparseable stamp falls to the oldest bucket rather than jumping to the top; empty buckets are dropped.

The buckets span every Agent and every Workspace, which is what makes this mode different in shape from the other two:

- **The folders moved out of the group.** Subagent / Scheduled / Archived rows load only when their folder is first expanded, so an unloaded Session has no known bucket and no bucket could honestly advertise a share of them. They render once below the buckets as a Project-wide set, with counts and fetch fan-out summed over all Agents (`totalCategoryCounts`). Without this the archived folder would be unreachable in this mode at all.
- **Paging moved out too.** The server pages by `created_at DESC` per Agent, so a fetched page lands in whichever bucket its rows' activity puts them; a per-bucket "More" would be claiming something it cannot deliver. One row below the buckets fetches the next page for every Agent that has one, and a bucket's own "More" only reveals further already-loaded rows.

Collapse state, conversation pins and the manual drag order all work as in the other modes. The manual order was already stored per Project **and** grouping mode, so the new mode starts with its own sequence and cannot scramble the others.

The header's create button follows the grouping mode, as it already did for the other two; time buckets are not something to create into, so it starts a plain new conversation.

## The Trace page

Its tree rows are Trace files keyed by Session id and carry no activity timestamp, so there is nothing there to cut time buckets on. `GroupMode` therefore splits: the shared preference key keeps its three values, and the tree narrows to `TreeGroupMode` (`workspace` | `agent`), reading a stored `time` as the Workspace grouping it falls back to. Its two-icon toggle is unchanged.

## Agent in the details card

`Agent` sits above `Model` — a conversation belongs to an Agent first, and the Model it runs on is one of that Agent's settings. Paired the same way as the Model row (the id the API speaks, then the display name), with the name dropped when the Agent has none of its own and the two would simply repeat.

## Incidental

`MoreRow` hardcoded `aria-label` to the shared "More" string. The new whole-list paging row reads "Load more chats", and an accessible name contradicting the visible text announces a different row than the one on screen, so the prop became overridable — defaulting to the old value, so no existing row changed.

## Verification

Run in the worktree with Node v24:

- `pnpm -r build` — clean
- `pnpm --filter @prismshadow/penguin-web test` — 89 files, 1155 tests, all passing (34 in `session-grouping.test.ts`, 7 in `group-list.test.ts`, covering the bucket boundaries, activity-not-creation bucketing, key collision-freedom, and the summed counts)

Driven in a real browser against a server on a scratch `PENGUIN_HOME` (Playwright + the built Web App), with conversations backdated in the DB to populate all three buckets:

- the three buckets render in order with their counts, and each row's relative timestamp agrees with its bucket
- switching the mode from the list-options menu persists to `penguin.sidebarGroupMode`
- with 13 active conversations and one archived: the whole-list paging row appears, one click pulls the remaining rows in (no duplicate rows across buckets), and the shared Archived folder shows its exact server share and loads its row on first expand
- the details card shows `Agent / default_agent General Agent` directly above `Model`


hiyouga:
Reviewed the diff in full. The design holds up: the bucket math, the key spaces, the `GroupMode` / `TreeGroupMode` split and the i18n pairing are all sound, and I found no correctness bug in the grouping or paging logic. Four small things fixed (one behavioural, three doc/convention), pushed as `80cb784a`.

### Fixed

- `packages/web/src/components/layout/sidebar.tsx:1768` — the time buckets carry only active conversations, so a Project whose conversations are all archived (or all subagent / scheduled) rendered "No Sessions yet" directly above the shared "Archived (N)" folder. `renderGroupBody` already refuses exactly that contradiction per group (its `empty` is `parts.active.length === 0 && folders.every(f => f === null)`); the whole-list line now does too — the shared folders are hoisted into `timeFolders` (`sidebar.tsx:1134`, same call, same order, same keys) and the line renders only when every entry of it is null.
- `changelog/unreleased/2026-08-21-session-list-time-grouping.md:10` — the English entry named the buckets in Chinese ("近一天 / 近一月 / 更早 (past day / past month / earlier)"); every other English entry under `changelog/unreleased/` is CJK-free past its `中文版` link. Dropped to the English names only — the zh entry already carries the Chinese ones, so the pair stays section-for-section equivalent.
- `packages/web/src/lib/session-order.ts:68` — `sessionOrderKey`'s doc still said "the two modes cut the same Sessions into different partitions … a drag in workspace mode scramble the agent-mode order (and vice versa)", which this PR makes three. Reworded to "the modes" / "another mode's order"; the same touch-up the diff already made to `setGroupMode`'s comment.
- `packages/web/src/components/layout/sidebar.tsx:16` — the file header promised "Groups can be pinned via the header's hover pin toggle" for every mode, but the time buckets render no `GroupPinButton` and are not run through `pinnedFirst`. Named the exception, and rewrapped the one line the diff had left running past the block's width.

### Findings not fixed

- `packages/web/src/lib/session-order.ts:136` — the newest bucket can be silently incomplete. The server pages by `created_at DESC` per Agent (no index on `last_active_at`), so a conversation created long ago but active today sits on a later page. In Workspace / Agent mode that only costs ordering within a group; in time mode it means "Past day" omits rows until "Load more chats" pulls their page in, and nothing on screen says so. Real limitation, already documented in that file, and a server-side ordering + index change to fix — out of scope here, but worth an issue.
- `packages/web/src/components/layout/sidebar.tsx:536` — `timeParts` / `timeGroups` are recomputed on every Sidebar render, which is `partitionSessions` plus a sort each time, and the sidebar re-renders on every status tick. The comment argues the case (the boundary is `Date.now()`, and the row timestamps are recomputed for the same reason) and it is defensible; a `useMemo` keyed on a minute-resolution clock held in state would keep the boundary just as fresh for one pass a minute. Left alone as a stated design decision.
- `packages/web/src/components/layout/sidebar.tsx:543` vs `:1655` — `projectAgentsFor` fans out purely from `countsByAgent`, where workspace mode unions the counts' Agents with the Agents already contributing loaded rows, precisely so a missing or stale counts entry cannot hide a folder's content. I could not construct a case where it bites (the store populates counts for every Agent it lists), so this is a note, not a defect.
- `packages/web/src/components/layout/sidebar.tsx:1780` — "Load more chats" fetches but does not raise the buckets' display caps, so if the page lands in a bucket already at its 10-row cap the click adds a bucket-level "More" rather than visible rows: two clicks for one page. Only reachable once a single bucket is over the cap, and it follows from "a bucket's own More only reveals further already-loaded rows", so I left it.
- `packages/web/src/features/chat/chat-page.tsx:1608` — `sessionAgent.name !== undefined ? agentDisplayName(sessionAgent) : null` is equivalent to `sessionAgent.name`, since `agentDisplayName` is `a.name ?? a.agentId`. Harmless; an Agent whose `name` is `""` would render an empty trailing span.
- `packages/web/src/components/ui/group-list.tsx:183` — the new `ariaLabel` override is correct, and the call site passes a value equal to the visible text. Worth noting that while `pending`, every `MoreRow` (old and new) shows the loading label under an unchanged accessible name — a pre-existing "Label in Name" wrinkle this PR neither causes nor worsens.

### What I checked and found clean

- Bucket boundaries: rolling 24h / 30d with the older side losing, future stamps in the newest bucket, unparseable in the oldest. Consistent with `formatRelativeShort`, which switches from `Xh` to `Xd` at exactly 24h — so no row can read "23h" under "Past month". A future stamp does read as an absolute date under "Past day", which the test asserts deliberately.
- React keys: buckets are `\0time-<bucket>`, rows are keyed by `sessionId` inside a per-bucket `<ul>`, and `timeBucketOf` is total — one Session, one bucket, no cross-bucket collision.
- Duplicate rows from the whole-list fan-out: `loadMoreFor` filters each response against the loaded id set, and two Agents cannot return the same Session.
- `totalCategoryCounts`: the `n > 0` guard matches `aggregateWorkspaceCounts`, so a missing key cannot poison a total with NaN; the empty map returns explicit zeros.
- State keyed per grouping mode: `\0time-*` cannot collide with a Workspace path or an Agent id, `sessionOrderKey` gains a third namespace rather than sharing one, and `setGroupMode` swaps the stored order in on the switch.
- `GroupMode` / `TreeGroupMode`: `initialGroupMode` accepts only `agent` / `time` and falls back to `workspace`, so an old or junk stored value is safe, and `traces-page` — the only other reader — narrows through `treeGroupMode`.
- i18n: both dictionaries gained the same four keys in the same shape, `en: Strings` makes a one-sided addition a type error, and every new user-facing string in the JSX goes through `S`.
- No stale closures: `trackedLoadMore` updates `pendingLoads` functionally, and `timeMoreAgents` is recomputed each render rather than captured.

### Verification

Run in the worktree with Node v24.18.0, after the fixes:

- `pnpm --filter @prismshadow/penguin-web test` — 89 files, 1155 tests, all passing
- `pnpm format` — no files rewritten; `pnpm format:check` — "All matched files use Prettier code style!"
- `git diff --check` — clean

Not run: `pnpm -r build` (typecheck needed no dependency `dist`) and the Playwright e2e suite (known pre-existing local failures).

PR #390, opened 2026-08-21
feat(web): agent settings — import under the list, dashed empty blocks

hiyouga:
Two layout changes on the Agent settings tabs, one commit each.

## 1. Skills tab: the import button moves under the installed list

`Import skill` sat in the tab header, opposite the description fold. The Vault and Schedules
tabs put their add button in the slot between the list and the prompt section — under the
enable-injection card the three tabs share — so the Skills tab was the odd one out. Moved it
into that same slot.

Behaviour is untouched: the same `openImport` handler, the same modal (chat install + zip
upload), the same overwrite confirmation and error toasts, and the same `disabled` gate while
the list is loading. With the button gone, the header row collapses to the bare `HelpFold`,
which now carries the same comment as the Vault and Schedules tabs.

## 2. One dashed, centered block for the settings empty states

Four sections rendered their empty state as the same bare paragraph, copy-pasted classes and
all — `py-2 text-xs text-gray-400 dark:text-gray-500`:

- Skills — "No skills installed yet"
- Vault — "No environment variables configured yet"
- Schedules — "No scheduled tasks configured yet"
- Tools → MCP Servers — "No MCP Server configured yet"

A single left-aligned line where a bordered table or list would be reads as a panel that failed
to draw. Added `SettingsEmpty` next to `EmptyState` in `components/ui/empty-state.tsx` — a
rounded, dashed-bordered block, `min-h-24`, message centered on both axes — and pointed all four
at it, so they cannot drift apart.

The block reuses the neighbouring container's radius (`rounded-md`) and the gray level the
placeholder text already had; dashed rather than solid keeps an empty slot from reading as a
rendered-but-blank container. Classes are template-literal Tailwind, colours from the existing
gray scale — no new palette classes, no styling dependency.

The Memory tab's per-scope empty line was deliberately left alone: it sits inside a bordered
scope card, where a second frame would double up.

No string keys were added or removed, so both dictionaries are unchanged.

## Verification

- `pnpm -r build` — pass
- `pnpm --filter @prismshadow/penguin-web typecheck` — pass
- `pnpm --filter @prismshadow/penguin-web test` — 89 files, 1147 tests, all pass (includes
  `disclosure-anchor.test.ts`, which guards the "?" anchoring rule the touched headers sit under)
- `git diff --check` — clean
- No Playwright spec drives the Agent-settings Skills tab or any of these empty states
  (`e2e/skills.spec.mjs` covers the skill *library* page), so no e2e run was warranted.


hiyouga:
Import button moved to the top-right of the list per review; the rest of the diff reviewed and found sound, with four observations left as-is below.

### Import button moved

`packages/web/src/features/agents/skills-tab.tsx:298-306` — `Import skill` now sits in its own right-aligned row directly above the installed-skills list, between the enable-injection card and the list slot. It is the shape `admin-users-page.tsx:51` already uses for a right-aligned primary action above a table (`<div className="flex justify-end">` holding a `size="sm" variant="primary"` button), and the 1rem gap to the list below it is the same header→list gap `mcp-servers-section.tsx:301` has between its section title and its table — no new spacing idiom, no new string key.

**Empty list.** The row is unconditional, so it renders identically in all three list states — skeleton, dashed `SettingsEmpty`, populated list. The button therefore never changes position as the list loads or empties, and the dashed block always has a header above it rather than the button floating with nothing to belong to. The alternative — folding the action into the empty block (`EmptyState` has an `action` prop for that) — was rejected because it would make the button jump between two places depending on whether any skill is installed.

**Behaviour is unchanged**: same `openImport` handler, same modal (chat install + zip upload), same 409 `skill_exists` overwrite confirmation, same error toasts, same `disabled={skills === null}` gate while the list loads. No string keys added, removed or renamed.

The paired changelog entry (`changelog/unreleased/2026-08-21-agent-settings-empty-block.md` and `.zh.md`) was rewritten section-for-section in both languages to describe the placement that actually ships — H1, lead paragraph and the first Details bullet.

### Findings not fixed

- `packages/web/src/components/ui/empty-state.tsx:35` — `SettingsEmpty` has no vertical padding; it relies on `min-h-24` plus `items-center`. At the narrowest settings pane the four current messages wrap to two lines at most, far short of 6rem, so nothing touches the dashed edge today. A much longer future message would sit flush against it. A `py-4` would be free insurance but changes nothing that ships now, so I left it.
- `packages/web/src/components/ui/empty-state.tsx:35` — no `break-words`. None of the four strings contains an unbreakable token, so nothing overflows; a future message carrying a long path or env key could push out of the box horizontally. Same call as above.
- `packages/web/src/components/ui/help-fold.tsx:39` — the `className` prop now has no caller anywhere in `src`. The Skills-tab header wrapper (`min-w-0 flex-1`) was its only user and this PR removed it. That is dead API surface, but a generic `className` escape hatch is standard across the `components/ui` primitives here, so dropping it is a taste call rather than a defect.
- The Skills tab now diverges from its three siblings: Vault (`vault-tab.tsx:218`), Schedules and MCP Servers (`mcp-servers-section.tsx:375`) all keep their add button *below* the list, left-aligned. That divergence is the explicit review request for this tab — recording it so it does not get "consistency-fixed" back later. Whether the other three should follow the Skills tab is a separate decision.
- Small correction to the PR body: `test/disclosure-anchor.test.ts` does not cover the touched header. It parses every `.tsx` under `src` looking for an `InfoPopover` with no title among its preceding siblings; the Skills tab header uses `HelpFold`, which that test never inspects — before or after this change. `test/help-fold.test.ts` is the one that covers the fold. The test file itself is untouched and still guards exactly what its own docstring claims.

### Reviewed clean

- `SettingsEmpty` duplicates nothing. `EmptyState` in the same file is the page-level `py-12` title/description/action stack; `models-page.tsx:926` is a dashed *button*, `drop-zone.tsx:142` a dashed drag overlay. Its `border-gray-300` / `dark:border-gray-700` pair is the one `models-page.tsx:926` already uses, and `text-gray-400` / `dark:text-gray-500` is the level the replaced paragraphs had — no new palette classes, both themes explicit.
- No call site lost a loading state, and the dashed box cannot flash on open. Skills, Vault and Schedules all keep `X === null ? <SkeletonList/> : X.length === 0 ? <SettingsEmpty/> : …`. The MCP section has no null gate and never needed one — `agent-settings-page.tsx:267` mounts it only once `data` is loaded, passing `data.config.mcpServers`.
- Nothing was missed. The bare `text-xs text-gray-400` lines still in the Agent settings area are the schedules session-picker dropdown (`schedules-tab.tsx:183`) and the Memory per-scope line (`memory-tab.tsx:684`); neither is a settings list slot.
- i18n: the diff touches neither dictionary and introduces no literal user-facing English. Dictionary parity is enforced by the `Strings` type at typecheck, and `placeholders-parity.test.ts` passes.
- Header row after the move: `skills-tab.tsx:293` is now a bare `<HelpFold label={S.agent.tabSkills}>…</HelpFold>`, structurally identical to `schedules-tab.tsx:438` and `vault-tab.tsx:166` — no leftover flex container, gap or `min-w-0 flex-1`.

### Verification

- `pnpm --filter @prismshadow/penguin-web typecheck` — pass
- `pnpm --filter @prismshadow/penguin-web test` — pass, 89 files / 1147 tests
- `pnpm format` then `pnpm format:check` — clean ("All matched files use Prettier code style!")
- Not run: `pnpm -r build` (typecheck resolved against the existing dists, so it was not needed) and Playwright e2e (no spec drives the Agent-settings Skills tab, and the local e2e run has pre-existing failures unrelated to this branch).

PR #391, opened 2026-08-21
Move the update check out of System settings into the account menu

hiyouga:
Both update affordances now sit **outside** the System settings dialog, on the row directly under its entry in the sidebar user menu: the web app's server update check, and — in the desktop shell's own window — the client update. The dialog's **Updates** page is removed.

## The reading

The request was 「桌面端和网页端的检查更新都放到设置页的外面(下方)」 — put both the desktop client's and the web app's check-for-updates outside the settings page, below it.

Implemented as: **the account menu, on the row under the System settings entry**, not as a footer inside the dialog. Two things settle it. The desktop client row already lived in the account menu, below the settings entry — it was never inside the dialog, so "put both outside" can only mean bringing the web one out to join it. And the two rows are alternatives for the same slot (a desktop-mode server hides the server check; a browser against one gets neither), so they belong in one place, which the dialog's Updates page was not.

## What moved

- `features/settings/updates-section.tsx` deleted; `"updates"` removed from `SettingsSectionKey`, the section registry, the rail icons and labels, and `S.settings.updatesTitle` from both dictionaries.
- New `components/account/server-update-row.tsx`, the web counterpart of `DesktopUpdateRow` and shaped identically: one row that reads "check for updates" and runs the forced lookup until a release is known, then names it in place and opens `UpdateDialog` instead. The running server version sits muted on the right, as on the client row.
- The check state and `UpdateDialog` are held by the `Sidebar`, not the row: the row unmounts with the menu, and a check clicked there must still settle and still toast — the same reason `DesktopUpdateRow` keeps its snapshot at module level and hands its install off to the Sidebar's confirm dialog.
- `SettingsDialog` lost `initialSection`, which existed only so the update reminder could jump to the Updates page (`resolveSettingsSection(active, sections)` still gates what renders, on every render).

## Behaviour held fixed

- The check stays lazy: `useVersionInfo(userOpen)` — nothing is fetched until the menu is first opened.
- The avatar's reminder dot, the trigger's tooltip and accessible name, and the draft page's "New version available" badge are untouched.
- `offersClientUpdate` still gates the desktop rows, and `!desktopMode` still gates the server one; the install-confirmation step is unchanged.
- Each manual check still reports exactly one outcome as a toast.

## Side effect worth naming

A non-admin outside desktop mode now has only Personal pages, so the settings rail draws no group headings for them — that is `settingsGroups`' existing "a lone heading implies the other group" rule, and the test that pinned the old two-group case was rewritten to pin the new one.

Two strings changed value: `update.foundNew` said the update entry was "below" the row that was checked; in the menu that row *is* the entry, so the word is dropped in both languages.

## Docs

The Web App guide's System Settings and Version-and-Updates chapters were rewritten for the new placement, in both languages, including the release-date hint moving off the row (the new-chat page's version line still carries it).

## Verification

- `pnpm --filter @prismshadow/penguin-web test` — 89 files, 1147 tests, all pass
- `pnpm --filter @prismshadow/penguin-docs test` — 7 files, 53 tests, all pass (content changed)
- `pnpm --filter @prismshadow/penguin-web build` — clean
- No Playwright spec asserts on the settings dialog's sections or on these menu rows, so none needed updating.


hiyouga:
Reviewed the full diff. The removal is complete and nothing dangles; the placement change holds up under the failure modes it claims to handle. One test regression introduced by the diff, fixed and pushed as `e4a2715a`; everything else is either correct or arguable and left alone.

### Removal audit

Grepped repo-wide (`.ts`/`.tsx`/`.mjs`/`.md`/`.json`/`.yaml`, excluding `node_modules` and `dist`) for `"updates"` / `'updates'` as a section key, `updates-section`, `UpdatesSection`, `updatesTitle`, `initialSection`, `settingsSection` / `setSettingsSection`, and `SettingsSectionKey` imports. All clean:

- **Deep links into the dialog's Updates page.** `main`'s only one was the sidebar's reminder row (`setSettingsSection("updates"); setSettingsOpen(true)`) reaching `SettingsDialog`'s `initialSection`. Both are gone, and `git grep initialSection main -- packages/` confirms the sidebar and the dialog were the only two production sites — no other caller ever passed it. `sidebar.tsx` no longer imports `SettingsSectionKey` at all.
- **Dictionary parity.** `S.settings.updatesTitle` is gone from both `strings.ts` and `strings-en.ts`; no key exists in one and not the other (`tsc` on `const en: Strings` plus `placeholders-parity.test.ts` cover the rest, both green). No orphans: the two strings the deleted page used — `S.update.lastUpdated` and `S.update.newVersionBadge` — are still live on the draft page's version line (`packages/web/src/features/chat/draft-view.tsx:884`, `:898`, `:901`), so dropping the release-date hint off the row does not strand either.
- **Docs.** No page in either language still puts the check inside System settings; `grep -ri 'Updates page|更新页|Server → Updates|系统设置.*更新'` over `packages/docs` and `packages/landing` returns nothing outside the frozen v0.1.4 / v0.2.0 blog posts, which describe shipped releases and are correct as history. Both new in-page anchors resolve: `#version-and-updates` and `#版本与更新` match `slugifyHeading` (`packages/docs/src/lib/toc.ts:15`), which keeps CJK verbatim.
- **e2e and tests.** `packages/web/e2e/` has no reference to the settings rail or to these menu rows — the PR's claim checks out. No dead imports, dead props or dead helpers left behind: `PrefRow`, `formatMonthDay` and `UpdateDialog` all keep other callers.

### Fixed

- `packages/web/test/account-menu.test.ts:72` — the diff added a second `{!desktopMode && (` to `sidebar.tsx` (the `ServerUpdateRow` gate at `:1729`, above sign-out's at `:1762`), so `it("keeps sign-out on its own desktopMode gate")`'s `expect(source).toContain("{!desktopMode && (")` silently stopped pinning sign-out — it would now pass with sign-out's own gate deleted. Re-anchored to the sign-out block: the nearest gate preceding `S.auth.logout`, with nothing but sign-out's own `<button>` between the two. Checked against a mutated copy of `sidebar.tsx` with that gate removed — the new assertion sees `["<S","<D","<b"]` and fails, where the old one passed.

### Findings not fixed

- **Mobile drawer drops the in-flight spinner.** `Drawer` returns `null` when closed (`components/ui/drawer.tsx:28`), so the drawer's `Sidebar` unmounts with it and takes `updateChecking` along. The check still settles and still toasts exactly once — the toast store is module-level (`components/ui/toast.tsx:33`) — but reopening the drawer shows "Check for updates" again, so a second concurrent check and a second toast are one tap away. Not a regression: on `main` the check lived in `UpdatesSection`, whose React owner was also the `Sidebar`. The pinned desktop sidebar, which is the case the new comment at `sidebar.tsx:337` describes, is CSS-hidden rather than unmounted and is genuinely unaffected.
- **`UpdateDialog` is mounted unconditionally** (`sidebar.tsx:1783`), including in desktop mode where nothing can open it (`updateDialogOpen` is only ever set by `ServerUpdateRow`, itself `!desktopMode`). Harmless — `Modal` renders nothing while closed — and it matches the already-unconditional client-install `ConfirmModal` beside it, so adding an asymmetric gate would cost more than it buys.
- **`foundNew`'s deixis** (`lib/strings.ts:141`, `lib/strings-en.ts:142`). Both readings are accurate — the check does not close the menu, so the row flips in place under the cursor — but the pointer is now redundant, and the Chinese 「该更新入口」 has no antecedent in a standalone toast. If you want it tighter, dropping the clause ("New version vX found" / 「发现新版本 vX」) says everything the row itself now shows. Arguable copy, so left as written.
- **Two unreleased entries still describe the deleted page.** `changelog/unreleased/2026-08-20-system-settings-page.md:18,28,30` and `2026-08-20-client-update-button.md:10,18` name "Server → Updates" as a page, and they ship in the same folder as this entry — so the release folder will document a page that never existed in any released build. Cross-linking, which this entry does, is the documented mechanism and rewriting a merged PR's entry is out of scope for this review; flagging it so it can be decided before `unreleased/` is renamed at release.

Checked and found correct, no action needed:

- **The two rows cannot collide.** `!desktopMode` and `offersClientUpdate = desktopMode && sessionVia === "desktop"` are mutually exclusive, and even co-rendered they share no state — `updateChecking` is server-only, `clientUpdate.checkPending` client-only, and their dialogs are separate.
- **No double toast, no setState-after-unmount, no dropped result** on the pinned sidebar: the re-entry guard plus `disabled={checking}` cover the double click (React flushes the discrete click before the next one, so the second handler sees `checking === true`), and the `[]`-deps listener effect in `use-version-info.ts:72` means a result landing after the menu closed still reaches the sidebar's state, not just the module cache.
- **The lazy gate does not refetch, cache-wedge, or strand a failure.** Closing and reopening the menu reuses the module-level `versionPromise` / `updatePromise`; a rejected lazy fetch nulls its promise so the next open retries; `forceUpdateCheck` nulls its own promise on rejection and rethrows, so `runUpdateCheck`'s `finally` clears `updateChecking` and the row is immediately clickable again.
- **`resolveSettingsSection(active, sections)` with `initialSection` gone.** `active` resets to `null` on open and resolves to `sections[0]`, which is always `general` (`visible: () => true`) — the same landing page `main`'s `initialSection = "general"` produced. A late-arriving admin flag widens the rail without moving the page, and a viewer who *loses* admin mid-dialog now falls back instead of continuing to render the admin form.
- **The `settingsGroups` re-pin is pinning the right thing.** `settings-sections.test.ts:96` now asserts the real non-admin case (`["personal"]`) rather than a synthetic filtered list, and the two-group case stays pinned by `admin` at `:92`. `PagedDialog` gates headings on `groups.length > 1` (`components/ui/paged-dialog.tsx:59`), so a non-admin gets General / Appearance / Account with no headings — which is what both docs edits now say.

### Verification

Run in the worktree on `feat/update-check-placement` at `e4a2715a`, Node v24.18.0:

- `pnpm --filter @prismshadow/penguin-web test` — 89 files, 1147 tests, all pass (run before and after the test fix)
- `pnpm --filter @prismshadow/penguin-docs test` — 7 files, 53 tests, all pass
- `pnpm --filter @prismshadow/penguin-web exec vitest run test/account-menu.test.ts test/settings-sections.test.ts` — 22 tests, all pass
- `pnpm format` then `pnpm format:check` — "All matched files use Prettier code style!"; `git diff --check` clean

Not run: `pnpm -r build` (typecheck resolved every dependency, so no stale `dist` to rebuild) and Playwright e2e (known pre-existing local failures, and no spec touches these surfaces).

PR #392, opened 2026-08-21
refactor(web): settings explanatory text follows the disclosure convention

hiyouga:
Two complaints about the same thing: the settings surfaces explain themselves on screen, so
every visit pays for prose that is read once.

The Web App already has a rule for this (`components/ui/info-popover.tsx`, enforced by
`test/disclosure-anchor.test.ts`): **semantics** — what a page or a row is, what it affects,
when it takes effect — go behind the circled "?" beside the title; **formatting** — the shape
a value must take — stays on screen in the field's hint. The settings pages predated it.

Two commits, one per complaint.

## 1. `fix(web)`: cut the security-policy wall of text

The Project settings security page opened with five sentences: both shell entry points
(`exec_command` and `input_command`), the literal `"Tool call denied by policy."` string the
model receives, the normalize-then-regex matching algorithm, and the run-time-assembly evasion
caveat. It sat above the rule list on every visit.

Two of those facts change what a reader does — a hit is refused whatever the approval mode
allows, and this is a guardrail rather than a sandbox — so `commandPolicyIntro` became
`commandPolicyInfo`, two sentences behind a "?".

The "?" needs a title to anchor to, and the section renders rows without repeating the tab's
name, so the anchor is the dialog's own pane heading: the tabs array grew an optional `info`
and the security tab is the first to carry one. The rule editor's `"Regular expression"`
placeholder is the format rule for the pattern field and stays visible.

## 2. `refactor(web)`: disclose the System settings prose

- **Preference rows.** `PrefRow` grew an `info` slot rendered beside its label — anchored,
  since the label is a title. Language, currency, CLI sessions, theme, terminal theme, font
  size, accent and change password moved their sub-labels into it. `hint` stays for what must
  remain visible; the updates page's build date is the one caller left, and it is a live fact
  rather than an explanation.
- **Page-level prose.** `PagedDialog` grew an `info` on `PagedDialogItem`, drawn beside the
  pane heading — the only title a page has, which is why this is a "?" and not a `HelpFold`.
  Proxy options' paragraph moved there, and `SectionShell`'s `description` slot lost its last
  caller and went with it.
- **Upload limits** was one string mixing both kinds, which the rule says to split rather than
  judge. The accepted MB range is read while typing and became each field's `hint`
  (`attachmentMaxMbHint` / `attachmentTotalMbHint`); the fixed per-message attachment count and
  the separate inline-image cap became `uploadLimitsInfo` behind the "?".

The renamed keys (`*Hint` → `*Info`) each had exactly one call site, all inside this diff. Both
dictionaries move together, as the `Strings` type requires.

## Not touched

- The Agent settings tabs already went through this pass — `HelpFold` on the tabs with no
  title, `info` on the runtime fields. The one paragraph left visible there,
  `mcpBudgetsHint`, is "Leave empty for defaults", which is formatting.
- Already-disclosed text: `title=` tooltips, `OptionMenu` descriptions, confirm bodies, toasts,
  empty states.
- Destructive-consequence lines next to a destructive control (`deleteProjectDesc`) and lines
  explaining why a control is absent (`deleteDefaultForbidden`, `deleteLastForbidden`) stay
  visible — hiding a warning behind a click is not the convention this PR is applying.
- The updates page, which is being moved in a concurrent PR.

## Verification

Node v24.18.0.

| Command | Result |
| --- | --- |
| `pnpm -r build` | pass (needed before typecheck — fresh worktree) |
| `pnpm --filter @prismshadow/penguin-web typecheck` | pass |
| `pnpm --filter @prismshadow/penguin-web test` | pass — 89 files, 1147 tests |
| `pnpm format` + `pnpm format:check` | pass — "All matched files use Prettier code style!" |

`test/disclosure-anchor.test.ts` (every `InfoPopover` has a title among its preceding
siblings) and `test/info-popover.test.ts` (the `<label>`/`<button>` retargeting trap) are the
ones guarding this area; both are in the run above. Nothing outside `packages/web` changed.


hiyouga:
Reviewed the whole diff and fixed the reported add-rule bug on this branch. The disclosure refactor itself holds up: no formatting rule was hidden, every new "?" is anchored to a real title, both dictionaries stay in step, and no renamed key is left dangling. Two commits pushed — `46de3bd` (the fix) and `a93e675` (a guard).

### Add-rule placement

**Root cause.** In `SecurityPolicySection` the "Add rule" button lives in the list's *header* row — `Rules · N` on the left, Restore / Add on the right — which sits **above** the rules. Clicking it set `editing = "new"`, and the `RuleEditor` for a new rule was rendered **after** `rules.map(…)`, i.e. below every existing rule. The nine seeded factory rules each render three lines (name, description, pattern) at `py-2.5`, so the list alone is ~680px tall while the Project settings dialog body is `max-h-[70vh] overflow-y-auto` (`packages/web/src/components/ui/modal.tsx:105`). The form therefore opened a few hundred pixels below the fold of that scroller, nothing moved, and the click read as inert. Nothing focused either, so a keyboard user got no cue at all.

**Fix** (`packages/web/src/components/layout/project-dialogs.tsx:1052`). The add form now renders as the list's **first** child, directly under the header row that holds the button — no scroll, no motion, and the form appears where the pointer already is. The applied rule joins at the **head** of the list (`setRules([nr, ...rules])`, line 1056) so it stays where it was typed instead of vanishing to the bottom the moment you press Apply. That is safe because deny rules are order-independent: `evaluateCommandPolicy` refuses on *any* enabled match, and list order only decides which rule *name* the refusal reports.

I preferred this over "append + `scrollIntoView`" because scrolling would fix visibility while leaving the two ends of the interaction ~700px apart — and would leave the applied rule off-screen again. The reasoning is inlined as a comment at the call site.

**Focus** lands in the editor's first field, the rule's name (`project-dialogs.tsx:890`, `autoFocus` — the same idiom `CreateProjectDialog` already uses at lines 142 and 163). The editor mounts only in response to an explicit Add or Edit click, so taking focus is what the click asked for; it also gives the browser's scroll-on-focus as a backstop on a short viewport where the header row is the last visible thing.

**Test.** `packages/web/test/command-policy-add-rule.test.ts` pins both halves — the JSX child order (parsed with the TypeScript parser, the way `disclosure-anchor.test.ts` checks anchoring, because sibling order is not something a regex can see) and the focused first field (rendered; `RuleEditor` is now exported for it). I verified the first assertion is not vacuous by moving the block back below the map: it fails with `expected 5 to be less than 3`.

Changelog: the paired entry's **Project settings** section gained a matching bullet in both files.

### Fixed

- `packages/web/src/components/ui/paged-dialog.tsx:109` and `packages/web/src/features/settings/setting-row.tsx:35` — the two new "?" anchors used `ICON_GAP.tight` (`gap-1`). The app already has a consistent split: `gap-1` is the **field-label** spacing (`components/ui/field.tsx:132`, and `project-dialogs.tsx:749`), while a "?" beside a **heading or a row title** uses `gap-1.5` in all nine existing places — including this PR's own third new anchor at `project-dialogs.tsx:293`, and `agent-settings-page.tsx:433`, which is structurally identical to `PrefRow`'s title. Changed both to `ICON_GAP.row`, so the mark reads the same everywhere and the PR is internally consistent.

### Findings not fixed

- **Dropped facts from `commandPolicyIntro`.** The two that went are which tools the policy sees (`exec_command`'s `cmd` and `input_command`'s `chars`) and the literal denial text the model receives. Neither is needed to write a regex, and both are in `docs/configuration` § *Command policy*, so nothing is lost from the product. The kept sentences do cover what a rule author needs — normalization, that a hit outranks the approval mode, and the run-time-assembly gap.
- **"Accident guardrail, not a defence against evasion" is now behind a click.** It is arguably a warning rather than a semantic, and misreading it means over-trusting the policy. It is not a destructive-consequence line next to a destructive control, which is where the convention says text stays visible, so I left it — flagging it because it is the one sentence on this page whose disclosure is a judgement call rather than a rule.
- **The rule editor's pattern field has no `info` of its own.** "Normalized for whitespace and quoting, then matched against the regular expression" is read *while writing a pattern*, but it now lives only at the pane heading. The editor's boxes are bare `Input`s with placeholders and no labels, so there is no title for a "?" to anchor to — fixing it properly means labelling those fields, which is more than this PR should carry.
- **`Type: refactor` on the changelog entry.** It now also describes a defect correction, and `changelog/README.md` allows exactly one `Type`. Happy to split the placement fix into its own `Type: fix` entry if you would rather `grep -rl 'Type:\*\* fix'` find it.
- **`PagedDialogItem.info` is `ReactNode` while the Project settings tabs array types `info?: string`** (`project-dialogs.tsx:256`). Harmless — the array is local and `InfoPopover` takes both — just not uniform.
- **`PrefRow` labels are still not programmatically associated with their controls.** `Switch` and `Segmented` carry no accessible name of their own, so those rows announce as unnamed controls. Pre-existing (the title was a `<p>` before this PR too) and not fixable inside `PrefRow`, whose children are opaque; worth a follow-up.

### Verification

Node v24.18.0, scoped to `packages/web` — `pnpm -r build` was not needed (typecheck resolved everything), and Playwright was not run.

| Command | Result |
| --- | --- |
| `pnpm --filter @prismshadow/penguin-web typecheck` | pass |
| `pnpm --filter @prismshadow/penguin-web test` | pass — 90 files, 1151 tests (was 89 / 1147: +1 file, +4 tests) |
| `pnpm format` then `pnpm format:check` | pass — "All matched files use Prettier code style!" |
| `test/command-policy-add-rule.test.ts` with the form moved back below the map | fails as intended — `expected 5 to be less than 3` |
| `test/disclosure-anchor.test.ts` with a deliberately orphaned `InfoPopover` dropped into `src/components/ui/` and `src/features/settings/` | fails and names both files, so the new anchor sites really are inside its scan |

Also checked by hand, all clean: the renamed keys (`*Hint` → `*Info`, and `uploadLimitsHint` → the split trio) have zero references left anywhere in the repo including `e2e/` and `docs/`, and each new key has exactly one call site; `SectionShell` has no caller left passing `description`; no e2e spec selects on a settings-page string; `PagedDialog` has a single caller and `HelpFold` appears only on the Agent settings tabs, so no page carries both disclosures.

I also extended `test/info-popover.test.ts` to cover the new `PrefRow` slot (commit `a93e675`): the rendered row contains no `<label>` at all, so the "?" cannot retarget the row from its control to the trigger — the trap the existing `Field` cases guard, on the one row primitive with no `Field` in between.

PR #393, opened 2026-08-21
fix(desktop): ship node-pty with the app so terminals can spawn

hiyouga:
Opening a terminal in the desktop app fails:

```
POST /api/terminals → 500 {"error":{"code":"terminal_spawn_failed","message":"Could not start a shell: node-pty is unavailable to the platform: Cannot find module 'node-pty'\nRequire stack:\n- .../packages/desktop/dist/server.js / no assets directory published"}}
```

Reported on macOS; it reproduces identically on Linux, so it is not platform-specific.

## Root cause

`node-pty` is a native module, and it is the one dependency the desktop bundler cannot absorb. The server loads it through a runtime `require` — `createRequire(import.meta.url)("node-pty")` in `packages/server/src/platform/terminal/pty-module.ts` — which esbuild leaves untouched, and node-pty's own loader then resolves `../build/Release/pty.node` or `../prebuilds/<platform>-<arch>/pty.node` relative to its `lib/` directory. Both need a real package directory on disk.

The desktop app has had none since https://github.com/Prism-Shadow/penguin-harness/pull/335 replaced `scripts/stage.mjs` (`pnpm deploy --prod`, which materialized the whole production dependency tree) with full bundling: `pnpm build` emits `dist/server.js` and `electron-builder.yml` ships no `node_modules` at all. Under pnpm, `node-pty` is installed into `packages/server/node_modules`, which a lookup anchored at `packages/desktop/dist/server.js` never reaches — so the require fails in a source run, and the packaged app has no copy at all.

The terminal feature (https://github.com/Prism-Shadow/penguin-harness/pull/285) merged 26 minutes after that packaging change on the same day, adding the first native dependency to a packaging path that had just stopped carrying any. Its `terminal-smoke.mjs` guard resolved node-pty from `packages/server/package.json`, so it verified the ABI and the spawn but never the lookup the shipped bundle performs, and passed throughout.

The second half of the message, `no assets directory published`, is not a separate defect: it is `loadNodePty`'s second attempt — the HMR pushed-asset fallback — correctly reporting that no push has published an assets directory. It only becomes visible because the first attempt failed.

## The fix

`pnpm build` stages a node-pty package directory into `packages/desktop/dist/node_modules/node-pty`, the first `node_modules` the resolver meets walking up from `dist/server.js`, which serves a source run and a packaged app with the same layout. No server change: the seam where `desktop` runs the unchanged server holds.

- `src/pty-payload.ts` (new, unit-tested) selects and copies the shipping subset: the manifest, the license, `lib/`, the `build/Release` binding compiled at install time, and `prebuilds/`. Dropped: sourcemaps, node-pty's own test build, its TypeScript sources, node-gyp inputs, the vendored winpty tree, and Windows `.pdb` / `.lib` files. That is 5.4 MB staged out of a 60 MB install, covering darwin arm64/x64 and win32 x64/arm64 alongside the host build. `scripts/deploy.mjs` already applies the same selection rule to the HMR push payload; the two are left separate — different destination, different transport — with a cross-reference in the new module.
- The staging step restores the exec bit on every `spawn-helper` it copies. node-pty publishes that darwin side binary as `0644`, which makes `posix_spawnp` refuse it; `packages/server/src/terminal/spawn-helper.ts` documents a packaging step that fixes the mode in the staged tree, which stopped existing when `stage.mjs` was removed. This restores it, and the runtime repair stays as the fallback for an npm install.
- `electron-builder.yml` lists `dist/node_modules/**/*`. This is load-bearing: electron-builder appends `!**/node_modules/**` to the app matcher unless a non-negated pattern names `node_modules`, and when one does it splices that exclusion *before* it — last match wins, so the app's only `node_modules` ships while nothing else can sneak one in.
- The `pnpm desktop` preflight checks the staged package, and `build-assets.mjs` fails the build when the resolved install carries no `pty.node`, so a broken payload is a build error rather than a 500 the user meets when they open a terminal panel.
- `scripts/terminal-smoke.mjs` now loads node-pty exactly as the server bundle does — a bare `require("node-pty")` anchored at `dist/server.js` — under Electron's own Node. CI's `ci-macos` job already runs it after `pnpm build`, so it now covers module resolution as well as the ABI and the spawn, and this class of regression fails there.

## Verification

Everything below ran on Linux (Node v24.18.0, Electron 43.2.0):

- **The reported 500, reproduced and fixed.** Booting `packages/desktop/dist/server.js` exactly as the shell forks it (`PENGUIN_DESKTOP_TOKEN`, `PENGUIN_PORT_FILE`, `PORT=0`), with the staged copy hidden, returns the reporter's message verbatim — `Cannot find module 'node-pty'` … `/ no assets directory published`, HTTP 500. With it in place, `POST /api/terminals` returns **HTTP 201** and a live pty.
- **The real Electron app.** `pnpm --dir packages/desktop start` under `xvfb-run` with a scratch `PENGUIN_HOME`: `POST /api/terminals` → 201, then `POST /:id/keys` with `echo …` + `Enter` and `GET /:id/capture` read the command's output back from the pty running inside the utilityProcess.
- **Electron ABI.** `node packages/desktop/scripts/terminal-smoke.mjs` → `node-pty spawn OK (Electron Node 24.18.0, ABI 148)`, now through the bundle's own resolution.
- **Packaging.** `electron-builder --linux --dir` ships all 34 staged files to `resources/app/dist/node_modules/node-pty`, with `spawn-helper` at `0755`.
- `pnpm -r build`, `pnpm typecheck`, `pnpm --filter @prismshadow/penguin-desktop test` (98 tests, 10 files), `pnpm format` + `pnpm format:check`, `git diff --check`.

**Still needs a macOS run.** Two things this box cannot exercise: the `prebuilds/darwin-*` binaries actually loading under Electron on darwin (Linux uses the node-gyp `build/Release` binding, and the darwin prebuilds are copied but never dlopened here), and `spawn-helper` being exec'd by `posix_spawnp` from the staged copy. Both are covered by the `ci-macos` job's `terminal-smoke.mjs` step, which now goes through the shipped path — a macOS failure there is a real failure, not a flake. The Windows payload is likewise copied and pattern-verified but not executed here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,11 +62,13 @@ jobs:
 
   # macOS: the same build/typecheck/test gates on macos-latest (arm64). This is the only job
   # where the terminal's pty suites run on darwin — node-pty compiles against the runner's
-  # toolchain and its darwin-only spawn-helper actually executes — and the standalone
-  # terminal smoke script then re-runs node-pty under Electron's own Node
-  # (ELECTRON_RUN_AS_NODE), the runtime the packaged app's server really uses. A macOS
-  # terminal regression that once shipped as a silently empty panel fails here instead.
-  # No format:check (same files as ubuntu) and no live-LLM e2e (ubuntu-only budget).
+  # toolchain and its darwin-only spawn-helper actually executes — and the terminal smoke
+  # script then re-runs node-pty under Electron's own Node (ELECTRON_RUN_AS_NODE), resolved
+  # from the built desktop bundle: the same module lookup, on the same runtime, that the
+  # packaged app's server performs. It runs after the build because that staged copy is a
+  # build output. A macOS terminal regression that once shipped as a silently empty panel
+  # fails here instead. No format:check (same files as ubuntu) and no live-LLM e2e
+  # (ubuntu-only budget).
   ci-macos:
     runs-on: macos-latest
     steps:
--- a/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.md
+++ b/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.md
@@ -1,0 +1,34 @@
+# Desktop terminals: node-pty ships with the app again
+
+- **Date:** 2026-08-21
+- **Type:** fix
+- **Scope:** `desktop`, `ci`
+
+[中文版](2026-08-21-desktop-terminal-node-pty.zh.md)
+
+Opening a terminal in the desktop app failed with `POST /api/terminals` → 500
+`terminal_spawn_failed`, reporting `Cannot find module 'node-pty'` from
+`packages/desktop/dist/server.js`. node-pty is a native module and the only dependency the
+bundler cannot absorb: the server reaches it through a runtime `require`, and node-pty's own
+loader then resolves its binary relative to its package directory. `pnpm build` now stages a
+node-pty package directory into `packages/desktop/dist/node_modules`, which both a source run
+and a packaged app resolve, and `electron-builder.yml` ships it.
+
+## Details
+
+- `src/pty-payload.ts` selects and copies node-pty's shipping subset — its manifest, license,
+  `lib/`, the `build/Release` binding compiled at install time, and the `prebuilds/` binaries —
+  dropping sourcemaps, node-pty's own tests, its TypeScript sources, node-gyp inputs, the
+  vendored winpty tree and 44 MB of Windows `.pdb` debug symbols. The staged copy is 5.4 MB and
+  covers darwin arm64/x64 and win32 x64/arm64 alongside the host build.
+- The staging step restores the exec bit on every `spawn-helper` it copies. node-pty publishes
+  that darwin side binary as `0644`, which makes `posix_spawnp` refuse it; fixing the mode
+  before the app is packed keeps a signed `.app` under `/Applications`, where the server's
+  runtime repair cannot write, out of that failure.
+- `scripts/build-assets.mjs` resolves node-pty from `packages/server`, where pnpm installs it,
+  and fails the build when the install carries no `pty.node`.
+- The `pnpm desktop` preflight checks the staged package, so a stale build says so at launch
+  instead of at the first terminal.
+- `scripts/terminal-smoke.mjs` now loads node-pty the way the server bundle does — a bare
+  `require("node-pty")` anchored at `dist/server.js` — under Electron's own Node, so CI's macOS
+  job covers module resolution as well as the ABI and the spawn.
--- a/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.md
+++ b/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-21
 - **Type:** fix
 - **Scope:** `desktop`, `ci`
+- **PR:** [#393](https://github.com/Prism-Shadow/penguin-harness/pull/393)
 
 [中文版](2026-08-21-desktop-terminal-node-pty.zh.md)
 
--- a/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.md
+++ b/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.md
@@ -17,17 +17,20 @@ and a packaged app resolve, and `electron-builder.yml` ships it.
 
 ## Details
 
-- `src/pty-payload.ts` selects and copies node-pty's shipping subset — its manifest, license,
-  `lib/`, the `build/Release` binding compiled at install time, and the `prebuilds/` binaries —
-  dropping sourcemaps, node-pty's own tests, its TypeScript sources, node-gyp inputs, the
-  vendored winpty tree and 44 MB of Windows `.pdb` debug symbols. The staged copy is 5.4 MB and
-  covers darwin arm64/x64 and win32 x64/arm64 alongside the host build.
+- `src/pty-payload.ts` selects and copies node-pty's shipping subset — its manifest, its own MIT
+  license and vendored winpty's, `lib/`, the `build/Release` binding compiled at install time,
+  and the `prebuilds/` binaries — dropping sourcemaps, node-pty's own tests, its TypeScript
+  sources, node-gyp inputs, the vendored winpty sources and 44 MB of Windows `.pdb` debug
+  symbols. The staged copy is 5.4 MB and covers darwin arm64/x64 and win32 x64/arm64 alongside
+  the host build.
 - The staging step restores the exec bit on every `spawn-helper` it copies. node-pty publishes
   that darwin side binary as `0644`, which makes `posix_spawnp` refuse it; fixing the mode
   before the app is packed keeps a signed `.app` under `/Applications`, where the server's
   runtime repair cannot write, out of that failure.
 - `scripts/build-assets.mjs` resolves node-pty from `packages/server`, where pnpm installs it,
-  and fails the build when the install carries no `pty.node`.
+  and fails the build when the staged copy carries no `pty.node` the build host can load.
+  node-pty prebuilds darwin and win32 but not Linux, so a Linux install whose node-gyp step
+  never ran still stages three bindings, none of which that app could load.
 - The `pnpm desktop` preflight checks the staged package, so a stale build says so at launch
   instead of at the first terminal.
 - `scripts/terminal-smoke.mjs` now loads node-pty the way the server bundle does — a bare
--- a/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.zh.md
+++ b/changelog/unreleased/2026-08-21-desktop-terminal-node-pty.zh.md
@@ -1,0 +1,27 @@
+# 桌面终端：node-pty 重新随应用分发
+
+- **Date:** 2026-08-21
+- **Type:** fix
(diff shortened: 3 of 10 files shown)


hiyouga:
@Myriad-Dreamin

hiyouga:
Reviewed the packaging change end to end on Linux. The approach is right and the shipping subset is complete for every target — I checked node-pty 1.1.0's actual install tree and its runtime require graph, and the 34 staged files cover darwin arm64/x64 and win32 x64/arm64 with nothing missing. Three fixes pushed in c018f3f2: one real hole in the build guard, one missing license, two comments that did not survive a check against their sources.

### Fixed

- `packages/desktop/scripts/build-assets.mjs:70` — the guard accepted a payload with no binding the build host can load. `nativeBindings()` matches any `*/pty.node`, and node-pty's tarball *always* carries `prebuilds/darwin-{arm64,x64}` and `prebuilds/win32-{x64,arm64}`. Linux is the only platform node-pty publishes no prebuild for, so an install whose node-gyp step never ran (`--ignore-scripts`, or a checkout that loses the `allowBuilds: node-pty` entry in `pnpm-workspace.yaml`) stages four foreign bindings, passes, and ships a Linux app that 500s on the first terminal — the exact failure this PR exists to prevent. Now gated on `hostBinding()`, which mirrors `lib/utils.js`'s own search order (`build/Release`, then `prebuilds/<platform>-<arch>`). Verified by hiding `node_modules/.pnpm/node-pty@1.1.0/node_modules/node-pty/build` and rebuilding:
  ```
  [build-assets] the node-pty install at …/node-pty has no pty.node for linux-x64 (only
  prebuilds/darwin-arm64/pty.node, prebuilds/darwin-x64/pty.node, prebuilds/win32-arm64/pty.node,
  prebuilds/win32-x64/pty.node) — reinstall it so its build script runs (pnpm-workspace.yaml allows it).
  ```
  The old check printed nothing and exited 0. No false failures: on each of the three CI runners the host binding is present (Linux `build/Release`, darwin/win32 their prebuild), and `desktop-build.yml` builds each OS on its own runner, so host == target. `packages/desktop/src/pty-payload.ts:110` adds `hostBinding()`; two unit tests cover the ordering and the rejection.

- `packages/desktop/src/pty-payload.ts:28` — `deps/winpty/LICENSE` was dropped while the binaries built from it ship. `prebuilds/win32-{x64,arm64}` include `winpty.dll` and `winpty-agent.exe`, compiled from node-pty's vendored `deps/winpty/` (MIT, © Ryan Prichard); node-pty's own `LICENSE` (© Christopher Jeffrey / Microsoft) does not cover it. Added to `SHIPPED_FILES`, with `descendsInto()` extended so a nested shipped file's directory is walked. +1085 bytes; staged tree is 35 files, still 5.4 MB. (`third_party/conpty/` — Microsoft's OpenConsole, also MIT — carries no notice file in the install at all, so there is nothing to copy for `conpty.dll` / `OpenConsole.exe`.)

- `packages/desktop/electron-builder.yml:37` — the splice claim is half right, and I could check it exactly rather than trust it. In app-builder-lib 26.15.3, `getMainFileMatchers` scans for the first non-negated pattern containing `node_modules/` and splices `!**/node_modules/**` immediately before it; with no such pattern it goes into `customFirstPatterns`, which lands at index 0 — i.e. *before* `dist/**/*`. Driving the real `createFilter` + `Minimatch` with both pattern lists:

  | path | with `dist/node_modules/**/*` | without it |
  | --- | --- | --- |
  | `dist/node_modules/node-pty/**` | KEEP | KEEP |
  | `dist/stray/node_modules/foo/index.js` | DROP | KEEP |

  So the staged copy ships either way; naming `node_modules` is what moves the exclusion *behind* `dist/**/*` and keeps any *other* `node_modules` under `dist/` out. Comment reworded to that (verified) statement. Also confirmed nothing staged is caught by electron-builder's own `excludedNames` / `excludedExts` — `LICENSE`, `.node`, `.dll`, `.exe`, `spawn-helper` all survive; `.pdb` is excluded twice over.

- `packages/desktop/src/pty-payload.ts:14` — "`scripts/deploy.mjs` applies the same rule … over the same set of files" is not true. `deploy.mjs`'s `wanted()` keeps `.map` and `*.test.js` and ships no `LICENSE`; this module drops the first two and ships both licenses. Reworded to state the actual relationship, since a reader who believes the old sentence will "sync" one to the other by assumption. Changelog pair updated for the license and guard wording, section for section.

### Findings not fixed

- **A shared constant for the two selection rules — judged not worth it.** `scripts/deploy.mjs` is a plain root-level `.mjs`; `pty-payload.ts` is TypeScript compiled by tsup into `packages/desktop/dist/`. Sharing would mean deploy importing a *build output* of a package it otherwise does not depend on. Leaving them separate is defensible, but the drift is real and already present (`.map` / `.test.js` / `LICENSE`), so the honest cross-reference matters more than the constant. Worth revisiting only if a third copy of the rule appears.
- **macOS signing and notarization of the staged binaries — only a real macOS run settles it.** `hardenedRuntime: true` + `notarize: true`, and the staged tree adds Mach-O files under `Contents/Resources/app/dist/node_modules/node-pty/` (two `pty.node`, two `spawn-helper`). @electron/osx-sign has to pick all of them up or notarization fails / the app is killed at exec. Precedent is good — before #335 the same binaries shipped at `resources/app/node_modules/node-pty/`, signed and notarized fine — but the path moved, and this box cannot check it.
- **`prebuilds/darwin-x64/spawn-helper` is an x64 Mach-O inside the arm64 `.dmg`** (and vice versa). Correct — it is never exec'd there — but if signing ever gets arch-scoped it is the first thing to trip.
- **The packed app is never asserted on.** `terminal-smoke.mjs` genuinely exercises the shipped lookup — `createRequire(pathToFileURL(dist/server.js))("node-pty")` resolves `packages/desktop/dist/node_modules/node-pty`, byte for byte the server bundle's line — but it reads the *source-tree* staged copy, not `out/*-unpacked/resources/app/dist/node_modules/node-pty`. ci.yml's `runtime` job already builds an unpacked app on all three OSes; a one-line existence check on that output would close the last gap between "staged" and "packed" on macOS and Windows, where the `--linux --dir` run in the PR body cannot speak for them.
- **The staged file list is not order-deterministic.** `fs.readdirSync` returns filesystem order, so `stageNodePty`'s return value (and the `[build-assets]` log line) varies by machine; the copied *bytes and file set* are deterministic, which is what packing consumes. The tests `.sort()` before asserting. Cosmetic, left alone.
- **Symlinked subdirectories inside the install would be skipped silently** — `entry.isDirectory()` is false for a symlink, so `copyInto` neither descends nor copies. Not reachable with pnpm's layout (verified: `lib/`, `build/`, `prebuilds/` are all real directories under `.pnpm/node-pty@1.1.0/`), but it is a silent drop rather than an error if some other install layout ever does it.
- **`.github/workflows/ci.yml:64` (pre-existing, context line, left alone)**: "node-pty compiles against the runner's toolchain" is no longer true on darwin. node-pty 1.1.0 ships `prebuilds/darwin-arm64`, and its install script (`scripts/prebuild.js`) exits 0 whenever that directory exists, so no node-gyp build happens on macOS or Windows — only on Linux. What ci-macos actually proves is stronger than the comment says: it loads the very prebuild that ships.
- **Exec bit on a win32 build host**: `fs.chmodSync(…, 0o755)` on the two `prebuilds/darwin-*/spawn-helper` copies is effectively a no-op on Windows (libuv maps chmod to the read-only attribute) and does not throw. Read, not run — though `ci-windows` runs `pnpm build`, so `stageNodePty` does execute there on every PR.

### Verification

Linux, Node v24.18.0, Electron 43.2.0. Everything below was run in this worktree at c018f3f2:

- `pnpm --filter @prismshadow/penguin-desktop build` — staged 35 files / 5.4 MB, `spawn-helper` at `0755` on both darwin prebuilds, `deps/winpty/LICENSE` present, no `.pdb` / `.lib` / `.map` / `.test.js`.
- `node packages/desktop/scripts/terminal-smoke.mjs` → `[terminal-smoke] node-pty spawn OK (Electron Node 24.18.0, ABI 148)`
- `node packages/desktop/scripts/preflight.mjs` → clean.
- Guard negative test (hide `build/`, rebuild) → exit 1 with the message quoted above; restored afterwards.
- Payload completeness: every `require()` in the 13 staged `lib/*.js` resolves to a Node builtin or a staged file; `loadNativeModule`'s computed `require`, `windowsPtyAgent`'s `fork(conpty_console_list_agent)`, `unixTerminal`'s `spawn-helper`, and conpty.cc's `conpty\conpty.dll` next to `conpty.node` all land on shipped files. The win32 prebuild dirs ship complete minus `.pdb`.
- electron-builder matching simulated against the installed app-builder-lib 26.15.3 (`createFilter` + `Minimatch`, real `excludedExts` / `excludedNames`), results in the table above.
- `pnpm --filter @prismshadow/penguin-desktop test` — 100 tests, 10 files, all pass.
- `pnpm typecheck` — 8 projects, clean. `pnpm format` + `pnpm format:check` — clean. `git diff --check` — clean.

**Not exercised here:** no macOS and no Windows run. Specifically unverified on this box — the darwin prebuilds actually dlopening under Electron, `spawn-helper` being exec'd by `posix_spawnp` from the staged copy, code signing / notarization of the staged Mach-O files, and the whole win32 payload (conpty and winpty are copied and pattern-verified, never loaded). `ci-macos`'s smoke step now covers the first two through the shipped lookup; the signing question and Windows remain open until a real run.

PR #394, opened 2026-08-21
feat(web): tabbed dock sidebars for all side elements (Codex-style right/bottom docks)

hiyouga:
Replaces the drawer-style chat side panels and the per-conversation terminal pane system with two uniform, Codex-style dock surfaces on the chat page — right and bottom — where **every side element is a tab**: the subagents panel, the Workspace files panel, the Memory panel, a new **Trace panel**, and any number of terminals.

## What changed

- **Two docks, uniform tabs.** Both docks can be open at once. Tabs switch by click, reorder by drag, move to the other edge by drag (edge bands + drop targets + landing preview) or via the dock's move button. Every tab carries an **always-visible ×** in its own slot — never overlapping the label (crowded tabs truncate the label instead); a panel tab's × closes the panel, a terminal tab's × kills the shell, and the last tab closing puts the dock away. Each dock has a "+" menu (panels / new shell / adopt a shell no conversation holds) and a detach button while a terminal is shown.
- **Toolbar = two pull-open buttons** (bottom dock, right dock — matching the Codex reference), replacing the per-element icons, the pin system and the create menu. An opened dock with no tabs shows a **centered picker** (agents / terminal with its Ctrl+` hint / workspace / memory / trace) choosing what to open there.
- **Per-conversation arrangement** (browser-window style, per the owner's direction): tabs, active tab and open docks switch with the Session; nothing leaks between conversations; a draft's arrangement is handed to the Session its first send creates (self-heal renames follow too); least-recently-used scopes age out of storage (cap 40). Dock sizes (right width via the shared `penguin.panelWidth`, bottom height ratio) stay global preferences. In-stream opens — file cards, subagent chips, memory rows, the spawn auto-open — land on the **right dock by default**.
- **Trace moved into the menu bar's reach.** The Trace panel (file pills + export + the timeline/event view reused from the traces page) is a picker/"+"-menu element; the left-nav Traces entry is removed while `/traces` stays routable for its deep links.
- **Subagents panel: jump to Session.** The identity strip's button opens the selected child's own Session in the chat area.
- **Ctrl+`** walks the conversation's terminal states: front when covered → hide → restore; adopts or creates a shell when this conversation has no terminal tab. Create failures surface as an error toast.
- **Narrow viewports (<1024px)**: the two docks render as one merged bottom surface; the stored arrangement splits back apart on widening. The mobile Sheet variant of the panels is removed.
- **Bug fix (dev-mode trace hang):** the Trace panel's first load died under React StrictMode's double-invoked effects — the rising-edge ref consumed the edge while the cleanup cancelled the fetch, so the skeleton stayed up forever ("switch to the trace tab shows nothing"). The fetch is now a tick-keyed effect (the WorkspaceBrowser pattern), verified against a real `vite` dev server.

## Round 3 (owner feedback)

- **Animations**: opening/closing a dock slides over 200ms (`use-dock-mount` keeps a closing dock mounted through its collapse; the bottom dock's height became measured pixels so its content sits at the settled size while the surface slides — no mid-transition reflow, xterm keeps its grid). Cross-dock moves and conversation switches are marked instant in the store (`instantVersion`) and skip the animation both ways.
- **Resize fix**: the right dock's boundary drag never worked — its handle is a layout *sibling* of the dock, so `closest()` could not reach the box it resizes; both handles now resolve the dock through a ref. The e2e sizes test gains right-width drag coverage.
- **Horizontal bottom picker**; **kill confirmation** on a terminal tab's × (it ends the shell for real); **detach round trip** — closing the detached window returns the tab to the conversation and dock it left, even if another conversation is on screen by then.
- **Per-conversation isolation re-verified over client-side navigation** (sidebar switches, a Models-page excursion, the draft page): nothing leaks; the leak reported during testing was the round-1 global-layout build.
- Also fixed en route: mounting the bottom dock used to slide from 0 even for instant changes (the host height state started at 0 and the late measurement re-triggered the transition) — the initial measurement is now synchronous.

## Round 4 (owner feedback)

- **Dock terminals start in the conversation's Workspace** instead of `~`: the shell opens where the agent's files are, and since the server names a shell after its directory the tab says where it landed. While drafting, the Workspace picked in the draft applies (the temporary Workspace has no directory until the Session exists, so it keeps the home default). The path is published by the surface that knows it — the chat page for a Session, the draft page for the picked one — because the Ctrl+` listener runs at module scope with no React context; a Workspace the server rejects (`cwd_not_found`, deleted since) retries once at home rather than leaving the user with no terminal. Covered by a new e2e test that runs `pwd` in a dock shell and matches the session's Workspace directory.

## Compatibility (user decision surfaced)

Stored dock state moves to a new localStorage key (`penguin.dock.layout`, now a per-scope map). The old keys (`penguin.terminal.dock`, `penguin.terminal.dockRatios`, `penguin.terminal.tabOrder`) are no longer read or written and are left in place — no migration. One-time effect: docks start closed after the update; shells keep running server-side and any shell no conversation holds can be pulled back in from a dock's "+" menu or the picker's terminal row. Recorded in the changelog entry's Compatibility section.

## Design notes

- The dock store (`features/dock/dock-state.ts`) is a pure module-level store (localStorage-guarded, scope map with LRU eviction, unit-tested); shell round-trips live in `dock-terminal.ts`; the chrome is `dock-panel.tsx` (strip, picker, "+" menu, drag, resize) with the 2-target drag overlay in `dock-drag.tsx`. AppLayout points the scope at the route's conversation (layout effect, before the docks paint). Panel bodies stay owned by the chat page and are injected via `renderPanel`; xterm views keep living in the existing pool and are adopted into tab bodies by DOM handoff, so tab/dock churn never reconnects a shell.
- Panel kinds are singletons per conversation; terminals are per-shell tabs owned by the conversation that tabbed them (`unownedTerminals` is the cross-scope adoption set). The subagents auto-open tracker is unchanged: auto-open on the current task's first live spawn, re-armed per task, manual close respected; task boundaries close nothing.
- Panel jump commands (locate file / locate memory / focus subagent) reset when their tab closes or the Session switches, so a re-added tab never replays a stale command.

## Verification

- Repo-wide `pnpm typecheck` clean; `pnpm format:check` clean; web unit tests **1157 passed** including the rewritten `test/dock-state.test.ts` (scope independence, hand-over/adopt, cross-scope prune and adoption, picker visibility, last-tab-close, toggle semantics, move/reorder, Ctrl+` walk, persistence round-trip incl. malformed storage).
- Playwright (local stack, mock LLM, fresh data root): `dock.spec.mjs` (9 tests: two-toggle + picker flow, mixed tabs, per-conversation isolation round trip, move/drag, resize+reload, detach, Ctrl+` lifecycle, failed-create toast), `mobile-dock.spec.mjs`, `files-switch.spec.mjs` (rewritten to the structural per-scope guarantee), `subagent.spec.mjs` (4), `background-subagent.spec.mjs`, `outline.spec.mjs`, `processes.spec.mjs`, layout rail — all pass. `chat.spec.mjs`'s one failure and two `layout.spec` failures reproduce identically on main's own build (pre-existing).
- **Dev-server (StrictMode) verification** of the reported hang: with `vite` serving the app, the Trace tab loads on first open and after switching among several tabs — no stuck skeleton, no page errors. Screenshot-driven visual checks of the two-toggle toolbar, both pickers, always-visible tab ×s, and the trace timeline.


packages/web/src/features/dock/dock-terminal.ts
```diff
@@ -0,0 +1,158 @@
...
+  }, 600);
+}
+
+/**
+ * Ctrl+`: hide the shown terminals, or bring them back — and with no terminal tab in
+ * this conversation, adopt or create a shell (the async tail the store's synchronous
+ * toggle hands off).
+ */
+export function toggleTerminal(): void {
+  if (!toggleTerminalDocks()) void openTerminalInDock();
```
hiyouga, line 141: **The module-scope Ctrl+` listener still runs on every route, but the docks moved into chat-page.tsx — off /chat it spawns a real pty into the invisible `~none` scope, which nothing renders and which `unownedTerminals` then marks as owned.**

On /agents (or /models, /skills, /login, or the detached /terminal window) AppLayout has set the scope to `~none`. Ctrl+` -> toggleTerminalDocks() finds no terminal tab and returns false -> openTerminalInDock() adopts or POSTs a shell -> addTerminalTab writes a tab + open:true into `~none`. No DockPanel exists on that route, so the user sees nothing while a pty runs; TerminalDockRuntime (mounted in AppLayout) still opens a WebSocket into a detached container. unownedTerminals() counts `~none` as an owner, so the shell disappears from every dock's "+" menu and picker. On /login the POST 401s and the user gets a stray "终端创建失败: POST /api/terminals → 401" toast. Later, setDockScope's hand-over (mayHandOver = scope === NO_SCOPE) can graft that surprise tab onto the first conversation opened that has no arrangement of its own. The comment at line 150 asserts the opposite ("the toggle just flips hidden state"), which was true only while the dock lived in AppLayout.

packages/web/src/features/dock/dock-panel.tsx
```diff
@@ -0,0 +1,841 @@
...
+  const bodyRef = useRef<HTMLDivElement | null>(null);
+  useEffect(() => {
+    const body = bodyRef.current;
+    if (!body || !active) return;
+    const container = terminalViewContainer(id);
+    body.appendChild(container);
+    return () => {
+      if (container.parentElement === body) body.removeChild(container);
+    };
+  }, [id, active]);
```
hiyouga, line 199: **TerminalBody's DOM-handoff effect keys on `[id, active]`, but the view pool destroys the pooled container as soon as the dock stops being visible — while use-dock-mount keeps the panel mounted for the 200ms collapse. Reopening inside that window leaves the terminal body permanently blank.**

Bottom dock shows terminal T. Click the dock × (or press Ctrl+`): toggleDock/hideView set open=false without bumping `instant`, so useDockMount lingers the same DockPanel (and the same TerminalBody, active still true) for DOCK_TRANSITION_MS. Meanwhile shownIds() drops T and TerminalDockRuntime runs `containers.get(T)?.remove(); containers.delete(T)` — the xterm node is yanked out of the still-visible collapsing dock, so the close animation slides an empty white box (exactly what the deleted comment in terminal-view-pool.tsx warned about: "tearing it down would make the panel slide in over a blank box"). Press the toggle again within 200ms: useDockMount cancels the linger and returns the view, but DockPanel never unmounted so TerminalBody's `[id, active]` deps are unchanged and the effect does not re-run; the pool mints a brand-new container and portals TerminalView into it, attached to nothing. The dock reopens with the terminal tab selected, data-status "ready", and an empty body until the user switches tabs and back.

packages/web/src/features/dock/dock-state.ts
```diff
@@ -0,0 +1,716 @@
...
+  switchScope(sessionId, true);
+}
+
+function switchScope(target: string, mayHandOver: boolean): void {
+  if (target === scope) return;
+  instant += 1;
+  persist();
+  const staged = scopes[scope];
+  // Never clobbers: a target with an arrangement of its own keeps it.
+  const handedOver = mayHandOver && staged !== undefined && scopes[target] === undefined;
```
hiyouga, line 285: **adoptDockScope() takes only the target id and hands over whatever `scope` happens to be current — both callers invoke it after several awaits, so a navigation mid-send transplants an unrelated conversation's arrangement onto the newborn Session and deletes it from its owner.**

draft-view.tsx:647 calls adoptDockScope(res.sessionId) after `await api.createSession` -> `await api.postTask` -> `await api.getSession`. If the user clicks conversation B in the sidebar while those are in flight, AppLayout's layout effect sets scope = B. When the send lands, switchScope(newId, true) sees `staged = scopes[B]` and `scopes[newId] === undefined`, so it deletes key B from `scopes` and re-keys B's whole arrangement under the new session id: B's docks are gone permanently and the new Session inherits tabs it never had. chat-page.tsx:954 (syncHealedSessionId, after `await reloadSessions()`) has the same shape. Fix shape: adoptDockScope(fromId, toId) that no-ops when fromId !== scope.

packages/web/src/features/dock/dock-terminal.ts
```diff
@@ -0,0 +1,158 @@
...
+    let created: TerminalInfo;
+    try {
+      created = await create(workspaceCwd ?? HOME_CWD);
+    } catch (err) {
+      if (workspaceCwd === null || !isBadCwd(err)) throw err;
+      console.warn(`[terminal] Workspace unusable as cwd, opening in ${HOME_CWD}:`, err);
+      created = await create(HOME_CWD);
+    }
+    noteTerminalCreated(created);
+    addTerminalTab(created.id, position);
```
hiyouga, line 71: **createShellInDock / openTerminalInDock capture the dock POSITION at click time but not the SCOPE, so they re-enter the store through the global current scope after their awaits — detachTerminal in the same file captures `currentDockScope()`, which makes this an inconsistency rather than a choice.**

On conversation A the user clicks "+ → New shell" (or presses Ctrl+`), then clicks conversation B in the sidebar while the POST is in flight. The POST resolves and `addTerminalTab(created.id, "bottom")` runs against scope B: B's bottom dock is forced open with a shell that was started in A's Workspace, and A never gets the tab at all. The cwd_not_found retry (line 68, a second round trip) and openTerminalInDock's `await fetchJson("/api/terminals")` before `addTerminalTab(adoptable, position)` (line 99) both widen the window.

packages/web/src/features/dock/dock-state.ts
```diff
@@ -0,0 +1,716 @@
...
+/** Brings an existing tab on screen: opens its dock, makes it active, focuses its dock. */
+export function activateTab(key: string): void {
+  const found = findTab(key);
+  if (!found) return;
+  activate(found.position, key);
+  persist();
+  notify();
+}
+
+function insertTab(tab: DockTab, position: DockPosition): void {
```
hiyouga, line 443: **insertTab() removes a tab from the other dock and repairs that dock's `active`, but never sets `open = false` when the source empties. Only moveTab compensates externally (line 485); openPanel, addTerminalTab and the picker do not.**

Bottom dock is open holding ONLY the Memory tab. From the right dock's "+" menu choose Memory -> openPanel("memory", "right") -> insertTab moves the tab, leaving `layout.bottom = {tabs: [], active: null, open: true}`. dockViews() still lists bottom, and DockPanel renders `tabs.length === 0 ? <DockPicker/>` — so a "choose what to open here" surface is left sitting at the bottom of the chat page after an action the user read as "move Memory to the right". Same via addTerminalTab's adopt rows and DockPicker's own choose. This violates the invariant removeTab's own comment states (line 465: "no empty surface (nor a surprise picker) should linger"). The unit test at test/dock-state.test.ts:47 asserts only `dockTabs("right")).toHaveLength(0)`, never isDockVisible, so it passes. These paths also skip the `instant += 1` bump moveTab does, so the cross-dock move plays the two-motion animation instantVersion exists to suppress.

packages/web/src/features/dock/dock-panel.tsx
```diff
@@ -0,0 +1,841 @@
...
+      }
+      if (!active || !candidate || candidate === tabHome(id)) return;
+      moveTab(id, candidate);
+    },
+    onCancel: clearTabDrag,
+  });
+
+  // ------------------------------------------------------------------------ boundary resize
+  const [resizing, setResizing] = useState(false);
+  const sideWidth = usePanelWidthValue();
```
hiyouga, line 457: **Deleting the three panel shells orphaned usePanelWidth() — it now has zero call sites, and with it the only window-resize clamp that kept the side panel from crowding out the chat column.**

usePanelWidth's effect (use-panel-width.ts:175-179, `window.addEventListener("resize", () => writeWidth(Math.min(readWidth(), maxWidthFor(window.innerWidth))))`) was registered by files-panel / memory-panel / subagents-panel, all deleted here. The dock subscribes with usePanelWidthValue(), which registers nothing. So: drag the right dock to ~700px on a wide monitor (allowed — maxWidthFor caps at min(720, w*0.5)), then shrink the window to ~1100px. maxWidthFor(1100) = 550, but nothing re-clamps: the dock stays 700px and the chat column is squeezed to ~390px and keeps shrinking. Only a reload (initialWidth re-clamps at first read) or a double-click on the handle fixes it. Below 1024px the docks merge so the symptom hides again, which makes it look intermittent. Also dropped with the hook: the `document.body.style.cursor/userSelect` lock during the drag.

packages/web/src/features/dock/dock-state.ts
```diff
@@ -0,0 +1,716 @@
...
+ * Ctrl+`'s toggle, on the store's synchronous half: hides the docks that hold this
+ * conversation's terminal tabs when one is shown; restores/activates the newest terminal
+ * tab when there are tabs but none on screen. Returns false when this scope has no
+ * terminal tab at all — the caller (dock-terminal.ts) then adopts or creates a shell,
+ * which is async.
+ */
+export function toggleTerminalDocks(): boolean {
+  const ids = terminalTabIds();
+  if (ids.length === 0) return false;
+  const holders = (["right", "bottom"] as const).filter((position) =>
```
hiyouga, line 677: **toggleTerminalDocks() computes `holders` as docks that CONTAIN a terminal tab, not docks that are SHOWING one, then closes all of them — so Ctrl+` hides a panel dock the user was reading, and the restore re-points it at the terminal instead of the panel.**

Right dock shows the Workspace panel and also holds an inactive terminal tab t2; bottom dock shows terminal t3. Ctrl+` -> `anyShown` is true (bottom), so BOTH docks close — the Workspace panel disappears even though no terminal was visible there. Ctrl+` again -> showTerminal(ids[ids.length - 1]); terminalTabIds() orders bottom-then-right, so the last id is t2: the right dock reopens showing the TERMINAL instead of Workspace, and the bottom dock stays hidden. The hide/restore pair is not an identity round trip.

packages/web/src/features/dock/dock-state.ts
```diff
@@ -0,0 +1,716 @@
...
+/**
+ * Below 1024px (the breakpoint the pre-dock panels used) the docks merge into one bottom
+ * surface. The store owns the media query so every narrow-aware decision (merged active,
+ * toggle semantics, the view models) lives in one place; components just render dockViews().
+ */
+let narrow = false;
+if (typeof window !== "undefined" && typeof window.matchMedia === "function") {
+  const query = window.matchMedia("(max-width: 1023.98px)");
+  narrow = query.matches;
+  query.addEventListener("change", (event) => {
```
hiyouga, line 327: **The narrow/wide media-query handler notifies without bumping `instantVersion`, so crossing the 1024px breakpoint makes useDockMount linger the disappearing dock for 200ms while the merged view already renders the same tabs — every panel body is mounted twice.**

Right dock open with the Workspace tab; the user drags the window narrower past 1024px. dockViews() collapses to one merged bottom view carrying right's tabs, but rightMount gets view=null with animate === true (instant unchanged), so use-dock-mount.ts:51 sets lingering=true and keeps the old right view rendered for 200ms. During that window chat-page.tsx:2074 and :2086 both render a DockPanel whose strip holds the workspace tab, and renderPanel('workspace', …) returns a second WorkspaceBrowser — whose initial listWorkspaceFiles effect (workspace-browser.tsx:200-215) runs on mount regardless of `active`, so the directory listing fetch fires twice. For a terminal tab, both TerminalBodys appendChild the SAME pooled container, so the node is yanked between them mid-transition. Repeats on every drag across the breakpoint. mobile-dock.spec.mjs pins its viewport with test.use(), so the transition is never exercised. The deleted use-files-panel.ts stated the invariant outright: "docked and Drawer are mounted mutually exclusively; mounting both at once would cause WorkspaceBrowser's own data requests to fire twice."

packages/web/src/features/dock/dock-terminal.ts
```diff
@@ -0,0 +1,158 @@
...
+ * the tab leaves the strip while the shell lives in the window, and closing the window
+ * puts the tab back into the conversation and dock it left — even if the user is looking
+ * at another conversation by then. There is no cross-window close event, so a slow poll
+ * on the handle's `closed` flag watches for it; the restore is skipped when the shell
+ * ended inside the window or the user already pulled it back in through a "+" menu.
+ * (The handle is why the window is NOT opened with "noopener" — window.open would then
+ * return null; the target is this same app's own page.)
+ */
+export function detachTerminal(id: string, position: DockPosition): void {
+  const fromScope = currentDockScope();
```
hiyouga, line 117: **detachTerminal captures the scope id at detach time; if adoptDockScope renames that scope before the popup closes, restoreTerminalTab writes the shell into a dead key — and for a draft that key is "new", so the shell surfaces in the NEXT new conversation.**

On /chat/new: Ctrl+` creates a shell (tab in scope "new"), click Detach -> fromScope = "new", popup opens, removeTab fires and scope "new" becomes empty so persist() drops it. Send the first message -> adoptDockScope(sessionId); `staged = scopes["new"]` is undefined so nothing is handed over. Close the popup -> the 600ms poll calls restoreTerminalTab("new", id, position), which takes the `scopeId !== scope` branch, builds a fresh emptyScope() and writes `scopes["new"] = {bottom: {tabs:[terminal], open:true}}`. The terminal never returns to the conversation it was detached from, and clicking "New chat" later pops the bottom dock open with a stranger's shell. Same class via syncHealedSessionId -> adoptDockScope(respondedId). unownedTerminals also counts the phantom scope as an owner, so the shell disappears from every "+" adopt list too.

packages/web/src/features/chat/draft-view.tsx
```diff
@@ -634,9 +641,9 @@ export function DraftView({
         const fresh = await api.getSession(res.sessionId).catch(() => null);
         add(fresh?.session ?? created.session);
         if (!keepDraft) discardDraft();
-        // The draft now has an id of its own, so its terminals move with it: every draft
-        // shares one dock scope, and anything left behind under that key would surface in
-        // the NEXT new conversation instead (terminal-dock-state.ts).
+        // The draft now has an id of its own, so its docks move with it: anything left
+        // behind under the draft's scope would surface in the NEXT new conversation
+        // instead (dock-state.ts).
         adoptDockScope(res.sessionId);
```
hiyouga, line 647: **adoptDockScope is wired only into the draft's send path; parkActiveDraft — the other way a draft gets an id of its own — never hands the scope over, so a parked draft loses its docks and the fresh new chat inherits them.**

On /chat/new (dock scope "new") the user opens a terminal and the Workspace tab, types some text, then clicks "New chat" in the sidebar. parkActiveDraft moves the typed draft into a `draft-<hex>` conversation row but navigates back to /chat/new, so the route — and therefore the dock scope — never changes. The parked draft row opens with an empty arrangement while the blank new chat inherits the previous draft's terminal and Workspace tabs: exactly the leak this file's own comment says adoptDockScope prevents ("anything left behind under the draft's scope would surface in the NEXT new conversation instead"). None of the parkActiveDraft call sites touch the dock scope: sidebar.tsx:794, chat-page.tsx:1343, app-layout.tsx:74, agents-page.tsx:162, skills-page.tsx:244, skills-tab.tsx:219.

packages/web/src/features/terminal/terminal-list.ts
```diff
@@ -137,7 +108,7 @@ export function refreshTerminals(): Promise<void> {
         if (!listed.has(id)) pendingKills.delete(id); // fully gone: nothing left to hide
       }
       const live = data.terminals.filter((t) => t.alive && !isPendingKill(t.id));
-      pruneAssignments(new Set(live.map((t) => t.id)));
+      pruneTerminalTabs(new Set(live.map((t) => t.id)));
```
hiyouga, line 111: **refreshTerminals() hands a concurrent caller the already-in-flight promise, so a GET issued before a shell existed can resolve after it — and pruneTerminalTabs now also CLOSES the dock, turning a stale race into "clicking + New shell does nothing".**

A refresh is already in flight (the 30s poll, the window-focus handler, or one of killTerminal's 300ms/1500ms timers). The user clicks "+ New shell"; the POST returns, noteTerminalCreated + addTerminalTab add the tab and open the dock; createShellInDock's `finally { void refreshTerminals(); }` hits `if (inflight) return inflight` and issues nothing. When the stale response lands, `live` does not contain the new id, so pruneTerminalTabs strips the just-added tab from every scope and — because it was the dock's only tab — pruneArea sets `state.open = false` and persists. The user kills a shell, clicks "+ New shell" a second later, and the dock closes again with no error; the pty keeps running, invisible, until the next 30s poll re-lists it as unowned. The dock-closing half is new in this PR (main's pruneAssignments only dropped assignments), and it now also silently closes docks in conversations the user is not looking at.

packages/web/src/features/dock/dock-terminal.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+
+/**
+ * Puts a terminal on screen in `position` (or the bottom dock): the newest live shell no
+ * conversation holds is adopted — one started through the API or the CLI, or whose tab
+ * was closed — rather than answered with a second shell running beside it; only when
+ * every live shell is already tabbed somewhere (or none exists) is a new one created.
+ */
+export async function openTerminalInDock(position?: DockPosition): Promise<void> {
```
hiyouga, line 94: **The single-flight guard around shell resolution/creation was dropped — openTerminalInDock has no in-flight set, so two quick triggers each run the full list-then-create path and spawn two ptys.**

The user presses Ctrl+` twice in quick succession in a conversation with no terminal tab (or double-clicks the picker's terminal row, dock-panel.tsx:721). Both calls run toggleTerminalDocks() -> false (the first create has not resolved, so no tab exists yet), both await GET /api/terminals, both find no adoptable shell, and both call createShellInDock() -> two ptys spawned, two tabs in the bottom dock. The deleted terminal-dock.tsx had exactly this guard: "/** One resolution at a time per pane — mount effects can fire in quick succession. */ const resolving = new Set<DockPosition>();" with a finally-delete. refreshTerminals in the same feature already shares its in-flight promise, so the pattern exists in the codebase.

packages/web/src/features/dock/dock-panel.tsx
```diff
@@ -0,0 +1,841 @@
...
+  const closeTab = useCallback((tab: DockTab, label: string) => {
+    if (tab.kind === "terminal") {
+      setConfirmKill({ id: tab.terminalId, label });
+      return;
+    }
+    removeTab(tabKey(tab));
+  }, []);
+  const killConfirmed = useCallback(() => {
+    if (!confirmKill) return;
+    void killTerminal(confirmKill.id);
```
hiyouga, line 330: **A failed kill leaves the UI and the server permanently disagreeing with no feedback: killTerminal never inspects `res.ok` and pendingKills actively hides the survivor for 10s, while the tab has already been removed optimistically.**

The user confirms the kill on a terminal tab's ×. killConfirmed fires `void killTerminal(id)` and immediately removeTab(...). The DELETE resolves non-2xx (403/409/500) — fetch only rejects on network failure, so terminal-list.ts's catch is never reached and the status is never checked. pendingKills.set(id, Date.now() + 10_000) filters the still-alive shell out of the reconciling refreshes at 300ms and 1500ms, so nothing contradicts the optimistic removal. After 10s the shell reappears in liveTerminals() but the tab is gone; only a "+" adopt list can find it, and only if no other scope claims it. Unlike createShellInDock, which toasts its failures, the user gets nothing. Needs `if (!res.ok) throw`, a toast, and pendingKills.delete(id) on failure.

packages/web/src/features/dock/dock-panel.tsx
```diff
@@ -0,0 +1,841 @@
...
+    },
+    onMove: (event) => {
+      const pane = rootRef.current?.getBoundingClientRect();
+      if (!pane) return;
+      if (horizontal) {
+        // The ratio's basis is the chat page column ([data-dock-host]), the same box the
+        // rendered height is computed from below.
+        const host = document.querySelector("[data-dock-host]")?.getBoundingClientRect();
+        if (!host || host.height === 0) return;
+        setBottomRatio((pane.bottom - event.clientY) / host.height);
```
hiyouga, line 478: **The bottom dock's resize drag (and the tab-reorder drag) persist the ENTIRE 40-scope layout map to localStorage on every pointermove, while the right dock's identical gesture three lines below defers persistence to onEnd.**

Dragging the bottom dock's boundary calls setBottomRatio on every pointermove (60-120 Hz); setBottomRatio -> persist() (dock-state.ts:196-225) spreads the scopes object twice, rebuilds every stored scope's area objects, maps every tab through tabKey, JSON.stringifies the whole payload and blocks on a synchronous localStorage.setItem — with MAX_SCOPES=40 conversations that is several KB re-written per frame, on the main thread, while the user is dragging. reorderDock (dock-state.ts:517) has the same shape: every midpoint crossing mid-drag is another full-map write. The correct pattern is in the same handler at line 485: `if (!horizontal) persistPanelWidth(); // once per drag, not per frame`, and use-panel-width.ts:145 spells out why. Split setBottomRatio into an in-memory setter plus a persistBottomRatio() called from onEnd/onDoubleClick.

packages/web/src/features/dock/dock-state.ts
```diff
@@ -0,0 +1,716 @@
...
+    // Re-inserted last so key order is least-recently-touched first, which is what the cap
+    // below evicts by.
+    const { [scope]: _previous, ...rest } = scopes;
+    scopes = { ...rest, [scope]: layout };
+  }
+  const keys = Object.keys(scopes);
+  if (keys.length > MAX_SCOPES) {
+    scopes = Object.fromEntries(keys.slice(keys.length - MAX_SCOPES).map((k) => [k, scopes[k]!]));
+  }
+  writeJson(LAYOUT_KEY, {
```
hiyouga, line 212: **persist() rewrites the entire `penguin.dock.layout` blob from a `scopes` map snapshotted once at module load, with no `storage` listener — and this PR gives the detached /terminal popup its own copy of the store plus the global Ctrl+` handler, so one keypress there overwrites the main window's whole arrangement.**

detachTerminal opens /terminal?id=… same-origin (deliberately without noopener), which boots the full bundle: dock-state.ts loads `scopes` from localStorage once, and dock-terminal.ts registers the hotkey. Ctrl+` in the popup (or in any second app tab) reaches addTerminalTab/toggleDock -> persist(), which serializes the popup's stale scopes map back to the single key. Every dock change the main window made since the popup opened — tabs added, docks opened, other conversations' arrangements — is silently discarded, and after a main-window reload the popup's `~none` entry becomes hand-over material for the first conversation opened. The old system wrote whole maps too, but it only held terminal panes; the blast radius now includes every panel arrangement in every conversation.

packages/web/src/features/dock/dock-terminal.ts
```diff
@@ -0,0 +1,158 @@
...
+  const listed = await fetchJson<{ terminals: TerminalInfo[] }>("/api/terminals").catch(() => null);
+  const live = (listed?.terminals ?? liveTerminals()).filter((t) => t.alive);
```
hiyouga, line 96: **openTerminalInDock adopts from a RAW `GET /api/terminals` and filters only on `t.alive`, bypassing terminal-list's `pendingKills` window — so it adopts the shell the user just killed instead of starting a new one.**

The user confirms a terminal tab's × (the dock's last tab). killConfirmed fires killTerminal(id) — which only records `pendingKills.set(id, now + 10_000)` and sends DELETE (SIGHUP; the pty exits asynchronously) — and removeTab, so the dying shell is now UNOWNED. The user presses Ctrl+` again: toggleTerminalDocks() returns false (no terminal tabs left), openTerminalInDock() GETs the server list directly, the shell is still reported alive:true, and `unownedTerminals(...).at(-1)` picks it — it is the newest, so this is the likely case rather than a rare one. addTerminalTab reopens the dock on a shell that is exiting; the next refreshTerminals() (which DOES apply isPendingKill, terminal-list.ts:110) prunes the tab and closes the dock again. Net effect: Ctrl+` flashes a dock open and shut and leaves the user with no shell. Note `liveTerminals()` — the fallback on the same line — is correctly filtered; only the preferred path is not. isPendingKill is module-private in terminal-list.ts (line 54, not exported), so this file cannot apply it even if it wanted to: the filtered view needs to be exported, or the adopt path should go through refreshTerminals()/liveTerminals().

packages/web/src/features/dock/dock-panel.tsx
```diff
@@ -0,0 +1,841 @@
...
+  const overlayActive = headerDrag.active || tabDrag.active;
+  const overlayCandidate = headerDrag.active ? headerDrag.candidate : tabDrag.candidate;
+
+  const terminalOrdinals = new Map<string, number>();
+  tabs.forEach((tab) => {
+    if (tab.kind === "terminal") terminalOrdinals.set(tab.terminalId, terminalOrdinals.size + 1);
+  });
+
+  const header = (
+    <header
```
hiyouga, line 631: **Both docks lost every accessible affordance the panels they replace had: no accessible name (the panels' <h4> titles are gone), no role="tablist"/role="tab"/aria-selected on the strip, and — because the mobile Sheet variant was deleted — no Escape-to-close, no role="dialog" and no focus restore on the narrow merged surface.**

A screen-reader user on desktop hears an unlabeled group of buttons where files-panel/memory-panel/subagents-panel used to announce themselves (`<h4 className="min-w-0 flex-1 truncate text-sm font-semibold">{S.files.title}</h4>`, files-panel.tsx:91 on main). The active tab is conveyed only by `data-active` plus a background colour, never by aria-selected, so which tab is current is never announced. Below 1024px the merged dock replaces the Sheet, which supplied `role="dialog" aria-modal="true" aria-label={title}` (sheet.tsx:276), an Escape handler (sheet.tsx:172) and focus restoration (sheet.tsx:183) — Escape now does nothing on the merged dock and nothing moves focus into it on open or back out on close. The e2e migration is the tell: every `getByRole("heading", { name: "智能体面板" })` / `getByRole("heading", { name: "文件" })` assertion had to be rewritten to a data-testid/data-tab-id attribute selector, because no role-based handle on these surfaces survives.

packages/web/src/features/chat/chat-page.tsx
```diff
@@ -1555,6 +1449,63 @@ export function ChatPage() {
...
+  const bottomMount = useDockMount(views.find((view) => view.position === "bottom") ?? null);
+  const terminalSupported = terminalApiSupported();
+
+  /**
+   * The panel tabs' bodies. Keyed by Session where the view starts over per conversation
+   * (agents / memory / trace); the Workspace browser instead re-binds through its props —
+   * its own handled-once request guard is what the conversation-switch e2e covers.
+   */
+  const renderPanel = (kind: PanelKind, active: boolean): ReactNode => {
+    if (!selected) return <EmptyState title={panelLabel(kind)} description={S.dock.draftEmpty} />;
```
hiyouga, line 1469: **Every panel tab body renders the DRAFT placeholder ("发送第一条消息后可用" / "Available once the conversation starts") whenever `selected` is null — which includes a real, existing conversation whose Session row has not loaded yet.**

Reload /chat/<sessionId> with a restored arrangement (say workspace + trace tabs). `sessions` starts empty, so `selected = sessions.find(...) ?? null` (line 373) is null, while the docks render unconditionally (lines 2073/2085, outside the `{selected && …}` toolbar guard). For the whole session-list fetch every dock tab claims the conversation has not started, while the chat column beside it correctly shows a loading Skeleton — two contradictory states side by side. The same wrong copy shows for a deep-linked Session for the duration of the `api.getSession` probe (chat-page.tsx:592-616), and right after the new openAsSession jump when the child Session is not yet in the paged list. `S.dock.draftEmpty` is documented in strings.ts:72 as "A session-bound panel's body on the draft page, where no Session exists yet" — the branch fires well outside that case. renderPanel has no sessionsLoading / probeFailedId distinction at all; it needs a Skeleton for "not resolved yet" separate from the draft placeholder.

PR #395, opened 2026-08-21
fix(web): an env-backed model counts as having a key

hiyouga:
## Problem

The has-key / no-key judgement in the chat **model picker** was not aligned with the one in the **model library**.

The library card already reads an env-backed row as configured: with no stored key it falls back to `envKeyMasked` and prints that variable's masked value. The picker filtered through `visibleChatModels` → `hasConfiguredKey`, which was `!!m.credential?.apiKeyMasked` only. A model whose key comes from `ANTHROPIC_API_KEY` was therefore pushed behind the "show models without key" expander and tagged with the struck-through key icon — on the same screen where the library printed its masked value.

The distinction that matters: `envKey` is only the **name** of a fallback variable and must keep counting for nothing. `envKeyMasked` is the server's proof that the variable currently holds a non-empty value (`project-config-service.ts`, DTO documented in `api/types.ts`). So the aligned rule is **stored masked key OR `envKeyMasked` present**.

## Change

One predicate, three surfaces.

- `hasConfiguredKey` (`model-grouping.ts`) is now `!!credential?.apiKeyMasked || !!envKeyMasked`, and `ModelCredentialRowLike` gained `envKeyMasked?: string`. Everything downstream follows without further edits: the picker's default list, its struck-through key icon and the "show models without key" count (derived from `visibleChatModels`).
- `models-page.tsx`'s `hasKey` was rebuilt on top of that predicate instead of restating it. It adds only what the DTO shape cannot carry: a key typed into the dialog counts before it is saved, and `clearApiKey` drops the **stored** key only — an environment variable cannot be cleared from there, so an env-backed row keeps its key through a clear. That is why `hasKey` masks out `credential` when `clearApiKey` is set rather than OR-ing the two notions together.
- The card's key status expression moved into `keyStatusText`, guarded by the same `hasKey`, so the card and the picker cannot drift apart again about which rows are "not configured". Display priority is unchanged: stored mask → typed-but-unsaved reads as configured → env mask → "no key".
- The chat credential guide (`chat-page.tsx`) gated its one-time "no model credential configured" dialog on `!def?.credential?.apiKeyMasked`; it now asks `hasConfiguredKey`. The comment right above it already said the guide exists so as not to nag "users who … use environment variables instead", so this was the same bug in the surface that reasoned about it explicitly.
- The three docstrings that argued for the old rule in prose (`model-grouping.ts`, `model-select.tsx`, `models-page.tsx`) were rewritten; stale reasoning left behind is worse than none.

## Deliberately left alone

- `models-page.tsx`'s `envHintKey` / `apiKeyHint` and the dialog's two masked-key blocks read `credential?.apiKeyMasked` on purpose: they choose **which source to display or promise**, and a stored key legitimately wins over the environment there. They are not "has a key" judgements.
- `packages/cli/src/commands/config.ts` masks `api_key` when dumping a Project's model table. That is a verbatim view of the TOML file with no env-fallback notion at all, not a judgement about whether a model can authenticate.
- `server/src/http/errors.ts`'s `modelCredentialMissing` is raised only when the provider SDK itself fails at client-construction time. The SDK reads the environment variable, so that path already accounts for the env fallback and is the real authority.
- No string changes: `S.models.noKey`, `S.models.showModelsWithoutKey` and `S.project.noCredentialBody` all stay accurate under the new rule.

## Verification

- `pnpm --filter @prismshadow/penguin-core... build` — needed so `@prismshadow/penguin-core` type declarations exist for the web typecheck. Success.
- `pnpm --filter @prismshadow/penguin-web test` — 90 files, 1166 tests, all passing. New cases: `test/model-grouping.test.ts` pins that an `envKeyMasked` row counts, that an `envKey`-name-only row still does not, and that `visibleChatModels` lists an env-backed model instead of hiding it (with the expander count dropping to the two genuinely key-less rows); `test/models-input.test.ts` pins `hasKey` / `keyStatusText`, including a cleared stored key that stays configured through its environment variable.
- Playwright e2e not run (known pre-existing local failures); no e2e spec references the key filter, the "no key" mark or the credential guide.

### DIFF
--- a/changelog/unreleased/2026-08-21-model-key-env-detection.md
+++ b/changelog/unreleased/2026-08-21-model-key-env-detection.md
@@ -1,0 +1,15 @@
+# A model backed by an environment variable counts as having a key
+
+- **Date:** 2026-08-21
+- **Type:** fix
+- **Scope:** `web`
+
+[中文版](2026-08-21-model-key-env-detection.zh.md)
+
+The chat model picker judged "has an API key" by the stored masked key alone, while the model library already showed a model backed by an exported environment variable as configured. A model whose key came from `ANTHROPIC_API_KEY` was therefore pushed behind the picker's "show models without key" expander and marked with the struck-through key icon, on the same screen where the library card printed that variable's masked value. Both surfaces were put on one rule: a stored key **or** an env fallback the server proved is set.
+
+## Details
+
+- `hasConfiguredKey` (`model-grouping.ts`) became the single predicate behind every surface — the picker's default list, its struck-through key icon and the "show models without key" count all read it — and `ModelCredentialRowLike` gained `envKeyMasked`. `envKey` still counts for nothing: it is only the NAME of a fallback variable and says nothing about whether that variable holds a value, while the server emits `envKeyMasked` only for one that currently does.
+- The model page's own `hasKey` was rebuilt on top of that predicate and now adds only what the DTO shape cannot carry: a key typed into the dialog counts before it is saved, and `clearApiKey` drops the **stored** key only — an environment variable cannot be cleared from there, so an env-backed row keeps its key through a clear. The card's key status line moved into `keyStatusText`, guarded by the same `hasKey`, so the card and the picker can no longer disagree about which rows are "not configured".
+- The chat credential guide — the one-time "no model credential configured" dialog — now asks `hasConfiguredKey` about the Project's default model instead of reading its masked key directly, so a Project whose default model runs on an environment variable is not nagged on first entry.
--- a/changelog/unreleased/2026-08-21-model-key-env-detection.md
+++ b/changelog/unreleased/2026-08-21-model-key-env-detection.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-21
 - **Type:** fix
 - **Scope:** `web`
+- **PR:** [#395](https://github.com/Prism-Shadow/penguin-harness/pull/395)
 
 [中文版](2026-08-21-model-key-env-detection.zh.md)
 
--- a/packages/web/src/features/chat/chat-page.tsx
+++ b/packages/web/src/features/chat/chat-page.tsx
@@ -106,7 +106,7 @@ import { buildInputHistory } from "./input-history";
 import { buildOutline } from "./outline-model";
 import { GoalStatusBanner } from "./goal-banner";
 import { handoffMessage, modelSwitchMessage } from "./agent-handoff";
-import { sameModelRef } from "../models/model-grouping";
+import { hasConfiguredKey, sameModelRef } from "../models/model-grouping";
 import { providerInfo } from "@prismshadow/penguin-core/model-catalog";
 import { FilesPanel } from "./files-panel";
 import { useFilesPanel } from "./use-files-panel";
@@ -998,6 +998,8 @@ export function ChatPage() {
   // gated by the server prefs' credentialGuideSeen — previously it checked "default model has no
   // key" and popped up a dialog on every visit to the chat page, which was repeated nagging for
   // users who simply don't intend to configure a key / use environment variables instead.
+  // "Has a key" is hasConfiguredKey, the same rule the model library and the model picker use: a
+  // model backed by an exported environment variable is configured and must not be nagged.
   useEffect(() => {
     if (!projectId) return;
     let cancelled = false;
@@ -1010,7 +1012,7 @@ export function ChatPage() {
         const { prefs } = await api.getPrefs();
         if (cancelled || prefs.credentialGuideSeen) return;
         const def = res.models.find((m) => sameModelRef(m, res.defaultModel));
-        const missing = !res.defaultModel || !def?.credential?.apiKeyMasked;
+        const missing = !res.defaultModel || !def || !hasConfiguredKey(def);
         if (missing) setCredentialGuide(true);
         // Mark as "seen" regardless of whether the dialog actually popped up: only ever once.
         void api.putPrefs({ credentialGuideSeen: true }).catch(() => undefined);
--- a/packages/web/src/features/chat/model-select.tsx
+++ b/packages/web/src/features/chat/model-select.tsx
@@ -156,14 +156,14 @@ export function PickerList<T>({
  * viewport no matter how many models there are.
  * Dropdown order mirrors the model library page (visibleChatModels): a top quick-search box
  * (the model page's rule — filters by id / display name / provider name); by default only
- * models with a configured API key are listed (stored masked key — the same standard as the
- * model page's key status; `envKey` is merely the NAME of a fallback env var and doesn't
- * count), with the selected and the default model always visible even without a key; a muted
- * bottom row reveals the remaining key-less models (marked by a struck-through key icon, with
- * the "no key" text in its title) without closing the menu or changing the selection — when
- * no model has a key at all, everything is listed directly. Rows carry the provider logo, the
- * light-yellow "Free" badge for zero-cost models (same as the model library card), the
- * project-default marker, and the selected checkmark.
+ * models with an API key are listed (hasConfiguredKey — a stored masked key or a masked env
+ * fallback, the same standard as the model page's key status; `envKey` is merely the NAME of a
+ * fallback env var and doesn't count on its own), with the selected and the default model
+ * always visible even without a key; a muted bottom row reveals the remaining key-less models
+ * (marked by a struck-through key icon, with the "no key" text in its title) without closing
+ * the menu or changing the selection — when no model has a key at all, everything is listed
+ * directly. Rows carry the provider logo, the light-yellow "Free" badge for zero-cost models
+ * (same as the model library card), the project-default marker, and the selected checkmark.
  */
 export function ModelMenuList({
   models,
--- a/packages/web/src/features/models/model-grouping.ts
+++ b/packages/web/src/features/models/model-grouping.ts
@@ -99,19 +99,29 @@ export function orderModelsLikeLibrary<T extends ModelRowLike>(rows: T[]): T[] {
   return groupModelRows(rows, "").flatMap((g) => g.rows);
 }
 
-/** Row shape for the configured-key filter: adds the read-only credential display (the DTO's ModelInfo is a superset). */
+/**
+ * Row shape for the configured-key filter: adds the read-only credential display and the masked
+ * env-fallback preview (the DTO's ModelInfo is a superset).
+ */
 export interface ModelCredentialRowLike extends ModelRowLike {
   credential?: { apiKeyMasked?: string };
+  /**
+   * Masked preview of the env-fallback value: the server emits it only for a variable that
+   * currently holds a non-empty value, so its presence is proof the environment can authenticate
+   * this entry.
+   */
+  envKeyMasked?: string;
 }
 
 /**
- * Whether the model has an API key configured: judged solely by the stored (masked) key — the
- * same standard as the chat credential guide and the model page's key status. `envKey` is only
- * the NAME of a fallback environment variable; its presence says nothing about whether that
- * variable is actually set, so it never counts as configured.
+ * Whether the model has an API key behind it — the single rule shared by the model library, the
+ * chat model picker and the chat credential guide: a stored (masked) key **or** a masked env
+ * fallback, since a user who exported the variable has configured the key just as deliberately as
+ * one who typed it into the dialog. `envKey` is only the NAME of that variable and says nothing
+ * about whether it is set, so it never counts on its own.
  */
 export function hasConfiguredKey(m: ModelCredentialRowLike): boolean {
-  return !!m.credential?.apiKeyMasked;
+  return !!m.credential?.apiKeyMasked || !!m.envKeyMasked;
 }
 
 /**
@@ -143,10 +153,10 @@ export interface VisibleChatModelsOptions {
 }
 
 /**
- * Candidate list for the chat model dropdown: library order → keep only key-configured models
- * (plus the selected and the default model, unless showAll) → the query then filters whatever
- * is visible. When NO model has a configured key, the key filter degrades to showAll
- * (everything listed), so the dropdown is never uselessly empty.
+ * Candidate list for the chat model dropdown: library order → keep only models with a key
+ * (hasConfiguredKey: stored or env-backed; plus the selected and the default model, unless
+ * showAll) → the query then filters whatever is visible. When NO model has a key, the filter
+ * degrades to showAll (everything listed), so the dropdown is never uselessly empty.
  */
 export function visibleChatModels<T extends ModelCredentialRowLike>(
   models: T[],
--- a/packages/web/src/features/models/models-page.tsx
+++ b/packages/web/src/features/models/models-page.tsx
@@ -75,7 +75,13 @@ import {
   resolveModelEnv,
 } from "@prismshadow/penguin-core/model-catalog";
 import type { FastModeProtocol, ModelProviderInfo } from "@prismshadow/penguin-core/model-catalog";
-import { groupModelRows, isFreeModel, sameModelRef, userProviderInfo } from "./model-grouping";
+import {
+  groupModelRows,
+  hasConfiguredKey,
+  isFreeModel,
+  sameModelRef,
+  userProviderInfo,
+} from "./model-grouping";
 import { protocolPathForModel } from "./protocol-path";
 import { ProtocolSuffixMenu } from "./protocol-suffix";
 import {
@@ -321,11 +327,31 @@ function isPreset(row: RowState): boolean {
   );
 }
 
-/** Whether this row already has (or will have, after this edit) an API key configured. */
-function hasKey(row: RowState): boolean {
-  return (
-    !row.clearApiKey && (Boolean(row.apiKeyInput.trim()) || Boolean(row.credential?.apiKeyMasked))
-  );
+/**
+ * Whether this row already has (or will have, after this edit) an API key: the shared
+ * hasConfiguredKey rule — a stored key or an env fallback the server proved is set — plus the two
+ * edit-only notions the DTO shape cannot carry. A key typed into the dialog counts before it is
+ * saved, and `clearApiKey` drops the **stored** key only: an environment variable cannot be
+ * cleared from here, so an env-backed row keeps its key through a clear.
+ */
+export function hasKey(row: RowState): boolean {
+  if (row.clearApiKey) return hasConfiguredKey({ ...row, credential: undefined });
+  return row.apiKeyInput.trim() !== "" || hasConfiguredKey(row);
+}
+
+/**
+ * The model card's key status line, most specific first: a stored key (not being cleared) shows
+ * its mask; a key typed but not yet saved has no mask of its own and just reads as configured;
+ * otherwise a detected first-party env fallback shows that variable's value under the same mask
+ * (the server reports envKeyMasked for official vendor entries only), so an env-backed row reads
+ * like a configured one. hasKey guards the whole ladder, so the card and the chat model picker can
+ * never disagree about which rows count as "not configured".
+ */
+export function keyStatusText(row: RowState): string {
+  if (!hasKey(row)) return S.models.noKey;
+  if (row.credential?.apiKeyMasked && !row.clearApiKey) return row.credential.apiKeyMasked;
+  if (row.apiKeyInput.trim() !== "") return S.models.keyConfigured;
+  return row.envKeyMasked ?? S.models.keyConfigured;
 }
 
 /** DTO -> row edit state (exported for unit tests): provider and modelId are both entry fields, never decomposed. */
@@ -1416,15 +1442,9 @@ function ModelCard({
     priced
       ? `${displayPrice(row.cacheRead, currency)} / ${displayPrice(row.cacheWrite, currency)} / ${displayPrice(row.output, currency)}`
       : null,
-    // Key status: the mask when configured; with no stored key, a detected first-party env
-    // fallback shows that variable's value under the same mask (the server only reports
-    // envKeyMasked for official vendor entries), so the row reads like a configured one;
-    // otherwise "not configured".
-    row.credential?.apiKeyMasked && !row.clearApiKey
-      ? row.credential.apiKeyMasked
-      : hasKey(row)
-        ? S.models.keyConfigured
-        : (row.envKeyMasked ?? S.models.noKey),
+    // Key status (see keyStatusText): the stored mask, a detected env fallback's mask, a plain
+    // "configured" for a key typed but not yet saved, or "not configured".
+    keyStatusText(row),
   ].filter((v): v is string => v !== null);
 
   const speedBadges =
--- a/packages/web/test/models-input.test.ts
+++ b/packages/web/test/models-input.test.ts
@@ -11,6 +11,9 @@
  *   model_id changes) moves the pointer along.
  * - Which env fallback variables a key hint may promise (detectedEnvKeys):
  *   only those the server reported a value for.
+ * - Whether a row has a key (hasKey) and what the card prints for it
+ *   (keyStatusText): the shared hasConfiguredKey rule (stored key or a masked
+ *   env fallback) plus the dialog's unsaved-edit notions.
  */
 import { describe, expect, it } from "vitest";
 import {
@@ -20,10 +23,13 @@ import {
   detectedEnvKeys,
   digitsOnly,
   fastModeState,
+  hasKey,
+  keyStatusText,
   nextPointers,
   rowRef,
   toRow,
 } from "../src/features/models/models-page";
+import { S } from "../src/lib/strings";
 
 describe("digitsOnly (context window)", () => {
   it("keeps digits only", () => {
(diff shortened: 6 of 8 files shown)


packages/web/src/features/models/model-grouping.ts
```diff
@@ -143,10 +153,10 @@ export interface VisibleChatModelsOptions {
...
 
 /**
- * Candidate list for the chat model dropdown: library order → keep only key-configured models
- * (plus the selected and the default model, unless showAll) → the query then filters whatever
- * is visible. When NO model has a configured key, the key filter degrades to showAll
- * (everything listed), so the dropdown is never uselessly empty.
+ * Candidate list for the chat model dropdown: library order → keep only models with a key
+ * (hasConfiguredKey: stored or env-backed; plus the selected and the default model, unless
+ * showAll) → the query then filters whatever is visible. When NO model has a key, the filter
+ * degrades to showAll (everything listed), so the dropdown is never uselessly empty.
```
hiyouga, line 159: **Regression for env-only users: this widening disarms the `!ordered.some(hasConfiguredKey)` escape hatch on line 167.**

The server emits `envKeyMasked` only when `envFallbackFirstParty` passes (`project-config-service.ts:115-133`): it is `false` for the `custom` group, every gateway group, every user-defined group, and any catalog preset whose `client_type` or `base_url` differs from the catalog's. Those entries still authenticate at runtime — `GenerativeModel`'s constructor omits `apiKey` and lets `AutoLLMClient` read the protocol's own variable (`packages/core/src/llm/generative-model.ts:1055-1066`).

Concretely, for a user with **no stored keys at all**, `ANTHROPIC_API_KEY` + `OPENAI_API_KEY` exported, and rows `anthropic/claude-sonnet-4-6` (catalog shape), `custom/my-vllm` (base_url set), `anthropic/…` re-pointed at a proxy:

```
BEFORE PR  default list: all three            hiddenCount 0
AFTER  PR  default list: claude-sonnet-4-6     hiddenCount 2
```

Before, `!ordered.some(hasConfiguredKey)` was `true` (nobody had `apiKeyMasked`) and the dropdown listed everything flat. Now the first-party row satisfies the predicate, the degrade stops firing, and the models the user actually works with are pushed behind "show models without key" and stamped with the struck-through key icon. That is the same defect this PR fixes, relocated onto a different audience.

The deeper fix: feed the predicate the process-wide fact the page already computes — `detectedEnvKeys(rows)` (`models-page.tsx:391`) — e.g. `hasConfiguredKey(m, envSet)` or a list-aware `keyedModels(list)`. "Is `$VAR` set" is a process fact, not a row fact, and every caller already holds the whole list.

packages/web/src/features/models/model-grouping.ts
```diff
@@ -99,19 +99,29 @@ export function orderModelsLikeLibrary<T extends ModelRowLike>(rows: T[]): T[] {
...
- * variable is actually set, so it never counts as configured.
+ * Whether the model has an API key behind it — the single rule shared by the model library, the
+ * chat model picker and the chat credential guide: a stored (masked) key **or** a masked env
+ * fallback, since a user who exported the variable has configured the key just as deliberately as
+ * one who typed it into the dialog. `envKey` is only the NAME of that variable and says nothing
+ * about whether it is set, so it never counts on its own.
  */
 export function hasConfiguredKey(m: ModelCredentialRowLike): boolean {
-  return !!m.credential?.apiKeyMasked;
+  return !!m.credential?.apiKeyMasked || !!m.envKeyMasked;
```
hiyouga, line 124: **`envKeyMasked` is a display-policy-gated field, not an auth fact — so the "single rule" is only correct for unmodified catalog presets.**

`envFallbackFirstParty` (`project-config-service.ts:115-133`) exists to stop the preview from *encouraging a misconfiguration* (pointing `OPENAI_API_KEY` at a reseller). It is a presentation rule with presentation reasons to change. Welding the auth predicate to it means:

- `openrouter/…`, `fireworks/…`, `siliconflow/…`, `qwen-*` (gateway groups), everything in `custom`, every user-defined group from the bulk import, and any preset re-pointed via `base_url`/`client_type` → `hasConfiguredKey` is `false` even when the exported variable authenticates them fine. `packages/server/test/models.test.ts:171-178` pins exactly those to `envKeyMasked === undefined` while still reporting `envKey`.
- Consequence on three surfaces at once: card prints "No key", `model-select.tsx:213` stamps the struck-through icon and hides the row, and if it is the Project default the credential guide nags — for a model that works.

Also worth naming: a future tweak to the *preview* rule (hide it when a stored key exists, gate it behind the owner-only credential role) would now silently change which models the picker lists, with no web-side test able to catch it.

packages/web/src/features/models/model-grouping.ts
```diff
@@ -99,19 +99,29 @@ export function orderModelsLikeLibrary<T extends ModelRowLike>(rows: T[]): T[] {
...
+ * env-fallback preview (the DTO's ModelInfo is a superset).
+ */
 export interface ModelCredentialRowLike extends ModelRowLike {
   credential?: { apiKeyMasked?: string };
+  /**
+   * Masked preview of the env-fallback value: the server emits it only for a variable that
+   * currently holds a non-empty value, so its presence is proof the environment can authenticate
+   * this entry.
+   */
+  envKeyMasked?: string;
```
hiyouga, line 113: **A whitespace-only variable now counts as a configured key.**

The server's guard is `envValue !== ""` (`project-config-service.ts:818`) — it does not trim — and `maskApiKey` returns `"***"` for anything ≤ 12 chars (`:94`). So `export ANTHROPIC_API_KEY=" "` (a stray `.env` line, a trailing-newline paste, a botched `docker -e`) yields `envKeyMasked: "***"`.

Pre-PR that only mis-painted the card's status line. Now it drives the picker filter (row listed by default, no struck-through icon) **and** permanently suppresses the one-shot credential guide, while every request on that model 401s.

If presence is to mean what the docstring above claims, the server guard should be `envValue.trim() !== ""`.

packages/web/src/features/chat/chat-page.tsx
```diff
@@ -1010,7 +1012,7 @@ export function ChatPage() {
         const { prefs } = await api.getPrefs();
         if (cancelled || prefs.credentialGuideSeen) return;
         const def = res.models.find((m) => sameModelRef(m, res.defaultModel));
-        const missing = !res.defaultModel || !def?.credential?.apiKeyMasked;
+        const missing = !res.defaultModel || !def || !hasConfiguredKey(def);
```
hiyouga, line 1015: **A transient env var now permanently burns the once-per-lifetime credential guide.**

`envKeyMasked` is a snapshot of `process.env[envKey]` at DTO-build time (`project-config-service.ts:816`). Widening the gate here suppresses the dialog, but line 1018 still writes `credentialGuideSeen: true` **unconditionally**.

So: first-ever chat visit on a box where the server has `ANTHROPIC_API_KEY` exported → no dialog, flag burned. The user then unsets the variable, restarts without the desktop's login-shell import, or opens the same account from a deployment that lacks it — the default model is genuinely key-less, the guide is permanently marked seen, and the only feedback is a send-time auth failure. Same outcome if the exported key is simply stale/revoked: `envKeyMasked` proves the variable is non-empty, not that it authenticates.

Worth considering: only write `credentialGuideSeen` when the answer was actually "configured", or re-evaluate the guide when a later `getModels` shows the default lost its key.

Minor, same line: `!res.defaultModel ||` is now fully subsumed by the `!def ||` term you added — `sameModelRef` returns `!!a && !!b && …`, so a falsy `res.defaultModel` already forces `def === undefined`.

packages/web/src/features/models/models-page.tsx
```diff
@@ -321,11 +327,31 @@ function isPreset(row: RowState): boolean {
...
-/** Whether this row already has (or will have, after this edit) an API key configured. */
-function hasKey(row: RowState): boolean {
-  return (
-    !row.clearApiKey && (Boolean(row.apiKeyInput.trim()) || Boolean(row.credential?.apiKeyMasked))
-  );
+/**
+ * Whether this row already has (or will have, after this edit) an API key: the shared
+ * hasConfiguredKey rule — a stored key or an env fallback the server proved is set — plus the two
+ * edit-only notions the DTO shape cannot carry. A key typed into the dialog counts before it is
+ * saved, and `clearApiKey` drops the **stored** key only: an environment variable cannot be
```
hiyouga, line 334: **The dialog never tells the user the env var takes over — so "an env-backed row keeps its key through a clear" is invisible exactly where the user makes that decision.**

The two masked-key blocks were left on the raw stored-mask rule:

- `:2340` — `form.credential?.apiKeyMasked && !form.apiKeyInput` — still renders the **stored** mask (with the checkbox ticked) while `clearApiKey` is set.
- `:2365` — `!form.credential?.apiKeyMasked && form.envKeyMasked !== undefined && !form.apiKeyInput` — suppresses the env line entirely, because the stored mask is still present.

Walkthrough: Anthropic preset, stored key `sk-o…1111`, `ANTHROPIC_API_KEY` exported (`envKeyMasked: sk-a…3456`). Tick "Clear stored API key" → the dialog reads as "this model will have no key". Save → `hasKey` is `true` via the env branch, the card flips to `sk-a…3456`, and the picker keeps listing it. **The user believes they revoked the model's access and they did not.**

The gate wants `clearApiKey` folded in, e.g. `(!form.credential?.apiKeyMasked || form.clearApiKey) && form.envKeyMasked !== undefined && !form.apiKeyInput`.

packages/web/src/features/models/models-page.tsx
```diff
@@ -321,11 +327,31 @@ function isPreset(row: RowState): boolean {
...
+  if (row.clearApiKey) return hasConfiguredKey({ ...row, credential: undefined });
+  return row.apiKeyInput.trim() !== "" || hasConfiguredKey(row);
+}
+
+/**
+ * The model card's key status line, most specific first: a stored key (not being cleared) shows
+ * its mask; a key typed but not yet saved has no mask of its own and just reads as configured;
+ * otherwise a detected first-party env fallback shows that variable's value under the same mask
+ * (the server reports envKeyMasked for official vendor entries only), so an env-backed row reads
+ * like a configured one. hasKey guards the whole ladder, so the card and the chat model picker can
```
hiyouga, line 347: **"the card and the chat model picker can never disagree" is true, but the dialog was left off the rule — and it contradicts the card on the same row.**

`envHintKey` (`:2137`) judges env coverage from `detectedEnvKeys` — *process-wide*, contributed by any sibling row — plus `liveEnvKey` (`:2009`), which resolves from `model_id`/`client_type` and **never looks at `base_url`**. `keyStatusText`/`hasKey` judge it from the *per-row* `envKeyMasked`, which the server withholds for exactly the rows `liveEnvKey` still answers for.

Reachable today: `OPENAI_API_KEY` exported, one first-party OpenAI preset present (so `detectedEnvKeys` has it) plus `custom/my-model` with a base_url and no stored key. Open `custom/my-model` → the API key field's placeholder reads *"Leave empty to use the OPENAI_API_KEY env var"*. Close it → the card behind it prints "No key", and (new in this PR) the chat picker hides that row behind the expander with the struck-through icon. Same for an Anthropic preset re-pointed at a proxy.

The hint is the truthful one here. One of the two rules has to give — either lift `detectedEnvKeys` into the shared predicate, or narrow `envHintKey` to `envKeyMasked`.

packages/web/src/features/models/models-page.tsx
```diff
@@ -321,11 +327,31 @@ function isPreset(row: RowState): boolean {
...
-  );
+/**
+ * Whether this row already has (or will have, after this edit) an API key: the shared
+ * hasConfiguredKey rule — a stored key or an env fallback the server proved is set — plus the two
+ * edit-only notions the DTO shape cannot carry. A key typed into the dialog counts before it is
+ * saved, and `clearApiKey` drops the **stored** key only: an environment variable cannot be
+ * cleared from here, so an env-backed row keeps its key through a clear.
+ */
+export function hasKey(row: RowState): boolean {
+  if (row.clearApiKey) return hasConfiguredKey({ ...row, credential: undefined });
```
hiyouga, line 338: **The `clearApiKey` branch discards `apiKeyInput`, contradicting this function's own contract ("or will have, after this edit") and all three submit paths.**

`rowToEntry` (`:488-489`) emits `apiKey` and `clearApiKey` as two *independent* `if`s, so a row with both produces `{apiKey: 'sk-new', clearApiKey: true}`. Downstream the three consumers disagree three ways:

| path | outcome |
| --- | --- |
| `putModels` (`project-config-service.ts:946-953`) | `delete next.api_key` then `next.api_key = entry.apiKey` → **key survives** |
| `testModel` probe (`:551`, `:605`) | `req.clearApiKey ? undefined : req.apiKey` → **runs keyless** |
| `hasKey` (here) | **"No key"** |

Unreachable through the current dialog only because the input's `onChange` resets `clearApiKey` (`:2332`) and the checkbox renders only while `!form.apiKeyInput` (`:2340`) — i.e. the invariant is enforced by JSX two thousand lines away, not by the state shape. `hasKey` is now an exported, unit-tested entry point (and the new test at `models-input.test.ts:120` hand-builds exactly this kind of composite row), so nothing stops the next caller from hitting it.

Separately, `hasConfiguredKey({ ...row, credential: undefined })` is an object-allocating spelling of `!!row.envKeyMasked`: it copies all ~20 `RowState` fields to switch off the one half of the delegate it would have shared, and it breaks *silently* the day `hasConfiguredKey` grows a third source (an OAuth mask, a group-level key) — `credential: undefined` would keep suppressing exactly one of them. `if (row.clearApiKey) return !!row.envKeyMasked;` says the actual rule with no allocation. (It is also a `TS2375` the day `exactOptionalPropertyTypes` is turned on.)

packages/web/src/features/models/models-page.tsx
```diff
@@ -321,11 +327,31 @@ function isPreset(row: RowState): boolean {
...
+ * otherwise a detected first-party env fallback shows that variable's value under the same mask
+ * (the server reports envKeyMasked for official vendor entries only), so an env-backed row reads
+ * like a configured one. hasKey guards the whole ladder, so the card and the chat model picker can
+ * never disagree about which rows count as "not configured".
+ */
+export function keyStatusText(row: RowState): string {
+  if (!hasKey(row)) return S.models.noKey;
+  if (row.credential?.apiKeyMasked && !row.clearApiKey) return row.credential.apiKeyMasked;
+  if (row.apiKeyInput.trim() !== "") return S.models.keyConfigured;
+  return row.envKeyMasked ?? S.models.keyConfigured;
```
hiyouga, line 354: **The `?? S.models.keyConfigured` arm is unreachable dead code, and this line hides the one truth-table cell the refactor did not preserve.**

Reaching line 354 requires `hasKey(row)` true, no usable stored mask, and a blank input. In the `clearApiKey` branch `hasKey` reduces to `!!row.envKeyMasked`; in the other branch, with the input blank and no stored mask, `hasConfiguredKey(row)` also reduces to `!!row.envKeyMasked`. Either way `envKeyMasked` is a non-empty string when this line runs — I brute-forced all 36 `(storedMask × clearApiKey × typedInput × envKeyMasked)` combinations and the `??` arm fires in none of them. `return row.envKeyMasked ?? S.models.noKey;` would be honest and lets the `hasKey` guard on 351 go away entirely.

It also masks a latent `!!` vs `??` split: `hasConfiguredKey` rejects `envKeyMasked: ""` (`!!""` is false) while this `??` would pass `""` straight through — harmless only because the guard runs first and the server can't emit `""` today (`optStr` maps `""` to `undefined`, `envValue !== ""`, `maskApiKey` ≥ `"***"`).

And the one behavioural delta vs the old inline expression: `(clearApiKey && typed && envKeyMasked)` used to print the **env mask** and now returns `S.models.keyConfigured` (it hits line 353 first). The other 22 of the 24 server-producible states are byte-identical, and this one is unreachable from the dialog — but it is undocumented, and it is exactly the state the previous comment is about.

packages/web/src/features/models/model-grouping.ts
```diff
@@ -99,19 +99,29 @@ export function orderModelsLikeLibrary<T extends ModelRowLike>(rows: T[]): T[] {
...
-/** Row shape for the configured-key filter: adds the read-only credential display (the DTO's ModelInfo is a superset). */
+/**
+ * Row shape for the configured-key filter: adds the read-only credential display and the masked
+ * env-fallback preview (the DTO's ModelInfo is a superset).
+ */
 export interface ModelCredentialRowLike extends ModelRowLike {
   credential?: { apiKeyMasked?: string };
+  /**
+   * Masked preview of the env-fallback value: the server emits it only for a variable that
+   * currently holds a non-empty value, so its presence is proof the environment can authenticate
```
hiyouga, line 110: **This docstring drops the first-party condition and asserts the false converse — and contradicts the sibling docstring added in the same PR.**

The server's own DTO doc is explicit (`packages/server/src/api/types.ts:383-385`):

> Absent = the variable is unset or empty, **or the entry is not first-party**.

This comment keeps only the first half and then upgrades it to "its presence is proof the environment can authenticate this entry", which `hasConfiguredKey` reads as *absence ⟹ cannot authenticate*. Meanwhile `models-page.tsx:346`, added by the same PR, states the missing half correctly ("the server reports envKeyMasked for official vendor entries only"). Two new comments, two different claims about one field — a reader cannot tell which is the contract, and the optimistic one is the one sitting on the predicate.

Stale reasoning left behind is worse than none (this PR's own premise); a comment that says the rule is complete is what will stop the next reader from re-opening it.

packages/web/src/features/chat/model-select.tsx
```diff
@@ -156,14 +156,14 @@ export function PickerList<T>({
...
- * the "no key" text in its title) without closing the menu or changing the selection — when
- * no model has a key at all, everything is listed directly. Rows carry the provider logo, the
- * light-yellow "Free" badge for zero-cost models (same as the model library card), the
- * project-default marker, and the selected checkmark.
+ * models with an API key are listed (hasConfiguredKey — a stored masked key or a masked env
+ * fallback, the same standard as the model page's key status; `envKey` is merely the NAME of a
+ * fallback env var and doesn't count on its own), with the selected and the default model
+ * always visible even without a key; a muted bottom row reveals the remaining key-less models
+ * (marked by a struck-through key icon, with the "no key" text in its title) without closing
+ * the menu or changing the selection — when no model has a key at all, everything is listed
```
hiyouga, line 164: **Scope expansion worth confirming: one exported vendor variable now promotes that vendor's entire preset catalog into the picker's default list.**

`envFallbackFirstParty` gates on group / `base_url` / `client_type` *shape*, not on the individual model (`project-config-service.ts:815`), so the server emits `envKeyMasked` for **every** unmodified catalog entry of a vendor whose variable holds a value.

A user who ran "sync presets" and happens to have `OPENAI_API_KEY` exported for some unrelated tool goes from ~2 rows in the collapsed list (the models they actually keyed) to every OpenAI preset in the catalog, inside the `max-h-56` scroller, while `showModelsWithoutKey(n)` collapses correspondingly.

The changelog frames this as the intent and it does match what the library card already printed — flagging it so it is a decision rather than a side effect, since the collapsed picker was the one surface where "configured" also meant "the user chose this model".

packages/web/src/features/chat/model-select.tsx
```diff
@@ -156,14 +156,14 @@ export function PickerList<T>({
...
+ * directly. Rows carry the provider logo, the light-yellow "Free" badge for zero-cost models
+ * (same as the model library card), the project-default marker, and the selected checkmark.
```
hiyouga, line 166: **The whole `visibleChatModels` pipeline runs twice per render (`:183` and `:187`), unmemoized, so both re-run on every keystroke in the quick-search box.**

Each call is `orderModelsLikeLibrary` → `groupModelRows` (an O(providers × rows) nested filter, plus a second pass for user-defined groups, a `Set`, a sort and a `flatMap`) followed by `some(hasConfiguredKey)`, the keep-filter with two `sameModelRef` per row, then `filter(matchesQuery)` — and `matchesQuery` re-normalizes the query and does a linear `MODEL_PROVIDERS.find` for *every* row.

The second call exists only to produce a count, and the collapsed list is a strict subset of the show-all list (same order, same query filter):

```ts
const all = useMemo(
  () => visibleChatModels(models, { showAll: true, query, selected: value, defaultModel }),
  [models, query, value, defaultModel],
);
const visible = showAll
  ? all
  : all.filter((m) => hasConfiguredKey(m) || sameModelRef(m, value) || sameModelRef(m, defaultModel));
const hiddenCount = showAll ? 0 : all.length - visible.length;
```

Identical output, one pipeline instead of two. (Pre-existing shape, but this PR re-blesses it in the rewritten header comment and makes `hasConfiguredKey` the predicate all three reads share.)

packages/web/src/features/models/models-page.tsx
```diff
@@ -321,11 +327,31 @@ function isPreset(row: RowState): boolean {
...
+
+/**
+ * The model card's key status line, most specific first: a stored key (not being cleared) shows
+ * its mask; a key typed but not yet saved has no mask of its own and just reads as configured;
+ * otherwise a detected first-party env fallback shows that variable's value under the same mask
+ * (the server reports envKeyMasked for official vendor entries only), so an env-backed row reads
+ * like a configured one. hasKey guards the whole ladder, so the card and the chat model picker can
+ * never disagree about which rows count as "not configured".
+ */
+export function keyStatusText(row: RowState): string {
```
hiyouga, line 350: **Altitude: the ladder is split across two modules, and the half that could be shared is the half that is stuck here.**

`hasKey` is pure policy but lives in a 2500-line React page and — after this PR — has exactly **one** production caller, `keyStatusText` on the next line; it is exported so a unit test can reach it (the file's convention is to say so, cf. `toRow` on `:357`). `keyStatusText` then fuses that policy with localized `S.*` output, which is precisely what blocks the ladder from moving next to `hasConfiguredKey`: `model-grouping.ts` is dependency-free today and cannot host it without acquiring an i18n dependency.

The factoring that would hold: a pure `keySource(row): "stored" | "typed" | "env" | "none"` in `model-grouping.ts`, with the string mapping left at the render site. Then this card, `model-select.tsx:213` (which re-derives its struck-through icon from `!hasConfiguredKey(m)`), the dialog's two masked-key blocks (`:2340`, `:2365`) and `envHintKey` (`:2137`) all read one enum instead of each re-deciding — which is what the two "can never disagree" claims in this PR actually need to be true. As shipped, the next surface that asks "why does this row have no key" writes a fifth copy.

packages/web/src/features/models/models-page.tsx
```diff
@@ -1416,15 +1442,9 @@ function ModelCard({
...
-    // envKeyMasked for official vendor entries), so the row reads like a configured one;
-    // otherwise "not configured".
-    row.credential?.apiKeyMasked && !row.clearApiKey
-      ? row.credential.apiKeyMasked
-      : hasKey(row)
-        ? S.models.keyConfigured
-        : (row.envKeyMasked ?? S.models.noKey),
+    // Key status (see keyStatusText): the stored mask, a detected env fallback's mask, a plain
+    // "configured" for a key typed but not yet saved, or "not configured".
+    keyStatusText(row),
```
hiyouga, line 1447: **The card can print a key status the server never accepted, and can look healthy while a destructive flag is armed.**

`persist`'s catch does `setRows(nextRows)` (`:601`), so a failed save leaves the draft's `apiKeyInput` **and** `clearApiKey` on the table rows:

- Type a key (or apply a group key) and save; the PUT fails → `hasKey` sees `apiKeyInput` non-blank → this card reads "Key configured" for a key that was never stored. Reload flips it back to "No key".
- Tick "Clear stored API key" and save; the PUT fails → the row keeps `clearApiKey: true` with its credential mask intact. Nothing on the card signals the armed flag. The user later clicks "sync presets", deletes an unrelated model, or applies a group key to a different provider — every one of those calls `persist(rows…)`, `rowToEntry` (`:489`) re-emits `clearApiKey: true`, and the stored key is deleted server-side without a second confirmation.

Both are pre-existing (the old `hasKey` counted `apiKeyInput` too), but they are the concrete counterexamples to the "the card and the picker cannot drift apart" claim this PR adds three lines above, so they are worth either fixing or scoping out of that claim.

changelog/unreleased/2026-08-21-model-key-env-detection.md
```diff
@@ -0,0 +1,16 @@
...
+
+## Details
+
+- `hasConfiguredKey` (`model-grouping.ts`) became the single predicate behind every surface — the picker's default list, its struck-through key icon and the "show models without key" count all read it — and `ModelCredentialRowLike` gained `envKeyMasked`. `envKey` still counts for nothing: it is only the NAME of a fallback variable and says nothing about whether that variable holds a value, while the server emits `envKeyMasked` only for one that currently does.
+- The model page's own `hasKey` was rebuilt on top of that predicate and now adds only what the DTO shape cannot carry: a key typed into the dialog counts before it is saved, and `clearApiKey` drops the **stored** key only — an environment variable cannot be cleared from there, so an env-backed row keeps its key through a clear. The card's key status line moved into `keyStatusText`, guarded by the same `hasKey`, so the card and the picker can no longer disagree about which rows are "not configured".
```
hiyouga, line 15: **The entry records a guarantee the code does not provide.**

> the card and the picker can no longer disagree about which rows are "not configured"

That holds only for first-party catalog rows. For every gateway / `custom` / user-defined / re-pointed row the server withholds `envKeyMasked` (`envFallbackFirstParty`, `project-config-service.ts:115-133`), so both surfaces agree on the *wrong* answer — and the config dialog, which was left on the raw stored-mask rule, disagrees with both (`models-page.tsx:2137`, `:2340`, `:2365`). `changelog/README.md` is explicit that an entry records what the change did rather than the state of the codebase, precisely because claims like this drift; here it is inaccurate on the day it lands.

Process note, not a defect in the change itself: `CLAUDE.md` (「Changelog 逻辑」) says *"changelog 用单独的汇总 PR 提交，不塞进各功能 PR"*, and this branch carries the entry pair in the same commit as the code (`54d181a4`) plus a follow-up link commit. Every recent PR (#389–#393) does the same, so the written rule and the actual practice have diverged — worth settling as "fix the rule or fix the workflow" rather than treating it as a one-off.

packages/web/test/models-input.test.ts
```diff
@@ -72,6 +78,66 @@ describe("detectedEnvKeys (variables a key hint may promise)", () => {
...
+      hasKey({ ...envBacked, credential: { apiKeyMasked: "sk-a\u20269999" }, clearApiKey: true }),
+    ).toBe(true);
+  });
+
+  it("the status line shows the most specific source it has", () => {
+    expect(keyStatusText(stored)).toBe("sk-o\u20261111");
+    expect(keyStatusText(envBacked)).toBe("sk-a\u20263456");
+    expect(keyStatusText(bare)).toBe(S.models.noKey);
+    // A key typed but not yet saved has no mask of its own.
+    expect(keyStatusText({ ...bare, apiKeyInput: "sk-new" })).toBe(S.models.keyConfigured);
```
hiyouga, line 129: **`keyStatusText`'s typed-key branch (`models-page.tsx:353`) is not actually covered by this suite — the assertion below it passes with that branch deleted.**

The fixture is `bare`: no `credential`, no `envKeyMasked`. Delete line 353 outright and control falls to line 354, `row.envKeyMasked ?? S.models.keyConfigured` → `undefined ?? S.models.keyConfigured` → the identical string. I ran all five `keyStatusText` cases in this `describe` against that mutant:

```
stored                     -> survives
envBacked                  -> survives
bare                       -> survives
bare + typed (line 129)    -> survives
cleared env (line 132-137) -> survives
```

The branch is load-bearing for exactly one row — typed key **and** `envKeyMasked` present, where 353 returns "Key configured" and 354 would return the env mask — and no test in either new suite constructs it. That is the same cell whose behaviour this PR silently changed from the old inline expression (see the note on `models-page.tsx:354`), so the one state that needed pinning is the one left unpinned. `expect(keyStatusText({ ...envBacked, apiKeyInput: "sk-new" })).toBe(S.models.keyConfigured)` would pin the priority and kill the mutant.

Secondary: both string assertions here compare against `S.models.*` — the same mutable `let S` binding (`strings.ts`) the implementation reads — so they pin the *choice between two keys*, never any literal text, and hold under any dictionary. Fine as far as it goes, worth knowing they are not copy tests.

PR #396, opened 2026-08-21
Unify how settings panels mark required and optional fields

hiyouga:
Settings panels marked mandatory fields three different ways: a red `*`, the word "optional"
appended to the neighbouring label, or nothing at all. This applies one rule everywhere:
**a required field carries the red `*`, an optional field carries no mark, and no label or
placeholder writes the word "optional".**

"Required" is grounded in behaviour, not in how the label sounds — a field is required when the
submit path refuses it empty. Two fields turned out to be mandatory and unmarked; none turned out
to be marked without being mandatory.

Explanations that had been folded into a label move out of it, following the split
[#392](https://github.com/Prism-Shadow/penguin-harness/pull/392) set: semantics disclose behind the
circled "?", the shape a value must take stays visible as the field's hint.

## Survey

Every field of Project settings (incl. the command-policy rule editor), Agent settings
(Overview / Prompt / Runtime / Tools & MCP / Vault / Schedules / Skills / Memory), System settings
(General / Appearance / Account / Proxy / Upload limits / Users) and the models add/edit and
group-key dialogs — 90-odd fields. **10 changed**; everything else already agreed with the rule and
was left alone.

## Fields whose marker or explanation changed

| Surface | Field | Was | Now | Why |
| System settings → Upload limits | Max attachment size | no mark | `*` | empty is refused with `invalid_attachment_limit`; save blocked |
| System settings → Upload limits | Max total per message | no mark | `*` | same |
| Agent settings → Memory → Add | Content or source | no mark | `*` | both dialog actions are disabled while empty (the code comment already said "required") |
| Project settings → Security → rule editor | Name | placeholder only | label + `*` | Apply is disabled without it |
| Project settings → Security → rule editor | Pattern | placeholder only | label + `*` | Apply is disabled without it |
| Project settings → Security → rule editor | Description | placeholder "Description (optional)" | label "Description", no mark | optional |
| New Project | Display name | label "Display name (optional, defaults to the Project id)" | label "Display name" + hint "Leave empty to use the Project id as the name" | optional; the fallback is a hint |
| Agent settings → Schedules | End at | label "End at (optional)" | label "End at" | optional |
| Agent settings → Schedules | Workspace | label "Workspace (optional; a temporary workspace is created when empty)" | label "Workspace" + hint "Empty = a temporary workspace is created" | optional; the fallback is a hint |
| Models → add group / model dialogs | base URL ×2, upstream model id, API key | hand-rolled label + hand-placed `*` | `FieldLabel` | same marks, one spelling |

Required/behaviour mismatches found: the three `no mark → *` rows above. No field carried a mark it
had not earned.

## One asterisk

The mark had been re-typed by hand in three places, and the trace viewer's copy
(`text-red-500` with no `dark:` variant) rendered a brighter red than the rest. It is now a single
`RequiredMark` in `components/ui/field.tsx`. The models dialogs hand-rolled their label rows
because a link or a detect button shares the line; they now use `FieldLabel`, which also makes
those titles visible to `disclosure-anchor.test.ts`.

## Tests

`test/required-mark.test.ts` (new) pins all three halves of the rule:

- the mark renders only when `required`, and is `aria-hidden` with `aria-required` on the control;
- an optional field renders no mark, no red ink and no `aria-required` — absence is the signal;
- no `.tsx` outside `field.tsx` hand-rolls a red `*` (TypeScript-parser scan, names file:line),
  mirroring how `icon-scale.test.ts` fails on a second copy of the close cross;
- neither dictionary says "optional" outside a named list of three entries where the word belongs
  to an explanation (the `call_description` argument, the memory bridge's "finish it in the chat",
  the budget format error) — plus a check that each exception still says it, so the list cannot rot.

Verified it fails on the pre-change state: re-introducing the trace copy and `endAt: "End at
(optional)"` produced `features/traces/trace-event-row.tsx:260` and `schedule.endAt`.

`disclosure-anchor.test.ts` and `info-popover.test.ts` needed no extension: no new "?" is added,
and the rule editor's new labels wrap no buttons.

## Decisions to review

1. **The two fallback notes became visible hints rather than "?" panels.** The frontend skill
   names "leave empty for unlimited" as *formatting*, which stays on screen, and the repo already
   spells 15+ such notes that way (`agent.nameHint`, `mcpCwdHint`, `baseUrlHint`,
   `chatDefaultsWorkspaceHint` …). Putting these two behind a "?" would have split the Create
   Project dialog from Create agent, and Schedules' Workspace from the identical field in Project
   settings → Defaults. Say the word and they move behind a "?" — together with their siblings.
2. **Project settings → General "Display name" gets no asterisk**, although Save is disabled while
   it is empty. It is an inline settings row (a `<p>` title next to a control), not a form field;
   the asterisk convention lives on `FieldLabel`. Marking it would mean giving settings rows a
   required state, which is a bigger idea than this PR.
3. **`memory.editRequirementPlaceholder` keeps its "(optional — you can finish it in the chat)"
   tail.** That sentence describes the bridge workflow rather than marking a field, so it is on the
   test's named exception list instead of being rewritten.
4. `project.name` folded into `project.displayName`: with the parenthetical gone the two keys held
   the same string, and the comment explaining why they differed described nothing any more.

## Verification

Node v24.18.0. Deps built first (`--filter penguin-skills/-core/-server build`) because the web
typecheck reads their `dist` types; the rest is scoped to the diff. Playwright not run (no
selectors or flows moved).

| Command | Result |
| --- | --- |
| `pnpm --filter @prismshadow/penguin-web typecheck` | pass |
| `pnpm --filter @prismshadow/penguin-web test` | pass — 91 files, 1165 tests (was 90 / 1159) |
| `pnpm format` + `pnpm format:check` | pass, no files rewritten |

Also rendered `RuleEditor` to static markup to confirm the new labels, the two asterisks and the
`sm:w-40` / `flex-1` split that moved onto the field wrappers.


packages/web/src/features/traces/trace-event-row.tsx
```diff
@@ -256,7 +257,7 @@ function ParamsTable({ schema }: { schema: Record<string, unknown> }) {
           <tr key={name} className="align-top">
             <td className="whitespace-nowrap pr-3 font-mono text-gray-700 dark:text-gray-300">
               {name}
-              {required.has(name) && <span className="ml-0.5 text-red-500">*</span>}
+              {required.has(name) && <RequiredMark />}
```
hiyouga: **Routing this copy through `RequiredMark` deletes the only required-ness signal the tool-schema table had for assistive tech.**

`RequiredMark` is `aria-hidden`, and its docstring justifies that with "the controls set `aria-required` from the same flag, which is what a screen reader announces." That premise holds for `Field`/`Input`/`Textarea`/`Select`/`OptionMenu`. It does not hold here: `ParamsTable` is a plain `<table>` of JSON-Schema properties — a `<td>` is not a form control, there is no `aria-required` anywhere in the file, no `<th>`, no caption and no legend.

Before this PR the cell rendered `<span className="ml-0.5 text-red-500">*</span>` with no `aria-hidden`, so the row announced "command star". After it, required and optional parameters are indistinguishable in the accessibility tree, while the function's own doc still promises "name / type / required / description".

This is the one place in the diff where the unification is a regression rather than a cleanup. `RequiredMark` needs an escape hatch for non-control call sites — a prop that drops `aria-hidden`, or an `sr-only` "required" beside it.

packages/web/src/features/settings/uploads-section.tsx
```diff
@@ -130,6 +131,7 @@ export function UploadsSection() {
       />
       <Input
         label={S.settings.attachmentTotalMb}
+        required
```
hiyouga, line 143: **Both fields are now marked required, but the refusal the changelog cites as justification is rendered on the second field only — so emptying the *first* one reddens the *second*.**

`save()` guards on `maxMb.trim() === "" || totalMb.trim() === ""` (line 69-70), but `limitError` is spread onto this `<Input>` alone (line 146). Clear "Max attachment size (MB)" — the field this PR just gave a `*` — and press Save:

- the empty field gets no error border, no `aria-invalid`, no `aria-describedby`;
- the correctly-filled total field goes red and is announced invalid;
- and because `Field` renders `error ? <FieldError> : hint ? <FieldHint> : null`, the total field silently loses its `${min}–${max} MB` bounds hint, which is the one piece of text a user needs to fix a value.

The server's `invalid_attachment_limit` lands the same way, and its text ("…with the total no lower than the per-file limit") is explicitly about both fields. The entry says "both … are refused when empty, and neither was marked. Both now carry the `*`" — the marking became symmetric while the feedback did not.

packages/web/src/components/layout/project-dialogs.tsx
```diff
@@ -164,7 +164,8 @@ export function CreateProjectDialog({
           />
         )}
         <Input
-          label={S.project.name}
+          label={S.project.displayName}
```
hiyouga, line 167: **Merging `S.project.name` into `S.project.displayName` makes one unmarked label mean two opposite things.**

Here in New Project the display name is genuinely optional, and the old string carried the distinction in its wording ("Display name (optional, defaults to the Project id)"). Project settings → General now renders the *same* `S.project.displayName` for a field that is refused when empty — `disabled={nameBusy || !name.trim() || …}` (line 392) and `if (!next || …) return;` (line 344) — with no `*`.

Under this PR's rule ("the absence of the asterisk is what says optional") the settings field now positively asserts something false. And it cannot be fixed at the call site: it renders inside `SettingRow`, whose props are `{title, description, children}` (line 214) with no `required`, and its bare `<Input>` has no `label`, no `aria-label` and no `required` — so it has neither an accessible name nor `aria-required`.

The diff also removed the only record of the asymmetry (`/** The display-name field in Project settings (required there, unlike the create dialog's optional one). */`) and re-attached the replacement to `displayNameHint`, a key only this dialog reads.

Same class, outside the audited surfaces: "Rename chat" (`sidebar.tsx:2009`, Save gated on `!renameText.trim()`, `confirmRename` returns early on empty) and the Skills import source field (`skills-tab.tsx:390`, `disabled={trimmedSource === ""}`) are both required and both unmarked.

packages/web/test/required-mark.test.ts
```diff
@@ -0,0 +1,157 @@
...
+        said,
+        'A field is optional because it has no red "*", never because its label says so.',
+      ).toEqual([]);
+    });
+  }
+
+  it("keeps the prose exceptions real (guards the allow-list against rot)", () => {
+    // A stale exception would quietly re-open the hole it was cut for.
+    const said = new Set([...optionalWording(zh, "zh"), ...optionalWording(en, "en")]);
+    for (const path of PROSE_EXCEPTIONS) expect(said, `${path} no longer says it`).toContain(path);
```
hiyouga: **This guard pins the violating wording in place: removing "(optional" from `memory.editRequirementPlaceholder` now *fails* the test.**

`memory.editRequirementPlaceholder` is not prose. It is the `placeholder` of a labelled `<Textarea>` in the Edit-memory modal (`memory-tab.tsx:810-817`), reading "Describe the change (optional — you can finish it in the chat)" — precisely the shape the rule in `RequiredMark`'s docstring forbids ("no surface writes that word into a label or a placeholder"). Its exception comment says "A bridge to the chat, not a form"; it sits between two other labelled Textareas in the same dialog, and the Add-memory dialog next door — structurally identical, same "bridge to the chat" framing — is one this PR marked `required`.

So the one genuine violation left in the tree is the one the allow-list carries, and this line then makes the correct fix a red build. A guard that fails on the repair is worse than no guard.

Separately, the union is wrong even for its stated purpose: `optionalWording(zh, "zh")` returns only `["agent.callDescriptionHint"]` — the other two match in `en` alone. Unioning the locales means a zh string that goes stale can never be reported, because its en twin keeps the path alive. Key the exceptions by `(path, locale)`.

packages/web/test/required-mark.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+  return out;
+}
+
+/** The element's `className`, when it is a plain string literal (the only form worth reading here). */
+function classNameOf(open: ts.JsxOpeningElement | ts.JsxSelfClosingElement): string | null {
+  for (const attr of open.attributes.properties) {
+    if (!ts.isJsxAttribute(attr) || attr.name.getText() !== "className") continue;
+    const init = attr.initializer;
+    if (init !== undefined && ts.isStringLiteral(init)) return init.text;
```
hiyouga: **"Spelled in exactly one place" only sees string-literal classNames and JSX *text* children, so the most likely form of a fourth copy walks straight past it.**

`classNameOf` returns `null` unless `ts.isStringLiteral(init)`, and the child check reads only `ts.isJsxText`. Missed:

```tsx
<span className={`ml-0.5 text-red-500`}>*</span>        // template literal
<span className={dim ? "text-red-500" : ""}>*</span>    // conditional
<span className="text-red-500">{"*"}</span>            // expression child
```

`className={…}` is not exotic here — `packages/web/src` has ~1695 `className=` against ~288 `className={`, so roughly 15% of class attributes are already outside the scan's reach, and a conditional mark is exactly what a re-typed asterisk tends to look like. The scan would report `[]` while a copy that dropped its `dark:` ink ships — the very bug this PR fixed in `trace-event-row.tsx`.

Two smaller things in the same function: the `| ts.JsxSelfClosingElement` arm is unreachable (the sole call site is guarded by `ts.isJsxElement`, and a self-closing element has no children to equal `"*"`), and there is no positive control — `disclosure-anchor.test.ts:108` and `placeholders-parity.test.ts:31`, both cited as the model, each assert their extractor still finds a known positive. Here the only known positive is excluded by construction (`if (path === FIELD) continue`), so if the walk ever returns nothing the test passes silently.

packages/web/test/required-mark.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  return found.sort();
+}
+
+/**
+ * Strings in a dictionary that say "optional" out loud, as `path -> value`. Functions are
+ * skipped: an interpolating entry is prose by construction, never a field label.
+ */
+function optionalWording(dict: unknown, locale: "zh" | "en", path = ""): string[] {
+  if (typeof dict === "string") {
+    const said = locale === "zh" ? /可选/.test(dict) : /\boptional\b/i.test(dict);
```
hiyouga: **The zh half of the rule is materially weaker than the en half, and zh is the source dictionary (`export type Strings = typeof zh`).**

`/可选/` is one literal substring. The idiomatic Chinese ways to say the same thing — 可留空 / 选填 / 非必填 / 可不填 — all pass. This is not hypothetical: `memory.editRequirementPlaceholder` in zh is "描述要怎么改（可留空，跳转后在对话中补充）", which evades the scan entirely and never needed its allow-list entry; only the English twin did.

The zh dictionary also mixes English freely (`sessionSearch: "搜索标题或 Session id…"`), and the very string this PR fixed was `workspace: "Workspace（可选，留空自动创建临时工作区）"` — half English. A future `endAt: "结束时间 (optional)"` is invisible to `/可选/`. Running **both** regexes over **both** dictionaries costs nothing and closes it.

In the other direction `/可选/` is unanchored, so 「该模型不可选」 or 「可选择的工作区」 would be reported as violations of a rule they do not break — forcing a bogus allow-list entry that, per the rot-guard comment above, cannot then be policed. `\boptional\b` likewise misses "optionally".

Also: `optionalWording` skips function-valued entries as "prose by construction", but there are 145 of them in `zh` and some are short user-facing chrome (`goalRoundBanner: (round) => \`目标 · 第 ${round} 轮\``), so an interpolating label is exempt by accident. And `Object.entries` on the 37 arrays yields index paths (`agent.placeholders.3.1`) that rot the moment a row is inserted above them.

packages/web/src/lib/strings-en.ts
```diff
@@ -958,7 +958,8 @@ export const en: Strings = {
     sessionSearch: "Search title or Session id…",
     sessionNoMatch: "No matching Session",
     sessionEmpty: "This agent has no Sessions yet",
-    workspace: "Workspace (optional; a temporary workspace is created when empty)",
+    workspace: "Workspace",
+    workspaceHint: "Empty = a temporary workspace is created",
```
hiyouga: **A PR whose thesis is "one thing gets spelled one way" adds a third wording of a sentence the dictionary already holds twice — and the changelog asserts a match that does not exist.**

```
project.chatDefaultsWorkspaceHint  "Empty = temporary workspace"                       / 留空表示使用临时工作区
schedule.workspaceHint (new)       "Empty = a temporary workspace is created"          / 留空自动创建临时工作区
chat.workspaceHint                 "Leave empty for an auto-created temporary …"       / 留空自动创建临时工作区；指定时必须是服务器上已存在的目录
```

The entry says the fallback "became a visible hint, matching the identical field in Project settings → Defaults" (zh: 「保持一致」). That field renders `S.project.chatDefaultsWorkspaceHint` (`project-dialogs.tsx:781`) — different text in both languages. The new zh value is also a character-for-character prefix of `chat.workspaceHint`, so a `grep` on the English will not find all three.

Worse at the call site: `schedules-tab.tsx:685` renders this hint directly under a `WorkspaceSelect variant="form"` whose own panel renders `S.chat.workspaceHint` (`workspace-select.tsx:289-292`). Open the picker and the user reads two differently-worded statements of the same rule at once, one of which drops the "must already exist" half.

Render `S.project.chatDefaultsWorkspaceHint` and add no key.

changelog/unreleased/2026-08-21-settings-field-markers.md
```diff
@@ -0,0 +1,42 @@
...
+- **New Project**: the display name read "Display name (optional, defaults to the Project id)".
+  The label is now just the name, and the fallback is a hint below the field — the same way the
+  Create agent dialog has always spelled it.
+- **Schedules** (Agent settings): "End at (optional)" and "Workspace (optional; a temporary
+  workspace is created when empty)" lost the wording; the workspace fallback became a visible
+  hint, matching the identical field in Project settings → Defaults.
+
+## One asterisk
+
+The mark had been re-typed by hand in three places, one of which had dropped its dark-mode ink and
```
hiyouga: **Two factual errors in this paragraph, both repeated in the zh entry.**

**"three places"** — the diff removes four hand-rolled copies: `models-page.tsx` ×3 (add-group base URL, model-dialog model id, model-dialog base URL) plus `trace-event-row.tsx`. Since the next clause identifies the trace copy as the one that dropped its dark-mode ink, it is inside the count, so the number is off by one. (zh:34 「手写了三份」.)

**"rendered a brighter red"** — backwards. The trace copy was `text-red-500` with no `dark:` variant; the others were `text-red-500 dark:text-red-400`. Tailwind's default red scale is not overridden anywhere in `styles.css` (only a brand scale is added), and red-400 `#f87171` is *lighter* than red-500 `#ef4444`. So in light mode the trace copy was pixel-identical to the rest, and in dark mode it rendered **dimmer**, not brighter. (zh:34 「红得比其他副本更亮」.)

While here — the sibling entry `2026-08-21-settings-disclosure.md:19-20` (and `.zh.md:16-17`) still states "the rule editor keeps its visible \"Regular expression\" placeholder", which this PR deletes in favour of a label. Both files ship unreleased side by side describing contradictory states of the same widget, and a release-notes reader hits the stale one first. That sentence needs fixing in this PR whether or not the two entries are folded together.

packages/web/src/components/ui/field.tsx
```diff
@@ -109,7 +122,7 @@ export function Field({
   error?: ReactNode;
   /** id for the error text — the control passes the same id in aria-describedby. */
   errorId?: string;
-  /** Renders a red "*" after the label to mark the field as required. */
+  /** Renders a red "*" after the label. An optional field passes nothing — no counterpart mark, and no "optional" in the label. */
```
hiyouga: **`required` is missing from `Field`'s early return, so under the new contract the flag can read as its own opposite.**

`if (!label && !hint && !error && !info) return <>{children}</>;` (line 135) — `<Input required />` with no label renders the bare control: no mark, no type error, no lint, and nothing in the new test can see it (it only renders `Input` *with* a label).

That matters more now than before the PR, because "the absence of the asterisk is what says optional" is being promoted from a habit to a contract. Three live call sites hit it — `models-page.tsx:1345`, `:2031`, `:2397` all pass `required` / `required={baseUrlRequired}` to a label-less `<Input aria-label={…}>` — and they only look right because a hand-placed `<FieldLabel required>` redraws the mark beside them. Nothing binds the two halves, so deleting either leaves a field mis-signalled in one direction; `project-dialogs.tsx:128/133` has *already* drifted the other way (the `FieldLabel` draws the mark, the `Input` never gets `required`, so no `aria-required` fires).

Adding `hint` to any of those three — the natural next edit, since `hint` is where the disclosure convention wants format rules — makes `Field` stop early-returning and render a wrapping `<label>` with a hint, still no label, still no mark.

packages/web/src/components/layout/project-dialogs.tsx
```diff
@@ -877,39 +878,48 @@ export function RuleEditor({
...
+            // and the browser's scroll-on-focus keeps the form in view on a short viewport.
+            autoFocus
+            onChange={(e) => setName(e.target.value)}
+          />
+        </div>
+        <div className="min-w-0 flex-1">
+          <Input
+            size="sm"
+            className="font-mono"
+            label={S.project.commandPolicyRulePattern}
```
hiyouga, line 905: **The pattern field now advertises a guarantee its gate does not deliver: `required` is checked untrimmed here and trimmed for the name.**

`apply()` is `const n = name.trim(); if (!n || !pattern) return;` and the button is `disabled={!name.trim() || !pattern}`. Type a name and a single space into Regular expression: `!pattern` is false for `" "`, `new RegExp(" ")` compiles, and `apply()` commits `{ pattern: " " }` — a command-policy rule matching any command containing a space, i.e. effectively every command, installed silently as a deny rule.

The asymmetry predates the PR, but the new `*` is what turns it into a stated promise: this file's own comment four lines up now says "name and pattern are both mandatory (Apply stays disabled without either), and a red \"*\" is what says so". `pattern.trim()` in both places.

packages/web/src/components/layout/project-dialogs.tsx
```diff
@@ -877,39 +878,48 @@ export function RuleEditor({
...
+              if (e.key === "Enter") apply();
+            }}
+          />
+        </div>
       </div>
       <Input
         size="sm"
-        className="w-full"
-        placeholder={S.project.commandPolicyRuleDesc}
+        label={S.project.commandPolicyRuleDesc}
```
hiyouga, line 922: **Now that the pattern field is a plain labelled `Field`, `invalid` + a detached `FieldError` is the wrong pair — `error={err}` is the prop this wants.**

`Input`'s own docstring scopes `invalid` to the case it no longer is: "use this when the input has a custom wrapper (prefix glyph, unit suffix, etc.) and the caller places the error text outside that wrapper" (`input.tsx:36-42`). The pattern field has no custom wrapper any more. `error={err}` renders the message inside the field's block and wires `aria-describedby={errorId}`, which `invalid` deliberately does not.

As written: type `[a-` and press Enter — the red border is on the pattern input at the top while the red text renders *after* the Description field, and the restructure just pushed them one label-height further apart. A screen-reader user on the pattern input gets `aria-invalid` with no described-by text at all.

One more in this block, pre-existing but adjacent: the master `Switch` is `disabled={busy}` only (line 1017) and nothing resets `editing`, so the policy can be switched off with the editor open. The container comment at line 1027 states that every control below disables on `!enabled` "since `opacity` alone would leave every control focusable" — the three `<Input>`s and both buttons here take no `disabled` at all, so the form reads dead at 50% opacity while still being tab-reachable, editable and able to commit a rule.

packages/web/src/features/models/models-page.tsx
```diff
@@ -2291,9 +2284,7 @@ function ModelDialog({
             own <label> wrapper, so this outer container is a <div> (a nested <label> is invalid). */}
         <div className="block">
           <span className="mb-1 flex items-baseline justify-between gap-2">
-            <span className="text-xs font-semibold text-gray-600 dark:text-gray-400">
-              {S.models.apiKey}
-            </span>
+            <FieldLabel block={false}>{S.models.apiKey}</FieldLabel>
```
hiyouga, line 2287: **This label row passes no `htmlFor`, and the control below it gets neither `label` nor `aria-label` — the API key box has no accessible name at all.**

`FieldLabel` accepts `htmlFor` and renders a real `<label>` when given one; none of the four new rows in this file pass it. For the other three the wrapping `<label className="block">` still associates them, but this row is a `<div>`, so the visible "API Key" text is an orphan `<span>` and the `PasswordInput` at line 2299 announces only its placeholder (`apiKeyHint`, which is the masked existing key or a generic hint). Clicking the visible label does not focus the box either.

The sibling `AddGroupDialog` does it correctly — `<PasswordInput label={S.models.apiKey} …>` at line 1307 — so the same field is accessible in one dialog and not the other.

Related, at line 2011: that row's wrapping `<label>` pulls the "Get model ids ↗" link text into the input's computed accessible name. A `FieldLabel htmlFor` scopes the name to the label text and fixes both.

packages/web/src/features/models/models-page.tsx
```diff
@@ -1319,12 +1319,9 @@ function AddGroupDialog({
                 picker — the add-model dialog's idiom, so the two flows read as one. */}
             <div className="block">
               <span className="mb-1 flex items-baseline justify-between gap-2">
-                <span className="text-xs font-semibold text-gray-600 dark:text-gray-400">
+                <FieldLabel required block={false}>
```
hiyouga, line 1322: **The asterisk was de-duplicated; the row around it was not — and `required` is now spelled twice per field.**

Four sites in this file (1321, 2011, 2286, 2367) still hand-roll the identical wrapper:

```tsx
<div|label className="block">
  <span className="mb-1 flex items-baseline justify-between gap-2">
    <FieldLabel required={…} block={false}>{…}</FieldLabel>
    {aside}
```

and two of them wrap the aside in a further `<span className="flex shrink-0 items-baseline gap-2.5">` while two do not — the same drift this PR just finished cleaning up in the mark, one level out. All four were rewritten here, so this was the moment to lift the shape.

The more consequential half is that the flag is now duplicated: once on the `FieldLabel` for the mark, once on the control for `aria-required` (1322/1345, 2012/2031, 2368/2397). That is the pairing that actually rots, and `project-dialogs.tsx:128/133` has already drifted. `field.tsx:142-158` already builds exactly this two-part structure for its `info` slot, so a `labelAside`/`action` slot on `Field` (or a local `LabelRow`) collapses all four to one-liners and takes the flag once.

The detect-protocol button is a second copy in the same hunks: lines 1325-1339 and 2373-2387 are verbatim identical but for `disabled` and the handler name, including a hand-drawn spinner `<span>` that has no home in `components/ui/icons.tsx` — which `test/icon-scale.test.ts` already polices for exactly this.

packages/web/test/required-mark.test.ts
```diff
@@ -0,0 +1,157 @@
...
+import { fileURLToPath } from "node:url";
+import ts from "typescript";
+import { Input } from "../src/components/ui/input";
+import { zh } from "../src/lib/strings";
+import { en } from "../src/lib/strings-en";
+
+const SRC = fileURLToPath(new URL("../src", import.meta.url));
+const FIELD = join(SRC, "components", "ui", "field.tsx");
+
+function tsxFiles(dir = SRC, out: string[] = []): string[] {
```
hiyouga, line 28: **Copy #4 of the src-tree walker, and it makes this one of the slowest files in the web suite.**

`SRC` + `tsxFiles()` is byte-identical to `test/disclosure-anchor.test.ts:25-32`; `icon-scale.test.ts:17-24` (`sources`) and `context-menu.test.ts:237-253` (`sourceFiles`) are variants #2 and #3, and the `ts.createSourceFile(…, /* setParentNodes */ true, ts.ScriptKind.TSX)` + `visit`/`forEachChild` block below is copy #3 of that (see also `command-policy-add-rule.test.ts:29-47`). The copies have already drifted (`.tsx` only vs `.ts`+`.tsx`, `sources` vs `tsxFiles`), and each has to re-learn the Windows `replaceAll(sep, "/")` invariant that `context-menu.test.ts` documents in a four-line comment about a green-on-Linux/red-on-Windows failure.

Cost: `disclosure-anchor.test.ts` is already the slowest file in the 90-file web suite at ~794ms against ~135ms for the next; this repeats the same 132-file walk and parse, so suite test-body time goes ~2.65s → ~3.4s. Two cheap wins if the scan stays: `setParentNodes: true` costs ~35% of the parse and is needed only by the single `attr.name.getText()` on line 41 (`attr.name.escapedText` is a drop-in — `node.getStart(source)` works without parent pointers), and the visitor reads the expensive `classNameOf` *before* the free children-text check that eliminates 2191 of 2192 candidates.

`readdirSync(dir, { withFileTypes: true })` would also drop 250 `statSync` calls per walk and let the recursion skip symlinked directories — it currently follows them with no cycle guard, and this repo's own layout is symlink-based.

A `test/source-tree.ts` (non-suite modules already live in `test/`, and `vitest.config.ts` collects only `test/**/*.test.ts`) exporting the walk, the parse and the location formatter would replace ~30 duplicated lines in three files.

packages/web/test/required-mark.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    expect(required).toMatch(/<span[^>]*aria-hidden[^>]*>\*<\/span>/);
+    expect(required).toContain('aria-required="true"');
+  });
+
+  it("leaves an optional field completely unmarked", () => {
+    // No counterpart mark and no wording: absence is the signal, so anything here would blunt it.
+    const optional = renderToStaticMarkup(
+      createElement(Input, { label: "Name", value: "", readOnly: true }),
+    );
+    expect(optional).not.toContain("*");
```
hiyouga: **`not.toContain("*")` scans the whole rendered markup including `class`, so an unrelated styling change fails this test with a message blaming the required mark.**

The assertion is not scoped to the label — it covers `<label class="block"><input class="w-full rounded-md … " …/></label>`. This package is on `tailwindcss ^4.1.0`, where `*:` (direct children) and `**:` (all descendants) are first-class variants. The moment anyone adds a `*:`-style utility or an arbitrary value containing `*` to `controlBase`, `baseClass`, `sizeClass`, `FieldLabel` or `FieldHint`, this fails as "leaves an optional field completely unmarked" and sends the next maintainer hunting a nonexistent asterisk regression. `not.toContain("text-red-")` has the same coupling to any red utility that reaches the Input chain.

The positive case three lines up already does it correctly — `expect(required).toMatch(/<span[^>]*aria-hidden[^>]*>\*<\/span>/)`. Assert the negative against the same shape.

PR #397, opened 2026-08-21
feat(core,docs): the memory prompt names when a fact is worth saving

hiyouga:
The built-in `memory.prompt` listed *what* belongs in memory but never said *when* to notice it, leaving the judgement entirely to the model mid-task. It now names the moments that produce a durable fact — and the whole block came out shorter than it went in.

## The prompt now

Four labelled lines in place of five dense paragraphs:

> **Save** who the user is and how they want you to work, with the why; goals and constraints not derivable from the code; pointers to external resources.
>
> **When:** a request repeated, a correction that outlives the task (with what they wanted instead), a habit or convention stated more than once, a reusable working detail you would otherwise ask for again. One clear statement is enough; a repeat only makes it obvious. Unsure? Ask.
>
> **Never** what code, config or git history states, unconfirmed guesses, transcript excerpts — asked anyway, save the non-obvious part. Anything else is savable if the user wants it; anyone who reaches this agent reads it all.
>
> **Index:** each `MEMORY.md` lists its memories one line each (`- [Title](file.md) — hook`, ~150 chars, no content), updated with the file — deletions included. Only the first N lines are injected: merge overlaps, drop stale entries. Before saving, check the index and extend the file that covers the subject; delete memories that prove wrong.

Three deliberate choices in the **When** line:

- **A repeat is a signal, not a threshold.** Requiring one would lose a standing preference stated clearly once, which is the most common way a user states one.
- **A correction saves what was wanted instead, not the complaint.** That is the reusable half.
- **Asking is the fallback, not the default.** It resolves the unsure case without turning every candidate fact into a question.

## The prohibitions were narrowed

`task progress`, `credentials` and `personal data` are no longer barred. They are things a user sometimes does want kept, and a blanket refusal made the Agent argue with a direct request. What stays barred is what memory cannot usefully hold: facts the code, config or Git history already states, unconfirmed guesses, transcript excerpts.

The visibility sentence stays, deliberately **as a fact rather than a rule** — anyone who can reach the Agent reads all of it, and it travels with export/import and snapshots. That is what someone needs to know before asking for something sensitive to be written down; it informs the decision instead of making it for them.

## Length

| | characters |
| --- | --- |
| `main` | 1739 |
| this branch | **1735** |

The new guidance is carried at no net cost. The savings are in the prose, not the content: the labelled lines lost their restatements, the index line lost a clause the rule above it already implies, and the lead and frontmatter comment lost words that were carrying nothing.

## Kernel generation

`memory.prompt` is a kernel-managed config leaf, so the pinned-hash guard fails until the history records the new value. The `2026-08-21` generation's hash is **revised in place** rather than a new generation appended: that generation was introduced by #376 and has not shipped (v0.2.3 was tagged 2026-08-19), so no config in the wild carries the stamp, and the module's own rule allows revising the newest entry while it is unfrozen.

The five older generations keep the previous hash `9380e238…`, which is the load-bearing part: `applyKernelUpdate` hashes an Agent's stored `memory.prompt`, still finds it among the recorded generations, and advances it to the new default. An Agent whose prompt was hand-edited hashes to nothing recorded, is conservatively treated as a customization, and keeps it — such an Agent does not receive the new guidance unless its owner restores the Memory tab's default.

## Docs

The Configuration guide's Memory chapter described what is worth saving but likewise not when; one sentence covering the same ground was added to both `configuration.en.md` and `configuration.zh.md`.

## Verification

Node v24.18.0.

| Command | Result |
| --- | --- |
| `pnpm --filter @prismshadow/penguin-core test` | pass — 44 files, 926 tests (5 skipped) |
| `pnpm --filter @prismshadow/penguin-docs test` | pass — 7 files, 53 tests |
| `pnpm format` + `pnpm format:check` | clean |

The pinned-hash guard was confirmed to work in both directions: before each history update it failed naming exactly `[ 'memory.prompt' ]` as the drifted leaf, and passed after.


packages/core/src/state/default-config.ts
```diff
@@ -241,6 +241,8 @@ updated_at: <YYYY-MM-DD>
 
 Worth saving: who the user is (role, expertise, preferences) and how they want you to work, with the why; ongoing work, goals and constraints not derivable from the code; pointers to external resources.
 
+Watch for the moments that produce them: the user asks for the same thing again, or is dissatisfied with what you produced — save what they wanted instead, and why; states a habit, convention or development practice more than once; or hands you reusable details you would otherwise ask for again (names, addresses, identifiers, endpoints — anything that reads like filling in a form). A repeat is a strong signal, not a requirement: one clear statement of a standing preference is enough. When you cannot tell whether something is worth keeping, ask.
```
hiyouga: **The new paragraph instructs the model to save PII and credentials that the same prompt forbids two paragraphs later.**

Added here:

> hands you reusable details you would otherwise ask for again (**names, addresses, identifiers, endpoints** — anything that reads like filling in a form)

Standing rule at line 246, unchanged, in the same prompt:

> Never save … **secrets** … Memory is readable by everyone who can reach this agent: **no sensitive personal data**.

And `packages/docs/content/configuration.en.md:250`: "every Project member who can reach the Agent can read it — so credentials and sensitive personal data never belong in it."

Names and postal addresses are the textbook definition of sensitive personal data; "endpoints" routinely *is* the credential (`postgres://user:pass@host/db`, a Slack webhook URL, a presigned S3 URL). The new instruction is affirmative, concrete and example-laden and arrives ~1,300 characters *before* the prohibition, which is a five-word trailing clause on a paragraph about `MEMORY.md` line lengths — specificity wins.

**Failure scenario:** the user pastes a shipping address or a staging DB connection string. The Agent writes `agent_state/memory/user/<slug>.md`; that file's index line is injected into the system prompt of *every* future Session, is readable by every Project member who can reach the Agent, and travels with export/import and snapshots (a whole scope exports as one JSON document). This also defeats the 0600 Vault the same assembled prompt spends two sections protecting — `{{VAULT}}` renders four sections *above* `{{MEMORY}}` and says "values never appear in your context", while `:430` says "Never read, copy or print … `.vault.toml` — they hold the user's secrets". The prompt now names two destinations for the same provisioning input and gives no rule for choosing the safer one.

Suggested repair: drop `names, addresses, identifiers, endpoints`, or narrow to non-personal reusable details (project identifiers, service names, conventions) with an explicit carve-out.

hiyouga: **"anything that reads like filling in a form" is a *shape* test with no source anchor, so tool output and file contents qualify — third-party PII, not just the user's own.**

Every other clause in the paragraph is anchored to the user ("the user asks", "what **you** produced", "**hands you**"). This trailing generalization drops the subject entirely and tests appearance, not provenance. The "Never save" list at line 246 excludes only "what code, config or git history already states" — which does not cover an API response, a CSV, a support ticket, a seed fixture, or a CRM export.

**Failure scenario:** the Agent runs `curl` against a staging endpoint and gets `{"name":"Jane Doe","address":"…","customer_id":"…"}`, or reads `fixtures/customers.csv`. That reads exactly like filling in a form, is "reusable", and is not stated by code/config/git — so under this clause it is written into a Markdown file that every Project member can read and that rides export/import and snapshots.

This is distinct from the category contradiction flagged separately: even if `names, addresses` were removed, an unanchored shape test still admits third-party data the user never volunteered. Repair: re-anchor the clause to the user ("details **the user gives you** that you would otherwise ask for again"), which is what the rest of the sentence already does.

packages/core/src/state/kernel-history.ts
```diff
@@ -345,7 +345,7 @@ export const KERNEL_HASH_HISTORY: Readonly<Record<string, Readonly<Record<string
     "compaction.mode": "e58fac0b4b9c0f29b3d224da119dff5f6517a40d139faf92e309705b98bd410a",
     "compaction.prompt": "8ed55781e8071083f8246ed70d9e063ee729d3872d5bc6a2bedfa9c99aed6ae5",
     "memory.enabled": "b5bea41b6c623f7c09f1bf24dcae58ebab3c0cdd90ad966bc43a45b44867e12b",
-    "memory.prompt": "9380e2382e5dd3d37a6470b0bc0d27c9ed113fa9a54ff78122784c57a16b992c",
+    "memory.prompt": "9fa69c31c4e35123f80994429133c361d7ee8dba2011d3624303ba0925e4c445",
```
hiyouga: **Revising this hash in place without bumping `KERNEL_VERSION` permanently strands every config already stamped `2026-08-21`.**

`KERNEL_VERSION` stays `"2026-08-21"` (line 29), and the update is offered on a *stamp* comparison, not a hash comparison:

```ts
return typeof kernelVersion !== "string" || kernelVersion < KERNEL_VERSION;   // :123
```

so `isKernelOutdated("2026-08-21")` is `false`. The Web App hard-gates on it:

```tsx
<Button size="sm" disabled={kernelUpdating || !data.config.kernelOutdated} …>  // agent-settings-page.tsx:503
```

and `agents-page.tsx:264` hides the update icon on the same flag.

**Failure scenario:** the `2026-08-21` generation landed on `main` in `3f8d1e8f` (#376). Any Agent created — or default-restored, or kernel-updated — by a build off `main` since then has `kernel_version: 2026-08-21` on disk *with the old prompt*. After this merges it reports "up to date", the Update kernel button is greyed out, and the new guidance can never arrive. `applyKernelUpdate` itself is purely hash-driven and would advance it correctly — the *affordance* is what is missing, and it does not self-heal; it stays broken until some future `KERNEL_VERSION` bump.

The PR body's justification ("no config in the wild carries the stamp") is true for releases — v0.2.3 was tagged 2026-08-19 and does not contain #376 — but not for anyone tracking `main`: every dogfood install, contributor worktree, and CI/e2e-seeded agent state. The changelog's claim at `.md:26-28` ("recognized as such and advanced by the kernel update") holds only for stamps ≤ `2026-08-20`.

Cheapest fix: bump `KERNEL_VERSION` and append a generation rather than revising this one. The date-keyed scheme blocks a same-day append, so a sortable suffix (`2026-08-21.2`) would keep `isKernelOutdated`'s lexicographic compare working unchanged.

changelog/unreleased/2026-08-21-memory-save-signals.md
```diff
@@ -0,0 +1,31 @@
...
+preference still qualifies; when the Agent cannot tell whether something is worth keeping, it is
+told to ask in the conversation.
+
+## Details
+
+- The prompt is a kernel-managed config leaf. The `2026-08-21` generation's `memory.prompt` hash was
+  revised in place (that generation is unreleased); the five frozen older generations keep the
+  previous hash, so an Agent whose stored prompt is still an old default is recognized as such and
+  advanced by the kernel update. An Agent whose prompt was edited by hand keeps that edit and does
+  not receive the new guidance — restoring the Memory tab's default adopts it.
```
hiyouga: **"restoring the Memory tab's default" names a control that does not exist — and the one whose name matches is destructive.**

`packages/web/src/features/agents/memory-tab.tsx` has no restore/reset affordance at all: two `<Textarea>`s and a Save button (`:700-757`). Grepping that file and `prompt-injection-controls.tsx` for `restore|reset|default` returns only unrelated prose comments about import modes.

The only restore in the product is **Restore default configuration** in the Kernel section (`agent-settings-page.tsx:508-516`, `variant="danger"`), whose own confirm copy reads (`strings-en.ts:556`):

> This overwrites the agent's existing configuration with the current defaults: the custom system prompt, tool list, model/compaction settings and MCP servers are all replaced, keeping only name and description. … this cannot be undone.

**Failure scenario:** a user who tweaked one sentence of their memory prompt follows this line, finds nothing on the Memory tab, uses the control whose name matches, and loses their custom system prompt, tool list, model/compaction settings and every configured MCP server — to adopt one paragraph.

Clearing the textarea is not an escape hatch either: `agent-config-service.ts:361` writes `""` verbatim and `agent-state.ts:351` (`config?.prompt ?? DEFAULT_MEMORY_PROMPT`) treats `""` as a real value by design, so the Agent then runs with an *empty* memory prompt.

Either correct the sentence to name the whole-config restore and state its blast radius, or build the per-field restore it describes. Same text in `.zh.md:20` ("在记忆标签页恢复默认即可采用").

packages/core/src/state/default-config.ts
hiyouga: **The "dissatisfied" trigger fires on `[user_steering]` and `[turn_aborted]` — the harness's own mid-run correction markers, which the same prompt frames as strictly task-local.**

174 lines below, `# System markers` teaches the model exactly what those two are:

> - `[turn_aborted]`: the previous round was interrupted…
> - `[user_steering]`: a user message sent mid-run, delivered between turns. **Not a new task: incorporate it immediately and adjust course within the current one.**

Those are the two highest-frequency dissatisfaction signals in the product, and both are explicitly scoped to the current task. Nothing in the new paragraph excludes them, and line 246 separately says "Never save … **task progress**".

**Failure scenario:** the user hits stop and steers "no, not that directory — use `src/`". The model reads the steering block as dissatisfaction, writes `prefers-src-directory.md`, and — in a temporary Workspace, where User is the only scope — every future Session of that Agent in *every* project now carries a one-off, project-specific correction as a standing preference. The prompt's own "delete memories that prove wrong" never fires, because the memory is not wrong so much as wrongly generalized.

Repair: exclude in-turn steering explicitly, or require that the preference be stated as a general rule rather than inferred from a mid-run course correction.

hiyouga: **The new triggers are user-scope-shaped, but they were added to `memory.prompt` while the scope tie-break lives in `memory.workspace_prompt` and resolves to Workspace.**

Line 265 (`DEFAULT_MEMORY_WORKSPACE_PROMPT`, untouched by this diff — its hash is unchanged):

> What would still hold in a different project goes in user memory; **when unsure, write here.**

The new triggers are dominated by cross-project, user-level facts — "one clear statement of a **standing preference**", "a habit, convention or development practice", "names, addresses" — yet they are stated in the always-injected half with no scope pointer, so in a persistent Workspace they inherit a tie-break that resolves to *Workspace*. (`memorySection` appends the workspace half directly onto the main prompt, so the model reads both in one block.)

**Failure scenario:** in a persistent Workspace the user says "always pnpm, never npm". The model is unsure which scope, follows "when unsure, write here", and files it under `memory/<workspace_key>/`. Every other project the Agent works in never sees it — precisely the failure the User scope exists to prevent, and the failure the new sentence's own words ("standing preference") presume it is avoiding.

Repair: attach a scope hint to the new triggers ("a standing preference belongs in user memory"), or widen the tie-break in `workspace_prompt` to except them.

hiyouga: **"When you cannot tell whether something is worth keeping, ask." is unbounded, and it reaches Sessions that have no user to answer.**

The sentence names no addressee, no timing and no frequency bound, and its trigger — "cannot tell whether something is worth keeping" — is the common case, which is why the paragraph was written at all. Meanwhile `# Stop rules` (lines 407-410) budgets interruption for exactly three reasons (success criteria met, an ambiguous *request*, an unresolvable blocker) and `# Personality` (`:396`) says "Communicate with the user precisely and concisely." Memory bookkeeping is none of those.

The same `{{MEMORY}}` block reaches every Session type — each leg checked:

- **Subagents** — `agent.ts:696` calls `childAgent.createSession(...)`, and `CreateSessionOptions` (`agent.ts:97-131`) has no memory-suppressing field; `agent.ts:279` resolves memory gated only on the Agent-level toggle, and `subagentDepth` never reaches `assembleSystemPrompt`. So the child carries this sentence with **no human channel at all** — its output is a `run_subagent` tool result.
- **Scheduled tasks** — `scheduler.ts:390` → `sessionService` → the same `agent.createSession(...)`, sender `"server"`.
- **Goal mode** — `goal-prompts.ts:105`: "Never use `blocked` merely because the work is hard, slow, or would benefit from clarification."

Scoping this honestly: the sentence says "ask", not "stop and ask", so a question folded into a turn's final reply is structurally cheap; goal mode survives it (`goal-prompts.ts:76-78` — ending a turn does not end the goal) and a scheduled session is answerable later in the Web App. **The subagent leg is the one with no recipient.**

**Failure scenario:** a subagent spends its final report asking whether to remember something and the parent must re-dispatch; in ordinary sessions, turns acquire "Before I continue — should I remember that you prefer `pnpm`?", which is the chatter `:396` forbids. A clarifying round-trip costs a full request (system prompt + tools + history), orders of magnitude more than the paragraph's own token cost.

Repair: bound it — "ask at the end of the turn, never mid-task, and never in a subagent or scheduled run" — or fold it into line 246, which already resolves the same uncertainty in the opposite direction.

packages/docs/content/configuration.en.md
```diff
@@ -276,7 +276,7 @@ updated_at: 2026-08-07
...
+These three fields are all the frontmatter there is — which layer a memory belongs to is expressed by its directory, so there is no `type` field (a `type:` line left in an earlier file is ignored as an unknown field). Worth saving: who the user is (role, expertise, standing preferences) and how they want the Agent to work, with the why; decisions, constraints and plans not derivable from the code; stable entry points into external systems, documents and services. The prompt also names the moments that produce such a fact — a request made a second time, dissatisfaction with a result, a habit or development practice stated more than once, or reusable details handed over that the Agent would otherwise ask for again — and tells it to ask in the conversation when it cannot tell whether something is worth keeping. A topic that turns out to be wrong is deleted together with its index line, and dates are written absolute (`YYYY-MM-DD`) — relative ones mean nothing to a later Session. What must never be saved: facts the code, config or Git history already states; short-lived task progress and debugging notes; credentials, tokens or secrets; unconfirmed guesses; long transcript excerpts.
```
hiyouga: **The docs sentence inverts the rule it describes: it presents repetition as the criterion and drops the qualifier that says it is not.**

The prompt says:

> A repeat is a strong signal, **not a requirement**: one clear statement of a standing preference is enough.

The docs render the trigger as "a habit or development practice **stated more than once**" and omit the qualifier entirely, in **both** languages. A reader concludes repetition is required — the opposite of what ships, and of what the PR description calls a deliberate choice ("A repeat is a signal, not a threshold"). The changelog gets it right ("A repeat is called a strong signal rather than a requirement"), so the PR's three artifacts disagree with each other.

Three further clause-level deltas against `default-config.ts:244`:

| prompt | docs |
| --- | --- |
| `states a habit, convention or development practice more than once` | drops *convention* (both changelog files keep it) |
| `— save what they wanted instead, and why` | absent; dissatisfaction is reduced to a noticing moment, hiding that it is a *write* instruction |
| `(names, addresses, identifiers, endpoints — anything that reads like filling in a form)` | absent |

The last omission is the consequential one: that clause is exactly what collides with line 250 of this same page ("credentials and sensitive personal data never belong in it"), 29 lines above. Dropping it means a docs reader can never see the conflict — and the changelog's claim that the docs "gained one sentence covering the same ground" (`.md:30-31`) is not accurate.

**Failure scenario:** an operator audits Memory against the Configuration guide, concludes "no PII, only abstract reusable details, and only repeated statements get saved", approves it for a team Project, and later finds customer names and internal endpoints in an exported user scope.

Same omissions in `configuration.zh.md:279`.

packages/core/src/state/default-config.ts
hiyouga: **The repetition-based triggers cannot survive compaction, and `compaction.prompt` was left untouched.**

"asks for the same thing again" and "stated **more than once**" require cross-turn observation of the raw transcript — which compaction destroys. `DEFAULT_COMPACTION_PROMPT` (`default-config.ts:558-563`, hash unchanged in this diff) enumerates what to carry:

> include everything needed to continue the task — the original request, current state, next steps, and any learnings

Stated preferences and dissatisfaction moments are named nowhere in that list, and `:420` tells the model the summary "is its only record — treat it as established context and continue from it."

**Failure scenario:** the user says "integration tests must hit a real DB" at turn 4; compaction fires at turn 40; at turn 55 they say it again. The model sees a *first* statement, not a repeat, and the "strong signal" never fires — so the trigger silently stops working in exactly the long sessions where a standing preference is most likely to surface twice. Symmetrically, a pending "should I remember this?" question asked just before compaction vanishes along with its answer.

This is one more reason the "a repeat is not a requirement" qualifier is load-bearing, and an argument for adding stated preferences to the compaction prompt's carry list in the same batch.

hiyouga: **Garden-path sentence: the em-dash imperative strands the subject of the two triggers this PR exists to add.**

> the user asks for the same thing again, or is dissatisfied with what you produced **— save what they wanted instead, and why;** states a habit, convention or development practice more than once; or hands you reusable details…

Subject trace:

| clause | subject |
| --- | --- |
| `the user asks for the same thing again` | `the user` |
| `or is dissatisfied with what you produced` | `the user` (gapped, adjacent) — ok |
| `— save what they wanted instead, and why` | **imperative, addressed to the model — voice switch mid-sentence** |
| `; states a habit … more than once` | **none** |
| `; or hands you reusable details…` | **none** |

`states` is 3sg present and needs a 3sg subject; the nearest NPs leftward are `why` and `what they wanted instead`, both inside the imperative aside and both nonsense. The real subject is 16 words back, behind *two* barriers that each say "new clause starts here" — a voice change and a top-level semicolon (which in English marks an *independent* clause and so cancels the subject-gapping the sentence depends on). Mixed coordination compounds it: two `or`s and two semicolons, with nothing to say whether `states a habit…` is a third alternative of the first `or` or a new list level.

**Failure scenario:** the model parses `states a habit … more than once` as a fragment and drops it, or attaches it to the imperative and reads it as an instruction that *it* should state a habit — inverting two of the three signals. Silent, per-turn, and no test catches it.

The PR already wrote the fix twice: the changelog and the docs both use subject-free noun phrases that need no gapping ("a habit, convention or development practice stated more than once"; "reusable details handed over that the Agent would otherwise have to ask for again"), matching the house pattern of line 242.

packages/core/src/state/default-config.ts
```diff
@@ -241,6 +241,8 @@ updated_at: <YYYY-MM-DD>
...
+
 Each directory's \`MEMORY.md\` is its index, injected below: one line per memory, under ~150 characters (\`- [Title](file.md) — hook\`), no content, updated in the same round as the file — deletions included. Only the first ${MEMORY_INDEX_MAX_LINES} lines of an index are injected — keep it well under that: merge overlapping entries, drop stale ones, move detail into the topic files. Before saving, check the index and update the file that already covers the subject instead of duplicating; delete memories that prove wrong. Never save what code, config or git history already states, task progress, secrets, unconfirmed guesses, or transcript excerpts — if asked to, save the non-obvious part instead. Memory is readable by everyone who can reach this agent: no sensitive personal data.
```
hiyouga: **Four new save triggers were added against this line's fixed index budget, with no counterweight.**

This line caps the injected index at `MEMORY_INDEX_MAX_LINES` (200) and tells the model to "keep it well under that"; line 230 says "**One file per fact**". The diff broadens what qualifies — four triggers plus an explicit floor-lowering ("one clear statement … is enough") — without touching the cap, the ~150-character line budget, or the consolidation guidance.

The form-filling trigger is the worst offender because its examples are individually atomic: a name, an address, two service endpoints and three identifiers are seven facts, hence seven files and seven index lines.

**Failure scenario:** a few setup-heavy sessions push the User index past 200 lines. Injection then truncates (`MEMORY_INDEX_MAX_CHARS` = 25,000 is the second backstop) and the model silently stops seeing its oldest memories — the recall failure the cap exists to make visible, now reachable through ordinary use rather than through a runaway file.

If the triggers stay as written, the consolidation guidance on this line probably needs strengthening in the same batch (e.g. group form-fill details into one topic file rather than one per field).

packages/core/src/state/kernel-history.ts
hiyouga: **The `2026-08-21` generation's own doc comment (lines 151-154, just above this table) was not updated and now under-reports the generation.**

It still reads:

> - `"2026-08-21"` (current) — background execution: `exec_command` / `run_subagent` gained `run_in_background`, `input_command`'s empty-poll default became 120000ms, and the `kill_command` / `kill_subagent` tools joined the set (**five `tools.builtin.*` leaves changed or appeared**).

After this PR that generation also carries a changed `memory.prompt` — a sixth leaf, and the only non-tool one. The diff edits the hash table without touching the prose that explains it.

This comment is the *only* human-readable record of what each generation means: there is no other doc, `design/specs/` never mentions the kernel, and there is no `KERNEL_GENERATIONS` list.

**Failure scenario:** a maintainer debugging why a kernel update reported `advanced: ["memory.prompt"]` reads "five `tools.builtin.*` leaves" and concludes the prompt was never in scope for this generation; the next person doing a same-day in-place revision has no idea `memory.prompt` was already revised here and edits the wrong generation — which no test would catch, since the pinned-hash guard compares only the newest entry.

hiyouga: **Latent (not triggered by this diff): the header's permission to revise the newest generation in place is missing the qualification that actually makes it safe.**

Line 12: "Several changes on the same day may revise that day's entry in place instead of appending (the entry is only frozen once a later generation exists)." Line 128: "only the newest entry may still be revised, on a same-day change." Both unqualified.

In-place revision is only safe for a leaf whose **pre-revision value survives in an older generation**. That is why *this* PR is fine — `memory.prompt`'s old hash `9380e238…` is recorded in all five older generations (lines 174, 204, 240, 276, 312), so a stored old default still hash-matches and advances. It is **not** true of the leaves introduced *by* the `2026-08-21` generation: `tools.builtin.kill_command` appears only at line 363 and `tools.builtin.kill_subagent` only at line 369, in this block alone.

**Failure scenario:** an Agent materialized from the `2026-08-21` defaults, then a same-day in-place revision of one of those two leaves. `historicalHashes(path)` (`kernel-update.ts:95-102`) would then contain only the post-revision hash, `storedHash !== defaultHash`, and the leaf falls into `else { kept.push(path); }` (`kernel-update.ts:176-177`) — permanently reported as a *user customization* the user never made, never advanced by any future update, with destructive restore-defaults the only recourse.

Cheap fix, and this PR is the natural place for it: one sentence in the header — *a leaf may be revised in place only if its previous hash is also recorded in an older generation; otherwise append a generation.* Nothing mechanical enforces the freeze rule today (the pinned-hash test compares only the newest entry, and no test asserts older generations are byte-stable), so the comment is the whole defence.

packages/docs/content/configuration.en.md
hiyouga: **The sentence asserts as universal fact something true only of the newest built-in default, breaking the chapter's own convention.**

"**The prompt** also names the moments … and tells it to ask" is phrased as a property of every Agent's Memory prompt. It is false for any Agent whose `memory.prompt` was hand-edited — this PR's own changelog says so — and for any Agent that has not run a kernel update (and, per the separate comment on `kernel-history.ts:348`, for any Agent stamped `2026-08-21`, which cannot run one).

The chapter is otherwise careful about exactly this distinction: line 285 says "**The default Memory prompt** declares the line cap".

Placement compounds it: the sentence was inserted into the paragraph documenting *frontmatter and policy*, taking it to 201 words in one unbroken block and splitting the what-to-save / what-never-to-save pairing the paragraph is built around. The paragraph that actually inventories what `memory.prompt` contains is line 283-285 ("what Memory is for, the save mechanics…"), and it was not updated.

**Failure scenario:** a user reads the guide, expects their existing Agent to ask before saving something borderline, and files a bug when it does not.

Suggested repair: "**The default prompt** also names…", moved next to the existing inventory sentence. Same wording issue in `configuration.zh.md:279`.

changelog/unreleased/2026-08-21-memory-save-signals.md
```diff
@@ -0,0 +1,31 @@
+# Memory prompt names when to save
+
+- **Date:** 2026-08-21
+- **Type:** change
```
hiyouga: **`Type: change` is not a valid value.**

`changelog/README.md:46`:

> | `Type` | always | **Exactly one of** `feature` (new user-facing capability), `fix` (defect correction), `refactor` (restructuring with no capability change), `process` (tooling, docs, skills, CI, release plumbing). |

Corpus tally across the whole `changelog/` tree: `feature` ×172, `fix` ×94, `process` ×50, `refactor` ×12 — and `change` ×2, **both of them this PR's `.md` and `.zh.md`**. (The two stray `feat` are pre-existing in the frozen `0.2.3/2026-08-15-session-fork` pair.)

**Cost:** the README deliberately ships no index file, so `Type` *is* the retrieval mechanism — `README.md:91` documents `grep -rl 'Type:\*\* fix' changelog/` and its siblings. This entry falls into none of the four buckets and is invisible to every Type-faceted query, permanently: once `unreleased/` is renamed at release the folder is frozen.

```suggestion
- **Type:** feature
```

`feature` matches the closest precedent — `changelog/unreleased/2026-08-20-compaction-threshold-default.md:4`, the same class of change (a kernel-managed config leaf default) with the identical `Scope: core, docs`. Same fix needed in `.zh.md:4`.

PR #405, opened 2026-08-23
fix(core): continue goal mode past a per-Task max_turns cutoff

### LINKED ISSUE #171: Goal mode aborts instead of continuing after a Task reaches max_turns

rootkiller6788:
Problem

  In Goal mode, when a Task reached the per-Task turn cap (max_turns), the entire goal ended as aborted instead of
  starting a new Task. This defeated goal mode for exactly the long-running tasks that most need continuation. Closes
  #171.

  Root cause

  The goal loop (packages/core/src/goal/goal-loop.ts) treated a final assistant text with stop_reason: "failed" as a
  terminal round. That signal conflates two different conditions that both end a round with a failed notice and no abort
  event:

  1. The engine's max_turns cutoff — a recoverable, per-Task bound that resets on the next Task.
  2. A non-recoverable failed finish such as an output-length truncation (finish_reason: "length").

  Change

  - Added isMaxTurnsCutoff(), which recognizes the engine's own [reached max turns (…); stopping] sentinel.
  - A round now records roundFailed only when the final assistant text is failed and not that sentinel. The max_turns
    cutoff therefore falls through as a round boundary: the loop starts the next round with a fresh turn budget, and the
    cut-off round's unsubmitted tool outputs (held as engine carry-over) are replayed into the next round, preserving
    workspace and progress.
  - Other failed final texts (e.g. output-length truncation) remain terminal, so their existing behavior is unchanged.

  The goal stays bounded by its token budget, the round cap (maxRounds, default 100), user aborts, and the model's own
  complete/blocked status.

  Docs updated in goal-mode.en.md / goal-mode.zh.md to describe the cutoff as a round boundary.

  Tests

  - packages/core/test/goal.test.ts: rewrote the prior "failed final text is terminal" case and added three — max_turns
    cutoff continues to the next round; a non-max-turns failed text stays terminal; a goal that hits max_turns every
    round still stops at maxRounds.
  - pnpm --filter @prismshadow/penguin-core typecheck — clean.
  - pnpm format:check — clean.
  - Full core suite: 925 passed; the 3 failures are unrelated Windows symlink EPERM errors in file-tools/memory
    (Developer Mode disabled).

  Changelog

  Bilingual entry added under changelog/unreleased/2026-08-23-goal-mode-max-turns-round-boundary.{md,zh.md}.


hiyouga:

rootkiller6788:
Hi @hiyouga — thanks for the quick response, and for the kind note.

To be transparent: I used an AI coding assistant to draft this change, but I reviewed the diff myself before submitting.
I fully understand the root problem: when a single Task hits max_turns, the entire Goal gets aborted rather than proceeding into the next round.
I have manually verified this fix locally: goal.test.ts passes, typecheck and format:check are clean.

I’d really like to keep this contribution if you are open to it.

Thanks for maintaining the project and your time.

PR #419, opened 2026-08-23
fix(skills): keep non-text files out of the library manifest

Myriad-Dreamin:
## The defect

`readSkillFiles` collected every regular file in a library skill directory with `fs.readFileSync(path, "utf8")`, and the other end of that chain writes them back the same way:

- `packages/skills/src/index.ts` — auxiliary files become `Record<string, string>`
- `packages/core/src/state/agent-state.ts:459` — `installSkill` writes each with `fs.writeFile(file, data, "utf8")`

So a file that is not UTF-8 text gets its bytes replaced with U+FFFD on the way in and is written out corrupted, with nothing reporting it. Latent today — the library ships only Markdown and SVG — but the only thing keeping it latent is that nobody has committed a PNG next to a `SKILL.md`.

## The fix

`decodeSkillFile(bytes)` returns the text or `undefined`, and `readSkillFiles` skips a file it cannot decode:

- `fatal: true` — malformed UTF-8 is rejected instead of being replaced.
- NUL byte rejected — that is what UTF-16 text looks like once its bytes happen to form valid UTF-8 (`68 00 69 00`).
- `ignoreBOM: true` — a leading BOM stays in the string, so a file that *does* decode carries exactly what the old `"utf8"` read produced. Without this, `TextDecoder` would have silently stripped the BOM and changed installed content.

Skipping rather than throwing: a bad file can only arrive through a commit to `packages/skills/skills/`, and the test below catches that in CI. At runtime, an omitted auxiliary file makes the agent's read fail where it is used; throwing would 500 the whole library listing for every user instead.

## Why not carry binaries

The manifest being text-only is a property of the built-in library, not of Skills in general — the archive-install route already carries uploaded files as `Uint8Array` (`packages/server/src/http/routes/skills.ts:101`), so a Skill installed from a zip can ship binary assets. Making the library manifest byte-preserving would mean a new field on `LibrarySkill`, the `SkillMetadataItem` DTO and the install writer, for a capability nothing asks for. Not built.

## Guard

A new test walks `skills/` — `SKILL.md` and `icon.svg` included, not only the auxiliary files — and lists any file that is not text.

Both halves confirmed against a planted `vllm/reference/logo.png` (a real PNG header):

- test: `non-text files in the skill library: expected [ 'vllm/reference/logo.png' ] to deeply equal []`
- loader: `librarySkill('vllm').files` returned `["reference/notes.md"]` — the PNG dropped, its text sibling kept.

## Verification

`pnpm format:check`; `skills` test (28 passed, up from 24) and `typecheck`; `skills` build + `server` typecheck, since this adds an export to a package the server imports.

## Note

One of five PRs from an audit of the shipped skill library.

### DIFF
--- a/changelog/unreleased/2026-08-23-skill-library-loader-keeps-non-text-files-out.md
+++ b/changelog/unreleased/2026-08-23-skill-library-loader-keeps-non-text-files-out.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-23
 - **Type:** fix
 - **Scope:** `skills`
+- **PR:** [#419](https://github.com/Prism-Shadow/penguin-harness/pull/419)
 
 [中文版](2026-08-23-skill-library-loader-keeps-non-text-files-out.zh.md)
 
--- a/changelog/unreleased/2026-08-23-skill-library-loader-keeps-non-text-files-out.md
+++ b/changelog/unreleased/2026-08-23-skill-library-loader-keeps-non-text-files-out.md
@@ -1,0 +1,23 @@
+# The skill-library loader keeps non-text files out of the manifest
+
+- **Date:** 2026-08-23
+- **Type:** fix
+- **Scope:** `skills`
+
+[中文版](2026-08-23-skill-library-loader-keeps-non-text-files-out.zh.md)
+
+`readSkillFiles` collected every regular file in a library skill directory by reading it with the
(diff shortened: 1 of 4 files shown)


Myriad-Dreamin:
Closed pending a decision on the approach; the defect and both candidate fixes are recorded in #559. The branch stays for whichever is chosen.

PR #426, opened 2026-08-24
Uploads without a size limit, and compression for large images

Myriad-Dreamin:
## What this does

Two things, one commit each.

**1. The attachment size limits are deleted.** The per-file cap, the per-message byte total, the admin form that set them, the request body cap derived from them, and the fixed 20MB refusal on an inline image — all gone. Every one of those numbers refused an upload the transport could in fact have carried, and the ceiling was whatever an operator had typed at some earlier point. The failure it produced was the worst kind: a 413 *after* the tab had read the file, base64'd it, and uploaded 33% more bytes than the file weighs.

**2. Neither API caps request size any more.** No `/api/*` cap, no `/api/hmr` cap. The only thing that remains bounded is what a gzip upgrade body *inflates* to — the part of the original PR that was a real defect rather than a coupling problem: `zlib.gunzipSync(gz).toString("utf8")` had no bound at all, so a few hundred kilobytes of gzip decided how many gigabytes this process allocated, and the request only failed once `JSON.parse` met whatever had been materialized.

## The image is compressed, not refused

The one number in the old scheme that was paying for something real was the image ceiling. An inline image is not an attachment: an attachment goes to the Session scratchpad and the model opens it by path, so its size is paid once, but an image enters the conversation and is written verbatim into the Trace JSONL — read back whole, into one JS string, on every history page and every Session resume. A large inline image is a Session that gets slower every time it is opened.

So that cost is now paid by making the picture smaller instead of turning it away. **System settings → Uploads** (admin, server-global) carries a switch and a threshold: an image over the threshold is fitted inside a 2048px box and re-encoded in its own format, **in the browser, before it is read into a data URL**. The upload shrinks with the picture rather than after it — which a server-side pass could not have done.

- Default: **on, above 4MB**. Threshold settable 1–64MB.
- 4MB is just under the ~5MB a provider commonly accepts for one image, and core's `read_image` cap. So the images this catches are largely ones that were heading for a refusal further down anyway — compressing makes more images work, not fewer.
- **Animated and vector images are never touched** (GIF, SVG): a canvas round trip returns a first frame or a rasterization, which is a different picture, not a smaller one. Only `image/jpeg`, `image/png`, `image/webp` go through, and the output keeps the input's type so alpha survives and nothing downstream meets a new format.
- Every failure path returns the original file — no `createImageBitmap`, a decoder that refuses, a canvas that will not read back. A re-encode that comes out no smaller than the source is discarded.

## What bounds remain

| | Before | After |
| --- | --- | --- |
| Per-file attachment | 100MB, admin-settable 1–200 | none |
| Per-message total | 120MB, admin-settable 1–200 | none |
| Inline image | 20MB, `413 image_too_large` | compressed above the threshold |
| `/api/*` body | derived, ~199MB at the default | **none** |
| `/api/hmr` body | the `/api/*` cap | **none** |
| `/api/hmr` *inflated* size | unbounded | `buffer.constants.MAX_STRING_LENGTH` |
| Files per message | 20 | 20 (`413 too_many_files`) |

**No number in this PR is chosen any more.** Verified on Node v24.12.0:

```
MAX_STRING_LENGTH  536870888  = 512.0MB
Buffer.alloc(520MB).toString("utf8")
  -> ERR_STRING_TOO_LONG | Cannot create a string longer than 0x1fffffe8 characters
```

The body is decoded into one string before `JSON.parse` sees it, so ~512MB is the platform's ceiling whether or not anything guards it. `readJson` now separates `ERR_STRING_TOO_LONG` out of the parse error it used to collapse into and answers `413 payload_too_large` naming that ceiling — so removing the cap removed the refusal, not the explanation. Reported as "must be valid JSON" it would have sent someone hunting a syntax error in a body that has none.

The upgrade channel keeps `gunzipSync`'s `maxOutputLength`, set to the same constant. That is not a size policy: without it the process allocates until it dies and never reaches the string that would have thrown. At `MAX_STRING_LENGTH` it refuses exactly the payloads that could never have been parsed, and nothing else — a real push is single-digit megabytes.

The per-message file **count** stays, because it bounds a different thing: a request the transport carries happily still fits hundreds of thousands of minimal `file` parts, which is that many sequential writes into one directory and that many marker lines on one message.

`services/attachment-limits.ts` is deleted with its last constant; `MAX_ATTACHMENT_COUNT` moved to `task-attachments.ts`, where it is enforced.

The messaging bridge's inbound image ceiling moved to `bridge.ts` as `MESSAGING_INBOUND_IMAGE_MAX_BYTES`. It was sharing the composer's constant, but it bounds a *download* from a remote chat anyone in that chat can post to — never the same question as what a composer may attach.

## What this trades away

Stated plainly, since these are the point of the change rather than oversights:

- **The server no longer refuses a large inline image from a non-browser client.** The Web App compresses; an API client posting a 400MB data URL writes 400MB into the Trace and pays that cost on every resume.
- **An authenticated caller can make the server buffer up to the string ceiling per request**, and concurrent requests multiply it. The removed `/api/*` cap was what bounded that — it was originally added because a chunked request carries no `content-length`, so a header check alone passes a body of any size. Nothing replaces it.

Both are the cost of removing defensive settings that, in the overwhelmingly common case, were blocking a person rather than an attacker.

## API changes

- `GET /api/me` reports **`uploadPolicy`** in place of `uploadLimits`: `{attachmentMaxCount, imageCompression, imageCompressionOverMb, imageCompressionMinMb, imageCompressionMaxMb}`. It shapes what a client uploads and gates nothing.
- `PUT /api/admin/settings` takes **`imageCompression`** and **`imageCompressionOverMb`** in place of `attachmentMaxMb` / `attachmentTotalMb`. Out-of-range is `400` `invalid_image_compression` (was `invalid_attachment_limit`), validated before any write as before.
- `413` `file_too_large` and `image_too_large` are no longer raised for composer input. (`file_too_large` still exists for the Workspace file endpoints, which have their own unrelated limits.) `payload_too_large` now means only "larger than this API can decode".

**On disk:** the `attachment_max_mb` and `attachment_total_mb` rows in `server_settings` are simply never read again. No migration, no compatibility code, nothing to remove later — an absent row already meant "the default", so a stale one costs nothing. Flagging it rather than deciding quietly: if you would rather they were deleted on boot, say so and I will add it.

## Tests

- `test/image-compress.test.ts` (new, web): the two DOM-free decisions — that the threshold is `>` and not `>=`, that GIF/SVG/AVIF are never handed to the canvas, and that fitting a 4000×1 strip inside the box cannot produce a zero-pixel canvas.
- `test/admin-settings.test.ts`: rewritten for the compression pair — defaults, the 1–64 range, the `/api/me` round trip, and that a PUT mixing a good proxy field with a bad threshold writes neither.
- `test/validate.test.ts`: `readJson` reports `ERR_STRING_TOO_LONG` as `413`, an ordinary parse failure still as `400`. Stubbed at the boundary, so it costs no memory to assert the thing that used to need a cap to observe.
- `test/hmr-upgrade-size.test.ts`: the gzip bomb — a few hundred KB on the wire, past `MAX_STRING_LENGTH` inflated — is refused with the message naming the *buffer* bound, so it cannot pass for the wrong reason. Verified to fail on `main`. A 33MB push still goes through. (`test/body-limit.test.ts` is deleted: it tested a cap that no longer exists.)
- `test/task-attachments.test.ts`: the per-file and per-total cases are gone; the count case stays.

`pnpm build`, `typecheck`, `format:check` clean. server 1418 passed (three consecutive full runs, after the bomb test's deflate was moved to 16MB blocks with an explicit timeout — at 1MB blocks it timed out once under a loaded suite), web 1534, docs 53. (`packages/desktop`'s `cli-link.test.ts` symlink case fails on `main` too — verified against a stashed tree.)


Myriad-Dreamin:
Split into three PRs on current main and closing this one — the branch was 159 commits behind, 22 of its 38 files conflicted, and its three commits contradicted each other (the third removed the `/api/hmr` cap the first had added).

- #801 — the inflate bound, the only part that was a defect rather than a direction. Based on main, lands on its own.
- #802 — the browser-side image compression and its admin setting. Deliberately additive: the limits stay, so it can be judged on its own merits.
- #805 — removing the limits and the body caps, stacked on #802. This is the part that contradicts #350 (which landed the admin-settable limits after this PR was opened), so it is a decision of its own rather than a conflict resolution.

Everything here is carried over, including the things this PR was careful about: the `maxOutputLength` bound, the `ERR_STRING_TOO_LONG` → 413 separation, the GIF/SVG exclusion and the return-the-original failure paths. Two things changed in the move: the messaging bridge's inbound file ceilings became its own constants (#513 landed inbound files after this PR, and that path fetches on a stranger's say-so), and the settings page is renamed **Uploads** in #802 rather than #805.

PR #431, opened 2026-08-24
Pick the Skills a new Agent starts with, from the library or from a project directory

hiyouga:
A new Agent started empty. It can now be seeded at creation from two sources.

## From the built-in library

A Skills field on the create dialog opens the multi-select panel the composer already uses, with a `N selected · Select all · Select none` row acting on whatever the search box currently leaves visible. It rides the seam core already had — `AgentPreset.skills` is documented as "Skills installed at initialization" — so nothing parallel was added: library names resolve through `librarySkill()` and install through core's `installSkill`.

## From a project directory

The same dialog can also install what a checkout already carries, which a coding agent leaves in `.agents/skills/` or `.claude/skills/`. Pick a directory and its Skills appear in a picker of their own.

**Why a separate field rather than more entries in the library picker.** A directory Skill may share a library Skill's name and still be the one installed. One flat list of picked names cannot express which was meant, so the two sources stay two fields, and the server takes two arrays.

**Decisions, each pinned by a test:**

- **`.claude` is very often a symlink to `.agents`** — it is in this repository — which would otherwise offer every Skill twice. The two roots are resolved through `realpath` and a shared target is read once. Where both exist as real directories and carry the same name, `.agents` wins: it is the canonical location and `.claude` is the compatibility alias.
- **A directory Skill installs over a library Skill of the same name** picked alongside it. Choosing a directory for this Agent is the narrower intent.
- **A directory carrying nothing installable is a stated fact in the field, not an error.** What is passed over is passed over silently: a directory with no `SKILL.md`, one whose `SKILL.md` has no frontmatter, a name the installer would reject, and a symlinked Skill directory.
- **The client sends names, never Skill content.** Creation re-reads the files from disk, so an install can never be driven by a body the client made up, and a megabyte of SKILL text never crosses the wire.
- **Both sources resolve completely before anything is written**, preserving `createAgent`'s existing contract: a name that has since left the library or the directory fails while the Agent still does not exist, so the id stays free.

**Reading a user-chosen directory** reuses the archive import's caps and its walk discipline rather than growing a second set — 200 files, 5MB per file, 20MB total, with symlinks and other non-regular entries skipped so nothing outside a Skill directory can be pulled in through one. Those constants moved into a shared module, which is why `skills.ts` appears in the diff. The discovery route mirrors the existing `dirs` route exactly on authorization and path handling: `projectId` is the anchor and the caller must have access to it, the path must be absolute, and it is resolved through `realpath` before anything is read. It is strictly narrower than that route — it never lists or descends anything but the two Skill layouts.

The response carries each Skill's `source` (`.agents/skills` or `.claude/skills`). The UI does not render it today — the field label already establishes the origin — but it is what makes the dedup rule observable from the API, and it is asserted in the tests.

## A pre-existing bug fixed alongside

`Modal` had `onClose` in its Escape-effect deps, and every call site passes an inline arrow, so the effect re-ran on every render and re-pushed the dialog's Esc layer *above* any menu opened inside it. One Escape then closed the whole dialog. With the new pickers that meant losing a typed id, name, description and a whole selection. `Modal` now reads `onClose` through a latest-callback ref, the guard `Dropdown` already documents. This also affects the schedule dialog's model and workspace pickers on `main` today.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check` — all green on Node 24. 926 core / 913 server / 292 cli / 1243 web / 100 desktop.
- **10 new server integration tests** in `test/directory-skills.test.ts` covering every decision above: both layouts with their source labels, the symlink-dedup case, `.agents` winning a real collision, the four things passed over rather than offered, the empty directory answering 200, relative/missing/outsider rejections, install with auxiliary files and icon, a directory Skill shadowing a library Skill, 404-with-no-Agent on a vanished name, and half the request pair being a 400.
- Existing server cases for the library path, web unit cases for the selection arithmetic, and `packages/web/e2e/agent-create-skills.spec.mjs` for the library flow; the e2e suite was baselined against a `main` worktree, where the same thirteen specs fail on both sides.
- **Driven in the real app** against a scratch data root: a temp checkout carrying two Skills in `.agents/skills` and one in `.claude/skills` reported "3 Skills found in this directory", and after Select all + Create the new Agent had exactly `repo-conventions`, `release-checklist` and `legacy-helper` on disk. Pointing the field at a directory with no Skills reported "This directory carries no installable Skills" rather than failing.
- `WorkspaceSelect` gained two optional copy overrides so the reused directory picker does not say "Temporary workspace" in a field that has nothing to do with Workspaces.


packages/server/src/services/directory-skills.ts
```diff
@@ -0,0 +1,173 @@
...
+): Promise<DirectorySkill | null> {
+  let content: string;
+  try {
+    content = await fs.readFile(path.join(abs, "SKILL.md"), "utf8");
+  } catch {
+    return null;
+  }
+  const metadata = parseSkillFrontmatter(content);
+  if (!metadata) return null;
+  const icon = await fs.readFile(path.join(abs, "icon.svg"), "utf8").catch(() => undefined);
```
hiyouga: **`SKILL.md` and `icon.svg` are read without the regular-file guard the module header promises**, so unlike `readAuxiliaryFiles` (which checks `entry.isFile()`) these two follow symlinks and open FIFOs/device nodes.

Two consequences, both reproduced against this branch:

1. **Arbitrary file disclosure.** `.agents/skills/leak/SKILL.md` (valid frontmatter) plus `icon.svg -> /home/svc/.ssh/id_rsa` makes `GET /api/projects/:p/dir-skills?path=<dir>` answer `200` with the private key verbatim in `skills[0].icon` — the route strips `content` and `files`, not `icon`. Creating the Agent then copies it to `skills/leak/icon.svg`, where the model can read it. An Agent that can write into any directory a user may pick can plant this.
2. **Availability.** `mkfifo .agents/skills/x/icon.svg` makes `fs.readFile` block forever; the `.catch(() => undefined)` never fires. With the default `UV_THREADPOOL_SIZE=4`, four such requests wedge every filesystem operation in the server. (Measured: the call never resolves.)

The header says the opposite: *"symlinks and other non-regular entries are skipped, so nothing outside the Skill directory can be pulled in through one."*

_Addressed in 00ca1fb3: both files go through `readSkillFile`, which `lstat`s first and reads only a regular file within the per-file cap._

packages/server/src/services/directory-skills.ts
```diff
@@ -0,0 +1,173 @@
...
+  const files = await readAuxiliaryFiles(abs);
```
hiyouga: **One unreadable or oversized Skill fails the whole directory.** `readSkillDir` try/catches only the `SKILL.md` read, so anything `readAuxiliaryFiles` throws propagates through `discoverDirectorySkills` and out of the route.

Reproduced: a checkout with `good/` and `big/` (one 6MB file) answers `413 skill_too_large` for the entire listing — `good` is never offered. An `EACCES` subdirectory inside one Skill gives an uncaught `Error` and a `500` for the listing. `resolveDirectorySkills` re-runs the same full scan, so the fat Skill also fails Agent creation for a *different*, tiny picked Skill.

This contradicts the entry's own rule that *"what is passed over is passed over silently"*.

_Addressed in 00ca1fb3: `readSkillEntry` is wrapped per Skill, so one bad Skill is passed over like any other non-installable directory; a new test pins it._

packages/server/src/http/routes/agents.ts
```diff
@@ -50,7 +57,32 @@ export function agentsRoutes(deps: AppDeps): Hono<AppEnv> {
       maxLen: 2000,
       label: "description",
     });
-    const item = await deps.agentService.createAgent(projectId, agentId, name, description);
+    // Library Skills to seed the new Agent with (the create dialog's picker); unknown names are
+    // rejected before the Agent directory exists.
+    const skills = optionalStringArray(body, "skills");
+    // Skills imported from a directory the user picked. The pair only means anything together, so
+    // half of it is a bad request rather than a silently ignored field.
+    const skillsDirectory = optionalString(body, "skillsDirectory", {
```
hiyouga, line 66: **`skillsDirectory` gets none of the path admission the discovery route applies** — no `path.isAbsolute`, no `realpath`, no `isDirectory` — only a length check.

Reproduced: `POST /api/projects/:p/agents {"agentId":"a","skillsDirectory":".","directorySkills":["x"]}` answers **201** and installs from `path.join(".", ".agents/skills")` relative to the *server process's cwd* — a tree the client never picked and could not have browsed to. The same input on `GET /dir-skills` is a `400 dir_not_absolute`. A non-directory degrades to a misleading `404 unknown_skill` instead of `not_a_dir`, and that message echoes the caller's raw path back.

The route header claims the two share path handling; nothing in the tests covers a relative `skillsDirectory` on POST.

_Addressed in 00ca1fb3: both routes and the `dirs` browser now share `requireProjectDir`; a new test pins the 400._

packages/server/src/services/directory-skills.ts
```diff
@@ -0,0 +1,173 @@
...
+    for (const entry of await fs.readdir(abs, { withFileTypes: true })) {
+      const relChild = rel ? `${rel}/${entry.name}` : entry.name;
+      if (entry.isDirectory()) {
+        await walk(path.join(abs, entry.name), relChild);
+        continue;
+      }
+      // Non-regular entries are skipped, symlinks included: a link is how a file outside the
+      // Skill directory would otherwise be read.
+      if (!entry.isFile()) continue;
+      if (relChild === "SKILL.md" || relChild === "icon.svg") continue;
```
hiyouga, line 136: **`SKILL.md` and `icon.svg` are excluded from every cap.** They are skipped here, and `readSkillDir` applies none of its own — so the "200 files, 5MB per file, 20MB total" the changelog advertises covers only auxiliary files.

Reproduced: a 31MB `icon.svg` is read whole and **returned in the GET response body** (icon is not stripped); a 31MB `SKILL.md` is read whole. Above ~512MB Node throws `ERR_STRING_TOO_LONG`, so the endpoint answers `500` rather than the intended `413`. The archive route caps every entry including these two.

_Addressed in 00ca1fb3: `readSkillFile` refuses either file above `MAX_FILE_BYTES`, from the stat._

packages/server/src/services/directory-skills.ts
```diff
@@ -0,0 +1,173 @@
...
+      const data = await fs.readFile(path.join(abs, entry.name), "utf8");
```
hiyouga: **The cap is consulted after the file is already a JS string**, so `MAX_FILE_BYTES` cannot bound the memory a directory pick costs: a 900MB file under a picked `.agents/skills/<name>/` is fully materialised (and UTF-8 decoded) before line 91 rejects it.

`collectSkillArchive`, the walk this says it mirrors, at least works from an already-decoded zip. A `stat` before the read is the cheap fix.

_Addressed in 00ca1fb3: the size comes off `fs.stat` and the read only happens under the cap._

hiyouga: **Auxiliary files are read as UTF-8 and written back as UTF-8, so any binary asset a checkout's Skill ships is silently corrupted.**

Reproduced: `logo.png` bytes `89 50 4e 47 …` come back as `ef bf bd 50 4e 47 …` — `installSkill` writes the mojibake with `fs.writeFile(file, data, "utf8")` and reports success.

As a knock-on, `Buffer.byteLength(decodedString)` is not the file's size: every invalid byte becomes 3, so a ~2.8MB binary measures ~5MB and can trip `MAX_FILE_BYTES` while a genuinely 5MB file was meant to pass. `collectSkillArchive` keeps `Uint8Array` end to end for exactly this reason.

`LibrarySkill.files` documents the UTF-8 assumption for *committed library text*; a user-picked checkout does not satisfy it.

**Still open** — a real fix needs `installSkill` to accept bytes (a core signature change, outside this diff), or the walk to skip non-text files deliberately. Worth a decision either way rather than silent corruption.

packages/server/src/http/routes/directory-skills.ts
```diff
@@ -0,0 +1,62 @@
...
+        404,
+        "dir_not_found",
+        `Directory does not exist or is inaccessible: ${target}.`,
+      );
+    }
+    if (!(await fs.stat(real)).isDirectory()) {
+      throw new HttpError(400, "not_a_dir", "Not a directory.");
+    }
+
+    const skills = await discoverDirectorySkills(real);
```
hiyouga, line 33: **Discovery reads the entire installable payload of every Skill just to build a name list, and the next line throws it away.** `readSkillDir` unconditionally calls `readAuxiliaryFiles`, and `count`/`total` are declared *inside* that function — so the caps are per Skill and there is no bound on one request.

Measured: 60 Skills carrying 1MB each = 63MB read, decoded and held simultaneously in `byName` for a response that is a few KB. `resolveDirectorySkills` then repeats the whole scan to satisfy one picked name. The picker re-fires this endpoint on every directory change.

_Partly addressed in 00ca1fb3: discovery is metadata-only now and the payload is read for the picked names alone. A per-request aggregate bound is still absent — with each of `SKILL.md`/`icon.svg` capped at 5MB, a checkout with very many Skill directories is still unbounded in aggregate._

packages/server/src/services/agent-service.ts
```diff
@@ -255,11 +265,23 @@ export class AgentService {
...
+      ? await resolveDirectorySkills(directory.path, directory.names)
+      : [];
+    const seedSkills = [...librarySeed, ...directorySeed];
     await coreCreateAgent({ root: this.root, projectId, agentId });
     try {
       await this.agentConfig.updateConfig(projectId, agentId, {
         config: { name: displayName, ...(description !== undefined ? { description } : {}) },
       });
+      for (const skill of seedSkills) {
+        await installSkill(this.root, projectId, agentId, skill);
```
hiyouga, line 283: **The "resolve everything before writing anything" contract has a hole: auxiliary *paths* are never validated at resolve time.**

`readAuxiliaryFiles` keys entries by raw filename, and a backslash is a legal filename character on Linux (common in trees copied from Windows). `resolveDirectorySkills` returns cleanly, `coreCreateAgent` runs, and then `installSkill`'s `assertSafeSkillFile` throws a **plain `Error`**, not an `HttpError`.

Reproduced with `.agents/skills/bs/notes\draft.md`: `500 {"error":{"code":"internal"}}` plus an `Unhandled exception` log line, from inside the create-then-`rm` cleanup window the resolve step exists to avoid. (Cleanup does work — the id stays free.)

_Addressed in 00ca1fb3: `isSafeSkillFilePath` mirrors core's rule and passes such a file over during the walk._

packages/web/src/features/agents/agents-page.tsx
```diff
@@ -120,6 +174,42 @@ export function AgentsPage() {
...
+  // Re-read whenever the picked directory changes. A directory that carries no Skills answers with
+  // an empty list, which the field states in place of its hint rather than treating as a failure.
+  useEffect(() => {
+    if (!createOpen || !skillsDir || !projectId) {
+      setDirSkills(null);
+      setDirSkillsError(null);
+      return;
+    }
+    let cancelled = false;
+    setDirSkillsError(null);
```
hiyouga, line 187: **Changing the picked directory never clears `dirSkills`**, so the previous directory's Skills stay listed, stay selected and stay submittable until the new fetch lands.

Pick `/repoA`, tick `deploy`, then pick `/repoB` and click 创建 before the response arrives: the body is `{skillsDirectory: "/repoB", directorySkills: ["deploy"]}` → `404 unknown_skill`, no Agent. If `/repoB` happens to carry its own `deploy`, **that** one is installed silently instead of the one the user inspected. Meanwhile the hint still reports repoA's count and the `S.common.loading` branch is unreachable on a switch, because it only renders while `dirSkills === null`. The scan is a full recursive read, so the window is seconds.

_Addressed in 00ca1fb3: `setDirSkills(null)` before the new fetch._

packages/web/src/features/chat/workspace-select.tsx
```diff
@@ -320,7 +331,7 @@ export function WorkspaceSelect({
         leading={folderIcon("")}
         label={label}
         {...(trimmed ? { labelClassName: "font-mono" } : {})}
-        title={trimmed ? `${S.chat.workspace}：${trimmed}` : S.chat.workspaceHint}
+        title={trimmed ? trimmed : (menuHint ?? S.chat.workspaceHint)}
```
hiyouga: **The tooltip lost its `Workspace：` label prefix for every pre-existing caller**, not just the new one — same change at line 356 for the pill.

`draft-view.tsx:762`, `sidebar.tsx:1536`, `project-dialogs.tsx:775` and `schedules-tab.tsx:679` all showed `工作区：/home/me/proj`. The trigger only renders the last path segment, so that tooltip was the one place the field's identity sat next to its value; it is now a bare path. Only the *empty* branch needed the new host's override, and neither changelog entry records the change (CONTRIBUTING.md: *"Every change ships with a changelog entry, in both languages"*).

_Addressed in 00ca1fb3: the prefix is back, using a new `fieldLabel` prop that defaults to `S.chat.workspace`._

packages/web/src/features/agents/agents-page.tsx
```diff
@@ -474,6 +582,102 @@ export function AgentsPage() {
...
+                  : S.agent.createSkillsPicked(createSkills.length)
+              }
+              muted={createSkills.length === 0}
+              title={S.agent.createSkills}
+              ariaLabel={S.agent.createSkills}
+              disabled={busy}
+              menuClass="w-[26rem]"
+            >
+              <SkillPickList
+                skills={library ?? []}
```
hiyouga, line 602: **`library ?? []` collapses "still loading" into "empty"**, so opening the picker before the fetch lands asserts 「技能库暂无可安装的技能」 / "The skill library has nothing to install." — a false statement that invites the user to close it and create a plain Agent.

The directory field right beside it distinguishes the three states (`dirSkills === null ? S.common.loading : …`), so the two adjacent fields disagree on the convention. On failure the state is also indistinguishable from a genuinely empty library except for the error text under the closed trigger.

_Addressed in 00ca1fb3: the empty hint says loading while `library === null`._

packages/server/src/http/routes/directory-skills.ts
```diff
@@ -0,0 +1,62 @@
...
+    return c.json({
+      path: real,
+      // Content and auxiliary files are deliberately not returned: the picker needs to describe a
+      // Skill, and creation re-reads it from disk, so a megabyte of SKILL bodies never crosses the
+      // wire and the install can never be driven by a body the client made up.
+      skills: skills.map(({ content: _content, files: _files, ...item }) => item),
```
hiyouga: **The response is projected with a two-field denylist rather than the repo's allowlist**, so it publishes fields `DirectorySkillItem` does not declare.

`parseSkillFrontmatter` emits `preinstall` when a SKILL.md carries `preinstall: false`, and the spread hands it straight to the client. `satisfies DirectorySkillsResponse` does not catch it — the mapped array is not an object literal, so there is no excess-property check. Any future `SkillMetadata` field is published here automatically while the two `toMetadataItem` endpoints correctly withhold it, and nothing fails to compile.

The existing projection is `toMetadataItem` (`routes/skills.ts:66`), used by both other Skill endpoints.

_Addressed in 00ca1fb3: `toMetadataItem` moved next to `resolveLibrarySkills` and all three endpoints share it._

packages/web/src/features/chat/workspace-select.tsx
```diff
@@ -288,7 +299,7 @@ export function WorkspaceSelect({
       )}
       {/* Hint text (bottom of the menu) */}
       <p className="px-0.5 text-xs leading-5 text-gray-400 dark:text-gray-500">
-        {S.chat.workspaceHint}
+        {menuHint ?? S.chat.workspaceHint}
```
hiyouga, line 311: (Anchored to the nearest changed line; the code this is about is the `{S.chat.workspaceClear}` link at line 297 and the `aria-label={S.chat.workspace}` at 205/335.)

**The copy overrides cover 2 of ~7 Workspace-flavoured strings, so the new field still speaks the old vocabulary.**

In *Import Skills from a project directory*, once a directory is picked the menu renders `S.chat.workspaceClear` — "Use a temporary workspace instead" / 「改用临时工作区」 — which here just blanks the field, and it is the only control that does so. Both triggers and the in-menu path input keep `aria-label={S.chat.workspace}`, so a screen reader announces the field as "Workspace" under a visible label reading "Import Skills from a project directory" (WCAG 2.5.3 label-in-name).

_Addressed in 00ca1fb3 with `fieldLabel` + `clearLabel`. The deeper form is still worth considering: almost nothing in this component is Workspace-specific, so a `DirectoryPicker` primitive with `WorkspaceSelect` as its Workspace-flavoured host would stop the next non-Workspace host adding a fifth override prop._

packages/web/src/features/agents/agents-page.tsx
```diff
@@ -135,9 +225,27 @@ export function AgentsPage() {
...
       if (description.trim()) body.description = description.trim();
+      // Picked Skills are seeded server-side inside the same create call, so a failure leaves no
+      // half-equipped Agent behind.
+      if (createSkills.length > 0) body.skills = createSkills;
+      // The pair only means anything together, so it is sent only when a directory actually
+      // contributed something — picking a directory and then no Skills from it is a plain Agent.
+      if (skillsDir && createDirSkills.length > 0) {
+        body.skillsDirectory = skillsDir;
+        body.directorySkills = createDirSkills;
+      }
```
hiyouga, line 252: (Anchored to the nearest changed line; the code this is about is `catch (e) { setIdError(apiErrorText(e)); }` a few lines below.)

**Every create failure is written to `idError`**, so the new Skill-shaped failures land on the *Agent id* field.

A picked directory Skill deleted between the picker fetch and Create gives `404 Skill is not in /w: alpha`, rendered under the id input whose hint is the id format rule. A `413 skill_too_large` lands there too. The user has no indication which of the two Skill fields to fix, and typing in the id box clears the message (`onChange` does `setIdError(undefined)`) even though the id was never the problem.

**Still open** — the field already carries a comment saying *"The id is the only validated create field"*, which stopped being true in this PR. A form-level error slot, or routing skill errors to the field that owns them, would fix it.

packages/server/src/services/agent-service.ts
```diff
@@ -255,11 +265,23 @@ export class AgentService {
...
+    // Both sources are resolved before a single file is written, so a name that has since left the
+    // library or the directory fails while the Agent still does not exist. Directory Skills are
+    // installed after the library ones and so win a name collision: the user picked that directory
+    // for this Agent specifically, which is a narrower intent than "install the built-in one".
+    const librarySeed = resolveLibrarySkills(skillNames ?? []);
+    const directorySeed = directory
+      ? await resolveDirectorySkills(directory.path, directory.names)
+      : [];
+    const seedSkills = [...librarySeed, ...directorySeed];
     await coreCreateAgent({ root: this.root, projectId, agentId });
```
hiyouga, line 277: **Seed resolution puts seconds of directory I/O between the duplicate-id check and `coreCreateAgent`, and the failure path of that window is `rm -rf` on a shared directory.**

The `taken` check is at line 265; the DB row is only written at line 294. Two `POST /agents` with the same `agentId` both pass the check. Request A then sits in `await resolveDirectorySkills(...)` walking a checkout — this PR's own caps allow 200 files x 5MB *per Skill*, and the whole directory is scanned to satisfy one picked name. Request B, with no `skills`/`skillsDirectory`, races through and returns `201`.

A resumes, `coreCreateAgent` takes the load path on the now-existing directory, and any error in the seed loop sends it into the catch — which runs `fs.rm(agentDir(root, projectId, agentId), { recursive: true, force: true })`, **deleting B's Agent on disk while B's DB row survives**. The result is an Agent that lists but has no directory, with an id that is now permanently 409.

The check-then-create window is pre-existing, but it used to be microseconds of `updateConfig`; this PR makes it unbounded and adds a new class of error (`unknown_skill`, `skill_too_large`, an unwritable auxiliary path) that reaches the `rm`.

**Still open** — the cheap mitigation is to re-check `taken` immediately before `coreCreateAgent`; the real fix is a per-`(projectId, agentId)` create lock, or making the cleanup conditional on this request having been the one that created the directory.

PR #435, opened 2026-08-24
修复切换项目时会话串入问题

ericperfect:
## Summary

Prevent a Session from another Project from being inserted into the currently selected Project's conversation list when switching Projects with an existing `/chat/:sessionId` route.

## Root cause

The Session store is correctly cleared and reloaded by `projectId` when the user switches Projects. However, the browser route can still reference the Session that was open in the previous Project.

The chat page supports deep links to Sessions that are not present in the currently loaded sidebar page. It resolves the routed Session through the global Session endpoint and previously inserted the result into the current Session store without verifying its `projectId`.

This produced the following sequence:

1. Open a Session in Project A.
2. Switch to Project B while the old `/chat/:sessionId` route remains active.
3. Project B correctly loads its own Session list.
4. The deep-link fallback fetches Project A's Session by ID.
5. The fetched Session is inserted into Project B's frontend list.

The database records and workspace files remain Project-scoped, but the UI can display and continue operating on a Session from another Project.

## Changes

- Validate that a directly fetched Session belongs to the currently selected Project before adding it to the Session store.
- Treat a cross-Project Session result as a failed deep-link probe.
- Let the existing navigation logic redirect to the target Project's latest conversation or new-chat page.
- Scope probe failure state by both `projectId` and `sessionId`, preventing stale failure state from being reused after another Project switch.
- Add regression tests covering matching Projects, cross-Project rejection, and Project-scoped probe keys.

## Expected behavior

After switching Projects:

- only Sessions belonging to the selected Project appear in the sidebar;
- a route pointing to another Project's Session is rejected;
- the user is redirected to a valid Session in the selected Project, or to the new-chat page;
- valid deep links within the same Project continue to work.

## Compatibility

No database schema, stored Session data, workspace files, or API contracts are changed. This is a Web App routing and state-isolation fix.

## Verification

- `pnpm --filter @prismshadow/penguin-web test`
  - 94 test files passed
  - 1,241 tests passed
- `pnpm --filter @prismshadow/penguin-web typecheck`
- `pnpm format:check`
- `git diff --check`

### DIFF
--- a/changelog/unreleased/2026-08-24-project-session-isolation.md
+++ b/changelog/unreleased/2026-08-24-project-session-isolation.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-24
 - **Type:** fix
 - **Scope:** `web`
+- **PR:** [#435](https://github.com/Prism-Shadow/penguin-harness/pull/435)
 
 [中文版](2026-08-24-project-session-isolation.zh.md)
 
--- a/changelog/unreleased/2026-08-24-project-session-isolation.md
+++ b/changelog/unreleased/2026-08-24-project-session-isolation.md
@@ -1,0 +1,14 @@
+# Keep deep-linked Sessions inside their Project
+
+- **Date:** 2026-08-24
+- **Type:** fix
+- **Scope:** `web`
+
+[中文版](2026-08-24-project-session-isolation.zh.md)
+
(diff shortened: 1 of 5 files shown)


hiyouga (approved):
LGTM

PR #436, opened 2026-08-24
feat: hot reload agents

Myriad-Dreamin:
## What

A platform hot push used to hard-stop every running agent Task: `manager.shutdown()` aborted the LLM stream, killed tool subprocesses, and denied pending approvals. This PR makes a running Task survive the swap, built around a per-session **monitor** (管程):

1. **`SessionMonitor`** (`runtime/session-monitor.ts`, new): one object per session owning ALL of its run state — status, pending approvals, the interrupt controller, queued follow-ups, display mirrors — with every operation serialized at the monitor boundary. The former entry table is gone and not exposed: `SessionManager` holds no per-session state and becomes the facade plus the generation's procedure set (`GenerationOps`) the monitor delegates to. The monitor itself stays small and stable: state, mutual exclusion, and the pointer-swap rule only — policy is always reached through the `current` pointer, so it is never frozen into surviving objects.
2. **Swap = pointer replacement at the event boundary.** Monitors ride the resource registry as ONE shared `agent-sessions:table`. A successor App's manager attaches itself to every monitor at construction (`setPending`); an idle monitor swaps its code pointer immediately (all accepted events are complete), a busy one at its run's settle — and the next event is processed by the new generation. The loaded `RuntimeSession` object is generation code: dropped at the swap, reloaded from the Trace (the durable state) by the next event.
3. **Engine — turn-boundary park** (`RunOptions.shouldPark`, ~15 lines): consulted before each turn's LLM request; a park holds the pending turn input as carry-over (the max-turns mechanism reused) and ends the generator **silently**. The dying generation's `quiesce()` requests it; the parked continuation is simply the session's next event, relaunched by the successor with an empty input — Trace replay's positional carry-over reconstruction brings the pending turn back, so the transcript continues without an aborted turn or redone side effects.
4. **No cross-generation forwarding.** Approvals, interrupts, status, steering and follow-up queues live in the surviving monitor, so during the lame-duck window they all just work through the successor's facade — including steering into the old run and queuing follow-ups that auto-start under the new generation. An incompatible successor hard-stops the table through the declared-resource gate and falls back to exactly the previous abort behavior; the process-exit sweep stops a lame duck that never found an adopter.

## Why this shape

At a turn boundary the run's continuation state is empty by construction and its durable form is the Trace, which the session loader already resumes from — so "migrate the run" degenerates to "reload and continue", with no engine rewrite. Making the session a monitor puts that handoff behind one door: there is no second representation of a busy session to keep in sync, and any future per-session surface is automatically swap-correct.

## Known limits (deliberate scope)

- The lame-duck window is the current turn **including its tool calls**; a `run_subagent` call stretches it to the whole subagent tree. Rescue remains available via the interrupt.
- Behavior pushed in an update applies from the parked run's next turn (the resumed run reloads through the new generation), not mid-turn.
- The old generation's session environment is disposed at the pointer swap, so background commands (dev servers) don't survive it — unchanged from before.
- Goal runs and compactions keep abort semantics (no parkable boundary contract yet), still gating the successor's boot within the drain grace.
- The web UI sees a brief `task_state` idle→running flip at the handoff instant.

## Verification

- `pnpm build`, `pnpm typecheck`, `pnpm format:check` — clean.
- Core suite: 931 passed (2 new engine tests: silent park + empty-input resume pairing the committed tool_call; park-before-first-turn + nothing-to-send no-op).
- Server suite: 964 passed (5 monitor handoff tests: park/adopt/resume with empty input over a shared table, interrupt-cancels-resume, approval decisions through the successor's facade with no forwarding, compaction hard-abort gating the drain, follow-ups queued before a swap auto-starting under the successor; plus 1 integration test driving a REAL kernel `upgrade()` over the full App — quiesce in the dispose effect, the table's registry entry, adoption and resume in the successor's `create()`).


Myriad-Dreamin:
Superseded: restarting on the event-loop architecture (the run loop inverted out of the engine into a stable per-session wrapper), per the design discussion. Branch kept for reference.

PR #440, opened 2026-08-24
feat(web): drag the sidebar's groups into your own order

hiyouga:
The chat sidebar's groups — the Workspace folders and the Agents — can now be dragged into a manual order, the way the conversation rows inside them already could. The order is stored per Project and per grouping mode, and it survives a reload.

`packages/web/src/lib/group-order.ts` is the row order's counterpart one axis up. It reuses `session-order.ts`'s sequence algebra (`applyManualOrder`, `moveInSequence`, `applyManualReorder`) rather than restating it — what differs is the key namespace, which modes may be ordered, and that there is no sort-mode switch.

## Decisions

**Implicit mode.** There is no second toggle. A Project and mode with nothing stored renders exactly the automatic sort it did before, because an empty order is the identity for `orderGroups`. The sidebar already has one sort toggle; a second one for groups would be a setting for a gesture that already says what it means.

**Workspace and Agent modes only.** The time buckets are last day / last month / earlier — a fixed chronological ladder where "put last month above last day" is not an order anyone can mean. `isOrderableGroupMode` is the single gate, and the *store itself* refuses to read or write for `"time"`, so the exclusion cannot be undone by forgetting a check at one call site. Time mode's render branch is untouched by this PR.

**The temp group drags like any other.** `TEMP_WORKSPACE_GROUP_KEY` is stable, so it stores like any other key. Its trailing position stays the default the automatic sort supplies; once dragged, the user's position wins.

**Unplaced groups go to the top**, keeping their automatic order among themselves — the rule rows use for new Sessions — so a Workspace used for the first time or a freshly created Agent does not hide at the bottom of a long list.

## Why nothing prunes

Originally a drop pruned stored keys whose group no longer existed, gated on a
`liveGroupKeys()` proof that the mode's key set was complete. Review showed the proof
does not hold, so the prune is gone rather than guarded harder.

- **Workspace mode** rested on each Agent's `workspaceCounts` covering *all* of that
  Agent's Sessions. It does not: `SessionService.listSessions` builds its row set with
  `listByAgent(…, { webOnly: !includeCli })` **before** counting, so with the "show CLI
  sessions" preference off — the default — a Workspace used only from the CLI is absent
  from a complete-looking picture and would be proved dead while alive.
- **Agent mode** rested on `agentsLoading === false`, which is not proof of a *successful*
  fetch (`reloadAgents` has no `catch`, and the provider clears `agents` first). An empty
  Agent list is indistinguishable from a rejected one, and `agents.some(…)` is vacuously
  true for an empty array — so one drag among two registered Workspaces could persist an
  order stripped of every real key.

What pruning bought was small: stale keys are already inert (`applyManualOrder` does
membership lookups only), and the array is bounded by the groups that were on screen at
some drop — a few hundred short strings. What it cost when the proof was wrong was an
arrangement the user cannot restore. A group that returns under the same key now resumes
its place, which is the honest reading of "same key, same group".

## A drop is a splice, not a rewrite

The rows' `applyManualReorder` moves the dropped partition's sequence to the **front** of
the stored array. That is sound for Sessions — a partition is one group's own list, and
the ids it displaces are read only within *other* partitions — but it does not lift one
axis up:

- A Workspace group exists only once one of its Sessions has paged in, so the keys a group
  drop displaces are its own peers. Front-loading demoted every not-yet-loaded Workspace to
  the tail, and it came back at the bottom.
- The two pin partitions share one array, so front-loading the dragged partition flung the
  other one down the list: arrange `A,B,C,D`, pin `D`, drag `C` up, unpin `D` — and `D`
  went from first to last, possibly past the 10-group display cap and out of sight.

`commitGroupOrder` instead materialises the rendered keys into the stored array **where
they already render** (unlisted ones at the front, which is where `applyManualOrder` puts
them) and then moves the dragged key next to its target inside that array. Keys that are
stored but not rendered keep their positions relative to everything else.

## What is shared with the row order

`orderWithinPinPartitions` (the pinned-cluster-then-manual-order kernel) and
`readStringArray` / `writeStringArray` (the localStorage body) now live once in
`session-order.ts` and are called from both axes; `orderGroups` is a one-line call.

The drag *wiring* is deliberately not shared. `groupDragProps` and `renderRows`'s drag
object have the same shape but differ in real ways — scope-keyed versus single key, gated
on a global sort mode versus implicit — and folding them together means designing an
abstraction over the row path, which this PR does not touch and which no test covers.

## Known, and not addressed here

- **One drag freezes the mode's automatic sort.** A drop commits every key of the
  partition, so after the first drag every group is "listed" and recency ordering no
  longer applies to groups the user never touched. There is also no way to clear a group
  order — the 排序方式 radios drive rows only. Both follow from "dragging is itself the
  intent"; giving groups a reset (or committing only what was dragged) is a design call.
- **Unlisted-goes-to-the-top** overrides two deliberate "sorts last" policies once an
  order exists: `groupSessionsByWorkspace`'s forced-last temp group, and
  `mergeRegisteredWorkspaces`'s empty registered Workspaces.
- **The drop zone is the header, but the indicator is drawn against the whole block**, so
  on an expanded group the "below" line paints far from the only place a drop lands.
- **The header is the whole drag handle**, so a press on one of its action buttons that
  drifts past the drag threshold starts a drag instead of clicking.

## Keeping manual order out of the Traces page

**By keeping the ordering decision out of the shared grouping function**, not by gating inside it. `session-grouping.ts` is unmodified in this PR: `groupSessionsByWorkspace` still returns its automatic sort, and the Traces page calls it and `pinnedFirst` exactly as before. The sidebar applies `orderGroups(...)` on top of that result, in the sidebar. There is no flag to pass and no call site to get wrong.

## The two axes are independent

Group order and row order are separate arrays under separate key namespaces (`penguin.groupOrder.<projectId>.<mode>` vs `penguin.sessionOrder.<projectId>.<mode>`), and `orderGroups` only permutes the outer list — group objects and their member sequences are passed through untouched. A group drag never writes the row-order key, and the row drag is unchanged. Both directions are asserted in the browser (below).

Group order composes with the group pins the way row order composes with the row pins: the pinned cluster stays first, and a drop stays inside its own pin partition, so dragging can reorder but never pin or unpin.

## Phone width

The group header is the drag handle rather than a new grip button, and the drop indicator is an absolutely positioned `inset-x-1` line — so neither costs layout width and the header keeps its up-to-three actions in the drawer. The handle is the header rather than the whole group block so the Session rows inside keep their own row-level drag. The indicator is drawn against the whole group block, so "below" reads as "after this group and its conversations" rather than "between the header and its own first row".

`draggable` is gated on the existing `canDrag` media query (HTML5 DnD never fires from touch), but a stored order still *applies* on such a device — unlike the rows' global sort mode there is nothing to degrade, so a phone renders the arrangement its owner made at a desk.

## Notes

- No new user-facing strings, so `strings.ts` / `strings-en.ts` are untouched and trivially the same shape.
- Like the row drag, this is pointer-only and has no keyboard equivalent; that parity gap is pre-existing and not widened here.

## Verification

Full CI-parity chain on Node 24 after the review fixes, all green:

- `pnpm -r build` — OK
- `pnpm typecheck` — OK
- `pnpm test` — 3680 passing, 5 skipped across the eight packages (`packages/web` 1282/1282, `packages/server` 927/927; no `terminal-stream` flake)

**Unit tests** (`packages/web/test/group-order.test.ts`, 30 tests): the drop-commit suite
now covers the splice — a not-yet-loaded group keeping its stored place, unlisted groups
folded in where they render, pin-then-unpin not flinging a group down the list, stale keys
carried along untouched — alongside the original coverage of the empty-order identity, the
temp group, per-Project and per-mode isolation, and degradation on malformed or hostile
storage (including the no-argument call that actually exercises the `?? localStorage`
branch). The prune suite is gone with the prune.

**Browser** (Playwright against a real server on a temp data root, four seeded Workspace
groups including the merged temp group):

- automatic order `[ws-gamma, ws-beta, ws-alpha, Temporary workspaces]` — unchanged;
- drag the temp group to the front → `[Temporary workspaces, ws-gamma, ws-beta, ws-alpha]`;
  **reload** → identical, so the order persisted;
- with a stored order whose first key belongs to a group that has **not** loaded, that key
  is still at index 0 after the drag (the front-loading commit moved it to the tail);
- the Session rows inside `ws-alpha` are byte-identical before and after the group drag,
  and no `penguin.sessionOrder.…` key exists at that point.


packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+   *
+   * Agent mode: the Project's Agent list is fetched whole, so once it has settled a
+   * stored key with no Agent is proof the Agent is gone. Workspace mode: the groups
+   * themselves surface as Sessions page in, but each Agent's `workspaceCounts` covers
+   * ALL of that Agent's Sessions rather than the returned page, so the union of the
+   * folded count keys with the rendered groups (which is where a registered but empty
+   * Workspace comes from) is the whole truth — once every Agent has reported.
+   */
+  const liveGroupKeys = (): ReadonlySet<string> | null => {
+    if (agentsLoading) return null;
```
hiyouga: **`liveGroupKeys()`'s completeness gate is vacuously satisfied when `agents` is empty — a drop then prunes and persists a group order stripped of every real key.**

`reloadAgents()` (`state/project.tsx:110-127`) is `try/finally` with no `catch`, and the provider effect clears `agents` first — so a rejected `listAgents` (server restart, 401, network blip) leaves `agents: []` with `agentsLoading: false`. In workspace mode `agents.some((a) => !workspaceCountsByAgent.has(a.agentId))` is then `false` for an EMPTY array, the gate reports “complete”, and `live` collapses to the localStorage-backed registered Workspaces (which still render as empty draggable groups via `mergeRegisteredWorkspaces`). One drag between two registered Workspaces runs `pruneGroupOrder(reordered, live)` and `saveGroupOrder` writes the truncated array: the whole manual group order for that Project is silently and permanently gone.

```suggestion
    if (agentsLoading) return null;
    // An EMPTY Agent list proves nothing, and `agentsLoading === false` is not proof of a
    // successful fetch: reloadAgents() clears `agents` up front and flips the flag in a
    // `finally`, so a rejected listAgents leaves exactly this state.
    if (agents.length === 0) return null;
```

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+   * and take the chance to prune keys of groups that no longer exist — a drop is the
+   * one moment the user restates the whole arrangement, and the array is being
+   * rewritten anyway, so no separate destruction event has to be listened for.
+   */
+  const commitGroupDrop = (sequence: readonly string[], targetKey: string, after: boolean) => {
+    if (dragGroup === null) return;
+    const seq = moveInSequence(sequence, dragGroup, targetKey, after);
+    // Identity guard, as for rows: a drop that changes nothing must not persist a fresh array.
+    if (seq === sequence) return;
+    const reordered = applyManualReorder(groupOrder, seq);
```
hiyouga: **`applyManualReorder` front-loads only the groups that have MATERIALISED, demoting every stored-but-unloaded Workspace key to the tail** — the opposite of the “a group that has merely not loaded keeps its place” promise.

`workspaceGroupSequence` derives from `workspaceGroups = mergeRegisteredWorkspaces(groupSessionsByWorkspace(sessions), …)`, i.e. only Workspaces with a LOADED Session (10 active rows per Agent), not from `workspaceCountsByAgent`.

Stored order `[/w6,/w5,/w4,/w3,/w2,/w1]`; on a fresh load only `/w1../w3` have a session in the first page. One small drag among them commits `applyManualReorder([...], ['/w3','/w1','/w2'])` = `['/w3','/w1','/w2','/w6','/w5','/w4']` and persists it. When `/w4../w6` later page in they render at the BOTTOM.

The comment above `groupDragProps` (“the mode's FULL ordered key list”) and the changelog only addressed the `groupCap` display slice, not session paging. `test/group-order.test.ts:275` actually encodes the demotion (`gone-a` moves index 0 → 2) under a name that claims the opposite.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -2162,6 +2310,39 @@ function DraftRow({
...
+  dropEdge: "above" | "below" | null;
+  children: ReactNode;
+}) {
+  return (
+    <div className="relative pt-2.5">
+      {dropEdge !== null && (
+        <div
+          aria-hidden
+          className={`pointer-events-none absolute inset-x-1 z-10 h-0.5 rounded-full bg-[var(--accent-bg)] ${
+            dropEdge === "above" ? "top-1" : "-bottom-1"
```
hiyouga, line 2353: **The “below” drop line is drawn against the whole `GroupBlock` while the only element that accepts a drop is the ~24px header — and moving toward the line erases it.**

`groupDragProps`'s handlers are spread onto `GroupHeader` only; the group body (up to 10 rows plus the three folders and a More row) never calls `preventDefault`, and the rows' own `onDragOver` bails because `dragSession` is null during a group drag.

Hovering the bottom half of an expanded Agent group's header paints the accent line ~340px lower — frequently below the fold of the `overflow-y-auto` scroller — and the moment the pointer moves down toward it the header's `onDragLeave` clears the hint and the line disappears. So “drop after this group” has feedback pointing at a place the user cannot aim at; the drop only ever works while the pointer is still on the header.

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+  groups: readonly T[],
+  keyOf: (group: T) => string,
+  opts: { pinned: ReadonlySet<string>; order: readonly string[] },
+): T[] {
+  const byPin = pinnedFirst(groups, keyOf, opts.pinned);
+  if (opts.order.length === 0) return byPin;
+  const pin = byPin.filter((g) => opts.pinned.has(keyOf(g)));
+  const rest = byPin.filter((g) => !opts.pinned.has(keyOf(g)));
+  return [
+    ...applyManualOrder(pin, keyOf, opts.order),
```
hiyouga: **The “unlisted groups go to the TOP” rule silently overrides `groupSessionsByWorkspace`'s forced-last temp group: once ANY group has been dragged, the merged 临时工作区 group renders FIRST the first time it exists.**

`orderGroups` short-circuits to `pinnedFirst` only when `opts.order.length === 0`; otherwise `applyManualOrder` returns `[...unlisted, ...listed]`.

Project with `/a`, `/b`, `/c` and no temp Sessions; the user drags `/c` above `/a`, so the stored order becomes `['/c','/a','/b']`. Later they start one chat without picking a Workspace, core auto-creates `<agentDir>/workspaces/tmp-1a2b3c4d`, `TEMP_WORKSPACE_GROUP_KEY` is unlisted, and the temp bucket renders above every real Workspace.

The changelog (“Its trailing position remains the default; once it has been dragged, the chosen position wins”) and the module header (“its forced-last position is only the default the automatic sort supplies before any drag”) both hold only for a Project that has never been dragged at all. The temp-group tests only cover the case where the stored order already contains its key.

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+    ...applyManualOrder(rest, keyOf, opts.order),
```
hiyouga: **The same rule inverts `mergeRegisteredWorkspaces`'s documented placement policy, fronting empty registered Workspaces ahead of every group that holds conversations.**

`lib/workspace-registry.ts:166-170` states it explicitly:

> The empty groups sort AFTER the session-derived ones (in registration order, newest first). They hold no conversations, and the list renders behind a 10-group display cap: **fronting them would push every group with real chats behind 更多分组** as soon as a handful of Workspaces were registered. The sidebar widens the cap when it registers one, so a just-added Workspace is still revealed immediately.

For any user who has dragged a group even once, `applyManualOrder` puts the newly registered (unlisted) Workspace first — exactly the outcome that comment says was avoided on purpose. It also invalidates the premise of `addWorkspace`'s `setGroupCap((c) => Math.max(c, workspaceGroups.length + 1))`, which exists only because the new group was assumed to land last.

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * An empty `order` returns the pinned-first list unchanged, which is exactly what the
+ * sidebar rendered before any group was ever dragged.
+ */
+export function orderGroups<T>(
+  groups: readonly T[],
+  keyOf: (group: T) => string,
+  opts: { pinned: ReadonlySet<string>; order: readonly string[] },
+): T[] {
+  const byPin = pinnedFirst(groups, keyOf, opts.pinned);
```
hiyouga: **One shared order array plus `applyManualReorder` front-loading the last-dragged partition means pinning or unpinning a group teleports it the full length of the sidebar — possibly past `groupCap`, out of sight.**

Walked against the real algebra:

1. Auto order `A,B,C,D`. Drag `D` to the top → stored `[D,A,B,C]`.
2. Pin `D` → renders `D,A,B,C`. ✓
3. Drag `C` to the top of the unpinned cluster: `partitionKeys = [A,B,C]`, `seq = [C,A,B]`, `applyManualReorder([D,A,B,C],[C,A,B])` = `[C,A,B,D]` → renders `D,C,A,B`. ✓
4. Unpin `D` → all four listed, `D`'s index is 3 → **`D` jumps from FIRST to LAST**.

With 14 Workspaces and `groupCap = 10` the unpinned group lands past the cap and vanishes from the sidebar until “more groups” is clicked. For rows this quirk is confined to one group's own list; for groups the two partitions ARE the whole sidebar.

packages/web/src/components/ui/group-list.tsx
```diff
@@ -258,9 +270,21 @@ export function GroupHeader({
...
+  onDragEnd?: () => void;
+  onDragOver?: (e: ReactDragEvent) => void;
+  onDragLeave?: () => void;
+  onDrop?: (e: ReactDragEvent) => void;
 }) {
   return (
-    <div className="group/header flex items-center gap-0.5 px-1 pb-0.5">
+    <div
+      className="group/header flex items-center gap-0.5 px-1 pb-0.5"
+      {...(draggable
```
hiyouga, line 289: **`draggable` on the whole header row makes every child button a drag source.**

HTML5 DnD drags the nearest `draggable=true` ancestor of a non-draggable descendant, and none of the children gets `draggable={false}`. The header is almost entirely children: the collapse `<button>` is `min-w-0 flex-1 self-stretch`, followed by up to three `h-7` action buttons (pin / new chat / Agent settings, or the workspace rename-remove overflow).

Pressing the Workspace `+` on a trackpad with 3px of hand movement starts a group drag instead of opening a draft; released over the same header `samePartition` is false (`dragging !== key`), so nothing happens and the button reads as dead. Same for the pin toggle, the gear, and the ellipsis. The group label also stops being selectable.

Separately — and unlike `SessionRow`, which adds `cursor-grab` when draggable — the header keeps its original class string, so nothing on screen says it is a drag source (there is deliberately no sort toggle for groups either, so the feature is invisible until someone presses and moves).

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+      dragging !== null && dragging !== key && pinnedGroups.has(dragging) === pinnedGroups.has(key);
+    /** Which half of the header the pointer is in — the edge the drop would land on. */
+    const edgeOf = (e: ReactDragEvent) => {
+      const rect = e.currentTarget.getBoundingClientRect();
+      return e.clientY - rect.top > rect.height / 2;
+    };
+    return {
+      header: {
+        draggable: true,
+        onDragStart: (e: ReactDragEvent) => {
```
hiyouga, line 694: **`dragGroup` is cleared only by handlers living on the SOURCE header, and it is the sole authorization for a drop — `GROUP_DRAG_MIME` is written and never read — so a drag whose source unmounts leaves the sidebar armed for an unrelated drag.**

If the source header unmounts mid-drag (a websocket `reload()` drops the group from the rendered slice, it falls out of `groupCap`, or `canDrag`/`searching` flips and `groupDragProps` takes its `{ header: {} }` branch, stripping every handler), `dragend` never reaches React and `dragGroup` sticks. Neither the Project-switch effect nor `setGroupMode` resets it, unlike `groupOrder`, which both reset.

From then on every same-pin-partition header computes `samePartition === true`: dragging a Session row — or a file from the desktop — over a group header calls `preventDefault()`, paints a phantom accent line, and on release runs `commitGroupDrop` with the stale key. Checking `e.dataTransfer.types.includes(GROUP_DRAG_MIME)` in `onDragOver`, or a window-level `dragend`, closes it.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+        },
+        onDragOver: (e: ReactDragEvent) => {
+          if (!samePartition) return;
+          e.preventDefault();
+          const after = edgeOf(e);
+          setGroupDropHint((prev) =>
+            prev?.key === key && prev.after === after ? prev : { key, after },
+          );
+        },
+        onDragLeave: () => setGroupDropHint((prev) => (prev?.key === key ? null : prev)),
```
hiyouga: **`onDragLeave` clears the drop hint whenever the pointer merely crosses onto one of the header's OWN children, so the indicator strobes off and on while aiming a drop.**

`dragleave` behaves like `mouseout` — it fires at the element being left with the child as `relatedTarget`, and it bubbles — and the header is almost entirely children. Crossing toggle → pin → new chat clears `groupDropHint` each time, and only the next bubbled `dragover` restores it; with the pointer at rest over a child, Firefox re-fires `dragover` only every ~350ms, so the line is invisible for most of the hover. Each toggle is also a full `Sidebar` re-render.

The repo already has the guard — `features/chat/drop-zone.tsx:108-112`, with the comment “`relatedTarget` is where the drag is going: still inside the region means an internal element boundary was crossed and nothing changed.”

```suggestion
        onDragLeave: (e: ReactDragEvent) => {
          const to = e.relatedTarget;
          if (to instanceof Node && e.currentTarget.contains(to)) return;
          setGroupDropHint((prev) => (prev?.key === key ? null : prev));
        },
```

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+   * ALL of that Agent's Sessions rather than the returned page, so the union of the
+   * folded count keys with the rendered groups (which is where a registered but empty
+   * Workspace comes from) is the whole truth — once every Agent has reported.
+   */
+  const liveGroupKeys = (): ReadonlySet<string> | null => {
+    if (agentsLoading) return null;
+    if (groupMode === "agent") return new Set(agents.map((a) => a.agentId));
+    if (groupMode !== "workspace") return null;
+    if (agents.some((a) => !workspaceCountsByAgent.has(a.agentId))) return null;
+    const live = new Set(workspaceGroupCounts.keys());
```
hiyouga: **The per-Workspace counts this gate trusts as “ALL of that Agent's Sessions” are filtered by the `showCliSessions` preference, so a Workspace used only from the CLI is pruned out of the stored order while the preference is off.**

`state/sessions.tsx` sends `cli: true` only when the pref is on; `SessionService.listSessions` then builds its row set with `listByAgent(projectId, agentId, { webOnly: !includeCli })` **before** computing `workspaceCounts` (`services/session-service.ts:169-171`, `:218-232`). Every Agent has reported, so the `agentsLoading` / `workspaceCountsByAgent.has(…)` gates pass — the report is just filtered.

User turns on “show CLI sessions”, arranges their groups (one exists only because of CLI Sessions), turns it back off; the next drag prunes that key and saves. Turning it back on returns the group as an unlisted newcomer at the TOP, its position gone.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+          e.dataTransfer.effectAllowed = "move" as const;
+          setDragGroup(key);
+        },
+        onDragEnd: () => {
+          setDragGroup(null);
+          setGroupDropHint(null);
+        },
+        onDragOver: (e: ReactDragEvent) => {
+          if (!samePartition) return;
+          e.preventDefault();
```
hiyouga, line 705: **`onDragOver` calls `preventDefault()` but never sets `e.dataTransfer.dropEffect`,** so with `effectAllowed = "move"` a modifier held during the drag resolves the effect to `"none"` and the `drop` event never fires.

Per the DnD processing model the resolved `dropEffect` comes from the user's requested operation filtered through `effectAllowed`; Ctrl requests “copy”, which “move” does not permit, so `dropEffect` becomes `"none"` and only `dragend` runs. The indicator shows normally, the user releases, and nothing is persisted with no explanation.

`features/chat/drop-zone.tsx:100-101` sets it for exactly this reason. (The row drag omits it too, so this is a re-exposure rather than a new class of bug.)

```suggestion
          e.preventDefault();
          e.dataTransfer.dropEffect = "move";
```

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+ * moveInSequence, applyManualReorder) this module reuses rather than restates; what
+ * differs is the key namespace, which modes may be ordered, and that there is no
+ * sort-mode switch:
+ *
+ * - **Implicit mode.** Dragging a group IS the intent — there is no second toggle
+ *   beside the rows' `recent | manual` one. A Project+mode with nothing stored keeps
+ *   the automatic sort untouched (an empty order is the identity), so a fresh profile
+ *   looks exactly as it did.
+ * - **Workspace and Agent modes only.** The time mode's buckets are last day / last
+ *   month / older: a fixed chronological ladder where "put 上个月 above 最近一天" is
```
hiyouga: **Chinese in a code comment in a pure-logic lib module, quoting two labels that exist in neither string dictionary.**

`CONTRIBUTING.md:124-128`:

> **English is the repository's working language** — code, comments, error/log messages, test names and fixtures, package metadata, and developer docs. Chinese appears only where it is the content itself: zh i18n catalogs and fields (`strings.ts` dictionaries, CLI `i18n.ts`, `titleZh`, `short_description_zh`), `*.zh.md` documents, and test literals that assert zh i18n output or exercise CJK-specific behavior.

The line is `a fixed chronological ladder where "put 上个月 above 最近一天" is not an order anyone can mean`. `group-order.ts` is none of the permitted cases, and the pre-existing Chinese in `sidebar.tsx` all quotes real zh dictionary values verbatim — which this cannot ride: `strings.ts:1093-1095` ships `近一天` / `近一月` / `更早` and `strings-en.ts:1133-1135` ships `Past day` / `Past month` / `Earlier`. The PR description already writes the same sentence in English.

changelog/unreleased/2026-08-24-sidebar-group-order.md
```diff
@@ -0,0 +1,42 @@
+# Drag the sidebar's groups into your own order
+
+- **Date:** 2026-08-24
+- **Type:** feature
+- **Scope:** `web`
```
hiyouga, line 5: **A user-facing ordering axis ships with no docs-site update** — hence `Scope: web` alone here.

`CONTRIBUTING.md:157-158`:

> New user-facing behavior should come with tests, and with docs updates when it changes documented behavior (README, docs site).

`packages/docs/content/web-app.en.md:44` documents the sidebar's ordering as one axis driven by one menu — “Grouping and sort order are switched in the **List options** menu … a clock for most-recent order, and up/down arrows for the drag-reordered manual order” — and `web-app.zh.md:44` mirrors it. After this PR there is a second drag-ordering axis with no List-options entry by design, so a reader following the docs cannot discover it.

Precedent: `changelog/unreleased/2026-08-23-trace-one-entry-point.md` carries `Scope: web, docs` and a `## Docs` section for exactly this.

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * The grouping modes whose groups can be dragged into a manual order. "time" is
+ * deliberately absent — see the header.
+ */
+export const ORDERABLE_GROUP_MODES = ["workspace", "agent"] as const;
+export type OrderableGroupMode = (typeof ORDERABLE_GROUP_MODES)[number];
+
+/** Whether a grouping mode's groups may be reordered at all (the one gate; time mode is excluded). */
+export function isOrderableGroupMode(mode: GroupMode): mode is OrderableGroupMode {
+  return mode === "workspace" || mode === "agent";
```
hiyouga: **`isOrderableGroupMode` hardcodes the two mode strings instead of deriving from `ORDERABLE_GROUP_MODES`, so the exported list and “the single gate” can drift apart.**

Adding a third orderable mode to the array widens `OrderableGroupMode` (so `groupOrderKey` and `saveGroupOrder`'s signature accept it) while this predicate keeps rejecting it — `loadGroupOrder`/`saveGroupOrder` would silently no-op for a mode the type system says is orderable, with no compile error. The module header calls this predicate “the single gate … so the exclusion cannot be undone by forgetting a check at one call site”, which is exactly the invariant the duplication breaks. (`group-order.test.ts:48-53` asserts the literal list against itself, so it cannot catch the drift.)

```suggestion
  return (ORDERABLE_GROUP_MODES as readonly GroupMode[]).includes(mode);
```

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+ *    order) — it decides where groups with no stored place sit, and IS the whole
+ *    answer while nothing is stored;
+ * 2. the pinned cluster first (pinnedFirst), then
+ * 3. the stored order within the pinned cluster and within the rest independently, so
+ *    a drag can never move a group across the pin boundary.
+ *
+ * An empty `order` returns the pinned-first list unchanged, which is exactly what the
+ * sidebar rendered before any group was ever dragged.
+ */
+export function orderGroups<T>(
```
hiyouga, line 129: **This module is a second copy of machinery `session-order.ts` already owns.**

- `orderGroups(g, keyOf, {pinned, order})` is output-identical to `orderSessionRows(g, keyOf, {pinned, sortMode: "manual", order})` (fuzz-checked over 20k random pin/order/stale-key combinations, 0 mismatches). It also runs `pinnedFirst` and then re-splits the same list with two filters — `pinnedFirst` preserves input order within each partition, so on the manual path its concat is undone immediately.
- `groupDragProps` + `commitGroupDrop` restate `renderRows`'s drag object + `commitManualDrop` handler for handler, and `edgeOf` is a third copy of `e.clientY - rect.top > rect.height / 2` (still inlined twice in the row path).
- `GroupBlock`'s indicator repeats `SessionRow`'s class string exactly, differing only in the edge offsets.
- `loadGroupOrder`/`saveGroupOrder` are the sixth copy of the same localStorage body (`session-order.ts`, `pinned-sessions.ts`, `sidebar.tsx`'s `loadGroupSet`/`saveGroupSet`, `workspace-registry.ts`, `model-group-expansion.ts`).

Every future change to the drag contract — the Firefox private-MIME rule, the hint dedupe, the pin-boundary guard, the drop-edge threshold — now has to be found in two or three places.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+          setGroupDropHint((prev) =>
+            prev?.key === key && prev.after === after ? prev : { key, after },
+          );
+        },
+        onDragLeave: () => setGroupDropHint((prev) => (prev?.key === key ? null : prev)),
+        onDrop: (e: ReactDragEvent) => {
+          if (!samePartition || dragging === null) return;
+          e.preventDefault();
+          const after = edgeOf(e);
+          const partitionKeys = sequence.filter(
```
hiyouga: **A drop commits the WHOLE pin partition, so the very first drag writes a stored index for EVERY group of the mode — and `orderGroups` has no escape hatch, so the automatic sort is dead from then on for groups the user never touched.**

`sequence` here is the full mode sequence, so `partitionKeys` is every unpinned key; `commitGroupDrop` then stores all of them via `applyManualReorder`. Once every rendered group is *listed*, `applyManualOrder` sorts purely by stored index and the recency sort is bypassed entirely.

Workspace mode, 12 Workspaces sorted by newest conversation. The user drags one group up one slot. From that moment, starting a conversation in the Workspace sitting 11th no longer floats it to the top — and never will again. They asked to move one folder and silently turned recency ordering off for the whole sidebar.

`orderSessionRows` survives the identical `applyManualReorder`-the-whole-partition design only because it is gated behind an opt-in `recent | manual` toggle (`if (opts.sortMode !== "manual") return byPin`). The group axis took the algebra and left the gate behind.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -374,6 +384,16 @@ export function Sidebar({
...
     loadSessionOrder(currentProjectId, initialGroupMode()),
   );
+  /**
+   * Manual GROUP order (Workspace keys / Agent ids), persisted per Project and grouping
+   * mode like the row order. Independent of `sortMode`: dragging a group is itself the
+   * intent, so there is no second toggle — an empty array is the identity and the list
+   * keeps its automatic sort. Empty in time mode, whose buckets are chronological
+   * (group-order.ts).
+   */
+  const [groupOrder, setGroupOrder] = useState<readonly string[]>(() =>
```
hiyouga, line 402: **There is no way to clear a group order.** `setGroupOrder` is only ever called with `loadGroupOrder(...)` (mount / Project switch / mode switch) or with the committed array from `commitGroupDrop` — never with `[]`. The list-settings 排序方式 radios call `setSortMode`, which touches rows only.

Read together with the previous comment this is the sharp edge: the header is the drag handle, so a slightly-dragged press on the pin button can drop a group somewhere unintended; that single accident freezes the mode's automatic sort permanently, and the one control a user would reach for to undo it (最近更新) does nothing to groups. Recovery is deleting `penguin.groupOrder.<projectId>.<mode>` from DevTools.

`session-order.ts` documents the row escape explicitly — “Leaving manual KEEPS the stored order — toggling back restores it.” The group axis shipped without one. Worth deciding deliberately: either a reset affordance, or a commit that stores only what was actually dragged.

packages/web/src/components/layout/sidebar.tsx
```diff
@@ -583,6 +618,115 @@ export function Sidebar({
...
+   * rewritten anyway, so no separate destruction event has to be listened for.
+   */
+  const commitGroupDrop = (sequence: readonly string[], targetKey: string, after: boolean) => {
+    if (dragGroup === null) return;
+    const seq = moveInSequence(sequence, dragGroup, targetKey, after);
+    // Identity guard, as for rows: a drop that changes nothing must not persist a fresh array.
+    if (seq === sequence) return;
+    const reordered = applyManualReorder(groupOrder, seq);
+    const live = liveGroupKeys();
+    const next = live === null ? reordered : pruneGroupOrder(reordered, live);
```
hiyouga: **Pruning lives only here, at the drop — so the one explicit “destroy this group” action in the UI never prunes.**

`confirmDeleteWorkspace` (从侧边栏移除, reached from `GroupOverflowMenu` in this very header) unregisters a Workspace and stops there. Compare the row-order counterpart in `confirmDeleteSession`, which prunes the pin, the `sessionOrder` entry and the seen marker at the moment of destruction.

User drags `/a` to the bottom, removes it from the sidebar, and never drops again — nothing prunes. Weeks later they re-add `/a` via 新建工作区: `addWorkspace` widens `groupCap` precisely so the fresh pick is visible, but `orderGroups` finds `/a` still listed and drops it back at its old buried position instead of the documented newcomer-to-the-top treatment. The cap widening looks like it did nothing.

(This is one symptom of a larger problem with drop-time pruning — see the comments on `liveGroupKeys()` and on the `showCliSessions`-filtered counts. The completeness proof the prune rests on does not hold, so the honest fix is probably to stop pruning rather than to prove deadness harder.)

packages/web/src/components/layout/sidebar.tsx
hiyouga, line 2353: **`top-1` and `-bottom-1` are not symmetric, so the one gap between two adjacent groups is drawn at two different heights depending on which header the pointer is over.**

The gap between A and B is B's `pt-2.5` (10px). Hovering the bottom half of A's header draws A's “below” line at `bottom: -4px`; hovering the top half of B's header draws B's “above” line at `top: 4px`. Both denote the identical insertion point, so the accent line visibly jumps ~2px as the pointer crosses between the two headers while the pending drop has not changed at all.

`SessionRow`'s indicator deliberately pairs `-top-px` with `-bottom-px` so its two edges coincide. The group copy pulled the “above” edge inside the block (presumably so the first group's line is not clipped) and did not compensate on the “below” side.

packages/web/test/group-order.test.ts
```diff
@@ -0,0 +1,332 @@
...
+
+  it("its key round-trips through storage like any other group key", () => {
+    const s = memStorage();
+    saveGroupOrder("p1", "workspace", [TEMP_WORKSPACE_GROUP_KEY, "/home/a"], s);
+    expect(loadGroupOrder("p1", "workspace", s)).toEqual([TEMP_WORKSPACE_GROUP_KEY, "/home/a"]);
+  });
+});
+
+describe("drop commit (the sequence algebra shared with the rows)", () => {
+  it("a drop commits the FULL group list, so groups past the display cap keep their places", () => {
```
hiyouga, line 294: **Every behavioural decision of this feature is untested; only the trivially pure module is.**

This test — “a drop commits the FULL group list, so groups past the display cap keep their places” — builds the 15-key list itself and never touches the display cap, the sidebar, or anything in group-order.ts. The cap property lives entirely in *which array* `groupDragProps` is handed (`agentGroupSequence` / `workspaceGroupSequence`). Change either memo to `.slice(0, groupCap)` — precisely the bug the comment above it describes — and this test still passes.

More broadly nothing in the suite reaches `groupDragProps`, `commitGroupDrop`, `liveGroupKeys` or `GroupBlock`, and `packages/web/e2e` has no spec for the gesture. The pure algebra has 28 tests; the decisions that can actually be got wrong have none.

packages/web/src/lib/group-order.ts
```diff
@@ -0,0 +1,148 @@
...
+
+/** Minimal storage interface, shared with session-order.ts; tests inject an in-memory implementation. */
+export type GroupOrderStorage = SessionOrderStorage;
+
+/**
+ * Storage key of one Project's manual GROUP order (sidebar key-naming convention,
+ * `penguin.sessionOrder.<projectId>.<mode>` &c.), scoped by grouping mode for the same
+ * reason the row order is: the modes have unrelated group lists.
+ */
+export const groupOrderKey = (projectId: string, mode: OrderableGroupMode): string =>
```
hiyouga, line 87: **`groupOrderKey` scopes the arrangement by Project only, so two accounts sharing a browser profile share one group order.** Worth deciding rather than inheriting — this is a brand-new key, so it is the cheapest moment it will ever be.

The repo already user-scopes the comparable personal sidebar state — `penguin.chatDrafts.${userId}.${projectId}`, `penguin.memoryCollapsed.${userId}.${projectId}.${agentId}` — and `e2e/draft-user-scope.spec.mjs` exists to keep it that way. A manual group order is the same kind of state: one person's arrangement of a Project several people belong to. On a shared or kiosk browser, user B logging in after user A inherits A's ordering, with nothing on screen explaining it.

The counter-argument is real: the sibling `penguin.sessionOrder.${projectId}.${mode}` is already un-scoped and shipped, so user-scoping only the group axis makes the two inconsistent. Either answer is defensible — but pick one deliberately, and if it stays per-Project, say so where the key is defined.

PR #443, opened 2026-08-24
feat(cli,server): terminal sign-in, signed sessions, and no auth secret on disk

Myriad-Dreamin:
There was no way to sign in to a PenguinHarness server from a terminal. The CLI works on the data root directly — `config`, `run` and `chat` all open files, never a socket — so nothing in it had ever needed to be a client, and "log in" had no answer beyond the browser's login page. The account bootstrap around it is rebuilt in the same pass: a fresh server prints a first-login link instead of a fixed password, and a forgotten admin password is recovered offline.

```bash
penguin auth login                      # password, against the server on this data root
penguin auth login --server https://penguin.example --user-id alice
penguin auth status
penguin auth logout
penguin auth token                      # no password: writes a session row on this machine
```

## Two ways in

**`login`** asks a *running* server for a session, exactly as the login page does. The target defaults to the server on this data root, so signing in to your own needs no URL — and it is read from a *live* lock (PID + port), so a password never goes to a stale lock's port that another process may now hold. Interactively it asks for the account first and names it in the password prompt; typing one account's password at another's is a mistake the prompt should prevent. Supply the password non-interactively (`PENGUIN_PASSWORD`, or `--password`) and neither question is asked: that is a script, and a script cannot answer one. Prefer the environment variable — a command line is world-readable through `ps`.

**`token`** takes no password at all. A session is a row in `web.db`, so it opens the database and inserts one: reading the data root already reaches every credential such a token could, so the write adds no authority that being able to read the root did not already carry. It needs no running server, is safe alongside one (WAL + busy timeout), and is for where there is no password to give — a machine whose admin password somebody set by hand, or a script that must not carry one. Bare output for `TOKEN=$(penguin auth token)`; `--mark` prefixes a marker line for a caller parsing it out of a shell that may print a banner.

The CLI session is remembered at `<root>/cli-session.json`, mode 0600 on create and overwrite, through a symlink-refusing write. `logout` ends it on the server, not only locally.

## First login and admin recovery

A fresh server's seed password is 24 base64url characters (144 bits), generated, hashed and discarded unseen — unguessable at the login endpoint from the moment the account exists, and read by nobody. Every start prints a sign-in link until a password is set.

The link carries a `setup` session: one that may set a password without knowing the old one (there is nothing to type — the current password is random and was never shown), and that opens no desktop-only route, because reading this boot's console proves someone reached the terminal, not that they own the machine the desktop shell runs on. It is redeemed at `GET /api/auth/claim`, which also serves the desktop shell's one-shot token. It is deliberately reusable *until claimed* — a link a mail client or browser may prefetch must not be spent before its reader opens it — and it dies the moment any password is set, through whichever door set it. A claimed server mints no such session at all.

`penguin server reset-admin-password` returns the admin to that unclaimed state offline (server stopped), deleting its sessions and producing no plaintext; the next start prints a fresh link.

## Sessions stay rows

A session is a 32-byte random token in the cookie, stored as its sha256 in `auth_sessions`. The row **is** the session: logout deletes it, an admin password reset deletes that user's rows, and a session survives a restart because it is on disk. Validity is **30 days** (previously 7), with sliding renewal topped up IN PLACE — the cookie value never changes, so no renewal creates a second identity for a revocation path to miss. Only a session whose own span reaches the renewal window slides: an hour-long `cli` token expires at its hour rather than stretching by being used.

This is worth stating because the branch went somewhere else first. An interim design made sessions signed tokens (`v1.<claims>.<hmac>`) with a memory-only key, to save the one indexed read per request. It was reverted in `5275059`: that read is a microsecond in a single-process SQLite the auth path never bottlenecks on, and the scheme was the source of most of this PR's bugs across three review rounds — renewal minting a second identity, revocation having to chase it across expiry, an owner-token file needed for bootstrap. `token-codec.ts`, `owner-token.ts`, the `auth_revocations` repo and `POST /api/auth/owner` are all gone; the row model is the conservative shape, chosen for stability rather than for the microsecond.

## Compatibility

The fixed seed password retires (`PENGUIN_SEED_ADMIN_PASSWORD` unset used to print `penguin-<4 digits>`): claim the account through the printed first-login link instead. A data root carried over from an earlier build has any `initial-admin-password` plaintext deleted at the next start.

Sessions in a v0.2.0 `web.db` are kept — `auth_sessions` gains its `via` column on the next start and existing rows read as ordinary password sessions — so an upgrade from a release does not sign anyone out. (Sessions issued by an unreleased build carrying the interim signed-token scheme are not rows at all, and simply stop working.)

The `Secure` flag on session cookies now requires the deployment's opt-in. `x-forwarded-proto` used to be believed unconditionally, which let anyone reaching a plain-HTTP port force a `Secure` cookie the browser then never sends back over that connection — a sign-in that never takes. An HTTPS deployment behind a reverse proxy sets `PENGUIN_TRUST_PROXY=1` (and keeps forwarding the header) to keep its cookies `Secure`; left unset, they are issued without the flag.

`penguin server auth-token` is now `penguin auth token`; `GET /api/auth/desktop-login` is now `GET /api/auth/claim`.

## Merged with main

main grew a local API token (`<root>/api-token`, `Authorization: Bearer` → the built-in admin) while this branch was rewriting the same auth surface, so `5f09771a` merges the two. Merged rather than rebased: rebasing 27 commits would have meant resolving main's changes against the interim signed-session design that does not survive to the tip.

- `SessionVia` is the union of both, `"password" | "desktop" | "setup" | "token"` — unrelated additions, one a stored row and the other never stored at all. `issueSession` narrows to `SessionViaValue`, the column's own type, so the per-request kind cannot reach a row.
- The boot's API token moves from a private field on `AuthService` to `AuthRuntimeState`. On main the service was a runtime singleton and could hold it; here auth is built per App, so a push would drop a token whose value is still on disk and still in every command subprocess's environment — exactly the "outlives a push, dies at a restart" rule that layer exists for. `claimRuntimeCapabilities` fills fields a runtime older than the platform never published, in place: copying the bag would strand the App's writes to it.
- `constantTimeEqual`, added here for the link comparison, gives way to main's `tokensEqual`, which digests both sides first and so leaks neither content nor length.

## Verification

`pnpm typecheck`, `pnpm test` (server 1307, web 1476, core 1021, cli 383, docs 53, landing 69, skills 27) and `pnpm format:check` — green.

The auth-focused suites: restart-surviving sessions and in-place renewal (including that an hour-long token does not slide), an admin password reset deleting that user's rows, first-login lifecycle (claim, reuse-until-claimed, death on password set through either door, no mint on a claimed server's restart, the link is actually printed), `penguin auth token`'s direct-write path incl. its behaviour alongside a running server, throttling without an account-existence oracle, CLI session-file modes on create and overwrite, and the local API token's Bearer path (no silent fallback to a cookie when the header is wrong).

An external agent review (codex) audited the branch; its findings are fixed in `2708d660` — the web change-password dialog honors `setup` sessions, `changePassword` on the admin ends a live link, `x-forwarded-proto` is gated behind `PENGUIN_TRUST_PROXY` with the operator docs to match, refusal text from a `--server` target is neutralized before it reaches the terminal, and a successful password set clears the login throttle so the claim flow cannot be 429'd after the password already committed.

---

First of a series splitting `feat/remote-install` into reviewable PRs. It is first because the machines work builds on the session minting from here.


hiyouga (approved):

hiyouga:
Security review of this PR (full diff plus the surrounding auth/routing/preview-isolation code at both refs). Three findings survived verification; **fixes for all three are pushed to this branch as `2708d660`**, so this comment is the record of what and why.

### 1. Medium — proxied HTTPS deployments silently lose the session cookie's `Secure` flag

`packages/server/src/auth/middleware.ts` (`cookieOptions`) now honors `x-forwarded-proto` only under `config.trustProxy`. The gate itself is right — on main the header was believed unconditionally (`packages/server/src/http/routes/auth.ts`), so anyone reaching a plain-HTTP bind could force a `Secure` cookie the browser then never sends back: a sign-in that never takes. But every operator-facing text still said the header alone was enough: the server README's reverse-proxy note ("forward `x-forwarded-proto` to enable Secure cookies"), the `PENGUIN_TRUST_PROXY` row (described as hot-update-gate only), the docs env table (no `PENGUIN_TRUST_PROXY` row at all), and this PR's changelog Compatibility section (silent on it).

Scenario: an HTTPS deployment that followed the README upgrades to this build; its session cookies silently stop being `Secure`; an active MITM induces any `http://same-host` request from a signed-in browser (an image, a typed hostname) and captures a 30-day sliding admin session in cleartext — exactly what `Secure` exists to prevent.

Fixed by documentation, since the code is correct: README (env row + proxy note), `packages/docs/content/configuration.{en,zh}.md` table row, the `config.ts` comment, and a new paragraph in this entry's Compatibility section (both languages).

### 2. Low — `penguin auth login` prints server-chosen bytes to the terminal unfiltered

`packages/cli/src/commands/auth.ts` (`serverSaid`): the JSON path returned `error.message` verbatim, and the raw path collapsed only `\s+` — ESC (0x1b) is not whitespace, so both paths deliver escape sequences to the operator's terminal. `--server` accepts any URL, so the refusal text is attacker-chosen: a hostile or MITM'd target can retitle the window, write the clipboard via OSC 52, or garble the screen of the machine signing in.

Fixed: `neutralLine` collapses whitespace, renders C0/DEL in caret form (the same convention `tool-render.ts` already uses for untrusted terminal-bound text) and drops C1; tests cover the OSC/CSI cases from both paths.

### 3. Low — the claim flow's completing sign-in can be 429'd from outside, after the password already committed

`packages/server/src/http/routes/me.ts:66-79`: the setup-session path first sets the password (deleting **every** setup session, this request's included — correctly), then calls `authService.login()` for the replacement cookie. That login runs under the per-user throttle, and anyone who can reach the port can hold the `admin` backoff window open indefinitely with junk `POST /api/auth/login` (at the cap the window is essentially always active). During such a spree the claimer's password set returns 429 **after the change committed**, with their setup session gone: failure reported for a change that took, signed out mid-bootstrap, and they must out-race the guesser to sign in with a password the dialog just told them failed.

Fixed in `AuthService.setPassword`: a successful password set clears the user's failure counter before the re-login — it proves at least what the successful login that already resets the counter proves (the old password, or a desktop/setup session), and it is unreachable without that proof. Regression test with a frozen clock (fails without the fix; mutation-checked).

### Verified clean

The load-bearing new surfaces held up: `/api/auth/claim` compares the printed token constant-time against the process value only (no cookie-out-of-any-valid-token), keeps refusals indistinguishable, tries the desktop one-shot first (sha256 + `timingSafeEqual`, consumed only on match), mints only while unclaimed, and every password-setting door kills all setup sessions including earlier boots'; the preview host still refuses `/api`, so no claim cookie can land on the preview origin; the request log records path only (no query string) and the errors table records only the error, so the link token stays out of both; `penguin auth token`'s FS-ownership model is coherent, with the TTL clamped at the single mint point and short tokens excluded from sliding renewal; the CLI session file is 0600 on create and overwrite with a symlink-refusing write; login still issues fresh tokens (no fixation), logout stays behind the content-type CSRF gate, and no interim-design leftovers (`owner-token`, `/api/auth/owner`, token codec) remain.

Two notes, not findings: (a) pre-existing on main and untouched here, `login()` skips scrypt for unknown userIds (`auth/service.ts`, `row !== null && await verifyPassword(...)`), so response timing still distinguishes existing accounts even though the throttle deliberately doesn't — worth a follow-up since the README bullet advertises "not an account-existence oracle"; (b) the design-repo sync for this redesign (05-ARCHITECTURE still describes the `penguin-1234` seed and 7-day sessions) looks pending — the one clause fix 3 touches is covered by Prism-Shadow/penguin-harness-design#63.

Verification for `2708d660`: server 996/996, cli 310/310, docs 53/53, landing 61/61, `pnpm typecheck` and `format:check` clean.

hiyouga (approved):

PR #446, opened 2026-08-24
feat(core,cli,desktop): one well-known server port

Myriad-Dreamin:
The server, the CLI and the desktop shell each had their own idea of what port to use, and the shell negotiated an **ephemeral one per launch**. `7364` is now an *address* rather than a default: a client finds "this user's server" by probing one known number, and can offer to install one when nothing answers.

**The number itself does not change** — what changes is that it is fixed.

## What follows

- The desktop shell's embedded server binds `DEFAULT_SERVER_PORT` instead of taking whatever the OS hands out. `--port` / `PORT` still override everywhere.
- **The remembered-port machinery is gone.** It existed to keep the app origin stable across launches — and with it the renderer's origin-scoped `localStorage` and cookies. A fixed port gives that by construction, so the memory was a mechanism maintaining what the address now guarantees.
- A port already in use is no longer routed around. The shell still attaches to a live server on the same data root before starting one, so this only bites when something *else* owns 7364 — worth an error rather than a silently different origin, since a different origin is a different set of cookies and a window that looks logged out for no reason anybody can see.

## What it costs

One server per user per machine by default; a second instance needs an explicit port. That is the trade the fixed address buys, and it is the same constraint an SSH tunnel imposes anyway — the local port has to equal the remote one, because preview URLs are built from the server's own bound port.

## Verification

`pnpm typecheck`, `pnpm test`, `pnpm format:check` — green.

Note: on the source branch this was two commits — one that moved the number to 7376 and one that moved it back to 7364. They are squashed here, so the diff is the semantic change alone rather than a number that churns and returns. The changelog entry also gained its missing metadata block and Chinese counterpart.

---

Independent of the rest of the `feat/remote-install` series. Split out to land on its own.


Myriad-Dreamin:
Closing: the case for this does not survive reading the code.

The stated rationale — a client finds this user's server by probing one known number — is
not a mechanism that exists. Every discovery site reads `server.lock` and verifies it live:
`cli/serve.ts`, `desktop/main.ts:224` (which already attaches to `existing.port`),
`server/index.ts`'s sole-instance check, `reset-admin-password.ts`, and the CLI's auth
commands. `DEFAULT_SERVER_PORT` is only ever a bind default (`config.ts`'s
`PORT || DEFAULT_SERVER_PORT`), never a discovery probe. The claim was also in direct
contradiction with the lock read a few lines away in the file this PR edits.

The tunnel argument is real but narrower than stated: preview URLs are built from the
server's own bound port, so both ends of an `ssh -L` must carry the SAME number — which a
discovered port satisfies just as well as a fixed one. `machines/connect-state.ts`'s
`pickTunnelPort` demonstrates exactly that: it prefers a remembered port and treats
`DEFAULT_SERVER_PORT` as a search START, never as an assumption.

What is left is one optional simplification (deleting `port-memory.ts` and its test, −156
lines) bought with a real failure mode this PR names itself: a hard error when something
else owns 7364, where today the shell falls back to an ephemeral port and still starts.

Nothing depends on this branch: `DEFAULT_SERVER_PORT` already exists on main, and the
machines stack is built on main's remembered-port behaviour. If the deletion is still
wanted later, the safe shape is DEFAULT_SERVER_PORT as the *preferred* port with the
existing PORT=0 fallback — simpler than main, without the regression.

PR #448, opened 2026-08-24
feat(machines): the Machines page — use a machine, stop using it, declare its ssh host, and open a workspace on it

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← #808 ← #718 ← #544 ← **#448** ← #450 ← #577 ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

## The Machines page: use a machine, stop using it, and open a workspace on it

> **Not reachable when this merges.** The sidebar row, the route and the picker's machine row live in #577, which stays the gate on top of #450 until the feature has settled.

The base of the Machines line: the page that drives all of it, and machines in the workspace picker, so a Session can start in a directory on another machine. A machine is something this Project runs agents on — **Use** one and this server installs the program, starts it, connects, hands over the Model config and stays connected; **Stop using** it and the machine is let go. What sits above it: #450 runs an agent on a machine, #577 opens the gate.

> **Folded in as real commits, not recorded merges** — the page lands as one PR, so its follow-ups belong here rather than above #450: #622 (the two verbs, the cards, the batch job runner — 16 commits) with #624 (an ssh host declared from the page — 5), #629 (the glyph-only selection bar and the server root on a card — 1), and #617 (the pre-migration machines table and the failed job's way out — 2). Cherry-picked in chain order onto `feat/machines-identity`; three shallow conflicts in the first #622 commit (the `Machines` interface's member list and the machines strings, where #450's old base had edited the same lines) resolved to #622's side. The closed PRs point here; their branches stay as they were. One changelog entry, `2026-09-09-machines-page`, with these as its sections.

### The window never moves

A call that names a machine is rewritten to `/server/<machineId>/api/…`, which the local server forwards (#574's proxy). There is deliberately **no window-wide "active server" mode**. One existed and was removed: pointing an entire window at another machine put every page behind a tunnel, `/api/me` included — so a tunnel that dropped left the app unable to answer whether anyone was logged in, and the control for getting back was rendered inside the layout that never mounted. An escape hatch behind the thing that breaks is not an escape hatch.

Naming the machine on the calls that actually concern it has no such failure: a dead tunnel breaks exactly the request that needed it, and says so. One string rule (`lib/server-context.ts`) covers `fetch` and the SSE subscriptions alike.

### Two verbs, one sentence per machine (was #622)

**Use** is one job — check, install or update, hand over, restart, connect, sync — and **Stop using** lets the machine go; install, update, restart, connect and disconnect stop being controls a person orders by hand. `POST …/machines/use` takes `{ machines: [...] }` (a few at once, the rest queued, `202`; refusals that need no ssh come back by id), `POST …/machines/stop-using` the reverse; the list carries `jobs` with the pipeline `phase` each is on. The per-machine routes remain.

The page is one card per machine, this server first, sorted by name so an update or a probe never moves one. A card at rest is the name and one sentence ("Connected and ready", "Its server is not running — Use starts it"), every sentence that names a problem fixed by the same button; the dot is the link blue for a held connection, amber and red for what needs a person, grey for settled. Selection is the card; a queued or working card grows a stepper; the chevron unfolds the build, dates, state and port, machine id, the job's full output and the forced install when the job offers it. Machines not yet in use live behind **Add machines…**, a search over the server's ssh config — the first matches, a fold for the rest, a confirmation once something is picked. **Update all** appears in the header when any machine in use is behind this server's build. Staying connected is the server's: a machine that drops is retried on a widening wait, a minute up to fifteen, until it is held again or someone stops using it.

### An ssh host declared from the page (was #624)

The **+** at the foot of **Add machines…** opens a short form — alias, address, optionally user, port and key file — and the server appends the `Host` block to its own `~/.ssh/config`, led by a comment naming PenguinHarness and the time (`machines/ssh-config.ts`: `validateHostEntry`, `renderHostBlock`; `transport/targets.ts`: `appendHostBlock`). One word per value, no `#`, no pattern alias, a real port; an alias the config already declares is refused, since ssh would take the earlier block. A block this app wrote can be reconfigured from the gear in a card's details (`findHostBlock` / `replaceHostBlock`, the marker comment is what makes it ours); a hand-written block is shown, not saved. `POST …/machines/ssh-hosts` (`201` / `400 ssh_host_invalid` / `409 ssh_host_exists`), `GET` and `PUT …/ssh-hosts/:alias` (`409 ssh_host_foreign`). Admin only.

### The bar, and the server root (was #629)

Select all, select none, use and stop using are glyphs alone — the box, the empty box, the plug, the plug pulled — with the word as tooltip and accessible name; verbs that stand alone (configure, new, force install) keep theirs. A card's details name the machine's **server root** (`PENGUIN_HOME`): the profile decides it, so a dev instance reaches a machine's dev installation and never the release one beside it, and two instances of this page can now say which server they mean.

### A machines table from before its migration (was #617)

A data root that ran the machines line before it was released had its machines tables adopted with the columns they had (migration 4 creates them `IF NOT EXISTS`), so a connect failed with `table machines has no column named session_pid`. Migration 10, `machines-columns`, adds `session_pid` and `platform` where they are missing; additive and swap-safe, so a hot push applies it. And every failed install or connect offers installing the program anyway, except the run that was that install and the one failure installing cannot mend (`canReplaceProgram: false` — this server has no build to send).

### Probing on a widening schedule

The page's own probe widens as answers stop changing — 15s, 30s, 45s, a minute, two, three… up to ten — and **any change resets it to the first step**. `checkedAt` is deliberately not part of the fingerprint: every probe moves it, so including it would pin the interval at 15 seconds forever. The schedule lives in the page, not the server: a timer over there would keep spawning ssh long after the last tab closed.

### Picking a workspace on a machine

A workspace is a directory **on** a machine, so only one whose filesystem is reachable right now can be browsed: this one always, any other with a held connection. Every machine in use is **listed regardless**, the unreachable ones disabled with the reason at the row — a list that silently omits its answer is indistinguishable from a broken feature. Identified by machine id, labelled by ssh alias, so a renamed host keeps its workspaces.

### A Project's plugin list reaches its machines, and what the page says in an outage (was #627's machine half, and #633)

The plugin list a Project owns (#383) is kept the same on the machines that Project uses: `machines/plugins-sync.ts` pushes the closure to every held machine with strict parity, and a machine that answers differently is reported rather than papered over. Two things the Machines page said during an outage were not true and are fixed here: a machine whose link was up but whose server had stopped read as connected, and a sync that had not happened read as done.

### The way in is not here

`main` today has the page component with no route, and keeps `machines` in `nav-group-collapse.ts`'s `UNRELEASED_NAV_KEYS` so the sidebar never offers the row. This PR leaves all of that as it is. The three lines that open it — lifting the gate, the `/machines` route, and the sidebar picker's machine row — are #577, to be reopened on top of #450 and merged on its own date.

### Verification

After the fold (Node 24): `pnpm -r build`, `pnpm typecheck` (0 errors), `pnpm lint`, `pnpm format:check` — clean. Web **126 files / 1677 tests** pass (`--maxWorkers=2`), including `machines-view`, `workspace-machines`, `machines-probe-schedule` and `server-context`; the server's machines suites run here — `machines-api`, `db-migrations`, `ssh-config-hosts` — **3 files / 105 tests** pass. The full server suite is CI's (it outgrows this box's memory in one process).


Myriad-Dreamin:
The machines stack is regrouped into two PRs. This one now carries what were #448, #449 and #451, rebased onto main (#447 merged as `c6d7633e`, so the 26 auth commits and 33 install commits below this work are gone from the branch — main has them squashed).

**Also here, moved down from #450:** `29935c79`, `36d9e053` and `33cf5b09` — "sign in to a machine", the ssh auto sign-in that creates `machines/signin.ts`, and the far-side-home fix that follows it.

They had to move. #451's first commit builds the model sync on `signInOnRemote`:

```ts
const signedIn = await this.#effects.signIn({ target });
if (signedIn.kind !== "signed-in") { say(`Models not synced — ${signedIn.detail}`); return; }
const cookie = signedIn.setCookie.map(...)   // the credential the sync pushes models with
```

`signin.ts` was created in #450, so project-config could not sit below run-agent without it. The three commits are unmodified — only their PR changed, and they belong here anyway: signing in to a machine is the primitive, not part of running an agent on one.

31 commits. `pnpm typecheck`, `pnpm format:check`, and `pnpm test` (server 1593, web 1587, core 1021, cli 383, docs 53, landing 69, skills 27) green.

#449 and #451 are closed in favour of this one. #450 continues on top of it with the rest of run-agent plus what was #452.

Two conflict resolutions worth a look:

- `packages/web/src/api/endpoints.ts` and `packages/server/src/app.ts` — main's new mounts (the model-OAuth callback, `/api/install`) and this stack's additions are independent; both kept.
- `packages/server/src/machines/service.ts` — the effects seam gained `signIn` and `loadConfig` from two different PRs' worth of work; both land here now.

hiyouga (approved):

hiyouga (approved):

PR #449, opened 2026-08-24
feat(server,web): reach a machine and work in it

Myriad-Dreamin:
> **Stacked on #448.** Review the diff against `feat/machines-identity`.

Connect to a machine, hold a tunnel to it, and pick a workspace on it. These arrived as three arcs on the source branch — the transport, the picker, and the ssh performance work — and are one PR because **none of them is usable alone**: perf is invisible, a tunnel with nothing to do with it is not a feature, and the picker cannot reach a machine.

## Connect

- Bring that machine's server up if it is down, forward its port, and proxy its API at `/server/<machineId>/…`. **Local and remote port are the same number by design**: preview URLs are built from the server's own bound port, so the two have to match.
- Addressed by **machine id, not ssh alias** — an alias is an address that can be re-pointed at a different host, and a proxy keyed on one would silently follow it.
- Cookies are namespaced per machine (`penguin_s_<hex>_…`), so a session on one is not a session on another.
- The tunnel is **unsupervised on purpose**. A dropped link or a rebooted machine is exactly what the person needs to see; a silent reconnect would hide it.
- Auto-connect on boot, because a tunnel is plumbing — having to establish it by hand before a machine will answer is a chore, not a decision.
- A push here becomes a push everywhere: an App booting is what a hot update produces, so it hands the same build to any machine carrying a different one. It **skips a busy machine rather than queueing** — automatic work must never make a person's install wait behind it.

## Pick a workspace on a machine

The picker browses the target machine's directories **over ssh**, not through its API. ssh is already the trust relationship — this server installed the program there — and asking the remote's HTTP API would 401 until somebody logged in over there, which is a second login for a directory listing.

Every installed machine is listed, and one that cannot be used says why **on its own row**. An earlier version filtered them out; a row that vanishes for a reason nobody can see is worse than a disabled row that gives one.

## ssh performance

Every ssh invocation is a TCP connect, a key exchange and an authentication — hundreds of milliseconds before the command runs, and these commands are tiny. So: **one shell held open per machine**, commands fed to its stdin, and `ssh -G` results cached. OpenSSH can do this itself with `ControlMaster`; Win32 OpenSSH cannot, and a controller may well be a Windows machine, so the multiplexing is here instead.

## What is not here

`7f98d39d` added an "enter a machine" mode that re-rooted the whole window at another server. `de957ff7` removed it: a dead tunnel left the window unable to load `/api/me`, so it rendered nothing and could not offer its own way back. Both commits are in this PR — the net effect is that the mode never ships, and the module survives only as a per-call address helper.

## Verification

`pnpm typecheck`, `pnpm test`, `pnpm format:check` — green. Verified against two real servers with a TCP-forwarding ssh stub.

Note: one conflict was resolved during the split — `71b6c11c` carries a `signInOnRemote` import belonging to the next PR, dropped here.


Myriad-Dreamin:
Folded into #448, which now carries identity + connect + project-config as one PR. Its 18 commits are on `feat/machines-identity`, rebased onto main; nothing was dropped. The branch `feat/machines-connect` is left in place but is no longer the head of anything.

PR #450, opened 2026-08-24
feat(machines): run an agent on a machine — its Sessions, Workspaces, Agents, Benchmarks and terminal, in this window

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← #808 ← #718 ← #544 ← #448 ← **#450** ← #577 ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

Pick a workspace that lives on another machine and the conversation runs there: that machine's server drives the agent, holds the messages, writes the trace, serves the terminal. The window never moves — it stays on this server and names the machine on the calls that concern it. And everything that reads an Agent's state — the Session list, the Workspace groups, the Agents page, the Evaluation Center, the terminal list — reads it off every machine this server holds a connection to, and says which machine an item is on.

## What is different from this branch's earlier version, and why

The earlier version was written before the stack: its server half was the pre-stack machines subsystem, and its page carried a connect-on-first-need loop — a widening schedule on the page that raised a forward whenever a call about a machine found none. That contradicts the rule the stack settled on in #574: **a connection someone asked for is held by the server**, never idles out, is re-established by the transport when it drops, and is restored after a restart or a hot push. So the page no longer connects anything. Reachability is the server's word (`MachineInfo.connection`), read from the machine list and re-read on a slow tick while any machine is out of reach; `raiseForward`, `setMachineConnector`, the `not_connected` retry in the fetch wrapper, `meOnMachine` probing and the autoconnect module are gone. Everything else in #450 that was about *Sessions living elsewhere* stands, and is re-implemented here on the current base.

## A Session's calls follow the machine that owns it

A Session lives on the server whose filesystem its workspace is on, so **every** call about it has to reach that machine — two dozen endpoints. Routing is a rule over the **path**: a request to `/api/sessions/<id>/…` goes wherever `<id>` was last seen, and the output stream with it (`lib/session-machines.ts`). Call sites stay unchanged; a Session's machine is recorded when a list hands one back and when one is created. An explicit `server` still wins for the calls that are not Session-scoped and know their target.

Two shapes escape a rule over the path and carry the machine by hand: the Trace endpoints, which bury a Session id in an Agent-level path, and bare URLs — a Workspace file's content, an attachment's scratchpad image, a Trace download — which never pass through the fetch wrapper. `trace-routing.test.ts` scans `endpoints.ts` so the next endpoint of the first shape is caught as it is written.

The mapping is in memory on purpose: rebuilt by the very list that displays the Sessions. Two caches that DO survive a reload sit beside it, per machine: the Sessions and Agents each was last seen holding, shown while the server is still re-holding its connection to that machine after a restart or a push, replaced wholesale by the machine's next answer (`lib/machine-cache.ts`).

## One list, several machines

The sidebar is every connected machine's Sessions, merged. Each server pages its own rows with its own cursor, so page state is keyed per source; counts are summed across the sources that answered; the merged list is ordered by a total order, so equal timestamps do not reshuffle between refreshes.

The line the list rests on: **a source that could not answer is not a source that answered nothing.** Only a 404 — this server has not got that Agent — is an answer. This server failing mid-swap abandons the reload and leaves the rows standing; a machine going quiet keeps its cached rows on screen and its cache intact; a machine with no connection contributes what it last held. `loading` is raised only when there is nothing on screen, because the chat page draws a skeleton in place of the open conversation while it is true. The chat page holds the routed Session through a refetch that momentarily stops naming it, and a failed direct lookup settles nothing while the machine that could hold the Session is out of reach.

## A Workspace group is a directory on a machine (was #646)

`/srv/app` on this server and `/srv/app` on a machine are two different directories; once the list merged machines, a path-keyed group folded them into one folder whose "+" created here, whose badge added both counts, and whose pages were asked of every machine. **`workspaceGroupKey(path, machineId)`** — one group per (machine, path); a machine of `null` keys exactly as before, so persisted collapse state, pins and manual order still resolve. The group is named for its machine (`app [SSH: prod-1]`, `nameOnMachine` / `S.chat.onMachine`) on its label, tooltip and the detail panel's Workspace line; this server's own stay bare. Counts fold per group (`workspaceCountsByAgent`, `aggregateWorkspaceCounts`); a group's `loadMoreFor` / `isLoadedFor` / `hasMoreFor` go to its own machine only (`sourcesFor`); the registry matches the pair in `loadWorkspaceRegistry`, `setWorkspaceAlias`, `unregisterWorkspace` and `mergeRegisteredWorkspaces`. The sidebar's `machineOfWorkspace` is gone — it read the registry by path and answered "this machine" for every session-derived group. `SessionsContext` exposes `machineLabels` (machineId → ssh alias) for the naming; an unlabelled machine falls back to its id.

## The list stays true without a reload

A Session created anywhere — the CLI, another tab, a schedule, an agent's child — is announced on the user channel (`session_created`, to the Project's owner and members) and the list fetches the row; a title set through `PATCH` is announced as `session_title`. Each connected machine's own event stream is opened through the proxy, since a Session there changes state on that machine's server and nothing else knows. A machine's `web_updated` is ignored: that is its web, not this window's.

## The composer follows the machine

The draft carries the workspace's machine with the path — in state, route state and the saved draft — offers that machine's Agents (last seen ones first, while it is asked), creates the Session on that machine, and re-records the routing under the id `postTask` hands back when the Session self-healed. Starting a chat from a Workspace group keeps the group's machine. A machine that cannot be asked for its Agents says so in the row and points at the Machines page; nothing here connects.

**Switching the model** (was #650) creates the new Session on the machine the source Session is on — `machineForSession(selected.sessionId)` as `createSession`'s machine — since the Workspace it carries over is a directory THERE and this server would refuse the path (`Workspace does not exist or is inaccessible`). The first task and the context trace follow the same routing.

## The Agents page and the Evaluation Center read every machine (was #653, #648)

An Agent is **one identity** across the machines it runs on, but its state directory — and its `benchmarks/` — is created wherever it has run. So the merge is by identity, never by machine, with the machine as attribution (`lib/benchmark-merge.ts`, pure and unit-tested: `mergeAgents`, `mergeBenchmarks`, `mergeBenchmarkCases`).

- **Agents page**: asks this server and every machine in `useSessions().machineIds` (`api.listAgents(projectId, machineId)`, `Promise.allSettled`) and merges by agent id. This server describes an Agent it also has; one that lives on exactly one machine carries `[SSH: alias]` beside its name. Its **New chat** hands the draft that machine with an empty Workspace (the temporary one ON it); settings, the stat shortcuts, usage and delete are disabled with the reason on hover (`S.agent.livesOnMachine`), rather than 404ing after the click.
- **Evaluation Center**: `listBenchmarks` / `listBenchmarkCases` / `listBenchmarkCaseFiles` / `benchmarkCaseFileUrl` take an optional machine; the page fans out over the same machine set, folds by benchmark id, and carries the machine on each evaluation. A scoreboard's append order *is* its sequence and wins over timestamps; a joined history is ordered by time, a single-source one is left as its file had it. First source (this server) wins title/description/`runs`; `caseCount` takes the max. The tree is built from the merged Agent list, so a machine-only Agent has a row; Cases are the union, and `BenchmarkCaseBrowser` reads files from the machine its listing came from.

Not in this PR: the Agent's own state (AGENTS.md, skills, memory, vault, schedules) is still a separate copy per machine, and `usage_records` is still per-server, so machine spend is missing from the cost center. Both need a design decision first, not a fan-out.

## Terminals

A shell opens on the machine its Workspace is on, including the fall-back to home, which is home *there*. The stream needs no runtime change — the seam it uses (`terminals()` / `attachStream`) is already on `main`: the platform answers `get` for `<terminalId>@<machineId>@<userId>` with a reference to the remote pty and relays the socket to that machine's own stream **through the held connection's SOCKS port**, like every other request to a machine's API (`machines/terminal-relay.ts`). The user id is what the runtime's owner check reads, so a client can only name itself; the relay then checks the user is an admin (`AuthService.isAdmin`, new).

The terminal list asks every connected machine, prunes a conversation's stored tabs only once every source answered and the machine set is published, adopts a shell already running on the conversation's own machine before starting another, and a detached pane carries its machine in the URL.

## Review fixes (folded 2026-09-17)

Six defects in the seam between this server and a machine, fixed in `8f00f28f1`:

- **A machine's own Agents are listed.** The list asks each machine about the Agents that machine runs, so a Session started on a machine-only Agent stays listed and reachable by deep link.
- **One Agent's failure is not the server's.** A reload is abandoned only when this server answered about nothing. An Agent whose list fails keeps its last rows, positions and counts while the others refresh, so the page no longer sticks on a skeleton.
- **The chat page settles out-of-reach Sessions.** It settles a Session it deleted itself while a machine is offline, and a routed Session whose machine is out of reach names the machine it is waiting on.
- **The draft waits for the route's machine** before validating the Agent the route names, so a machine-only Agent is no longer rejected and replaced by the fallback.
- **The terminal relay listens before it dials.** It holds frames typed while the stream opens and pushes back when the viewer falls behind, and a pane closed mid-dial no longer leaks a remote pty stream.
- **A machine's events match the current Project.** Its `session_created` and `schedule_fired` are compared against the current Project like this server's, since a machine's Projects carry this server's ids.

## Plugins and their settings per machine (#778 merged in, 2026-09-18)

- **A plugin can run on some machines only.** A Project's `[plugins]` table is shared by every machine, and a `[plugins.<machineId>]` table, keyed by the machine's stable id, adds to it or overrides it for that machine alone. The sync to machines hands each one the shared table plus its own. The Plugins page switches between all machines, this server and each machine, and an install or removal edits the table of the view it was made in.
- **Plugin settings can be edited on each machine.** Each server keeps its own values. The Settings dialog's Plugins page has a machine picker, and every read, save and action goes to the picked machine through this server's tunnel. Nothing is copied between machines.
- Changelog: `2026-09-17-plugins-per-machine` and `2026-09-17-plugin-config-per-machine`, both citing #450.

## Verification

After the last fold (Node 24): `pnpm -r build`, `pnpm typecheck` (0 errors), `pnpm lint`, `pnpm format:check` — clean. Web **134 files / 1745 tests** pass (`--maxWorkers=2`), including the folded suites: `benchmark-merge` (7), and the machine-pair cases in `session-grouping`, `workspace-registry` and `session-aggregation`; core 53 / 1109 pass. The folded commits touch no server file, so the server numbers stand from the branch's own run — **123 files / 1748 tests**, with `session-created-events` (3) and `terminal-relay` (3) new — as do the web suites from that run: `session-machines` (6), `session-merge` (3), `machine-cache` (5), `terminal-machines` (5), `trace-routing` (3), `route-session-hold` (5).


Myriad-Dreamin:
The machines stack is regrouped into two PRs. This one is the second: what remains of #450 after its three sign-in commits moved down to #448, plus what was #452. Base retargeted to `feat/machines-identity`.

10 commits on top of #448:

- `48d09e12` … `2eb3cbd0` — a Session's calls follow the machine that owns it, starting an agent on a workspace's machine, the composer offering that machine's Agents, and the far-side path fix.
- `5d958994` … `f73a2593` — what was #452: one Session list merged across machines.

Moved to #448 instead: `29935c79`, `36d9e053`, `33cf5b09`. #451's model sync is built on `signInOnRemote`, which `36d9e053` creates, so project-config could not sit below run-agent without them. See the note on #448.

`pnpm typecheck`, `pnpm format:check`, and `pnpm test` (server 1593, web 1617, core 1021, cli 383, docs 53, landing 69, skills 27) green on this branch.

One conflict resolution to look at: `packages/web/src/state/sessions.tsx`, where the cross-machine fan-out met main. `showCliSessions` is gone from the store since #466 made the CLI a thin client, so the fan-out is kept and every `cli: true` parameter and the preference hydration that fed it are dropped with it — the merged list is now the web list across machines, with no CLI-sessions toggle behind it.

hiyouga (approved):

PR #451, opened 2026-08-24
feat(server): machines belong to a Project, and get its config

Myriad-Dreamin:
> **Stacked on #450.** Review the diff against `feat/machines-run-agent`.

Two things that turn out to be the same thing: a machine belongs to a Project, and a Project's config has to reach it.

## An Agent over there needs the credential over there

Picking a model here and starting the Session there failed on the far side with **"Model is not in the Project config"**. Model config is per-server — `<data root>/<project>/.project_config.toml`, credentials inlined — and nothing carried it across. An Agent running on a machine calls the model endpoint *from* that machine; running the conversation remotely does not move where the key has to be.

This is the one secret that travels, and it is worth saying why it is not the admin password: that one never has to move, because the far-side scripts read the machine's own off its own disk. A model key has no local source.

It goes **through the tunnel**, as an ordinary authenticated call to that machine's own models API — which validates, invalidates its cached runtimes and tells its open tabs. A far-side script would have to re-implement the config writer, and `penguin config model add` would put the key in argv where `ps` reads it.

**The merge is where this lives or dies.** `PUT /models` is a whole-table replace, so every entry the machine already had is re-sent — and re-sent *without* `apiKey`, because omitting it is exactly what keeps the stored value. A GET reports keys masked, so echoing one back would overwrite a working credential with `sk-1…abcd`.

A Project the machine lacks is **created there, under the same id** — which is what stable ids are for: a Session started on that machine sends this id for it to resolve.

## Machines belong to a Project

Machines sat in the same nav group as Agents, Skills and Models, under the Project switcher — and were the only row that did not change when the Project did. That mattered more than tidiness once models started syncing: this server was handing **every shared Project's credentials to every machine it could reach**.

Split, because "which machines does this Project use" and "what is on that host" are different questions with different owners: `<project>/machines.json` holds membership; `machines-installs.json` stays as what is true of a machine whoever asks. One host runs one program, and two Projects must not be able to disagree about which. So a machine installed for another Project is neither yours nor blank — it reports `elsewhere`, and the button says "add to this project" instead of offering a 30 MB transfer to write one line of JSON.

The default Project inherits what the server already had: an **absent** list is seeded from the install records, an **empty** one is left as somebody left it.

## A token, so a real password stops locking a machine out

Reaching a machine's API needed a session, and the only way to get one was to read its **seeded** admin password. That works until somebody sets a real one — which is what setting a machine up properly means — and then never again: the machine stays connected, healthy, and impossible to configure or update. Found on a live deployment, where a machine's config was eight days stale while the page showed it current.

`penguin server auth-token` (from #443) mints a session from the data root instead. What authorizes it is the ssh account's access to that root, not a secret. Used by both paths that needed a session — the model sync and the hot update — each falling back to the seeded password for builds too old to mint.

## Verification

`pnpm typecheck`, `pnpm test`, `pnpm format:check` — green. The merge rules are tested against a fake machine and against a **real server over a real socket**: the Project is created with its agents and DB row, the keys land, the default resolves to the pair this side named, and the machine's own key survives untouched.


Myriad-Dreamin:
Folded into #448, which now carries identity + connect + project-config as one PR. Its 7 commits are on `feat/machines-identity`, rebased onto main. Note that moving project-config below run-agent required #450's three sign-in commits to move down with it — the model sync here is built on `signInOnRemote`; see the note on #448.

PR #452, opened 2026-08-24
feat(web): one Session list across machines

Myriad-Dreamin:
> **Stacked on #451.** Review the diff against `feat/machines-project-config`. Last of the series.

A Session lives on the server whose filesystem its workspace is on, so a Project's list was never one server's answer — but that is all it ever fetched. Sessions started on a machine showed up while browsing, because `add()` holds each new row in memory, and **vanished on refresh** when the list was rebuilt from this server alone. Five conversations on a machine, and a sidebar reading `penguin-harness 0`.

## The merge

`reload()` and `loadMoreFor()` now fan out over this server plus the Project's machines.

**The page cursor moves per source**, which is the part that fails quietly. Every server pages its own Sessions with its own offsets, so one shared cursor asks machine B for rows only machine A had reached — skipping whatever sat between, with a list that still looks plausible. The page key carries the source for that reason, and `isLoadedFor` is true only when **every** source has answered: one machine's first page arriving does not make a folder complete, and calling it loaded hides the rest behind a "load more" that never appears.

Counts are **summed**, or a folder badge contradicts the rows under it. Ownership is recorded as rows land, since that map routes every later call about a Session — a row listed without it is asked of this server, which does not have it — and is forgotten on a Project switch, with the rows it described.

## The 401 that made it look like nothing

The fan-out was correct and **every request to a machine answered 401**: a machine is a separate server with its own accounts, and the browser had no session on it. The merge caught the failure and treated it as "that machine has not got this Agent" — a genuinely ordinary case, since an Agent is per-server — so the rows were missing with nothing said.

Both halves fixed: the browser's sign-in route now asks the machine's own CLI for a token first (the seeded-password path stops working on any machine whose admin password somebody set), and the provider establishes that session **before** naming a machine as a source. A machine that cannot be reached either way is left out entirely rather than added as a source that silently answers nothing — the list is then this server's, which is what it was before, instead of a list with holes.

## A self-healed Session

`postTask` answers with the **current** id — a Session with no Trace whose process restarted heals into a new one. The draft's send path recorded everything under the id it created and navigated to the one it got back. The routing map is the half that failed silently: a remote Session left mapped under the old id has its new id resolve to "this server", which does not have it.

## Verification

`pnpm typecheck`, `pnpm test`, `pnpm format:check` — green. Three new tests pin the merge, the per-source cursor, and the 404 tolerance; verified end to end against a real machine through the proxy path the sidebar uses.

### DIFF
--- a/packages/server/src/machines/service.ts
+++ b/packages/server/src/machines/service.ts
@@ -1066,10 +1066,23 @@ export class MachinesService {
     if (resolved === null) {
       return { kind: "failed", detail: "ssh could not resolve that host." };
     }
-    return await this.#effects.signIn({
-      target: { alias: machine.alias, user: resolved.settings.user },
-      assets: this.#assets,
-    });
+    const target = { alias: machine.alias, user: resolved.settings.user };
+    // The machine's own CLI first, for the same reason the model sync prefers it: it needs no
+    // password, so it still works on a machine whose admin password a person has set — which
+    // is precisely the machine a browser could not otherwise be signed in to from here.
+    const minted = await mintTokenOnRemote(target, (t, command) => this.#effects.runOn(t, command));
+    if (minted.kind === "minted") {
+      // Synthesized as the Set-Cookie that machine's own login would have sent: the caller's
+      // job is to rename one into the machine's namespace, and it should not have to know
+      // this one was minted rather than issued. No Secure — the browser is talking to THIS
+      // origin through the proxy, not to the machine.
+      return {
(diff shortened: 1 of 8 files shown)


Myriad-Dreamin:
Folded into #450, which now carries the rest of run-agent plus this merged Session list as one PR, on top of #448. Its 5 commits are on `feat/machines-run-agent`, rebased onto main. The `showCliSessions` plumbing they carried is gone with #466 — see the note on #450.

PR #454, opened 2026-08-25
feat(server,cli): the owner token — nothing auth-shaped rests on disk

Myriad-Dreamin:
> **Stacked on #443.** Review the diff against `feat/penguin-auth`.

The stateless scheme in #443 still kept one long-lived secret at rest: `auth-token-secret`, a permanent, admin-equivalent signing key — a leaked **backup** of the data root meant forging sessions against the live server, forever. This PR replaces that anchor with a deliberately short-lived one, and moves the key where no backup can reach it.

## The owner token

`<root>/owner-token` — a fresh random value at every process start, 0600. Presenting it to `POST /api/auth/owner` over loopback proves what ownership has always meant here (*the ability to read the data root*) and is exchanged for an ordinary signed session, capped at the session TTL.

- It **signs nothing and derives from nothing**; a copy leaked through a backup is overwhelmingly a value the server no longer honors.
- A wrong value and an unknown account answer **identically**, so a caller without the token cannot enumerate accounts.
- A magic-cookie file + loopback HTTP rather than a Unix socket, **on purpose**: identical behavior on every platform Node runs on, no named-pipe DACL code to get wrong.
- The honest cost: the ownership check happens at *read* time — a value exfiltrated mid-boot stays usable until the next restart. A bounded window, where the signing key's was unbounded.

## The signing key moves into memory

Generated at process start, written nowhere. Nothing to leak, nothing to rotate — **a restart is the rotation**. Who pays:

| holder | on restart |
| --- | --- |
| hot pushes | nothing — App swaps keep the process and the key |
| CLI / machine tokens | nothing — 1h lifetime, re-minted on demand |
| browser session | one re-typed password, after a *real* restart only |

## Minting follows

`penguin auth token` redeems the owner token against the running server — the common case; a controller minting on a machine has just probed its server. Server stopped ⇒ there is no key anywhere ⇒ fall back to the one session shape that needs none: a **legacy database row**, which verification has always honored. The row path's foreign key keeps it honest about accounts, and a stale lock (crash leftovers) falls through to the row rather than failing.

The denylist keeps its meaning with a sharper edge: revocation must hold **for as long as the key lives**, and the key lives across App swaps. The test pins that with a deliberately shared key — and pins separately that *without* sharing, a restart voids every signed session by signature alone.

## Removed

- `auth/token-secret.ts` and the `auth-token-secret` file (never generated again).
- The `no_database` mint outcome — a stopped server with a database mints a row; a fresh root is honestly `no_user`.

## Not in this PR (stated follow-ups)

- The **first-login link** replacing `initial-admin-password` — that file is untouched here.
- Unifying the desktop shell's one-shot token onto `POST /api/auth/owner` (it is the same primitive).

## Verification

`pnpm typecheck`, `pnpm test` (all packages, 90/90 server files), `pnpm format:check` — green. 13 auth tests including: owner redemption end-to-end, indistinguishable refusals, mint-over-loopback against a **real listening server** (signed token, zero rows written), the row fallback with the server stopped, restart-voids-everything, and revocation across a shared-key "App swap".


Myriad-Dreamin:
Folded into #443 — same tree, one commit, so the whole auth design reviews as a single diff.

Keeping both would have meant reviewing two files that do not survive the series: `auth/token-secret.ts` (added, then deleted when the signing key moved into memory) and `cli/commands/auth-token.ts` (added, then deleted when minting collapsed to one command with `--mark`).

Closing in favour of #443.

PR #456, opened 2026-08-25
fix: a stopped background command reports as stopped, not as a failure

hiyouga:
## Problem

A `run_in_background` command that ends on a SIGTERM reaches the conversation as:

```
Background command failed: `export PATH=…; pnpm dev:web` (process_id proc-2483f6cc) — terminated by signal SIGTERM
```

A SIGTERM is almost always somebody stopping the process on purpose. The notice reads like a crash, and a notice arriving while the Session is idle **starts a Task of its own** (`startBackgroundNoticeTask`) — so pressing **Stop** on a dev server in the Web App's process list woke the agent up and invited it to restart exactly what had just been stopped.

Two separate holes produced it:

1. The host stop path (`POST /processes/:id/kill` → `Session.killBackgroundCommand` → `CommandSessionManager.kill`) never disarmed the armed completion report, unlike `input_command`'s `kill`, which calls `clearExitWatchers()` first. The kill's own SIGTERM then came back as a failure notice.
2. `armCommandDoneReport` classified *any* signal exit as `failed`, so an external stop (a Ctrl-C in a terminal sharing the process group, a `pkill`, a supervisor) read the same way.

## Change

- **`CommandSessionManager.kill` disarms the report.** Every stop that comes through it was asked for by someone who reads the outcome directly — the model (the call's own result) or the user (the row they watched stop). Same silence `input_command`'s `kill` and an explicitly aborted subagent round already get.
- **Deliberate stops that still report use a new `status: "stopped"`** instead of `failed`: an outside stop signal (`SIGTERM`/`SIGINT`/`SIGHUP`, via the new `isStopSignal`) and a stop the harness forced without going through `kill` (capacity eviction, idle reap, via the new `ManagedSession.stopRequested`). For a stopped status, the `[background_task_done]` block's leading sentence says so in as many words: *"…was stopped on purpose — someone ended it, it did not crash. Do not restart it unless you are asked to."* The notice's head line reads `Background command stopped: … — stopped by SIGTERM`.
- **`failed` keeps the outcomes nobody asked for**: a spawn error, a non-zero exit, a hard kill or a fault signal — so an OOM kill still reads as the failure it is.
- `input_command`'s `kill` on an already-exited process uses the same vocabulary (`stopped by SIGTERM` rather than `terminated by signal SIGTERM`).
- Web App: the notice card carries the third outcome ("Background command stopped" / 「后台命令已停止」) with a new muted square `StatusIcon` state, rather than the red failure X.

`resultForExit` (the foreground `exec_command` result) is deliberately untouched: that outcome is read inside the call that produced it, where there is no out-of-context notice to misread.

### Traces

New Traces can carry `status: stopped` in a `[background_task_done]` block. Existing Traces are unaffected and parse exactly as before — the parser accepts the two old values unchanged, so no migration and no compatibility shim. (An older build reading a *new* Trace would fall back to `completed` for that one field; nothing else changes.)

## Verification

Node v24.18.0, full CI-parity chain in a clean worktree:

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (core 985+, server 970, web 1289, cli 299, desktop 106, docs 53, landing 61, skills 25)

New tests in `packages/core/test/background-execution.test.ts`:

- a stop signal from outside reports `stopped` with `detail: "stopped by SIGTERM"` (POSIX-only);
- the host's stop button sends no completion report at all;
- a harness-forced stop (eviction/reap path) reports `stopped` on every platform;
- the `stopped` marker block round-trips and carries the do-not-restart instruction, while `failed` keeps the plain lead;
- a stopped command's notice head reads `Background command stopped: …`, never `failed`.

The new stop glyph was rendered at 13px next to the existing `done`/`failed` icons to check its weight.


hiyouga:
Closing: the fix over-reached. A SIGTERM stop should still report back to the conversation — only the wording needs to change (`stopped` instead of `failed`, with `failed` kept for real errors). Reopening as a narrower PR.

PR #464, opened 2026-08-25
feat(server,web): messaging integration — Feishu + Telegram behind one channel seam; a saved config per channel, at most one connected

hiyouga:
A Session picked in the web sidebar can be connected to a messaging bot — **Feishu (Lark) and Telegram are the first two channels** of a channel-connector seam. A Session keeps **at most one saved config per channel** (both may sit saved side by side) and **at most one of them is enabled** — the enabled one holds the live connection. Messages sent to that bot flow into the Session as ordinary user input, and the Session's completed AI replies are relayed back to the chat. Design PR: penguin-harness-design [#61](https://github.com/Prism-Shadow/penguin-harness-design/pull/61).

## What's in here

**Storage** — SQLite `messaging_bindings`: primary key **`(session_id, channel)`** (`channel` discriminator — `feishu` | `telegram`), `account_id` + unique `(channel, account_id)` index, `config_json` for the channel's own config, **`enabled` as stored connection INTENT** (runtime status stays in memory), last-chat memory, timestamps — with `MessagingBindingsRepo`. Cross-session uniqueness is per bot account per channel (409 `feishu_app_in_use` / `telegram_bot_in_use`); the **one-enabled-per-Session** invariant is enforced by the state route, not the repo, which stays a dumb store. Secrets live inside the config document, plaintext at rest like the proxy address in `server_settings`, masked at every API surface; a cleared secret is stored as the empty string so the row, its non-secret fields and its account identity survive.

**Save/Enable split** — saving and connecting are separate concerns: PUT persists credentials only and never opens or closes a connection, with one deliberate exception — an **enabled** binding's connector restarts with the just-saved credentials, so stored config and live connection never diverge. The explicit `POST …/messaging/<channel>/state {enabled}` toggle connects with the STORED credentials or terminates. New configs start disabled; server startup connects only enabled bindings.

**One enabled channel per Session** — the state toggle refuses to enable a channel while the Session's other channel is enabled (**409 `another_channel_enabled`** — turn that one off first) and to enable a config whose secret was cleared (**400 `feishu_secret_required` / `telegram_token_required`**). The Web App grays the switch with the reason on screen before the request is ever made. The session row's small channel glyph therefore marks the **enabled** channel; a merely saved config leaves the row unmarked.

**Removing a credential is a per-field clear, not an unbind** — PUT takes `clearAppSecret` / `clearBotToken` (the models page's clear idiom: a typed secret wins over a checked box), refused with **409 `messaging_disable_before_clear`** while that channel is enabled, so a live connection can never outrun the credential its store still has. `DELETE …/messaging/<channel>` still removes a whole channel config and is kept for API completeness — the Web App no longer offers it.

**Runtime** — `runtime/messaging/`: `MessagingBridge` (modeled on the scheduler; started in `platform.ts`, stopped in the dispose effect) manages channel connectors behind the `MessagingChannelConnector` seam (`createClient`/`connect`, normalized inbound messages, ready/error callbacks). Because at most one of a Session's channels is enabled, the bridge holds at most one inbound event connection per Session — the enabled binding's — and every entry carries its channel, so a saved-but-dark channel reads as `disconnected` and its test message gets a fresh client instead of the connected channel's cached one. The seam survived the second channel unchanged except one widening: `checkCredentials` resolves an optional **account label** so a channel can name the account the credentials sign in as (`FeishuConnector` returns null). `FeishuConnector` runs over the injectable `FeishuSdk` factory (lazy dynamic import of `@larksuiteoapi/node-sdk`). Per-binding in-memory status: `disconnected | connecting | connected | error` + lastError + changedAt.

**Telegram channel** (`telegram-connector.ts` + `telegram-api.ts`) — the whole credential is one **bot token** (`<numeric id>:<secret>`, as issued by @BotFather); the id half is the account identity, so a rotated token keeps its binding and a cleared config keeps that identity. No SDK: the transport is a thin fetch wrapper over `https://api.telegram.org` (hardcoded; no domain field), injectable for tests exactly like the Feishu factory. Inbound is an **offset-based `getUpdates` long poll in the connector** — no public URL: a `getMe` probe surfaces a bad token immediately, then a one-time **backlog drain** (`offset: -1`) so messages sent while no connection existed never replay as a task flood (matching Feishu, where missed events are simply gone), then long polls advancing the offset per processed update. Failures back off exponentially (1s → 60s), report **once per outage**, and recovery re-probes and re-fires ready. Only `message` updates are processed (non-text → a bilingual "text only" reply); Telegram message ids are chat-scoped, so the connector packs `chatId:messageId` into the seam's opaque reply ref and group replies thread via `reply_parameters`. Outbound `sendMessage` rides the shared 4000-char chunker — inside Telegram's 4096-character hard cap. The credential test surfaces the bot's **@username** in the success feedback.

**Inbound** — every event first records the chat as the reply target; text messages start a Task as **plain `userText(text)`** — exactly as if typed into the web composer, no marker block, no special sender (the model deliberately does not learn the origin; explicit product decision) — via `startTask(..., {queueIfBusy: true})` (busy Sessions queue as follow-ups, never 409). Other message types get a bilingual "text only" reply. `packages/core` is untouched (zero diff against main).

**Outbound** — the bridge subscribes to the Session's in-process channel, accumulates main-conversation completed assistant text (skips partials/thinking/tools, subagent `origin` traffic, and compaction windows), and flushes on the `task_state` idle flip: reply-to-message in group chats, plain send in direct chats, chunked at 4000 chars preferring newline boundaries. ALL completed tasks mirror once a chat is known (web-initiated turns too); joining mid-run skips that run's partial tail. `approval_request` sends a one-line notice pointing at the web UI.

**API** — the channel-agnostic **`GET /api/sessions/:sessionId/messaging`** returns `{bindings: [...]}` — **every** saved channel config (`channel`-discriminated, masked, with its own `enabled` intent, runtime status and `lastChatKnown`) — and the channel-aware editor loads only this one. Per channel, the same verb set: `GET|PUT|DELETE …/messaging/{feishu|telegram}`, `POST …/state`, `POST …/test`, `POST …/test-message`; each subtree sees only its own channel's row, so the other channel is never disturbed. Masked-secret fields are **omitted** when nothing is stored, which is what tells the editor to render "not configured" instead of a placeholder. Telegram specifics: 400 `telegram_token_invalid` when the numeric id cannot be read, 409 `telegram_bot_in_use`, 409 `telegram_no_chat`, `botUsername` in the test response. `SessionInfo.messagingChannel` carries the **enabled** channel for list rows. Documented in `packages/docs/content/server-api.{en,zh}.md`.

**Web UI**
- **Two management surfaces, ONE shared channel-aware binding editor** (`features/messaging/messaging-binding-{form,editor,modal}` — moved from `features/chat/feishu-*`; a `useMessagingBinding` hook + `MessagingBindingBody` + `MessagingBindingHelp`, neither host forks the form): the session-row menu's channel-neutral **"Messaging binding…"** dialog and the **Messaging dock panel** (消息软件). The `Segmented` channel selector **no longer locks** — it switches between two independently savable forms, each showing its own configured / enabled state and each keeping its own edits across a switch. Fields swap per channel: App ID / App Secret / API domain vs a single Bot Token.
- **The form opens on its fields.** The explanation that used to lead it moved into **three collapsed folds below the Save area**: the channel's setup steps (ending in its 前往教程 / tutorial link), what binding does, and troubleshooting. `HelpFold` grew an optional `title` so a stack of folds can say what each one holds instead of showing a column of identical "More info" rows; it stays the same WAI-ARIA disclosure (real button, `aria-expanded`, panel `hidden` in flow).
- **Links moved onto the fields.** The channel's 前往开发者后台 / developer-console link now sits at the top-right corner of its leading credential field — the models dialog's "get API key" idiom — instead of a links row above the form. Feishu → https://open.feishu.cn/app, Telegram → https://core.telegram.org/bots/api.
- **Secrets behave exactly like the models page's API key**: the field always starts empty, a stored secret shows only as its site-wide mask on the line beneath, and a "clear stored …" checkbox next to that mask applies on Save (typing unchecks it, and a typed replacement wins over a checked box). While the channel is enabled the checkbox is disabled with the reason on screen — a disabled control does not reliably fire hover, so the hint is not a `title`. **There is no unbind button** in either host, and no unbind confirmation: removing a credential is the clear.
- Per-channel identity glyphs (`CHANNEL_ICONS`): a feather for Feishu, the paper plane for Telegram — on the session row's **enabled** indicator (tooltip/sr names the channel). Full bilingual i18n for every new string, including the two new error codes.
- Session-row **hover affordance** (from the earlier commits): archive + a solid-dot ellipsis "more" button opening the row's context menu anchored at itself; delete lives only inside the menu.

**Dependency** — `@larksuiteoapi/node-sdk` ^1.73.0 in the server package (CJS; external in the server dist, inlined by the desktop/deploy esbuild bundles). Telegram adds **no dependency** (global fetch).

## Verification

- `pnpm -r build`, `pnpm -r typecheck` — clean (all packages); `git diff origin/main...HEAD -- packages/core` is empty.
- Server suite: 90 files, **1012 passed**. `messaging.test.ts` (17) and `messaging-telegram.test.ts` (18) cover the reshaped key and the new rules: both channels saved side by side with the second save leaving the first one's live connection alone, the channel-agnostic GET listing both rows with per-row `enabled` + status, **409 `another_channel_enabled`** on the second enable and the same enable succeeding once the first is switched off, the clear flag refused while enabled (**409 `messaging_disable_before_clear`**) then dropping the secret with the row and account identity kept, a typed secret winning over the flag, **400 `*_required`** when enabling a cleared config, and the session list marking only the ENABLED channel. Plus everything from before: masking + keep-on-blank, per-channel account 409s, save/enable split with restart-on-save-while-enabled, both probes, inbound → plain user input, group reply threading, chunking, connect-time backlog drain, poll failure → recovery, startup connect + stop. No test opens real network (injected Feishu SDK and Bot API transport).
- Web suite: 100 files, **1318 passed** — `messaging-binding-form.test.ts` (14) for the multi-config load, the clear-checkbox mapping and the dirty/testable rules, and the new `messaging-binding-editor.test.ts` (6) pinning what a type checker cannot see: the form opens on fields with no prose above the first input, the console link rides the credential field's corner per channel, the clear checkbox is the removal affordance and is gated on screen while enabled, the gating hint names the other channel, and the three folds render collapsed with the relocated explanation and per-channel tutorial link inside.
- `pnpm format` + `format:check` + `git diff --check` clean.


hiyouga:
Merged `main` in and applied the review feedback on the binding editor and the sidebar mark.

## Conflict with `main`

`git merge origin/main` conflicted in exactly one file, `packages/web/src/api/endpoints.ts`, and only inside the `import type { … }` list: `main` had inserted the four `ModelOAuth*` DTOs (#465) where this branch had inserted the three `Messaging*` DTOs. Both sets were kept, in alphabetical order (`Messaging*` sorts before `Model*`). Nothing else in the file was touched — the `Feishu*` / `Telegram*` type imports and every messaging endpoint wrapper auto-merged, as did `strings.ts`, `strings-en.ts` and `pnpm-lock.yaml`.

Verified rather than assumed: `git diff origin/main...HEAD -- packages/web/src/api/endpoints.ts` is **63 insertions, 0 deletions** — this branch's additions only, nothing from `main` dropped. `pnpm install --frozen-lockfile` was re-run after the merge (agenthub floor moved to `^0.4.8`, `pnpm-workspace.yaml`'s supply-chain exemption changed); it resolves clean.

The branch was merged a second time after #463, #472 and #473 landed; that merge was clean. Nothing in either merge reads or writes existing on-disk state (`web.db`, `system_config.yaml`, `.project_config.toml`) — the resolution is a browser-side type-import list.

## The editor: controls lead the form

`MessagingBindingBody` now renders, top to bottom:

1. channel selector (`Segmented`)
2. enable toggle + live status
3. the two probe buttons
4. `toggleHint` (conditional)
5. the selected channel's credential fields

Everything above the fields is channel-independent in height, which is what actually fixes the jump — the two channels differ only in how many fields they have, and those now sit last.

**`toggleHint` trails the probes**, rather than sitting under the switch it explains. Reserving a line under the switch was the alternative, but the hint does not fit one line at every width: at the dock's minimum width (320px) the English `otherEnabledHint` — "Only one channel can be enabled per conversation: turn off the Feishu connection first" — wraps to two lines, so a one-line reservation would still let the probes move, and reserving two lines leaves permanent dead space and is one narrower panel or one longer translation away from being wrong again. Trailing the probes makes the guarantee structural instead of numeric: nothing above the hint can move, whatever it grows to. It is still inside the control block, one row from the switch and above the fields, and each of the three hint strings names the switch's own label ("…then enable the connection", "turn off the X connection first"), so its referent stays unambiguous.

The status line's `lastError` was checked, not assumed: it truncates inside its own flex track, so the row stays exactly one line high (`switch → probe` gap is 36.00px in both the `connected`/`error` and `disconnected` states). The probe labels only ever get *shorter* in flight (`Testing…`, `Sending…`), and at the narrowest dock width the resting pair measures 129.8 + 9 + 153.7 = 292.5px against 293px of usable width, so they share one row at rest and cannot wrap when the labels swap.

Block comments updated: the module header, `MessagingBindingBody`'s doc comment, the inline "the fields lead the form" comment, `MessagingBindingHelp`'s doc comment and `messaging-panel.tsx`'s header now state the arrangement that ships.

## Measured, not asserted

Driven through the Playwright e2e harness (mock LLM, scratch `PENGUIN_HOME`), reading `boundingBox().y` of the enable `Switch` and of both probe buttons while switching channel. Two Sessions: one with Feishu saved **and enabled** (so Feishu shows a live status and Telegram shows the "other channel enabled" hint — the asymmetric case), one with nothing saved (both channels gated by the missing-credential hint). Dock measured at its minimum width, 320px.

**Dock panel — feishu → telegram, absolute viewport y**

| case | element | before | after |
| --- | --- | --- | --- |
| one channel enabled | switch | 476.81 → 249.44 (**−227.37**) | 182.97 → 182.97 (**0.00**) |
| one channel enabled | test / send | 512.81 → 316.92 (**−195.89**) | 218.97 → 218.97 (**0.00**) |
| nothing saved | switch | 422.84 → 249.44 (**−173.40**) | 182.97 → 182.97 (**0.00**) |
| nothing saved | test / send | 490.33 → 316.92 (**−173.41**) | 218.97 → 218.97 (**0.00**) |

The gated case is covered on both sides of the table: with `toggleHint` visible on the Telegram side only (one channel enabled) and on both sides (nothing saved), the toggle and the probes do not move either way.

**Dialog — offsets inside the dialog, feishu → telegram**

| element | before | after |
| --- | --- | --- |
| switch | 372.86 → 185.95 (**−186.91**) | 119.48 → 119.48 (**0.00**) |
| test / send | 408.86 → 253.44 (**−155.42**) | 155.48 → 155.48 (**0.00**) |

**One residual in the dialog, and it is not this layout.** In absolute viewport coordinates the dialog's controls still move 168.08 → 245.80 (+77.72) when the channel switches. That is the shared `Modal`'s vertical centering (`sm:items-center`): the dialog's own height changes 622.80 → 467.38 as the field list shortens, so a centered box's top drops by half of that. Everything in the dialog translates by the same +77.72 — the title, the channel selector the user just clicked (125.34 → 203.06), the toggle, the probes and the footer Save — so nothing moves *relative to anything else* any more, but the box as a whole does re-centre. Removing that means either top-aligning every dialog in the app or giving `Modal` a one-off alignment prop for this one dialog; both are design calls, so I left it and am flagging it. The dock panel, which is top-anchored, is at a flat 0.00.

Same figures reproduce in English (the toggle row wraps to two lines at 320px in English — identically in both channels, so `switch → probe` is 62.99px on both sides and still 0.00 movement).

## Sidebar: one relay glyph

`CHANNEL_ICONS` had exactly one consumer, so the two-entry map is gone rather than reduced to a single branch. In its place `MESSAGING_RELAY_ICON` — the paper plane — with a doc comment stating what the mark means (a Session relaying through a messaging channel) and why one shape is safe here. The row's `title` and `sr-only` text are untouched: both still read `S.messaging.enabledIndicator[channel]`, so the channel identity is carried by text, which is what the `icons.tsx` comment about shape not being the carrier was already claiming and now actually needs.

Verified in the captured frames and by reading the DOM: a Feishu-bound row and a Telegram-bound row render the identical `d` (`M22 2 11 13M22 2l-7 20-4-9-9-4z`) with tooltips "飞书连接已启用" and "Telegram 连接已启用".

`MESSAGING_ICON` (the chat bubble on the dock tab and the row-menu action) does **not** end up adjacent to the new mark: the bubble is a dock-tab / menu-row glyph, the paper plane is inline after a Session title. After merging #472 I also re-read the merged Session-row block — its `UpdateDot`s land on the page-nav rows and the account avatar, both outside the Session list and separated from it by the nav collapse strip, the "Session" header and the workspace group header; a red 6.75px disc at a row's right edge and a dim gray 12px stroke glyph after a title do not read as one badge system.

## Verification

`pnpm build`, `pnpm typecheck`, `pnpm --filter @prismshadow/penguin-web test` (102 files / 1359 tests), `pnpm --filter @prismshadow/penguin-server test` (91 files / 1037 tests, `messaging.test.ts` 17 + `messaging-telegram.test.ts` 18 among them), `pnpm --filter @prismshadow/penguin-core test` (997 passed / 5 skipped), `pnpm format` + `pnpm format:check`, `git diff --check` — all pass. Screenshots taken at `deviceScaleFactor: 2` for both channels × both hosts × light and dark, plus the sidebar; checked for home-directory paths, bot tokens, app secrets and API keys — the only credential-shaped strings in frame are the fixture App ID and the site-wide mask of the fixture secret.

The changelog pair was updated: the entry said the form opens on its fields and that the row glyph is the channel's own, neither of which is true any more.

hiyouga:
Second round: re-merged `main` after #466, changed the outbound relay to per-message delivery, and the Windows failure is resolved.

## Merge with #466

`git merge origin/main` conflicted in the two files you named, both purely additive collisions where each PR inserted into the same list:

- **`packages/server/src/services/session-service.ts`** — `SessionServiceDeps` gained #466's `controlEnv?: (ctx: ControlEnvContext) => Record<string, string>` and this branch's `messagingChannel?: (sessionId) => MessagingChannel | null` at the same position. Both kept; `controlEnv` placed next to `proxyEnv`, which is the family it belongs to.
- **`packages/server/src/app.ts`** — the `new SessionService({…})` argument gained `controlEnv,` from #466 and the `messagingChannel` lookup lambda from this branch. Both kept.

I did not trust the marker-free result. Both files were diffed **in both directions**:

| file | vs `origin/main` | vs the pre-merge branch tip |
| --- | --- | --- |
| `app.ts` | only this branch's messaging additions, **0 deletions** | only #466's additions (`mintApiToken`/`storeApiToken`, `ControlEnvContext`, `canonicalApiUrl`, `controlEnv`, both wiring sites) |
| `session-service.ts` | only the `MessagingChannel` import + the dep + its use in `toInfo` | only #466's (adoption sweep, `includeCli` removal, `controlEnv`, the `client` hint) |

And read end to end rather than inferred: `controlEnv` is defined at `app.ts:596` and threaded through **both** core entry paths — `createCoreSessionLoader(config.root, sessionSources, { proxyEnv, controlEnv })` on the manager's loader and `controlEnv` on `SessionService` — while the messaging repo, bridge, `AppDeps` entries and `app.route("/api/sessions", sessionMessagingRoutes(deps))` all survive. `platform.ts` (auto-merged) holds both sides' lifecycle: `deps.messaging.start()`/`.stop()` and `deps.sessionService.adoptUnmanagedTraceSessions()` sit side by side. `session-service.ts` consumes both deps — `messagingChannel` in `toInfo`, `controlEnv` spread into `createAgent` next to `proxyEnv`.

`pnpm install --frozen-lockfile` re-run after the merge: clean.

## The control-plane env and the messaging bridge

Checked, and the answer is that nothing needed to change. The bridge **spawns no processes and touches no CLI surface** — the only `exec` in `packages/server/src/runtime/messaging/` is `TELEGRAM_TOKEN_RE.exec`, and there is no `spawn`, `child_process`, `createAgent` or `createSession` anywhere under it. It starts work through `this.deps.runner.startTask(...)`, i.e. the server's `SessionManager`, whose sessions are loaded by the single `createCoreSessionLoader(…, { proxyEnv, controlEnv })`. So a messaging-driven Task's command subprocesses get the control variables through exactly the same path as a web-driven or scheduled one, composing with `hostEnvForChild` → vault → `controlEnv` → `HARDENED_ENV` rather than around it. No messaging-specific env assembly exists on this branch.

## The `ci-windows (packages)` failure was a flake

The old head `ae240ea7` failed `test/exec-session.test.ts > … > a PENGUIN_* nobody has invented yet is stripped` with `Received: "[exit code: 1]"`. That is the shell-probe signature, not a strip regression: a broken strip prints `X=[leaked]`, and a broken assertion still prints the `X=[` prefix — an empty stdout with exit 1 means the spawned `node -e` never ran. Its siblings in the same `describe`, which set the same variables and spawn the same way, all passed, as did 45 of 47 core test files. And `git diff origin/main...HEAD -- packages/core` on this branch is **empty**: the branch touches no core code at all.

**CI on the new head is fully green, `ci-windows (packages)` included** — run [33003866194](https://github.com/Prism-Shadow/penguin-harness/actions/runs/33003866194), all seven jobs `success`. Did not reproduce.

## Outbound relay: per message, as it completes

The bridge buffered every completed assistant text of a run and, at the idle flip, sent `buffer.join("\n\n")` as one message. A chat therefore saw nothing during the run and then one wall of text. Now each completed assistant message is relayed on its own the moment it completes, so a run that writes working notes between tool calls before its answer reaches the chat as that same sequence.

**Ordering — a promise chain per entry, not a queue object.** `BridgeEntry.sendChain: Promise<void>` starts at `Promise.resolve()`, and `relay()` does `entry.sendChain = entry.sendChain.then(() => this.deliverReply(entry, body))`. Enqueueing happens inside `observe`, which is synchronous on the channel event, so enqueue order is completion order and the chain preserves it. A queue object would have been the same guarantee with a hand-rolled drain loop; the chain is three lines and has no drain state to get wrong.

**Not stalling or breaking the run.** I checked what the old `void this.flush(...)` did before changing it: `flush` wrapped everything in `try/catch` and called `recordError(…, "messaging_send_failed")`, so it never rejected and never threw into the caller. `deliverReply` keeps that catch verbatim, which is also what keeps the chain from ever rejecting — no unhandled rejection, no need for a `.catch` on the tail. `relay()` returns synchronously, so `observe` never awaits a send. Net effect of a slow or failing channel: that Session's later messages queue behind it, and nothing else is affected.

**Gates unchanged, now per message.** `armed` moved from the idle-flip check to `relay()`, which is the same protection at the moment a message completes: a connection made mid-run has `armed === false` (it is initialised `runState === "idle"`) and relays nothing from that run, arming at the next idle. `inCompaction` was not touched at all — it is still tested in `observe` before the assistant-text branch, so the streamed compaction summary still cannot escape. There is now a test that pins that.

**Chunking** still applies per message: `chunkMessagingText` is unchanged and `deliverReply` loops over its chunks.

**Reply threading — the run's first message only.** In a group the first outbound chunk of a run threads onto the inbound message and everything after it is a plain send. The reply-to relation exists to say which incoming message is being answered; one of them says it, and repeating it across N progress messages × M chunks stacks a quote header over the whole conversation. Both clients group consecutive messages from one sender anyway, so the follow-ups still read as one block. `threadedThisRun` resets only on a genuine idle→running edge — `task_state` is re-published mid-run for queue and steering changes, and re-arming there would thread every message — which is what still gives `entry.active` a reader now that the buffer reset is gone.

**`buffer` is gone entirely**; `active`, `armed` and `inCompaction` all still have readers.

## Tests

The old suites passed unchanged against the new code, which confirms your suspicion directly: **nothing covered a multi-message run**. Every fake session yielded exactly one `assistantText`, so the joined-vs-separate distinction was invisible. Added, and verified they fail against the old bridge (both new behaviour tests time out on it, in both suites):

- `messaging.test.ts` — *"relays every completed assistant message on its own, as it completes, never joined"*: a three-message run whose last message is held behind a gate. It asserts the first two are already in the chat **while `manager.statusOf()` is still `running`**, then releases the gate and asserts all three arrived in order, none containing `\n\n`. That pins both halves — separate messages, and sent at completion rather than at the end.
- `messaging.test.ts` — *"in a group only the run's first message threads onto the inbound one"*: exactly one `reply` followed by two `send`s.
- `messaging.test.ts` — *"the streamed compaction summary is not a reply and never reaches the chat"*: a run yielding `compactionBegin` → an assistant message → `compactionEnd` → the real answer; only the real answer is sent.
- `messaging-telegram.test.ts` — the mirror pair: gated two-message run (first send observed mid-run, order preserved) and one `replyTo` followed by a plain send in a supergroup.

## Verification

`pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, `pnpm test` across all nine packages — skills 25, docs 53, landing 63, core 1000 (+5 skipped), server **1052** (92 files; `messaging.test.ts` 20 + `messaging-telegram.test.ts` 20 among them), cli 374, web 1359, desktop 106 — plus `pnpm format` + `pnpm format:check`, `git diff --check`, and `sh scripts/test-installer.sh`. All pass.

The bilingual changelog pair now describes the per-message relay, the per-Session send ordering and the one-thread-per-run rule in both languages.

PR #465, opened 2026-08-25
App attribution for OpenRouter and TokenDance, plus a TokenDance provider group

hiyouga:
Model requests now identify PenguinHarness to the gateways that read an app-attribution header, and the built-in catalog gained a TokenDance provider group. The model library's provider groups were resequenced at the same time.

## App attribution

`attributionHeaders(baseUrl)` joined `packages/core/src/state/model-catalog.ts` and is wired into `GenerativeModel`, which passes it to AgentHub as `defaultHeaders` — the constructor option added in `@prismshadow/agenthub` 0.4.7 (already the pinned version, see `changelog/unreleased/2026-08-25-agenthub-0-4-7.md`). AgentHub forwards it to whichever client it routes to, so every protocol is covered by the one call site.

| Endpoint | Headers |
| --- | --- |
| `openrouter.ai` | `HTTP-Referer: https://penguin.ooo/`, `X-OpenRouter-Title: PenguinHarness`, `X-OpenRouter-Categories: cli-agent,personal-agent` |
| `tokendance.space` | `X-App-URL: https://penguin.ooo/` |

Two decisions worth stating:

- **Keyed on the endpoint host, not the provider group.** The group is a display bucket; the header is a property of the server being called. A model filed under `custom` that points at OpenRouter is still PenguinHarness talking to OpenRouter and is attributed the same way. The cost is the mirror case: an entry with no `base_url` of its own gets no headers even when `OPENAI_BASE_URL` sends it to a gateway, because that variable is read inside AgentHub and never reaches this side. Matching is suffix-anchored, so `notopenrouter.ai` is not OpenRouter.
- **Two categories, not four.** OpenRouter accepts at most two categories per request and silently drops slugs outside its fixed list, and `productivity` / `coding` are group headings in its docs rather than slugs. `cli-agent,personal-agent` is what goes on the wire.

`listEndpointModels` was deliberately left alone: a `/models` listing is not an inference call and carries no attribution meaning on either gateway.

## TokenDance

A `tokendance` gateway group reaching `https://tokendance.space/gateway/v1` over Chat Completions (`client_type = "openai-chat"`, blank credential falling back to `OPENAI_API_KEY`, like every other gateway group), with six presets: `deepseek-v4-flash-0731`, `deepseek-v4-flash-vision-exp`, `deepseek-v4-pro-0813`, `glm-5.3`, `kimi-k3`, `qwen3.8-max`.

Context windows and vision flags come from the public catalog API (`GET https://tokendance.space/gateway/v1/models`, no credential needed). Prices are the gateway's own CNY rates from the model detail pages, which need a signed-in session — they cannot be re-read anonymously, so the block comment records the read date and the figures. `qwen3.8-max` is stored at its running 20%-off promotion (list 1.5 / 12 / 36 CNY), matching how the OpenRouter rows already store what the gateway actually bills; a lapse there would silently put the real cost 25% above the catalog.

Projects created before this do not gain the group automatically — presets are copied into `.project_config.toml` at creation and nothing rewrites them. The models page's "sync presets" appends it. Nothing already on disk changes shape, so there is no compatibility shim and no `Breaking` field.

## Provider order

`MODEL_PROVIDERS` was resequenced to DeepSeek, OpenRouter, Fireworks AI, Google Gemini, OpenAI, Anthropic, SiliconFlow, TokenDance, Z.AI (GLM), Moonshot (Kimi), MiniMax, Qwen Pay-As-You-Go, Qwen Token Plan, Custom. The list no longer groups gateways together, so the "gateways first, then first-party vendors" narration moved out of the block comment and the gateway-credential explanation moved onto the array's own doc comment. `MODEL_CATALOG`'s physical clustering is unchanged (it never tracked this order) and the order is read at render time, so nothing on disk moves.

## Verification

`pnpm -r build`, `pnpm format` + `pnpm format:check`, `pnpm typecheck`, and the `core`, `web`, `server`, `cli`, `docs` and `landing` package tests all pass on Node 24.18.

`attributionHeaders` is covered directly (both gateways, the custom-entry case, blank/unparseable base URLs, lookalike hosts, and a catalog invariant that every OpenRouter and TokenDance row carries its scheme). The `GenerativeModel` wiring has no committed test: asserting it means reaching through `AutoLLMClient._client._client._options.defaultHeaders`, two libraries deep. It was verified by hand instead —

```
openrouter      => {"HTTP-Referer":"https://penguin.ooo/","X-OpenRouter-Title":"PenguinHarness","X-OpenRouter-Categories":"cli-agent,personal-agent"}
tokendance      => {"X-App-URL":"https://penguin.ooo/"}
deepseek-direct => undefined
```

— and the field name is typed on AgentHub's constructor options, so a rename fails typecheck.

The TokenDance glyph was rendered at 20px and 40px on light and dark backgrounds; its viewBox is cropped to the mark's ink so it fills an icon box like the square brand glyphs do.


hiyouga:
## Review

The mechanism is sound and I could not break it. `attributionHeaders` is genuinely suffix-anchored — I ran the lookalikes (`evil-openrouter.ai`, `openrouter.ai.attacker.com`, the punycode homograph, and `https://openrouter.ai@evil.com/v1`, which parses to `hostname === "evil.com"`) and every one returns `undefined`; uppercase hosts and userinfo normalise correctly, and since the payload is an app name and a public URL, a false positive could not leak a credential anyway. The wiring is complete too: `defaultHeaders` is a typed constructor option on `AutoLLMClient` in the installed agenthub 0.4.7 and every one of the twelve `dist/*/client.d.ts` declares it, `autoClient.js:53,129` forwards it unconditionally, and all five production `new GenerativeModel(...)` sites pass the entry's own `base_url` — so the session LLM, the post-compaction rebuild, the bare/meta LLM behind titles and the vision describer are all covered, and `GenerativeModelConfig` has no user-headers field for it to collide with. The provider resequencing is display-only in fact and not just in claim: the only two order-sensitive consumers are `defaultExpandedProviders()` (hard-codes `"deepseek"`) and `groupModelRows`, `MODEL_CATALOG`'s clustering tracks neither the old nor the new order, and both docs tables match. Findings below are one contradiction in the pricing prose, one over-claim in a comment, and two silent-false-negative edges.

---

**1. `packages/core/src/state/model-catalog.ts:912` — "like the other gateway groups" is false; two of the five other gateway groups state the opposite rule.**

Only the OpenRouter block stores billed rates (`:311`, "these rows store what OpenRouter actually BILLS ... unlike the direct-vendor rows"). Qwen Token Plan says the reverse 66 lines below this one — `:978`, "official CNY list prices; limited-time promotions such as the 20%/50% off discounts are not stored" — and Qwen Pay-As-You-Go the same at `:1032`. Failure: a maintainer re-reading TokenDance prices next month follows "like the other gateway groups" over to the Qwen blocks and applies the discount rule there, silently under-costing every Qwen session by 20-50%; or reads the Qwen rule first and restores `qwen3.8-max` on TokenDance to list, over-costing it by 25%. The decision itself is fine — it matches OpenRouter — but the sentence generalises it to groups that reject it.

**2. `packages/core/src/state/model-catalog.ts:967` — the catalog's one un-re-readable promotional price is also the one with no expiry recorded.** *(Judgement call — flagging, not changing the number.)*

Every OpenRouter row that stores a discounted rate names the mechanism for noticing the lapse: `pricing.discount` on the endpoints API (`:313`, `:534`), which is anonymous and mechanical. This row's own comment says the opposite (`:908`: the detail pages "need a signed-in session, so they cannot be re-read anonymously"), so the row that will drift is the one nobody can check. Failure: the 20% promotion ends, no signal exists anywhere, and the cost center reports every TokenDance `qwen3.8-max` session 25% low indefinitely — the comment already says exactly this will happen. Storing the list price (as both Qwen groups do for exactly this reason) or recording the promotion's end date would bound it. Your call; I have left the figure alone.

**3. `packages/web/src/components/ui/provider-logo.tsx:97` — the viewBox does not make the mark "fill an icon box like the square glyphs do".**

`viewBox="50.79 100 378.72 280.5"` is 1.35:1, and the wrapper `<svg>` sets no `preserveAspectRatio`, so the default `xMidYMid meet` renders it 24x17.8 inside a 24x24 box, letterboxed top and bottom. The crop is right and the wide ratio matches `fireworks` and `openrouter`, so nothing needs redrawing — but the claim is checkable and false, and the same sentence is in the PR body. Failure: someone trusting it skips comparing the row's visual weight against its neighbours, or a later reader treats the letterboxing as a regression and "fixes" a glyph that was correct.

**4. `packages/core/src/state/model-catalog.ts:1599` — a fully-qualified host with a trailing dot is silently unattributed.**

WHATWG `URL` keeps the dot: `new URL("https://openrouter.ai./api/v1").hostname === "openrouter.ai."`, which is neither equal to `openrouter.ai` nor a `.openrouter.ai` suffix, so `attributionHeaders` returns `undefined`. Failure: a user pastes the trailing-dot form (legal DNS, and the request itself succeeds), and attribution vanishes with no error — the exact outcome the function exists to produce reliably. No leak; only a false negative.

**5. `packages/core/src/state/model-catalog.ts:1589` — a scheme-less base URL is likewise inert.** *(Low; no fix proposed.)*

`new URL("openrouter.ai/api/v1")` throws, so `endpointHost` returns `undefined`. The models page rejects scheme-less URLs before this is reachable through the UI (`packages/web/src/features/models/protocol-types.ts:73`), and a hand-edited `.project_config.toml` carrying one would almost certainly fail the request itself, so this is worth knowing rather than guarding.

---

**Checked and clean, for the record.** `(provider, model_id)` uniqueness holds and the pins are load-bearing: unpinned, the six new ids resolve to `DEEPSEEK_API_KEY` / `ZAI_API_KEY` / `MOONSHOT_API_KEY` and — for `qwen3.8-max` — to nothing at all; with `client_type: "openai-chat"` all six short-circuit to `OPENAI_API_KEY` before any of those branches. `fastModeProtocol` returns `"openai"` for all six, identical to SiliconFlow's existing `zai-org/GLM-5.2` row, so the gateway-versus-direct fast-mode divergence is pre-existing and not this diff. Ports are ignored by the host match, which is correct — a different port on the real host is still the real host. Changelog pairs are format-compliant (metadata order, `Scope` values, counterpart links, section-for-section parity, no reasoning sections). Docs provider tables are in sync with the new order in both languages, and `packages/docs` `toc`/`search`/`content` tests still pass with the new section. Leaving `listEndpointModels` unattributed is right — `GET /models` is not an inference call. Every factual claim in the PR body verified except the glyph sentence in finding 3, including the agenthub 0.4.7 pin, the changelog cross-reference, and "`MODEL_CATALOG`'s physical clustering ... never tracked this order" (it is `deepseek > openrouter > fireworks > siliconflow > tokendance > qwen-token-plan > qwen-pay-as-you-go > minimax > google > anthropic > openai > zhipu > moonshot`, matching neither sequence).

Fixing 1, 3 and 4 in a follow-up commit on this branch; 2 and 5 left as-is.

hiyouga:
Fixed findings 1, 3 and 4 in `fff481e9` (a new commit on top — nothing rebased, so the branch stacked on `251072e1` is unaffected).

- **1** — the TokenDance discount note now names the OpenRouter rows as the peer it follows and points out that both Qwen groups take the opposite rule, so neither can be carried across to the other.
- **4** — `endpointHost` strips a fully-qualified trailing dot, so `https://openrouter.ai./api/v1` is attributed instead of silently skipped. The comparison stays suffix-anchored; a new assertion pins that `openrouter.ai.attacker.com.` is still refused, alongside the positive case.
- **3** — the glyph comment now says the cropped viewBox spans the icon's full width and letterboxes at 1.35:1, as the other wide brand marks do, rather than claiming it fills a square box.

Left alone as deliberate: **2** (the `qwen3.8-max` promotional rate — your call whether an un-re-readable figure with no expiry is worth the 25% drift) and **5** (scheme-less base URLs, unreachable through the UI).

Verified on Node 24.18: `pnpm --filter @prismshadow/penguin-core test` 993 passed / 5 skipped, `pnpm --filter @prismshadow/penguin-web test` 1296 passed, `pnpm --filter @prismshadow/penguin-docs test` 53 passed, `pnpm --filter @prismshadow/penguin-server exec vitest run test/models.test.ts` 22 passed, `pnpm typecheck` clean across all nine packages, `pnpm format` + `pnpm format:check` clean, `git diff --check` clean.

hiyouga:
Added a seventh TokenDance preset: `glm-5.3-flash` (GLM-5.3 Flash), 1M context, vision, `client_type: "openai-chat"` on the group base URL.

- **Protocol pin.** The public catalog API (`GET https://tokendance.space/gateway/v1/models`, no credential) reports `supported_protocols: ["openai:chat-completions"]` for this id alone — every other row in the group also advertises `openai:responses` and/or `anthropic:messages` — so the `openai-chat` pin is the only shape it serves.
- **Vision.** TokenDance describes it as the GLM-5 series' first natively multimodal model, and this group routes through AgentHub's generic `openai-chat` client, which converts `image_url` parts.
- **Pricing — list, not billed.** Stored at TokenDance's official list price, CNY 0.23 cache hit / 0.8 input / 2.8 output → `cny(0.23, 0.8, 2.8)`, with `cache_write` carrying the input price per the group's no-cache-write-fee convention. TokenDance is currently running a limited-time 50% off on it (its catalog entry is tagged 【官方同价，限时 2 周五折】) which is deliberately **not** stored.

That makes the group mixed — `qwen3.8-max` still stores the 20%-off rate it is actually billed — so the block comment no longer claims the whole group holds billed rates; it now states the convention per row. `qwen3.8-max`'s stored numbers are unchanged.

Also folded into the existing changelog pair (six presets → seven) rather than opening a new entry.

**Verification:** `pnpm build`, `pnpm typecheck`, core `model-catalog.test.ts` (28 passed), web `model-grouping.test.ts` + `protocol-path.test.ts` (31 passed), server `models.test.ts` (22 passed), `pnpm format:check`, `git diff --check` — all green. No change needed in `protocol-path.ts` (the path is `/chat/completions`), no glyph-map change (no new group), and `resolveModelEnv` resolves on the pinned `openai-chat` → `OPENAI_*`, matching the rest of the group.

hiyouga:
Two follow-ups pushed: the pricing convention is now uniform across the TokenDance group, and `main` is merged in.

## Pricing: the whole group stores list prices

The group was internally inconsistent — `qwen3.8-max` held the discounted rate it is billed while `glm-5.3-flash` held the list price. Every row now stores TokenDance's official CNY **list** price, the same convention the two Qwen groups below it already follow.

- `qwen3.8-max`: `cny(1.2, 9.6, 28.8)` → `cny(1.5, 12, 36)`, i.e. `[0.214286, 1.714286, 5.142857]` USD/Mtok after the 7:1 conversion. `model-catalog.test.ts`'s pinned row moves with it.
- The group comment no longer describes the group as mixed. It states the single convention, keeps both promotions named with their rates (`glm-5.3-flash` on a 2-week 50% off, `qwen3.8-max` on a limited-time 20% off) so the stored figures stay auditable, and names the re-check mechanism the group previously lacked: `GET https://tokendance.space/gateway/v1/models` needs no credential and opens a discounted model's `description` with a bracketed 限时 tag. Verified live — `qwen3.8-max` carries 【指定服务商限时调用八折】 and `glm-5.3-flash` 【官方同价，限时 2 周五折】.
- `glm-5.3-flash`'s per-row note lost its now-redundant restatement of the convention; its multimodal and single-protocol facts stay.
- OpenRouter, Fireworks AI and SiliconFlow are untouched — their documented convention is to store what the gateway actually bills.

No on-disk data changes: presets are copied into `.project_config.toml` at Project creation and nothing rewrites them; the models page's "sync presets" is the only path that updates catalog-owned fields. Both changelog entries say so now, in both languages.

## Conflict with main

`packages/web/src/features/models/models-page.tsx` was the only file both sides touched. #467 re-indented ~300 lines of the group header when it wrapped the `<section>` in a `relative` div for the drop indicator, which is why the conflict looked larger than it was.

Resolved by taking `main`'s structure whole and re-applying this branch's one hunk into it: the header row keeps `{...drag.header}`, the `cursor-grab` class, the drop indicator, and the `if (e.target !== e.currentTarget) return;` guard in `onDragStart` exactly as #467 wrote them, and the OAuth button is re-inserted between "添加模型" and "手动设置密钥". The resolved file is byte-identical to `main`'s except for the five additive hunks this branch already had over the merge base.

Nothing else conflicted or clashed semantically: `model-grouping.ts` (`groupOrder` parameter), `model-group-order.ts` and `model-select.tsx` come from `main` alone; the i18n dictionaries and `packages/core/src/state/model-catalog.ts` from this branch alone — `main` did not touch the catalog.

## Verification

`pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck` — all clean. `penguin-core` 997 passed / 5 skipped, `penguin-web` 1316 passed (99 files, `model-group-order.test.ts` 20/20 among them), `penguin-server` 1002 passed, `docs` 53 passed. `pnpm format` made no changes, `format:check` clean, `git diff --check` clean.

Driven through the Playwright e2e harness (mock LLM, scratch `PENGUIN_HOME`, `deviceScaleFactor: 2`, zh-CN):

- The TokenDance header renders **添加模型 · 自动获取密钥 · 手动设置密钥 · 测速 · 前往密钥管理 ↗**, in that order. DeepSeek — a group with no key-minting flow — renders the same four minus 自动获取密钥, so the button really is driven off the catalog descriptor.
- Dragging the TokenDance header row to the top reorders the groups (`DeepSeek, OpenRouter, …, TokenDance, …` → `TokenDance, DeepSeek, OpenRouter, …`) and the order survives a reload.
- A drag started on the "前往密钥管理 ↗" link leaves the order untouched, before and after a reload. Instrumented so the negative is not vacuous: the gesture does reach `dragstart` with `target` = `<A>`, and the guard turns it away.
- `qwen3.8-max`'s card reads `1M · $0.2143 / $1.7143 / $5.1429`.

The throwaway spec and its screenshots are not committed.

hiyouga:
Second round: `main` moved again (https://github.com/Prism-Shadow/penguin-harness/pull/469, `7bbe1846`) and has been merged in. `pnpm install --frozen-lockfile` re-run against its `@prismshadow/agenthub` `^0.4.8` floor.

**One conflict**, in `packages/core/src/state/model-catalog.ts`, and only in the file header's "data verified as of" list — both sides appended a date. Kept both: `… the TokenDance group: 2026-08-25, its glm-5.3-flash row: 2026-08-26; the GLM-5.3 Flash rows (direct + OpenRouter) and qwen3.8-flash: 2026-08-26 …`. #469's rewritten `glm-5v-turbo` exclusion rationale (AgentHub's GLM client now gates image support per model) came through untouched.

Everything else merged cleanly and was checked rather than assumed:

- `packages/core/test/model-catalog.test.ts` — both sides' exact-order assertions survived intact: the `zhipu` list carries `glm-5.3-flash`, `openrouter` carries `z-ai/glm-5.3-flash`, `qwen-pay-as-you-go` carries `qwen3.8-flash`, and the `tokendance` block keeps its seven rows with the recomputed `[0.214286, 1.714286, 5.142857]`. The suite passes, which is the real check on the merged order.
- `packages/core/package.json`, `packages/cli/package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml` — `main`'s side wholesale; this branch has no opinion about the dependency.
- `packages/docs/content/models.{en,zh}.md` — both sides' edits present (`tokendance` in the gateway-group table and the `OPENAI_*` sentence from here, the three new ids from #469).

**One silent semantic clash found and fixed** (`4a6f6e02`): #469 wrote that `glm-5.3-flash` "appears twice" — the direct Z.AI row and OpenRouter's `z-ai/glm-5.3-flash`. True of the catalog it was written against; wrong once this branch's TokenDance group lands a third row under the bare id. Both languages now say three, name the TokenDance row alongside the OpenRouter one as the routes whose generic `openai-chat` client carries images for any id, and split the price clause three ways (Z.AI list / TokenDance list / OpenRouter's discounted rate). All three rows are `vision = true`, so nothing about vision needed correcting — and #469's own changelog entry is left as written, since it accurately records what that PR did.

## Re-verified on the merged result

`pnpm build`, `pnpm typecheck` clean. `penguin-core` 997 passed / 5 skipped, `penguin-web` 1316 passed, `penguin-server` 1002 passed, `docs` 53 passed, `skills` 25 passed. `pnpm format` no changes, `format:check` clean, `git diff --check` clean.

Playwright, same harness as before (mock LLM, scratch `PENGUIN_HOME`, `deviceScaleFactor: 2`, zh-CN):

- TokenDance header still **添加模型 · 自动获取密钥 · 手动设置密钥 · 测速 · 前往密钥管理 ↗**; DeepSeek still the same four minus 自动获取密钥.
- Header drag still reorders and persists across a reload; a drag started on 前往密钥管理 ↗ still leaves the order untouched, with `dragstart` observed firing at `<A>`.
- TokenDance's `qwen3.8-max` card reads `1M · $0.2143 / $1.7143 / $5.1429`, and #469's rows render alongside unaffected — Z.AI (GLM) now 5 models with `glm-5.3-flash` at `$0.03 / $0.15 / $0.5`, OpenRouter 39, Qwen Pay-As-You-Go 6.

PR #466, opened 2026-08-25
The CLI becomes a thin client of the server, and agents get a local control plane

hiyouga:
## What

Rebuilds the CLI on the server API and adds the local-control plumbing that lets agents drive the harness through that CLI.

### CLI

- `run` / `chat` now execute on the server: the CLI parses args, sends HTTP requests, subscribes to the session SSE stream via fetch (dependency-free parser, `Last-Event-ID` reconnects, heartbeat-tolerant) and feeds OmniMessages to the existing renderer. Approval requests arrive as `approval_request` server events, answered through `POST /approvals/:toolCallId`; Ctrl-C posts `/abort`; REPL typing mid-run posts `/steer`; `/compact` posts `/compact`.
- New commands: `ls`, `input <session>`, `logs <session>`, `agent ls|create`, `project ls`, `cost`, `schedule ls`. Session arguments accept a full id or a unique fragment (the 8-hex tail `ls` prints); ambiguity errors list candidates. `run` gains `--session`, `--background`, `--json`, `--goal`, `--server`.
- Connection resolution: `--server` > `PENGUIN_API_URL` > live `server.lock` at the data root > auto-start (detached `node <cli entry> server` with `PORT=0`, logs to `<root>/logs/server-auto-<date>.log`; a lost spawn race attaches to the winner's lock). Token: `PENGUIN_API_TOKEN` > `<root>/api-token` (loopback only; 401 re-reads once); remote targets require the env var explicitly.
- Defaults: `--project-id` ← `PENGUIN_PROJECT_ID` ← `default_project`; `--agent-id` ← `PENGUIN_AGENT_ID` ← `default_agent`; workspace = the CLI's cwd; approval mode stays allow-all.

### Agent-side ergonomics (second round)

- **Caller-context defaults**: session creation inside a harness agent (`PENGUIN_SESSION_ID` present) defaults each unspecified field to the calling session's live values (`GET /api/sessions/$PENGUIN_SESSION_ID`) — workspace, model pair, approval mode, thinking level — the `run_subagent` inheritance applied to the CLI surface. Per field: flag > caller > plain fallback; failed lookup warns (dim) and falls back; outside an agent nothing changes.
- **`--timeout <duration>`** on `run` (non-background), `input`, `logs -f`: soft-yield wait budget mirroring the command tools' yield window — on expiry detach cleanly (exit 0, dim still-running line; `status:"running"` under `--json`), task keeps running server-side. Accepts `30s`/`5m`/`2h` or bare seconds; conflicts with `--background`/`--no-wait` are rejected.
- **Bare `penguin input <session>` (no `-m`) polls**: prints the most recent complete assistant text (idempotent snapshot, thinking/tool output skipped), mirroring `input_subagent`'s empty prompt; waits out a running turn first (bounded by `--timeout`); `--no-wait` without `-m` rejected. A watcher with no approval prompt now leaves pending approvals untouched.
- Smoked live against a real model: bare poll returned the last reply; a `PENGUIN_SESSION_ID`-simulated caller had its workspace/model inherited by the child session; a `sleep 25` task soft-yielded at 6s (exit 0), polled as still-running at +3s, and polled "Done." after finishing.

### Server

- **Local API token**: minted per boot (`auth/api-token.ts`), written to `<root>/api-token` (0600, atomic), installed on the runtime AuthService; `authMiddleware` accepts `Authorization: Bearer` (constant-time compare) as the built-in admin — SSE endpoints included; `sessionVia: "token"`. The hmr comment about the earlier removed on-disk token now records the deliberate reversal: local filesystem access to the data root **is** admin authority (the reset-admin-password rule), and agents driving their own server through the CLI is the feature.
- **Control env injection**: server-driven sessions inject `PENGUIN_API_URL` / `PENGUIN_API_TOKEN` / `PENGUIN_PROJECT_ID` / `PENGUIN_AGENT_ID` / `PENGUIN_SESSION_ID` into every tool subprocess through the new `CreateAgentOptions.controlEnv` seam (a policy getter mirroring `proxyEnv`, bound per Session, re-evaluated per spawn, overriding vault entries of the same name; subagents evaluate it with their own ids; SDK-direct embedding injects nothing). Threaded through both core entry paths (loader + SessionService).
- **CLI-session toggle retired**: `POST .../sessions` takes an optional `client` hint (`"cli"` from the CLI, default `"web"`) stored on the existing column as provenance; list endpoints stop filtering (`cli=1` removed on sessions and agent-traces, `webOnly` removed from the repo, `includeCli` gating removed from trace-service); per-list adoption is replaced by a **startup adoption sweep** (`adoptUnmanagedTraceSessions`, fired at platform create, mtime-gated, idempotent); `UiPrefs.showCliSessions` removed (stale stored keys are ignored by the shallow merge).

### Web

Removed the "Show CLI sessions" toggle, its state plumbing in `state/sessions.tsx`, the `cli` params in `api/endpoints.ts`, and the strings in both dictionaries. (Rebased over #462, which removed the standalone trace page — the trace-tree cli plumbing went with it.)

## Verification

- `pnpm -r build`, `pnpm -r typecheck`, `pnpm format:check`, `git diff --check` — clean.
- Tests: core 992 (new: controlEnv injection into command env, precedence over vault, per-spawn re-read), server 987 (new: Bearer auth incl. SSE + hmr gate + CSRF interaction, token file 0600/rotation, adoption sweep, client field; rewritten: cli=1 removal), cli 327 (new: connection resolution order, SSE parser, fragment resolution, full command wiring against an in-process fake server incl. the chat REPL end-to-end), web 1296.
- End-to-end smoke on a scratch `PENGUIN_HOME` with the built CLI and a real model: `penguin ls` with no server auto-started one and listed; `penguin run -m "say hi"` streamed, rendered and exited 0; a run that echoed `$PENGUIN_API_URL $PENGUIN_SESSION_ID …` proved control-env injection; a run whose agent executed `penguin ls` **from inside its own session** listed itself as running (the headline loop); `penguin input` / `logs --tail` / `agent create` / `project ls` / `cost` / `schedule ls` all exercised against the live server.

Design docs PR: Prism-Shadow/penguin-harness-design (docs/cli-on-server).


hiyouga:
## Merged `main` in — one conflict, resolved

`main` moved to `6446684a` (through #467, #471, #469, #465, #473, #463, #472) and it is now merged into this branch (`086b820d`). `git merge` reported exactly **one** conflict:

**`packages/core/src/environment/tools/command/session-manager.ts`**, between #471's note on the child-process env spread and this branch's control-var injection layer.

The two sides were never in tension — #471 documented the seam in anticipation of this layer, and its PR body says so outright ("Control-var injection (PR #466) will cover the 'call the same harness's API' need without any inheritance"). Both survive.

### The spread order

```ts
env: {
  ...hostEnvForChild(this.proxyEnv?.() ?? null),
  ...this.vault,
  ...(this.controlEnv?.() ?? {}),
  ...HARDENED_ENV,
},
```

`hostEnvForChild` is where the strip runs: every `PENGUIN_*` by prefix plus `PORT` / `HOST` / `FORCE_COLOR` / `CLICOLOR_FORCE`, removed rather than blanked, matched case-insensitively. Because it only ever governs the copy of the host env, everything spread in after it — the vault, then `controlEnv` — carries authoritative values into the child, stripped names included. That ordering is what makes `PENGUIN_API_URL` / `PENGUIN_API_TOKEN` / the Session coordinates reach an Agent's commands while no inherited copy of any `PENGUIN_*` does. Reversing it would leak the harness's own data root into every command an Agent runs, which is the regression #434 and #471 exist to prevent.

The guard is core's `an explicit injection layered after the strip wins, for any PENGUIN_* name` (`packages/core/test/exec-session.test.ts`). It passes, alongside this branch's three `controlEnv` tests.

### The comment

#471's reasoning is load-bearing and was kept, but its forward-looking phrasing ("An injection layer added to this spread later composes with the strip for free") now describes code sitting directly beneath it, so it is restated in the present:

> The strip governs inheritance only: it runs inside `hostEnvForChild` and never re-applies to entries spread in after it, so the vault and `controlEnv` carry authoritative values into the child — stripped names, `PENGUIN_*` included, and all. Both guarantees hold because of that order: no `PENGUIN_*` reaches a command by inheritance, while the control variables the host does sanction arrive as injected values rather than as surviving copies. Pinned by the "an explicit injection layered after the strip wins" test.

### Also in this batch

#463 landed the shipped `penguin-orchestration` skill, written against this branch's CLI surface as of 2026-08-25. Its command block still spelled `input <session_id>` and `logs <session_id>` as required and listed `--agent-id` on neither — but `6556d4f6` on this branch made the id optional (omitted, it is the agent's most recent session, with `--agent-id` picking whose). Since this PR is what changes that surface, the skill is updated here rather than in a follow-up (`97ec7d4f`, skill version 1 → 2), and both changelog files gained a line for it. `packages/docs/content/cli.*.md` already documented the optional form on this branch and needed no change.

Nothing else conflicted, and no on-disk data or config shape changed as a result of the merge — the compatibility entry stands as written.

### Verification (Node 24.18.0, full CI chain)

| Step | Result |
| --- | --- |
| `pnpm install --frozen-lockfile` | clean (agenthub floor `^0.4.8`, `pnpm-workspace.yaml` exemption from main) |
| `pnpm build` | pass, 9 packages |
| `pnpm typecheck` | pass, 8 projects |
| `pnpm test` | pass — 3970 passed, 5 skipped, 0 failed across the 8 test-bearing packages |
| `pnpm format` + `pnpm format:check` | pass, no rewrites |

Targeted, on top of the full run:

- `packages/core` `exec-session.test.ts` — 31/31, including the strip-versus-injection guard and the three `controlEnv` tests
- `packages/cli` `dev-entry-isolation.test.ts` (#471's port/data-root drift guard) — 5/5
- `packages/skills` 25/25 and `packages/docs` 53/53 (incl. `skills-sync.test.ts`) after the skill edit

No known-flake failures appeared locally; ci-windows is the only place the two known ones (`Failed` worker crash, `environment.test.ts` truncation timing) can show up.

### CI on `97ec7d4f`

All checks green on the first run — no rerun needed, and neither known ci-windows flake appeared:

`build` · `ci` · `ci-macos` · `ci-windows (packages)` · `ci-windows (server)` · `runtime (ubuntu | macos | windows)` — pass. `deploy` skipped.

PR #467, opened 2026-08-25
Drag the models page's provider groups into your own order

hiyouga:
The model library's provider groups can be dragged into a manual order, the way the chat sidebar's groups already could (#440). The order is kept per Project in `localStorage`, survives a reload, and the chat page's model picker follows it.

Drag-only, per the requested scope: no pinning. With a dozen-odd groups, dragging the two you actually use to the top is the whole of what pinning would buy, and it keeps a sixth icon off a group header that already carries five actions.

## What reuses what

The splice is `session-order.ts`'s `moveInSequence`, and `readStringArray` / `writeStringArray` are the storage body. New here is `features/models/model-group-order.ts` — key namespace, load/save, and the two halves of the placement rule — plus the `groupOrder` parameter threaded through `model-grouping.ts` and the drag wiring on the group header.

`model-grouping.ts` stays pure: every entry point takes an optional `groupOrder` and the caller supplies it. That is what keeps `orderModelsLikeLibrary`'s documented contract — "rows ordered exactly as the model page shows them" — true once a user has arranged their groups, and it is why the chat picker follows for free.

## Decisions

- **An unplaced group TRAILS.** This is the opposite of the sidebar's rule, and the difference is what an unplaced group *means* on each side. There, it is a Workspace or Agent that would be lost at the bottom of a long list, so it surfaces. Here, it is almost always a group the user just created — through a control that sits below every group on the page — so leaping to the top would leave the part of the page being looked at. Trailing also matches the automatic order, which already puts user-defined groups after the built-in ones.
- **`orderModelGroups` and `commitModelGroupOrder` are one rule stated twice**, so they live side by side. A drop that materialised unplaced groups at the opposite end from where they render would shove groups the user never touched.
- **A drop commits against every group the library could show**, not the groups on screen. `groupModelRows` drops a built-in group holding no models, so committing only what renders would leave the empty ones unplaced and make each one a newcomer the day a model is added to it. `allGroupKeys` is that full sequence.
- **A private MIME type, distinct from the sidebar's.** The sidebar renders alongside this page, so one shared `application/x-penguin-group-key` would let a Workspace group dragged out of the conversation list paint a drop line on the models page and commit on release. Two unrelated orders that happen to be dragged the same way.
- **Refused while searching.** The query filters the page to matches; an arrangement made against that subset is not the arrangement of the library.
- **Order and expansion stay separate arrays** under separate keys, so a drag never folds or unfolds a group.
- **Nothing prunes stale keys**, following `group-order.ts`'s reasoning: a key matching no group is one failed membership lookup, and a provider that leaves the catalog and returns resumes its place.

## Known limits

The gesture is mouse-only — HTML5 drag-and-drop never fires from touch, and there is no keyboard equivalent. A stored order still *applies* everywhere, since the arrangement is implicit and per Project, so a phone renders what its owner arranged at a desk. This matches the sidebar's group drag; giving either one a keyboard path is a separate change that should do both.

## Verification

`pnpm -r build`, `pnpm format` + `pnpm format:check`, `pnpm typecheck`, and the full `web` suite (99 files, 1316 tests) pass on Node 24.18.

`test/model-group-order.test.ts` covers the storage (per-Project keys, no-Project inert, malformed values, storage that throws from the `useState` initializer path) and the ordering (empty order is the identity, unplaced groups trailing, stale keys inert, custom's always-shown rule, search still filtering, the picker agreeing, and a drop leaving an empty provider at its catalog place).

The gesture itself was driven by hand against a real server through the e2e harness, since neither the sidebar's group drag nor this one has a committed e2e spec. Dragging Anthropic onto the top half of DeepSeek's header moved it from position 8 to position 1, one drop indicator was present mid-drag (902x2.25 at the group's top edge, `--accent-bg`, `z-10`), and the order was identical after a reload:

```
BEFORE:   DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go
AFTER:    Anthropic, DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan
RELOADED: Anthropic, DeepSeek, OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan
```

A second run covered the placement rule: after that drag, a user-defined group named `aaa-my-gateway` — early enough alphabetically that name-sorting and trailing disagree — was created and landed last, after Custom, with the dragged Anthropic still first.

The throwaway specs were removed; only the unit tests are committed.


hiyouga:
## Review

The ordering algebra is sound. I traced `orderModelGroups` against `commitModelGroupOrder` for the cases the tests do not spell out — stale keys interleaved with live ones, an empty built-in sitting between the drag and its target, a user group created after an order was already stored, and the no-op identity path — and the two halves agree in every one of them, so what the drop indicator promises is what re-renders. The materialised tail (`allGroupKeys` order) matches the unplaced tail (`[...builtin, ...extras]`), which is the invariant that makes the trailing rule survive a second drop. The drag wiring reproduces the sidebar's `dragover` / `dragleave`-with-`relatedTarget` / `types`-authorises-the-drop pattern faithfully, including the guard that makes a `dragGroup` stranded by a `canDrag` flip harmless. The 300-line hunk in `models-page.tsx` is a re-indent plus the wrapper and nothing else — I compared the token multiset of the whole `groups.map` body against `main` and there are no removals. Findings below, most severe first; only the first is a functional misfire.

---

**1. `packages/web/src/features/models/models-page.tsx:1040` — the header's external link is still natively draggable, so a link drag is silently converted into a group reorder.**

The header `div` is now `draggable`, but `<a href={group.provider.apiKeyUrl}>` inside it is draggable by default. The drag source resolves to the *anchor*, not the header; `dragstart` fires at the anchor and bubbles into the header's `onDragStart`, which then calls `setData(MODEL_GROUP_DRAG_MIME, …)`, overwrites the link drag's `effectAllowed` with `"move"`, and sets `dragGroup`. From then on the drag satisfies `isModelGroupDrag`, so every other header paints an indicator and commits on release.

Failure scenario: on the OpenAI header, press on "Get API key ↗" and drag it toward a new tab. The drag carries `text/uri-list` *and* the reorder MIME with `effectAllowed = "move"`, so the link drop it was aimed at is refused (a link drop needs `copy`/`link`); release it anywhere over another group header instead and the model library silently rearranges — OpenAI jumps to that group's edge and the new order is written to `localStorage`. The user never touched the drag handle.

The sidebar's group headers have no anchor, which is why the pattern it was copied from does not hit this. A one-line guard in `onDragStart` covers this and any future natively-draggable child, since `dragstart` fires *at* the source node: bail unless `e.target === e.currentTarget`.

**2. `packages/web/src/features/chat/model-select.tsx:187` — the comment is not true at HEAD: the order is re-read on every render, not on each open.**

`loadModelGroupOrder` is called in the render body, so it runs a `localStorage.getItem` + `JSON.parse` on every render of `ModelMenuList` — and the component now subscribes to `useProject()`, whose context value is rebuilt whenever the project store changes (agent reloads, project-list refresh), so it re-renders for reasons unrelated to the picker as well.

Failure scenario: nothing breaks, but typing eight characters into the picker's search box performs eight storage reads and eight parses, and a reader who takes the comment at its word will look for a remount that is not what gates the read. `useMemo` keyed on the project id makes the code do what the comment says.

**3. `packages/web/src/features/models/model-group-order.ts:77` — `orderModelGroups` is `applyManualOrder` with the two output halves swapped. Judgement call, not a defect.**

Compare with `packages/web/src/lib/session-order.ts:133`: the `order.length === 0` identity return, the position `Map`, the single partition loop, and the stable sort by stored position are identical line for line; the only difference is `[...placed, ...unplaced]` versus `[...unlisted, ...listed]`. That is two representations of one rule, and the placement difference the PR body argues for is one boolean. A shared primitive in `session-order.ts` taking "unplaced first | last" — the way `orderWithinPinPartitions` already factors the axis-agnostic half out — would leave one implementation of the sort.

I have not applied this: it edits the sidebar's ordering path from a models-page PR, and the header comment already documents the divergence as intentional, so whether the cost of a second copy is worth paying is the author's call.

**4. `packages/web/src/features/models/models-page.tsx:516` — change narration in code that has no previous state.**

"a `dragGroup` left behind by a header that unmounted mid-drag **can no longer** make the page swallow an unrelated drag" describes a fix to something that never shipped — the module is new in this PR. The repo's rule keeps the counterfactual present ("without this, X happens") and drops the narration; the sentence is one word away from being in that form.

**5. `packages/web/src/features/models/model-grouping.ts:122` — the order stated in `orderModelsLikeLibrary`'s doc is backwards.**

"built-in provider groups in MODEL_PROVIDERS order, then user-defined groups, custom last" — `custom` is the *last entry of `MODEL_PROVIDERS`*, and this file's own `groupModelRows` appends the user-defined groups after it. The new test asserts exactly that (`[…, "custom", "alpha-vendor", "zeta-vendor"]`). The PR rewrote this line, so it is fair game; "in MODEL_PROVIDERS order (custom last), then user-defined groups" is the accurate reading.

---

**PR-body claims I checked and that hold:** the splice is `moveInSequence` and the storage body is `readStringArray`/`writeStringArray`; `model-grouping.ts` takes `groupOrder` at every entry point and loads nothing itself; a drop commits against `allGroupKeys`, not the rendered set, so an empty built-in keeps its catalog place; the MIME is distinct from the sidebar's and `isFileDrag` still ignores both; the drag is refused while searching while the stored order still applies to the filtered view; order and expansion are separate keys; nothing prunes stale keys, and they are inert. The three `ModelSelect` hosts (`chat-input`, `project-dialogs`, `schedules-tab`) are all inside `ProjectProvider` and all resolve to `currentProject`, so the new `useProject()` neither throws nor reads another Project's array.

**Checked and clean:** the drop-indicator wrapper does not clip (`space-y-3` gap is 13.5px at this app's 18px root, the line sits at 6.75px), the indicator is `pointer-events-none` so it cannot strobe its own `dragleave`, `dragstart` is a discrete-priority event in React so `dragGroup` is committed before the first `dragover` (no stale closure in `onDrop`), the changelog pair is section-for-section parallel with the metadata block in the required order, and `pnpm --filter @prismshadow/penguin-web test` (99 files / 1316 tests), `pnpm typecheck`, `pnpm format:check` and `git diff --check` all pass on the branch as it stands.

hiyouga:
Pushed `01ac0521` with fixes for findings 1, 2, 4 and 5. Finding 3 (the `applyManualOrder` / `orderModelGroups` duplication) is left alone — it is your call.

**1 — the link drag.** Confirmed in real Chromium before fixing, with a page reduced to the same shape (a `draggable` header holding a button and an `<a href>`):

```
start on the anchor: dragstart target=a1  same=false  effectAllowed=uninitialized  types=[text/uri-list,text/plain,text/html]  -> DROP on the other header
start on the button: dragstart target=h1  same=true   effectAllowed=uninitialized  types=[]                                    -> DROP on the other header
```

So the anchor really is the drag source, its native payload is already on the DataTransfer when the header's handler runs, and the reorder commits. Since `dragstart` fires *at* the source node, `e.target === e.currentTarget` is exactly "the header row itself was picked up", and the guard is one line in `onDragStart`. Re-running the probe with the guard in place: the anchor drag produces no drop on the other header (the link keeps its own payload and `effectAllowed`), the button drag still reorders. A drag begun on a text selection inside the header now also declines to reorder, which is the same correct answer.

No changelog change: the entry already says the group header is the drag handle, which is what the code now does.

**2** — `useMemo` on the project id, so the read happens once per open as the comment claimed rather than once per keystroke.

**4** — the drop-authorisation comment is now in the counterfactual present ("without it, … would …").

**5** — `orderModelsLikeLibrary`'s doc now reads "MODEL_PROVIDERS order (custom last), then user-defined groups".

Verification on `01ac0521`, Node 24.18.0: `pnpm --filter @prismshadow/penguin-web test` — 99 files / 1316 tests passed; `pnpm typecheck` — all 8 packages clean; `pnpm format` then `pnpm format:check` — clean; `git diff --check` — clean.

PR #469, opened 2026-08-26
feat(model-catalog): add GLM-5.3 Flash (direct + OpenRouter) and Qwen 3.8 Flash presets

hiyouga:
Three low-cost rows join the built-in model catalog, each verified against the vendor's own page on 2026-08-26.

## Rows

| Provider | Model id | Context | cache_read / cache_write / output (USD per MTok) | Vision |
| `zhipu` | `glm-5.3-flash` | 1,000,000 | 0.03 / 0.15 / 0.50 | `false` |
| `openrouter` | `z-ai/glm-5.3-flash` | 1,048,576 | 0.015 / 0.075 / 0.25 | `true` |
| `qwen-pay-as-you-go` | `qwen3.8-flash` | 1,000,000 | CNY 0.1 / 1 / 3 via `cny()` | `true` |

## Why the two GLM-5.3 Flash rows disagree

**Price.** The catalog's stated convention is that direct-vendor rows record the vendor's list price while gateway rows record what the gateway actually bills. Z.AI lists $0.15 input / $0.03 cached input / $0.50 output and is running a 50% promotion through 24:00 on 2026-09-09 (UTC+8), so the direct row keeps the list price and its comment names the promotion and its expiry, keeping the stored figure auditable. OpenRouter bills the promotional rate (50% off via ZAI through 2026-09-09 16:00 UTC — the same window in UTC), so the gateway row stores $0.015 / $0.075 / $0.25 and names $0.03 / $0.15 / $0.50 as the rates to restore when it lapses. A lapsed promotion silently doubling the real cost is exactly the drift that hit the `gpt-5.6-terra` / `gpt-5.6-luna` rows before, which is why the OpenRouter section requires the note.

**Vision.** GLM-5.3 Flash is natively multimodal (images, video, files), but the two rows take different routes into AgentHub and only one of them can carry an image:

- The direct row auto-routes on its `glm-5` substring to AgentHub's unified GLM client, which throws `"GLM-5 does not support image inputs."` on any `image_url` part (`src_ts/src/glm5_3/client.ts` on agenthub `main`, checked for this PR). That is the same limitation that keeps `glm-5v-turbo` out of the catalog entirely, and core pins `@prismshadow/agenthub ^0.4.7`, which does not have GLM image input. So the row is `vision = false`, and the code comment states the concrete condition for flipping it: a release inside the pinned range whose `glm5_3` client forwards `image_url` parts.
- The OpenRouter row pins `client_type: "openai-chat"`, and AgentHub's generic Chat Completions client does convert `image_url` parts, so vision genuinely works on that path. `vision = true`.

Marking the direct row `true` today would send images down a route that rejects them at request time; marking the gateway row `false` would hide a capability that works.

## Routing

- `glm-5.3-flash` carries the `glm-5` substring, so `resolveModelEnv` returns the `ZAI_*` pair and `fastModeProtocol` returns `undefined` (the GLM client raises `UnsupportedParameterError`) — both fall out of the existing rules with no change, and a test pins the env pair.
- Both gateway rows pin `openai-chat` plus their preset base URL, as their siblings do. The pin is load-bearing on each: unpinned, `z-ai/glm-5.3-flash` would match `glm-5` and reach the first-party GLM client aimed at OpenRouter, and `qwen3.8-flash` would match no branch at all and throw.
- No new id is a harmful lookalike: `qwen3.8-flash` matches none of AutoLLMClient's substrings, and `glm-5.3-flash` matches only the branch it should.
- `protocol-path.ts` needed no change — all three rows resolve to `/chat/completions`, the direct one through the `zhipu` group default and the gateways through their `openai-chat` pin. `provider-logo.tsx` needed none either: no group is new.
- `modelHomepageUrl`'s `docs.z.ai/guides/llm/<id>` pattern still lands correctly — `guides/llm/glm-5.3-flash` 308-redirects to the model's `guides/vlm/` page.

## Ordering

Each row sits where the file's rule puts it (dictionary order by model id within a provider, newer versions of a series first): `glm-5.3-flash` after `glm-5.3`, `z-ai/glm-5.3-flash` after `z-ai/glm-5.3`, `qwen3.8-flash` before `qwen3.8-max`. The catalog test's exact-order assertions for all three groups were updated, and the header's "data verified as of" list gained 2026-08-26 for these rows.

## Shipped skill

The `agenthub-models` skill (bumped to v16) is the id table an agent consults before writing AgentHub code, so the feature would arrive incomplete with the Flash ids missing from it. Each family row followed its own existing convention rather than a single imposed one: GLM 5.3 is generation-named and already the kind of row that holds several ids, so `glm-5.3-flash` and `z-ai/glm-5.3-flash` joined it; the Qwen rows are variant-named (`Qwen 3.8 Max`, like `Kimi K2.7 Code`), so `qwen3.8-flash` got its own `Qwen 3.8 Flash` row. No column was added — the table is Family / Official IDs / Gateway variants and stays that way.

The vision caveat therefore lives in prose, next to the table's existing id-level gotcha: `glm-5.3-flash` is multimodal but the GLM client rejects image parts, so the direct id is text-only in practice and images need the `z-ai/glm-5.3-flash` variant on `openai-chat` — the same fact and the same reason as the catalog row's `vision = false`. The `openai-chat` bullet, which already named Qwen's gateways, gained DashScope's endpoint so the new gateway id is actually usable.

`packages/docs`'s `skills-sync.test.ts` only asserts that every library skill name appears in `skills.{en,zh}.md`; no skill was added, so it needed no docs-side change and stays green. The bilingual skill table rows for `agenthub-models` carry its description, which is unchanged.

## Compatibility

Nothing on disk changes. Existing Projects do not pick the presets up on their own — presets are copied into `.project_config.toml` at Project creation and nothing rewrites them afterwards; users bring the rows in through the models page's explicit **sync presets**, which appends missing catalog entries and updates catalog-owned fields but deletes nothing and touches neither the stored default model nor any credential. The changelog entry says exactly that. No compatibility shim was added, so this batch has no `backward-compatibility.md`.

## Verification

Node v24.18.0, run in the worktree after `pnpm install --frozen-lockfile`.

- `pnpm build` — pass (exit 0, all packages).
- `pnpm typecheck` — pass (exit 0).
- `pnpm --filter @prismshadow/penguin-core exec vitest run test/model-catalog.test.ts` — 24/24 pass.
- `pnpm --filter @prismshadow/penguin-web exec vitest run test/model-grouping.test.ts test/protocol-path.test.ts` — 31/31 pass.
- `pnpm --filter @prismshadow/penguin-server exec vitest run test/models.test.ts` — 22/22 pass.
- `pnpm --filter @prismshadow/penguin-docs test` — 53/53 pass (content and search-index suites, for the docs edit), including `skills-sync.test.ts` 3/3.
- `pnpm --filter @prismshadow/penguin-skills test` — 25/25 pass (skill frontmatter and library validation, for the skill edit).
- Note: `.prettierignore` excludes `*.md`, so the skill's Markdown table alignment is hand-maintained; the two touched rows were padded to the table's existing 18/71/145 cell widths.
- `git diff --check` — clean.

Vendor pages re-read for this PR: https://docs.z.ai/guides/vlm/glm-5.3-flash, https://docs.z.ai/guides/overview/pricing, https://openrouter.ai/z-ai/glm-5.3-flash, https://www.qianwenai.com/models/qwen3.8-flash. AgentHub's GLM client checked at `git show origin/main:src_ts/src/glm5_3/client.ts`.


hiyouga:
The direct `glm-5.3-flash` row now carries `vision = true`. The condition its comment was written to wait on has shipped: [agenthub v0.4.8](https://github.com/Prism-Shadow/agenthub/releases/tag/v0.4.8) teaches the GLM client to forward `image_url` parts for `glm-5.3-flash` — in a prompt and in a tool result — while every other GLM id still refuses them.

**The flag is true in fact, not only in the catalog.** `core` and `cli` raise their declared `@prismshadow/agenthub` range to `^0.4.8` (previously `^0.4.7` and `^0.4.6`) and the lockfile resolves both to 0.4.8, collapsing the two copies the workspace used to install into one. `pnpm-workspace.yaml`'s `minimumReleaseAgeExclude` moves to `0.4.8` — pnpm 11's supply-chain gate would otherwise withhold a version published today — following that entry's own instruction to roll it forward at the next agenthub bump. Nothing else in the lockfile moved.

Verified against the installed 0.4.8 through `AutoLLMClient`, the same entry point `packages/core` uses — request building only, no network:

```
glm-5.3-flash  PROMPT      -> [{"role":"user","content":[{"type":"text",...},{"type":"image_url","image_url":{"url":"https://example.com/penguin.png"}}]}]
glm-5.3-flash  TOOL RESULT -> [{"role":"tool","tool_call_id":"call_1","content":[{"type":"text",...},{"type":"image_url","image_url":{"url":"https://example.com/penguin.png"}}]}]
glm-5.3        PROMPT      -> THREW: GLM glm-5.3 does not support image inputs.
glm-5.3        TOOL RESULT -> THREW: GLM glm-5.3 does not support images in tool results.
```

Prose that asserted the split is corrected rather than annotated: the row comment and the file header state the current rule (the header's `glm-5v-turbo` exclusion said image input was unsupported by the GLM client at all — it is now gated per model, and `glm-5v-turbo` still refuses, verified); the `agenthub-models` skill goes to v17 with the text-only caveat replaced by what the client does, including both refusal messages and the case-insensitive match that covers `z-ai/glm-5.3-flash` and `zai-org/GLM-5.3-Flash`; the bilingual `models` docs and the changelog pair drop the vision split and keep the price split, which is real; and the OpenRouter row's own comment loses its now-false contrast ("Unlike the direct row, this one carries vision"). The catalog test now pins vision per row across the whole direct Z.AI group.

One correction worth flagging: the compatibility note does **not** say existing Projects keep a stale `vision = false`. All three rows are new in this PR and have never been in a release, so no Project on disk carries the row or its flag — the first `sync presets` appends it as it now stands, with `vision = true`.

`origin/main` was merged in (`ab79125d`) rather than rebased, to keep the pushed history intact and avoid a force-push; `#471`'s `packages/cli/package.json` script edit and the dependency bump here sit in different sections and merged without conflict.

**Verification** (Node 24): `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm typecheck`, core `model-catalog.test.ts` (24), web `model-grouping` + `protocol-path` + `catalog-sync` (38), server `models.test.ts` (22), `skills` (25), `docs` incl. `skills-sync` (53), `pnpm format` + `format:check`, `git diff --check` — all pass.

PR #470, opened 2026-08-26
Authorize a new TokenDance API key from the models page

hiyouga:
Stacks on [#465](https://github.com/Prism-Shadow/penguin-harness/pull/465) (the TokenDance provider group and `attributionHeaders`). GitHub will retarget this to `main` once #465 merges.

TokenDance publishes an OAuth-style flow that mints a **new** API key without the user visiting the console. This wires it end to end: an action on the TokenDance group header, authorization in the browser, and the key applied to every model in the group.

## Surface

**Catalog (`packages/core`).** `ModelProviderInfo` gained an optional `oauth` descriptor — `authorizeUrl`, `exchangeUrl`, `keyName` — set on the `tokendance` entry. The Web App renders the action for any group whose provider declares one, so no provider id is hardcoded on that side, and the server accepts a flow only for such a group. `APP_URL` became an export rather than a second copy: it is what the flow sends as `app_url`, so a key minted here is stamped with the same app URL `attributionHeaders` already carries.

**Routes (`packages/server`), all owner-only.**

| Method | Path | |
| --- | --- | --- |
| POST | `/api/projects/:p/model-oauth/start` | `{provider, mode?}` → `{flowId, authorizeUrl}` |
| GET | `/api/projects/:p/model-oauth/callback` | where TokenDance redirects; exchanges, applies, answers a small HTML page |
| GET | `/api/projects/:p/model-oauth/:flowId` | `{status, provider, error?}` for the dialog to poll |
| POST | `/api/projects/:p/model-oauth/:flowId/code` | `{code}` → redeems a pasted code |

`ProjectConfigService.setGroupApiKey` writes the key onto the group with the same `api_key` / `created_at` pair `updateModels` writes; the route then invalidates cached runtimes and publishes `credentials_updated` through the helper `models.ts` already owned.

**UI (`packages/web`).** An **Authorize new API key** button in the group header's action row, following the existing pattern exactly (`size="sm" variant="ghost" className="shrink-0"`, `GlyphIcon` at 13, `aria-label` + `title`, label hidden below `@3xl`) with a new sign-in glyph. The dialog states that a **new** key will be created and will replace the group's current one, opens the authorization page, polls, reports the outcome, and offers the manual paste path. Every string is in both dictionaries.

## Security decisions

- **The verifier and the key never enter the browser.** The verifier is generated server-side, lives only in this process's memory, and is dropped the moment it is spent. The minted key goes from the exchange response straight into the model table — it is in no response body, no log line, no URL, and no error message. The redirect page reports the outcome and nothing else.
- **S256 always**, in both modes, not only where the docs make it mandatory: a code observed in a redirect is useless without the verifier.
- **The catalog chooses the endpoints.** The client sends a provider id; the authorize and exchange URLs come from `providerInfo(...).oauth`. A group that declares none is a 400, so no request can point the challenge or the key request anywhere else.
- **A flow is a capability bound to one user in one Project**, single-use, 10-minute TTL (matching the upstream code's lifetime, which also bounds how long a verifier sits in memory). Another user, another Project, an unknown id and an expired id all 404 alike, so a flow id is never a probe. The store sweeps on access and evicts the oldest per owner rather than growing without bound.
- **Callback origin comes from the request's own URL**, so loopback, a LAN address and a custom port work unconfigured. `x-forwarded-proto` / `x-forwarded-host` are honoured only under `PENGUIN_TRUST_PROXY=1` — the same opt-in the hot-update network gate requires, and for the same reason: otherwise anyone who can reach the bind could choose where the authorization lands.
- **A lost key is reported honestly.** TokenDance reveals the full key exactly once, so a key that arrives but cannot be stored ends the flow as `apply_failed` and the copy tells the user to authorize again and delete the orphan. The alternative — echoing the key so it is not lost — is precisely what this design refuses.
- The exchange's failure reasons are reduced to five codes the frontend phrases; a transport failure carries no detail, since a fetch error message can quote the request body, and the request body holds the code and the verifier.

## Compatibility

Nothing already on disk changes shape. `oauth` is a code-side catalog field; the key is stored as the same `api_key` / `created_at` pair every other credential write uses. No config file, database or stored preference is migrated, and no compatibility shim was added.

Existing Projects still do not pick up catalog changes on their own — presets are copied into `.project_config.toml` at Project creation and only the models page's "sync presets" updates them. The action therefore shows wherever a Project's table already holds TokenDance entries; a Project older than that group syncs presets once first. The changelog entry says so.

## Left out deliberately

The `TokenDance-Recovery-Action` response header (`top_up_balance` / `reauthorize_api_key` / `api_key_quota`) is not handled. It is a natural follow-up: `reauthorize_api_key` in particular could drive this same flow from a failed request instead of from the models page.

## Verification

Run in the worktree after rebasing onto the current `feat/model-attribution-tokendance` (`fff481e9`), Node 24:

- `pnpm build` — pass
- `pnpm typecheck` — pass, all 8 packages
- `pnpm --filter @prismshadow/penguin-core test` — 995 passed, 5 skipped
- `pnpm --filter @prismshadow/penguin-server test` — 1002 passed
- `pnpm --filter @prismshadow/penguin-web test` — 1296 passed
- `pnpm --filter @prismshadow/penguin-docs test` — 53 passed
- `git diff --check` — clean

New tests (`packages/server/test/model-oauth.test.ts`, 25 cases) cover the PKCE challenge against the RFC 7636 vector and the verifier's alphabet/length, authorize-URL construction (`callback_url` and `app_url` encoding, `key_name`, S256, and manual mode omitting the callback), flow ownership / expiry / single-use, the non-oauth-provider rejection, the exchange's 400 / 403 / 5xx / malformed-body / transport mapping, forwarded-header origin derivation, and the routes end to end — owner vs member, the key reaching every model of the group and the config file, `credentials_updated` published, a replayed callback refused, and the manual paste path. Two cases assert the redirect page and the responses contain neither the code nor the key. `packages/core/test/model-catalog.test.ts` pins that TokenDance is the only group declaring `oauth` and that its `app_url` is the attribution `APP_URL`. No test touches the network: `exchangeCode` takes its fetch as an argument and the route cases stub the global one.

Not verified by hand against the live tokendance.space endpoint.


hiyouga:
Relabelled and reordered the provider-group header actions (e7d690f3).

The five labels were sentence-length; they are now one or two words in both dictionaries, and the automatic key action comes before the manual one:

| # | zh | en |
| --- | --- | --- |
| 1 | 添加模型 | Add model |
| 2 | 自动获取密钥 | Authorize key |
| 3 | 手动设置密钥 | Set key |
| 4 | 测速 | Speed test |
| 5 | 前往密钥管理 ↗ | Manage keys ↗ |

Only the string values moved — no key was renamed, so every call site follows automatically. Two of the retexted keys are shared beyond the header: `models.getApiKey` also labels the provider link in the model settings dialog (right of the **API key** field, next to `Get model IDs ↗`) and at the foot of the group key dialog; "Manage keys" is the same trip to the provider's console in all three, so they stay on one string. `models.addToGroup` is header-only, but the empty-group hint quoted it and was updated with it. The dialog titles the buttons open (`groupApiKeyTitle`, `oauthTitle`, `addTitleVendor`) are unchanged — they are the explicit long form the short button no longer carries.

A user-defined group still shows **Add model · Set key · Speed test · Delete group**: it has neither an authorization flow nor a key console, so the delete button never competes with entries 2 or 5.

The Models doc now names the action the way the button spells it, and the changelog pair gained one clause on the relabel.

**Verification** — `pnpm typecheck`, `pnpm --filter @prismshadow/penguin-web test` (1296 pass), `pnpm --filter @prismshadow/penguin-docs test` (53 pass), `pnpm format` + `pnpm format:check`, `git diff --check`. No test or e2e spec asserts on the old label text.

Rendered through the Playwright harness (mock LLM, scratch data root) at row widths from 30rem to 64rem in both languages: every label lays out on a single line box, the row never overflows (`scrollWidth == clientWidth` throughout), its height stays 45px at every width, and all five actions plus the collapse chevron remain present in the icon-only state below `@3xl`. Button labels return at 48rem of row width and the link's at 56rem, as before.

PR #471, opened 2026-08-26
Dev CLI gets its own data root; agent-env and MCP hygiene guarantees pinned

hiyouga:
## What

Environment-hygiene and port/data-root isolation pass over every entry point, in two parts.

**Part A — audit of the port/data-root landscape, with one fix.** Every place a port or a data root is chosen was enumerated (`ports.ts`, package.json script lines, vite configs, the desktop shell, the e2e harness, `penguin-hmr`, `run-with-env.mjs`). The matrix of server-holding entry points:

| Entry point | Port | Data root | When a second server hits the root |
| --- | --- | --- | --- |
| installed `penguin server` / `penguin web` | 7364 | `~/.penguin/data` | `server`: refuses with the URL; `web`: opens the existing instance |
| installed desktop | none fixed (sticky `PORT=0`) | `~/.penguin/data` (shares the CLI root by design) | attach mode |
| `pnpm dev:server` (also via `pnpm dev`) | 7368 | `~/.penguin/dev-data` | refuses, exit 3 |
| `pnpm penguin web` (dev CLI) | 7369 | **was `~/.penguin/dev-data` → now `~/.penguin/dev-data-cli`** | opens the existing instance |
| `pnpm desktop` (dev shell) | none fixed (sticky, own userData per #292 fix) | `~/.penguin/dev-data` | attach mode |
| web e2e (`packages/web/e2e/run.sh`) | 8930 (+8931 mock) | `mktemp -d`, deleted after | n/a |
| `penguin-hmr` | whatever the pushed CLI binds (7364 default) | `~/.penguin/data` | deliberate: a pushed CLI acts as the installed one |

Non-serving dev ports: vite web 7365 (proxy → 7368, or `PORT`/`PENGUIN_API_PROXY`), landing 7366, docs 7367 — no data roots.

Findings:

- **Installed × dev holds everywhere**: ports 7364 vs 7365–7369/ephemeral, roots `data` vs `dev-data*`, and `run-with-env.mjs` already names any default an exported variable displaces.
- **One genuine conflict, fixed**: the dev CLI shared `~/.penguin/dev-data` with `pnpm dev:server`. The lock is per root, so the pair that port 7369 exists to let run simultaneously (`ports.ts`: a harness started as `pnpm penguin web` is exactly what asks an Agent to run `pnpm dev` in this repo) could not actually coexist — the Agent's `dev:server` exited 3 against the harness's own lock. `pnpm penguin` now defaults to `~/.penguin/dev-data-cli` (root + cli scripts). Trade-off stated in CONTRIBUTING and the changelog: aiming a dev CLI command at the `pnpm dev` dataset is now said per command (inline `PENGUIN_HOME` or `--root`).
- **Deliberate sharing, now pinned instead of implicit**: the dev desktop stays on `~/.penguin/dev-data` — a second server on a locked root is its attach-mode case, not a failure, and the two surfaces used alternately sharing one dataset is the point. A new `packages/cli/test/dev-entry-isolation.test.ts` pins the (port, root) disjointness of the simultaneous pairs, the root/cli script agreement, and the desktop's root-sharing choice; the `ports.ts` table now carries the data-root column plus the desktop/e2e rows.

**Part B — exec_command / MCP environment hygiene.** The strip itself landed in #434 (every `PENGUIN_*` by prefix, plus `PORT`/`HOST`/`FORCE_COLOR`/`CLICOLOR_FORCE`; removal not blanking; case-insensitive for Windows; vault re-adds), so this PR verifies and pins rather than reimplements:

- **Layering seam made explicit** for injection layers added to the spawn spread later (e.g. control-var injection): the strip runs inside `hostEnvForChild` and never re-applies to later spread entries, so `{ ...hostEnvForChild(...), ...vault, ...HARDENED_ENV }` composes with any future `...injected` for free — inherited copy gone, injected value stands. Asserted in a comment at the spread and a new test ("an explicit injection layered after the strip wins, for any `PENGUIN_*` name") that pollutes the parent with one value and injects another.
- **MCP stdio checked for the same leak: there is none.** Its child env is the MCP SDK's safe-inherited allowlist (`HOME`/`PATH`/`SHELL`-class names only) plus the entry's own `env` — the spec'd shape — so no `PENGUIN_*`/`PORT` can arrive by construction. That guarantee had no local coverage (only the SDK's implementation provided it); a new test pins it (fixture probe now reports `PENGUIN_HOME`/`PORT`), and the provider comment now names the invariant. Policy kept as-is: commands strip a blocklist from a full inherit, MCP inherits an allowlist — different mechanisms, same outcome (harness config never reaches a child), each fitting its spec.
- The web terminal was considered and left alone: it is the signed-in user's own shell on the host, full-env inheritance is its documented design, and a shell running as the server's OS user could read the server's environment regardless — stripping there would be cosmetic.

## Compatibility

- The #434 break (agent shells no longer inherit `PENGUIN_HOME` etc.) is unchanged here; its entry documents the vault migration. Control-var injection (PR #466) will cover the "call the same harness's API" need without any inheritance; sharing a data root with the running harness remains a vault decision.
- The dev CLI root move is contributor-facing only; the changelog entry's Compatibility section states the one-time move and the explicit-root escape hatch. No compat code was added, so no dedicated backward-compatibility entry.
- Drive-by: the #434 entry's English file used the Chinese compat heading; normalized to `## Compatibility` per changelog/README.

## Verification

- `pnpm -r build`, `pnpm -r typecheck` — green (Node 24).
- `pnpm --filter @prismshadow/penguin-core test` — 991 passed, 5 skipped; `--filter @prismshadow/penguin-server test` — 977 passed; `--filter @prismshadow/penguin-cli test` — 306 passed.
- `pnpm format` + `pnpm format:check`, `git diff --check` — clean.

### DIFF
--- a/.agents/skills/penguin-harness-manual-test/SKILL.md
+++ b/.agents/skills/penguin-harness-manual-test/SKILL.md
@@ -18,8 +18,9 @@ workspace deps); `pnpm desktop` runs a full `pnpm -r build`, so it is slow to st
 | `pnpm dev:docs` | http://localhost:7367 | none (static) |
 
 Other fixed ports (`packages/core/src/internal/ports.ts`): 7364 installed server, 7368 dev backend,
-7369 `pnpm penguin web`. On a shared box, `ss -tln` before assuming one is free; `PORT=` inline
-moves it.
+7369 `pnpm penguin web` (data root `~/.penguin/dev-data-cli` — its own, so it can serve while an
+Agent it hosts runs `pnpm dev`; the lock is per root). On a shared box, `ss -tln` before assuming
+one is free; `PORT=` inline moves it.
 
 The user's installed app, server and CLI all use `~/.penguin/data` — their real Agents, Sessions
 and keys. Never point a dev run there. Both surfaces print the root they took (`Data root: …`,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,10 +45,18 @@ that only re-sync when the package's `build` script runs via pnpm
 cache — on the old build; if a running dev web app still serves stale core after a manual
 rebuild, delete `packages/web/node_modules/.vite` and restart.
 
-Dev entry points that touch data (`pnpm dev`, `pnpm dev:server`, `pnpm penguin`,
-`pnpm desktop`) default to a separate data root, `~/.penguin/dev-data`, kept apart from
-the installed CLI/server's `~/.penguin/data` — hacking on the repo never mixes state with
-your real agents. Need a different root? Pass `PENGUIN_HOME` inline, for the single
+Dev entry points that touch data default to separate data roots, kept apart from the
+installed CLI/server's `~/.penguin/data` — hacking on the repo never mixes state with
+your real agents. `pnpm dev`, `pnpm dev:server` and `pnpm desktop` share
+`~/.penguin/dev-data`; `pnpm penguin` takes its own `~/.penguin/dev-data-cli`, because a
+data root admits one server at a time (`<root>/server.lock`) and the dev CLI's
+`penguin web` is exactly the harness that then asks an Agent to run `pnpm dev` — on a
+shared root that `dev:server` would refuse to start against the harness's own lock (the
+same coexistence that already gives it port 7369, see
+`packages/core/src/internal/ports.ts`). To aim a dev CLI command at the `pnpm dev`
+dataset anyway, say so per command: `PENGUIN_HOME=~/.penguin/dev-data pnpm penguin
+config model list`, or `--root` where the subcommand takes it. Need a different root?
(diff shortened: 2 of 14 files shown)


hiyouga:
CI note: GitHub Actions is currently in a major outage (githubstatus.com shows Actions: major_outage), so the `pull_request`-event run for this PR has not been created — a close/reopen did not produce one either. The full workflow was run on this exact head SHA (`d61fb951`) via `workflow_dispatch` instead and is green across all seven jobs (ci incl. live e2e, ci-macos, ci-windows packages+server, three runtime matrices): https://github.com/Prism-Shadow/penguin-harness/actions/runs/32987166291. The PR checks should attach once Actions recovers; if not, any push or re-run then will.

PR #472, opened 2026-08-26
Update badges lead from the outermost menu to the update control

hiyouga:
A phone-style red dot now marks the chrome whenever something is updatable, and every dot leads down an unbroken path to the control that performs the update. A dot the user cannot follow to an action is a bug, so every gate in this change is written as "is there a control the user can reach".

Two trails carry a badge:

- **Software** — a newer release for a browser install, or a downloaded client build for the desktop shell's own window. Anchors: the `<md` top bar's menu button, the collapsed rail's avatar, the pinned sidebar's avatar → the user menu's update row → the update dialog.
- **Agent kernel** — any Agent in the current Project whose kernel stamp is behind. Anchors: the same menu button, the pinned sidebar's Agents entry, the collapsed rail's Agents icon, the Agents list card of each outdated Agent, then that Agent's settings page — the Overview tab, the Kernel section, and the enabled "update kernel" button beside it.

**Deliberately out of scope:** the sidebar's conversation/Agent *groups* in the session list are not badged. That list is about conversations, not Agent configuration; a dot there would sit on a path that leads to no action.

## One badge component, one place the colour lives

`packages/web/src/components/ui/update-dot.tsx` is the only place the badge colour is written, and every dot in the feature renders through it — including the sidebar avatar's, which previously drew its own `bg-[var(--accent-bg)]` mark inline.

The user asked for red, phone-style, and `lib/tone.ts` is documented as "the one place a status colour is spelled out". That constraint is handled deliberately rather than worked around: tone.ts's five tones are each a *judgement about a thing's state* — `danger` is failed, destructive or over a limit; `attention` is unfinished work waiting on someone. An available update is neither. Nothing has gone wrong, nothing is stuck. What this mark says is "there is something new further down this path", which is a category tone.ts explicitly keeps out of scope, so putting it there would have to either re-mean `danger` or add a sixth tone that is not a status. The component's header states that, cross-references tone.ts, and records the measured contrast: `red-500` is 3.76 : 1 on white, 3.66 : 1 on gray-50, 5.58 : 1 on this app's gray-950 (`#000000`) and 5.16 : 1 on its gray-900 (`#0d0d0d`) — clearing the 3 : 1 WCAG 1.4.11 asks of a graphical object on all four surfaces a dot lands on.

Two sizes, named by anchor: `chrome` (11.25px with a 2px ring in the sidebar's own background, for an avatar or a rail icon button) and `inline` (6.75px bare, for a text row or a lone glyph). Both are absolutely positioned and `pointer-events-none`, so no row changes height and no hit area moves.

## Accessibility

Every dot is `aria-hidden` and carries no meaning of its own. The anchor states what is updatable, in its `title` and in its accessible name, following the pattern the sidebar avatar already used (`title: <what is waiting>`, `aria-label: "<anchor> · <that same wording>"`). A row with a visible label gets the note alone in the tooltip and the label as the accessible name's prefix; an icon-only anchor (the rail, the menu button) gets both in both. The Kernel section title is not interactive, so it carries an `sr-only` sentence instead of relying on a tooltip.

## The lazy check

The release check used to be deliberately lazy — nothing fired until the user opened the sidebar's bottom dropdown — so on a fresh load the outermost dot could not appear at all. The app layout is now the single owner and activates it once on mount.

The trade-off was checked rather than assumed:

- It is still **one request per browser session**: the module-level cache and shared promise in `use-version-info.ts` are untouched.
- It costs **no extra outbound request**. The server caches a successful lookup for an hour, caches failures for ten minutes, and shares one in-flight call — its own header already reasons about "a sidebar that opens on every page load".
- The **opt-out stays authoritative**: `PENGUIN_UPDATE_CHECK=off` answers without dialing out, arrives as `updateAvailable: false`, and produces no dot and no notice. A failed lookup is fail-soft the same way.
- It is not a new behaviour in spirit: the draft page's version line already called `useVersionInfo(true)` on mount, and `/` redirects to `/chat`, so the default landing route already fired the check on load.

`use-version-info.ts`'s header was rewritten to describe the behaviour as it now is, with no narration of the change.

## A broken trail this fixes

In desktop mode the sidebar hides the server-release row entirely (updating is the shell's job) but the avatar dot was gated on `update.updateAvailable` regardless — so a newer server release put a dot on the avatar whose menu had no row to act on it. The software gate now takes which row the running mode actually offers: in the shell's own window the avatar marks a client build the shell has *downloaded and is waiting to install*, and a browser signed into a desktop-mode server gets neither row and no badge. A client download still in flight is deliberately not badged — there is nothing to act on until it lands, and a badge the user cannot clear by acting on it is noise.

## Compatibility

Nothing here reads or writes state on disk. No new `ui_prefs` key, no new localStorage key, no `web.db` or `system_config.yaml` involvement, no API shape change — the kernel flag rides along on the Project's existing Agent list and the release check already existed. `Tabs` gained an optional `badge` prop, which is additive.

## Verification

Node 24, in a worktree off `origin/main`.

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (all 8 packages)
- `pnpm --filter @prismshadow/penguin-web test` — 99 files, 1317 tests, pass (21 of them new)
- `git diff --check` — clean

New unit tests (`packages/web/test/update-badges.test.ts`) cover the logic, not the pixels: the release gate including the `disabled`, failed-lookup and available-but-unnamed paths; the client-build gate across every shell state; the "release row not offered" case; "any Agent outdated"; and the combined note that must name neither trail.

Driven by hand through the Playwright harness against a scratch `PENGUIN_HOME` with the mock LLM, screenshotting at `deviceScaleFactor: 2`, with a fabricated release response and a real Agent whose `kernel_version` stamp was rolled back on disk (so the settings page, the confirm dialog and the update itself were exercised for real, not mocked):

- pinned sidebar, collapsed rail, mobile top bar and mobile drawer all show the dots, in both light and dark
- the software trail reaches the user menu's "新版本 v0.9.9 可用" row
- the kernel trail reaches the Overview tab dot, the Kernel section dot, and an enabled "update kernel" button
- running the update clears the dot on the section, the tab, the card and the sidebar entry — the page's existing config refresh is enough
- with `disabled: true` the software dot is absent
- English copy on the combined anchor reads `Sessions · Updates available`; the Agents entry reads `Agents · Kernel update available`

The throwaway spec and its screenshots were not committed. The full e2e suite was not run — about 8 specs fail on `main` on this box already, and no committed selector or flow moved.


hiyouga:
## Update-badge restyle — pale red, and two anchors moved

Pushed as `750b8c4a`.

### 1. Nav row dot: right edge, vertically centred

`sidebar.tsx` — `position="right-2.5 top-1/2 -translate-y-1/2"`. The `NavLink` row was
already the `relative` anchor; `right-2.5` matches the row's own `px-2.5`, so the dot lines
up with the sidebar's content column rather than floating in the gutter. Row height is
unchanged and the dot clears the label at both sidebar widths (`w-64` at `md`, `lg:w-72`).

The boxed chrome anchors are untouched — the collapsed rail's icon buttons, both avatars and
the mobile hamburger keep their corner overhang. A square icon button has a corner and no row
to centre in, and side by side the two treatments read as one rule rather than two.

### 2. `red-500` → `red-400` everywhere, with the numbers redone

One value, in `UpdateDot`. Measured WCAG 2.x contrast against every surface a dot lands on:

| | white | gray-50 | nav row active fill |
|---|---|---|---|
| light | 2.89 : 1 | 2.76 : 1 | 2.46 : 1 |

| | gray-950 `#000000` | gray-900 `#0d0d0d` | gray-800 `#1f1f1f` |
|---|---|---|---|
| dark | 7.27 : 1 | 6.73 : 1 | 5.71 : 1 |

Dark clears WCAG 1.4.11's 3 : 1 on all three. **Light misses on all three, by 4% to 18%**, and
the header now says so rather than claiming otherwise. Accepted because the dot is
`aria-hidden` and decorative — every anchor names what is updatable in its own `title` and
accessible name, so 1.4.11's "required to understand the content" does not attach — and
because `red-500` is the palest rung on Tailwind's ramp that clears 3 : 1 in light
(3.81 / 3.64 / 3.25 : 1). There is no rung that is both paler than that and compliant, so
"lighter" and "compliant" could not both be had here.

The old header quoted 3.76 / 3.66 / 5.58 / 5.16 for `red-500`; recomputed from Tailwind
4.3's oklch values those are 3.81 / 3.64 / 5.52 / 5.10, and the nav row's active fill was
never among them. Both are fixed.

### 3. Kernel dot straddles the update button's corner

`agent-settings-page.tsx` — `position="right-0.5 top-0.5 -translate-y-1/2 translate-x-1/2"`,
putting the dot's centre on the rounded corner's arc so it sits half in, half out of the
border rather than inside the box. `pointer-events-none` keeps it clear of the hit area, and
the `sr-only` sentence stays inside the `Button`, so the enabled control's accessible name is
still "Update kernel · Kernel update available" — the same wording the trail carries from the
sidebar down.

### 4. The Agents-card capsule went pale too

The `UpdatePill` was `bg-red-700` / `dark:bg-red-900` with white text. Beside the new pale
dots it was by far the loudest thing on the Agents page — the trail read whisper → shout →
whisper. It is now a tinted capsule in the same badge family as the `v1` pill next to it:
`bg-red-100 text-red-800` (6.85 : 1) light, `bg-red-950 text-red-300` (8.42 : 1) dark, hover
fills `red-200` / `red-900` at 5.75 : 1 and 5.22 : 1. Same button, same geometry, same label.

This revises a choice made one commit earlier (`b66f602b` picked dark red *because* the pill
has an interior to read). The contrast requirement that motivated it is still met — a pale
surface just pays it with a dark ink instead of a white one. **Say the word and I will put the
solid dark red back.**

### Badge red vs destructive red in the Kernel row — honest read

The dot now shares a row with the `danger`-tone "Restore default configuration" button.

- **Light:** fine. The button is red-600 *text* (`#e7000b`), the dot is red-400 (`#ff6467`) —
  visibly lighter and pinker, and a 6.75px filled circle next to a word is not confusable.
  Lightening the dot *helped* here; at red-500 it sat much closer to the destructive ink.
- **Dark: no hue separation at all.** `tone.ts` gives `danger` ink `dark:text-red-400`, so the
  notification dot and the destructive label are now *the same hex*. It still reads correctly
  — different shapes, and the dot straddles the button that means the opposite thing — but
  colour is doing none of the separating.

If you want that separated, the fix I would suggest is a dark-only paler rung for the dot,
`dark:bg-red-300` (10.94 / 10.13 / 8.59 : 1 on gray-950 / gray-900 / gray-800), at the cost of
the "one value across both themes" rule the module currently keeps. I did not do it unasked.
Not restyling the destructive button either way.

### Verification

`pnpm typecheck`; `pnpm --filter @prismshadow/penguin-web test` (99 files, 1317 tests, all
green — `test/update-badges.test.ts` is gate logic and needed no change); `pnpm format` +
`pnpm format:check`; `git diff --check`. Driven through the Playwright harness with the mock
LLM against a scratch `PENGUIN_HOME`, screenshotted at `deviceScaleFactor: 2` in both themes:
sidebar nav row at both widths, collapsed rail, mobile hamburger, Agents list card, Overview
tab strip (still not clipped by the strip's `overflow-y-hidden`) and the Kernel row. Running
the update clears every dot on the trail — asserted, not just looked at.

The bilingual changelog pair is updated where it described placement or colour.

PR #480, opened 2026-08-26
feat(desktop,docs): install the bundled penguin command automatically

hiyouga:
## What

The desktop app carries the CLI inside itself, but only the **deb** package put `penguin` on a user's PATH by itself. macOS, Windows and the AppImage showed a one-time **Install / Not Now** dialog on first launch and never asked again — and the flag that gated it was written *before* the dialog was answered, so it recorded that the question was asked, not what was answered. An app that bundles its own CLI and updates itself has to keep that command current; a stale `penguin` against a newer server is a live compatibility surface now that the CLI is an HTTP/SSE client.

The shell now installs and repairs the command at **every launch**, without asking.

## How

New `packages/desktop/src/cli-link.ts` — pure `node:fs`, no Electron import — owns the whole disk-side decision, so the automatic path and the menu item cannot drift apart. `cli-install.ts` is the Electron adapter around it.

| Form | What happens |
| --- | --- |
| macOS | `/usr/local/bin/penguin` → `<app>/bin/penguin`. The unprivileged write is tried first; `osascript` escalation happens **only** on a real `EACCES`/`EPERM`, never as an opening move. A dismissed prompt is recorded and the automatic path stops asking; the menu item clears it. |
| Windows | `<app>\bin` appended to `HKCU\Environment` `Path`. Unchanged mechanism, no elevation, append-only. |
| Linux AppImage | `~/.local/bin/penguin` wrapper, rewritten when the AppImage moves. |
| Linux deb | unchanged — its postinst already does it. |

Two refusals, both enforced in `cli-link.ts` rather than left to a caller remembering to check:

- **Never replace a `penguin` this app did not write.** Ownership is a symlink into a `resources/app/bin/penguin` layout, or the `LAUNCHER_MARKER` text every generated launcher has carried since 0.2.2 (so an install made by an older version is repaired, not treated as foreign). Anything else — a regular file where a link belongs, a symlink elsewhere, an unreadable or oversized target — is `skipped` and the reason recorded in `userData/cli-command.json`. This is the `~/.local/bin/penguin` collision with `install.sh` that I reproduced in the audit: **ours is the side that gives way.** `force` exists only behind the menu item's Replace/Cancel dialog.
- **Never install from a location that will not persist.** `/Volumes/…` (running from the mounted dmg) and `/AppTranslocation/…` (Gatekeeper running a quarantined bundle) both produce a link that dies with the mount, and the old code would then never re-offer. The install defers; the next launch from `/Applications` does it.

Running every launch is what makes this *repair* and not just *create*: a dangling link from a moved or updated app, or a wrapper naming a moved AppImage, is rewritten. A correct `penguin` costs one `lstat`.

State moves from the `cli-install-offered` marker to `userData/cli-command.json`, which separates the *outcome* of the last attempt from a *decision* — and only one thing sets a decision: dismissing the macOS administrator prompt.

### Tests and CI (the first commit, previously this PR's whole content)

- `packages/desktop/scripts/verify-packed-cli.mjs` checks the `electron-builder --dir` tree and runs the bundled launcher; ci.yml's existing `runtime` job invokes it on Linux, macOS and Windows.
- `launcher.test.ts` gained a `packaged CLI` block pinning `bin/**/*` in `files`, `asar: false`, `productName`/`executableName` vs the launcher constants, the tsup entry vs `CLI_ENTRY_RELPATH`, and the deb postinst/postrm link, plus a `LAUNCHER_MARKER` test — the marker is now load-bearing for ownership.
- `cli-link.test.ts` (24 tests) runs against **real temporary directories**, not mocks: symlinks, modes and dangling targets are the whole subject.

### Docs

`quickstart-desktop.{en,zh}.md` gained a `penguin` command section covering all four forms and both refusals, and stopped sending desktop users to `install.sh` for a command they already have.

**Overlaps [#481](https://github.com/Prism-Shadow/penguin-harness/pull/481):** at that PR's request I also removed the stale macOS quarantine / Windows SmartScreen first-launch instructions from `quickstart-desktop.{en,zh}.md` and `quickstart.{en,zh}.md` — #481 removed the same copy from the landing page and both READMEs and left the docs site out of scope. Wording matches its phrasing so the two sites do not disagree. No file is touched by both PRs.

## Compatibility

`Breaking` is set on `changelog/unreleased/2026-08-27-backward-compatibility.md`: the pre-0.2.7 `cli-install-offered` marker is **not** migrated. It was written before its dialog was answered, so an install, a "Not Now" and an unanswered dialog all left the same file — it cannot be read as a decline, and reading it as one would deny the command to the entire existing install base. It is ignored and removed on the first new state write (one `rmSync`; `TODO(compat)` at the code site names 0.3.0 as the removal point). The consequence, stated in the entry: a user who answered "Not Now" to the old offer gets the command installed on their next launch.

Nothing a user installed by another route is moved or replaced — that is the hard rule above, and it is what the tests pin.

**`install.sh` is untouched.** Its `ln -sf ~/.penguin/bin/penguin ~/.local/bin/penguin` is still unconditional and will still overwrite this app's wrapper. That is a separate decision and is left open.

## Verification

Node v24.18.0, Linux x64. Full chain: `pnpm -r build`, `pnpm typecheck` (9 packages), `pnpm format` + `format:check`, `git diff --check`, `sh scripts/test-installer.sh` — all pass. `desktop` 137 tests, `cli` 374, `docs` 53.

Exercised for real on this box, against a genuine `electron-builder --linux AppImage` build, driving the production `syncWrapper` in a throwaway `HOME`:

| Scenario | Result |
| --- | --- |
| fresh HOME, silent install | `installed`, mode 755, `penguin --version` → `v0.2.6` |
| second launch | `current`, file not rewritten |
| our wrapper naming a moved AppImage | `installed` (repaired), → `v0.2.6` |
| `install.sh`'s symlink already at the path | `skipped`, symlink intact, → `tarball-cli-0.0.1` |
| a foreign regular file at the path | `skipped`, → `somebody-elses-penguin` |
| menu item's forced replace | `installed`, → `v0.2.6` |
| decline recorded / legacy marker | `decision=declined`; `cli-install-offered` removed |

Also: `electron-builder --linux dir`, `resources/app/bin/penguin --version` → `v0.2.6` directly and through a deb-style symlink; `verify-packed-cli.mjs` passes and was negative-tested (missing `bin/`, cleared exec bit, renamed entry each exit 1); each of the nine config mutations the new unit tests cover fails exactly one assertion.

**Not exercisable on this Linux box:** macOS and Windows packaging, the `/usr/local/bin` symlink and its `osascript` escalation, the `reg.exe` PATH write, App Translocation, and any live auto-update. Those paths are covered by unit tests over the pure logic and by the CI step on all three runners — not by a real run here.


hiyouga:
## Review

Reviewed the full diff in a dedicated worktree, exercised the packaged tree, and mutation-tested the new units. **The shipped logic is sound** — the four red CI jobs were assumption errors in the *check* and in *two tests*, not in `cli-link.ts`. Fixes pushed as 10ae6611.

Highest-severity thing I looked for and did **not** find: the `osascript` escalation is injection-safe. I reversed the AppleScript escaping out of `adminSymlinkAppleScript`, ran the resulting shell command through a real `/bin/sh -c` for eight hostile bundle paths (`'; touch …`, `$( )`, backticks, backslash, `"`, newline, `$HOME;rm -rf`, apostrophe) and each produced exactly one correct symlink and no side effect. The `shellQuote` → `appleScriptString` ordering is right; a closing `"` cannot be produced unescaped.

---

### Blockers, fixed

**1. `verify-packed-cli.mjs` demanded a version the build never prints — `runtime` red on all three runners.**

```
resources/app: penguin --version printed "v0.2.6-g6a92db7", want v0.2.6.
```

The CLI prints the *build identity* from core's `composeDescribe`, not a bare version. `actions/checkout@v5` fetches depth 1 with no tags, so `git describe --tags --always` returns a bare sha and the identity renders `v<version>-g<sha>`. Only a release build (`BUILD_DATE` stamped) prints the bare `v0.2.6` the check demanded — i.e. the check could only ever pass on the one build shape CI does not produce.

`namesThisBuild()` now accepts `v<version>` and `v<version>-<suffix>`. Still a real assertion — the predicate matrix I ran accepts `v0.2.6`, `v0.2.6-g6a92db7`, `v0.2.6-3-g074f4b75`, `…-dirty` and rejects `v0.2.5-3-gabc1234`, `v0.2.60`, `0.2.6`, empty, and `penguin: unknown option '--version'`.

**2. Two POSIX-only assertions ran on Windows — `ci-windows (packages)` red.**

- `cli-link.test.ts:75` — `fs.symlinkSync` on Windows stores the target as a backslash path, so `BUNDLED_LAUNCHER_PATH` no longer matches; the link dangles so the marker cannot be read either, and the app's own stale link classifies `foreign` → `"skipped"`.
- `cli-link.test.ts:136` — `statSync().mode & 0o111` is 0; Windows carries no exec bit.

Neither code path ever runs on Windows (`cliInstallKind` returns `"windows"`, which is the registry form). Gated with `it.skipIf(process.platform === "win32")`, the idiom `test/pty-payload.test.ts:17` already uses. 23 of 25 still run there and no assertion is weakened where it means something. The exec-bit check moved into its own POSIX-only test rather than being deleted from the shared one.

**3. `cli-install.ts:132` — a successful-but-unparseable `reg query` replaced the user's entire PATH.**

Pre-existing code, but this PR moves it from *once, after the user clicked Install* to *silently, every launch*, which is what makes it worth fixing now. If `reg query HKCU\Environment /v Path` **succeeds** and the output does not match the regex, `current` stays `null`, `mergeWindowsUserPath(null, dir)` returns just `dir`, and `reg add … /f` writes the app's `bin` directory **in place of the whole user PATH** — with no dialog, because the automatic path shows none. Any `Path` value whose type or line shape the regex does not cover reaches this. `attemptWindows` now returns `failed` and writes nothing. Code-reading finding; I have no Windows box, but the fix is a pure refusal-to-write and cannot regress the fresh-profile case, which reaches the `catch`.

**4. `cli-link.test.ts:181` passed vacuously.** `does not read a large binary sitting at the target` used `Buffer.alloc(128 * 1024, 7)`, which contains no marker — so deleting `stat.size > MAX_MARKER_READ_BYTES` from `hasLauncherMarker` left it green (verified by mutation). The fixture now carries the launcher marker *past* the cap, so what makes the file foreign has to be the refusal to read it. Removing the cap now flips the classification to `ours` and fails the test named for it.

---

### Design docs are out of sync — needs its own design-repo PR

`specs/06-PROTOTYPE.md` § CLI 可用性 still reads:

> macOS / Windows / AppImage：原生菜单「Install 'penguin' Command…」加首启一次性询问（标记存 userData）

This PR deletes that behaviour outright: no first-launch question, automatic install/repair every launch, `cli-install-offered` replaced by `userData/cli-command.json`, plus two refusals (foreign target, volatile app location) the spec does not describe at all. Per `AGENTS.md` design is ground truth and must move with the code, and `design/` is a symlink so it cannot ride in this PR. Design PR: **[penguin-harness-design#65](https://github.com/Prism-Shadow/penguin-harness-design/pull/65)** — it also records the two refusals, the escalation rule and the state file, none of which the spec described.

---

### Not fixed — for you to decide

**5. First launch on a stock Mac now pops an admin password prompt with no preceding explanation.** The "unprivileged first, escalate only on real `EACCES`" rule is honoured, but on a Mac with no Homebrew and no `/usr/local/bin` that failure *is* the ordinary case, not the exception — and the dialog this PR removes (`Install the 'penguin' command line tool? [Install] [Not Now]`) was the consent step. That is a product call, not a code defect, so I left it to you.

**6. `cli-link.ts:193` / `cli-install.ts:78` overstate the `declined` decision.** Both JSDoc say it is "set only by an explicit act of the user", but `cli-install.ts:112` sets `declined: true` on *any* `osascript` rejection — a missing binary, a failing `mkdir`/`ln` under root, a suppressed dialog — permanently stopping the automatic install. The conservative behaviour is right and the user-facing `detail` already says "cancelled or failed"; only the comments (and two changelog sentences saying "only") overstate. Left alone rather than churn four prose files.

**7. Nits.** `cli-link.ts:132` — `chmodSync(tmp, 0o755)` is unpinned (`writeFileSync(…, {mode})` already leaves the owner exec bit under any umask, so the mutation survives; pinning needs process-global `umask()`, which races vitest). `cli-link.ts:237` — the state file is written non-atomically while core has `atomicWriteFile`, though it is in `internal/` and not on the barrel; worst case reads back as "no decision" and self-heals. `verify-packed-cli.mjs:97` — `spawnSync` has no `timeout`, so a hung launcher burns the whole job budget.

---

### Verified sound

- `force` is unreachable from the automatic path (`ensureCliCommand` → `attempt(kind, false)`; only `installCliCommand` passes `true`, behind Replace/Cancel).
- `writeWrapper`'s `renameSync` replaces the symlink itself, never its target — `install.sh`'s `~/.penguin/bin/penguin` is never written through.
- `LAUNCHER_MARKER` really is unchanged across v0.2.2–v0.2.6 (checked at each tag), and no other generator in the repo emits that text.
- A declined decision is honoured; corrupt/partial/unknown-value state JSON reads as *no* decision rather than a decline; the legacy marker removal is a single `rmSync`.
- Windows PATH append cannot grow across launches (`mergeWindowsUserPath` normalizes case, quotes, trailing slashes).
- All three changelog pairs conform to `changelog/README.md`; en/zh mirror section for section; docs' signing claim matches `electron-builder.yml` + `release.yml`'s `require_*_signing: true`.
- `launcher.test.ts`'s new `packaged CLI` block has teeth (removing `- bin/**/*`, flipping `asar: true` each fail exactly one assertion).

### Verification

`pnpm install --frozen-lockfile`, `pnpm --config.verify-deps-before-run=false -r build`, `desktop` **138 passed** (was 137), `docs` 53 passed, `typecheck` clean, `pnpm format` + `format:check` clean, `git diff --check` clean. Real `electron-builder --linux --dir` build → `verify-packed-cli.mjs` **ok: 1/1**, `penguin --version -> v0.2.6-3-g074f4b75` (the old code rejected exactly this). Windows-gate simulated by inverting `skipIf`: exactly the 2 POSIX assertions skip, 23 passed. Six mutants of `cli-link.ts` run; the size-cap one went from 0 failures to 1 after the fixture fix.

PR #247, opened 2026-08-11
feat(skills,core): remote-claude-code skill, preinstall:false manual-install marker

hiyouga:
## Summary

- **New library skill `remote-claude-code`** (AI App Development group): drive Claude Code on a remote Linux host over SSH — a persistent expect-driven login session (started with `exec_command`, driven through `input_command`), headless `claude -p` with the `< /dev/null` stdin fix, the interactive TUI inside a remote tmux via `send-keys` / `capture-pane`, and multi-turn continuity (`--session-id`/`--resume`, stream-json). Credential handling is strict: hosts/users/passwords are placeholders resolved at runtime from the user or the agent key vault, the expect template reads the password from the environment (never baked into a file), and the body carries explicit sanitization and cleanup rules. Ships with a custom line-art `icon.svg` and the mandatory `## Before you start` section.
- **New optional frontmatter marker `preinstall`** — only the literal `false` is recognized: such skills stay in the library for manual install but are excluded from default_agent's preinstalled set. `remote-claude-code` is the first (and only) skill using it, per the requirement that it must not be auto-installed.
- Both default_agent init paths honor the marker via the new `loadPreinstalledSkills()` export: the project-provisioning preset (`builtinProjectAgentPresets`) and the no-preset CLI first-run fallback in `loadOrInitAgentState`.
- Docs `/docs/skills` (en + zh): `preinstall` field row, the install-policy exception, and the new skill's library-table row.
- API and web UI unchanged: `GET /api/skills` keeps serving the skill (installable from the existing library page); `preinstall` is not exposed in responses. Existing installed agents are untouched.

Design sync: Prism-Shadow/penguin-harness-design PR (spec § "Skill 库": frontmatter field, install policy, group membership trued up to the current library; id-section note; changelog entry) — link follows in a comment once opened.

Changelog: per the changelog workflow this PR carries no `changelog/` entry; it goes into the next batch changelog PR.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check` all pass locally (Node 24).
[...]

PR #481, opened 2026-08-26
docs(landing,readme): correct copy that had stopped being true

hiyouga:
Sweep of the landing site (`packages/landing/`) and both root READMEs for statements a reader at HEAD could falsify, plus a test that keeps the Skill lists from drifting again.

## Why

The download page's first-launch FAQ was telling every visitor that "current builds are not signed yet", then walking macOS users through `sudo xattr -rd com.apple.quarantine` and Windows users through clicking past SmartScreen. Neither has applied for several releases:

- `.github/workflows/release.yml` calls `desktop-build.yml` with `require_macos_signing: true` and `require_windows_signing: true`.
- `desktop-build.yml` signs the macOS artifacts with a Developer ID cert, notarizes and staples them, and verifies the result (`codesign --verify --deep --strict`, `stapler validate`); it signs the Windows installers with EVSign and verifies the Authenticode signature.
- `changelog/0.2.2/RELEASE.md` announced macOS signing + notarization; `changelog/0.2.4/RELEASE.md` and `changelog/0.2.6/RELEASE.md` both state "so neither platform needs a first-launch unblock". v0.2.6 shipped Windows installers signed by `NaisNet Technology Co., Ltd.`

Telling a user to strip the quarantine flag off a notarized app is worse advice than silence, so both items are deleted rather than reworded. The Linux AppImage execute bit is a genuinely separate problem — browsers still strip the bit — so that item stays and is now the only one.

Once in the file, the rest of the sweep followed the same test: could a reader at HEAD verify this?

## What changed

**First-launch FAQ** (`strings.ts` / `strings-en.ts` / `download.tsx` / `links.ts`, `README.md` / `README.zh.md`) — macOS and Windows items removed; intro replaced with the signing statement; `MAC_UNQUARANTINE_CMD` removed with the item it fed.

**Built-in Skill lists** — five Skills had joined `packages/skills/skills/` without reaching the landing section or the README tables: `bento-slides` (0.1.1), `skill-porting` (#195), `remote-claude-code` (#247), `humanizer` (#256), `penguin-orchestration` (#463). Added in the order `packages/docs/content/skills.{en,zh}.md` lists them. **Now guarded** — see below.

**Supported-model table** (READMEs) — untouched since #43, and the table's own rule is "each family's latest generation only":

| Row | Was | Now |
| --- | --- | --- |
| GLM | `GLM 5.2` — Z.AI, OpenRouter, Fireworks AI, SiliconFlow, Qwen ×2 | `GLM 5.3` — Z.AI, OpenRouter, TokenDance |
| Gemini | `Gemini 3.6 Flash` | `Gemini 3.7 Flash` |
| GPT 5.6 | OpenRouter | OpenAI, OpenRouter |
| DeepSeek V4 | (no TokenDance) | + TokenDance |
| Kimi K3 | Moonshot AI, OpenRouter, Qwen PAYG | + Fireworks AI, TokenDance |
| Qwen 3.8 Max | (no TokenDance) | + TokenDance |

Derived by enumerating `packages/core/src/state/model-catalog.ts` rather than by reading the prose.

**Feature grid subtitle** — "One-to-one with the web interface's menu" / "与 Web 界面的菜单一一对应". The Web App's nav is six entries (Chat, Agents, Skills, Models, Cost Center, Evaluation Center) against nine feature cards; the standalone `/traces` page is gone and a Trace is read in its conversation's panel. Now says the capabilities are in the web interface, which is what the grid shows. This is the only line in the diff where the false claim predates the Trace-page removal — it was already loose, and losing a menu entry made it plainly wrong.

## The Skill lists are now pinned

Five of nineteen went missing because **nothing guarded those lists**. `packages/docs/test/skills-sync.test.ts` covers the docs pages only; `packages/skills/test/skills.test.ts` covered that package's own README — which is precisely why *that* table was still correct while the other three had rotted. Both unguarded surfaces are now derived from the library.

Two homes, because each surface has to be able to fail its owner's `pnpm test`:

- **`packages/landing/test/skills-sync.test.ts`** (new) — the landing package owns `src/lib/strings{,-en}.ts` and no other package can import them. It asserts each dictionary's Skills section names **exactly** the directories under `packages/skills/skills/`, names no Skill twice, and that the two dictionaries agree. It reads the directory over a relative path — the mechanism docs' `skills-sync.test.ts` already uses — so landing takes no new dependency and the lockfile is untouched.
- **`packages/skills/test/skills.test.ts`** (extended) — the root READMEs belong to no package, and this one owns the library their tables claim to list, and already had exactly this assertion for its own README. Its single `it` became a loop over `packages/skills/README.md`, `README.md` and `README.zh.md`, each parsed by locating the table's own `Group` / `分组` header row (so the three files parse identically despite different headings and `,` vs `、` separators) and compared as a set in both directions.

**Membership, not order.** The docs guard does not pin order, `SKILL_GROUPS`' exact order is already pinned in `skills.test.ts`, and a test that fails on a reordering nobody cares about gets deleted.

**Every failure names the Skill.** Verified by mutation — each case run, then reverted:

| Mutation | Result |
| --- | --- |
| drop `penguin-orchestration` from `strings-en.ts` | ✗ `Skills that ship but are missing from strings-en.ts: expected [ 'penguin-orchestration' ] to deeply equal []` |
| drop `bento-slides` from `strings.ts` | ✗ `Skills that ship but are missing from strings.ts: expected [ 'bento-slides' ] to deeply equal []` |
| drop `humanizer` from `README.md` | ✗ `Skills that ship but are missing from README.md: expected [ 'humanizer' ] to deeply equal []` |
| drop `skill-porting` from `README.zh.md` | ✗ `Skills that ship but are missing from README.zh.md: expected [ 'skill-porting' ] to deeply equal []` |
| drop `remote-claude-code` from `packages/skills/README.md` | ✗ `Skills that ship but are missing from packages/skills/README.md: expected [ 'remote-claude-code' ] to deeply equal []` |
| add a non-shipping `ghost-skill` to `strings-en.ts` | ✗ `Skills listed in strings-en.ts that no longer ship: expected [ 'ghost-skill' ] to deeply equal []` |
| list `vllm` in two groups in `strings.ts` | ✗ `Skills listed in more than one group in strings.ts: expected [ 'vllm' ] to deeply equal []` |
| rename `README.md`'s `Group` column to `Category` | ✗ `no Group/分组 table found in README.md: expected 0 to be greater than 0` (the parser fails loudly instead of silently passing on an empty set) |

All eight reverted; both suites green again.

## Checked and deliberately left alone

- **`penguin` CLI surface.** `penguin server` / `web` / `chat` / `run -m` / `config model add --provider --model-id --api-key --set-default` all still exist with those spellings; port 7364 is still `DEFAULT_SERVER_PORT`; `/compact` `/clear` `/exit` still exist. The new `ls`/`input`/`logs`/`agent`/`project`/`cost`/`schedule` family is missing from the README's CLI block, but nothing there is *false*, and documenting a new command family is a different change.
- **Hero stats, `+992 more`, the LangChain 1× / 100× comparison, the customer scenarios, the `$0.02` RAG figure.** Marketing claims, not verifiable either way from the tree; none changed under this diff.
- **Benchmark copy.** Re-derived the four cost multiples from `benchmark-data.ts`: 1/70, 1/35, 1/58, 1/39 all still round correctly.
- **`install.offlineNote`.** "one package per target (Linux / macOS in x64 and arm64, Windows in x64)" omits `penguin-universal.tar.gz`, but is true per target; the README already mentions the universal package.
- **Download links and platform requirements.** Every asset name in `links.ts` matches the real v0.2.6 Release; electron-builder still emits dmg arm64/x64, nsis x64, AppImage/deb x64.
- **The 0.2.1 / 0.2.2 blog posts** still say builds are unsigned. They are dated release records and were true when published.
- **`CONTRACT.md`** does not exist as a file, but the section renders the covenant inline and links nothing — no broken reference.

## Out of scope, needs a follow-up

`packages/docs/content/quickstart-desktop.{en,zh}.md` and `packages/docs/content/quickstart.en.md` carry the *same* false unsigned-build instructions, including the `sudo xattr` command. Not touched here to keep this diff to the landing site and the READMEs.

## Verification

Node 24.18.0, in `../penguin-harness-wt/stale-copy`.

- `pnpm install --frozen-lockfile` — clean
- `pnpm build` — all packages built
- `pnpm typecheck` — 8/8 packages pass (the new test file is inside landing's `tsconfig` `include`)
- `pnpm --filter @prismshadow/penguin-landing test` — 9 files, 69 tests pass (was 8/63; +6 from the new guard)
- `pnpm --filter @prismshadow/penguin-skills test` — 27 tests pass (was 25; the one README `it` became three)
- `pnpm --filter @prismshadow/penguin-docs test` — 7 files, 53 tests pass (incl. `skills-sync`)
- `pnpm format` + `pnpm format:check` — "All matched files use Prettier code style" (note: `*.md` is in `.prettierignore`, so the README and changelog tables were hand-aligned)
- `git diff --check` — clean
- Mutation runs for the new guard — the eight rows above

Rendered the site with Vite and read the changed pages via Playwright at `deviceScaleFactor: 2` in **en/zh × light/dark**: download page (FAQ crop + full page), `#features`, `#skills`. Horizontal-overflow probe `document.documentElement.scrollWidth === clientWidth` at 1440px and at 390px, on both the download page and the home page, in both languages. The AI App Development skill card grows from 6 chips to 8 (five wrapped rows) and stays inside its card.


hiyouga:
## Review

This PR's whole thesis is that certain statements were false and its replacements are true, so I re-derived every factual claim independently rather than reading the prose. **They hold** — including the ones that were easiest to get wrong. One fix pushed as 9ed28452, plus a design-repo PR.

### Claims re-derived from the tree, all confirmed

- **Signing.** `release.yml:89-94` calls `desktop-build.yml` with both `require_*_signing: true`. macOS: the notarize step hard-fails under `set -euo pipefail` (`notarytool submit --wait`, `stapler staple`, `stapler validate`, :413-434) plus `codesign --verify --deep --strict` (:539). Windows: `Get-AuthenticodeSignature` on the installer, the app exe and elevate.exe, with `Write-Error` on anything not `Valid` under `$ErrorActionPreference = "Stop"` (:320-340).
  I looked specifically for an **overcorrection**, since a blanket "builds are signed" would be a new false claim if any download on the page were unsigned. It is correctly scoped: the download page links only dmg/exe/AppImage/deb + `SHA256SUMS.desktop` (`download.tsx:260`; the CLI is a link-out at :353), Linux is explicitly carved out, and the README sentence sits inside `### 🖥️ Desktop app` right after the desktop installer table. The universal tarball and CLI zip are not on that page.
- **Model table.** Re-derived all 106 rows from `model-catalog.ts`. Every rewritten row is right — **including the four providers dropped** from the GLM row (fireworks / siliconflow / qwen ×2 carry only GLM-5.2, so they cannot appear on a GLM-5.3 row). "Latest generation only" still holds after the edit, and en/zh agree row for row (diffed mechanically).
- **Skill lists.** All 19 directories under `packages/skills/skills/` appear exactly once across all five files, in the right group and right order, matching `SKILL_GROUPS` (`packages/skills/src/index.ts:195-227`) and `packages/docs/content/skills.en.md:62-80`.
- **The mutation table in the description** reproduces exactly, all eight rows, including the header-rename fail-loud. The parser is separator-agnostic by construction — it regex-matches backtick spans and never splits on `,` / `、`.
- **`MAC_UNQUARANTINE_CMD`** has no dangling references anywhere. `Strings = typeof zh` + `en: Strings` as a direct object literal enforces key parity in both directions; both typechecks pass.
- **Feature grid.** `router.tsx:69-75` has no `/traces`; `app-layout.tsx:83-89` is exactly Chat + 5 pages. All nine cards do live in the web UI (schedules → `features/agents/schedules-tab.tsx`; multi-user → `strings.ts:297` 用户管理).
- **Spot-checked the "left alone" list**, since a wrong claim there would matter more than the ones you changed. Benchmark multiples: 19.41/0.552=35, 38.48/0.552=70, 220.08/3.812=58, 146.97/3.812=39 — all four round correctly. CLI surface intact (`server`/`web`/`chat`/`run -m`/`config model add …`, `DEFAULT_SERVER_PORT = 7364`, `/compact` `/clear` `/exit`), and the undocumented `ls`/`input`/`logs`/`agent`/`project`/`cost`/`schedule` family exists as described. `CONTRACT.md` is indeed not a file and `contract.tsx` links nothing.
- **Changelog pair** conforms to `changelog/README.md`; `"agent tuning"` lowercase at `strings-en.ts:486` is deliberate and consistent with `"agent hub"` :425 and `"agent evaluation"` :449; both READMEs' tables are correctly hand-aligned; `分组` in `skills.test.ts` is a load-bearing CJK literal, explicitly permitted.

---

### Fixed — the new guards pinned membership but not grouping

The guards asserted *which Skills appear*, not *which group each appears under* — and grouping is exactly the dimension this PR's own edit had to get right, since it filed five Skills into three different cards.

Confirmed by mutation, not asserted:

- moving `humanizer` from Office Productivity to Agent Tuning in `README.md` → **27 passed**, green.
- moving `humanizer` between cards in `strings-en.ts` **only**, leaving `strings.ts` disagreeing → **69 passed**, green.

That is the same class of drift the lists were just corrected for, so the guard should catch it. The README guard now parses the table into rows and matches each row against the `SKILL_GROUPS` group its heading names — `title` for the two English tables, `titleZh` for `README.zh.md` — comparing the Skills under a heading as a **set**, so order inside a row stays free and rows are matched by heading rather than by position. Your "membership, not order" principle is preserved; only the group dimension is added. The landing guard compares the two dictionaries card by card instead of as one flattened list.

Test counts are unchanged (69/27/53) — this adds assertions, not cases. All twelve mutations re-run: your original eight still fail naming the Skill, and the four that slipped through now fail naming the group and the file, e.g.

```
Skills under "Office Productivity" in README.md: expected [ 'bento-slides', …(2) ] to deeply equal [ 'bento-slides', …(3) ]
```

**Residual gap, left alone:** if *both* landing dictionaries move a Skill to the same wrong card, nothing catches it. Closing that needs `SKILL_GROUPS` inside `packages/landing`, i.e. the new dependency and lockfile change you deliberately avoided. The READMEs are now fully pinned to the manifest, so the drift would at least surface as a landing/README disagreement.

---

### Design docs are out of sync — separate PR opened

`specs/06-PROTOTYPE.md` § 打包与更新 → 下载入口 still specifies the FAQ as the **未签名首启 FAQ**, "一条折叠项 per platform (macOS quarantine deletion with a copyable `xattr` command; Windows SmartScreen 更多信息 → 仍要运行; Linux AppImage 补执行位), 访客所在平台默认展开". This PR deletes two of those three items and rewrites the intro. Per `AGENTS.md`, design is ground truth and must move with the code; `design/` is a symlink, so it cannot ride in this PR.

**[penguin-harness-design#66](https://github.com/Prism-Shadow/penguin-harness-design/pull/66)** carries the one-line spec change.

(Worth knowing: the shared local `design/` checkout is **8 commits behind `origin/main`**, and the relevant line differs between them. I verified against `origin/main` and branched from there.)

---

### Not fixed

**The docs site carries the same false instructions this PR calls "worse than silence".** `packages/docs/content/quickstart-desktop.{en,zh}.md:20,25,28` and `quickstart.{en,zh}.md:14` still say builds are unsigned and still walk the user through `sudo xattr -rd com.apple.quarantine /Applications/PenguinHarness.app` and the SmartScreen click-through. Your description declares this an out-of-scope follow-up and I have honoured that — noting only that #480 removes exactly this copy from those four files, so the two PRs together close it. Worth confirming the wording agrees; I checked, and it does.

**Two flagged, neither actionable from the tree:**

- `specs/06-PROTOTYPE.md:277` names the Windows signing route as "云签名服务（Azure Trusted Signing / DigiCert KeyLocker / SSL.com eSigner 三选一）", but the implementation uses EVSign (`mc.evsign.cn`, `desktop-build.yml:245`). Pre-existing desync, unrelated to this PR, not touched.
- `README.md:139` / `README.zh.md:139` both say "macOS 11+". `packages/desktop/package.json:23` pins Electron `^43.2.0` and `electron-builder.yml` sets no `minimumSystemVersion`, so the real floor is Electron's — which I could not determine from the tree, so I left it alone. Unconfirmed suspicion only. (Your PR body says "macOS 12+"; that is the description being loose, not the README.)

### Verification

`pnpm install --frozen-lockfile`, `pnpm --config.verify-deps-before-run=false -r build` (exit 0), landing **69 passed**, skills **27 passed**, docs **53 passed**, `pnpm -r typecheck` **8/8**, `pnpm format` + `format:check` clean, `git diff --check` clean. Counts read from printed output, not exit codes. Twelve mutations run and reverted; tree clean after each.

PR #482, opened 2026-08-26
fix(desktop,tooling): ESM bundles declare `__dirname` and `__filename`, not just `require`

hiyouga:
## The report

Connecting Feishu **from the desktop app** on 0.2.6 fails with `连接错误 / __dirname is not defined`. An npm or CLI install is fine.

## Cause

`packages/desktop/tsup.config.ts` bundles `../server/dist/index.js` and `../cli/dist/penguin.js` into one self-contained ESM file each, with `external: ["electron"]` only — so `@larksuiteoapi/node-sdk`, added by the messaging feature in #464, is absorbed whole. Bundling CJS into ESM drops the module wrapper that supplied `require`, `__filename` and `__dirname`, and the banner declared only the first of the three.

The SDK's use is narrow and unguarded:

```js
// @larksuiteoapi/node-sdk/es/index.js — getSdkVersion()
const candidates = [
  path.resolve(__dirname, '..', 'package.json'),
  path.resolve(__dirname, '..', '..', 'package.json'),
];
```

It reads its own `package.json` to put a version into its User-Agent header. `feishu-sdk.ts` loads the SDK lazily, so the undefined identifier surfaces on the first connect or credential test rather than at server startup — which is why it reads as a connection error.

Scanning the emitted `dist/server.js` for identifier references found three modules touching these globals. `@aws-sdk/core` and `undici` both guard with `typeof __dirname !== "undefined"`; the Lark SDK is the only unguarded one.

`compileEntry` in `scripts/deploy.mjs` carried the identical banner, so a hot-pushed platform would have hit the same wall. Those were the only two sites — `cli`, `core`, `server` and `skills` keep third-party deps external and absorb no CJS.

## Fix

The banner moves to one definition, `scripts/esm-cjs-banner.mjs` (sibling of the existing shared `build-git-stamp.mjs`), declaring all three. Both bundling sites import it.

The module doc states the limit plainly rather than implying full CJS emulation: inside a bundle `__dirname` names **the bundle's** directory, not the directory the dependency was published in. A dependency reading `__dirname` to find a file it ships beside itself still looks in the wrong place — the crash becomes a wrong path. The Feishu SDK stays inside that limit: its lookup already falls back to `"unknown"`, so the cost is a version tag in a request header.

**Telegram was never affected** — `telegram-api.ts` calls the Bot API with `fetch` and loads no SDK. Only Feishu.

## Verification

Reproduced end to end against the real emitted bundle: `packages/desktop/dist/server.js` spawned on a scratch `PENGUIN_HOME` exactly as `server-process.ts` forks it, then driven over HTTP through login → session create → `POST /api/sessions/:id/messaging/feishu/test`.

Before:

```
===== FEISHU TEST RESULT =====
HTTP 200
{ "ok": false, "error": "__dirname is not defined" }
```

After (same driver, same endpoint — a genuine Feishu API rejection of the fake credentials, i.e. the SDK loaded, built its client and reached the network):

```
===== FEISHU TEST RESULT =====
HTTP 200
{ "ok": false, "error": "failed to get tenant_access_token, code: 10003, msg: invalid param" }
```

New test `packages/desktop/test/esm-cjs-banner.test.ts` pins it three ways: the banner's own values, a real esbuild round-trip through a CJS fixture that reads `__dirname`/`__filename` unguarded, and both bundling sites sourcing the shared constant instead of re-inlining a partial copy.

Each fixture is evaluated in a **child `node` process**. The first draft imported them in-process and passed even with the banner reverted — Vite's SSR runner, which vitest evaluates test modules through, defines `__filename` and `__dirname` itself and masked the exact bug under test. Reverting `ESM_CJS_BANNER` to the 0.2.6 string now fails 2 of the 5 assertions.

Full chain, all green: `pnpm build`, `pnpm typecheck`, `pnpm test` (9 packages, 4037 tests, 0 failures), `pnpm format` + `pnpm format:check`, `git diff --check`, and `sh scripts/test-installer.sh` (touched `scripts/deploy.mjs`).

`scripts/esm-cjs-banner.d.mts` exists so desktop's typechecked test can import the constant the build actually uses rather than restating it; without it `tsc` raises TS7016 on the untyped `.mjs`.

## Compatibility

Emitted bundles only — no on-disk data or configuration format changes.


hiyouga:
## Review

Reviewed the full diff in a dedicated worktree, re-ran the audit the PR describes against freshly built bundles, and mutation-tested the new suite. **The diagnosis and the fix are both right** — and the audit is complete: there really are only two bundling sites, and the Lark SDK really is the only unguarded consumer. Two fixes pushed as 8046e6cf and 00f89e0f.

---

### Blockers, fixed

**1. `esm-cjs-banner.test.ts:30` — the ci-macos failure.**

The temp root came straight from `os.tmpdir()`, but Node resolves a module's real path before evaluating it, so every `__filename` a child process reports is a real path. On macOS `os.tmpdir()` is `/var/folders/…`, a symlink onto `/private/var/folders/…`, so the exact-equality assertions compared an unresolved expectation against a resolved value.

Every path in the file derives from that one root, so **all four** assertions were affected, not just the two the log named.

Fixed by `fs.realpathSync`-ing the root once. Exact `toBe` equality kept throughout — no `toContain` weakening, since exact equality with what CJS would have produced is the whole point of the test. The root is now reached through a symlink the test creates itself (`fs.symlinkSync(target, link, "junction")` — junction so Windows needs no elevation), so **Linux and Windows exercise the same resolution macOS forces**: reverting the `realpathSync` now fails on every runner rather than on one. Verified — reverting it gives `Tests 3 failed | 109 passed` on Linux with exactly the ci-macos failure shape.

`fs.realpathSync` and deliberately not `.native`: the ESM loader's `finalizeResolution` uses the former, and the two disagree about Windows 8.3 short paths (`C:\Users\RUNNER~1`). Expanding here while the loader does not would have broken ci-windows, which passes today. That is noted in the code so nobody "upgrades" it later.

**2. `scripts/esm-cjs-banner.mjs:34-36` — `const` in the banner is a latent hard failure, introduced by this PR.**

This one is worth a close look. Declaring `__filename` / `__dirname` with `const` makes the banner **collide** with any bundled module that declares the same names — and unlike `require`, esbuild does not rename them.

Reproduced against esbuild 0.28.1 directly: a bundled module's own top-level `require` is renamed (`→ require2`), so the banner's `const require` could never collide. `__filename` and `__dirname` are left spelled as written, so the banner's lexical declaration and the module's produce:

```
SyntaxError: Identifier '__filename' has already been declared
```

— at parse time, so **the whole bundle fails to load**. All four declaration forms in the bundled module (`const`, `let`, `var`, `export const`) crash under a `const` banner; all four pass under a `var` one.

The trigger is ordinary: `const __dirname = dirname(fileURLToPath(import.meta.url))` is the ESM boilerplate half the ecosystem ships. Nothing in the tree triggers it today — I scanned all five emitted bundles and none carries a top-level declaration of either name — but that is a property of the current dependency set, not of the banner, and it changes with the next bump. The failure it would produce is strictly harder than the `__dirname is not defined` this PR exists to fix.

All three declarations are now `var`. Two `var`s of one name are legal, and inside a bundle both compute the same path from the same `import.meta.url`, so the redeclaration is harmless. esbuild's `__require` shim reads the outer scope (`typeof require !== "undefined" ? require : …`) and is satisfied identically — verified in the rebuilt `server.js`. A new case bundles a module declaring both names and runs the output in a child node process; reverting to `const` fails it with the exact SyntaxError. One mirrored paragraph added to each changelog file, since the shipped bundle prelude changed.

---

### Audit re-run independently — all clear

- **Only two bundling sites.** Grepped every `tsup.config.ts` and build script for `banner` / `noExternal` / `external`. `cli` sets `noExternal: ["@prismshadow/penguin-core"]` only; `core`, `server`, `skills` set none, so tsup externalizes their deps. `packages/desktop/tsup.config.ts` and `compileEntry` in `scripts/deploy.mjs` are the only two that bundle third-party CJS. No third site.
- **No other CJS globals.** Rebuilt `packages/desktop/dist/` and scanned all five bundles. Exactly three `__dirname`/`__filename` sites, matching the PR's claim: `@aws-sdk/core` and `undici` are `typeof … !== "undefined"`-guarded, the Lark SDK's `path.resolve(__dirname, "..", "package.json")` is not. `require.resolve`: 0. `__non_webpack_require__`: 0. `require.main`: 2 hits, both inside commander doc comments. `module.exports` / `exports` only inside esbuild's `__commonJS` wrappers, which supply them.
- **The CJS-output hazard is unreachable.** Both sites are ESM-only (`format: ["esm"]` / `format: "esm"`), so the banner can never land in a `.cjs` output where `const require` would shadow the wrapper's.
- **The tests are load-bearing.** Reverting `ESM_CJS_BANNER` to the 0.2.6 require-only string gives `Tests 2 failed | 109 passed` with `ReferenceError: __dirname is not defined in ES module scope`. The PR's account of why the first draft passed in-process (Vite's SSR runner defines these itself) and why child processes were needed is accurate.
- **`scripts/deploy.mjs`** changes nothing beyond the banner string and its doc comment.
- **Design docs:** `grep` over `design/specs/` for banner / createRequire / `__dirname` / noExternal / esbuild / tsup returns zero hits. The bundling banner is not part of the documented contract, so nothing needs syncing here.
- **Changelog pair** conforms to `changelog/README.md`; English-only holds; no comment cites a design doc.

### Not fixed

`scripts/esm-cjs-banner.d.mts:6` — nothing mechanically checks the shim agrees with the `.mjs`. In practice the only export is a `string` and the test imports it and asserts on its content, so a rename or removal fails the suite rather than passing silently. Machinery would cost more than it catches.

### Verification

`pnpm --filter @prismshadow/penguin-desktop test` → **11 files, 112 passed** (was 111). `typecheck` clean. `pnpm format` + `format:check` clean. `git diff --check` clean. `sh scripts/test-installer.sh` → "Installer bundle, offline, rollback and online tests passed." All five bundles rebuilt and `node --check` clean under the `var` banner. Three mutations run and reverted, results above.

PR #483, opened 2026-08-26
test(server): pin both web.db upgrade directions

hiyouga:
## Summary

An audit of what PenguinHarness writes to a user's disk and what happens to each artifact across an upgrade found the upgrade paths themselves clean — but the property that makes them clean is undocumented, untested, and rests entirely on discipline. This PR pins it. No production code changes.

`openDatabase` had coverage for one of the three shapes a released build meets on disk: a `sessions` table formed before a column existed. Two more are added, both derived from the real `SCHEMA_SQL` rather than a hand-copied DDL replica, in the style the file already used.

- **New table on an old database.** `messaging_bindings` shipped in 0.2.5 ([#464](https://github.com/Prism-Shadow/penguin-harness/pull/464)), so every `web.db` written by 0.2.4 or earlier meets this path. The table and its `idx_messaging_account` uniqueness index arrive on open, existing Sessions are kept, and a binding can be written immediately after. The test's second `upsert` claims an account another Session already holds, so the index is asserted to be *enforcing*, not merely present.
- **A database from a newer build.** An unknown table, an unknown `sessions` column and an index over that column all survive the open untouched, and this build goes on reading and writing the rows it does own. This is the shape a user produces by updating, disliking the update and reinstalling the previous release.

## Why this, and why only this

Nothing stamps a schema version on `web.db` — `PRAGMA user_version` is `0` and there is no `schema_version` table. The migration mechanism is `CREATE TABLE IF NOT EXISTS` plus a hand-maintained `ensureColumn` list in `database.ts`. That works today only because every schema change so far has been purely additive, and it fails silently rather than loudly the first time one is not: an older build cannot tell it is looking at a newer database, so it simply proceeds.

Adding a schema-version stamp is a design decision with a long tail (what an older build should *do* on a newer database — refuse, warn, proceed?), and it changes what is on disk. That is the user's call, so it is described rather than built. The test states the additive-only condition at the exact point where a future change would break it.

## Verification

Exercised for real, not only read — three genuine checkouts (`v0.2.2`, `v0.2.4`, `v0.2.6`) built and run against scratch data roots, never `~/.penguin`:

- **Upgrade.** A `web.db` created and seeded by v0.2.4's own `openDatabase` (14 tables, no `messaging_bindings`), opened by this build: the table and its index are created, all 12 seeded rows across every table intact.
- **Downgrade, DB level.** A 0.2.6 database holding a telegram binding, opened by v0.2.4 and by v0.2.2: both open without error and leave the binding and its stored secret untouched. v0.2.2 also inserts a `sessions` row without `fork_count` / `thinking_level` / `last_active_at` — legal because every `ensureColumn` column carries a `DEFAULT` (SQLite requires it for `ALTER TABLE ADD COLUMN NOT NULL`).
- **Round trip.** 0.2.6 reopens that database: the v0.2.2-written row reads back correctly (`last_active_at` NULL is absorbed by `mapRow`'s coalesce), the binding is intact, and the stale `idx_usage_session` v0.2.2 recreated is dropped again.
- **End to end.** 0.2.6 booted on a fresh scratch root, state added, root copied; v0.2.4 booted on the copy — no errors, and every file under the Project tree byte-identical (md5 over 44 files). 0.2.6 booted again on that root: still byte-identical across the whole 0.2.6 → 0.2.4 → 0.2.6 cycle, and v0.2.4 correctly left the newer `kernel_version: 2026-08-24` stamp alone.
- **Mutation-tested.** Removing `messaging_bindings` from `SCHEMA_SQL` fails the new table test; making `openDatabase` drop an unrecognized table fails the new downgrade test *and nothing else* — the assertion is exactly targeted.

CI chain: `pnpm format` + `format:check` clean, `git diff --check` clean, `pnpm --filter @prismshadow/penguin-server typecheck` clean, `pnpm --filter @prismshadow/penguin-server test` — 92 files, 1054 tests, all passing.


hiyouga:
## Review

Reviewed in a dedicated worktree: read `database.ts` before the tests, then mutation-tested both new cases plus eight mutations the PR did not run. CI is green here, so everything below is something CI cannot catch.

The PR is sound in construction — the fixtures really are spliced into the real `SCHEMA_SQL` rather than hand-copied, the downgrade assertion really is exactly targeted, temp-dir hygiene is correct, and `Type: process` / `Scope: server` are right. **But one of the two headline claims was vacuous**, and the reasoning behind the other covers only half its cases. Fixes pushed as 1a2e6790.

---

### Blocker, fixed

**1. `db-upgrade.test.ts:359` — the uniqueness-index assertion proved nothing about the index.**

The test asserted `repo.upsert(...)` returns `{ok: false, reason: "account_in_use"}` and called that *"the index is asserted to be enforcing, not merely present"*. It is not. `MessagingBindingsRepo.upsert` short-circuits at `packages/server/src/db/repos/messaging-bindings.ts:127-130` on a `findByAccount` SELECT in application code, which answers identically over a non-unique index.

Measured: changing `CREATE UNIQUE INDEX IF NOT EXISTS idx_messaging_account` to `CREATE INDEX …` in `schema.ts:137` left the **entire 1054-test server suite green**. The PR description and both changelog files repeated the claim.

Uniqueness is now asserted on a raw `INSERT` that goes around the repo's pre-check (`toThrow(/UNIQUE constraint failed/)`), with a comment saying why the `upsert` result alone proves nothing. That mutation now fails exactly one test.

---

### Should-fix, fixed

**2. `db-upgrade.test.ts:402-404` — the DEFAULT argument covers only half the downgrade cases, and the missing half is untested.**

The comment argued that a future `NOT NULL` column must carry a `DEFAULT` "because that is what SQLite's `ALTER TABLE ADD COLUMN` requires". True for a newer build *ALTERing* an existing database — false for one that *forms* it, since `CREATE TABLE` accepts a defaultless `NOT NULL` happily.

Both shapes measured:

| Newer build's schema change | Result under this build |
| --- | --- |
| `sessions` + `future_required TEXT NOT NULL` (no default) | `openDatabase` succeeds silently, then `SessionsRepo.insert` throws `NOT NULL constraint failed: sessions.future_required` |
| `CREATE UNIQUE INDEX … ON sessions(project_id, agent_id)` | open succeeds, second insert throws `UNIQUE constraint failed` |

Both changes are **additive**, so the PR's stated invariant ("holds while every schema change stays additive") is not sufficient as written — that is exactly the kind of gap a test claiming to "pin both upgrade directions" should not leave. New test `"a defaultless NOT NULL column or a new unique index still breaks this build's writes"` covers both; its `sessions` fixture is spliced into the real `SCHEMA_SQL` and **throws if its anchor moves**, so it cannot silently stop splicing. The DEFAULT comment and both docstrings now state the real condition: additive **and** defaulted **and** no new unique index.

**3. `db-upgrade.test.ts:326-334` — handle leak on assertion failure.** `beforeTables` was opened, asserted on, then closed; a failing `expect` between the two leaks the handle, which is precisely what the file's own `afterEach` comment says Windows `rm` cannot survive. Closed before asserting.

**4. Changelog rewritten in both languages.** The uniqueness sentence now describes what is actually asserted, a third bullet records the boundary case — and the standing-state sentence *"Nothing stamps a schema version on `web.db`"* is gone: `changelog/README.md:62` forbids describing the current state of the codebase, and that sentence was also the one asserting a durability property the design docs do not carry (see below).

---

### Mutation results

| # | Mutation | Caught? |
| --- | --- | --- |
| 1 | Remove `messaging_bindings` from `SCHEMA_SQL` (PR's) | yes — though it fails in the seed helper (`no such table`), not at an assertion |
| 2 | `openDatabase` drops every unrecognized table (PR's) | yes — exactly 1 of 1054 |
| 3 | `idx_messaging_account` UNIQUE → plain INDEX | **no — 1054/1054 green.** After fix: 1 of 1055 |
| 4 | Remove `idx_messaging_account` entirely | yes |
| 5 | `ensureColumn` skips `sessions.fork_count` | yes |
| 6 | `openDatabase` drops every unrecognized index | yes |
| 7 | `ensureColumn` skips `trace_files.page_stats` | **no — 1055/1055 green** |
| 8 | `ensureColumn` skips `auth_sessions.via` | **no — 1055/1055 green** |
| 9 | `ensureColumn` skips `sessions.thinking_level` | **no — 1055/1055 green** |
| 10 | Vacuity check on the added test (schema splice made a no-op) | yes — the new test is not vacuous |

---

### Not fixed — worth their own follow-ups

**Rows 7–9: half the `ensureColumn` list is unpinned.** Pre-existing, and outside this PR's stated scope (the column direction is the one it did not touch). Deleting any of `database.ts:33,34,35` — `sessions.thinking_level`, `auth_sessions.via`, `trace_files.page_stats` — leaves the full suite green. Only `client` / `has_trace` / `fork_count` / `last_active_at` are pinned, and they are pinned by the hand-copied pre-#139 DDL at `database.ts:69-80` — itself the "hand-copied DDL replica" pattern this PR's new fixtures rightly avoid. A future cleanup of that list breaks every existing `web.db` silently. ~15 lines to fix: a `seedPreColumnDb` helper deriving from `SCHEMA_SQL` with `ALTER TABLE … DROP COLUMN` per column.

**Design docs say nothing about the direction this PR asserts.** `design/specs/05-ARCHITECTURE.md:1211` specifies only the forward path (`CREATE IF NOT EXISTS` + `ensureColumn` + one-time backfill, 存量 web.db 数据不丢). A repo-wide grep for 降级 / schema 版本 / `user_version` / 向下兼容 finds nothing about a newer database opened by an older build, and nothing about a version stamp. Removing the overclaim from the changelog means nothing in the diff now outruns the spec, so this PR no longer *needs* a design change — but the spec is silent on a property the tests now pin, and writing that paragraph is the natural companion to the schema-version-stamp question you deliberately left open. **Your call**; happy to open the design PR if you want it.

**A renamed table strands data with no error.** A newer build renaming `sessions` → `session_index`, then this build opening it: `openDatabase` recreates an empty `sessions` (0 rows), the user's rows sit invisible in `session_index`, no exception raised. Not a defect here — renaming is not additive, so it is outside the stated invariant — but it is the worst-shaped downgrade failure and worth knowing about. A changed column *type* is harmless (SQLite dynamic typing; verified).

### Verification

`pnpm --filter @prismshadow/penguin-server test test/db-upgrade.test.ts` → **10 passed** (was 9). Full server suite → **92 files, 1055 passed**. `typecheck` clean, `pnpm format` + `format:check` clean, `git diff --check` clean. Test file run 3× consecutively, 10/10 each time, no leftover temp dirs.

One aside: `pnpm --filter @prismshadow/penguin-server test packages/server/test/db-upgrade.test.ts` matches nothing — vitest's filter is package-relative, and `--passWithNoTests` makes that **exit 0 with "No test files found"**. `test/db-upgrade.test.ts` is the form that actually runs it.

PR #484, opened 2026-08-26
fix(server,core,web): Telegram poller conflicts, a missing Workspace misreported as a missing shell, and duplicate inbound messages

hiyouga:
Three defects reported against 0.2.6 — two from a user driving a Session from Telegram in the desktop app, one from the duplicate-follow-up report — plus one latent defect found on the way.

## Bug 1 — a Telegram binding parks at "connection error"

The user saw `getUpdates failed: Conflict: ter…` and no inbound message ever arrived again. The connector's module doc anticipates this 409 and claims it "resolves through the same retry path". It cannot, because the retry path was inert.

**Root cause.** The poll loop's recovery branch was gated on `bot.getMe()`. `getMe` is not the call that conflicts — Telegram's one-poller-per-token rule shows up only on `getUpdates` — so after any successful first connect the loop ran:

1. `getUpdates` → 409 → `failures = 1` → `onError` → `ready = false` → sleep `retryDelayMs(1)`
2. `getMe` → succeeds → `ready = true`, **`failures = 0`**, `onReady` → status flips back to `connected`
3. back to 1

Measured against a stub transport, 150 ms of a persistent 409 produced **162 `onError` calls, 162 `onReady` calls, 162 polls**, and the backoff the loop would have slept was `[1000]` — one value, forever. In production that is a `getUpdates` **every second indefinitely**, an `messaging_connect_failed` record every second, and a status flapping `connected`/`error` far faster than the panel polls it — so the panel reads as parked on the error while no message is ever delivered. The 1 Hz re-poll is also what makes the conflict permanent rather than transient: every retry terminates the other instance's long poll, so neither side ever gets a quiet window.

**Fix.** Readiness is proven by a `getUpdates`, never by `getMe`. The recovery poll runs with `timeoutSec: 0` so the end of an outage is observed immediately instead of after the next 30 s window, and only its success clears `failures` and fires `onReady`. One outage now reports once and walks the backoff to its documented 60 s ceiling.

Three things ride with it:

- **`deleteWebhook` once per connection**, before the first poll, keeping pending updates. A webhook and `getUpdates` are mutually exclusive on the Bot API, so a bot pointed at a webhook before it was bound here could never be polled at all. **This writes to the user's bot** — see *Deliberate side effect* below.
- **The two actionable 409s lead with the action.** `Conflict: terminated by other getUpdates request; make sure that only one bot instance is running` becomes `another program is already polling this bot — one bot token can serve only one PenguinHarness server at a time`; the webhook 409 gets its own sentence.
- **The panel shows the whole message.** `lastError` shared one flex row with the switch, the "enabled" label and the status word, truncating to `Conflict: ter`. It now gets its own line under the toggle, clamped to two lines with the full text on hover. A Telegram binding's troubleshooting fold gained a bilingual entry naming the one-program-per-token rule.

**Two pollers.** Traced `abort.abort()` / `closed` through save, disable, re-enable and shutdown. Inside one server the late-close in `MessagingBridge.connect` does close a racing connector — a new test pins that two racing `sync()` calls leave exactly one parked poller and start exactly one Task. Two *separate* servers sharing a token cannot be detected from either side; that case is now named in the error text and in the troubleshooting fold rather than fixed.

## Bug 2 — `[spawn error: spawn bash ENOENT]` was never about bash

**Node reports an unusable working directory as `spawn <command> ENOENT`.** The error carries the *command's* name, not the directory's, so a Workspace that has been deleted or moved is indistinguishable from a missing shell — and sends everyone who reads it hunting for a `bash` that was there all along.

Reproduced through the real production path (`CommandSessionManager.spawn` → `ManagedSession` → `child_process.spawn`), with `/bin/bash` present and `PATH` untouched, only the `cwd` missing:

```
bash on PATH -> true
cwd exists   -> false
model sees   -> [spawn error: spawn bash ENOENT]
```

Byte-identical to the reported string.

**What the Session sees.** A Workspace is validated exactly once — when the Session's *runtime* loads (core `resumeSession` stats `meta.workspace`; the no-Trace self-heal path stats `row.workspace` and raises 409 `workspace_missing`). It is never checked again for the life of that runtime, and neither task-start path re-checks it: `POST /:sessionId/tasks` and the messaging bridge both go straight to `manager.startTask`. So a directory that disappears under a live Session stays invisible until a command runs — and then the tool misnames the cause.

**Fix.** `CommandSessionManager.spawn` now rejects a working directory that is missing or is not a directory, before starting the child, and names the path:

```
[spawn error: working directory does not exist: /…/tmp-1a2b3c4d. The Session's Workspace
may have been deleted or moved — restore that directory, or start a new Session in one that
exists.]
```

The same check covers an `exec_command` `workdir` argument that does not resolve — a model typo now reads as a bad path instead of a missing shell. The directory is **not** auto-created: `Agent.createSession` and `assertWorkspaceAllowed` both refuse to conjure a Workspace so a typo cannot silently start working in the wrong place, and this keeps that rule.

**On "only the Telegram-driven conversation fails".** I swept both paths for anything that could make them compute a different `cwd`, and there is nothing. Both call the same `SessionManager` instance, the same `ensureEntry`, the same loader, and one `Environment` per Session; `sessionShell()` is resolved once per process and cached; nothing re-resolves, re-reads or defaults `workspace` at task start on either side. Workspace paths are stored absolute (`path.resolve` in `Agent.createSession`, or an absolute temp workspace under the Agent directory), and `~` is never expanded for a Workspace on any path. The complete list of real differences is auth, input shape, attachment writes, `queueIfBusy` (hard-coded `true` on the bridge), and error handling.

Exactly **one** surface-correlated variable exists, and it cuts the other way: `ENTRY_IDLE_MS` is 30 minutes, so a Telegram-bound Session usually sits idle long enough to be evicted and is *re-loaded* on the next inbound message — which re-runs `resumeSession`'s `fs.stat(meta.workspace)`. That makes the check **stricter**, turning a missing Workspace into a 409, not an ENOENT.

So "web works, Telegram doesn't, same install" is far more consistent with **different Sessions** than with different code paths: a binding attached to an older Session whose Workspace has since been removed or renamed, whose runtime entry was still warm in `this.entries` (where `ensureEntry` returns without any `fs` call), while the web conversations are newer Sessions with live Workspaces. The web surface also has three independent honest signals a Telegram user does not — the 409 at load, the Files panel's `workspace_missing` 404 (`WorkspaceFilesService.realBase`, and that panel refreshes every turn), and the Workspace picker — so the same dead Workspace announces itself in the browser and stays mute in the chat.

**I cannot confirm from here that this user's Workspace is in fact missing, and I am not going to claim it.** Two cheap checks settle it:

1. That Session's Workspace path (session list / binding row) — does the directory exist on their Mac?
2. **Restart the desktop app and message the bot again.** A cold load re-stats the Workspace, so if this is a dead `cwd` the error *changes* — `409 workspace_missing` / "The original Session's Workspace no longer exists" (and, because the bridge swallows a `startTask` failure, likely silence in the chat). If it stays `spawn … ENOENT` after a restart, the `cwd` is fine and the hunt continues.

Either way the fix stands on its own: the reported string could not distinguish the two, and now it can.

## Bug 3 — a redelivered chat message ran twice

Reported separately: a queued follow-up sometimes delivered twice, over Feishu *and* in the Web App. #485 established the drain is exactly-once, so the duplication is upstream of the queue — and the unguarded step is the **enqueue**. `queueIfBusy` appends unconditionally and `MessagingBridge.onInbound` had no idempotency, so any channel-side redelivery queued the same text twice and both ran. Both surfaces showed it because a queued follow-up's input is published to the Session channel.

Before, driving the bridge through the Feishu fake with one replayed event:

```
× a redelivered message starts one Task, not two
  → expected [ … ] to have a length of 1 but got 2
× dedupes a replayed non-text message too, so the text-only notice is sent once
  → expected [ … ] to have a length of 1 but got 2
× remembers across a connector restart, so a re-enable does not replay a message
  → expected [ … ] to have a length of 1 but got 2
× the same Telegram message delivered under two update ids starts one Task
  → expected [ … ] to have a length of 1 but got 2
```

**Fix.** `MessagingBridge` drops a message its binding has already processed, ahead of the chat record and the text-only reply as well as the Task start — a replayed sticker was otherwise answered with the notice twice.

- **Identity, never text.** The key is the channel's own message id (Feishu `message_id`, Telegram `chatId:message_id`), which `MessagingInboundMessage` already carries channel-neutrally. A user really does ask "status?" twice; that is two message ids and still two runs. Pinned by a test that passes both before and after the fix, so the naive-dedupe failure mode cannot creep in later.
- **Bounded on both axes** — the last 64 ids per binding, ten-minute window — because an unbounded set on a server that runs for months is a leak. A `Map`'s insertion order is the ring; eviction is a `delete` of the first key.
- **Scoped to the binding, not the connection.** Re-saving an enabled binding restarts its connector, and a message already turned into a Task must not run again because the connector object is new. Dropped when the binding or its Session is removed.
- **A connector that mints no id opts out** rather than having every message after the first read as a duplicate.

**Across a reconnect and a restart, stated:** a reconnect is covered — the memory outlives the connection. **A server restart is not**: it is in-process, so a redelivery that survives a restart still duplicates. That is deliberate. A channel replays within seconds of its own reconnect, and the restart has almost always already closed that window; persisting it would put a disk write on every inbound message to cover the remainder. If that trade is wrong, the store is the natural place and it is a contained change.

Note this and the 409 fix are the same user-visible symptom from two ends: two pollers on one token is a way the same update reaches `onInbound` twice, and cross-process duplication is something in-memory dedupe can never catch — so removing the conflict matters independently of making the path idempotent.

## Latent defect found on the way — POSIX shell resolution

Separate from the report, and **not** its cause. `resolveShell` returned a bare `"bash"` on every non-Windows platform unconditionally — no existence check, no `sh`, no `$SHELL`, no absolute path — while the Windows branch has had a fallback chain since 0.1.3/0.1.5. A machine whose `PATH` has no bash therefore fails every command before it runs. On macOS `/bin/bash` always exists, so this is not what the reporter hit; it is a real gap for trimmed GUI-session PATHs and for distributions that do not install bash.

POSIX resolution is now `PENGUIN_SHELL` → `bash` on `PATH` (still the bare name, so the common case is byte-identical) → `/bin/bash`, `/usr/bin/bash`, `/usr/local/bin/bash`, `/opt/homebrew/bin/bash` → `$SHELL` when it points at an existing shell → `sh`. The probe is a `PATH` directory walk, not a subprocess, so the existing "POSIX never probes" cost guarantee holds. A shell resolved through a non-bash step reports its own name to the model (`zsh`, `sh`), because that is the syntax it must write.

`hostEnvForChild` was checked as asked: `PATH` is not in `STRIPPED_ENV_KEYS`, has no `PENGUIN_` prefix and is not a proxy key, so it passes through, and neither `controlEnv` nor `HARDENED_ENV` touches it.

## Deliberate side effect, for review

`deleteWebhook` writes to the user's bot. It is the only way a bot previously pointed at a webhook can ever be polled — the two are mutually exclusive — and enabling a binding is an explicit instruction to make that bot this Session's. `drop_pending_updates` stays `false`, so nothing beyond the webhook registration is discarded. Say the word and it comes out; it is a five-line revert.

## Not done, on purpose

- **Task start is not blocked on a missing Workspace.** A Session whose Workspace vanished can still read and write absolute paths, search the web and answer; blocking the whole Task would take that away, and for a messaging-driven Task a thrown 409 becomes a recorded error with *no reply to the chat at all* — strictly worse than a tool result the model can relay. The spawn-time check puts the truth exactly where it is consumed.
- **The bridge still swallows a `startTask` failure** into an error record with no reply to the chat (`MessagingBridge.onInbound`'s catch). A Telegram user whose Session cannot load gets silence — which is also what makes the restart diagnostic above read as "nothing happened". Real gap, out of scope here — flagging rather than fixing.
- **`SessionsRepo.replaceId` does not cascade to `messaging_bindings`.** After a no-Trace self-heal the Session gets a new id, the `sessions` row follows it, and the binding row does not — so the next inbound message resolves a Session id that no longer exists, `ensureEntry` raises 404, and the catch above swallows it. The web client is handed the new id in the 202 response and follows it; a bound chat goes permanently silent. Separate bug, touches stored binding rows, so it is the user's call — reporting it rather than fixing it here.
- **No jitter on the backoff.** Two servers sharing a token alternate 409s and cannot self-heal at any delay; jitter would add untested nondeterminism to a case whose only fix is removing one of them.
- **Windows shell resolution untouched.** `spawn bash ENOENT` there would mean `where bash` and `spawn` disagreed — a real hazard (the resolver probes the parent's `PATH`, spawn uses the child's), but not one that can be exercised on Linux.
- **The agent vault can still overwrite `PATH`.** `...this.vault` is spread over the host env with no key filtering, so a vault entry named `PATH` replaces it wholesale for every `exec_command`. Intentional as a feature, worth knowing as a footgun; left alone.

## Verification

Node 24.18.0, Linux.

Bug 1, against a stub transport, a persistent 409 arriving after a healthy connect (150 ms of wall clock, backoff recorded rather than slept):

```
before: {"errorsTotal":162,"readiesTotal":162,"pollCalls":163,"distinctWouldSleep":[1000]}
after:  {"errorsTotal":1,  "readiesTotal":1,  "backoff":[1000,2000,4000,…,60000]}
```

The new suite run against the pre-fix connector:

```
× reports once and walks the backoff up …            → expected 126 to be 1
× recovers as soon as getUpdates stops conflicting … → waitFor timed out
× clears a stale webhook once per connection …       → expected +0 to be 1
```

Bug 2, the new tests run against the pre-fix `CommandSessionManager` — the observed value is the reported string, produced through the real `exec_command` tool with bash present:

```
× names the missing Workspace instead of blaming the shell
  → expected '[spawn error: spawn bash ENOENT]' to contain 'working directory does not exist'
× names a workdir argument that does not resolve, rather than the shell
  → expected '[spawn error: spawn bash ENOENT]' to contain 'working directory does not exist'
× names a workdir that is a file, not a directory
  → expected '[spawn error: spawn ENOTDIR]' to contain 'is not a directory'
```

and after:

```
[spawn error: working directory does not exist: /tmp/penguin-workspace-that-was-deleted-…]
```

The POSIX chain, real spawn on a `PATH` carrying `sh` and no `bash`:

```
before: spawn error -> [spawn error: spawn bash ENOENT]   output -> ""
after:  spawn error -> none                               output -> "test\nshell=/bin/bash\n"
```

Chain: `pnpm build`, `pnpm typecheck`, `pnpm --filter … test` for `core`, `server`, `web` and `desktop`, `pnpm format` + `pnpm format:check`, `git diff --check` — all clean.

**Not exercised here:** Windows and macOS packaging cannot be built on this box, so the MinGit bundle path, `PENGUIN_BUNDLED_SHELL` resolution in a packaged app, and the macOS login-shell probe are covered only by injected unit tests. The real-spawn test skips on `win32`. No test opens real network — the Telegram work runs against the injectable transport, and the transport's error mapping against a stubbed `fetch`.

## Tests added

- `packages/core/test/exec-session.test.ts` — the missing/unusable working directory suite, driven through the real `exec_command` tool: a Workspace removed under a live Session, an unresolvable `workdir`, a `workdir` that is a file, and the unchanged happy path.
- `packages/server/test/messaging-telegram.test.ts` — a persistent-409 suite (one report per outage, the backoff climbing, recovery when the conflict clears, the once-per-connection webhook clear, `close()` really ending the loop), the two-racing-syncs poller test, one Telegram redelivery test, and four transport error-mapping tests.
- `packages/server/test/messaging.test.ts` — inbound deduplication: a replayed event, a replayed non-text message, memory surviving a connector restart, the bound (an id past the cap is forgotten rather than kept forever), and the genuine-repeat guard.
- `packages/core/test/shell-resolver.test.ts` — the POSIX fallback chain plus one **real spawn** proving the resolved shell runs a command on a bashless `PATH`.
- `packages/web/test/messaging-binding-editor.test.ts` — the error detail renders whole, in its own element, outside the toggle row.


hiyouga:
## Review

Reviewed all four pieces in a dedicated worktree, mutation-tested each one, and re-derived the measured claims. **The three fixes are correct and the diagnoses hold up** — including the one I most expected to find overstated, the "Telegram-only" analysis in bug 2, which is honest about what it cannot confirm. One real defect found in the new POSIX chain, several records corrected, and one decision I'd push back on. Fixes pushed as 2ba0b07b; design PR [penguin-harness-design#68](https://github.com/Prism-Shadow/penguin-harness-design/pull/68).

---

## `deleteWebhook` — you asked; I would not ship it

Enabling a binding is consent to make that bot serve this Session. It is not consent to silently un-register a webhook the user may have pointed at a different service.

Three things compound:

- **The write is unconditional.** It fires even when there is no webhook to clear.
- **It repeats on every connection**, not once ever — so every server restart and every credential re-save clobbers a webhook the user may have re-registered in between.
- **Nothing tells them.** No log line, no status text, no UI notice. The other integration just stops receiving updates, with nothing to correlate it to. A user debugging *that* has no path back to this app.

The decisive point is that **this PR already ships the non-destructive fix for the same failure.** The mapped 409 — "a webhook is set on this bot, which blocks polling — remove it and re-enable the connection" — is actionable, correct, and puts the destructive step in the user's hands where it belongs.

And shipping both makes the good one mostly dead code (`telegram-api.ts:107`): the loop only reaches `getUpdates` after `deleteWebhook` has *succeeded*, since a failure re-enters the catch and retries the delete rather than the poll. The webhook 409 stays reachable only if someone sets a webhook while the server is already polling — and there the advice is right but the loop then backs off forever, because `webhookCleared` is already `true`. So the two ride-alongs are near-alternatives, and the PR ships the worse-behaved one in front of the better one.

If you want the convenience, gate it: an explicit confirm at enable time, or a `getWebhookInfo` probe that reports the URL and asks. **Your call — I did not revert it.**

---

## Fixed

**1. `shell.ts` — `$SHELL` accepted things that are not shells.** This is the one real defect in the new code.

The step took any existing absolute path, while the code comment, the EN changelog and the ZH changelog all promised "an existing shell". `$SHELL` carries the login shell of whatever account the server runs as, and on a service account that field is routinely `/usr/sbin/nologin` or `/bin/false`. **Both exist. Both would have been selected ahead of `/bin/sh`.** `-lc` then turns every single command into a silent non-zero exit with no output — strictly worse than the `spawn bash ENOENT` this chain was added to prevent, because it fails quietly.

Now gated on a `POSIX_LOGIN_SHELLS` allowlist of basenames known to accept `-lc` (`bash zsh fish ksh ksh93 mksh dash ash sh`); anything else falls through to `sh`, which does. New test plus mutation.

**2. `connector.ts:23` — the dedupe key had no contract.** `messageId`'s doc described it only as a reply target, while `bridge.ts` silently promoted it to the dedupe key with `""` as an opt-out. No connector author could have seen that. The seam doc now states one-id-per-message-across-redeliveries and the `""` opt-out.

**3. Module docs that contradicted the code.** `shell.ts:4` and `session.ts:4` both still asserted POSIX runs `bash -lc` unconditionally — `shell.ts` contradicted itself five lines later. Also `telegram-connector.ts:195`, whose lifecycle comment says "one `getMe` probe" when `getMe` runs ahead of *every* recovery attempt (the behaviour is right — a token revoked mid-outage stops reading as a stale conflict — the comment was wrong), and `session-manager.ts:176`, which now states the `statSync`→`spawn` TOCTOU window.

**4. Four changelog claims that overstated.** "a process still pays nothing for it" → the real cost. `$SHELL` "an existing shell" → what the code now actually enforces. The shell name reaches the model *through `{{SHELL}}`* → see below. And "**an error record** landed on every retry" → `onError` fired on every retry; `ErrorRecorder`'s 2 s coalescing window (`error-recorder.ts:99`) kept most of them off disk. The 1 Hz `getUpdates` and the status flapping were real; the disk write was not.

---

## Needs your call — the shell name does not always reach the model

`agent-state.ts:654`. The PR says "a shell resolved through a non-bash step reports its own name to the model". That is true for any Agent whose template renders `{{SHELL}}`, i.e. everything created since #79. **It is false for a pre-`{{SHELL}}` Agent on POSIX**: `withShellLineFallback` returns early on `platform !== "win32"`, so no `- Shell:` line is injected and the model goes on writing bash syntax into `zsh`, `fish` or `sh`.

The gate's own justification is *"bash was always implied there"* — which is precisely what this PR invalidated. `05-ARCHITECTURE.md:420` states the POSIX byte-identity guarantee as a design decision, so changing the gate is a spec-level change.

I implemented it (gate on `shell === "bash"` instead of on platform, keeping byte-identity wherever bash resolves), confirmed it works, then **reverted it**: it breaks an existing assertion at `state.test.ts:613` and `AGENTS.md` requires a design change be confirmed with you first. Blast radius is narrow — a pre-#79 Agent on a bashless box — but on such a box it is exactly the population that gets a non-bash shell. I qualified the changelog claim instead of leaving it absolute. Say the word and I'll open it as its own PR with the design change.

---

## Verified sound

The measured claims reproduce, and the parts most likely to hide a bug do not have one.

- **The poll loop is clean.** The recovery `getUpdates({ timeoutSec: 0 })` cannot drop or skip updates: `offset` only advances after `onMessage`, `allowed_updates: ["message"]` is applied uniformly inside the transport including the drain, and a zero-timeout poll at the same offset returns the same pending set. `failures` resets on exactly one path — a successful `getUpdates` — and nowhere else. `defaultRetryDelayMs` = `min(1000·2^(f−1), 60_000)`: no off-by-one (f=1 → 1000 ms), ceiling at f=7, stable past `2^1024 → Infinity`. `close()` can neither leave a hot spin (`sleep` resolves on abort, `while (!isClosed())` exits) nor leave the loop alive. The late-close in `MessagingBridge.connect` really does close the loser of two racing `sync()` calls.
- **The dedupe ring is correct.** `check()` deletes-then-sets, so re-dating an expired id moves it to the tail and a within-window hit is not re-dated — insertion order stays a true ring. Telegram's key is `${chat.id}:${message_id}`, so two chats cannot collide; Feishu uses its globally unique `message_id`. `messageId: string` is required by the type, so `""` is the only opt-out and it is a genuine one. The drop is at `bridge.ts:440`, ahead of `recordChat`, the text-only reply and `startTask` — pinned by the replayed-sticker test. The memory survives a connector restart. The >64-in-<10-min hole is the documented bound, not a bug.
- **The `cwd` check is well-placed.** One `statSync` per `exec_command` spawn (sole caller `exec-command.ts:87`); no cost guarantee in `design/specs/` that this violates. It only ever *rejects*, so it cannot widen what paths are reachable, and `workdir` is `path.resolve(ctx.workspaceDir, …)`'d before the check. Symlinked dirs pass. A permission-denied dir still fails at spawn with `EACCES`, unchanged. The error reaches the model through the identical `[spawn error: ${message}]` envelope, so nothing downstream sees a new shape. (Minor: the shipped string differs slightly from the one quoted in the description.)
- **The common shell case is byte-identical** — bare `"bash"`, returning before any absolute path. `onPath` walks `path.posix.delimiter`, skips empty entries (POSIX says that means CWD; skipping is right here — the Workspace is model-writable, and picking a shell out of it would be an execution primitive — now commented), never expands `~`, and costs O(PATH entries) `existsSync` calls once per process.
- **The real-spawn test does not silently skip on Linux** — I checked, because a skipped test would have hidden finding 1. Mutation 6 ran it and it failed with `spawnSync bash ENOENT`. The only core skips are `llm.e2e.test.ts` (network).
- Pre-fix `exec_command` output is byte-identical to the reported string. `drop_pending_updates: false` really is sent.

### Mutations — all eight caught

| # | Mutation | Result |
| --- | --- | --- |
| 1 | Revert poll loop to `getMe`-gated readiness | 3 failed / 27 passed |
| 2 | Remove the `deleteWebhook` call | 1 failed / 29 passed |
| 3 | Remove `isRedelivery` from `onInbound` | 4 failed / 21 passed — genuine-repeat guard still passed |
| 4 | Key dedupe on `msg.text` instead of the id | 1 failed / 24 passed — **only** the genuine-repeat guard, so it is not vacuous |
| 5 | Remove `assertUsableCwd(opts.cwd)` | 3 failed / 32 passed |
| 6 | Revert `resolvePosixShell` to bare `bash` | 6 failed / 20 passed (incl. the real spawn) |
| 7 | Revert the web error line to the truncated span | 1 failed / 6 passed |
| 8 | Drop my new `POSIX_LOGIN_SHELLS` guard | 1 failed / 26 passed |

---

## Also reported, not fixed

- **`bridge.ts:270`** — `unbind`/`unbindSession` drop the dedupe memory, but `DELETE /agents/:agentId` (`routes/agents.ts:119`) calls `sessionsRepo.deleteByAgent` and never touches `deps.messaging`, so a bulk Agent/Project delete leaves the entry until the next `start()` reconcile. The live *connection* already leaks the same way (the bridge's own comment says so), and this is ≤64 short strings per orphaned binding.
- **`telegram-connector.ts:238`** — `handlers.onError?.(err)` sits inside the catch, so a throw there escapes `poll()`, which is `void`ed at `connect()`; the poller would die silently with the status frozen. Unreachable in production (`ErrorRecorder.record` never rethrows, `setStatus` is an assignment). Left alone.

## Cross-PR with #485

**No conflict, textual or semantic.** #484 edits `strings-en.ts:1863-1873` / `strings.ts:1824-1834` (the messaging troubleshooting block); #485 edits `strings-en.ts:2051-2057` / `strings.ts:2016-2022` (the `errorCode` map). ~180 lines apart in both files, disjoint objects, disjoint subject matter — #484 adds `messaging.troubleOnePoller`, #485 adds `errorCode.follow_up_started`. Both also merge cleanly against current `main`.

Worth noting the pairing: **#485's reported bug 1 is fixed here, not there.** #485 established the drain is exactly-once and shipped a regression pin; this PR is where the enqueue actually becomes idempotent.

## Verification

core **1013 passed / 5 skipped**, server **1067 passed**, web **1360 passed**, desktop **106 passed**, cli **374 passed**, `pnpm -r typecheck` clean, `pnpm --config.verify-deps-before-run=false -r build` clean, `pnpm format` + `format:check` clean, `git diff --check` clean.

hiyouga:
## Option B implemented — `891495ec`

Per the discussion above, `deleteWebhook` is out; a read-only `getWebhookInfo` probe takes its slot.

**What changed**

- `TelegramBotClient.deleteWebhook()` → `getWebhookInfo(): Promise<TelegramWebhookInfo>` (`{ url: string }`; the empty string is the Bot API's own encoding of "none", not a normalization).
- The connect path probes once, before the first poll. If a webhook is registered it throws with **the URL in the message** and never reaches `getUpdates`:

  > `a webhook is registered on this bot at https://…, which blocks polling — remove it there and the connection recovers on its own`

  That URL is the entire point of the change. The mapped 409 could only ever say a webhook *was set*, never *which one* — and finding it is the whole of the user's task.
- The flag latches **only on a clean answer**, so every retry re-probes. Removing the webhook elsewhere needs no re-enable: the next probe passes, the drain runs, `onReady` fires, status returns to `connected`. The loop notices, not the user.
- The generic 409 copy is corrected accordingly: "remove it and **re-enable the connection**" → "remove it and **the connection recovers on its own**". It was wrong even before this change — traced through the loop, a cleared webhook always self-healed within the ≤60 s backoff.
- Changelog pair rewritten in both languages: the bullet now describes the probe, says why the delete is not this app's call, and states the no-re-enable property.

This also resolves the dead-code problem from the review: with nothing clearing the webhook, the `"webhook is active"` 409 mapping is reachable again in the case it was written for.

**Not added:** a troubleshooting-fold entry for the webhook. The one-poller entry exists because that 409 *cannot* name the other program; this one names the webhook in the error itself, so a fold entry would only restate it.

**No spec moves with this** — `grep` over `design/specs/` for `webhook` / `getUpdates` returns nothing. [penguin-harness-design#68](https://github.com/Prism-Shadow/penguin-harness-design/pull/68) is unaffected; it covers the shell chain and the `cwd` check.

**Mutations** (each reverted; all caught by the new test alone, none by anything else):

| Mutation | Result |
| --- | --- |
| probe result ignored, webhook tolerated | 1 failed / 30 passed |
| message drops the URL | 1 failed / 30 passed |
| flag latches even when a webhook was found (no self-heal) | 1 failed / 30 passed |

**Verification:** server **1068 passed** (was 1067; +1 for the new case), web **1360 passed**, core **1013 passed / 5 skipped**, `pnpm -r typecheck` exit 0, `pnpm format` + `format:check` clean, `git diff --check` clean.

Still open from the review: the `- Shell:` line is not injected for a pre-`{{SHELL}}` Agent on POSIX, so on a bashless box such an Agent keeps writing bash syntax into `zsh`/`sh`. That one changes a stated guarantee in `05-ARCHITECTURE.md:420`, so it wants its own PR with the design change — say the word.

🤖 Implemented with [Claude Code](https://claude.com/claude-code)

PR #485, opened 2026-08-26
fix(server,web): a queued follow-up is recallable however it was queued

hiyouga:
Two linked reports on 0.2.6, from a Session that was busy when the user sent a message and showed the queued-follow-up chip. One of them reproduces; the other, on investigation, is not a defect of the follow-up queue and no speculative fix for it is included here. Both findings are below.

## Reported bug 2 — recall says "already sent" on a message still queued

Reproduced deterministically. The messaging bridge queues an inbound Feishu/Telegram message by calling `SessionManager.startTask(sessionId, [userText(text)], { queueIfBusy: true })` — the manager directly, not the HTTP route — and passes no `recall` store, because only the route knows the parts of one the input does not carry (the pre-attachment text, and where each file attachment landed in the Session scratchpad). `QueuedFollowUp.recall` was optional, so such an entry was queued without it, and two things followed:

- `followUpInfo` reported `text: ""`, so the Web App composer drew the queued line with nothing after its colon;
- `recallFollowUp` refused any entry with no recall store, as 409 `not_pending` — which the Web App renders as "该消息已发出，无法撤回。" / "This message already went out and can no longer be recalled." The message had not gone out. It was still in the queue after the refusal, so the line stayed on screen, and clicking recall again said the same thing.

Before the fix, the new test's first assertion fails on the bridge path (`text: ""` where the message belongs), and a probe against the DELETE returned `409 not_pending` with `pendingFollowUpCount === 1` still standing. After it, both pass.

**Fix.** `startTask` derives the recall store from the input when the caller supplies none (plain user text joined, inline `image_url` payloads collected, `files: []` — only the HTTP route writes attachments to disk). `QueuedFollowUp.recall` became required, which removes the "queued but not recallable" state rather than papering over it, and `recallFollowUp`'s only remaining refusal is "the id is not in the queue". No change to `bridge.ts` was needed.

**Copy.** The two recalls no longer share one code, because they mean different things to the user. A follow-up refused because it already became a task of its own is 409 `follow_up_started`; a steering message the model already received keeps `not_pending`. Both dictionaries split with the codes — zh 「该跟进消息已开始发送，无法撤回。」/「该插话已随本轮送达模型，无法撤回。」, en "This follow-up already started and can no longer be recalled." / "This steering message already reached the model and can no longer be recalled." Unmapped codes fall back to the server's English message, so a stale bundle degrades rather than breaks. `server-api.{en,zh}.md` and `RecalledMessageResponse`'s doc record the split.

## Reported bug 1 — "the message is sometimes delivered twice"

**Does not reproduce as a follow-up-queue defect, and this PR ships no fix for it — only a test pinning the invariant.**

- `startQueuedFollowUp` has exactly **one** trigger site: `drive`'s `finally`. The two other lines that look like triggers are a guard inside `startQueuedFollowUp` itself and `startBackgroundNoticeTask`'s own guard, which belongs to the background-notice queue, not this one.
- Its guard-then-`shift()` sequence runs entirely synchronously inside `withLock`; no `await` lands between the length check and the shift, so there is no re-entry window. `drive`'s `finally` is likewise synchronous from the idle flip to the fire-and-forget call, so no HTTP handler can interleave between them.
- Exercised at the manager/service level with a fake session across: run completion, abort (with one and with two queued entries), a task POST fired from inside the idle broadcast, a compaction started in the same gap, an agent-config generation bump mid-run, and the existing recall-versus-drain race. Every interleaving delivered each queued input exactly once, in order.
- The composer cannot send twice for one action either: steer and queue are exclusive branches of `send()`, and the steer→409 fallback runs only when the server refused the steer.

What *is* unguarded is the **enqueue**: `queueIfBusy` has no idempotency, and `MessagingBridge.onInbound` does not dedupe by `messageId`. Any channel-side redelivery — a Feishu long-connection reconnect, or a second Telegram poller, which is exactly the 409 poll conflict under repair on another branch — queues the same text twice, and both copies run. Both runs are then visible in the Web App as well, because a follow-up's input is published to the Session channel when it starts; that is the most likely reason one duplicate was reported as two channels. `packages/server/src/runtime/messaging/` is another agent's working tree, so nothing there was touched.

## Tests

- `packages/server/test/followup.test.ts`
  - **new** — "recall (#287) works on a follow-up queued straight through the manager (the messaging bridge's path)": pins the queued line's content and a successful recall on the non-route path. Fails before this change.
  - **new** — "one queued follow-up starts exactly one task, whether the run completes or is aborted": pins exactly-once across both ways a run ends, with a settle window that would catch a second launch. Passes before and after — it is a regression pin, not a reproduction.
  - the existing recall test's 409 expectation moved to `follow_up_started`.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm format` + `pnpm format:check`, `git diff --check` — all clean. Tests: server 1054 passed (92 files), web 1359 passed (102 files), core 1000 passed / 5 skipped (46 files), docs 53 passed.


hiyouga:
## Review

Reviewed in a dedicated worktree, mutation-tested the new tests, and checked the compatibility question the type change raises. **No blockers — the production change is sound.** The recall derivation is correct and complete, the 409 split is complete on both sides, and the exactly-once claim holds under the reasoning you give. One test fix pushed as 6b94e8d7, plus a design-repo PR that closes a gap predating this change.

### The compatibility question, answered

Making `QueuedFollowUp.recall` **required** is the highest-risk edit here, so I chased it: `followUps` lives only on `RuntimeEntry` in `SessionManager.entries` (a `Map`). No `db/schema.ts` table, no repo, no file, no snapshot; `QueuedFollowUp` is not exported, and there is exactly one producer. The queue dies with the process. The wire type `PendingFollowUpInfo` did not change — its `text`/`images`/`files` were already required; only the *value* went from `""` to the real text on the bridge path.

So **no `Breaking:` line is owed**, and removing the "queued but not recallable" state rather than papering over it is the right call. Bar check for the 409 rename: the 0.2.6 entry that *deleted* two model-facing tools (`kill_subagent` / `kill_command`) carries no `Breaking:` line either.

### Verified sound

- **Derivation is complete, not just adequate.** `parseTaskInput` (`sessions.ts:289-329`) only ever builds `userText(...)` and `imageUrlMessage(...)` — exactly the two shapes `recallStoreOf` handles, so there is no third part kind it silently drops. `files: []` is correct on every non-route path: the bridge sends text only, and `scheduler.ts:418` never passes `queueIfBusy`, so it 409s instead of queueing.
- **The route cannot be shadowed.** `sessions.ts:952` always passes `recall: recallStore(...)`, and `??` only fires on `undefined`.
- **The 409 split is complete.** Both codes in both dictionaries — 72 keys each, same order, no en-only or zh-only key (checked programmatically, and `en: Strings` makes typecheck enforce it). The fallback is real: `api-error.ts:24-25` is `byCode[err.code] ?? err.message`. Repo-wide `not_pending` grep: only `recallSteering` throws it; every other hit is steer docs/tests/comments or the frozen `changelog/0.2.3/` entry, correctly untouched. Nothing in `cli`/`desktop`/`core` switches on either code. One line each in `server-api.{en,zh}.md` is enough — there is no other code table.
- **Exactly-once holds.** `startQueuedFollowUp` has one definition and one call site (`session-manager.ts:1833`). No `await` between the length check and `shift()` (`findById` is synchronous better-sqlite3); no `await` in `drive`'s `finally` between `entry.status = "idle"` and the fire-and-forget call (`touchRow` and `publishState` are both sync). `withLock` chains via `prev.then(fn)`, so an HTTP handler landing during the idle publish serialises and merely sees `status !== "idle"`.
- **Web side.** `steeringSummary` (`chat-input.tsx:669`) already normalised empty text, so the now-populated text just fills the line. The steer→409 fallback (`chat-input.tsx:1818`) keys off `status === 409` from `POST /steer`, which only ever emits `not_running` — untouched by the split. The bridge's channel tap reads only `task_state.state`, so the populated text never echoes back to Feishu/Telegram.

---

### Fixed — the abort half of the exactly-once test pinned nothing

`followup.test.ts:258-264` aborted a run with an **empty** queue: the follow-up had already become run 2 and drained before the abort landed. So the abort half asserted nothing about the drain, while its own comment claimed the opposite ("a queued follow-up deliberately survives an abort of the run under way").

Demonstrated: mutating `abortTask` to `entry.followUps = []` — a direct violation of the specced 「中止**不**丢弃排队消息」 — leaves the whole file green.

The test now queues a third message behind run 2 before aborting it, asserts exactly one third run and an empty queue, and settles the last run so teardown does not race it. That mutation now fails. Ran the file 8× solo: no flake.

Also dropped "the SDK" from `recallStoreOf`'s doc — `SessionManager` is on no export of `@prismshadow/penguin-server`, so `MessagingBridge.onInbound` (`bridge.ts:365`) is the only caller that supplies no store.

---

### Design docs — the gap is older than this PR

Worth knowing: **the recall feature never reached the specs at all.** `DELETE /steer/:steerId` and `DELETE /follow-ups/:followUpId` shipped in 0.2.3 (#287) and appear in neither the API prototype list nor the two bullets describing the queues they act on. The follow-up bullet says only 「会话运行状态携带排队数」 — a count — while the implementation has published a per-entry mirror with content since that release, which is exactly what the composer's queued line renders.

This PR then changes behaviour on top of that undocumented ground. **[penguin-harness-design#67](https://github.com/Prism-Shadow/penguin-harness-design/pull/67)** closes both: the steering and follow-up bullets gain their recalls, the follow-up bullet gains the per-entry mirror and the rule this PR establishes (*every queued entry carries its content whichever path queued it*) plus `follow_up_started` as its own code, and the composer bullet in `06-PROTOTYPE.md` stops describing a bare count.

---

### One thing to state plainly

**Reported bug 1 ("delivered twice") is not fixed by this PR.** Your investigation is right — the drain is exactly-once, the duplication is upstream at the enqueue, and `queueIfBusy` plus `MessagingBridge.onInbound` have no idempotency. But what ships here for that report is a regression pin on an invariant that was never broken. The reported symptom needs its own change, and #484 is where it lands: that PR adds the `messageId` dedupe in `MessagingBridge`. Worth saying in the description so the report is not read as closed by this PR.

The scope call itself is correct — this PR is about recall, and the messaging directory was another working tree.

### Not fixed

- `session-manager.ts:441-450` — the image half of `recallStoreOf` omits the "no `origin`" condition the text half gets from `isPlainText`. Unreachable today: no origin-tagged input is ever queued, since the only `queueIfBusy` callers are the route (own store) and the bridge (one plain `userText`). Latent trap, not a live bug; guarding a caller that does not exist would be defensive code against the file's style.
- Pre-existing, out of scope: an evicted or restarted runtime answers a recall with `follow_up_started` ("already started") when the message will in fact never be sent. And `server-api.md`'s SSE wire sample at line 373 still shows `task_state` without `queued` / `pendingFollowUps`.

### Verification

server **1054 passed** (92 files), web **1359 passed** (102 files), docs **53 passed**, `followup.test.ts` **6 passed** ×8 solo with no flake, `pnpm -r typecheck` 8 projects clean, `pnpm --config.verify-deps-before-run=false -r build` clean, `pnpm format` + `format:check` clean, `git diff --check` clean.

Mutations: full revert of `session-manager.ts` → the bridge-path test fails on `text: ""` and the existing recall test fails on `follow_up_started` (confirms "fails before this change"); `startQueuedFollowUp` launching twice → 4 of 6 red; `abortTask` clearing the queue → **missed before, caught after**.

PR #488, opened 2026-08-27
ci: fan the gates out into short parallel jobs, and absorb the Windows flakes

hiyouga:
CI was **4.3 minutes** of wall clock and failed often enough on Windows to cost re-runs. Both come from the same shape.

## What was slow

Measured on a clean run (33026406079), per step:

| | ubuntu | macOS | Windows |
| --- | --- | --- | --- |
| `pnpm/action-setup` | 5s | 9s | **21s (63s at its worst)** |
| `setup-node` | 3s | 13s | 16–20s |
| install | 7s | 12s | 11s |
| build (tsup, all nine packages) | 27s | 34s | 34–40s |
| tests | 78s | 105s | 93–**138s** |
| **job total** | 180s | 217s | 228s / **254s** |

Every platform ran build → style → typecheck → test → installer → e2e as serial steps in one job, so the slowest job set the whole run. `ci-windows (server)` was the pole at 254s.

## What changed

**Each gate is its own job.** `style` needs no build at all — Prettier reads sources — so it is now the cheapest job in the run instead of 17s buried behind a 27s build.

**Tests are sharded by package**, and a shared composite setup action (`.github/actions/setup`) builds **only the packages a shard needs**. Measured locally: core's closure builds in 8s against 22s for the full graph, server's in 15s.

**Coverage is unchanged.** The shards sum to all 297 test files on every platform, with no file running twice — verified by running each shard command and adding up:

| Platform | shards | files |
| --- | --- | --- |
| ubuntu | core 46 + server 92 + web-cli 130 + rest 29 | 297 |
| macOS | core 46 + server 92 + rest 159 | 297 |
| Windows | core 46 + server-1 46 + server-2 46 + rest 159 | 297 |

Windows splits the server suite across two vitest `--shard`s on top of the per-package split, because it is the longest suite on the slowest runner.

**pnpm comes from corepack** rather than `pnpm/action-setup`. `packageManager` pins the same version either way; the action was simply the most expensive single step on Windows.

## Flakes, handled at the level each one occurs at

`retry: 2` on win32 in core's and the server's vitest configs covers the **per-test** kind. These suites spawn real shells and read their output under deadlines, and the Windows runners lose those races at a measurable rate — over one batch of reruns during the 0.2.7 release, the Windows shard failed **4 attempts in a row and then passed twice**, on four *different* tests, from a diff (a version-string bump) that could not change behaviour at all. Retrying the individual test costs nothing when everything passes, and a real regression still fails all three attempts. POSIX keeps `retry: 0` so a flake introduced there is seen the first time.

What per-test retry cannot catch is a failure with **no failing test**: a tinypool `ERR_IPC_CHANNEL_CLOSED` raised while the worker pool tears down, after every file has already reported green. The Windows test step therefore retries the whole command once, and says so in a `::warning::` when it does. A real regression fails both attempts.

Deliberately **not** done: `dangerouslyIgnoreUnhandledErrors`. It would silence the teardown crash for free, but it would also silence a genuine Windows-only unhandled rejection — which is exactly the class of bug this gate exists to catch.

## One thing a maintainer has to do

`ci` is now an **aggregate job** gated on every other one, and it is the name worth marking required in branch protection. The individual shard names are an implementation detail and will change again as the suite grows; `ci` keeps meaning what it has always meant. **Branch protection currently lists the old job names**, so the required-checks list needs updating to just `ci` when this merges, or the old names will sit pending forever.

## Verification

Every shard command was run locally and its file/test counts recorded (the table above). `pnpm -r typecheck` clean, `pnpm format` + `format:check` clean. The workflow and the composite action both parse as YAML.

Real wall-clock numbers from this PR's own CI run are in a follow-up comment — that is the measurement that matters, and it is what this change is for.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -21,9 +21,26 @@ inputs:
 runs:
   using: composite
   steps:
+    # corepack fetches the pinned pnpm tarball on first use -- ~20s on the Windows runner,
+    # which is most of what replacing pnpm/action-setup was meant to save. Cache what it
+    # unpacks, keyed on the packageManager pin, and the fetch happens once per pin per OS.
+    - name: Cache corepack's pnpm
+      uses: actions/cache@v4
+      with:
+        path: |
+          ~/.cache/node/corepack
+          ~/AppData/Local/node/corepack
+        key: corepack-${{ runner.os }}-${{ hashFiles('package.json') }}
+
     - name: Enable pnpm (corepack)
       shell: bash
-      run: corepack enable
+      env:
+        COREPACK_ENABLE_DOWNLOAD_PROMPT: "0"
+      run: |
+        corepack enable
+        # Materialise the pinned version now so the cache above has something to save, and
+        # so a download failure surfaces here rather than inside an unrelated step.
+        pnpm --version
 
     - uses: actions/setup-node@v5
       with:
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,51 @@
+# Shared prologue for every CI job: pnpm on PATH, Node 24, dependencies installed.
+#
+# pnpm comes from corepack rather than pnpm/action-setup. The version is the one
+# `packageManager` pins either way, but the action was the single most expensive
+# step on the Windows runners — 21s typical and 63s at its worst against ~5s on
+# ubuntu — while `corepack enable` is a shim install with no download of its own.
+# It has to run before setup-node, whose `cache: pnpm` resolves the store path by
+# executing pnpm.
+name: Set up pnpm, Node and dependencies
+description: Enables pnpm via corepack, installs Node 24 with a warm pnpm store, and runs a frozen install.
+
+inputs:
+  build:
+    description: >-
+      pnpm --filter arguments naming what to build, comma separated (e.g.
+      "@prismshadow/penguin-core..."). Empty builds nothing — a job that only reads
+      sources, such as the Prettier gate, needs no dist at all.
+    required: false
+    default: ""
+
+runs:
+  using: composite
+  steps:
+    - name: Enable pnpm (corepack)
+      shell: bash
+      run: corepack enable
+
+    - uses: actions/setup-node@v5
+      with:
+        node-version: 24
+        cache: pnpm
+
+    - name: Install dependencies
+      shell: bash
+      run: pnpm install --frozen-lockfile
+
+    # Only the requested packages and their dependencies. A shard that tests core
+    # alone does not pay for the desktop bundle, which is the slowest target here.
+    # verify-deps-before-run is off because it wants a TTY it does not have on a runner.
+    - name: Build (tsup)
+      if: inputs.build != ''
+      shell: bash
+      run: |
+        args=""
+        IFS=','
+        for f in ${{ inputs.build }}; do
+          args="$args --filter $f"
+        done
+        unset IFS
+        echo "pnpm$args build"
+        pnpm --config.verify-deps-before-run=false $args build
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -37,13 +37,23 @@ runs:
     # Only the requested packages and their dependencies. A shard that tests core
     # alone does not pay for the desktop bundle, which is the slowest target here.
     # verify-deps-before-run is off because it wants a TTY it does not have on a runner.
+    #
+    # The empty check is in the script, not a step-level `if`. A composite step whose `if`
+    # reads `inputs` silently evaluated false here, so every job skipped its build and only
+    # the shards that never import a built package stayed green -- a failure that looks like
+    # a broken test, not a broken workflow. The echo makes the decision visible in the log.
     - name: Build (tsup)
-      if: inputs.build != ''
       shell: bash
+      env:
+        BUILD_FILTERS: ${{ inputs.build }}
       run: |
+        if [ -z "$BUILD_FILTERS" ]; then
+          echo "No build requested for this job."
+          exit 0
+        fi
         args=""
         IFS=','
-        for f in ${{ inputs.build }}; do
+        for f in $BUILD_FILTERS; do
           args="$args --filter $f"
         done
         unset IFS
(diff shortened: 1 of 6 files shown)


hiyouga:
## Measured results

Iterated against real CI. Wall clock is first-job-start to last-job-end on the run's own jobs.

| | before | after |
| --- | --- | --- |
| **wall clock** | **254s** | **129s** |
| slowest job | `ci-windows (server)` 254s | `test-windows (server-2)` 123s |
| jobs | 7 | 22 |

The whole distribution moved, not just the pole:

```
 123s  test-windows (server-2)      66s  test-macos (server)
 119s  test-windows (server-1)      60s  test (web-cli)
 115s  test-windows (server-3)      57s  test (server)
 114s  runtime (windows --win)      49s  runtime (ubuntu --linux)
 111s  test-windows (core-1)        44s  installer-e2e
 108s  test-windows (rest)          41s  test (core)
 108s  installer-windows            39s  test (rest)
 107s  test-windows (server-4)      38s  style (prettier)
  97s  runtime (macos --mac)         2s  ci
  90s  test-macos (rest)
  85s  test-windows (core-2)
  75s  typecheck
  69s  test-macos (core)
```

### Where the time went

Three measurements, each acted on:

1. **The build was running seven times** and building everything each time. Now each shard builds only what it loads — and the measurement that mattered was which packages a suite *actually* needs, not which it nominally depends on. The server shards were building `penguin-server...`, which drags in web because the server package ships `web-dist`; the tests read none of it. Building `penguin-core...` into a clean tree gives the same **1073 tests**, so that is what they build. Same for `rest`: `penguin-desktop...` alone covers landing, docs and skills.

2. **The Windows installer checks were riding the longest shard**, adding 20s to the critical path from a job they have nothing to do with. Own job now.

3. **Windows setup costs 55–81s before a test runs** — the same work, that much variance. With the build already minimal, the only remaining lever on that platform is giving each job less to do, so core splits in two and the server in four *on Windows only*. ubuntu and macOS keep whole suites; their setup is cheap enough that more jobs would cost more than they save.

`pnpm/action-setup` → `corepack` also went in (same pinned version, and that action was the single most expensive step on Windows).

### Coverage is unchanged

Every package's tests run exactly once per platform. Verified by running each shard command and adding up — ubuntu 46+92+130+29, macOS 46+92+159, Windows (23+24)+46+46+... — each summing to the repo's **297 test files**, with core's split reproducing its 46 passing plus 1 skipped.

### Two bugs this PR shipped and then fixed, both worth knowing

- **A double-quoted string inside a `${{ }}` expression.** Only single quotes are legal there. The run had **zero jobs**, was named after the file, and left no annotation on the run, the check suite or the commit status. `actionlint` reads it in under a second, so the `style` job now runs it over every workflow.
- **A composite step whose `if:` read `inputs` silently evaluated false**, so the build never ran in *any* job. Only the shards that import nothing built stayed green — which reads as a broken test, not a broken workflow. The empty check moved into the script, and the step now echoes what it decided.

The second one is why my local pre-flight was not enough: I had validated the build commands and the test commands separately, and never the **pairing**. Every shard is now checked as build-into-a-clean-tree then run-its-tests.

Pass-rate measurement over repeated runs is in flight; numbers to follow.

hiyouga:
## Correction, and the final numbers

An earlier comment on this PR reported **129s**. That was a single lucky sample and I should not have led with it. Measured properly — repeated `workflow_dispatch` runs, wall clock as first-job-start to last-job-end, the same metric applied to the baseline:

| config | jobs | wall clock | pass |
| --- | --- | --- | --- |
| **before** | 7 | **254 / 255 / 257s** (queue 1–2s) | — |
| 22-job split | 22 | 180 / 166 / 191 / 139 / 193 / 201 / 179 / 202 / 199s — **median 191s** | 10/10 |
| **this PR** | 18 | 264 / 154 / 137 / 179 / 152s — **median 154s** | 5/5 |

**15/15 green across both configurations**, so the >95% target is met with room. Wall clock is **254s → median 154s, ~40% off**, but it is **not under two minutes** and I want that stated plainly rather than buried.

### Why two minutes was not reachable

Splitting further stops helping, and then starts hurting. On GitHub-hosted runners a job waits for a machine before any step executes — measured per OS, from one representative run:

| | jobs | queue: median | max |
| --- | --- | --- | --- |
| ubuntu | 8 | **0s** | 47s |
| macOS | 4 | 42s | 74s |
| Windows | 9 | **70s** | 83s |

That wait is paid **per job**, so each extra Windows shard buys a fresh provisioning delay instead of removing work from the critical path. The 22-job version proved it: more shards, *worse* median (191s vs 154s). The old 7-job workflow queued 1–2s in total — sharding is what introduced the queueing.

On top of that, a Windows job costs ~55s of its own setup. Best case ever observed for *any* Windows job, with zero queue, is a 78s finish. So a run containing a Windows gate has a floor around 120–130s before its slowest suite is considered, and lands ~150–200s in practice.

### What would get under two minutes, and why it is not in this PR

Taking macOS and Windows off the pull-request path (running them on pushes to main and on dispatch) puts PR feedback at **~75s** — the ubuntu jobs never queue. I implemented that, and **@hiyouga rejected it**, correctly: it trades away PR-time platform coverage, which is a maintainer's call and not mine to make. Three `if:` lines would do it if that trade is ever wanted.

### One thing I tried that failed, recorded so nobody repeats it

The server suite is I/O-bound, so oversubscribing vitest workers looked promising, and at four cores it was a large win: **48.9s at one worker per core, 29.7s at two per core, 33.0s at three**. But `availableParallelism() * 2` unbounded breaks the suite on a bigger machine — on 8 cores (16 forks, with `isolate: false`) it dies with `Error: Terminating worker thread`. It would need a cap and its own validation, so it is **not** in this PR.

### Branch protection needs one change

`ci` is now an aggregate job gated on all the others, and the individual job names have changed. **The required-checks list must be set to `ci`** when this merges, or the old names will sit pending forever on every PR.

hiyouga:
### Correction: no branch-protection change is needed

I claimed above that the required-checks list has to be repointed at `ci`. I checked after merging, and that is **wrong for this repo** — there is no required-checks list to update.

`main` is governed by a ruleset (id `19162097`), not classic branch protection, and its rules are `deletion`, `non_fast_forward`, `creation`, `pull_request` and `copilot_code_review`. There is no `required_status_checks` rule in it, nor in the `release` tag ruleset. So the renamed jobs break nothing, and nothing will sit pending.

The `ci` aggregate job is still worth having — it is one name that stays meaningful as the shard list changes, and it is the sensible thing to require *if* required checks are ever turned on. But that is an option, not a chore this merge created.

PR #490, opened 2026-08-27
feat(server,web): enabling a messaging connection is the binding

hiyouga:
A Feishu app or a Telegram bot used to belong to one Session forever. The first Session to save its credentials owned the account — a unique index on `messaging_bindings(channel, account_id)` made the second Session's **save** fail, answered as 409 `feishu_app_in_use` / `telegram_bot_in_use` — and the Web App offered no unbind affordance at all. A user who wanted to move their bot to another conversation had no path through the product.

This replaces that with the model the switch already looks like: **enabling the connection is the binding, disabling it is the unbind.**

## Semantics

1. **Saving never conflicts across Sessions.** Any number of Sessions may keep the same Feishu app or Telegram bot saved side by side, each with its own stored config and its own remembered chat. `MessagingBindingsRepo.upsert` lost its failure mode and now returns the row; both PUT handlers lost their 409.
2. **Enabling is refused only while another Session has the same `(channel, account_id)` enabled** — new 409 `account_enabled_elsewhere`. One account has one event stream, so two live connections on it are meaningless; two saved configs are not.
3. **The per-Session rule is unchanged**: at most one of a Session's own channels is enabled at a time (409 `another_channel_enabled`).

`findByAccount` (single row, any state) became `findEnabledByAccount(channel, accountId)` — at most one row can match for the same reason `findEnabled(sessionId)` is at-most-one: the state route refuses the second enable, and its guard-then-flip runs in one synchronous block, so two requests cannot interleave between the check and the write.

## The authorization decision on the conflict message

**The refusal names nothing about the holder**, deliberately. Authorization is per Project, and the state route only proved the caller's access to *its own* Session; the Session holding the connection may sit in a Project this user cannot see, so its title or id would be a leak. Naming it only when the caller happens to have access was the alternative, and it was rejected on three grounds: it adds a branch whose only job is to not leak (the failure mode of which is a leak); the two branches would differ in shape and timing, which is itself a signal; and the web's `apiErrorText` maps a code to *localized* text, so a server-supplied holder name could only reach the user as untranslated English. The remedy — turn that connection off where it is on — does not depend on hearing which Session it is.

## Database migration, and its one-way half

- `schema.ts`: the `CREATE UNIQUE INDEX idx_messaging_account` became a plain `CREATE INDEX idx_messaging_by_account` over the same two columns, so the enable guard's by-account lookup stays indexed and no query plan changes. **The new index needs the new name**: `openDatabase` runs `SCHEMA_SQL` before its drops, so reusing the old name would have the drop delete the index just created — the same reason `idx_usage_session` became `idx_usage_session_ts`.
- `database.ts`: `DROP INDEX IF EXISTS idx_messaging_account` next to the existing `idx_usage_session` drop. Dropping is safe because an index is derived, never data: no row is touched, and on a database that never had it the statement is a no-op.
- **One-way.** A build from before this change recreates the unique index as part of its own `SCHEMA_SQL`, and that `CREATE UNIQUE INDEX` fails once duplicate `(channel, account_id)` rows exist. `openDatabase` runs the schema first, so the older build fails *at open* — it does not start against that database at all. Downgrading after two Sessions have saved one bot means deleting the extra rows first. `changelog/unreleased/2026-08-26-backward-compatibility.md` records this, including the query to find such rows and when the drop line may be retired (with the `idx_usage_session` drop beside it, in a release allowed to break existing `web.db` files).

Nothing is asked of the user: the drop runs automatically on the first open of an updated build, and no binding, credential or remembered chat is altered.

## Web

The two dead error codes left both dictionaries and `account_enabled_elsewhere` joined them, next to `another_channel_enabled` (its sibling in the same guard). The binding editor keeps its shape — this is copy plus doc, not a redesign:

- the connection switch carries the bind/unbind sentence as its own `title`, not as a line under it. Per the frontend convention, semantics are *disclosed* rather than parked on screen, and a permanent sentence there would push the whole form down for a fact read once;
- the "what binding does" fold gained the channel-neutral half: several conversations may keep one bot saved, only one may have it enabled, and moving it needs no credential deleted.

## `bridge.ts` was not touched

Its module doc states the per-Session rule, which this change leaves intact, so nothing there went stale. A defensive guard in `start()` was considered and rejected: no API path can produce a database with two enabled bindings on one account (the guard-then-flip is synchronous, and one data root has one server), and if a hand-edited database ever did, the Telegram connector already reports the resulting conflict with an actionable message. That kept the diff out of a file two other branches are editing.

## Verification

- `pnpm -r build` — all 9 packages built.
- `pnpm typecheck` — clean across all 8 projects.
- `pnpm test` — 4115 passed, 5 skipped, 0 failed (core 1013, server 1074, web 1361, cli 374, desktop 144, landing 69, docs 53, skills 27).

New/changed coverage: `messaging.test.ts` and `messaging-telegram.test.ts` now pin that two Sessions may save one account, that the second enable is refused with `account_enabled_elsewhere` while the first holds it (and that the refusal does not carry the holder's id), that a save on the dark Session leaves the live connection alone, and that after the first disables the second connects on its own credentials while the first keeps its saved config. `db-upgrade.test.ts` gained a database seeded with the old unique index: the open drops it, keeps the existing row, then accepts a second Session on the same account — and recreating the unique index over the two rows throws, which pins the one-way direction where the tolerance is defined.


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`: full server suite, the web messaging and locale tests, prettier, and a route-level probe to confirm the main finding by running it.

The model is the right one and the migration half is careful — `SCHEMA_SQL` runs before both drops, `DROP INDEX IF EXISTS` is idempotent and re-runnable, and the one-way downgrade is honestly documented and reproduces as described. One hole, now fixed, plus two smaller ones.

### 1. A `PUT` could move an enabled binding onto an account another Session held

Exclusivity moved from save time to enable time, but only the state route enforced it. Both PUTs write `accountId` unconditionally and then `if (saved.enabled) await deps.messaging.sync(...)` — so an already-enabled binding could be re-pointed at an account another Session was holding, and would come up live on it. Two Sessions enabled on one account: the one state the whole design asserts is unreachable, reached without ever passing the enable gate.

Confirmed by running it (two Sessions, one Project, owner cookie): A enables Feishu app `cli_app_AAAA`, B enables `cli_app_BBBB`, then `PUT /api/sessions/B/messaging/feishu {appId: "cli_app_AAAA", …}` returned **200** with both rows `enabled = 1` and two live connections. Both persist, so `bridge.start()` reconnects both on every subsequent boot. On Telegram that's two `getUpdates` pollers on one token — the Bot API 409 the 0.2.7 entry already documents. On Feishu the inbound dedupe is keyed per *binding*, so one chat message becomes a Task on both Sessions and both reply into the same chat. The path is mundane: the App ID / Bot Token inputs and Save stay live while the switch is on — only the clear-secret checkbox is gated.

**Fixed** in 57254e57: the account half of `guardEnable` is now its own `guardAccountFree`, asked by the enable *and* by a save that would leave the row enabled. A disabled binding stays exempt — that is the point of the model — and its own enable is still gated. Route tests added for both channels, each mutation-verified (removing the guard turns the new test red).

This also restores the PR body's stated reason for leaving `bridge.ts` alone. It was false as written: no API path can now produce a database with two enabled bindings on one account.

### 2. An orphaned enabled binding turned the 409 into a dead end

`destroyProject` and `deleteAgent` remove `sessions` rows without touching `messaging_bindings`; only `bridge.start()` cleans those up, at the next server start. Delete the Project holding an enabled Feishu app and the row stays `enabled = 1` — another Session's enable then answers "turn it off there first", where "there" no longer exists and, by deliberate policy, is not named. No in-product recovery short of restarting the server. This predated the PR, but the enable guard is now the only exclusivity gate, so it is now the only dead end.

**Fixed** in the same commit: when the holder's Session no longer resolves, the guard drops the stale row and lets the enable proceed — the same reconciliation `start()` already performs. Test added.

### 3. The upgrade test never upgraded a database whose binding was enabled

`seedUniqueAccountIndexDb` seeded a disabled row and asserted `findEnabledByAccount(...)` is `null`. The realistic upgrade is a user who already had a live binding, and nothing pinned that such a row keeps `enabled = 1` across the open and still holds the account. **Fixed** — second case added.

### Docs and changelog

Both said the PUTs had lost their 409 outright, which the fix makes untrue. `server-api.{en,zh}.md` and the changelog pair now state the one case that remains, mirrored in both languages.

### Checked and clean

- Migration ordering, idempotency and the one-way downgrade claim — all reproduce.
- Both `_in_use` codes are gone everywhere in `packages/`; only the frozen 0.2.6 entry and the new entries' past-tense prose retain them.
- Routes / `api/types.ts` / `endpoints.ts` / docs agree, and the en and zh API docs mirror each other.
- `Strings = typeof zh` makes the two dictionaries compiler-enforced; all three new keys are in both. `faqWhatBinding` reads correctly under the fixed behaviour.
- No new Chinese outside `*.zh.md`, `strings.ts` and the mandated `[中文版]` links.
- Both changelog pairs match `changelog/README.md` — metadata order, counterpart link, no banned sections, `Breaking` present with `## Compatibility`, sections mirrored. The "0.2.5 and later" claim matches the root CHANGELOG's account of the 0.2.5 npm-only tag.
- Server suite 1078/1078, typecheck clean, prettier clean.

### Merge

- The messaging spec still describes the old binding model in eight places (`05-ARCHITECTURE.md:1277,1278,1282,1345,1346,1347`; `06-PROTOTYPE.md:68,71`) — 跨 Session 唯一性按账号, the `(channel, account_id)` unique index, the two retired `_in_use` codes, 解绑 as the DELETE alone, and 没有整体解绑按钮. Those live in the still-open design PR #61, and per `CLAUDE.md` the spec change lands with the behaviour change. I am preparing that separately.
- Merge order: this before #494 (stacked on it) and before #493 — both write the same `2026-08-26-backward-compatibility` pair and the same `ensureColumn` line, so whichever lands second rebases. `490 → 494 → 493` costs one rebase; putting #493 first costs two.

hiyouga:

PR #491, opened 2026-08-27
fix(web): dismiss an anchored menu only for scrolls that moved it

hiyouga:
While a conversation is streaming, the Session list's row context menu "keeps getting wiped and won't open".

## What was happening

`Dropdown` registers its scroll listener on `window` in the **capture phase**, which it has to: scroll events do not bubble, so a bubble-phase window listener would never hear a nested scroller. The cost is that capture on `window` hears *every* scrolling element in the document, including ones the panel is nowhere near.

For a pointer-anchored panel — `anchorRect` set, which is exactly what a context menu is — the handler's response was to dismiss, unconditionally:

```js
if (anchorRect) setOpenRef.current(false);
else place();
```

While the model streams, the message list auto-follows the output (`stream-follow.ts` writes `el.scrollTop` on every chunk), and each of those writes fires a scroll event on the chat pane. Every one of them reached this listener and closed a menu anchored in the **sidebar** — a part of the page that had not moved at all. Hence "wiped instantly, and unopenable for as long as output is streaming".

## The fix

The dismiss rule is kept, not removed. A pointer anchor is a position rather than an element: once the content under that position moves there is nothing left for the panel to follow, so dismissing is the correct behaviour and the usual context-menu one. The defect was that it fired for scrolls that moved nothing.

The rule is now conditional on the scroll actually moving the anchored content:

- `Dropdown` takes a new optional `anchorOwner?: () => HTMLElement | null` — the element the anchor was measured from. An accessor rather than a raw node, matching `returnFocus` next to it, and read through a latest-callback ref so a caller passing a fresh arrow each render does not tear down and re-register the listeners.
- `scrollMovesAnchor(target, owner)` is the whole decision, exported so it can be unit-tested: dismiss when the scrolled container contains the owner. A page-level scroll targets `document`, which contains every node in the page, so full-page scrolling still dismisses with no case of its own; scrolling the sidebar's own Session list still dismisses too.
- `useRowContextMenu` exposes the row it already holds in a ref as `anchorOwner`; `sidebar.tsx` — the only `anchorRect` caller — passes it.
- A caller that supplies no owner keeps today's behaviour (any scroll dismisses), so a future `anchorRect` caller that cannot name an element is not silently changed.
- The `resize` branch still dismisses unconditionally: a resize really does move every anchor point on the page.

Nothing else the panel installs is affected. The outside-click listener responds to `mousedown`, which a programmatic scroll does not produce, and an outside click anywhere genuinely is a dismiss; the Escape layer is keyboard-only. `usePortalPanel` (OptionMenu, InfoPopover, Select) has the same capture-phase reach and also closes on any scroll — but it anchors to a real trigger element and repositions nothing, and none of its call sites sit next to a continuously scrolling container, so it is left alone rather than changed speculatively.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (web 102 files / 1366 tests; full workspace green)

The regression is pinned in `packages/web/test/context-menu.test.ts`: a scroll in an unrelated container leaves the menu open, while a scroll in the container holding the anchored row (and the row itself, and the document) still dismisses. Both new assertions were confirmed to fail against the pre-fix behaviour — the containment rule short-circuited to "always dismiss", and the sidebar's `anchorOwner` wiring removed — and to pass with the fix in place. Playwright e2e was not run: no selectors or flows moved.


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`, with the package's tests and a mutation pass over the fix itself.

The containment rule is the right shape for this bug. It avoids the fragile alternatives — no rect diffing, so no sub-pixel comparison, no sensitivity to zoom or a `transform` ancestor — and it fails toward dismissing on every ambiguous input (null owner, null target, a target with no `contains`), which is the safe direction. Tracing the layout confirms the reported case is genuinely fixed: `<aside>` and `<main>` are siblings, the Session-list scroller is a real ancestor of the row, and the document itself never scrolls.

Three things were worth changing, all pushed in f8c5294d.

### 1. The line that fixes the bug had no test behind it

The new cases exercise `scrollMovesAnchor` against stand-in nodes and grep `sidebar.tsx` for the prop, but nothing asserted that `dropdown.tsx` *consults* the predicate. Verified by mutation: replacing the new branch with the pre-fix `if (anchorRect) { setOpenRef.current(false); return; }` left all 31 tests green. The node-only suite (`environment: "node"`, no jsdom) can't reach the component, so nothing else would have noticed — a later refactor of `onScroll` could drop the guard and bring the streaming-wipes-the-menu bug back with CI green.

The file already runs a source scan for exactly this class of unreachable wiring, so the assertion went there. Re-mutated afterwards to confirm it now fails.

### 2. `scrollMovesAnchor` moved to `lib/context-menu.ts`

It's pure DOM logic — no React, no hooks, only `Node.contains`. `lib/context-menu.ts` says in its own header that the deciding parts live there "so they can be tested in this package's node-only vitest environment", and `vitest.config.ts` says the same. Leaving it in `dropdown.tsx` made `context-menu.test.ts` pull React and `react-dom`'s `createPortal` to reach one predicate, and made #495 point a positioning hook at a component module to borrow it. No cycle existed, but the dependency ran the wrong way.

Doing it here rather than as a follow-up keeps #495 a genuine two-line consumer. #495 has been rebased on this and imports from `lib/` now.

### 3. The `anchorOwnerRef` comment recorded a constraint the code doesn't have

It said the ref exists so the scroll listener "must not be torn down and re-registered because a caller passed a fresh arrow". That effect's deps are `[open, portal, place, anchorRect]`; the only anchored caller passes `portal={{ ... }}` inline, and `place`'s identity changes every render because `children` sits in its `useCallback` deps. The listeners already re-register on every `SessionRow` render. Reworded to what the ref actually buys: the owner is read at event time, so the accessor's identity never reaches the dependency list.

### Not changed

`Dropdown`'s element-anchored branch still calls `place()` for every scroll in the document — a synchronous layout read plus a re-render per streamed chunk while any portal-mode dropdown is open, for a position that is numerically identical. That's pre-existing behaviour from before this PR and it changes the re-place path rather than the dismiss path, so it's left alone deliberately rather than widened into this fix. One guard before the `if (anchorRect)` branch would cover both.

### Checks

- Design: no companion PR needed. Grepping `design/` for the row context menu returns nothing — it was never specced, so its dismissal nuance sits below the spec's altitude. (That the feature is absent from `06-PROTOTYPE.md` is a pre-existing gap from the PR that introduced it.)
- Changelog: compliant. Bilingual pair in `unreleased/`, metadata in Date/Type/Scope/PR order under the H1, counterpart link on the next line, `## Details` / `## 细节` mirroring 1:1 at five bullets, no banned sections, `Issue` correctly omitted. Every claim matches the diff.
- No new strings, so the two-dictionary contract is untouched; no new Chinese outside the `.zh.md`.
- `context-menu.test.ts` 32/32 pass; prettier clean; `tsc --noEmit` on `packages/web` reports nothing in the changed files.

### Merge

Merge this before #495 — that branch imports `scrollMovesAnchor` and its changelog relative-links this entry's files.

PR #492, opened 2026-08-27
fix(web): the Trace panel refreshes when a turn ends

hiyouga:
The Trace ("轨迹") dock panel fetched only when its tab became visible. A Trace opened while the Session kept working therefore stayed at the size and the contents it happened to be read at — the pills still showed the old byte counts, and the timeline below them was whatever the last round had written minutes earlier. The Files panel already solves this: the chat page bumps a counter on the settled-turn edge and the browser re-lists and re-reads the open file. This hooks the Trace panel onto that same counter.

## What now triggers a refresh

- **A turn settling while the Trace tab is showing.** The chat page's counter — bumped on the `running` → `idle` edge of the selected Session, guarded against the phantom `idle` a detaching stream emits and against a session switch — is now handed to `TracePanel` beside `WorkspaceBrowser`. It was renamed `filesReloadSignal` → `settledTurnSignal` for its second reader, doc comment included.
- **The file listing and the file on screen.** The panel's `listTick` doubles as `TraceFileView`'s `reloadSignal`, so every edge that re-lists also re-reads the selected file. That matters more than the re-list: a Trace file is *appended to* while the Session runs, so the ordinary refresh is the same `index` at a larger size, which a load keyed on `index` alone never re-runs for.
- **The re-show edge, unchanged.** It is also what makes dropping a hidden signal safe.

## What deliberately does not

- **A hidden tab fetches nothing.** The dock keeps inactive tabs mounted; firing a request per settled turn for a tab nobody is looking at is waste. The signal is consumed rather than queued, and the hidden → visible edge re-fetches once, however many turns went by.
- **A refresh does not disturb what the user is reading.** The selected pill survives a re-list while its file is still listed (selection is held by index, and a vanished pick falls back to the newest file); the "+N" expansion and the timeline highlight live above the fetch and are untouched. Inside `TraceFileView`, clearing the view — skeleton, collapsed round cards, pinned row — moved out of the load effect into a render-phase reset keyed on the file identity (React's documented "adjust state when a prop changes" pattern, as `workspace-browser.tsx` already uses for a session switch), so it happens only when a *different* file is selected. A refresh of the file on screen overwrites in place; nothing is blanked up front, so the outcome of the request is what sets or clears the error.
- **That holds for the default selection too, not just a clicked one.** `fileIndex` starts `null`, which `activeTraceFile` resolves to the *newest* file — re-resolved on every re-list. Left that way, a reader who never clicked a pill would be walked onto a compaction shard the instant one appeared: `index` changes, the render-phase reset fires, and analysis, events, collapsed rounds, pinned row and scroll go with it. The listing's `.then` now pins the default once — `setFileIndex((cur) => cur ?? sorted[0]?.index ?? null)` — so after the first listing only a pill click moves the selection.
- **A failed refresh takes nothing down.** The listing's `.catch` no longer empties `files`: it keeps the listing already on screen and reports the failure as a `toneStrip.danger` strip above the pills, so a blip mid-read cannot unmount the pills, the header, the download link and the whole file view under them (there is no manual refresh control, so recovery would have meant waiting for the next settled turn or switching dock tabs). The full `EmptyState` path is reachable only where there is nothing to keep — no listing yet, or a listing with no file in it. Same shape one level down: `TraceFileView` falls through to its error line only when `analysis === null`, and otherwise renders the line above the summary it kept. This follows `workspace-browser.tsx:329-331`, which already scopes a failed re-read the same way.
- **Pricing is no longer refetched per turn.** `api.getModels(projectId)` rode in the same `Promise.all` as the analysis and events fetches, so a project-wide catalog no turn produces was refetched on every settled turn — and the `setModels((prev) => m ?? prev)` guard existed only to stop that refetch dropping the cost column when it failed. It moved to its own effect keyed on `[projectId]`; `setModels(m)` needs no fallback, and a refresh went from three requests to two.

## The StrictMode split

`trace-panel.tsx` kept its two-effect shape: an edge-detecting effect that bumps `listTick`, and a fetch effect keyed on that tick. Collapsing them would lose the FIRST load under StrictMode's dev-mode double invoke — the edge tracker advances on the first run, whose fetch the cleanup cancels, and the second run sees no edge and never refetches, leaving the skeleton up forever. The new trigger goes through the same tracker rather than into the fetch effect: only the *decision* moved out, into `trace-refresh.ts`, seeded from the panel's first render so a mount is never itself an edge. `TraceFileView`'s load effect is not edge-detecting (it fetches whenever its deps change, mount included), so adding `reloadSignal` to its dependency list is safe there.

## Tests

`packages/web/test/trace-refresh.test.ts` — the suite runs in a node environment and renders no React, so the rule is unit-tested where it is separable from the effect applying it (same approach as `panel-task-scope.test.ts`), with the wiring pinned against the components' source:

- a settled turn while showing re-fetches, once per bump; renders alone do not;
- a bump while hidden fetches nothing, and three of them collapse into the single fetch the re-show does;
- both edges in one observation are one fetch, and the panel's own first render is not an edge;
- the selected file survives a re-list that returns it grown or adds a newer one, and a vanished selection falls back to the newest;
- the same re-list held as `null` *does* move — the characterisation of the yank the pin exists to prevent;
- the panel pins the default from the first listing and writes `fileIndex` nowhere else but the pill click;
- a failed re-list keeps `files` and reports itself as a strip, the whole-panel failure state is reachable only from the two nothing-to-keep guards, and `TraceFileView` gives its view over to the error only when `analysis === null`;
- `getModels` is out of the file's load effect and on an effect whose deps are exactly `[projectId]`;
- the counter reaches `TracePanel` (asserted to be the same identifier `WorkspaceBrowser` gets) and `TraceFileView`, whose load effect depends on it and clears no view state.

Each wiring test was mutation-checked: reverting its source change individually turns exactly that test red.

## Verification

From the worktree root, Node v24.18.0:

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (web 103 files / 1372 tests, whole repo green)
- `pnpm format` + `pnpm format:check` — pass ("All matched files use Prettier code style!")

Re-verified for the review-fix commit: `pnpm --config.verify-deps-before-run=false -r build` — pass; `packages/web` `npx tsc --noEmit -p tsconfig.json` — clean; `npx vitest run` in `packages/web` — 103 files / 1377 tests pass (`trace-refresh.test.ts` 17); `npx prettier --check` on the six changed files — pass.

Playwright e2e (`pnpm test:e2e`) was not run: it is outside the CI-parity chain above, its runner binds fixed ports 8930/8931, and this box is running several worktree agents in parallel. No e2e spec asserts Trace fetch timing; `dock.spec.mjs` only opens and moves the tab, and `compaction.spec.mjs` reads the panel after a finished run, which this change does not alter.


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`, with the package's tests and mutation passes over each fix.

The refresh/dedupe logic itself is correct and I could not break it: the tracker's edge rules hold (a bump only ever fires while `active`, both edges in one observation collapse to one fetch, a bump while hidden is consumed not queued), the two-effect StrictMode split is preserved, and every fetch has a `cancelled` cleanup, so an overlapping turn or an unmount mid-flight cannot write stale state. The gaps were at the edges the PR's own prose claimed to have covered.

### 1. The "a refresh doesn't disturb what you're reading" guarantee only held after the user clicked a pill

`fileIndex` initialised to `null` and was written only by the pill `onClick`, so for a user who never clicked one `activeTraceFile(files, null)` resolved to `files[0]` — the newest file, re-resolved on every re-list. Since the listing now re-runs on every settled turn while the tab shows, the panel silently followed a newly rotated shard mid-read.

Concretely: a long session compacts during turn N; `ContextEngine.write` defers the rotation, so `<session>_002.jsonl` appears with turn N+1's first write. When turn N+1 settles the panel re-lists, `activeFile` moves from #001 to #002, `TraceFileView`'s `index` prop changes, and the render-phase reset blanks `analysis`/`events`/`collapsed`/`pinnedRow` and drops the scroll position. The reader is looking at the near-empty new shard having touched nothing. Before this PR the same jump was only possible when they re-showed the tab. The test that claimed to cover this only exercised `selected = 1`; the `null` case one test below asserted the yank as correct without connecting it to the panel's default state.

**Fixed** in 5210370a — the listing `.then` pins the default (`setFileIndex((cur) => cur ?? sorted[0]?.index ?? null)`), so after the first listing only a pill click moves the selection. Mutation-verified.

### 2. A failed settled-turn re-list destroyed the whole panel

The listing's `.catch` did `setFiles([])` + `setError(...)`, and the render returned the load-failed `EmptyState` before anything else — pills, header, download link and `TraceFileView` all unmounted, taking `collapsed`, `pinnedRow`, the loaded analysis and the scroll with them. With no manual refresh control, recovery meant waiting for the next settled turn or switching dock tabs. A dev-server restart, a proxy 502, or the trace being removed while the panel is open was enough. Previously that could only happen on a re-show — user-initiated, nothing on screen to lose; now it fired automatically on every turn end while someone reads. `TraceFileView` had the same shape one level down.

**Fixed** — the catch keeps the previous listing and renders the error as a `toneStrip.danger` strip above the pill row; the full `EmptyState` path is reached only when there is nothing to keep. `TraceFileView` falls through to its error line only when `analysis === null`. This follows `workspace-browser.tsx`, which already scopes its listing error the same way. Three separate mutations verified.

### 3. The project-wide model catalog was refetched on every settled turn

`api.getModels(projectId)` rode in the same `Promise.all` as the analysis and events fetches, so it re-ran on every bump even though pricing is a project-level catalog no turn produces — and the `setModels((prev) => m ?? prev)` guard existed only to stop that refetch dropping the cost column when it failed. **Fixed** — its own effect keyed on `[projectId]`; the refresh is two requests instead of three, and the fallback is gone.

### Deferred

`analyze()` and the events endpoint each read and JSONL-parse the entire shard (`trace-service.ts:817,833`), so a settled-turn refresh costs two full parses of a growing file, per viewer, per turn. Past 1000 events the fixed `offset 0, limit 1000` also can't surface anything new. The listing already returns `sizeBytes` and `mtime`, so skipping the re-read when the selected file's size is unchanged would make the common "turn wrote to a *different* shard" case free — worth a follow-up, not this PR.

### Checks

- Design: no companion PR needed. `06-PROTOTYPE.md` documents panel *features*, not refresh triggers, and the sibling change that introduced this pattern (#460) landed with no spec entry. (Separately and pre-existing: the spec's Trace section still describes the standalone browsing page `trace-panel.tsx` says was removed — drift this PR neither causes nor worsens.)
- Changelog: compliant, and the selection bullet plus two new mirrored bullets now match the landed behaviour. PR body updated.
- No new strings, no new colours; the new strip uses the semantic tone token.
- `trace-refresh.test.ts` 17/17 (was 12), full web suite 1377/1377, `tsc --noEmit` fully clean, prettier clean.

PR #493, opened 2026-08-27
fix(server): a redelivered chat message survives a server restart without running twice

hiyouga:
Reported on the desktop app, driving a Session from Feishu: *"right after switching the connection on, an empty follow-up message still occasionally shows up, and then two are sent in a row."* This PR fixes the half of that report that still reproduces on 0.2.7. The diagnosis of the other half is below, because "still" was the load-bearing word and the answer to it is *no longer, since 0.2.7*.

## What reproduces on 0.2.7, and what does not

**The empty follow-up does NOT reproduce.** #485 fixed it and this PR adds nothing to it. Driven end to end through the real bridge and the Feishu fake into a busy Session, the queued line carries its content:

```
pendingFollowUps: [{"id":"af1b…","text":"second from feishu","images":0,"files":0}]
```

`MessagingBridge.onInbound` → `startTask(…, {queueIfBusy:true})` → `recallStoreOf` → `followUpInfo` → `task_state`. `userText()` builds a `model_msg`/`text`/`role:"user"`, which `isPlainText("user")` matches, so the text survives. The SSE subscribe snapshot reads the same `pendingFollowUpsOf` the live event does, so a reload cannot draw an empty line either — already pinned by `followup.test.ts` and `packages/web/e2e/reloaded-followup.spec.mjs`. On **0.2.6**, `QueuedFollowUp.recall` was optional and `followUpInfo` read `f.recall?.text ?? ""`, and the bridge supplies no recall store — so *every* bridge-queued follow-up rendered with nothing after its colon. That is the report, and it is a 0.2.6 report.

The two symptoms are one event on 0.2.6: a Feishu redelivery queued the message twice (no dedupe existed yet) and *both* copies rendered empty. "An empty follow-up shows up, and then two are sent in a row" is one redelivery, described from both ends.

**The duplicate DOES reproduce**, in exactly the shape #484 named as deliberately open. Rebuilding the bridge over the same database — what `hmr/platform.ts` does across a generation and what a desktop-app relaunch does — and replaying the event:

```
× remembers across a SERVER restart: the binding row's watermark survives the process
  → expected [ [ { type: 'text', …(3) } ], …(1) ] to have a length of 1 but got 2
```

That matches the report's timing precisely: quit with an event un-acked, relaunch, **switch the connection on**, and the channel pushes what it never saw acknowledged.

Everything else in that family was probed and holds on 0.2.7, so nothing here touches it: a concurrent redelivery of one id collapses (`isRedelivery` records synchronously before any `await`); two racing `sync()` calls on a **Feishu** binding leave one open connection and start one Task, so #484's late-close pin is not Telegram-specific; a connector restart inside one process was already covered.

## The cost: one extra `UPDATE` per inbound message, deliberately

#484 declined to persist this because it "would put a write on every inbound message". It does, and that is the right price.

An earlier revision of this PR avoided it by riding the statement `onInbound` already issues — `repo.recordChat(...)` runs unconditionally for every inbound message, ahead of the text check — so the watermark cost one more bound parameter and nothing else. Review caught that as a blocker, correctly. `recordChat` runs **before** `startTask` is even attempted, and it has to: the outbound relay reads `last_chat_id` off the row, and the run can emit its first assistant message before `onInbound` returns. A watermark on that statement is therefore committed before the work exists, and the follow-up a busy Session queues from it lives only in `SessionEntry.followUps`, in memory. A process dying in that window — measured at ~904ms below, and the Feishu ack is lost outright if the socket closes inside it — left a durable "already processed" mark over work that no longer existed, and `connect`'s seeding then turned the channel's replay into a complete no-op: the message never ran and nothing ever answered. On 0.2.7 the successor's empty ring would simply have run it. Trading a recoverable duplicate for a silent swallow, in the exact crash window this PR exists to cover, is not a saving.

So the write is split. `recordChat(sessionId, channel, chatId, isDirect)` stays where it was, keeps its four arguments and still goes first. `recordInboundWatermark(sessionId, channel, messageId)` is a second `UPDATE`, issued only once `startTask` — or the text-only reply for a non-text message — has resolved. One extra statement per inbound message buys at-least-once instead of a silent swallow, and a `startTask` that threw leaves the id unwritten so the replay runs the message.

## The change

`messaging_bindings.last_inbound_message_id` (TEXT, nullable), advanced by `recordInboundWatermark` after the message has been dealt with. `MessagingBridge.connect` folds the stored id back into the binding's in-memory ring before the stream can deliver anything, so the ring and the row are one memory rather than two guards: the ring holds the last 64 ids for the life of the process, the row holds the last one for the life of the binding. An account change clears it with the remembered chat — the other bot's ids mean nothing to the new one.

The statement is `last_inbound_message_id = COALESCE(?, last_inbound_message_id)`, so an inbound message with no channel identity — which also opts out of the in-memory ring at `isRedelivery` — leaves an existing watermark standing instead of nulling it and stripping the binding of its restart guard until the next id-bearing message. No shipped connector emits one; the code anticipates the case, so it should not handle it in the erasing direction.

The row holds **one** id, which is what Feishu's WebSocket resume needs and all it needs: that stream replays events it never saw acknowledged, and `WSClient` acknowledges each one only after our handler returns, so at most the message in flight when the process ended is still owed. **Telegram redelivers nothing across a restart** — `telegram-connector.ts` advances `offset = update.update_id + 1` *before* `await handlers.onMessage(msg)`, and a new connection drains with `getUpdates({offset: -1})`, discarding whatever was sent while nothing was connected. The row is written and seeded on that channel and simply never has a replay to catch. The fix is Feishu-only in practice, and the code comment and both changelog files now say so rather than generalising over "a connector".

**The ten-minute age bound on the in-memory ring is removed** (the count bound stays, and is what keeps the memory bounded). Its stated justification was a channel that reuses message ids. Neither channel does: Feishu's `om_*` are globally unique, and Telegram's key is `chatId:message_id`, whose counter only climbs. With no reuse to guard against, expiry could only forget a redelivery still owed — and a channel resuming a stream after a long outage replays what it never saw acknowledged, however much later that is. `Map<string, number>` becomes a `Set<string>`.

## Acknowledgement timing: deliberately unchanged

The other candidate was acking Feishu promptly instead of awaiting `startTask` inside the event handler. The mechanism is real — `WSClient.handleEventData` in `@larksuiteoapi/node-sdk@1.73.0` awaits `eventDispatcher.invoke` and only then sends the response frame, and `sendMessage` silently no-ops unless `readyState === OPEN`, so a socket that closes during the handler drops the ack outright. Measured, first message after enabling, with a cold Session load:

```
handler (= the Feishu ack) held for 904ms; loader cost 900ms
```

The ack latency *is* the task-start latency, unbounded. It is still not the right change on its own: acking early does not shrink that window, it moves it from "a replay we can recognize" to "a message lost outright if the process dies before `startTask` runs". Once a replay is harmless — which is what this PR makes it — acking early buys nothing and opens a new failure mode. Left as it is.

## Contested files touched

- `packages/server/src/db/schema.ts`, `database.ts`, `db/repos/messaging-bindings.ts` — one column, one `ensureColumn` line, one new `recordInboundWatermark` method beside an unchanged `recordChat`, and their `mapRow`/`upsert` companions. Nothing else in those files is read or reshaped.
- `changelog/unreleased/2026-08-26-backward-compatibility.{md,zh.md}` — started from `feat/messaging-bind-by-enable`'s exact content and added one section plus a lead-sentence count and one Compatibility sentence, so the resolution against that branch is additive. The `PR:` field needs the second link once both are known.
- `packages/server/src/runtime/messaging/bridge.ts` — inbound path only; the outbound half is untouched.

## Left undone on purpose

- **Feishu group messages hand the model the raw mention placeholder.** `feishu-connector.ts`'s `textOfContent` returns the content JSON's `text` verbatim, and Feishu's group payload is `{"text":"@_user_1 …"}` with the real name only in a `mentions` array the connector never carries. Every group message therefore reaches the model prefixed with `@_user_1 `, and an @-only message passes the `text.trim() === ""` guard and starts a full run on a placeholder. Separate defect, separate PR.
- **A `startTask` failure still consumes the in-memory ring entry.** `isRedelivery` records the id synchronously, before the Task is attempted, so within one process a failed start is not retried — unchanged from 0.2.7. What this PR does fix is the durable half: the row is not advanced, so the channel's next replay (across a restart, where the ring is empty again) runs the message. A same-process retry is a different change.

## Verification

Node 24.18.0, Linux. First revision: full CI-parity chain from the worktree root — `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check`, `git diff --check`, all clean, with core 1013 passed / 5 skipped, web 1360, cli 374, desktop 144, landing 69, docs 53, skills 27.

Review revision (the diff is confined to `packages/server` and the changelog pair; nothing outside `packages/server` calls the repo API that changed): `npx prettier --check` on all six changed files, clean; `npx tsc --noEmit -p tsconfig.json` in `packages/server`, clean; `npx vitest run test/messaging.test.ts test/messaging-telegram.test.ts test/db-upgrade.test.ts` → **72 passed**; full server suite `npx vitest run` → **1079 passed (92 files)**.

Every behavioural change was mutation-tested: the source reverted, the test confirmed red, the source restored.

| Change | Mutation | Result |
| --- | --- | --- |
| Watermark after the work | `recordInboundWatermark` moved back above the branch | red — row carries `om_start_threw`; with the row assertions dropped, the successor's replay is swallowed and `waitFor` times out |
| `COALESCE` on the watermark write | back to `last_inbound_message_id = ?` | red — the id-less message nulls a good watermark |
| `upsert`'s `keepChat = false` branch | the new `CASE` made a no-op (arity kept) | red — the watermark survives an account change |
| Restart-test controls | `check()` returns true whenever the ring is non-empty (seeding poisons it wholesale) | red — both controls time out while the original assertions still pass |

Tests added or extended in `packages/server/test/messaging.test.ts`:
- "a message whose Task never started is left unwatermarked, so the replay runs it" — a bridge whose `startTask` rejects handles the event; the chat is recorded, the watermark is not; a successor over the same database replays and runs it.
- "a message with no channel id leaves the binding's watermark where it was".
- "remembers across a SERVER restart" — now fires a second, distinct event on the successor and asserts it runs, so a successor that never wired `onMessage` cannot pass.
- "a replayed non-text message does not re-send the text-only notice across a restart", with a fresh sticker as its control.
- "re-saving a binding onto a different bot account drops the watermark with the chat" — repo-level: same account keeps `lastChatId` / `lastChatIsDirect` / `lastInboundMessageId`, a different account resets all three. Cover for the `keepChat` branch and for the lifecycle claim the changelog makes.

And in `packages/server/test/db-upgrade.test.ts` — the column is ALTERed onto a `messaging_bindings` formed before it existed; the pre-existing binding keeps its credentials, intent and reply target, grandfathers in as `NULL`, and takes the new write.


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`: full server suite, a scratch harness driving a real restart across a real database, and mutation passes over each fix.

The mechanism is well chosen for the bug it targets — identity is exact set membership, not an ordered comparison, so none of the lexicographic-vs-numeric, out-of-order, or clock-skew failure modes apply, and the migration is correct and idempotent. The problem was ordering.

### 1. The watermark was committed before the work, so a crash in the window dropped the message

`recordChat` persisted `last_inbound_message_id` before `startTask` was even attempted, and the queued follow-up it produces lives only in `SessionEntry.followUps` — in memory. A process dying between those two points left a durable "already processed" mark over work that no longer existed, and `connect()`'s seeding then made the channel's replay a complete no-op.

Reproduced with a scratch harness: with a `startTask` that rejects, the row still carried the id, and a successor `MessagingBridge` over the same database processed the replay **zero** times. The window is real — the un-acked Feishu handler window is ~904ms, and the ack is lost outright if the socket closes during the handler.

So the PR traded "runs twice" — recoverable, and what the ring already did — for "never runs", silent and unrecoverable, in exactly the crash window it exists to address. A single concurrent burst could produce both halves at once. The PR body's claim that the gap was unchanged from 0.2.7 was false at precisely the restart boundary, and its own argument against early-acking ("a message lost outright if the process dies before `startTask` runs") describes the failure the change introduced by another route.

**Fixed** in c10424dc — the write is split. `recordChat` stays first, because the outbound relay reads `last_chat_id` before the run can emit; a new `recordInboundWatermark` is issued after `startTask` / `replyInbound` resolves, so a throw skips it. That is one extra `UPDATE` per inbound message, and it buys at-least-once instead of a silent swallow. Mutation-verified in both directions: moving the write back above the branch turns the new test red, and with the row assertions dropped the successor's replay is swallowed exactly as described. The PR body's cost argument was rewritten to state the trade plainly.

### 2. `keepChat = false` was new SQL with no coverage

A third `CASE WHEN ?` and a third bound parameter were added. The order was correct, but nothing in the suite re-saved a binding onto a different `accountId` at any level, so a transposition would have shipped silently — and this is the one lifecycle claim the changelog makes about the watermark. **Fixed** — repo-level test covering both directions, mutation-verified.

### 3. The restart test had no liveness control

It asserted `runs` stays at 1 after the successor replays. A successor that connected but never wired `onMessage`, or one whose seeding poisoned the ring wholesale, passed identically — nothing anywhere asserted that a *new* message still gets through after the watermark is seeded. **Fixed** — a second distinct event asserts `runs` reaches 2, plus a non-text control. Mutation-verified: poisoning the ring turns both controls red while the original "still 1" assertions stay green, which is the blind spot exactly.

### 4. The comment's justification was false for Telegram

It reasoned "a connector acknowledges each event as it finishes with it". Telegram does the opposite: `offset = update.update_id + 1` advances *before* `await handlers.onMessage(msg)`, and a new connection drains with `getUpdates({offset: -1})`, discarding everything sent while no connection existed. A Telegram message in flight at process death is never redelivered, so the persisted watermark is inert there — the fix is Feishu-only in practice. **Fixed** in the comment and in both changelog files.

### 5. An id-less message erased an existing watermark

`msg.messageId === "" ? null : msg.messageId` wrote NULL over a good watermark, and such a message also opts out of the ring, so the binding lost its restart guard entirely until the next id-bearing message. Latent — no shipped connector emits `""` — but the code anticipated the case and handled it in the erasing direction. **Fixed** with `COALESCE(?, last_inbound_message_id)`. (`node:sqlite` cannot bind `undefined`, so the parameter stays `string | null`; the semantics are as intended.)

### Still open before merge

- **Design.** `05-ARCHITECTURE.md:1277` enumerates the `messaging_bindings` columns exhaustively and needs `last_inbound_message_id`; `:1280` is the inbound contract and describes no dedupe at all, while Schedule's structural sibling at `:1268` has explicit 重启不重复触发 text. `08-CHANGELOG.md:9` also records that on this table `enabled` is intent and 连接实况仅存内存 — this is the first runtime-state column on it. Being prepared as a PR onto design #61's branch.
- **Merge order.** `2026-08-26-backward-compatibility.{md,zh.md}` is ADDED by this PR and by #490, and both add an `ensureColumn` line at the same spot in `database.ts`. This branch's copy describes the index drop and carries a `Breaking: yes` that belong to #490, and relative-links a file only #490 adds. Land #490 → #494 → this, then rebase so the entry becomes a real edit adding only its own section and appending `#493` to `PR:`. I have the resolution staged.

### Checks

- Server suite 1079/1079, typecheck clean, prettier clean.
- The `messaging-inbound-watermark` pair is compliant — metadata order, counterpart link, three bespoke sections mirrored, no banned headings, no new Chinese outside the `.zh.md`.
- Known limit, unchanged and out of scope: `isRedelivery` rings the id synchronously before the Task is attempted, so a failed start is still not retried *within* one process. Only the durable half is fixed.

hiyouga:

hiyouga:
**Rebased and merged.** `main` took #490 and #494 first, so this needed the rebase I flagged above. Four files conflicted; all four resolutions were additive:

- `db/database.ts` — both `ensureColumn` lines kept (`line_per_message` from #494, `last_inbound_message_id` from this PR). No version scheme, so no collision: `ensureColumn` is idempotent.
- `test/messaging.test.ts` — #494's `boot()` helper and this PR's `restartBridge()` helper both kept.
- `changelog/unreleased/2026-08-26-backward-compatibility.{md,zh.md}` — merged to the union: `PR:` now lists #490, #493 and #494, the lead covers three pieces of existing state rather than two, and the six sections (retired index / inbound watermark column / one-way half / how long the tolerance stays / `line_per_message` column / compatibility) mirror one-for-one in both languages.

One real fixup the rebase surfaced: the new repo test asserted `upsert(...).ok`, which #490 removed when it took the failure mode off `upsert` — it returns the row now. Retargeted onto `accountId`.

Server suite **1133/1133**, typecheck clean, prettier clean after the rebase. CI hit the known macOS `terminal-stream.test.ts` PTY flake twice (a file this PR does not touch); the third run is green.

PR #494, opened 2026-08-27
feat(server,web): deliver a messaging reply one message per line

hiyouga:
> **Stacked PR.** Based on `feat/messaging-bind-by-enable` (#490), which touches the same files. **Retarget this to `main` once #490 merges** — the diff shown here is this change alone.

A relayed assistant reply reached the chat as one message, however it was written. This adds a per-binding option that delivers it as **one message per non-blank line** instead — the shape a role-play answer written as several spoken lines is meant to arrive in. Off by default, and off is byte-for-byte the previous send sequence.

## Storage: a column, not the config document

`messaging_bindings.line_per_message INTEGER NOT NULL DEFAULT 0`, beside `enabled` and `last_chat_id`. `config_json` is the channel's own credential shape, owned by its connector; this is a channel-agnostic delivery preference and means the same thing on every channel. It is ALTERed onto existing databases by the `ensureColumn` list in `openDatabase`, which exists for exactly this.

**The default reproduces today's behaviour exactly**, so an existing `web.db` needs nothing from the user. On whether that deserves a line in the batch compatibility entry: **it got one** — a short "purely additive, no decision needed, nobody does anything" section. Not because it is a risk, but because that file is the one place a reader looks to answer "does my database need anything?", the batch already had the file open for the index drop, and three sentences saying "no" cost less than the reader's uncertainty. An older build simply ignores a column it has never heard of.

## API

`linePerMessage: boolean` on `FeishuBindingInfo` / `TelegramBindingInfo`, and optional on both PUT bodies where omitted keeps the stored value and a new binding starts `false`. No new route: it is an ordinary form field applied on Save, like the rest. The web form always sends it — an omitted flag means "keep", which would leave no way to turn the option back off.

## Bridge

`deliverReply` builds a list of bodies first — `row.linePerMessage ? splitReplyLines(text) : [text]` — and the existing chunk/thread/send loop runs over each. With the flag off the list is one element and the send sequence is unchanged; the hunk is one line plus a nested `for`, kept deliberately small and separable because `../msg-media` is editing the same method's outbound path.

`splitReplyLines` is deliberately literal: every non-blank line becomes its own message, **fenced code blocks included**, blank lines dropped, each line trimmed. No "smart" grouping — the option's value is that its behaviour is predictable. Each body still goes through `chunkMessagingText`, so a single line over the channel cap is split rather than rejected.

**The cap is `MESSAGING_MAX_LINE_MESSAGES = 20`, and it is a rate limit, not a size limit.** Telegram's documented per-chat allowance is about one message a second, with a burst to a single group tolerated up to roughly 20 a minute before it answers 429; Feishu's IM send limit is counted per app and is far looser, so Telegram sets the number. 20 spends the tightest channel's burst allowance for one chat on one reply. Past it the remaining lines are **combined into one final message**, not dropped — silently losing the tail of a reply is the worst failure available here. The number and this reasoning are documented where the constant is defined.

Unchanged, and pinned by tests: the group-chat rule (the run's FIRST outbound message threads onto the inbound one, everything after is a plain send — many lines must not become many quote headers), and the scope (relayed assistant replies only, never the approval notice, the text-only notice or the test message).

## Web

A `Switch` closing the binding form, for both channels, saved by the existing Save action — so it rides `MessagingFormState`, `bindingsToForm`, `formToPut` and `formDirty`. `formDirty` is the one that would have failed silently: on Telegram the only dirty signals were the token and the clear box, so flipping the switch would have left Save disabled.

On the explanatory text: what the option does to a reply is **semantics**, so per the repo's semantic-versus-formatting convention it is disclosed rather than parked on screen — an `InfoPopover` anchored to the option's `FieldLabel`, which is the form the convention prescribes when a title is present (and which `test/disclosure-anchor.test.ts` enforces). A permanent line would be read once and then stepped over on every later visit.

## Verification

- `pnpm -r build` — all 9 packages built.
- `pnpm typecheck` — clean across all 8 projects.
- `pnpm test` — 4129 passed, 5 skipped, 0 failed (server 1086, web 1363, core 1013, cli 374, desktop 144, landing 69, docs 53, skills 27).

New coverage: `splitReplyLines` unit tests (blank-line dropping and trimming, the literal code-fence split, the cap, the combined remainder that loses nothing, and the default cap). Through the real bridge and a fake Feishu SDK: the flag off delivering a multi-line reply as one message with the original text; the flag on delivering three ordered messages (asserted as a whole array — three messages out of order is a different bug with the same members); a line over `MESSAGING_TEXT_CHUNK_CHARS` still chunked, with the pieces reassembling; the cap reached, with the tail combined and nothing lost; the notices arriving whole; and a group chat still threading exactly once. Route coverage on both channels for save / keep-on-omit / turn-off, and `db-upgrade.test.ts` for a `messaging_bindings` formed before the column — it arrives, existing rows read `false`, the rest of the row is untouched, and it is writable immediately.


hiyouga:
## Review

Reviewed in a dedicated worktree against `feat/messaging-bind-by-enable` (#490's branch), with the full server suite, the web messaging tests, and mutation passes over each fix.

The approach is right and the base is solid: flag off is byte-for-byte the old send sequence, the migration is idempotent and additive with a correct default, and there is no version scheme to collide with #490's `DROP INDEX`. What was missing was the failure story for a burst of 20 messages on a channel the PR itself argues will rate-limit it.

### 1. One failed send silently dropped the rest of the reply, and nothing paced the burst

The `try` wrapped the whole `for (body) { for (chunk) { await send } }` loop, so the first throw abandoned every remaining line. Meanwhile `sendMessage` has no 429 handling — the exponential backoff in `telegram-connector.ts` is the poll loop only — and `deliverReply` fired back-to-back with zero delay, against the very limit the new constant's own comment cites ("about one message a second… roughly 20 a minute before it starts answering 429").

Proven with a scratch test against the real bridge: a 6-line reply with the client throwing `Too Many Requests: retry after 5` on send 3 → exactly `["l1","l2"]` reached the chat, one `messaging_send_failed` landed in `error_records`, and the chat showed a reply that just stops mid-sentence with no marker. With the flag off the same reply is one message, so the feature created this failure mode and it grows with line count.

**Fixed** in 0c4aedf1 — bodies and chunks flatten into one list, each send gets its own try/catch (a refused message costs only itself), and a 1s pace sits between messages when `linePerMessage` is set. The delay is injectable through the deps so only the pacing test actually waits. Both halves mutation-verified. The recorder's dedup window means a dead binding still lands one error row, not twenty.

### 2. The documented cap did not bound the number of outbound messages

The comment opened "How many messages one reply may become" and the changelog said "at most 20 messages", but the cap counted *bodies* — and each body then went through `chunkMessagingText`, including the combined tail, which is by construction the longest body of all. Measured: a 200-line reply of ~120-char lines produced **25** messages; 60 lines of 500 chars also **25**. Ordinary long replies, not pathological input, exceeding the rate-limit ceiling the cap exists to protect by 25%.

**Fixed** by spending a real-message budget instead: both cases now land at exactly 20, byte-intact. The comment states what it actually bounds, including the one honest exception.

### 3. Per-line `trim()` destroyed indentation

The comment promised literal splitting "inside a fenced code block included", but each line was `trim()`ed — a content change, not a split. `splitReplyLines("```python\ndef f():\n    return 1\n```")` returned `["```python", "def f():", "return 1", "```"]`; the Python arriving in the chat does not run, and the same applies to nested lists and any indented block. Nothing anywhere said indentation was stripped. **Fixed** with `trimEnd()`, which drops trailing whitespace and blank lines as advertised and leaves leading indentation alone. Five tests go red on the mutation back to `trim()`.

### 4. The approval notice could land mid-burst

`deliverNotice` was fired with bare `void`, off the `sendChain` that exists precisely to keep outbound order. Before this PR `deliverReply` held the client for one send; it now holds it for up to 20 sequential round-trips, so the interleaving window grew by more than an order of magnitude — a 12-line reply followed by a tool needing approval puts the notice between line 3 and line 4. **Fixed** by putting it on the chain; the notice already never throws, so it cannot break it. Mutation reproduces the interleave at index 1 exactly.

### Not done

A 429 retry honouring `retry_after` would be the more complete fix, but it does not stay small: `sendMessage` goes through the shared `call<T>()`, which flattens the envelope into an `Error` string, so honouring it means carrying `parameters.retry_after` on `TelegramEnvelope`, scoping the retry to sends only (a loop inside `call` would also retry `getUpdates`/`getMe`), and threading an abort signal so a retry cannot outlive a closed connection. That is a transport-contract change and it interacts with the pacing just added. Raising it here rather than leaving a TODO in the code.

### Still open before merge

- **Design.** `05-ARCHITECTURE.md:1277` (the column list), `:1281` (outbound described as size-chunked only), `:1345-1346` (the two PUT bodies), and `06-PROTOTYPE.md:68,71` (the editor's field list) all enumerate exactly the surfaces this extends. `git grep line_per_message` across every branch of the design repo returns nothing. Being prepared as a PR onto design #61's branch.
- **Base ordering.** #490 must merge first, then retarget this to `main` and rebase. Both touch `bridge.ts`, `messaging.ts`, `messaging-bindings.ts` and the same `2026-08-26-backward-compatibility` pair, and #493 collides on the same pair plus the `ensureColumn` line — `490 → 494 → 493` costs one rebase, the other order costs two.

### Checks

- Server suite 1091/1091, the three messaging suites 84/84, typecheck clean, prettier clean.
- Web: typecheck clean, the two binding suites 24/24, placeholder parity 3/3. Both `linePerMessageHelp` dictionaries gained a clause about the pacing — a 20-line reply now lands over ~19s, which is user-visible and belongs in the help text.
- Pre-existing row default reads `false` and reproduces the old send sequence exactly; `db-upgrade.test.ts` seeds a realistic pre-column table and asserts read + write.
- Changelog pair compliant and updated for the split semantics, the message budget, the pacing and the failure behaviour; the shared backward-compatibility pair left untouched, since it is being resolved at rebase time.
- `server-api.{en,zh}.md` left alone — already correctly vague about the cap.

hiyouga:

PR #495, opened 2026-08-27
fix(web): close a portal panel only for scrolls that moved its trigger

hiyouga:
> **Stacks on #491.** Based on `fix/context-menu-scroll-dismiss`, because it reuses the `scrollMovesAnchor` rule that PR exports. **Retarget to `main` once #491 merges** — the diff below is this branch's own two commits' worth of change; review it after #491.

#491 fixed the Session row's context menu. The same capture-phase defect lives in `usePortalPanel`, the shared portal-panel hook, and its blast radius is wider.

## What was happening

`usePortalPanel` registers its scroll listener on `window` in the capture phase, which it has to — scroll does not bubble. Capture on `window` then hears **every** scrolling element in the document, and the handler closed the panel for all of them:

```js
const onScroll = (e) => {
  if (panelRef.current?.contains(e.target)) return;
  onCloseRef.current();
};
```

`stream-follow.ts` writes `el.scrollTop` on the message list on every streamed chunk. Each of those writes fires a scroll event that closed:

- **the chat header's context ring** (`features/chat/context-gauge.tsx`) — the panel a user opens *while a run streams* to watch the context fill. It closed on the next chunk, every time;
- **`Select`** and **`OptionMenu`** in the composer, the message list's immediate neighbour;
- **`InfoPopover`** anywhere on a page whose neighbour scrolls.

## The fix

One conditional in the hook. The panel now consults `scrollMovesAnchor` (from #491) with the trigger element the hook already holds in `triggerRef`, so a scroll closes it only when the scrolled container actually contains that trigger.

Three things are deliberately preserved:

- **The panel's own internal scroll still does not close it.** The existing `panelRef.contains` guard runs first and is untouched.
- **`resize` still closes unconditionally.** A resize moves every trigger on the page at once; the asymmetry with scroll is now stated in a comment so it does not read as an oversight.
- **A null trigger still closes on any scroll.** `scrollMovesAnchor` answers "close" for an owner it cannot see, so a consumer that attaches no ref degrades to today's behaviour rather than to a panel that scrolling can never dismiss. All four current consumers attach the hook's own `triggerRef` to an always-rendered `<button>`, and a test now pins that.

Stale comments in `use-portal-panel.ts`, `select.tsx`, `info-popover.tsx` and `context-gauge.tsx` — each of which specified "closes on … scroll …" — were rewritten.

## Two things left unmade, deliberately

**1. Re-placing instead of closing is probably the better long-term behaviour, and this PR does not do it.** `Dropdown` already re-places an element-anchored panel on scroll, and its own comment gives the reason: the trigger can live in a horizontally scrolling toolbar on phones, where tapping a partially visible control scrolls it into view and would otherwise dismiss the panel the same gesture just opened. That hazard applies verbatim to the context ring and the composer's `Select`/`OptionMenu`, which sit in exactly such a toolbar — so the two primitives differ for no reason a user would recognise. Converting would mean extracting `usePortalPanel`'s one-shot `useLayoutEffect` placement into a callback and changing observable behaviour across five surfaces, which is not what was reported. Recommended as a follow-up, not done here.

**2. The context panel's composition breakdown is still an open-time snapshot.** Its comment justified that partly with "the panel closes on any scroll anyway" — a rationale this PR removes, since the panel now survives a streaming run. The headline percentage and token figures stay live (they come from the ring's own props); what holds still is the segment breakdown, for as long as the panel is left open. The other half of the rationale stands — the endpoint re-reads a whole Trace shard, so polling it through a run is not free — so the behaviour is unchanged and the comment now says what is actually true. Making the breakdown live is a separate call.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass
- `pnpm test` — pass (web 103 files / 1371 tests; core 1013, server 1073, cli 374, desktop 144, landing 69, docs 53, skills 27)

`packages/web/test/portal-panel-dismiss.test.ts` is new. A behavioural test cannot see a missing prop from a node-only suite, so it pins the wiring against the source: the hook's `onScroll` answers `scrollMovesAnchor` with `triggerRef.current`, the own-scroll exemption survives, `onResize` stays free of any ownership test, and every consumer discovered by scanning for `usePortalPanel({` attaches `ref={triggerRef}` — so a fifth consumer is covered the day it is written, and the scan asserts it found at least the four that exist plus the reported one by name.

Both new assertions were confirmed to fail without the change: the `scrollMovesAnchor` assertion fails against the pre-change hook, and the consumer scan fails — naming `features/chat/context-gauge.tsx` — when that file's `ref={triggerRef}` is removed. The two guard assertions (own-scroll exemption, unconditional resize) pass on both sides by design; their job is to fail on a future edit.

Playwright e2e was not run: no selectors or flows moved.


hiyouga:
## Review

Reviewed in a dedicated worktree against `fix/context-menu-scroll-dismiss`, with the package's tests and a mutation pass.

The one-line rule change is correct and fixes the reported defect. `MessageStream`'s scroller is a sibling of the composer wrapper, so `stickToBottom`'s `scrollTop` writes now correctly leave the ring's panel open. Listener registration (capture on both add and remove), effect deps, teardown, and the own-panel and resize guards are all sound, and because the rule is pure containment with no numeric threshold, sub-pixel rects, zoom and `transform` ancestors are simply not in play.

Consistency with #491 is genuinely good — same helper, same target, same phase, same `Node.contains` comparison, same null-owner fallback, same unconditional `resize`. Three fixes pushed (branch rebased on the updated #491, so this is a force-push).

### 1. The test passed with the fix's polarity inverted

`expect(onScroll).toContain("scrollMovesAnchor(")` is a presence check. Dropping the `!` turns the rule into "close only when the trigger did *not* move" — strictly worse than the bug being fixed — and all five new tests stayed green. Deleting the call is the mutation the assertion caught; inverting it is the likelier future edit.

Now matched as `/if \(!scrollMovesAnchor\(/`. Verified both ways: green as written, red with the `!` removed.

### 2. The context ring is in the composer toolbar, not the chat header

`context-gauge.tsx:2` says "Context usage gauge in the composer toolbar" — in this PR's own diff — and the only render site is `chat-input.tsx`, inside the bottom toolbar's right group. The hook comment, the test's header comment and both changelog files all said "chat header". Corrected in all four places; the argument is unchanged, since the composer is a sibling of the message-list scroller either way.

### 3. The hook no longer reaches into a component module

`use-portal-panel.ts` — the leaf positioning hook — was importing `./dropdown`, which pulls `react-dom`'s `createPortal` and `modal`'s esc-layer state. `scrollMovesAnchor` moved to `lib/context-menu.ts` in #491, and this branch now imports it from there.

Also re-wrapped an edited comment line that had run to 120 columns; prettier doesn't reflow block-comment prose, so `format:check` passed it through.

### Not changed

`Dropdown`'s element-anchored `place()` path is still ungated — every scroll in the document forces a layout read and a re-render for a position that is numerically identical, once per streamed chunk while a portal-mode dropdown is open. That's the remaining divergence between the two primitives, but it's a re-place concern rather than a dismiss one and it predates #491, so it's deliberately left rather than folded in here.

### Checks

- Design: no companion PR needed. `06-PROTOTYPE.md:16` and `08-CHANGELOG.md:13` spec the portal panel only as "not clipped by ancestor overflow, Esc captured and propagation stopped" — scroll-dismissal granularity is below that resolution.
- Changelog: compliant. Metadata order, counterpart link on the next line, `## Details` / `## 细节` mirroring four bullets each, no banned sections.
- `context-gauge.tsx`'s +6/-3 is comment-only — nothing unrelated smuggled in.
- 37/37 tests pass across both dismissal suites; prettier clean on all five touched files.

### Merge

Merge #491 first, then retarget this to `main`. The import and the changelog's relative links both depend on that branch.

PR #496, opened 2026-08-27
feat(server): images into the chat, generated files back out

hiyouga:
Media in both directions for the messaging channels: a picture pasted into Feishu or Telegram now reaches the Agent, and a run's reply is followed into the chat by the files it named. Both directions run over the existing connector seam, so a third channel implements them the same way.

## Inbound images (channel → Agent)

An inbound image becomes exactly what the web composer submits: the caption (where the channel has one) as user text, then one `image_url` part per image carrying a base64 `data:` URL, passed through core's `imageUrlMessage`. A model without vision folds those into scratchpad files itself, as it does for a pasted image.

- **Feishu** — `im.message.receive_v1` with `message_type: "image"` carries `{"image_key": …}`; the bytes come from `im.v1.messageResource.get({ path: { message_id, file_key }, params: { type: "image" } })`. Checked against the installed `@larksuiteoapi/node-sdk@1.73.0`: that call resolves a stream wrapper (`getReadableStream`/`headers`), not a `{code,msg}` envelope, and the local `LarkClient` interface says so.
- **Telegram** — the largest variant of `message.photo` (compared on pixel count, `file_size` as the tie-break, since `width`/`height` are the required fields), fetched via `getFile` then the **file** endpoint. That URL embeds the bot token, so nothing echoes it: failures report the status and `fetchErrorText`'s cause only. `message.caption` becomes the message's text.

### The seam

`MessagingInboundMessage` gains `images?: readonly MessagingInboundImage[]`, where an image is a **handle**, not bytes:

```ts
export interface MessagingInboundImage {
  fetch(maxBytes: number): Promise<{ data: Buffer; mimeType: string }>;
}
```

Two reasons it is a handle. First, inbound dedupe drops a redelivery before anything else happens — a connector that downloaded eagerly would pay a full image transfer per replay (there is a test for exactly this). Second, the cap rides *into* the fetch, so an oversize transfer is refused at the byte that crosses it rather than buffered and measured afterwards; `collectUnderCap` counts the chunks actually received, so a lying `content-length` buys nothing. A caption needs no field of its own — a captioned image's caption **is** that message's `text`.

A caption is that message's text **only for an actual photo**. Telegram attaches captions to documents, video and audio too, and none of those is delivered, so falling back to the caption there would run the model on a description of a file it never received. A group photo's caption gets the same `@mention` stripping a text message does.

**Per image: `INLINE_IMAGE_MAX_BYTES` (20MB)**, the server's existing inline-image ceiling, reused rather than reinvented: these images land in the conversation and the Trace exactly as the composer's uploads do. It also happens to match Telegram's own 20MB bot-download limit.

**In aggregate: a rolling per-binding byte budget** (two full-size images per ten minutes). Per image the ceiling was always there; without a budget, anyone who can DM the bot could write hundreds of MB of base64 into one Session's Trace, which `services/attachment-limits.ts` says is read back whole — into a single JS string — on every history page and every resume. The composer reaches that ceiling only through an authenticated user; this path has no authentication at all. The window's remainder rides into the fetch exactly as the ceiling does.

An image over the ceiling, one past the budget and one the channel refuses answer with **three different** bilingual notices, and none of them starts a Task, caption included — a model asked "what is wrong with this?" about a picture it never received answers confidently about nothing. A refusal caused by the bot's own app permissions names the scopes and carries the channel's grant link. Refusals are recorded through `recordError`; a size or budget refusal states its limit and stays out of the error log. (These started as one notice covering both — see *Follow-up fixes* for why that was wrong.)

**Image-only input is accepted** as-is: `SessionManager.startTask` does not inspect input parts (the HTTP route validates, and it already permits an image-only `input`), and core's `imagesToScratchpadPaths` / `appendAttachmentLines` handle "no user text at all" explicitly — an image with no caption is already a complete steering message on that path. No filler text is sent.

`MESSAGING_TEXT_ONLY_NOTICE` now names what is supported. Stickers, voice, video and files still get it.

## Outbound files (Agent → channel)

The bridge aggregates the run's relayed assistant text per entry and, at the `task_state → idle` flip, extracts the paths that reply **mentions**, resolves them against the Session's Workspace, stats them, and sends the ones that exist **and that this run wrote** as a batch.

Two filters, both load-bearing. The mention picks the one output that was the point out of the dozen a run writes. The modification time — at or after the run's start, with a two-second grace for filesystems whose mtime granularity is coarse — is what keeps that from becoming a read primitive: the reply is steerable by whoever is in the chat, and on a mention-only rule a group member who asks "what's in `secrets.json`?" gets the file uploaded by the very sentence that declined to paste it. A file the run did not write is dropped **silently**; a reply mentioning the config it read is the ordinary case, and announcing every one of those would bury the notices that matter.

`reply-files.ts` takes path-like tokens from the whole reply — prose, list, code block alike — splitting on whitespace, Markdown decoration and CJK punctuation, and trimming sentence punctuation off the edges. Existence is what makes that safe. It is deliberately looser than the Web App's file card, which reads inline code only: see *Follow-up fixes*. Dedupe is keyed on the normalized path, so a file named once relatively and once absolutely is one file.

- Resolution, containment and existence come from the **same `WorkspaceFilesService` the Files panel reads through** (`statExisting` + a bounded `read`), reached by widening `MessagingSessionIndex` to `{ workspace: string } | null` and adding one narrow `MessagingWorkspaceFiles` dep. A path leaving the Workspace — by `..`, absolutely, or through a symlink — is refused there, not by a second copy of the rules.
- `.png/.jpg/.jpeg/.gif/.webp` go as images, everything else as files. SVG stays a file on purpose: neither client renders it, and it is a scriptable document.
- Feishu uploads then sends (`im.v1.image.create` → `image_key`, `im.v1.file.create` → `file_key`, then `msg_type: image|file`); those two endpoints resolve to their inner `data` object, so a refusal arrives as a missing key and is reported as one. `file_type` is mapped from the extension onto Feishu's enum (`pdf`/`doc`/`xls`/`ppt`/…, `stream` otherwise) so the client previews what it can.
- Telegram posts `sendPhoto` / `sendDocument` as multipart, sharing the envelope handling with the JSON calls. `fetch` and `FormData` both come from **undici**, as a matched pair — see *Follow-up fixes*, where the mismatch is the reason no upload ever left the process.
- Image-versus-file is decided from the file the read actually reached, not from the name the reply spelled: an in-Workspace `report.png -> report.pdf` (the everyday "latest" symlink) would otherwise be posted to the image endpoint under the image ceiling and arrive named `report.pdf`, which Telegram rejects.

### Caps, and where the numbers come from

| Cap | Value | Source |
| --- | --- | --- |
| Per image | 10MB | min(Feishu image 10MB, Telegram `sendPhoto` 10MB) |
| Per file | 30MB | min(Feishu file 30MB, Telegram `sendDocument` 50MB) |
| Per run | 5 files | judgement about the medium, not an API limit |

The byte numbers are the tighter of each pair because the file is read once for whichever channel the Session happens to be bound to; the same reply behaving differently in two chats would be worse. A file over its ceiling is **named** in the chat, and a batch cut short by the count says how many were left behind — the Web App still has all of them. The read asks for one byte past the ceiling rather than the whole file, so measuring a 2GB log does not mean loading it.

Sends ride the existing per-entry `sendChain`, queued at the idle flip, so files land behind the reply that named them and in that order. They are always plain sends: a run that mentions a file has already sent the text that mentions it, and that message took the group's one reply-to. A failed upload is recorded through `recordError` and does not stop the files behind it; nothing here throws.

Every way a mentioned file fails to arrive is now **said in the chat**: over a byte cap, past the count cap, no such file inside the Workspace, or an upload the channel refused. The last of those previously reached `error_records` and nowhere the person in the chat could look, so the feature simply appeared not to work.

## Deliberately out of scope

- **Inbound documents/files.** The ask was images. It is the natural next step: the seam already carries a list of handles, and a `files?:` sibling with the same lazy `fetch(maxBytes)` shape would ride core's `[attached file: …]` path rather than the conversation.
- **Rich-text (Feishu `post`) messages with embedded images**, and Telegram pictures sent *as documents* — both answer with the not-supported notice. (The document case briefly did not: a captioned document reached the model as caption-only, with the file silently dropped. Fixed — see *Follow-up fixes*.)
- **Outbound files for a run the connection joined midway** — that run relays no text, so it has nothing to mirror files from.

## Follow-up fixes (after the first round was run for real)

Two bugs came back from real use, and both had passed a green suite. That is the finding worth recording: **the tests substituted a hand-written implementation of our own seam** (`FeishuSdk`, `TelegramTransport`). That proves the bridge drives the seam correctly and proves nothing about whether the adapter behind it speaks the actual protocol — and both bugs lived on exactly that blind side.

### 1. Every Feishu image failed, at any size

The download is coded correctly — URL, `:path` params and the resource type all match `@larksuiteoapi/node-sdk@1.73.0`, verified by driving the real SDK. What was broken is that the **reason never survived**.

`im.v1.messageResource.get` is issued with `responseType: "stream"`, and that applies to the failure too: on a refusal the SDK returns an error whose `response.data` is a **stream**, not a `{code,msg}` envelope. `larkErrorText` looked for `msg` there, found nothing, and reported `Request failed with status code 403` — discarding the one actionable fact, `app has no permission to access resource`.

Downloading message resources is **scoped separately** from receiving messages on the Feishu open platform, so an app set up for the existing text feature is refused here until that scope is granted. No code change can fix that; the fix is that the code now says it. The error body is read (bounded at 64KB — it is untrusted) and parsed, so the API's `msg` and `code` reach the chat and the error log.

### 2. Returning files produced nothing, on either channel

Extraction matched **inline-code spans only**. Measured against thirteen ways a reply really does hand a file over, eleven matched nothing:

| reply | before | after |
| --- | --- | --- |
| ``the chart is `chart.png` `` | hit | hit |
| `I've written the chart to chart.png for you.` | **miss** | hit |
| `Saved it at /ws/out/chart.png — take a look.` | **miss** | hit |
| `已经把图表保存为 chart.png，请查收。` | **miss** | hit |
| `Wrote **chart.png** to the workspace.` | **miss** | hit |
| `See [the chart](chart.png).` | **miss** | hit |
| `The report is report.md.` | **miss** | hit |

The card can afford the strict rule because it decorates Markdown that is already rendered — a path it misses simply is not clickable, and the file stays one click away in the Files panel. Here the extraction **is** the feature: what it misses never reaches the user. Tokens now come from the whole reply, split on whitespace, Markdown decoration and CJK punctuation (a Chinese sentence puts no spaces around a filename), with sentence punctuation trimmed off the edges; the ASCII colon and backslash are deliberately not boundaries, so a Windows path survives.

The cost is precision — a file the reply merely *read* can now be sent — bounded by three things: it must exist in the Workspace, at most 5 ride along per run, and the reply said the name out loud. `https://…/logo.png`, `v1.2`, `~/.penguinrc.toml`, `../../etc/passwd` and absolute paths outside the Workspace are all still refused.

### The notice was part of the bug

"That image could not be delivered (too large, or the download failed)" names two unrelated problems and therefore neither — which is precisely why the first report of bug 1 was indistinguishable from a size complaint. Now:

- **too large** — `That image is larger than the 20MB limit, so it was not sent to the Agent. Try a smaller one. 该图片超过 20MB 上限，未发送给智能体，请改用更小的图片。` Nobody's fault, so it is not an error record.
- **refused/failed** — `That image could not be downloaded from the chat, so nothing was sent to the Agent. 无法从会话中下载该图片，未发送给智能体。(<the channel's own reason>)`, recorded through `recordError` with that reason so it is diagnosable from the dashboard, not only from a chat bubble.

The two are distinguished by a typed `MessagingMediaTooLargeError` thrown by the shared capped reader, so a channel cannot report one as the other.

### Closing the testing gap

`packages/server/test/messaging-wire.test.ts` (18 tests) drives the **real adapters** against the transport each actually calls:

- **Feishu** — `createLarkSdk({ httpInstance })`, a new injection point that hands the real SDK a stand-in for axios (reproducing its response interceptor), so the SDK's URL building, `:path` filling, param serialization and response unwrapping all run for real. It asserts the resource URL and `type` param, that the type is read off the bytes rather than a lying header, that a 403 with a **streamed** error body surfaces `…no permission to access resource (code 99991672)`, and the upload path's real response shape (the upload endpoints resolve to their inner `data`, unlike the JSON endpoints' `{code,msg}` envelope — a difference no seam-level fake could have exposed).
- **Telegram** — a stub handed to the transport through its own `fetch` seam: the two-hop download, the `/file/bot<token>/…` endpoint, the multipart field names (`photo`/`document` + `chat_id`), a Bot API refusal's description, and that the bot token never reaches an error message though the download URL embeds it. Plus one test that stubs **nothing**: it builds the transport exactly as production does, points undici's global dispatcher at a local `http.createServer`, and asserts on the bytes that arrive.

Verified by reverting each fix in turn: the Feishu refusal test fails against the old adapter and passes with it, and the extraction matrix fails against the inline-code-only rule.

Both channels' seam-level fakes now enforce the size cap through the **real** `collectUnderCap` instead of an imitation — one of them had already drifted into disagreeing with production about which failure a cap is, which is the same class of miss in miniature.

### The one that made outbound files look unimplemented

`sendPhoto` / `sendDocument` posted the literal 17 bytes `[object FormData]` with `content-type: text/plain`, and the Bot API answered `Bad Request: there is no document in the request`. `installGlobalProxyDispatcher()` replaces `globalThis.fetch` with **undici's** so outbound traffic follows the proxy dispatcher; undici's fetch brand-checks a FormData body against **its own** class, and `new FormData()` is Node's built-in — a different class. The check failed, and the body fell through to `String(body)`.

Measured against a local server, all five combinations:

| fetch | FormData | on the wire |
| --- | --- | --- |
| undici | **global** | `text/plain`, 17 bytes — **what the server did** |
| undici | undici | `multipart/form-data`, correct |
| undici | undici + `node:buffer` Blob | `multipart/form-data`, correct |
| built-in | global | `multipart/form-data`, correct — **what CI did** |
| built-in | **undici** | `text/plain`, 17 bytes — the inverted bug |

So it failed 100% in production and 0% in CI: only the real entry installs that fetch. The transport now imports `fetch` **and** `FormData` from undici, with a comment at the construction site saying why — this file is the only `FormData` construction in `packages/server/src` and `packages/core/src`, so there is no precedent to copy from. Proxy support is unchanged (undici's fetch resolves the global dispatcher per call, which is what `net/proxy.ts` is written against), and `Blob` needs no treatment: `globalThis.Blob` **is** `node:buffer`'s Blob, and undici's FormData accepts it — verified on the wire, not assumed.

The guard is the test that asserts on what arrives, not on what the code intended. Both mismatch directions turn it red.

**The Feishu side is unaffected, from evidence:** `@larksuiteoapi/node-sdk@1.73.0` contains zero `fetch` calls and zero `FormData` references; it depends on axios, which serialises multipart through the `form-data` package over `node:http`. Different transport, different realm question. The Feishu upload failure is genuinely the missing scope and was not masked by this.

### The two Feishu failures the adapter could not report

- **A missing upload scope.** `code: 99991672` — `One of the following scopes is required: [im:resource:upload, im:resource]` — with a per-app grant URL in the same `msg`. That reached `error_records` and nowhere the person in the chat could look. It is now its own case, detected by **code** (the prose is localized and would not match reliably), and the chat notice names the scopes and carries Feishu's own console link. Nothing else off the error travels with it: the raw SDK error carries `config`/`request`/`response`.
- **A 400 whose reason was thrown away.** `im.v1.messageResource.get` is fetched as a stream, so a refusal arrives as the response **stream**, unread — `complete: false` in the owner's paste confirms nobody drained it. Read straight off the error it yields "Request failed with status code 400". The body is now drained and parsed, and `code`, `msg` and `log_id` are reported the way the ordinary calls already report them; if the code is `99991672`, the grant link applies here too.

### Review fixes

- **A captioned document ran the model on the caption alone, with the file silently dropped.** Only a photo's caption is that message's text now; every other media kind keeps the not-supported notice. Captions get the same `@mention` stripping as text.
- **`./chart.png` and every dotfile were destroyed before path resolution.** The edge-punctuation trim stripped leading dots along with trailing ones, so `./out/report.pdf` became an absolute path outside the Workspace and `.eslintrc.json` a file that is not there. (The `seg === "."` branch in `toWorkspaceRelative` had been unreachable dead code, which was the proof.)
- **The reply uploaded files it had explicitly declined to disclose.** Containment was right — the Workspace and nothing narrower — but inside it the only gate was "the reply said the name out loud", and the reply is steerable by anyone in the chat. Now the run must also have written the file.
- **A Feishu media call had no timeout and sat on the serial send chain.** The SDK's default transport is a bare `axios.create()` with no `timeout`, so a request the API accepted and never answered parked every later reply, approval notice and file for that Session — silently, no error record, no status change — until the binding was toggled or the server restarted. Every call this adapter makes now has a deadline.
- **Nothing bounded inbound image bytes in aggregate.** See the rolling budget above.
- **An upload that failed was the only drop invisible in the chat.** Now named, with the channel's reason, or with the scopes and grant link when permissions caused it. A mention that resolves to no file in the Workspace is named too.
- **Image-versus-file was decided from the mentioned name** while the uploaded name came from the canonical path. Both now come from the read.

### Extraction coverage the review asked for

`./chart.png`, `./out/chart.png` and `.eslintrc.json` joined the matrix. `KNOWN_EXTENSIONS` was checked against what an Agent plausibly hands over: `pdf`, `zip`, `xlsx`, `png`, `csv`, `json`, `log`, `sql` and `gz` are all present — the notable absences are `jsonl`, `tsv`, `parquet`, `db`/`sqlite`, `7z`, `tar.*` compound suffixes, `patch`/`diff` and `mdx`. Not widened here; a file with an unlisted extension is never a candidate, so it is silently not sent, which is the one drop arm still without a notice.

### Inbound test gaps

A captioned non-photo message; `./path` in the extraction matrix; a download that dies **mid-stream** after some chunks are already buffered; and an inbound "image" whose bytes are not an image, where the sniff fails and the header (or `image/png` as the last resort) has to label it.

## Verification

- `pnpm -r build` — pass
- `pnpm -r typecheck` — pass
- `pnpm -r test` — pass (server 1176, core 1021 + 5 skipped, web 1421, cli 374, desktop 144, docs 53, landing 69, skills 27)
- `prettier --check .` — "All matched files use Prettier code style!"

Rebased onto `origin/main` (past #490, #494, #497, #502, #503) and re-run from scratch; the numbers above are from that run. Every behavioural fix was mutation-tested: the change reverted, the new test confirmed red, the change restored.

New coverage in `packages/server/test/messaging.test.ts` and `messaging-telegram.test.ts` (both channels' fakes record downloads and uploads; the one test that opens a socket binds `127.0.0.1` on an ephemeral port): an inbound image becoming an `image_url` part with the right cap and message/file key; a photo's caption arriving as text ahead of the image; the largest photo variant being the one fetched; oversize and failed downloads degrading to the notice with no Task; a redelivered image downloading nothing; a non-text non-image message still getting the notice; a reply mentioning two existing files sending both, in order, after the reply text, as image + file; the same file named twice sending once; a mentioned path that does not exist sending nothing; `..`, an absolute outside path and a symlink escape all refused; the count cap with its notice; the per-image byte cap naming the file and the batch continuing; a failing upload not stopping the next file. Plus unit coverage of the media helpers (extension classification, magic-byte sniffing, the capped read).


hiyouga:
## Review

The largest PR in the batch, and the one that moves user files across a network boundary in both directions, so it got the most adversarial reading — then three live failure reports from running it against a real Feishu app and a real Telegram bot.

**The architecture is sound, and unusually careful where these paths normally go wrong.** The inbound cap rides *into* the fetch through `collectUnderCap`, so a lying `content-length` buys nothing and the reader is cancelled on the way out. Inbound bytes never touch disk under a platform-controlled name — core writes `upload-<8hex>.<ext>` into its own scratchpad. Outbound containment genuinely reuses `resolveRead`'s realpath-then-IO-on-canonical rule rather than a second copy, so `..`, absolute paths and symlink escapes are all refused with tests for each. And `deliverFiles` does **not** repeat the "one failure abandons the rest" bug #494 had just fixed next door — the per-file try/catch is inside the loop, with a test.

What follows is what the reading and the field reports turned up.

### 1. The feature had never once worked outside the test suite, on Telegram

Reported from a real run: `sendDocument failed: Bad Request: there is no document in the request (code 400)`.

Telegram was telling the exact truth. The server was posting the literal 17-byte string `[object FormData]` with `content-type: text/plain;charset=UTF-8`.

`index.ts` calls `installGlobalProxyDispatcher()`, which sets `globalThis.fetch = undiciFetch` — the **`undici` package's** fetch. But `telegram-api.ts` built `new FormData()`, Node's **built-in** copy. undici's fetch brand-checks the body against its own `FormData` class, the check fails, and it falls through to the generic `String(body)` branch. I measured all the combinations against a local server:

| | result |
|---|---|
| undici fetch + global FormData (what shipped) | `text/plain`, 17 bytes, `"[object FormData]"` |
| undici fetch + undici FormData | `multipart/form-data; boundary=…`, 295 bytes ✓ |
| built-in fetch + global FormData | `multipart/form-data; boundary=…`, 295 bytes ✓ |

So the multipart construction was never wrong, and neither was the deliberately-omitted content-type header — `headers: {}` does let fetch derive the boundary, and the doc comment saying so is correct.

**Why CI was green:** nothing exercised the real transport with the real global fetch. The suite uses the fake transport, and where it stubs `globalThis.fetch` it is still Node's built-in — the working combination. `installProxy()` only runs in the real server entry. 100% failure in production, 0% in CI.

**Fixed** — `telegram-api.ts` imports `fetch` **and** `FormData` from undici, with a comment at the site. Swapping only the FormData would have inverted the bug and broken the tests instead. `Blob` needs no change (measured: `globalThis.Blob` is `node:buffer`'s and undici's FormData accepts it). The guard test drives the production transport at a real `http.createServer` and asserts on the bytes that arrive; **both** mismatch directions turn it red.

Worth recording: the first draft of that fix named the resolved fetch `call`, shadowing the existing JSON-call helper and causing infinite recursion — caught by the tests, renamed. And Feishu is unaffected *from evidence*, not assumption: `@larksuiteoapi/node-sdk@1.73.0` has zero `fetch` calls and zero `FormData` references; it goes through axios and `form-data` over `node:http`.

### 2. A captioned Telegram document ran the model on the caption, with the file dropped

`text: typeof msg.text === "string" ? msg.text : (msg.caption ?? null)` gave *every* captioned media message a non-null `text`, while `images` came only from `msg.photo`. So a document with a caption reached the plain-text branch and started a Task on the caption with nothing attached and no notice — the exact outcome this PR's own inbound design forbids ("a model asked 'what is wrong with this?' about a picture it never received answers confidently about nothing"), and the opposite of what the PR body claimed. The common case is a picture Telegram sent as a document because it was dragged in uncompressed.

**Fixed** — the caption falls back only when a photo is present, and photo captions now carry `caption_entities` through the same mention stripping (the type's own comment had predicted this: "the day this connector reads captions, it reads that").

### 3. A reply that declined to disclose a file uploaded it anyway

Containment was exactly the Workspace and nothing narrower — `~/.ssh/id_rsa`, `web.db` and the `api-token` file are all out of reach, and that part is right. But *inside* the Workspace the only gate was "the reply said the name out loud", and the reply is steerable by anyone in the chat.

The concrete case: a group member asks "what's in secrets.json?". The agent does the right thing and answers "I won't paste secrets.json here." → the extractor picks up `secrets.json`, `statExisting` confirms it, and the server uploads the file. **The refusal was the disclosure.**

**Fixed** by making the feature match its own title — the bridge records `runStartedAt` at the idle→running edge and sends only files whose `mtime` is at or after it. A loose mention-matcher became a "files this run produced" rule, which is what the changelog and docs already claimed it was.

### 4. `./chart.png` and every dotfile were destroyed before resolution

`EDGE_PUNCTUATION = /^[.,;:!?]+|[.,;:!?]+$/g` stripped the leading dot too, so `./report.pdf` became `/report.pdf`, read as absolute and refused. An agent writing "saved to ./out/report.pdf" produced zero files, silently — and `.eslintrc.json` became `eslintrc.json`. The unreachable `if (seg === "." )` branch was itself proof `./x` was meant to work. Fixed to `/^[,;:!?]+|[.,;:!?]+$/g`.

### 5. A Feishu media call could wedge a Session's messaging permanently

The Lark SDK's default http instance is a bare `axios.create()` with **no timeout**, and `createLarkSdk` passed none — while Telegram had 60s on both directions. Because every send is appended to `entry.sendChain`, a request that never settles is not one lost file: the chain is the single ordering point for every subsequent reply, approval notice and file for that Session. A 30MB upload TCP-stalling behind a proxy blackhole meant that Session silently mirrored *nothing* to Feishu, with no error record and no status change, until the binding was toggled or the server restarted. **Fixed** — a deadline on every Feishu wire call, with a test seam.

### 6. Nothing bounded inbound image bytes in aggregate

The 20MB cap was per image; no rate or byte budget, no sender allowlist. Each accepted image is written into the conversation as a base64 data URL, and `services/attachment-limits.ts` states the consequence itself: an inline image "is written verbatim into the Trace JSONL, and that file is read back whole — into a single JS string — on every history page and every Session resume." Anyone who can DM the bot could send ~19 images just under the cap and pass 512MB of base64, after which every history page and every resume of that Session fails on the V8 string limit. The web-composer path reaches that ceiling only through an authenticated user; this one had no authentication at all. **Fixed** — a rolling per-binding budget, with the remainder riding into the fetch as its cap.

### 7. Your two Feishu reports

**The upload scope (`99991672`)** is a permission on your app, not a bug — but you had no way to know that, since it only reached `error_records`. Now detected by numeric `code` (not prose — the message is localized) and answered in the chat with the missing scopes and the app-specific grant URL Feishu itself returns. Nothing from `config`/`request`/`response` reaches the chat or the log; that object carries credentials.

**The inbound image 400** could not be diagnosed at all, and that was the real defect: `im.v1.messageResource.get` is fetched as a stream, so on an error axios returns the response stream and Feishu's actual error body sat inside it unread (`complete: false` in your paste confirmed it). Every inbound failure collapsed to an HTTP status. The stream body is now drained and reported as `msg (code N, log_id X)`. **I still do not know your 400's cause** — that needs the drained body from your app, which the branch now surfaces. The `type=image` hypothesis looks unlikely: the key is `img_v3_…` and only `messageType === "image"` reaches that call.

### 8. Smaller ones

A failed upload was the only drop invisible in the chat — every other arm was named. Now `noteFileFailure` in the per-file catch, plus a notice for mentions that resolve to nothing. Image-vs-file was classified from the *mentioned* path while the name sent came from the canonical one, so an in-Workspace `report.png → report.pdf` symlink was routed to the image endpoint under the image cap and rejected; now classified after the read.

### Answers to the three questions I asked

- **Is `deliverFiles` unreachable for a channel?** No. It is channel-agnostic, both connectors implement `sendImage`/`sendFile`, both register in `app.ts`. No capability gate.
- **What does a Telegram document get now?** Still the bilingual not-supported notice. Inbound files remain out of scope — a follow-up, not a bug.
- **Could the bind-time probe catch a missing upload scope?** **No, not cheaply.** `checkCredentials` only exchanges a tenant token, which needs no scopes and succeeds for an app with none. Detecting `im:resource:upload` would need a real `im.v1.image.create` — a side-effecting upload of throwaway bytes, which is a different kind of probe and its own PR.

### One remaining silent drop, deliberately left

An extension not in `KNOWN_EXTENSIONS` is never a candidate, so nothing can name it. `pdf`/`zip`/`xlsx`/`png`/`csv`/`json`/`log`/`sql`/`gz` are all present; notable absences are `jsonl`, `tsv`, `parquet`, `db`/`sqlite`, `7z`, compound `tar.*`, `patch`/`diff`, `mdx`. Widening the set is a judgement call I did not want to make silently — flagging it so you can.

Also deliberate: a file the run did not write is dropped without a notice, because announcing every config an agent mentions reading would bury the notices that matter.

### Checks

- Server **1176/1176**, whole monorepo green, `pnpm -r typecheck` clean, prettier clean, build clean. Every fix mutation-tested.
- Being rebased onto `main` again — it has taken #499, #501, #505, #506 and #508 since.
- **Design:** still outstanding. `05-ARCHITECTURE.md:1275` (the connector seam's method list), `:1280` (inbound images, the two distinct notices, the lazy fetch handle) and `:1281` (the reply-mention extraction, the caps, the run-scoped mtime rule) all need it. I am writing that.

hiyouga:
**Rebased onto current `main`.** All seven commits replayed, none skipped, new head `95aaebd7` on top of #504. `main` moved twice during it — #507, then #504 landing on `app.ts` and the server-api docs — so it went through two rounds.

Conflicts were in `bridge.ts`, `connector.ts`, `telegram-api.ts`, `telegram-connector.ts`, both server-api docs and two test files. The one resolution worth naming: in `onInbound`, this branch's early `return`s became `if/else` so that **every** path — text notice, task start, image notice, image task — still falls through to `main`'s `recordInboundWatermark` write. Left as returns, an inbound image would have skipped the watermark and been re-processed after a restart, quietly undoing #493.

All eleven post-rebase checks confirmed against the final head, including the six shipped fixes from `main` that a careless resolution reverts: leading-only `stripBotMention` with `cutSpan` gone, the un-awaited Feishu identity lookup, `topicIdOf`'s `is_forum && is_topic_message` gate, the 400-only thread retry, `deliverReply`'s single `lastInboundMessageId` snapshot, and QQ's redelivery-guarded `noteInbound`.

**Three fixups, each amended into the commit that caused it — and all but one failed `tsc` while the suite was green.** That is the trap this batch keeps setting: a hand-written structural type or a stale fixture is invisible to the tests. Two were type drift (`TextPayload[][]` → `InputPayload[][]` once a run's input mixes `image_url` parts; 17 call sites from `allSends()` to `allTexts()` once `sends` became a `SentText | SentMedia` union). The third was a real failure: #505's test *"stamps arrival for a message it cannot act on"* used a Feishu `image` as its example of "cannot act on" — which this PR makes actionable. The invariant under test is untouched, the test's own comment already calls the message a sticker, and its two siblings already use `messageType: "sticker"`, so the fixture now matches them.

Server **1287/1287**, web **1475/1475**, both typechecks clean, prettier clean across the whole repo, build clean. No flakes this round.

---

**One follow-up the rebase surfaced, for you to decide.** `deliverFiles`, `noteFileFailure` and `inboundImageParts` record through `recordError` (`messaging_image_fetch_failed`, `messaging_file_send_failed`, `messaging_send_failed`) rather than `recordDeliveryFailure`. That is this branch's own code and #505 never touched those paths, so there was no conflict and no mechanical answer. But by #505's own stated rationale — "a failure after an inbound message was accepted belongs on the binding's panel, because the error table is a Project-wide dashboard nobody visits to debug one bot" — a failed image download or a refused upload arguably belongs on `lastDeliveryError` too. Switching them would have been new behaviour invented during a rebase, so it was left alone. Worth its own change if you agree.

PR #497, opened 2026-08-27
fix(server): group messages reach the model as what the user wrote, not as mention placeholders

hiyouga:
The user drove a Session from a Feishu group and the model answered:

> 我看到你发的是 `@_user_1 你好`，但我不太确定 `_user_1` 指的是谁或什么

A whole turn spent on a token that was never part of the message. Addressing a bot in a group means mentioning it, so this is not an edge case — it is nearly every group message a binding receives, on both channels.

## What each channel was doing wrong

**Feishu** never puts a mention in the text. `im.message.receive_v1` writes a placeholder (`@_user_1`) into the content JSON's `text` and carries the real identity in a parallel `message.mentions` array — verified against the installed `@larksuiteoapi/node-sdk@1.73.0`, whose event type declares `mentions?: Array<{ key: string; id: { union_id?, user_id?, open_id? }; mentioned_type?: string; name: string; tenant_key? }>`. `FeishuInboundEvent` did not carry it, so `textOfContent` returned the placeholder verbatim.

**Telegram** delivers the handle as real text, so the model saw `@penguin_test_bot ask @alice about the build` — readable, but it announces the channel the message came through, which the bridge deliberately does not tell the model.

Both are reproduced by the new tests; against the pre-change source:

```
× resolves mention placeholders to names and drops this bot's own
  → expected '@_user_1 @_user_2 check the build' to be '@Alice check the build'
× names this bot's mention like any other when its identity is unavailable
  → expected '@_user_1 status?' to be '@PenguinHarness status?'
× replaces the longest placeholder first, so a tenth mention is not corrupted by the first
  → expected '@_user_10 and @_user_1 ship it' to be '@Jules and @Alice ship it'
× a message that is nothing but a mention starts no Task
  → expected 'Reply text' to be 'Only text messages are supported for …'
× strips this bot's own @mention out of a group message, keeping everyone else's
  → expected '@penguin_test_bot ask @alice about th…' to be 'ask @alice about the build'
× matches the bot's handle case-insensitively and recognizes a text_mention by user id
  → expected '@Penguin_Test_Bot deploy' to be 'deploy'
× a group message that is nothing but the bot's mention starts no task
  → expected 'Reply text' to be 'Only text messages are supported for …'
```

The two `'Reply text'` lines are the second half of the defect: an `@`-only message was starting a full LLM run on a bare placeholder, because `@_user_1` passes the `text.trim() === ""` guard.

## Feishu

`FeishuInboundEvent` gains `mentions`, mapped from the SDK payload in the production adapter. `resolveFeishuMentions` (exported from the connector, where the reduction of Feishu events to the bridge's normalized shape already lives) replaces each placeholder with the mentioned party's name.

**Longest key first.** Feishu numbers placeholders from 1 upward, so `@_user_1` is a literal prefix of `@_user_10`; replacing in event order corrupts the longer key once a message mentions ten people. Length order is the only order that is safe for any key set, and it has a test.

**This bot's own mention is dropped, not named** — the bridge feeds inbound text to the model exactly as if it had been typed into the web composer, and a leading `@PenguinHarness` announces the channel. Identity comes from `/open-apis/bot/v3/info`, one lookup per connection, held for that connection's lifetime by the closure that needs it. The SDK exposes no typed member for that endpoint, so it goes through the same generic `client.request` its own internals use. `botOpenId()` resolves **null** instead of throwing on every failure shape — a thrown request, a non-zero envelope, a body without the field — so a bot whose capability is off, or a transient failure, degrades to naming its own mention like anyone else's and never refuses a connection. Two tests cover both directions.

Name resolution alone would already have fixed the reported confusion; the identity lookup is what the `@`-prefix rule needs, and it met the bar — one cached GET, no guesswork about the call shape.

## Telegram

`getMe` already runs before every up-streak of the poll loop, so the bot's `id` and `@username` cost nothing extra; the loop just keeps the answer now. `stripBotMention` cuts the bot's own `mention` entities (span compared case-insensitively — Telegram matches handles that way and a hand-typed handle rarely matches capitalization) and `text_mention` entities (matched by `user.id`, the only handle a username-less bot has).

Spans are cut **back to front** so each remaining offset still indexes the same character, and one separating space is swallowed so `@bot hello` becomes `hello` rather than ` hello`. Entity offsets are UTF-16 code units, which is exactly what a JavaScript string index is — no conversion anywhere, written down on `TelegramMessageEntity` so nobody adds one later.

Other people's mentions stay on both channels: on Telegram they are literally what the user typed, and on Feishu a resolved `@Alice` is a name the model can use.

**Captions:** `TelegramMessage.entities`' doc names `caption_entities` as where the same shapes live for a captioned media message, and `stripBotMention` takes `(text, entities)` so a caption path passes its own pair. No unused wire field is declared for it — that belongs to the change that reads captions.

## A message with no words

On both channels a message left with nothing but mentions takes the existing empty-text branch (the bilingual text-only notice) instead of starting a run. On Feishu the emptiness is judged with **every** placeholder removed, not only this bot's, so it holds whether or not the identity lookup succeeded — the rule cannot depend on a network call.

## Contested files touched

All four are the connector/SDK files the media work is rewriting. Each hunk is confined to inbound text normalization and nothing else:

- `feishu-sdk.ts` — `FeishuMention`, one field on `FeishuInboundEvent`, `botOpenId()` on `FeishuApiClient`, `request` on the local `LarkClient` interface, and the mentions mapping inside the existing dispatcher handler. The outbound methods are untouched.
- `feishu-connector.ts` — `resolveFeishuMentions` plus the identity lookup in `connect`. `connect` became `async`; its return type was already a Promise.
- `telegram-api.ts` — `TelegramMessageEntity` and one field on `TelegramMessage`. Purely additive.
- `telegram-connector.ts` — `stripBotMention` and friends; the poll loop's only change is keeping the `getMe` result it already fetched and passing it to `normalizeUpdate`.

`bridge.ts` is **not** touched: the empty-text branch it already had is what a words-less message now reaches.

## Left undone on purpose

- **The text-only notice is reused for a mention-only message.** "Only text messages are supported for now" is not what happened — the user sent text, it just had no words in it. A notice of its own would be better copy; it is new user-facing strings for a case that is mostly a slip of the send key, so it is called out rather than invented here.
- **`mentioned_type` is not used.** The SDK's event type declares it but documents nothing about its values, so the bot is recognized by open id, which is unambiguous.

## Verification

Node 24.18.0, Linux. Full CI-parity chain from the worktree root: `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check`, `git diff --check` — all clean. Tests: server 1081 passed (92 files), core 1013 passed / 5 skipped (47 files), web 1360 passed (102 files), cli 374, desktop 144, landing 69, docs 53, skills 27. All seven new tests were run against the pre-change source and fail there (output above).

Tests added:
- `packages/server/test/messaging.test.ts` — a group message with two mentions (bot's dropped, other named), the identity-unavailable fallback, the `@_user_1`/`@_user_10` prefix ordering, a mention-only message, and a direct message whose text merely looks like a placeholder staying untouched.
- `packages/server/test/messaging-telegram.test.ts` — the bot's own handle stripped with another user's kept, case-insensitive matching plus `text_mention` by user id, and a mention-only group message.


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`: full server suite, mutation passes, and the offset arithmetic replayed by hand.

The Telegram half was correct as written — offsets are treated as UTF-16 code units, which is what a JS string index is, and an astral-cluster prefix slices exactly right. The Feishu half is where the problems were.

### 1. The bot-identity lookup blocked server boot on an untimed HTTP round trip

`connect()` awaited `(await this.sdk.createClient(creds)).botOpenId()` before returning, and that sits on the boot critical path: `index.ts:48 await server.buildDeps()` → `bootAppDeps` → `platform.ts:207 await deps.messaging.start()` → `bridge.ts:235 await this.connect(row)` (a *sequential* loop) → `connector.connect()`. `server.listen()` is line 52, i.e. after. `botOpenId()` swallows rejections but cannot swallow a promise that never settles, and the underlying call has no timeout — the Lark SDK's http instance is a bare `axios.create()` and passes none, and `formatPayload` first awaits an equally untimed tenant-token fetch.

So a server with one enabled Feishu binding restarting behind a stalled proxy, a captive portal or a blackholing firewall never comes up at all — no HTTP listener, no routes — because a best-effort *display-name* lookup stalled. Even the benign `ETIMEDOUT` case costs ~2 minutes per enabled binding on every boot and every hot swap, and the same await parked the enable request. It also broke the seam contract documented at `connector.ts:79-84` ("Resolves as soon as the connection is constructed and connecting") and diverged from Telegram, which does the identical lookup non-blockingly inside its poll loop.

**Fixed** in b9661072 — `connect()` awaits nothing; the id is filled in the background into a variable the `onMessage` closure captures, so at worst the first message or two name the bot's mention instead of dropping it, which was already the documented degraded mode. Mutation-verified: restoring the await makes the new test time out at 5105ms, which is the reported failure exactly.

Not done: reusing the bridge's client. The bridge has no client at connect time (`clientFor` builds one lazily on the first outbound send), so sharing would mean changing the connector seam — and Telegram already builds its own client inside `connect()`, so one client per connection is the established idiom.

### 2. The Feishu substitution re-scanned its own output

The loop rewrote `out` in place, so each later iteration re-scanned text earlier iterations had already substituted; the longest-key sort fixed the prefix problem but not this one. And `split(key).join(...)` replaced *every* occurrence, while Feishu emits exactly one per key — so a token the user literally typed was rewritten too.

Verified: `resolveFeishuMentions("@_user_1 and @_user_10 ship it", [{key:"@_user_1",name:"Alice"},{key:"@_user_10",name:"_user_1"}], bot)` → `"@Alice and @Alice ship it"` — the second person's display name re-substituted into Alice. And a user quoting this very bug (`"@_user_1 why did you say @_user_1 to me?"`, head mention being the bot) → `"why did you say  to me?"`, their own literal token silently deleted.

**Fixed** — one left-to-right pass over a single longest-first alternation regex, immune to both the prefix problem and the re-scan, plus a `Set` so each key substitutes at its first occurrence only. Both halves mutation-verified.

### 3. Own-mention removal was unconditional, so mid-sentence uses arrived garbled

Both channels cut *every* mention of the bot anywhere in the message, not just the addressing prefix the doc comments describe. Verified: `"hey @penguin_test_bot, what changed?"` → `"hey , what changed?"`; `"what is @penguin_test_bot's status?"` → `"what is 's status?"`; `"@penguin_test_bot summarize what @penguin_test_bot said yesterday"` → `"summarize what said yesterday"`. On Feishu a mid-sentence placeholder left a double space, because Telegram swallowed one following space and Feishu swallowed nothing.

**Fixed** — leading-only on both channels, which is both what the doc comments claimed and what preserves the sentence; a later mention is now named like anyone else's. The whitespace policy is literally the same on both sides (cut the span, trim), so `cutSpan` is gone. Mutation-verified per channel. Both doc comments and the changelog pair were updated, with a new mirrored section for the cross-channel rule.

### 4. Nothing pinned the UTF-16 claim

`telegram-api.ts:28-33` and the changelog both state offsets are UTF-16 code units and that nothing converts them, but every new Telegram case put the mention at offset 0 of an all-ASCII string — where code-point, UTF-16 and byte indexing all agree. **Fixed** — a leading `text_mention` whose span covers an emoji-bearing display name is the case where the arithmetic is still observable once cutting is leading-only (a code-point implementation returns `"ploy"` instead of `"deploy"`), plus three non-leading cases and four new Feishu cases. Mutation-verified with a `[...text]` re-index.

### Still open before merge

**Design.** `05-ARCHITECTURE.md:1280` says a text message becomes 普通用户输入 verbatim and that 非文本消息 takes the 「仅支持文本」 branch — it says nothing about mention normalization, and it is now also wrong that only non-text messages take that branch (a mention-only text message does too). `:1279` enumerates each channel's connect sequence (Telegram's 先 `getMe` 探测凭据 is there) and needs Feishu's best-effort identity lookup. Both live in still-open design PR #61; being prepared as a PR onto that branch.

### Checks

- Server suite 1087/1087, typecheck clean, prettier clean.
- Changelog compliant — metadata order, counterpart link, sections mirrored, no banned headings, claims match the diff.
- Nit left alone deliberately: `feishu-connector.ts:71` has 你好 in a source comment. There is close precedent (`报告.pdf` in `task-attachments.ts`, `第 N 轮` in `message-window.ts`), so I read it as within the load-bearing-fixture allowance rather than residue.

hiyouga:

PR #498, opened 2026-08-27
Scheduled tasks target the current Session by default

hiyouga:
When an Agent sets up a scheduled task and the user has not said where it should fire, the task should land in the conversation the user is having — not in a fresh Session nobody is watching. This settles that in the built-in Schedules Prompt.

## The prompt change

The `# Scheduled Tasks` section's field rules used to end:

> `end_at` (optional) must be later than start_at; a periodic task stops after it. **Each trigger starts a new Session by default: `workspace` (optional) picks its working directory, and `provider` + `model_id` pick its model — always both or neither; omit both to use the Project's default model. Setting `session_id` instead sends the prompt into an existing Session, and cannot be combined with workspace / provider / model_id.**

It now ends:

> `end_at` (optional) must be later than start_at; a periodic task stops after it. **`session_id` picks where the prompt is delivered, and by default it is this Session: write the id from the Environment section's Session ID line, so every trigger arrives in the conversation you are in now, with its context intact. Omit it when the user asks for a separate Session, or when the task is better off starting clean — a nightly report that should not inherit this conversation's history. Each trigger then opens a new Session, in which `workspace` (optional) picks the working directory and `provider` + `model_id` pick the model — always both or neither; omit both to use the Project's default model. `session_id` cannot be combined with workspace / provider / model_id.**

with a new paragraph after it:

> **A repeating task aimed at this Session grows its context on every trigger and nothing ends it on its own: bound it with `end_at`, use a one-shot task for a one-time reminder, and keep the work each trigger asks for small.**

and two lines added to the fenced example, so the example demonstrates the default target and the bound rather than contradicting them:

```toml
end_at = 2026-09-01T09:00:00Z
session_id = "<session_id>"
```

The model can resolve `<session_id>`: the default template's `# Environment` section already carries `- Session ID: {{SESSION_ID}}`, substituted by `assembleSystemPrompt`. `<session_id>` is the same angle-bracket-pattern convention the section already uses for `<app_data_dir>` and `<agent_id>`.

Guidance only — nothing in the TOML format moved. `schedule-file.ts` still opens a new Session for a file that carries no `session_id`, and still rejects a file that combines `session_id` with `workspace` / `provider` / `model_id`.

## Why the hygiene sentence is not optional

`penguin-orchestration`'s own caution says an agent that messages itself — "through a schedule aimed back at its own session" included — "keeps spending until someone stops it". This change makes exactly that shape the default, so the prompt has to carry the antidote rather than leave it in a skill the Agent may not have installed: a recurring task firing into its own Session accumulates context on every trigger and terminates only when a person terminates it. Hence `end_at`, one-shot for "remind me later", and small per-trigger work — stated in one sentence, in the same section that creates the hazard.

The counter-example is kept honest rather than papered over: a nightly report that should not inherit the conversation's history is a genuine reason to open a fresh Session, and the prompt says so and lets the model make that call.

## Existing Agents

`system_config.yaml` is materialized at Agent creation and never auto-upgraded, so this follows the repo's kernel-version mechanism exactly:

- `KERNEL_VERSION` advanced `2026-08-24` → `2026-08-26`.
- `KERNEL_DEFAULT_TAB_HASHES.schedules` rewritten to the recomputed hash; the previous one appended as the first (and so far only) entry of `KERNEL_SUPERSEDED_TAB_HASHES.schedules`, with a generation-list line saying what moved.
- Verified rather than assumed: hashing `{ schedules: { enabled: true, prompt } }` over `HEAD`'s prompt text reproduces `79123643…` exactly, so an untouched Agent really does hash-match the frozen record and its Schedules tab advances whole instead of being conservatively kept.

**What a user sees:** an existing Agent keeps its stored `schedules.prompt` and runs the old guidance. Its settings page and the agents list show the kernel-update hint. On **Update kernel**, a Schedules tab still equal to the previous built-in default is rewritten to the new prompt; a Schedules tab the user edited in any way is kept whole and listed as kept, with restore-default-config as the full-refresh recourse. No migration was invented and nothing on disk is rewritten without that explicit action.

## The CLI skill, and one thing deliberately not implemented

`packages/skills/skills/penguin-orchestration/SKILL.md` documented `penguin schedule add`'s target flags as simply "optional", which would now contradict the prompt. Brought into line: the recipe shows `--session-id "$PENGUIN_SESSION_ID"` first and keeps the untargeted `daily-report` as the labelled clean-context case; the flag bullet states the default and when to leave it off; the "No runaway loops" caution now names the bound (`--end-at`, or no `--period` at all) instead of only warning. Skill `version` 2 → 3 with `updated` moved, per the repo's rule that every content change bumps both.

**Recommendation, not implemented here** (this PR is scoped to the prompt): `penguin schedule add` should default `--session-id` from `PENGUIN_SESSION_ID` when the caller is inside a harness Agent session and no target flag is given. The CLI already has the convention — `run`'s caller-context defaults fill `--workspace`, the model pair, `--approve` and `--thinking` from the caller's live session, with precedence *explicit flag > caller value > plain fallback* — and `PENGUIN_SESSION_ID` is injected into every command subprocess (`server/src/app.ts`). Doing it there would make the CLI surface agree with the prompt by construction instead of by documentation, and would remove the `"$PENGUIN_SESSION_ID"` boilerplate from the skill. It needs its own decision because it changes an existing command's behaviour for callers who omit the flag today.

## Verification

Full CI-parity chain, from the worktree root, Node v24.18.0:

- `pnpm -r build` — exit 0
- `pnpm typecheck` — exit 0 (9 packages)
- `pnpm test` — exit 0; 4114 passed, 5 skipped (core 1014, server 1073, web 1360, cli 374, desktop 144, landing 69, docs 53, skills 27)

New coverage in `packages/core/test/prompt-sections.test.ts`: the rendered prompt's `# Scheduled Tasks` instruction and the `# Environment` section's `Session ID` line are pinned to one shared literal, so renaming the Environment label fails the test until the section is re-pointed; the fresh-Session escape, the exclusivity rule and the `end_at` bound are asserted alongside it. `core/test/kernel-version.test.ts` (the pinned-hash guard and the tab-merge semantics) passes on the advanced record.

### DIFF
--- a/changelog/unreleased/2026-08-26-schedule-target-current-session.md
+++ b/changelog/unreleased/2026-08-26-schedule-target-current-session.md
@@ -7,13 +7,14 @@
 
 [中文版](2026-08-26-schedule-target-current-session.zh.md)
 
-The built-in Schedules Prompt now tells the Agent to aim a scheduled task at the Session it is already in — `session_id` written from the Environment section's Session ID line — so a task arranged in conversation reports back into that conversation instead of into a fresh Session nobody is watching. Opening a new Session per trigger became the deliberate alternative: when the user asks for a separate Session, or when the task is better off starting clean.
+The built-in Schedules Prompt now tells the Agent to aim a scheduled task at the Session it is already in — `session_id` written from the Environment section's Session ID line — so a task arranged in conversation reports back into that conversation instead of into a fresh Session nobody is watching. Opening a new Session per trigger became the deliberate alternative: when the user asks for a separate Session, when the task is better off starting clean, or when it has to outlive the conversation.
 
 ## Details
 
-- The `# Scheduled Tasks` field rules lead with `session_id` and its default, and the fenced example carries `session_id` alongside an `end_at`. `session_id` still cannot be combined with `workspace` / `provider` / `model_id`, and a file that omits `session_id` still opens a new Session on every trigger — the TOML format itself is untouched.
-- A hygiene line bounds the shape the new default makes ordinary: a repeating task aimed at its own Session grows that Session's context on every trigger, so it asks for an `end_at`, a one-shot task for a one-time reminder, and small per-trigger work.
-- The `penguin-orchestration` skill's `penguin schedule add` guidance follows the same default (`--session-id "$PENGUIN_SESSION_ID"`), and its runaway-loop caution names the bound that ends such a task.
+- The `# Scheduled Tasks` field rules now describe `session_id` and its default before the new-Session form, and the fenced example carries `session_id` alongside an `end_at`. `session_id` still cannot be combined with `workspace` / `provider` / `model_id`, and a file that omits `session_id` still opens a new Session on every trigger — the TOML format itself is untouched.
+- The same rules name where this Session is the wrong target. A Session is not the durable identity of a conversation — switching the model or the agent opens a new one, and deleting the conversation strands a task bound to it — so anything that has to outlive the conversation takes the new-Session form. A subagent, which renders this same section against its own short-lived Session, omits `session_id` unless its caller gave it an id to target.
+- A hygiene line bounds the shape the new default makes ordinary: a repeating task aimed at its own Session grows that Session's context on every trigger, so it asks for an `end_at` when the request has a natural horizon, a one-shot task for a one-time reminder, and small per-trigger work. An open-ended reminder keeps no `end_at`: the Agent leaves it off and says in its reply that the task runs until the user removes it, instead of inventing an expiry date the reminder would then silently stop on.
+- The `penguin-orchestration` skill's `penguin schedule add` guidance follows the same default (`--session-id "$PENGUIN_SESSION_ID"`), and its runaway-loop caution names the bound that ends such a task, on the same condition.
 
 ## Existing Agents
 
--- a/changelog/unreleased/2026-08-26-schedule-target-current-session.md
+++ b/changelog/unreleased/2026-08-26-schedule-target-current-session.md
@@ -1,0 +1,19 @@
+# Scheduled tasks target the current Session by default
+
+- **Date:** 2026-08-26
+- **Type:** feature
+- **Scope:** `core`, `skills`
+
+[中文版](2026-08-26-schedule-target-current-session.zh.md)
+
+The built-in Schedules Prompt now tells the Agent to aim a scheduled task at the Session it is already in — `session_id` written from the Environment section's Session ID line — so a task arranged in conversation reports back into that conversation instead of into a fresh Session nobody is watching. Opening a new Session per trigger became the deliberate alternative: when the user asks for a separate Session, or when the task is better off starting clean.
+
+## Details
+
+- The `# Scheduled Tasks` field rules lead with `session_id` and its default, and the fenced example carries `session_id` alongside an `end_at`. `session_id` still cannot be combined with `workspace` / `provider` / `model_id`, and a file that omits `session_id` still opens a new Session on every trigger — the TOML format itself is untouched.
+- A hygiene line bounds the shape the new default makes ordinary: a repeating task aimed at its own Session grows that Session's context on every trigger, so it asks for an `end_at`, a one-shot task for a one-time reminder, and small per-trigger work.
+- The `penguin-orchestration` skill's `penguin schedule add` guidance follows the same default (`--session-id "$PENGUIN_SESSION_ID"`), and its runaway-loop caution names the bound that ends such a task.
+
+## Existing Agents
+
+The config kernel advanced to `2026-08-26`, moving the Schedules tab. An Agent already on disk keeps its stored `schedules.prompt` and runs the old guidance until its owner takes the kernel update on the Agent settings page; a Schedules tab still holding the previous built-in default is rewritten by that update, while an edited one is kept whole.
--- a/changelog/unreleased/2026-08-26-schedule-target-current-session.md
+++ b/changelog/unreleased/2026-08-26-schedule-target-current-session.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-26
 - **Type:** feature
 - **Scope:** `core`, `skills`
+- **PR:** [#498](https://github.com/Prism-Shadow/penguin-harness/pull/498)
 
 [中文版](2026-08-26-schedule-target-current-session.zh.md)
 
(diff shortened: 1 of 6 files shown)


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`. The mechanics all check out: the kernel-history bookkeeping is done exactly right — I reproduced the frozen superseded hash independently by hashing `{schedules:{enabled:true,prompt}}` over `origin/main`'s prompt and got `79123643…`, so an untouched Agent really does advance whole rather than being conservatively kept. The TOML format and `schedule-file.ts` are untouched, so this is guidance only, which is a sound and cheap way to ship it.

The defects were in the instruction text the PR exists to write.

### 1. The `end_at` rule was unconditional, so an open-ended reminder silently expired

"bound it with `end_at`" was stated as a rule for every repeating task aimed at this Session, and the skill hardened it to "needs an `--end-at`". No carve-out for a task the user wants running indefinitely — the most common shape of the very request this PR optimizes for.

"Remind me every weekday at 9 to check the build" → the model writes `period = "1d"` plus an `end_at` it invents (the example hands it one a month out). On that date `evaluateEntry` returns on `slotInWindow(def, slot) === false` with no event and no error record, and `nextFireAt` returns `undefined`. The only signal anywhere is an amber `expired` chip on Agent settings → Schedules, a page the user has no reason to open. The reminder just stops arriving.

**Fixed** in 578c34a1 — the bound is conditional, and when the user wants it open-ended the model leaves `end_at` off and says so in its reply. Mirrored in the skill's *No runaway loops* bullet.

### 2. "This Session" is not durable, and nothing said so

The prompt promised "every trigger arrives in the conversation you are in now", but a Session is not the durable identity of a conversation. `/model` and `/agent` open a **new** Session while the old one still resolves and `statusOf` reports it idle — so every trigger keeps landing off-screen, forever. Deleting the conversation makes the next fire record `schedule_session_missing` and `markInvalid(..., "session_missing")`: the task never fires again, the delete route does no schedule cleanup and issues no warning, and recovery means hand-editing the TOML.

**Fixed** at the prompt level — the field rules now say to omit `session_id` for anything that has to outlive the conversation, and name both ways a bound task is stranded.

### 3. A subagent read the same default, inverting the PR's own rationale

`run_subagent` spawns via `childAgent.createSession({ source: "subagent" })`, which runs the same `assembleSystemPrompt` with the child's own Session ID — so the section rendered identically inside a subagent. A delegated "set up the nightly report" bound the task to the child session, released when idle and visible only in the parent's subagent panel: precisely the "fresh Session nobody is watching" the change set out to avoid. **Fixed** with one clause reusing the template's existing self-test idiom.

### 4. The skill's new recipe hard-failed after 2026-09-01

`--start-at now` was paired with a literal `--end-at 2026-09-01T09:00:00Z`; `upsert` reparses the serialized file and 400s on `end_at must be later than start_at`. After that date an agent following the recipe gets `Invalid schedule configuration: end_at must be later than start_at` and has to work out that the skill's own example is the problem. **Fixed** — a placeholder, matching the skill's own convention.

### 5. Two smaller ones

The changelog claimed "the field rules lead with `session_id`" — they don't; the paragraph still opens with `prompt` and reaches `session_id` last. What moved was the ordering *within the target discussion*. Fixed in both languages. And the test pinned the Environment-label coupling but not the one the fenced example actually depends on: `session_id = "<session_id>"` only resolves because the `# File system` section declares angle-bracket names substitutable — and that lives in the `system_prompt` tab, a *different* kernel tab, which can advance or be user-edited independently. Now asserted.

**Kernel hash re-derived**: `schedules` moves to `c722922e…`. `KERNEL_VERSION` stays `2026-08-26` (same day's generation) and the superseded list is untouched — deliberately not appending the intermediate `4b8e9372…`, since that generation exists only inside this unmerged PR and no config on disk can carry it.

### Raising separately, not fixed here

The server never validates `session_id` on write. `upsert` already rejects a bad `(provider, model_id)` via `validateScheduleModelRef`; nothing does the equivalent for the session, so a dead target — or a literal unsubstituted `<session_id>`, which parses, registers, and lists as `active` with a `nextFireAt` — dies only at its first fire, after showing the user a healthy task. That is a server change outside this PR's scope, but this PR is what makes the unvalidated field the default path, so it shouldn't stay silent. Related: deleting a Session does no schedule cleanup and issues no warning.

### Still open before merge

**Design.** `05-ARCHITECTURE.md:1266` § 定时任务 still reads 目标二选一：`session_id`… 或缺省**每次新建 Session**. True of the file format, but it no longer describes what an agent will actually do, and neither the built-in Schedules Prompt's default target nor this kernel generation appears anywhere in `design/specs/`. Precedent is mixed — the 2026-08-20 and 2026-08-24 kernel bumps each got an `08-CHANGELOG.md` entry, the 2026-08-21 memory-prompt reword did not — but this one flips the ordinary target of every agent-created task, so it outruns the spec even though it doesn't contradict it.

### Checks

- Core suite 1014 passed / 5 skipped, `prompt-sections` + `kernel-version` 45/45, skills 27/27, typecheck clean, prettier clean.
- Skill frontmatter well-formed, `version` left at 3 (unpublished, so editing in place is right), `short_description_zh` still matches the English.
- Grep sweep: the schedules prompt text exists only in `default-config.ts`; the old hashes appear nowhere but `kernel-history.ts`; the docs' schedule tables describe the TOML/CLI format this PR doesn't change.

hiyouga:
**Design sync is up:** [penguin-harness-design#71](https://github.com/Prism-Shadow/penguin-harness-design/pull/71), against design `main`.

It appends to the § 定时任务 `配置` bullet rather than changing what that bullet says about the **file format** — the format's 目标二选一 is still true; what was missing is that the built-in prompt now gives agent-created tasks a default, with the three carve-outs this PR's review added (conditional `end_at`, outlive-the-conversation, subagent). The framing 提示词层面的默认、不是格式的缺省 mirrors § 选择方式's existing 提示词层面的软约定，不是运行时机制.

On the `08-CHANGELOG.md` bar: it clears it. The closest precedent is 2026-08-20 压缩阈值缺省值提到 256000 — a built-in default with no user-visible surface change, entered because it changes what every new Agent does, and carrying the same 内置默认变更照例前进配置内核版本 rider. The counter-precedent holds too: the memory-prompt reword got no entry. The difference is that this flips the target of every agent-created task, which is behaviour rather than 措辞.

The entry is dated **2026-08-26**, to agree with this PR's changelog entry and the `2026-08-26` kernel generation label, so it sorts below the two 2026-08-27 entries. Say if you would rather the design log carry the day it was recorded.

Also confirmed while there: the spec promises **no** write-time validation of `session_id` — only reconcile-time 非法文件跳过并记异常 and fire-time 绑定的 Session 已删除时记异常并在 SQLite 标记失效. The only write-time rejection it records for schedules is the model reference. So the gap I raised above is an implementation gap the spec never promised away, and nothing was added to the spec to paper over it.

PR #499, opened 2026-08-27
feat(server,web): a Telegram bot that cannot hear a group now says so

hiyouga:
New report: in a Telegram **group**, messages to the bot get no response at all. Nothing in the product is broken — the poll loop is healthy, the status reads `connected`, no error is recorded, and `getUpdates` simply returns nothing.

## The cause is Telegram's Group Privacy, and it is on by default

Verified against https://core.telegram.org/bots/features rather than assumed. Under privacy mode a bot receives only:

> Commands explicitly meant for them (e.g., `/command@this_bot`). General commands (e.g. `/start`) if the bot was the last bot to send a message to the group. Inline messages sent via the bot. Replies to any messages implicitly or explicitly meant for this bot.

A plain sentence is not on that list — **and neither is a bare `@thisbot 你好`**, which is the thing everyone assumes works. Two more facts from the same page decide the copy:

> Privacy mode is enabled by default for all bots, except bots that were added to a group as admins (bot admins always receive all messages).

> the bot will need to be re-added to the group for this change to take effect.

So the failure is total, silent, and the default. This PR cannot fix it — the setting belongs to the bot's owner in @BotFather — so it makes it *visible*, which is the part that was missing.

## What the user sees now

**The credential test reports the setting.** `getMe` answers `can_read_all_group_messages` — *"True, if privacy mode is disabled for the bot. Returned only in getMe"* — and `TelegramBotUser` did not carry it. It now rides `MessagingAccountInfo.readsGroupMessages` through the same path the bot's `@username` already takes, and the route states it as `groupPrivacy` (the inverse), because **Group Privacy** is the name the user has to go and look for in @BotFather. A successful test whose bot still has it on adds a second, separate toast: the credentials really are fine and the bot really will answer a direct chat, so folding this into the success line would misreport both.

**It reports the setting, never a verdict on a group.** `can_read_all_group_messages` is account-wide, and Telegram overrides privacy mode wherever the bot is an administrator — nothing here asks about per-group admin status, so the copy says *"it receives no ordinary messages in any group where it is not an administrator"* and stops there. See the review-fixes section: the first revision asserted the outcome, and its advice would have broken a working admin bot.

**Unknown is reported as unknown.** The flag is only set when `getMe` actually answered with a boolean. A response that omits it says nothing, because sending someone to @BotFather over an absent field wastes the trip. Pinned by a test.

**The troubleshooting fold carries the durable version**, beside the one-program-per-token entry #484 added — that fold is where a user who has not run a credential test will look, and where remedies survive longer than a four-second toast. It leads with the remedy that touches nothing else (make the bot an administrator of the group), then the `/setprivacy` route with its re-add caveat. Both dictionaries, keys aligned, order pinned by a test.

## Considered and rejected: showing it in the connection status

The status row reports connection health. A bot with privacy on is connected and working — it answers direct chats normally — so a marker there would read as a fault on a binding that has none, on every one-to-one binding in existence. It stays out.

Also left out: `can_join_groups` is not surfaced. It is false only for a bot whose owner disabled group membership, in which case the bot cannot be in the group to begin with and the user is not looking at this panel. Adding it would be a second warning for a case the first one cannot arise in — so rather than declare it and leave it unread, the field was removed from `TelegramBotUser` in the review round.

## Review fixes (second commit)

1. **The report false-alarmed on an admin bot, and its advice would have broken it.** The copy asserted *"it receives no ordinary group messages"* from a bot-global flag, then told the user to *"remove the bot from any group it is already in and add it back"* — which strips exactly the administrator status that made a working group work, and the warning would have repeated forever, since `getMe` never learns about admin status. Fixed in all four places (`testPrivacyOn` in both dictionaries, `TelegramTestResponse.groupPrivacy`'s doc, both `server-api` rows), and the fold entry now leads with the admin route as the standalone fix it is.
2. **The changelog recorded reasoning and was written in the present tense.** `## Not shown in the connection status` was an alternatives-considered section under a bespoke name; it lives here instead, where `changelog/README.md` says it belongs. The rest is rewritten as what the change did. Its claim that *"any two of them leave the bot just as mute"* was wrong — the admin route alone fixes it, and `/setprivacy` + re-add fixes it without admin.
3. **The privacy toast could not be read before it disappeared** — 45 words of three-step instruction through a 4s `info` toast with no hover-pause. It now carries the diagnosis and points at the Troubleshooting fold, which holds the steps; a test pins that the pointer is the fold's own title, so renaming the fold cannot strand it. The key moved up beside `testOk`/`testOkAs`/`testFail`.
4. **The zh fold entry promised 三选一 and listed two.** Gone with the rewrite; both dictionaries now name the same two routes in the same order, asserted per dictionary (`S` is the zh dictionary at import time, so a single-dictionary assertion is vacuous).
5. **Stale seam doc and a dead field.** `MessagingAccountInfo`'s header still described a single label payload whose absence was `null`; it now describes a payload of independently optional members. `TelegramBotUser.can_join_groups` was declared, documented, and read nowhere — deleted.
6. **Tests covered the flag, not the new behaviour.** The response-to-notices decision was extracted as a pure `telegramTestNotices` (the `updateCheckOutcome` idiom already in this repo), so the two-line case is pinned by value rather than sitting behind a hook a node-only vitest suite cannot render; the dispatch of each tone to its toast is pinned against the source (`account-menu.test.ts` convention). `FakeBotClient`'s username became optional, covering `checkCredentials` returning a payload instead of `null` when `getMe` names none.

## Files touched

- `packages/server/src/http/routes/messaging.ts` — three lines inside the existing `/telegram/test` handler.
- `packages/web/src/features/messaging/messaging-binding-editor.tsx` — `telegramTestNotices` plus the loop that shows what it returns, one `<li>` in the troubleshooting fold, two imports.
- `packages/web/src/lib/strings.ts` / `strings-en.ts` — two new keys each (`troubleGroupPrivacy`, `testPrivacyOn`) plus the doc comment above the trouble entries, which enumerates them and would otherwise go stale.
- `telegram-api.ts` / `telegram-connector.ts` / `connector.ts` — one optional field added to `TelegramBotUser` and one to `MessagingAccountInfo`, `checkCredentials` returning them, `can_join_groups` removed, and both seam docs rewritten.
- `bridge.ts` — the credential-probe passthrough only. Neither the inbound nor the outbound path is touched.
- `packages/docs/content/server-api.{en,zh}.md` — the `/messaging/telegram/test` response row, in both languages.

## Verification

Rebased onto `main` after #497 landed (which carried a third commit this branch predated); `stripBotMention` still cuts only the leading mention and the Feishu identity lookup is still off `connect()`'s await path — `feishu-connector.ts` is byte-identical to `main`.

Node 24.18.0, Linux. `pnpm install --prefer-offline`, `pnpm -r build`, `npx tsc --noEmit` in `packages/server` and `packages/web`, `npx prettier --check` on every changed `.ts`/`.tsx` — all clean. Tests: **server 1129 passed (92 files)**, **web 1428 passed (105 files)**, including `placeholders-parity`.

Every behavioural change was mutation-tested — source reverted, suite run, red confirmed, restored:

| Reverted | Test that went red |
| --- | --- |
| the `groupPrivacy === true` push in `telegramTestNotices` | `adds the privacy notice as a second line, leaving the success line untouched` |
| the `info` branch of `testConnection`'s dispatch | `is shown by testConnection, one toast per notice, each in its own tone` |
| the admin remedy back to the tail of the fold entry (both dictionaries) | `leads the fold entry with the admin route, then the @BotFather one` |
| the toast back to repeating the remedies | `points the toast at that fold instead of repeating the remedies in it` |
| `checkCredentials` answering `null` when `getMe` names no username | `POST /test reports Group Privacy for a bot getMe names no username for` |

Tests added:
- `packages/server/test/messaging-telegram.test.ts` — privacy on → `groupPrivacy: true`, off → `false`, the field absent from `getMe` → absent from the response, and a bot with no `username` → `{ok, latencyMs, groupPrivacy}` with no `botUsername`.
- `packages/web/test/messaging-binding-editor.test.ts` — `telegramTestNotices` for the two-notice, one-notice, unknown and failure cases; the tone-to-toast dispatch; the fold's remedy order and the toast's pointer, per dictionary; the Telegram fold carries both Telegram-only entries and the Feishu fold carries neither.


hiyouga:
## Review

Reviewed in a dedicated worktree, then rebased onto the new `main` and fixed.

The mechanism is sound and needed no change: `getMe`'s `can_read_all_group_messages` is exactly the field that means privacy mode is disabled, the route inverts it into the name the user has to look for in @BotFather, and the unknown case is correctly withheld rather than guessed. Error handling holds up too — `getMe` was already behind the existing 15s `AbortSignal.timeout`, and every failure shape (timeout, 4xx, a 429 envelope, a malformed body) lands on the pre-existing `{ok:false, error}` path. Nothing can be `undefined` on the wire: all three hops spread-guard, the built `dist/api/types.d.ts` carries `groupPrivacy?`, and the web reads it with a strict `=== true`.

What the change got wrong was the boundary of what the flag proves.

### 1. The report false-alarmed on an admin bot, and its advice would have broken that bot

`can_read_all_group_messages` is the **bot-global** @BotFather setting. Telegram overrides privacy mode **per group** when the bot is an administrator there — admins receive every message regardless. Nothing in the code asks about per-group admin status (no `getChatMember`, no `getChatAdministrators`), so it cannot distinguish "muted" from "admin, hearing everything". The copy asserted the outcome flatly: *"it receives no ordinary group messages."*

The concrete case: a user adds `@mybot` to a group **as an administrator** and leaves privacy at its default ON. The bot hears and answers everything; the binding is perfect. They click Test connection and are told Group Privacy is on, it receives no ordinary group messages, and to *"remove the bot from any group it is already in and add it back"* — which **strips the administrator status that was making it work**. And it was unfixable from their side: a user who took the third remedy (make it an admin) would get the identical warning on every subsequent test forever, because `getMe` never learns about it.

**Fixed** in 7e2ecc55 — the copy now reports the *setting* rather than the outcome ("in any group where it is not an administrator, it receives no ordinary messages"), in all four places: both dictionaries, `TelegramTestResponse.groupPrivacy`'s doc, and both `server-api` rows. The fold entry now leads with the admin route as the standalone fix it is, since it is the one remedy that does not require touching group membership. The same admin/global conflation was fixed in `MessagingAccountInfo.readsGroupMessages` and `TelegramBotUser.can_read_all_group_messages`. Mutation-verified.

### 2. The rebase hazard, and how it was avoided

This branch was based on `c872c2e7`, but #497 gained a third commit after that — the one that took the Feishu identity lookup off `connect()` and narrowed `stripBotMention` to the leading mention — and *that* is what `main` carries as `db3bbd52`. A plain `git rebase origin/main` would have replayed #497's superseded commits and invited a resolution that silently reverts both: the all-mentions back-to-front strip, and a server-boot hang behind a stalled Feishu endpoint.

Rebased with `--onto origin/main c872c2e7` instead, dropping those commits outright — which is "take main's side" mechanized. Zero conflicts, and the diff against `main` is exactly this PR's 15 files. Both hazards verified afterwards: `cutSpan` is gone from the tree, `stripBotMention` is a single leading-mention `find`, and `feishu-connector.ts` is byte-identical to `main`.

### 3. The changelog recorded reasoning and was written in the present tense

`## Not shown in the connection status` was an alternatives-considered section under a bespoke name, which `changelog/README.md` bans outright — deleted from both files and moved into the PR body, where it belongs. The prose was present tense throughout ("PenguinHarness cannot change that setting… What it can do is…", "the fold **gains** an entry"); rewritten past tense in both languages. One claim was also simply false — "any two of them leave the bot just as mute" — since making the bot an admin alone fixes it, and `/setprivacy` plus a re-add fixes it without admin.

### 4. Three smaller ones

The privacy toast was 241 characters carrying a three-step instruction, pushed as `info` with a fixed 4s duration and no hover-pause — roughly 13s of reading in a 4s window, stacked under the success toast. Cut to the diagnosis plus a pointer at the fold, which is this PR's own stated split; the durable version already lives there.

The zh fold entry read `三选一：…；或…` and then listed two options. Gone with the rewrite — both dictionaries now name the same two routes in the same order.

`MessagingAccountInfo`'s header still described a single-label payload that "returns null" when there is no cheap label, which stopped being true when `checkCredentials` began returning `{}`. Rewritten for a payload of independently optional members. `can_join_groups` was declared with a doc comment and read nowhere — deleted.

### 5. Tests covered the flag, not the new behaviour

The server side did exercise all three flag states, which is the right shape. But `FakeBotClient.getMe()` always set a username, so the one contract change — returning an object instead of `null` when there is none — was never exercised; and on the web, `testConnection` was stubbed to a noop, so nothing pinned that a `groupPrivacy: true` response actually produces a toast.

Both covered now. One deviation worth naming: `testConnection` genuinely cannot be driven in this suite — `vitest.config.ts` is `environment: "node"` by a documented decision, and `useMessagingBinding`'s `form` is filled by a `useEffect` that SSR never runs, so the call returns early. Rather than reverse that decision for one test, the notice selection was extracted into a pure `telegramTestNotices` classifier and tested by value — the shape `updateCheckOutcome` in `use-version-info.ts` already uses for exactly this problem.

A trap caught mid-flight, worth recording: `S` in web tests resolves to the **zh** dictionary, so a first assertion written against an English phrase passed vacuously. Only the mutation run exposed it. Rewritten to assert per dictionary — which is strictly better here, since finding 4 was about the two drifting apart.

### Checks

- Server **1129/1129**, web **1428/1428**, both typechecks clean, prettier clean, workspace build clean. Five mutations confirmed red then restored.
- Design: a companion PR is needed and I am writing it — `05-ARCHITECTURE.md:1275` (the connector seam's credential-check payload) and `:1282` (the binding editor's test line). `:1279` deliberately does **not** change: it documents the `connect()` sequence, and this PR touches only `checkCredentials`, which is the test endpoint's path. Grepping the specs finds no existing mention of any of these identifiers, so the lines are additive rather than corrective.

### Merge

#505 and #507 are stacked on this branch, so their diffs read doubled until they are rebased — that is expected and I am handling it.

PR #501, opened 2026-08-27
feat(server,web): bind a conversation to a QQ bot

hiyouga:
Deleting the data files and restarting brought the app back with the old Workspace still selected.

## Two mechanisms, one fixed

The Workspace survives a data-root wipe through two independent paths:

1. **Server side — deliberate, unchanged.** `ProjectService.provisionInitialProject` adopts an existing `default_project` directory without overwriting its config. Delete `web.db` but not the project directory and `.project_config.toml`'s `[default_chat].workspace` is adopted as-is. That is intended behaviour and stays.
2. **Browser side — fixed here.** The Workspace also lives in `localStorage`, which has no relationship to the data root at all: in the desktop app it sits in Electron's userData directory, so no amount of deleting `PENGUIN_HOME` touches it. The keys are `penguin.chatDraft.<userId>.<projectId>` (carrying `workspace`, `agentId`, approval mode, model selection and the unsent composer text) and `penguin.sidebarWorkspaces.<projectId>`, among ~26 others — and **both ids are compile-time constants** (`ADMIN_USER_ID = "admin"`, `DEFAULT_PROJECT_ID = "default_project"`). A wipe-and-restart re-provisions the same user and Project, the keys match again, and the old state returns.

Only (2) is addressed. (1) is a separate, intentional seam.

## What this adds

- **`<root>/install-id`** — an opaque `randomUUID` minted the first time a data root is used, next to the existing `api-token`. It is an **identity, not a credential**: it authorizes nothing, contains no host name, user name or timestamp, and gets ordinary file permissions rather than `0600`. Only an *absent* file counts as a new root; any other read failure returns "unknown" so a transient error can never masquerade as a wipe, and a root that cannot be written also returns "unknown" rather than a fresh id on every boot.
- **`GET /api/install`** — public, mounted in the **platform** (`createApp`, just above its auth gate, the shape #502 used for the OAuth callback). A hot push carries platform + cli + web dist as one version and never the runtime, so the route travels with the bundle that calls it; mounted in the runtime shell it would have been missing on exactly the installations that received this web bundle by push. It is deliberately absent from `RUNTIME_PREFIXES` for the same reason — the platform must serve it, not decline it. The data root is still the runtime's; `deps.config.root` reaches the platform either way, so ownership of the file and ownership of the route are different questions. Public is a requirement, not a convenience: the browser must ask before it knows whether anyone is signed in, and a just-wiped root has nobody signed in at all. The id is read per request, so a root deleted under a running server is noticed on the next page load rather than only after a restart.
- **`packages/web/src/lib/install-scope.ts`** — one stored `penguin.installId`, compared on boot; a mismatch sweeps the install-scoped keys.

### Where the sweep runs, and why that is early enough

In `main.tsx`, **before `createRoot`**. The state it may clear is read from `useState` initializers scattered through the tree (sidebar pins and order, the composer's draft), and those run during the first render — a sweep arriving one effect later would let a stale draft be read once and then deleted underneath the component holding it, which is worse than either doing nothing or doing it in time. Before `createRoot` there is provably no component to have read anything.

It costs one same-origin request to the server this page was just served by. It is bounded by a 3s timeout: it can delay first paint, never prevent it. Giving up resolves to "unknown", which changes nothing.

### The sweep alone is not enough: two live copies of what it removed

**Modules evaluated before it ran.** ES module evaluation of the whole static import graph finishes before `main.tsx` runs a statement, so `dock-state.ts` — which parses `penguin.dock.layout` into module state at module scope — is already holding a copy of keys the sweep then deletes, and the first route resolution's `setDockScope` writes the entire pre-wipe map back, reopening a bottom dock pointing at a shell that belonged to the deleted root. The id matches by then, so nothing ever sweeps it again.

A boot that actually swept therefore **reloads instead of mounting**; the second pass evaluates every module against the cleaned store and reconciles to "unchanged". Making `dock-state` lazy would fix `dock-state` and leave the class open for the next module written that way, and a sweep only happens when the data root was actually replaced. A sweep whose marker could not be written back is reported apart (`swept-unrecorded`) and mounts, so the reload cannot repeat forever.

**Other tabs.** The sweep is per page load, so a tab already open against the old root keeps that state in React and module memory and restores it on the next pin, reorder or draft keystroke. The cross-tab `storage` event for `penguin.installId` now sweeps and reloads that tab too — only for an id *replaced* by a different id, since a first recording is another tab's adoption and a removal is site data being cleared.

### One id plus a sweep, not the id in every key name

Key names stay as they are, so no module that builds a key changes and nothing on disk needs migrating. Embedding the id would also create orphans forever — a key carrying a dead install id is something nothing ever reads or deletes, and each wipe would leave another full set behind. Walking the store collects orphans from *every* earlier root in one pass.

## Key classification

`grep -rhoE "penguin\.[A-Za-z]+" packages/web/src` finds 28 storage keys. They are classified in `KEY_RULES`, each with a one-line reason next to it. Two strings that grep finds are deliberately excluded: `penguin.chatRouteApplied.*` is `sessionStorage` (scoped to one tab's history, cannot outlive a data root) and `penguin.ooo` is the product's domain inside an example URL in a doc comment.

**Kept (browser preferences — 12):** `theme`, `fontScale`, `accent`, `currency`, `terminal.theme`, `lang`, `sidebarCollapsed`, `panelWidth`, `sidebarGroupMode`, `sidebarSortMode`, `sidebarNavGroupCollapsed`, `steerMode`. None names anything on the server; a new root can honour every one of them exactly, and wiping a data root is not a request to reset someone's theme.

**Swept (install-scoped — 16):** `chatDraft.*`, `chatDrafts.*`, `sidebarWorkspaces.*`, `pinnedSessions.*`, `sessionOrder.*`, `sessionSeen.*`, `groupOrder.*`, `sidebarCollapsedGroups.*`, `sidebarPinnedGroups.*`, `lastProjectId`, `lastAgentId.*`, `memoryCollapsed.*`, `modelsExpandedGroups.*`, `modelsGroupOrder.*`, `dock.layout`, `terminal.page.id`. Each references Session ids, Agent ids, Workspace paths, Project ids, terminal ids or per-Project provider group names.

Rules are **exact key or dot-terminated family**, never a plain prefix, because two pairs would otherwise be misclassified: `penguin.sidebarCollapsed` (preference) is a prefix of `penguin.sidebarCollapsedGroups.<projectId>` (install-scoped), and `penguin.terminal.` covers both `penguin.terminal.theme` (preference) and `penguin.terminal.page.id` (a terminal id).

An **unclassified** key is left alone. Never deleting something we do not understand is the safe default; the cost is that a new install-scoped key must be added to the table — which is now **enforced rather than asked for**: a test reads `packages/web/src` off disk, collects every `penguin.*` key literal (block comments stripped, since prose names key prefixes and the product's domain) and fails on any the table does not cover, listing the file it came from. The two deliberate exclusions above are on an explicit list, and a second test fails if either stops appearing in the source.

### The arguable ones

- **`penguin.dock.layout` → swept.** The entry is overwhelmingly per-conversation tab arrangement keyed by Session id and holding terminal ids, but it also carries `bottomRatio`, a size preference, in the same JSON. Sweeping loses that number. Splitting the entry would mean a stored-format change and a migration for a value the user re-drags in a second, so the whole key goes.
- **`penguin.modelsExpandedGroups.*` / `penguin.modelsGroupOrder.*` → swept.** Most provider group ids come from the built-in catalog, which ships with the app, so the arrangement would still "work" against a new root. But the keys are per-Project and include user-defined group names that live in that Project's config on disk. Project-scoped state goes when the Project is new; the cost of being wrong here is that the models page returns to catalog order.
- **`penguin.sidebarGroupMode` / `sidebarSortMode` / `sidebarNavGroupCollapsed` → kept.** These are sidebar state, and the sidebar is otherwise install-scoped, which is what makes them arguable. They are also documented in-code as single global user preferences ("switching it anywhere switches it everywhere") and none of them names a server entity — they are the *mode*, not the groups or the order, both of which are separate keys and are swept.
- **`penguin.panelWidth` → kept.** Its own module says it outright: "Width is a layout preference, not session data: it is never reset on a Session switch."

## Upgrade behaviour, and what a user must still do by hand

A browser that holds keys but has never recorded an install id **cannot be told apart** from one whose root was wiped. It therefore adopts the current id and sweeps nothing — destroying legitimate state on upgrade would be a far worse bug than the one being fixed.

Stated plainly: **shipping this does not clean state that is already stale.** A user who wiped their data root before this release still sees the old Workspace, and the only way to clear it today is by hand — clearing site data for the app's origin, which also resets the preferences listed above. Only wipes from this release onward are handled automatically.

A "reset local UI state" action in settings would cover that gap in one click (it would call the same sweep with a forced mismatch, so it needs no new machinery beyond a button and its copy). Not built here — proposing it rather than adding unasked UI to a fix.

## Notes

- **Two roots, one origin** is the case that matters — the same host and port re-pointed at a different data root; different origins already have separate `localStorage`. A plain server restart on an unchanged root sweeps nothing, which is pinned both in the unit tests and in the manual run below.
- **Storage may throw or be empty.** Every read and write is wrapped and `localStorage` is resolved inside the `try` (the convention the existing storage modules already follow). A store that throws on every access resolves to "adopted" having stored nothing, destroys nothing, and lets boot proceed.
- `GET /api/install` is documented in `packages/docs/content/server-api.{en,zh}.md`.

## Verification

From the worktree root, on Node v24.18.0:

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (8 packages)
- `pnpm test` — pass: core 1014 passed/5 skipped, server 1085, web 1376, cli 374, desktop 144, docs 53, landing 69, skills 27. Includes the two new files: `packages/server/test/install-id.test.ts` (12) and `packages/web/test/install-scope.test.ts` (16).

By hand, in a real browser (headless Chromium with a persistent profile, the built server serving the built web dist on one `localhost:7371` throughout, isolated root `~/.penguin/dev-data-installid`, since removed):

```
== phase 1: fresh root A ==
  install id A = be3eba7f-10a1-4268-9c6a-9b9a784e4e04
  <root>/install-id = be3eba7f-10a1-4268-9c6a-9b9a784e4e04
  localStorage after using the app on root A:
    penguin.installId = be3eba7f-10a1-4268-9c6a-9b9a784e4e04
    penguin.dock.layout = {"scopes":{},"bottomRatio":0.4}
    penguin.theme = dark
    penguin.chatDraft.admin.default_project = {"text":"","workspace":"","approvalMode":"allow-all","agentId":"default_agent","modelRef":…
== restart on the SAME root (must sweep nothing) ==
  install id after restart = be3eba7f-10a1-4268-9c6a-9b9a784e4e04 (same: true)
  every key survived the restart: true
== delete the data root, restart on the same host:port ==
  install id B = f272227f-6f13-44fe-9d17-dd48b55667ee (different: true)
  localStorage after reload on root B:
    penguin.theme = dark
    penguin.installId = f272227f-6f13-44fe-9d17-dd48b55667ee
```

The draft and the dock arrangement are gone, the theme survived, and the app lands on the login page — which is what a wiped root should look like.


hiyouga:
Follow-up commit `236ac854`: the two QQ setup links now point at pages that resolve. Every URL in `CHANNEL_LINKS` was checked over the wire (no proxy, `curl -sSL`, status without redirect / after redirect / body size):

| URL | raw | final | bytes | verdict |
| `https://q.qq.com/qqbot/dashboard/` | 200 | 200 | 11161 | **new console** |
| `https://q.qq.com/qqbot/dashboard` | 404 | 404 | 0 | trailing slash is load-bearing |
| `https://q.qq.com/qqbot/#/developer/developer-setting` | 200 | 200 | 700 | **old console** — the fragment is never sent, so this landed on the shell |
| `https://q.qq.com/qqbot/` | 200 | 200 | 700 | shell, confirming the above |
| `https://bot.q.qq.com/wiki/` | 200 | 200 | 49090 | **new tutorial** |
| `https://bot.q.qq.com/wiki/develop/api-v2/` | 200 | 200 | 24658 | old tutorial — resolves, but it is the API reference |
| `https://open.feishu.cn/document/develop-an-echo-bot/introduction` | 200 | 200 | 7464 | unchanged, healthy |
| `https://open.feishu.cn/app` | 200 | 200 | 9649 | unchanged, healthy |
| `https://core.telegram.org/bots/tutorial` | 200 | 200 | 43985 | unchanged, healthy |
| `https://core.telegram.org/bots/api` | 200 | 200 | 860076 | unchanged, healthy — see note below |

On the tutorial target: the wiki **root** turned out to be the onboarding guide itself — its `<title>` is `介绍与接入指南 | QQ 机器人官方文档` and its section anchors are `#_4-创建应用`, `#_7-1-配置沙箱环境`, `#_8-开发基础设置`, `#_8-1-机器人-appid`, with the body naming `AppID` / `Token` / `AppSecret` directly. That is exactly what a reader of the setup fold is after, so the root is the right target and no deeper link is needed; `/develop/api-v2/` stays out of the UI as a reference that answers a question the reader is not asking yet.

Both traps are pinned by tests so they cannot be tidied away: the console assertion requires the trailing slash and explicitly rejects the slashless form.

**One finding I did not act on.** Telegram's `console` is `https://core.telegram.org/bots/api` — the API reference — while the corner link it renders is labelled "Open developer console" and hangs off the Bot Token field. Telegram has no web console; the token comes from `@BotFather` in the app (`https://t.me/BotFather` → 200, 11163 bytes). So the label and the target disagree. Nothing is broken, and it is another channel's pre-existing copy, so changing it here would put an unrelated behavioural change in a QQ PR — flagging it for a change of its own instead.

Re-verified after the commit: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (server 1109, web 1369, core 1013, cli 374, desktop 144, docs 53, landing 69, skills 27), `pnpm format` + `pnpm format:check` — all clean.

hiyouga:
## Review

Reviewed in a dedicated worktree, then rebased and fixed. A third channel is the real test of the connector seam, so that is where I started.

**The seam held.** `connector.ts` gains one optional `replyBudget`, `bridge.ts` one line, and Feishu and Telegram are byte-identical — both suites pass unchanged. The QQ work itself is well built: the budget/tail design is coherent, the passive-window reasoning is sound, credentials never reach a log or an error string, every OpenAPI call carries `AbortSignal.timeout`, the token cache collapses concurrent refreshes, and `openGateway` awaits nothing on the boot path. Against the four bugs this batch has just fixed in the sibling connectors, QQ repeats **one** — and that one was a blocker.

### 1. A redelivered `msg_id` reset the passive-reply ledger, refusing the rest of the answer

`noteInbound` unconditionally reset `msgId`/`receivedAt`/`spent`, with no check that the id was the one it was already anchored to — and the bridge's redelivery guard runs *inside* `handlers.onMessage`, strictly after. Same shape the design spec's inbound bullet forbids (watermark before the work), applied to the connector's own state.

Reproduced against the real `QQConnector`: fire `msg_1`, send two replies (`msg_seq` 1, 2), re-fire the same `msg_1` — which the platform documents it does to guarantee delivery, and this file's own comment says so — then send a third → `msgSeq` came out `[1, 2, 1]`. In production that third send carries a `(msg_id, msg_seq)` pair already spent, which QQ **refuses** as `40054005 消息被去重` rather than deduplicating. So a gateway resume replaying one message mid-run kills every remaining reply: the user gets half an answer, the chat goes quiet, and the bridge writes `messaging_send_failed`.

**Fixed** in 4ac59d44 — `noteInbound` returns early on the same anchor, before the timer and carry branches. Mutation reproduces `[1, 2, 1]` exactly.

### 2. `PUT /messaging/qq` skipped the save-path guard

The PR wired `guardEnable` into `POST /qq/state` but not the save half. `main` guards both, because an **enabled** binding re-pointed at another account restarts its connector and would stand two Sessions on one bot without ever passing the enable gate — the design spec states it outright (`上面那条重启例外绕不过这道闸门`). Verified: Session B PUTs Session A's App ID, the save succeeds, `sync()` restarts B's connector, and two gateway connections race on one event stream. **Fixed** — `guardAccountFree` on the QQ PUT, mirroring feishu and telegram, documented in both `server-api` files.

### 3. The per-chat ledger outlived the connection

`close()` cancelled the gateway only, leaving pending flush timers armed and the ledger map — keyed `appId:kind:openid` on the long-lived connector singleton — never released. Verified: spend the three immediate slots, withhold a fourth body, `close()`, wait → the withheld tail still reaches QQ, i.e. **a message arrives after the user turned the binding off**. Two more consequences from the same root: a stale tail carried onto the first message after a re-enable, and — because the key is the account, not the binding — a second Session holding a stale `lastChatId` could spend the live Session's budget. Plus an unbounded map of 20KB tails on a server that runs for months.

**Fixed** — `close()` drops that credential set's ledgers, and `ledgerFor` evicts expired ones before growing the map.

### 4. A failed flush dropped the whole withheld tail

`flush` cleared `ledger.tail` *before* awaiting the send, so a transient 5xx on the one message carrying the coalesced end of a long answer lost that end silently — contradicting the module doc's "nothing is ever dropped". Fixed by restoring the head on rejection with `unshift`, so a body withheld *during* the await keeps its place.

### 5. The gateway had no handshake or heartbeat-ACK watchdog

`startHeartbeat` fired on an interval and never checked that `HEARTBEAT_ACK` came back — the `default:` arm ignored it — and nothing covered "socket opened, HELLO never arrived". A half-open socket (NAT or proxy drop with no FIN, and this path deliberately inherits the admin proxy dispatcher) keeps `readyState === OPEN`, so heartbeats write into a dead pipe, no `close` fires, `onError` is never called, and the binding sits at `connected` receiving nothing, indefinitely.

**Fixed** — an ACK watchdog at two intervals of silence and a 30s handshake deadline, both calling a new `dropSocket` that detaches and calls `fail()` **directly** rather than waiting on a `close` event a half-open socket may never emit.

### 6. `qq-api.ts`'s production adapter was untested

Of 680 lines the tests reached two pure functions. New `messaging-qq-transport.test.ts` (14 tests) covers the token cache (reuse, in-flight collapse, quoted and absent `expires_in`, and a refused exchange asserted not to contain the secret), the OpenAPI send (endpoint per scene, payload shape, 401 replay-once, `err_code` over `code`), and the gateway over a fake socket (HELLO→IDENTIFY→READY, RESUME after 4009, fresh identify after 4006, INVALID_SESSION, RECONNECT, fatal 4914, both watchdogs).

Worth noting how: `vitest.config.ts` runs `isolate: false` and `test/isolation.test.ts` guards that no file mocks a module, so `vi.mock("undici")` was off the table. `createSocket` and `handshakeTimeoutMs` were added to `QQTransportOpts` as explicit "Test hook:" seams, matching the existing `gatewayRetryMs`.

### On the mention-normalization question

I asked whether #497's defect class applies here. **It does not, and there is nothing to fix.** On `GROUP_AT_MESSAGE_CREATE` the platform strips the bot's own `@` prefix from `content` server-side, leaving only the whitespace it sat in — which is why the connector's entire mention handling is one `trim()`. The dispatch frame carries no mentions array and no placeholder scheme. So both halves of #497 are structurally absent: no placeholder to resolve, and no own-handle to cut. A third party's mention is ordinary content and reaches the model verbatim, which is the same policy #497 settled on for a non-prefix mention.

That claim is now pinned rather than assumed — a test asserts the trim is the only edit (a third-party token passes through byte-for-byte) and that a message left with nothing but the stripped mention reads as `""`, which the bridge answers with the text-only notice. **One honest gap:** the literal shape QQ uses for a third-party mention in group v2 `content` is stated nowhere in the code, comments or design docs, so the test uses an opaque token rather than assert a format I could not source.

### The rebase

`main` moved three times during this work (#493, #506, #499 all overlap this diff), so the branch chased the tip rather than sit on a base that conflicts. Beyond the predicted conflicts, `app.ts`, `bridge.ts`, `messaging-binding-editor.tsx`, both dictionaries and the editor test all needed resolving — `main`'s flattened, paced `deliverReply` kept, with the `replyBudget` clamp re-applied on top. The "six messages → four sends" test passes against that rewritten shape unchanged, and `git diff origin/main -- changelog/unreleased/` shows exactly the two new QQ files.

### Still open before merge

**Design.** There is no QQ section in `design/specs/` at all — a whole new channel is unambiguously spec material. Note two things are new *normative* behaviour with no counterpart today: the ledger's connection-scoped lifetime, and the two gateway watchdogs. (`05-ARCHITECTURE.md:1278` needs no edit — the code was the side out of line, and finding 2 brings it into conformance.) I am writing that.

### Checks

- Whole monorepo: **server 1178**, **web 1435**, core 1021 (+5 skipped), cli 374, desktop 144, docs 53, landing 69, skills 27 — all green. Typecheck clean across all 8 projects, prettier clean, build clean. Every mutation confirmed red then restored.

hiyouga:
⚠️ **This PR's description is not its own** — and it reached `main`.

The body on this PR is #508's ("Deleting the data files and restarting brought the app back with the old Workspace still selected…" — the install-scoped browser state work). Nothing in it describes the QQ channel. Because this was squash-merged, the same text is now the commit body of `483af5db` on `main`:

```
feat(server,web): bind a conversation to a QQ bot (#501)

Deleting the data files and restarting brought the app back with the old
Workspace still selected.
...
```

That matters more than a cosmetic slip. `changelog/README.md` deliberately keeps reasoning off disk on the grounds that "an agent that wants it pulls the PR description, `git blame`, or `git log`" — so for this change, all three of those now lead somewhere wrong, and the changelog entry (correctly) does not carry the reasoning either.

I have not touched it. Rewriting the body is yours to do, and the commit message on `main` cannot be corrected without rewriting published history, which is not something to do for this.

For what it is worth, the review of what this PR *actually* does is in [the review comment above](https://github.com/Prism-Shadow/penguin-harness/pull/501#issuecomment-5438069224), and design PR [penguin-harness-design#70](https://github.com/Prism-Shadow/penguin-harness-design/pull/70) now records the QQ channel in the specs — those two are the durable account of it until the body is fixed.

Everything about QQ in that review and in the design text was read from the merged source on `origin/main`, not from this description.

PR #502, opened 2026-08-27
fix(server): receive the provider key-minting redirect without a session, so it works in the desktop app

hiyouga:
TokenDance's **Authorize key** action fails in the desktop app with a bare

```json
{"error":{"code":"unauthorized","message":"Not signed in or the sign-in has expired."}}
```

and works in the browser. This makes it work in both.

## The asymmetry, and why it existed

1. `models-page.tsx` opens the provider's authorization page with `window.open(flow.authorizeUrl, …)`.
2. In the desktop app that hits `setWindowOpenHandler` in `packages/desktop/src/main.ts`. The target is not `isLocalSurfaceUrl`, so the shell answers `shell.openExternal(target)` + `{ action: "deny" }` — the authorization page opens in the **system browser**.
3. The provider redirects that browser to `GET /api/projects/:projectId/model-oauth/callback?flow=…&code=…`. It can reach it at all because the desktop window's origin is `http://localhost:<port>` (`main.ts` boot / `server-process.ts` `appOriginFor`), and the callback URL is derived from the `/start` request's own origin — so the redirect target is a loopback URL any browser on that machine can open.
4. The system browser is not the Electron window. It holds no `penguin_session` cookie for that origin, `authMiddleware` at the `/api/*` gate throws before the handler runs, and the message above is that middleware's own string, verbatim.

In the browser the popup is a tab of the same browser that holds the session, so it authenticates and the flow completes. That is the entire difference.

Reproduced end to end, both in-process (`app.request` with no cookie → 401 with exactly that body) and against a real bind on an isolated data root.

## The fix: the receiver takes the code, the owner's poll spends it

A loopback OAuth redirect receiver cannot require a session — the browser that arrives is, by construction, not necessarily the one that started the flow. So `GET …/model-oauth/callback` is mounted ahead of the auth gate in `createApp` and authorizes on the flow id, which was already a capability rather than a name:

- `randomBytes(32)` base64url, minted server-side, never guessed and never enumerable;
- bound in `ModelOAuthService` to a user, a Project, a provider, and a PKCE verifier that never leaves the process;
- 10-minute TTL, and single-use.

**But the route mounted outside the gate does not get to write a credential.** The redemption is split in two:

- `GET /callback` (no session) calls `ModelOAuthService.deposit`, which stores the code on the flow and leaves it `pending`. It reaches no provider and touches no Project. The page it answers says "Authorization received — return to PenguinHarness".
- `GET /:flowId` (owner-only, behind the gate) calls `poll`, which redeems any deposited code — `exchangeCode` + `applyGroupKey` — and reports the outcome. This is the route the dialog already hits every `OAUTH_POLL_MS` while it waits, so the desktop flow still completes hands-free and the provider's one-time code is still redeemed within seconds of the redirect.

The state machine that introduces:

| Flow state | `GET /:flowId` answers | Dialog renders |
| --- | --- | --- |
| opened, no redirect yet | `{status: pending}` | spinner, "waiting…" |
| code deposited | redeems inline, then `{status: done}` (or `{status: error, error}`) | toast + close, or the localized failure |
| code deposited, a concurrent poll is already redeeming it | `{status: pending}` | spinner — the next tick reports the outcome |
| exchange failed | `{status: error, error}` | localized failure, **Retry** opens a new flow |
| expired / unknown / not yours | 404 | `oauthTimedOut`, failure phase |

No Web App change is needed: `pending` is the state it already renders as the spinner, and a failure now reaches it as `{status: error, error}` instead of being displayed on the redirect page in the other browser — which on the desktop is a browser the user may not even be looking at. A failed exchange leaves the flow ended, not wedged. `applied` is set on the redeeming poll alone, so `credentials_updated` is published exactly once.

Single use survives the move and gains a second half. A flow has **one deposit slot** — a second redirect neither overwrites the code nor re-arms the exchange — and `redeem` nulls both the verifier and the code **before its first `await`**, so two redirects racing deposit once and two polls racing redeem once. The TTL bounds the deposited-but-unexchanged state exactly as it bounded the verifier.

Two more narrowings on the exempt route:

- **`HEAD` answers 405.** Hono re-dispatches `HEAD` as `GET` before routing (`hono-base.js`: `if (method === "HEAD") return new Response(null, await this.#dispatch(request, …, "GET"))`), so `HEAD …/callback?flow=…&code=…` reached the handler and spent the flow — on a method HTTP requires to be safe. The re-dispatch keeps the original `Request`, so `c.req.method` still reads `"HEAD"` inside the handler.
- **A `mode: "manual"` flow is refused.** Manual mode exists so that no callback URL is ever handed out; the flow record did not store the mode, so a deployment that deliberately chose it had gained an unauthenticated path it had opted out of. `Flow` now records `mode`, and `deposit` 404s a manual flow — the same answer an unknown flow gets.

The exemption is as narrow as the router allows — exactly this literal path, exactly `GET`:

| Request, no cookie | Before | After |
| --- | --- | --- |
| `GET …/model-oauth/callback` | 401 | reaches the handler (deposits only) |
| `HEAD …/model-oauth/callback` | 401 | 405 |
| `GET …/model-oauth/callback/extra` | 401 | 401 |
| `POST …/model-oauth/callback` | 401 | 401 |
| `POST …/model-oauth/start` | 401 | 401 |
| `GET …/model-oauth/:flowId` | 401 | 401 |
| `POST …/model-oauth/:flowId/code` | 401 | 401 |

This follows the precedent already in `app.ts` — `/api/desktop/update` mounts its own `authMiddleware` on its own subtree rather than being listed as an exception inside the middleware, and the desktop one-shot `/api/auth/desktop-login` sits outside it entirely. No path-exemption list was added to `authMiddleware`; the ordering *is* the statement. Registering the literal path first also preserves (and now pins) the property that `:flowId` can never swallow `callback`.

## Threat analysis

**Correction to this PR's first revision.** It argued the exemption cost nothing because "local filesystem access to the data root *is* admin authority", so a local attacker already had better options. That is false for a **different unprivileged OS user on the same machine**. `<root>/api-token` is `0600`, so they have no admin path — but the flow id is world-readable to them: `shell.openExternal` on Linux dispatches `xdg-open <url>`, and the authorize URL carries `callback_url=http://localhost:<port>/api/projects/<pid>/model-oauth/callback?flow=<id>` in that process's argv. The consequence was also worse than the "billing misattribution" the first revision claimed: the key written into the victim's Project belongs to the **attacker**, so the victim's traffic runs on the attacker's provider account — visible in their console, revocable at will. Verified against the first revision: a request with no cookie, and a request carrying a *different signed-in user's* cookie, both returned 200 "API key created" and wrote the key.

**What the unauthenticated route can do as it now stands.** Deposit one code on one pending, unexpired, callback-mode flow whose id the caller already knows, having beaten the provider's own redirect to that flow's single deposit slot. That deposit is inert: no provider is contacted, nothing is stored, and nothing about the Project changes. It becomes a key only if the flow's own owner, in their own session, polls that flow — and the same poll is what would have redeemed the legitimate code. So the worst case is no longer "an attacker writes their key into your Project"; it is "an attacker who won that race gets *your dialog* to redeem *their* code instead of yours, in the seconds while you are looking at it".

**The residue, stated plainly.** That race is not zero. If the attacker reads the flow id out of argv the moment the user clicks **Authorize**, and returns a code minted against that flow's challenge before the provider's redirect arrives, the owner's next poll redeems it. What the split buys is that nothing can land without the owner's session acting, that a closed dialog (or a user who never presses the button) is unreachable, and that the route sitting outside the auth gate holds no write authority at all — which is the property that has to be true for it to sit there.

**The rest of the envelope is unchanged:** 256 bits from `randomBytes`, not guessable and not enumerable; the same generic HTML page for unknown, expired, foreign-Project, manual-mode and already-used, so a flow id is not a probe for what exists; ten minutes; one deposit; in-memory and per-App, so a restart drops it; and no key, code or verifier ever appears in a response, a log line or a URL.

**The alternatives, for the record.**

- *Open the authorization page in a window carrying the app's session.* Worse: it embeds a third-party sign-in page inside the app's own session partition, and providers commonly refuse embedded views. It also puts the fix in `main.ts`, the least-updatable code in the system — reaching users only through a new installer, where this fix reaches them by push.
- *Use the flow's existing `mode: "manual"`.* It stays exactly as it is, and is now protected from the receiver: it remains the right answer for deployments the redirect genuinely cannot reach. It is not the right default, because the desktop path would keep failing first and the user would have to discover the switch after seeing a raw `unauthorized`.

## What was deliberately not changed

- **The desktop shell.** No change to `main.ts`, `setWindowOpenHandler` or `isLocalSurfaceUrl`. Sending the authorization page to the system browser is correct; the callback was the broken half.
- **The Web App.** No change to the dialog, its poll, the mode default, or the manual-mode escape hatch — the restructure was designed to land inside states it already renders.
- **The API types.** `ModelOAuthStatusResponse` is byte-identical; the extra `applied` the service returns to the poll route is consumed there and not put on the wire.
- **`/start`, `/:flowId/code`, and the status route's gate.** Still owner-only, still behind `authMiddleware`, asserted as such.
- **`authMiddleware`.** No exemption list; the mount order carries the exemption.

`ModelOAuthService.providerOf()` — dead code with a false comment, three lines above the signature this diff edits — **was** deleted, unlike in the first revision.

Nothing on disk changes shape — no stored Trace, config, or credential format is touched, so there is no compatibility entry.

## Verification

From the worktree root, all passing:

- `npx prettier --check` on all ten changed files — "All matched files use Prettier code style!".
- `packages/server`: `npx vitest run test/model-oauth.test.ts` — **42 passed**.
- `packages/server`: `npx vitest run test/authz.test.ts test/auth.test.ts test/api-token.test.ts test/isolation.test.ts test/hmr-http-seam.test.ts test/id-validation.test.ts test/errors.test.ts` — 7 files, **104 passed**.
- `packages/server`: `npx vitest run` — 92 files, **1090 passed**.
- `packages/server`: `npx tsc --noEmit -p tsconfig.json` — clean.
- `packages/web`: `npx tsc --noEmit -p tsconfig.json` — clean (no web change was needed).
- `packages/web`: the five suites that touch the models page (`protocol-detect`, `models-input`, `messaging-binding-editor`, `catalog-sync`, `group-import`) — **77 passed**.
- `packages/docs`: `npx vitest run` — 7 files, **53 passed**.

**Mutation-tested**, one targeted source mutation at a time, each restored afterwards:

| Mutation | Result |
| --- | --- |
| `deposit` redeems immediately (`void this.redeem(flow, input.code)`) | **13 failed** / 29 passed |
| HEAD guard removed | **1 failed** / 41 passed |
| manual-mode rejection removed | **2 failed** / 40 passed |
| deposit slot overwritable (drop `flow.code !== null`) | **3 failed** / 39 passed |
| verifier/code nulled *after* `await exchangeCode` | **1 failed** / 41 passed |
| `credentialsChanged` dropped from the poll route | **1 failed** / 41 passed |

New cases pin: the receiver depositing without redeeming; a different signed-in Project member depositing but being unable to poll (403) or land a key; two callbacks racing (one 200, one 400, then one exchange); two polls racing on a deposited code (one exchange); `HEAD` → 405 leaving the flow unspent; a manual flow refused while its paste box still works; and a deposited code expiring with its flow. The pre-existing cases for unknown flow, foreign Project, TTL, replay and path/method scope are all retained.


hiyouga:
## Review

Reviewed as an authorization review rather than a style one, since the PR deliberately removes an authentication boundary.

The diagnosis is right and the mechanics were sound: `appOriginFor` is always `http://localhost:<port>`, so the callback lands on the canonical App host and the preview-host guard still 401s `127.0.0.1`; the mount order really does confine the exemption to that one path (verified — trailing slash, uppercase, longer path, POST and all three siblings still 401); and the single-use guard was genuinely atomic, `flow.verifier = null` executing before the first `await`.

What was under-weighed is what the flow id alone buys, and where that id is readable.

### 1. The trade rested on a premise that does not hold for a different local user

The argument was that the exemption costs nothing because "local filesystem access to the data root *is* admin authority". `<root>/api-token` is `0600`, so a different unprivileged local user has no admin path — but the flow id is world-readable to them, because `shell.openExternal` on Linux dispatches `xdg-open <url>` and the authorize URL carries `callback_url=…/model-oauth/callback?flow=<id>` in argv. The consequence is also worse than billing misattribution: the key written into the victim's Project belongs to the *attacker*, so the victim's traffic runs on the attacker's provider account — visible in their console, revocable at will.

Shared Linux host, and the repo ships AppImage and deb. Alice clicks Authorize key; Mallory runs `grep -a <provider-host> /proc/*/cmdline`, reads Alice's port, project id and flow id, walks the authorization page as herself, and issues the callback against Alice's loopback — which is not per-user. I verified the server-side half: no cookie, and a *different* signed-in user's cookie, both returned 200 "API key created" and wrote the key.

**Fixed** in d5798572, with the deposit-only design: `GET /callback` now only stores the code (`deposit()`), leaves the flow `pending`, and answers "Authorization received — return to PenguinHarness". It reaches no provider and touches no Project. The owner's existing authenticated poll redeems it inline. An attacker who knows a flow id can now at most cause the owner's own session to redeem a code — nothing lands without the owner.

The atomic-before-first-await discipline is preserved and now covers both secrets: `redeem` claims the flow with `flow.verifier = null; flow.code = null;` before its first `await`. Mutation-verified — moving the claim after `exchangeCode` turns a test red, and `deposit` also calling `redeem` turns 13 red.

The dialog needed no change; `pending` already renders as the spinner, so the wire shape, `ModelOAuthStatusResponse` and both string catalogs are byte-identical.

### 2. "Exactly this path and exactly `GET`" was false — HEAD reached the handler and spent the flow

Hono re-dispatches HEAD as GET before routing, so `HEAD …/callback?flow=<id>&code=abc` with no cookie returned 200, incremented the exchange counter and spent the flow — on a method HTTP requires to be safe. The new test asserted POST → 401 and stopped there. **Fixed**: the handler rejects HEAD with 405 as its first line (the re-dispatch keeps the original Request, so `c.req.method` is still `"HEAD"`), and the claim was wrong in six places — `app.ts`, the route doc, both changelog files, both `server-api` docs and the PR table. All corrected.

### 3. A `mode: "manual"` flow was redeemable through the unauthenticated receiver

Manual mode exists so no callback URL is ever handed out — the docs sell it as the answer "for deployments the redirect cannot reach" — but `Flow` recorded no mode, so the callback could not tell one flow from the other. Verified: `POST /start {mode:"manual"}` then a cookie-less callback → 200, one exchange. Before this PR that was a 401, so a deployment that deliberately opted out had gained an unauthenticated path. **Fixed** — `mode` is on `Flow`, and `deposit` 404s a manual flow with the same message as an unknown one.

### 4. The user-facing docs contradicted the new behaviour

`server-api.{en,zh}.md` was updated but `models.{en,zh}.md` — the page users actually read for this feature — was not. Line 106 still said only the owner can *finish* an authorization; line 104 still named "a desktop window that hands external links to the system browser" as a case where the redirect cannot come back, i.e. told desktop users to switch to manual mode for a flow this PR fixes. **Fixed** in both languages, and 106 now states it accurately: the owner starts it, the owner's own session finishes it, and the redirect is received without a session but only hands the code over.

### 5. Two smaller ones

The abuse tests covered unknown flow, foreign Project, expiry, sequential replay and path/method scope, but missed the two cases carrying the safety argument — concurrent callbacks on one flow id, and HEAD. Both added, along with concurrent polls, manual mode, a different signed-in user (a Project *member* is the sharper case, since a non-member gets 404 from `requireProjectOwner`), the deposit TTL and the deposit slot. And `providerOf` was dead code with a false comment three lines above the signature this diff edits — deleted, along with `FAILURE_TEXT`, which became unreachable once the exchange left the callback.

The changelog also asserted current codebase state in present tense, which `changelog/README.md` bans; rewritten past tense in both languages.

### Still open before merge

**Design.** Nothing in the specs is contradicted — model-oauth appears nowhere in `design/` and `## API Server` is an empty placeholder — but the specs do maintain an explicit security-invariant list for the one existing unauthenticated surface (`/preview`, "the token is the credential"), and `06-PROTOTYPE.md:210` states the desktop rule *with its rationale*: app-origin new-window requests stay inside the app because 交给系统浏览器只会得到 401. This PR creates the first app-origin URL the system browser is expected to reach and succeed on, which amends that stated rule. A companion design PR should add the second capability-authorized surface to the invariant list and amend that desktop rule with its exception.

Also worth knowing: open PR #455 (a Security Model page) will need this route listed if it lands after.

### Checks

- `model-oauth.test.ts` 42/42; the seven auth-adjacent suites 104/104; full server suite 1090/1090; server and web typecheck clean; docs 53/53; prettier clean. CI on the pushed head: 21 success, 1 skipped.

PR #503, opened 2026-08-27
Render LaTeX math on every Markdown surface

hiyouga:
Formulas now render. A pasted reply whose formula was
`\[\text{采集} \rightarrow \text{解码} \rightarrow \text{干预}\]` came through as prose littered with
backslash commands; it now renders as centred display math, CJK included.

Math is added to the plugin lists all Markdown surfaces share, so chat messages, Trace events,
benchmark cases and Workspace file previews gained it in one change. Those lists moved from
`remark-autolink-boundary.ts` (which is a plugin, not a pipeline) to a new
`packages/web/src/lib/markdown-plugins.ts`, now carrying a `REHYPE_PLUGINS` stage as well — the
five `ReactMarkdown` call sites take both from there rather than assembling anything themselves.

## Delimiters, and why `\(`/`\[` needed a syntax extension

Three delimiter pairs are supported: `\[…\]` and `$$…$$` render as display math, `\(…\)`
inline. Single `$` is deliberately off (below). `remark-math` brings the dollar form; the TeX
bracket forms are new (`packages/web/src/lib/remark-math-brackets.ts`).

`remark-math` decides display-versus-inline for `$$` by run length, not by TeX's meaning: only the
fenced form — `$$` alone on its own lines — is display, so `$$E=mc^2$$` on one line and
`text $$E=mc^2$$ text` both came out as *inline* math. That is nowrap, typeset cramped (small
integrals, squeezed fractions) and unlike `\[…\]` for content nobody writes meaning inline. A
`$$` pair has been display since plain TeX and single-dollar math is off here, so a twelve-line
remark plugin (`packages/web/src/lib/remark-math-dollars.ts`) re-classes every dollar-delimited
text-math node to display. The delimiter is read back from the source rather than from the class
list, because `remark-math` and `remarkMathBrackets` both produce `inlineMath` nodes.

They could not be done as an mdast pass. By CommonMark a backslash before ASCII punctuation is a
character escape, so `\[x\]` has *already* become the literal text `[x]` before any tree exists —
the backslashes are gone, and nothing separates it from a bracket someone typed. Rewriting the
source string before parsing is the other shortcut and is worse: to avoid corrupting code it would
have to re-derive which spans are fences, indented blocks and code spans of arbitrary backtick
runs, i.e. reimplement CommonMark on the streaming path.

A micromark construct removes both problems by construction. It is only consulted where inline
content is read, so a `\[` inside a fenced block or a code span is never offered to it. And because
`micromark-util-combine-extensions` splices extension constructs ahead of the built-ins at the same
character, it gets first refusal on every `\` and falls back to `characterEscape` when it declines —
which is exactly what makes a half-arrived `\[` mid-stream stay ordinary text and settle into math
when its closer lands.

Two things the implementation had to get right, both found by testing rather than reasoning:

- **Line endings must be their own tokens.** `micromark-util-subtokenize` stitches an inline token
  stream back onto the per-line chunks a paragraph was split into by counting void tokens that span
  a break. One value token swallowing two newlines reports one boundary where there are two, and the
  stitch throws `RangeError: Cannot access index …`. A crash, not a misrender.
- **No token may be closed empty.** micromark's development build (the one vitest resolves) asserts
  it, so `\[\]` and `\[\n…` aborted the parse until the tokenizer stopped opening a value token
  before it had a character to put in it.

Display math is emitted as a `<span class="math math-display">`, not the `<pre><code>` pair
`remark-math` builds for `$$`. `\[…\]` is phrasing-level syntax that can appear mid-sentence, and a
`<div>` or `<pre>` inside the surrounding `<p>` would be invalid nesting; KaTeX's own
`.katex-display` wrapper supplies the block layout either way.

An unclosed opener used to scan the rest of its paragraph, once per opener — 1000 stray `\[` in an
18k-character paragraph measured at ~520ms of parse, and chat bodies re-parse as they stream. Since
a scan that reaches the end proves no closer exists after *any* later opener of the same kind, the
tokenizer records that and short-circuits: same input now ~6ms, in line with GFM's own baseline.

## Single `$`: disabled, with evidence

`remark-math` rejects a span whose content starts or ends with whitespace, which saves a lone `$5`.
It does not save a pair, and agent transcripts are full of pairs. Measured through this repo's own
pipeline with single-dollar math enabled:

| input | rendered as |
| --- | --- |
| `Set $PATH and $HOME before running.` | `Set `**`PATH and `**`HOME before running.` |
| `It costs $5 and $10 in total.` | `It costs `**`5 and `**`10 in total.` |
| `The range is $5-$10 per seat.` | `The range is `**`5-`**`10 per seat.` |
| `echo $PATH; echo $HOME` | `echo `**`PATH; echo `**`HOME` |
| `Prices: $1,200 and $3,400.` | `Prices: `**`1,200 and `**`3,400.` |

Every one loses its dollar signs, its spacing and its meaning. The failure in the other direction is
that someone writing `$x^2$` sees `$x^2$` — the source, legible, re-typable as `\(x^2\)` or `$$x^2$$`.
Corrupting prose that was already correct is the worse trade, so single `$` is off.

**Residual shapes that still misfire**, all requiring a literal `$$`:

- A doubled dollar in prose: `it costs $$5` next to another `$$` later in the same paragraph.
- Shell that reaches the renderer unfenced and uses `$$` — the PID (`kill $$`) or an escaped awk
  field (`awk '{print $$1}'` in a Makefile recipe). Inside a code span or fence these are safe; only
  bare in prose, and only in pairs, do they combine into math.
- A doubled `$$` inside a table cell or a heading, same rule.

Anything with a single `$` — `$PATH`, `$5`, `$5-$10`, `$HOME/.penguin` — is now inert, and that is
the shape this product actually produces.

## Malformed input

`rehype-katex` owns `throwOnError` and omits it from its options type: it renders once strictly and,
on any error, re-renders with `throwOnError: false`, falling back to a `.katex-error` span holding
the original source. That is the required degradation already, so there is nothing to override — a
model emitting `\text(采集)` or `\frac{1` gets a visible formula or visible source, never a blank
reply. What is set: `strict: "ignore"` (KaTeX's default warns to the console per CJK character in
sloppy `\text`, unactionable to anyone reading a transcript), `trust: false` restated as the
boundary it is (no `\href`, `\url`, `\includegraphics` from model output), and `maxSize` so
`\rule{9999em}{9999em}` cannot blow the message layout apart. `maxSize` is 5em — 80px at a 16px
root, which a message body absorbs; the 100em first written here still clamped to a 1600x1600 black
square, and no overflow rule helps when the element's own height is the problem.

`errorColor` is `currentColor` rather than KaTeX's `#cc0000`, which is an inline style and so cannot
adapt to the dark theme, where it lands under 4:1 on black. Failed expressions render as their own
source in the body colour, marked with a dotted underline, KaTeX's parse error in the `title`.

## CJK

KaTeX's faces carry no CJK glyphs. It tags every CJK run `cjk_fallback` and then defines nothing for
that class, so `\text{采集}` inherits `.katex { font: … KaTeX_Main, "Times New Roman", serif }` and
lands on whatever the browser picks last. `styles.css` points the class at the application's own
stack, which is now a named variable so the body rule and the math rule cannot drift apart.

Verified by rendering, not by reasoning: the sample page below was screenshotted in both themes
against the built CSS and font assets, and the computed `font-family` on `.cjk_fallback` read back
as the app stack with real glyph metrics.

## Bundle size

`packages/web/dist`, before and after:

| | before | after |
| --- | --- | --- |
| total | 3,892,732 B | 4,456,715 B (**+551 KiB, +14.5%**) |
| files | 34 | 53 |
| main JS (gzip) | 384.47 kB | 468.16 kB (**+83.7 kB**) |
| CSS | 93,040 B | 120,270 B |
| fonts | — | 256,168 B across 19 woff2 |

KaTeX lists woff2, woff and truetype in every `@font-face`, and Vite emits an asset per URL, so a
verbatim import cost **+1,383,266 B and 93 files**. A ten-line `enforce: "pre"` Vite plugin drops
the two fallback formats, which no browser this SPA runs in has needed since 2016 and which the
Electron shell certainly does not; that is the +551 KiB above. Fonts are fetched lazily by face —
the whole sample page pulled three of the nineteen.

## Streaming: the rehype stage waits for the settle

`md.tsx` memoizes by text identity, so a settled message is skipped entirely — but the message that
is *streaming* re-parses on every delta, roughly eight times a second, and the rehype stage was
unconditionally inside that loop.

KaTeX itself is cheap (~0.3ms for a typical formula). The stage around it is not: `rehype-katex`
re-parses KaTeX's ~2.5KB of markup per formula back into hast, and React then builds several
hundred elements from it. Measured through `<Md>` with `renderToStaticMarkup`, medians:

| | before | after |
| --- | --- | --- |
| 3.6KB reply, 60 formulas, one delta | 178ms | 6.7ms |
| the same reply streamed in 40 deltas | 2571ms | 112ms |
| 736B reply, 16 formulas, one delta | 17.5ms | 1.6ms |
| one 16KB formula, one delta | 907ms | 4.1ms |
| 20KB paragraph wrapped by a stray `\[ … \]` | 857ms | 4.1ms |

So `rehypePlugins={streaming ? NO_REHYPE_PLUGINS : REHYPE_PLUGINS}`, two module constants flipped
once on settle — the same shape as `STREAMING_COMPONENTS`/`SETTLED_COMPONENTS` and
`highlight={!streaming}` next to it, and the same reason those exist. The settle render still pays
the full ~110-180ms for that 60-formula reply, once. Until then a formula shows its own TeX source.

Memoizing the produced hast per formula in a bounded `Map` was prototyped and measured as the
alternative, and it is not enough: it cannot help the formula that is still growing — which is the
one being re-rendered — and it left 1853ms of the 2352ms in place on the streamed reply. It also
wants `unist-util-visit-parents` and `hast-util-to-text` as new direct dependencies, or a
hand-rolled copy of `rehype-katex`'s element-selection rule.

One consequence needed handling: with no rehype stage, `remark-math`'s block form reaches the chat
renderer as the `<pre><code class="language-math">` pair it really is, which `MdPre` would turn into
CodeBlock chrome — a language label and a copy button appearing for the length of the reply and
then vanishing. `MdPre` now renders a `language-math` fence as a plain `<pre>`, so the settle
changes the typesetting and nothing else.

## Wide formulas

`.md-body .katex-display` already scrolls inside its own block. Inline math had no such guard, and
KaTeX sets `.katex .base { white-space: nowrap }`, so a formula with no top-level break opportunity
— a long `\sqrt{…}`, a long `\text{…}` — was one unbreakable run. Measured at 760px with a 640px
column: the paragraph rendered 1421px wide and the page scrolled sideways to 1514px.
`.md-body :not(.katex-display) > .katex` now gets `display: inline-block; max-width: 100%;
overflow-x: auto; vertical-align: middle`, after which the page stays at 760px and the formula's
1421px scrolls inside its own 574px box.

`vertical-align: middle` is load-bearing rather than decorative: an inline-block whose overflow is
not `visible` puts its bottom margin edge on the baseline, which floats every inline formula above
the line it sits in. Screenshotted side by side in both themes — with it, the sentence is
indistinguishable from before the guard; without it, `E=mc^2`, `\frac{a}{b}` and `x_{ij}` all
hang. No colour is introduced, so nothing here needs a tone token.

## Left out

- A `\[…\]` broken by a blank line spans two paragraphs and is not matched. A blank line inside
  display math is not valid input.
- `\[a\\]` — a TeX line break butted straight against the closer — closes at the second backslash.
  Ambiguous in TeX too.
- **CommonMark's escaped brackets are now math.** `\[` and `\]` are also the documented escape for
  a literal bracket, so `See \[1\] and \[2\] for details.` renders as `See` + a centred `1` +
  `and` + a centred `2` + `for details.`, and likewise `Write \[TODO\] in the file.`,
  `Use \(foo\) to capture a group.`, `C:\Users\(test\)\file`. Two syntaxes over one character
  and only one can win; it goes to the one models actually emit, thousands of times a day, against
  an escape almost nobody reaches for in a chat message. Pinned by a test so the trade stays on the
  record. The same collision reaches `$$` prose: `it costs $$5` paired with another `$$` later in
  the paragraph, or an unfenced `kill $$` / `awk '{print $$1}'`, is now a centred block rather than
  garbled inline math — a more visible misfire of an already-broken case.
- **A backtick between the delimiters declines the formula.** The tokenizer is only entered where
  inline content is read, so an opener inside a code span is never offered to it — but a *closer*
  inside one is just more raw characters. `In LaTeX you open with \( and close with` a code-spanned
  `\)` used to find that closer, destroy the code span and strand a backtick in the prose. It now
  bails at the backtick, which costs the rare formula that contains one (TeX's open-quote pair
  inside a `\text{}`); whether the run has a partner later is not checked, because preferring prose
  to math there is the safe direction.
- A ` ```math ` fence renders as math rather than as code. That is `rehype-katex`'s documented
  opt-in and it shares the `language-math` class with `$$` block math, so it cannot be disabled
  without breaking the latter. Plain fences are unaffected.
- Existing content changes appearance: a Workspace `.md` preview or a stored Trace containing
  `\[…\]` used to render as literal brackets and now renders as a formula. No stored data or config
  is read or rewritten, so there is nothing to migrate.

## Verification

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (8 packages); `packages/web` alone reports no errors at all
- `pnpm -r test` — pass, 4142 tests across 8 packages, 0 failures (`packages/web`: 103 files, 1389
  tests, of which 29 are `test/markdown-math.test.ts`)
- `pnpm format:check` — "All matched files use Prettier code style!"
- `vite build` in `packages/web` — 19 woff2 assets emitted, no woff or truetype, and no `http(s)://`
  reference anywhere in the emitted CSS

Every behavioural change carries a test that was confirmed red with the source change reverted: the
rehype flip, the `language-math` fence guard, `$$…$$` as display, the backtick bail, and the
`maxSize` clamp.

The full web suite was additionally run twice under saturating CPU load to confirm the one timing
assertion in the new tests does not flake; it compares against a control render in the same process
rather than a wall clock, precisely so a loaded box moves both numbers together.

Beyond the automated tests, the user's arrow chain, a partial derivative, an integral, a 3x3 matrix,
an inline `E=mc^2`, a deliberately malformed formula, the mangled `\text(采集)` paste, `$PATH`/price
prose, inline code, a fenced block, a clamped `\rule`, a one-line `$$…$$` against its `\[…\]`
twin, and an inline formula too wide for its column were rendered through the real pipeline against
the built CSS and font assets and screenshotted in both light and dark themes. All legible; no
failed requests, no console output, no CDN reference anywhere in the built bundle; wide display math
and wide inline math each scroll inside their own box while the page does not.


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`, with the package's tests, a build, and measurement through the shipped `<Md>`.

Three things held up under attack and needed nothing: the **parser layer** is right — a micromark construct really is where this belongs, and I confirmed code spans, fenced and indented blocks, link destinations, image alt text and HTML blocks never offer a `\` opener to it, while the give-up memo collapses the 1000-stray-`\[` case to 140ms against a 70ms control rather than the 235× it could have been. **Security** is clean: `trust: false` leaves `\href`, `\url`, `\includegraphics` and the `\html*` family inert (they render as literal `\command`), `\text{<script>…}` is escaped, the `title` on `.katex-error` is escaped, a `\def` bomb is stopped by `maxExpand`, and rehype-katex hands React *hast elements* rather than a `raw` node, so no `dangerouslySetInnerHTML` is involved. And the **font pipeline** does what it claims — reproducing the `enforce: "pre"` transform in an isolated build emits 19 woff2 assets, zero woff/ttf, and no external host anywhere in the emitted CSS, so the offline/desktop claim holds and there is no CDN risk.

The problems were on the rendering side.

### 1. KaTeX re-rendered every formula in the streaming message on every delta

`md.tsx` memoizes by text identity, so the settled transcript is skipped — but the actively-streaming message re-renders ~8×/second by its own module comment, and the rehype stage was unconditionally inside that loop. Raw `katex.renderToString` is cheap (0.30ms), but rehype-katex then re-parses ~2.5KB of markup per formula through `hast-util-from-html` and React rebuilds several hundred elements, ~5.5ms per formula, paid every frame.

Measured through the real plugin lists: a 4.4KB reply with 60 formulas went 49–107ms (gfm only) → 31–63ms (math parsed, not typeset) → **375–543ms** (full pipeline). Streamed in 40 deltas: 159ms per delta with math vs 21ms without. That is the same O(n²)-over-a-long-reply shape the memo and `highlight={!streaming}` were both written to prevent.

**Fixed** in 345c1951 with the flip the file already precedents — `rehypePlugins={streaming ? NO_REHYPE_PLUGINS : REHYPE_PLUGINS}`. The memoizing wrapper was prototyped and measured first, then rejected: it left 1853ms of 2352ms in place and did nothing at all for the growing formula, which is the one case actually being re-rendered. Numbers after the fix:

| | before | after |
| --- | --- | --- |
| 3.6KB reply, 60 formulas, one delta | 178ms | 6.7ms |
| same, streamed in 40 deltas | 2571ms | 112ms |
| one 16KB formula | 907ms | 4.1ms |
| 20KB paragraph caught by a stray `\[ … \]` | 857ms | 4.1ms |

One consequence handled: with no rehype stage, `remark-math`'s block form reaches `MdPre` as `<pre><code class="language-math">` and would have grown CodeBlock chrome — language label and copy button — for the length of the reply and then lost it. `MdPre` now renders a `language-math` fence as a plain `<pre>`.

### 2. A single-line `$$…$$` rendered as *inline* math — `white-space: nowrap`, no overflow guard

Only the fenced form became display math, so `$$E=mc^2$$` on one line rendered as `class="katex"` with no `katex-display`, and the new CSS scoped its scroll guard to `.katex-display` only. `.md-body table` and `.md-body pre` both carry `overflow-x: auto` precisely because this app's containers do not clip.

Measured at 760px with a 640px column: an ordinary one-line integral-and-sum rendered **1421px wide and the page itself scrolled to 1514px**. Written as `\[…\]` the identical content scrolled correctly — two documented-as-equivalent forms behaving differently.

**Fixed** both halves: a new `remark-math-dollars.ts` re-classes dollar-delimited math to display (reading the delimiter back from the source, since both plugins emit `inlineMath`), plus an inline guard. With it the page stays at 760 and the 1421px scrolls inside its own box. `vertical-align: middle` is load-bearing there — A/B screenshotted in both themes, the default baseline visibly floats `E=mc^2` and `\frac{a}{b}` above the line. No colour introduced, so no tone token needed.

### 3. Three smaller ones

A `\]` or `\)` **inside an inline code span** closed a math span opened in prose, destroying the code span — the module's stated invariant held for the *opener* only. ``In LaTeX you open with \( and close with `\)`.`` was enough to reproduce it. Fixed by declining at a backtick run, with the cost documented.

`maxSize: 100` did not do what its comment claimed: it clamps to 100em, still a ~1600px black box, and `overflow-y: hidden` cannot help because the element's own height *is* 100em. Dropped to 5.

"Four delimiter forms" is three — single `$` is deliberately off — and `$$…$$` was documented as display when it was not. Corrected in the module, both changelog files, a test name and the PR body.

### Recorded rather than fixed

`\[` / `\]` is CommonMark's escape for a literal bracket, so `See \[1\] and \[2\] for details.` now renders as centred display blocks splitting the sentence — likewise `Use \(foo\) to capture a group.` and `C:\Users\(test\)\file`. The ambiguity is real and the PR resolves it deliberately; it is now in the changelog's Left-out list, the module's Known limits, the PR body, and pinned by a test.

### Checks

- Design: no companion PR needed. `06-PROTOTYPE.md:75` describes the Markdown body in one clause and enumerates neither GFM nor Shiki nor table scrolling; `08-CHANGELOG.md`'s header explicitly excludes 局部渲染; and the closest precedent — the 2026-08-24 autolink-boundary change — shipped with no design entry.
- Autolink behaviour did not vanish: `remarkAutolinkBoundary` is unchanged and still second in the list, the one-line test edit only re-points the import, and its 8 assertions now run through the math-enabled list and pass. Plugin order checked — the math plugins register parser extensions spliced ahead of `characterEscape` at `\`, while the only tree transformer is the autolink one, still after GFM.
- `THIRD-PARTY-NOTICES.md` correct: `node_modules/katex/LICENSE` is verbatim MIT with no separate font licence.
- `markdown-math` 29 tests (was 24) + autolink 8, `tsc --noEmit` zero errors, `vite build` clean, `pnpm -r test` **4142 passed / 5 skipped** across 8 packages, prettier clean.

PR #504, opened 2026-08-27
feat(server,web): bind a QQ bot by scanning a QR code

hiyouga:
Binding a QQ bot no longer requires copying an App ID and an App Secret out of the developer console. The QQ section of the binding editor now leads with a QR code: scan it in QQ, pick the bot, and the server has the credentials. The typed fields stay below it as the fallback.

Rebased onto `main` after #501 merged (as `483af5db`), which dropped this branch's copy of that PR's superseded commits — `main` had already gained #501's own review fixes, and the link fix that used to sit on top here landed there too, so it replayed as empty. This PR now carries only the scan flow.

## The protocol

Three calls against `q.qq.com` — a different host and a different envelope (`{retcode, msg, data}`) from the bot OpenAPI, which is why it lives in its own module rather than in `qq-api.ts`:

1. `POST /lite/create_bind_task` with `{key}`, a fresh base64 32-byte AES key → `task_id`.
2. `https://q.qq.com/qqbot/openclaw/connect.html?task_id=…&source=…&_wv=2` is **encoded into a QR code**, never fetched by this server.
3. `POST /lite/poll_bind_result` with `{task_id}` every 2s → status plus, on completion, the App Secret AES-256-GCM-encrypted under the key from step 1: base64, IV = first 12 bytes, auth tag = **last** 16, ciphertext between.

The completed `data` is a bare **object**, not an array — I read the published dist to check. Tencent's own client wraps it in a length-1 array, which is presumably where a future multi-bot response would surface, but the wire returns one bot. So there is no bot picker: it would be UI for a state the protocol does not currently produce. The parser normalizes both shapes anyway, and the App ID lands in the form where the user can see which bot bound.

## The key is the whole security model

The AES key decrypts an App Secret, so it is exactly as sensitive as the secret. It is generated, held, used and dropped in `qq-scan.ts`. The browser is handed a task handle, a URL and a status; the decrypted credentials go straight into storage without passing back through it. That is the same never-round-trip-the-secret rule the PUT handlers already follow, applied to a secret that arrives from outside rather than from the user. Every route assertion in the tests greps the whole response body for the key.

Around that:

- Tasks are **in memory only**. Persisting a key that decrypts an App Secret would buy nothing but a second place for it to leak from; a restart mid-scan loses the task and the user starts another, which is a strictly better failure than a key at rest.
- Tasks are **scoped to the Session that started them** — another Session's poll reads as unknown, at the route as well as at the service — and bounded **per Session** (four in flight) rather than process-wide. That is the ceiling that binds: starting a scan is reachable by anyone who owns a Project, so a table bounded only across all callers would let one of them in a loop evict everybody else's code. Eviction can only ever drop the caller's own oldest task; the process-wide bound stays as a memory backstop far above any real load.
- A task is **claimed by the poll that resolves it**, before the upstream request rather than after it. The client's interval fires whether or not the previous request came back, so a slow platform leaves several polls of one task overlapping; without the claim each of them decrypts the secret and binds again. Both a replay and a second concurrent poll now read as 404. A request that itself failed resolves nothing and puts the task back.
- A task nobody came back for is swept at its ten-minute TTL by an **unref'd, self-re-arming timer**, not at whenever somebody happens to start the next scan — which may be never. Cancel (including leaving the editor, and the panel giving up after a run of failed polls) forgets the key at once.
- Every scan route is **owner-only**. The flow ends in a stored credential however little of it the caller types, so it sits on the same side of the vault split as PUT.
- A completed scan **saves and stops**. Enabling stays the separate, exclusive act, so a scan cannot take a bot away from a conversation currently holding it. Starting a scan while this Session's own QQ connection is enabled is refused with 409 `messaging_disable_before_scan` — the rule the button, the copy and the test already stated — and a scan completing onto a connection switched on meanwhile runs the same account guard the PUT does.

## Dependency decision

`@tencent-connect/qqbot-connector@1.2.0` implements exactly this and is **not used** — I verified the package rather than taking the brief's word for it: `"license": "UNLICENSED"` with no LICENSE file, a dist that is machine-obfuscated (mangled identifiers plus string-array indirection), and a `qrcode-terminal` dependency for printing a QR to a console. Any one of those is disqualifying for a server inside a distributed product that maintains `THIRD-PARTY-NOTICES.md`. The protocol is ~50 lines and is reimplemented behind an injectable seam in the style of `telegram-api.ts`.

For the QR itself I added **`uqr`** to `packages/web`: MIT (plus Project Nayuki's MIT notice for the algorithm), **zero dependencies**, TypeScript types included, and an `encode()` that returns a plain module matrix — so the SVG is built in JSX rather than injected as markup.

`THIRD-PARTY-NOTICES.md` is **not** touched. An earlier revision added a `uqr` section; `main` has since restated that file's scope as a third-party work redistributed *as its own file*, explicitly excluding npm packages compiled into the application bundle (React, Shiki, xterm.js, KaTeX's own code). `uqr` is one of those, and a selectively complete notices file is more misleading than one with a stated narrow scope. Its licence is recorded in `package.json` and `pnpm-lock.yaml`.

## Rendering the QR, and the polling panel

Generated in the browser and inlined as `<svg>`, so no third party learns the task handle from an image request. Two details are asserted in tests because they are the difference between a code that scans and one that does not: the four-module quiet zone the spec requires, and **dark-on-white in both themes** rather than theme-following colours — a code inverted for a dark background is read unreliably by phone cameras that assume dark modules on a light field. Modules are drawn 1.02 wide so neighbours meet, since a hairline gap at fractional device pixel ratios is what makes a rendered code hard to read.

The loop's decisions live in a pure `qqScanStep` (the `updateCheckOutcome` idiom), so they are tested by value in a package whose vitest runs in `node` by a documented decision:

- A poll that failed is ridden out for a couple of attempts before the code comes down. The old behaviour tore the QR down on the first transient error without cancelling the task, so a user walking to their phone came back to a scan that bound nothing while the server kept the key.
- Giving up releases the task; an `expired` answer does not, because the server consumed it on that poll.
- An expired code is replaced automatically, but a bounded number of times. A platform answering `expired` to tasks it just created — clock skew, throttling, an outage returning one fixed code — would otherwise loop create/poll for as long as the panel stays open.

`scanHint` moved out of the button row into the QQ setup fold: it is semantics, and per the frontend skill a neighbouring `<Button>` is not a title for a standing sentence. The zh scan steps no longer say "the code on the left" — the panel is `flex flex-wrap`, so in the dock or on a phone the text wraps below the code.

## Contested files touched

`routes/messaging.ts` (three routes appended to the `/qq` subtree, one line in its DELETE), `api/types.ts` (two response types), `app.ts` (one dep, one override), `strings.ts` / `strings-en.ts` (keys inside the `qq` section #501 added), and `messaging-binding-editor.tsx` (`sessionId` and `adoptBinding` on the editor state, the scan block at the top of the QQ branch, one paragraph in the module doc, one line in the setup fold). `pnpm-lock.yaml` gains `uqr`.

## Verification

From the worktree root, after `pnpm install --prefer-offline`:

- `pnpm -r typecheck` — exit 0 (8 projects)
- `packages/server` `npx vitest run` — 96 files, **1214 passed**
- `packages/web` `npx vitest run` — 107 files, **1471 passed**
- `npx prettier --check .` — "All matched files use Prettier code style!"

`packages/server/test/messaging-qq-scan.test.ts` (22 tests) — the AES-GCM round trip against node's own crypto (the only thing that can prove the IV/tag layout, since a wrong offset authenticates as a failure rather than as wrong-looking output), wrong-key and truncated-payload rejection, the envelope and status mapping, the task table's ownership / claim / expiry / cancel rules, the per-Session ceiling, the sweep timer, and the routes including the replay 404, the cross-Session poll, the member 403 and both disable-first refusals. `packages/web/test/qq-scan-connect.test.ts` (9 tests) — the inline SVG, the quiet zone, the fixed palette, the enabled-connection gate, and every branch of `qqScanStep`. Each behavioural change was mutation-checked: the source change reverted, the test confirmed red, the change restored. No test opens a socket or a real request to QQ.


hiyouga:
Carried the QQ link fix onto this branch as `eae95a11` (cherry-pick of `236ac854` on #501), so this PR does not read as reverting it while the stack is unmerged. The link hunk therefore shows up in this diff too — as the fix, not as a revert — because this branch's merge base with #501 still predates it; it disappears once #501 merges or the stack is rebased. Rationale and the over-the-wire link audit are in the comment on #501.

Re-verified on this branch after the cherry-pick: `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck`, `pnpm test` (server 1124, web 1373, core 1013, cli 374, desktop 144, docs 53, landing 69, skills 27), `pnpm format` + `pnpm format:check` — all clean.

hiyouga:
## Review

Reviewed as an authentication review — a QR login mints a long-lived credential from a short-lived handle, so that is where the danger lives.

**The credential path is genuinely well built and needed no rework.** The AES key never leaves the process. All three scan routes sit behind `authMiddleware` + `requireProjectOwner` — verified by probe: 403 `owner_required` for a real Project member, 404 for an outsider. The `taskId` is **not** the authorization: knowing one gets you nothing without owning the Session's Project, and a cross-Session poll answers 404. This is explicitly **not** the #502 OAuth shape — there is no endpoint that attaches a credential on a bearer handle alone. The dependency choice is right too (`uqr`: MIT, zero deps, 68K, client-side), and I probed all four setup links live.

Two majors and a set of smaller ones.

### 1. The 64-task ceiling was global, so one user could starve every other tenant's scans

`QQ_SCAN_MAX_TASKS = 64` bounded one process-wide `Map`, and `sweep()` evicted oldest-first **across all Sessions and all users**. No per-user or per-Session cap, and `start()` is reachable by anyone who owns any Project — which is every signed-up user, since signup provisions one.

Verified: start one task for a victim Session, then 64 for an attacker's, and the victim's poll returns `null` — their poll route answers 404 and their QR is dead. A loop on `POST …/messaging/qq/scan` keeps the table permanently full so no other tenant can ever complete a scan, and fires one `create_bind_task` to `q.qq.com` per iteration, risking a rate-limit on the deployment's egress IP.

**Fixed** in d9b6df62 — a per-Session ceiling of 4 that evicts only the caller's own oldest (and only after the create succeeds, so a failed create costs nothing), with the process-wide bound kept as a memory backstop at 4096 and now picking by `createdAt` rather than Map insertion order. Mutation-verified.

### 2. A completed task was not consumed atomically

The upstream request ran *before* the delete, so overlapping polls all resolved the same task — while the module doc, `api/types.ts`, the route comment and both docs pages all state a replayed poll is a 404 rather than a second bind. Verified: two concurrent polls both returned `{status:"completed", bot:{…}}`. The client makes this reachable — `setInterval(…, 2000)` fires regardless of whether the previous request returned, and the upstream call has a 15s deadline, so a slow `q.qq.com` yields up to 7 concurrent polls, each decrypting the secret and running `upsert` + `sync`.

**Fixed** — the task is now *claimed* before the upstream request rather than deleted after it, and released only while unresolved, including on a transport throw so a failed request leaves the code pollable. Both halves mutation-verified.

### 3. The scan route did not enforce the rule its own UI, copy and test all claimed

The button is disabled when `enabled`, `S.qq.scanDisableFirst` says "Disable the connection before rebinding by scan", and the test asserted "a rebind would swap the credentials under a live connector, so it is refused" — but the route refused nothing. Verified: with a binding enabled on one bot, the scan started and the completed poll rewrote the row to a different `accountId` while `enabled` stayed `true`, then ran `sync`. Since `guardEnable` was never consulted, that could point a live connector at an account another Session — possibly in a Project the caller cannot see — already had enabled.

**Fixed** with two guards, made consistent with the `guardAccountFree` `main` added to the QQ PUT in #501's own review: a 409 `messaging_disable_before_scan` before any `q.qq.com` request, and `guardAccountFree` on the completed poll — reachable only if the connection was switched on mid-scan, which is why both exist.

### 4. Two client loops that never ended

A transient poll failure tore the QR down without cancelling the server-side task, leaving it and its AES key resident indefinitely — the exact "key at rest" outcome the module doc argues against — because the TTL was checked lazily inside `poll()` and the sweep ran only when someone started another scan. And a repeated `expired` looped create → poll → expired → create forever, ~2 outbound requests every 2 seconds for as long as the panel was open.

**Fixed** — three consecutive failures tolerated before giving up, and giving up now releases the task; an unref'd self-re-arming sweep on the server that stops when the table empties; and a bounded auto-refresh (3) with a new string for the give-up case.

### 5. The "owner-only" test did not exercise the owner check

It provisioned a brand-new user who was not a member of the Project at all and asserted 404 — which comes from `requireProjectAccess`, not `requireProjectOwner`. Deleting all three owner checks would have left it green. Now adds a real member and asserts 403 `owner_required` on all three routes; mutation-verified by removing one of the three.

### 6. Smaller ones

The zh changelog said 黑底白码 (black plate, white modules) where the English said dark-on-white and the code renders `fill="#ffffff"` then `fill="#000000"` — the inverse of both, and inverted codes are exactly what that paragraph argues against. Fixed to 白底黑码.

The zh `scanSteps` said 扫描**左侧**二维码 while the English said only "Scan the code" — but the container is `flex flex-wrap`, so in the narrow dock panel or on mobile the text wraps *below* the QR. Dropped, and pinned by a test asserting neither dictionary names a position.

`scanHint` was always-on semantic prose beside a button, which the frontend skill routes to a fold ("a control is not a title"). Moved into the QQ setup fold and reworded to lead with "Or connect by QR:" now that it sits under the manual steps; the two transient in-panel strings left alone as live task guidance.

The module doc's scan paragraph had been spliced into the middle of an existing sentence, leaving a dangling "Secrets follow that" and a duplicated restart — resolved during the rebase, since `main` had rewritten that sentence anyway.

### The `THIRD-PARTY-NOTICES.md` decision

I recommended dropping the `uqr` section: react, react-dom, react-router, shiki, xterm, zustand and remark-gfm are compiled into the same shipped bundle and were not listed, so a reader of the new section would reasonably conclude `uqr` was the only third-party source redistributed — and a selectively-complete notices file is more misleading than one with a stated narrow scope.

That got easier while the work was in flight: `main`'s preamble now says outright that "What earns an entry is a third-party work redistributed **as its own file**. An npm package whose JavaScript is compiled and minified into the application bundle (React, Shiki, xterm.js, KaTeX's own code, …) is not listed." So taking `main`'s side in the conflict *was* the fix, and the file is untouched by this PR.

### The rebase

#501 merged, so this dropped its base commits — and `eae95a11` replayed as **empty**, because `main` already carried all of it via #501 and #506 (the `credentialSource` rename plus the wiki-root and trailing-slash link fixes). `main` then moved twice more during the fixes (#505, #507, both touching `routes/messaging.ts`, the binding editor and both dictionaries). Rather than assume, that was checked: the new `main` merged onto a throwaway branch and the whole chain run against it — build clean, typecheck clean, prettier clean, **server 1238**, **web 1475**.

### Still open before merge

**Design.** There is no QQ section in `design/specs/` at all, so #501's channel and this PR's scan flow both need writing: the three `q.qq.com` calls, the AES-GCM framing, the task table with its Session scoping / TTL / per-Session ceiling / claim-on-resolve rule, that a completed scan **saves only** (enabling stays the separate exclusive act), the owner-only routes and the `channel ∈ feishu|telegram|qq` widening. I am writing that.

### Checks

- Server **1214/1214**, web **1471/1471** on the branch; typecheck clean across all 8 projects, prettier clean, build clean. Every fix mutation-verified.
- Not covered, deliberately: the React effect shell around the extracted `qqScanStep` classifier (interval teardown on unmount, `setPhase` clearing, the `goneRef` late-start cancel). The classifier itself is tested by value in every branch — the `updateCheckOutcome` / `telegramTestNotices` idiom — and covering the effects needs a DOM environment the package excludes by documented decision. No jsdom was added.

PR #505, opened 2026-08-27
feat(server,web): a messaging binding reports what it has actually seen

hiyouga:
> **Stacked PR.** Based on `fix/telegram-group-privacy-mode` ([#499](https://github.com/Prism-Shadow/penguin-harness/pull/499)) → `fix/messaging-group-mentions` ([#497](https://github.com/Prism-Shadow/penguin-harness/pull/497)). Retarget to `main` once those land; only the commit `feat(server,web): a messaging binding reports what it has actually seen` belongs here.

**No fix for the reported symptom, because it does not reproduce.** A user's Telegram bot answered nothing in a supergroup; after re-enabling the connection the full round trip works — receive → task → reply — in the same group. The two explanations that fit the evidence (a manual `curl getUpdates` competing for the token, or the poll loop parked in backoff) both self-heal on a fresh connect, which is what re-enabling gave it. The group path is audited below and has no defect.

What this PR ships is the reason that took a round trip to the user, a hand-run API call and a lucky reconnect to establish: **the panel could not say any of it.**

## "Nothing happened" had three causes and one appearance

1. Telegram never delivered the message (privacy mode, the bot not really in the group, another program holding the poll).
2. It arrived and `startTask` threw — `MessagingBridge.onInbound` catches that into `messaging_inbound_failed` and returns. #484 named this and left it open.
3. The Task ran and the send failed — `messaging_send_failed`, recorded, silent.

`grep -rn "messaging_inbound_failed\|messaging_send_failed" packages/web/src` returned nothing, so all three read as `connected`, no error, nothing on screen.

## The runtime status now reports the traffic, not just the socket

`MessagingRuntimeStatus` gains three fields, all filled by the live connection and all rendered under the toggle beside the existing connection-error line (the arrival line is deliberately not gated on `connected` — it reports traffic rather than the socket, and `connecting` and `error` are where a reader most needs it):

- **`lastInboundAt`** — when this binding last accepted an inbound message. Absent while none has, and that reading is the point: *connected, no error, nothing has ever arrived* is exactly what a withholding channel looks like. It is also what the user's manual `getUpdates` established the hard way — four updates queued, none consumed.
- **`lastDeliveryError`** — `{at, stage, detail}` for the most recent failure *after* a message was accepted. `stage: "inbound"` sends the reader to the Session; `stage: "send"` sends them to the bot's rights in the chat. Both are identical silence in the chat, so the stage is stated rather than folded into one sentence. Arrival is still reported alongside it: a send failure is not a delivery failure, and reporting only the failure would send someone hunting for a delivery problem that does not exist. A later success does not clear it — an intermittent failure would otherwise be erased by the next ordinary message, the same blind spot `lastConnectionError` closes — so the panel line carries the failure's own time rather than parking it in a hover-only `title=`.
- **`lastConnectionError`** — `{at, detail}` for the last connection failure, **kept after recovery**. `lastError` belongs to the `error` state and is wiped the moment the state leaves it, so a connector that flaps — precisely what a second program taking turns on one bot token produces — reads as a clean `connected` in every snapshot taken between its failures. This is the trace it leaves behind.

All three are in-process and scoped to one **connection**, not to one server run: `connect()` builds a fresh `BridgeEntry`, and `sync()` runs it on the state toggle *and* on any credential save made while the binding is enabled. The copy says exactly that — "No message has arrived since this connection opened" — and the group-silence entry below treats that line as evidence only about a message sent after reading it. A line that read as "never" would send a user who had just re-enabled a working binding off to kick the bot out of a healthy group.

## Messaging errors were filed nowhere a user could reach

`MessagingBridge.recordError` passed `ctx: { sessionId }` and nothing else. `GET /api/projects/:id/usage/errors` selects by project, and a record with no project is *unattributed* — served only to admins (`includeGlobalErrors`). **An ordinary member could never see a failure of their own binding anywhere in the product.** Records now carry the Session's `projectId` and `agentId`, read from the sessions index the bridge already holds (`MessagingSessionIndex.findById` widened from `object | null` to the two fields); a Session deleted since leaves the record unattributed rather than dropping it. Pinned by two tests that assert the code turns up in `errorsRepo.recent(projectId)`.

## Audit of the group path — no defect, one regression pin, two findings

**The real supergroup shape is now a fixture.** No test in the repo used a chat id like `-1004475424385`: negative, 14 digits, and its reply ref `-1004475424385:4` puts a leading `-` right next to `parseReplyRef`'s `indexOf(":")`. Traced and now pinned end to end — `wireChatId` keeps it a safe integer, the ref round-trips (`indexOf(":")` is 14, not 0), the row records it as a non-direct chat, and the reply threads onto message 4 in that chat.

**`allowed_updates: ["message"]` is correct for groups.** Confirmed against the Bot API docs: a supergroup message arrives as `message`, not a distinct type. Nothing group-specific needs a second update type.

**`#497`'s mention normalization does not empty a real group message.** Two regression pins added, both passing before and after (this is an audit result, not a reproduction): a Telegram message whose bot handle sits *after* a surrogate pair (`🚀 @penguin_test_bot 帮我看看 build 好了没`) — the fixture asserts its own offsets against the text, so it cannot drift into agreeing with the code, and a code-point-based implementation would cut a character early — and the Feishu counterpart. The user's actual payload, `@hiyouga_bot hi`, is the shortest thing that could have been emptied by mistake; it now has its own test and produces `hi`.

Two findings that are **not** shipped here:

- **Telegram channel posts are a permanent, invisible dead end.** They arrive as `channel_post`, which `allowed_updates` excludes and `normalizeUpdate` ignores, so a bot in a channel is silent forever with no error anywhere. Supporting it is a product decision with a real cost (a busy channel means a Task per post), so it is **named in the troubleshooting copy** instead — the requirement was that it stop being invisible, not that it be supported.
- **Forum/topic supergroups get their later replies in the wrong topic.** A run's first outbound message threads onto the inbound one via `reply_parameters`, so it lands in the right topic; every message after it, and the approval notice, are plain sends with no `message_thread_id` and therefore land in **General**. Real, separate, and unshipped: it lives in `bridge.ts`'s outbound half, needs the thread id plumbed through the reply ref or the binding row, and raises a design question (what does a binding remember when a user writes in two topics?). Reported rather than guessed at.

## The retest recipe, and the trap it names

The one-program-per-token entry now says the thing that cost this investigation the most time: **a `getUpdates` you run by hand is that other program.** It must be run with the connection disabled, and reading updates by hand *confirms* them — the app will never see those messages, so a retest needs a freshly sent one. The new group-silence entry names the rest: the bot is still in the group, a Group Privacy change needs the bot removed and re-added, nothing else is polling the token, and channel posts are not supported.

## Contested files touched

- `packages/server/src/runtime/messaging/bridge.ts` — inbound path, `statusOf`, and `recordError`. The outbound half is untouched except the two existing `catch` blocks now routing through `recordDeliveryFailure`.
- `packages/web/src/features/messaging/messaging-binding-editor.tsx` — three `<p>` blocks under the existing status block, one import.
- `packages/web/src/lib/strings.ts` / `strings-en.ts` — five new keys each, plus one sentence appended to `troubleOnePoller`.
- `packages/server/src/api/types.ts`, `packages/docs/content/server-api.{en,zh}.md` — additive.

## Verification

Node 24.18.0, Linux. Full CI-parity chain from the worktree root: `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check`, `git diff --check` — all clean. After the review pass: server **1095** passed (92 files), web **1365** passed (102 files), docs 53; `tsc --noEmit` clean for server and web; `prettier --check` clean on every changed file. Earlier full-chain run: core 1013 passed / 5 skipped (47 files), cli 374, desktop 144, landing 69, skills 27. (Counts include #497 and #499, which this is stacked on.)

The four observability tests were run against the pre-change source and fail there:

```
× reports when a message last arrived, and says plainly that none has
  → expected 'undefined' to be 'string'
× a reply that never went out is reported on the status and filed under the Project
  → waitFor timed out
× an inbound message whose Task cannot start > is reported on the binding's status instead of being silent everywhere
  → waitFor timed out
× MessagingBindingBody > says whether anything has arrived, and which end failed when something did
  → expected '<div class="space-y-3"><div role="gro…' to include undefined
```

Tests added — `packages/server/test/messaging.test.ts`: arrival reported and the empty case stated; a send failure reported with `stage: "send"` and attributed to the Project; a connection failure surviving recovery; a whole new suite driving a Session that cannot load, asserting `stage: "inbound"`, no chat output, and the attributed record. `packages/server/test/messaging-telegram.test.ts`: the real supergroup id end to end, and the surrogate-pair offset pin. `packages/web/test/messaging-binding-editor.test.ts`: the three status lines, and that the retained connection error does not double up while the connection is actually down.

## Review pass

Seven findings, all addressed in one follow-up commit. No change to the accounting itself — `lastInboundAt` is still stamped on acceptance, after the redelivery drop and before the work.

- **The arrival record is per connection, not per server start.** `connect()` builds a fresh `BridgeEntry` and `sync()` calls it on the state toggle *and* on a credential save made while the binding is enabled. Six places claimed server-start scope, and the new group-silence entry told the reader to act on the empty line — so a user who re-enabled a working binding (which the one-poller entry prescribes) could be sent to remove the bot from a healthy group. The copy now names the scope in both languages, that entry treats the line as evidence only after a fresh message, and the type docs, both `server-api.{en,zh}.md`, both changelog files and the editor comment are corrected.
- **`lastDeliveryError` is kept, not cleared, and its line carries the time.** Kept deliberately: clearing on the next success would erase every failure that comes and goes — a send right revoked for a minute, a rate limit — which is the blind spot `lastConnectionError` exists to close; and a `send` success says nothing about whether `inbound` acceptance recovered. The staleness is carried by the timestamp instead, which moves out of the hover-only `title=` and into the sentence, matching `lastConnectionError`'s shape, in both dictionaries and for both stages.
- **`lastConnectionError` was missing from the docs**, and the changelog said "two fields" over three bullets and "Both are in-process" over three. Both docs paragraphs gain it; the counts read three / all three; the missing space before `` `connected` `` in the zh docs is fixed.
- **The arrival line no longer hides in `connecting` and `error`.** It was gated on `connected`, so it vanished during Telegram's handshake and backlog drain and during the flapping-token case this PR was written for. It is a fact about traffic, not the socket; it now renders whenever a connection exists.
- **The changelog sections are rewritten in past tense**, per `changelog/README.md`, in both languages, mirrored.
- **Seven tests added.** The reset on re-enable and on a credential re-`PUT`; a redelivery not moving the stamp (driven on the bridge's injected clock, since two wall-clock writes in one millisecond are indistinguishable); a non-text message still stamping it; the failure record surviving a later successful send; a superseded connect's failure filing nothing; and web-side the `stage: "inbound"` branch, which had never been rendered, plus the arrival line under `connecting` and `error`. Each was mutation-checked: the source change reverted, the test confirmed red, the source restored.
- **`connect()`'s two byte-identical connection-failure blocks** fold into one `recordConnectionFailure`, behind the stale-entry guard `setStatus` already used — so a callback from a superseded attempt no longer files an error record against a binding that has since been reconnected — and the six raw `new Date(this.now()).toISOString()` into `nowIso()`.


hiyouga:
## Review

Reviewed in a dedicated worktree, with the real routes driven against the fake Feishu SDK.

The accounting itself is sound and needed no change: `lastInboundAt` is stamped on acceptance, after the redelivery drop and before the work, which is the right instant for "did the channel deliver anything"; no hot-path cost; no concurrency hazard (single-slot assignments on a single-threaded loop, and the only added DB read is on the failure path). What was wrong was the honesty layer around the numbers.

### 1. "Connected, but no message has arrived yet" was scoped to the connection, not the server — and the copy told users to act on it

`connect()` builds a fresh `BridgeEntry` with all three fields null, and `sync()` → `connect()` runs on the state toggle **and on any credential PUT while the binding is enabled**. So the reset is per *connection*. Six places claimed server-start scope: the `types.ts` doc comment, both `server-api` docs, both changelog files, and the editor comment.

Verified by running the real routes: bind, enable, fire one inbound message, then `POST /state {enabled:false}` + `{enabled:true}` → `statusOf` comes back `connected` with **no `lastInboundAt`**. A plain re-`PUT` of the same credentials does the same.

That matters because the new `troubleNoGroupInbound` entry read the empty line as proof — *"Then Telegram is not delivering it and nothing on this machine can change that … remove the bot from the group and add it back."* And `troubleOnePoller`, in the same PR, prescribes toggling the connection. So the documented path was: toggle the connection, see "no message has arrived yet" for a binding that received one seconds ago, and go kick the bot out of a working group.

**Fixed** in 0bc59773 — the copy is scoped to the datum ("since this connection opened" / 「本次连接建立以来」), the fold says the line covers the current connection alone and is evidence only about a message sent *after* reading it, and all six "server start" claims are corrected. The constraint is now recorded on `BridgeEntry`'s doc so a future edit cannot re-assert a scope the code lacks. Mutation-verified against a genuine per-server-start implementation.

### 2. `lastDeliveryError` was never cleared and its visible line carried no time

Nothing reset it on a later success, and unlike `lastConnectionError` ("The connection dropped **at {when}**: …") the delivery strings interpolated only `detail` — the timestamp existed but lived in `title=`, hover-only and unavailable on touch.

Verified: with `failSend` set, fire a message → error recorded. Clear `failSend`, fire a second, wait for it to land → `statusOf` still returns the *original* failure byte for byte, with `lastInboundAt` updated beside it. So: a group admin revokes send rights for a minute, one reply fails, rights are restored and everything since has gone through — and the panel says "the task ran but its reply never went out" for the life of the connection, with nothing on screen saying it was three days ago.

**Fixed**, and the choice is worth recording: the error is **kept, not cleared**, and the time moved into the sentence instead. Clearing on the next success would reintroduce exactly the blind spot `lastConnectionError` exists to close — an intermittent failure (rights revoked briefly, a rate limit, a cap only long replies trip) would be wiped by the next ordinary message and never seen. A successful `deliverReply` is also a `send` fact and says nothing about whether `inbound` acceptance recovered, so clearing an inbound-stage record there would assert a recovery the code never observed.

### 3. The docs and changelog documented two of the three new fields

`lastConnectionError` was absent from both `server-api` files entirely, and the changelog said "gains **two** fields" / 「新增**两个**字段」 then listed three bullets and closed "**Both** are in-process" over three. Fixed in all four places, plus a missing space before `` `connected` `` in the zh docs.

### 4. The arrival line disappeared exactly when the user was looking hardest

It was gated on `state === "connected"`, so during `connecting` (Telegram's `getMe` → `getWebhookInfo` → backlog drain → first `getUpdates`) and during `error` it rendered nothing — including the flapping-token case this PR was written for, where the binding sits in `error` half the time. Now `state !== "disconnected"`: it is a fact about traffic, not about the socket, which is the stated rationale for holding it off `entry.status` in the first place.

### 5. Tests pinned that a number goes up, not that it tells the truth

All seven gaps closed, plus one the review did not list: the redelivery case needed the bridge's **injected clock**, because with the wall clock two writes in one millisecond are indistinguishable and the assertion would have been vacuous. Server 34→41 tests, web 9→12. Eight mutations confirmed red then restored.

The duplication cleanup (`nowIso()`, `recordConnectionFailure`) turned out to carry a real behavioural change — `setStatus` guards against a stale entry but the two raw `lastConnectionError` assignments beside it did not, so a superseded connect's failure used to file an error record. That got its own test.

### Still open before merge

**Design.** `05-ARCHITECTURE.md:1279` enumerates the runtime status in full — 「连接状态（disconnected / connecting / connected / error + 最近错误与时间）仅存内存」 — and that list is now incomplete. The reset scope matters especially there, because the same bullet already contrasts `enabled`（连接意图）with 连接实况仅存内存. `06-PROTOTYPE.md:68`/`:71` also describe a singular 连接状态行 where the panel now has up to four lines. I am writing those.

### Checks

- Server **1095/1095**, web **1365/1365** (including `placeholders-parity`), docs 53/53, both typechecks clean, prettier clean.
- Being restacked onto `main` now that #497 and #499 have merged.

PR #506, opened 2026-08-27
fix(web): the Telegram binding's links lead where their labels promise

hiyouga:
The Bot Token field's corner link was labelled **"open developer console"** and pointed at `https://core.telegram.org/bots/api` — the Bot API reference. Telegram has no developer console, and that page issues no tokens. So a user standing on the one field that needs a token, following the one link that offers to supply it, landed in an 860KB API manual with no way to get what they came for.

Branched off `main` and independent of the QQ stack (#501 / #504), where this was found during a link audit but deliberately not fixed.

## What must not survive

A label and a target that disagree. There were three ways to resolve that, and the evidence picks one.

**Keeping the target and relabelling it "API reference" is wrong.** The corner link's whole job in this editor — the models-page "get API key" idiom it copies — is *where does this value come from*. Relabelling would make the copy truthful and the link useless, leaving the field's actual question unanswered.

**So the target moves, and the label follows it.** `https://t.me/BotFather` is where the credential is issued. Telegram's own page for it is titled `Telegram: Launch @BotFather` and describes it as the bot "to create new bot accounts and manage your existing bots" — it serves a real interstitial with an action button and a `tg://resolve?domain=BotFather` hand-off, so it works with or without a desktop client installed.

**And the label changes too, because the interaction is not the one the shared wording describes.** You asked me to check how the corner link behaves for a `t.me` target: `ExternalLink` renders `target="_blank" rel="noreferrer noopener"` with a `↗` glyph, so a click opens a new tab which then offers to launch the app. The `↗` stays accurate — a tab really does open — but "open developer console" does not, in two ways at once: it names a thing Telegram has never had, and it promises a documentation-style destination when the click ends in an app. Telegram's own verb for it is "Launch". So this channel gets `S.telegram.openBotFather` — 打开 @BotFather / Open @BotFather — and the shared `S.messaging.console` stays untouched for the channels that genuinely have a console.

One implementation detail worth flagging, because the obvious version is a bug: the label is chosen **at render**, not stored beside the URL in `CHANNEL_LINKS`. `S` is a `let` the locale provider reassigns (`setActiveStrings`), so a label captured in that module-level table would freeze at whatever language was active when the module first evaluated and never follow a language switch.

## The tutorial half

You asked whether `/bots/tutorial` still reads coherently once `console` changes. It does not, and this is the part I'd have missed by pattern-matching.

`https://core.telegram.org/bots/tutorial` is titled **"From BotFather to 'Hello World'"**. Its headings run: Introduction → Getting Ready → Obtain Your Bot Token → **Download an IDE → Pick a Framework or Library → Create Your Project → Add Framework Dependency → Start Coding**. It is a guide to *writing* a bot. Nobody using this form is writing one — PenguinHarness is the bot; the user only needs a token.

`https://core.telegram.org/bots/features#botfather` is the BotFather section, which opens: *"Below is a detailed guide to using @BotFather, Telegram's tool for creating and managing bots… **Creating a new bot** — Use the `/newbot` command to create a new bot. @BotFather will ask you for a name and username, then generate an authentication token…"* That is this fold's own setup steps, one level deeper, and nothing more.

So the pair now reads coherently: both surfaces are about BotFather, at the level each needs — the corner link is *do it now*, the fold link is *read how*. I also checked `/bots` ("Bots: An introduction for developers", a "What Can You Do with Bots?" overview) and rejected it as marketing.

On the fragment: `#botfather` is a plain document anchor (`<a class="anchor" name="botfather">`) present in the served HTML, so it lands where it says. Worth stating explicitly because the sibling QQ audit found the opposite case — a `#/developer/developer-setting` hash route that the server never sees, which silently landed on a shell page.

## What each URL returned

Checked over the wire, no proxy in the environment, `curl -sSL` recording status-without-redirect / status-after-redirect / body size:

| URL | raw | final | bytes | verdict |
| `https://t.me/BotFather` | 200 | 200 | 11162 | **new console target** |
| `https://core.telegram.org/bots/features#botfather` | 200 | 200 | 116925 | **new tutorial target** (anchor confirmed in the served HTML) |
| `https://core.telegram.org/bots/api` | 200 | 200 | 860075 | old console target — resolves, but it is the API reference |
| `https://core.telegram.org/bots/tutorial` | 200 | 200 | 43985 | old tutorial target — resolves, but teaches you to write a bot |
| `https://core.telegram.org/bots` | 200 | 200 | 24954 | considered, rejected as an overview |
| `https://telegram.me/BotFather` | 200 | 200 | 11162 | the legacy alias, byte-identical; `t.me` is the canonical one |

Feishu's two links were audited in the same pass and are healthy and correctly labelled, so they are untouched: `https://open.feishu.cn/app` (200, 9649 bytes) really is a developer console, and `https://open.feishu.cn/document/develop-an-echo-bot/introduction` (200, 7464 bytes) really is its setup walkthrough.

## Tests

The existing corner-link test now asserts the new target and that the API reference is gone. A new test pins the thing that actually broke — that the label and the target name the same thing — by requiring `S.telegram.openBotFather` in the Telegram render and asserting `S.messaging.console` is absent from it, while still requiring it in the Feishu render so the shared wording is not quietly lost. The help-fold test pins the new tutorial URL.

## Verification

Full CI-parity chain from the worktree root on Node v24.18.0, all passing:

- `pnpm install --frozen-lockfile` — exit 0
- `pnpm -r build` — exit 0
- `pnpm typecheck` — exit 0 (8 projects)
- `pnpm test` — exit 0: core 1014, server 1073, web 1361, cli 374, desktop 144, docs 53, landing 69, skills 27

Diff is 4 files (49 insertions, 6 deletions) plus the changelog pair, confined to the Telegram channel's links, their label, and their tests.

### DIFF
--- a/changelog/unreleased/2026-08-27-telegram-botfather-link.md
+++ b/changelog/unreleased/2026-08-27-telegram-botfather-link.md
@@ -3,29 +3,35 @@
 - **Date:** 2026-08-27
 - **Type:** fix
 - **Scope:** `web`
+- **PR:** [#506](https://github.com/Prism-Shadow/penguin-harness/pull/506)
 
 [中文版](2026-08-27-telegram-botfather-link.zh.md)
 
 The Bot Token field's corner link was labelled "open developer console" and pointed at
 `core.telegram.org/bots/api`, the Bot API reference. Telegram has no developer console, and
 that page does not issue tokens — so a user standing on the one field that needs a token,
-following the one link that offers to supply it, landed in an 860KB API manual. The link now
-opens `@BotFather`, which is where the token comes from, and says so.
+following the one link that offers to supply it, landed in an 860KB API manual. The link was
+repointed at `@BotFather`, where the token actually comes from, and the label was reworded to
+name it.
 
 ## Details
 
-- `console` for Telegram is `https://t.me/BotFather`. Telegram's own page for it is titled
-  "Launch @BotFather" and describes it as the bot "to create new bot accounts and manage your
-  existing bots".
-- The label follows the target rather than the other way round: this channel uses its own
-  `openBotFather` string instead of the shared "open developer console" wording, which stays
-  in place for the channels that have one. The label is read at render rather than captured
-  in the module-level link table, because `S` is a live binding the locale provider swaps.
-- The setup fold's companion link moves from `https://core.telegram.org/bots/tutorial` to
+- Telegram's corner link was repointed at `https://t.me/BotFather`. Telegram's own page for it
+  is titled "Launch @BotFather" and describes it as the bot "to create new bot accounts and
+  manage your existing bots".
+- The label was made to follow the target rather than the other way round: this channel took
+  its own `openBotFather` string instead of the shared "open developer console" wording, which
+  stayed in place for the channels that have one.
+- The link table's key was renamed `console` → `credentialSource`, and the comments around it
+  stopped calling that slot a console — the next channel added to the table reads a name and a
+  description that fit a credential issued from a chat as readily as one issued from a page.
+- The setup fold's companion link was moved from `https://core.telegram.org/bots/tutorial` to
   `https://core.telegram.org/bots/features#botfather`. The former is "From BotFather to
   'Hello World'", which past its token section is about downloading an IDE and picking a
   framework — writing a bot, which nobody doing this is doing. The latter is the BotFather
   guide itself, opening on `/newbot` and the token it returns, which is what the fold's steps
   already say.
-- Tests pin both halves: the corner link's target, and that its label and its target name the
-  same thing.
+- Tests were extended to pin both halves — the corner link's target, and that its label and
+  its target name the same thing, asserted as one anchor — and to run over both dictionaries,
+  so the label cannot regress in the locale the suite does not activate.
+- The design spec was updated to match ([penguin-harness-design #70](https://github.com/Prism-Shadow/penguin-harness-design/pull/70)).
--- a/changelog/unreleased/2026-08-27-telegram-botfather-link.md
+++ b/changelog/unreleased/2026-08-27-telegram-botfather-link.md
@@ -1,0 +1,31 @@
+# The Telegram binding's links lead where their labels promise
+
+- **Date:** 2026-08-27
+- **Type:** fix
+- **Scope:** `web`
+
+[中文版](2026-08-27-telegram-botfather-link.zh.md)
+
+The Bot Token field's corner link was labelled "open developer console" and pointed at
+`core.telegram.org/bots/api`, the Bot API reference. Telegram has no developer console, and
+that page does not issue tokens — so a user standing on the one field that needs a token,
+following the one link that offers to supply it, landed in an 860KB API manual. The link now
+opens `@BotFather`, which is where the token comes from, and says so.
+
+## Details
+
+- `console` for Telegram is `https://t.me/BotFather`. Telegram's own page for it is titled
+  "Launch @BotFather" and describes it as the bot "to create new bot accounts and manage your
+  existing bots".
+- The label follows the target rather than the other way round: this channel uses its own
+  `openBotFather` string instead of the shared "open developer console" wording, which stays
+  in place for the channels that have one. The label is read at render rather than captured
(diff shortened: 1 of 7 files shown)


hiyouga:
## Review

Reviewed in a dedicated worktree, with every URL checked over the wire.

The link fix itself is right and needed nothing: all six URLs return 200, `t.me/BotFather` is titled "Telegram: Launch @BotFather" and says "Use it to create new bot accounts and manage your existing bots", `name="botfather"` is a real anchor in the served `/bots/features` HTML, and Feishu's two links genuinely are a developer console (开发者后台 - 飞书开放平台) and a tutorial (三分钟快速开发). All four links now match their labels, they share one `ExternalLink` with `rel="noreferrer noopener"`, and the i18n parity is compiler-enforced.

Four things were worth changing, all pushed.

### 1. The tests only exercised the zh dictionary — the same bug could return in English unguarded

`S` is `zh` unless a test calls `setActiveStrings`, and no test in this file did, so both label assertions only ever compared against the Chinese strings. Proven: setting `en.telegram.openBotFather = "Open developer console"` left `pnpm typecheck` green (the key exists, the type is `string`) and all 1361 web tests green — while every English user would see the original label-promises-a-console bug back, on a link that goes to `t.me`.

Now looped over both dictionaries with the house `afterEach(() => setActiveStrings(zh))` pattern that ten other files in this directory already use, and label and target are tied in one assertion so a right-href-wrong-anchor render can't slip through. Mutation-verified in both directions: the new test fails under that mutation, the old one passes.

### 2. Five comments still asserted what the PR stopped being true, and the record key was a lie

The module doc still said the credential field "carries the developer-console link at its label's top-right corner"; `CHANNEL_LINKS`' entry was still documented as "the console where the credential is fetched"; and the key was literally `console` holding a `t.me` deep link. The concrete risk: the QQ channel work (#501/#504) adds a third entry to that record, and its author reads the doc and the key name, follows the pattern, and hangs `S.messaging.console` on a field no console issues — this exact defect, reintroduced from the comments this PR left standing.

`CHANNEL_LINKS.console` is now `credentialSource`, and the docs describe a credential *source* rather than a console. The i18n key `S.messaging.console` deliberately keeps its name — its value literally is "Open developer console", and only Feishu reads it, so renaming it while the value stays console-specific would recreate the same name-promises-X defect one level up. Its doc now says so and points at `telegram.openBotFather` for channels that issue elsewhere.

### 3. The changelog was missing its `PR:` field and was written in the present tense

`changelog/README.md` marks `PR` required when one exists, and both other entries in `unreleased/` carry it — without it the documented `grep 'PR:'` query returns nothing for this change once the folder is renamed at release. Added to both files.

The Details bullets were all present tense ("`console` for Telegram **is** …", "the companion link **moves** …", "Tests **pin** both halves"), and one was pure rationale — "because `S` is a live binding the locale provider swaps" — which the spec routes to the PR description, where it already appears. Recast past tense in both languages, mirrored, with the rationale dropped.

### 4. The spec pinned the label, for both channels

`06-PROTOTYPE.md:68` and `:71` and `05-ARCHITECTURE.md:1282` all read 凭据字段右上角为该渠道「前往开发者后台」链接 — "that channel's", in sentences that had just named both. Telegram has no developer console; the bot and its token both come out of the @BotFather conversation, so the label was promising a destination the link could not lead to. Left as-is, the next person reconciling code to spec would revert the Telegram label and reintroduce the bug **with the spec on their side**.

Amended in all three places on design PR #70's branch, per-channel, with an `08-CHANGELOG.md` entry explaining why the old wording could not stay: [penguin-harness-design#70](https://github.com/Prism-Shadow/penguin-harness-design/pull/70). The changelog pair here cross-references it in the house form.

### Rebased

`main` moved a long way under this branch (ten PRs today). Rebased onto it; the conflicts were doc-comment prose where `main` had newer content (the bind/unbind semantics from #490, the per-line option from #494) and this PR had the credential-source rewording — resolved by keeping both.

### Checks

- `messaging-binding-editor.test.ts` 8/8, full web suite **1422/1422** after the rebase, `tsc --noEmit` clean, prettier clean, workspace build clean.
- Design: the URL half sits below spec altitude — the specs contain no external URLs at all, only two `http://localhost:<port>` examples. Only the label crossed it, and that is now recorded.

PR #507, opened 2026-08-27
fix(server): Telegram replies stay in the forum topic they were asked in

hiyouga:
> **Stacked PR.** Based on `fix/messaging-delivery-observability` ([#505](https://github.com/Prism-Shadow/penguin-harness/pull/505)) → `fix/telegram-group-privacy-mode` ([#499](https://github.com/Prism-Shadow/penguin-harness/pull/499)) → `fix/messaging-group-mentions` ([#497](https://github.com/Prism-Shadow/penguin-harness/pull/497)). Retarget to `main` once those land; only the two commits `fix(server): Telegram replies stay in the forum topic they were asked in` and `fix(server): mint a Telegram forum topic only where one can be sent back to` belong here.

The finding reported out of #505's group-path audit, now fixed. A forum supergroup splits one chat into topics, and every message carries the one it was written in. Only the run's **first** reply stayed there — Telegram infers a reply's topic from the message it quotes — while everything else was a plain send with no `message_thread_id` and landed in **General**: the run's later completed messages, every chunk after the first of a long one, the tool-approval notice, and the test message. Half an answer arrived where it was asked for and the rest turned up somewhere else.

## The topic rides the chat id, not a new column

Taken as suggested, and it holds up: `MessagingInboundMessage.chatId` is already documented as opaque in exactly the way `messageId` is — the connector mints it and the connector consumes it in `sendText` — so `<chat id>:<topic id>` reaches every outbound path without the channel-neutral seam, the repo or the schema learning what a forum topic is.

I checked every reader before committing to it. `lastChatId` is touched in six places outside the repo: four are `!== null` guards (`lastChatKnown` in two response builders, the two `telegram_no_chat`/`feishu_no_chat` 409s), and the other two hand it to `MessagingClient.sendText` — which *is* the connector-side parse. Nothing displays it, compares it to a raw id, or passes it to a channel API. `sendTestMessage` is one of the two `sendText` callers, so it goes through the parse for free; a test asserts it reaches the topic, since it is the path that reads the row most directly.

Three consequences worth naming:
- **Persistence for free.** The remembered topic survives a restart in the column that already persists the chat, with no migration and no additive column in `schema.ts` / `database.ts` / `messaging-bindings.ts` — three files #493, #494 and the QQ stack are all editing.
- **Old rows parse as themselves.** No separator means no topic, which is what every chat outside a forum topic writes anyway. A test writes a bare id straight into the column, the way an older build would have, and sends to it.
- **A Telegram-only concept stays off the seam,** where Feishu and QQ have nothing to map it to.

The binding remembers the **most recent** topic, matching what it already does with the chat: a test moves the conversation from topic 91 to topic 77 and the replies follow.

Because the column changes meaning on disk, the batch gets its own `changelog/unreleased/2026-08-27-backward-compatibility.{md,zh.md}`. The strategy recorded there is **permanent dual-format tolerance**, with the reason it has no removal date: the bare id is not a legacy encoding, it is the only encoding a chat with no topic has. The entry also records the downgrade leg — an older build hands `-100…:91` to Telegram as a chat id, `Number(…)` is `NaN`, and every reply fails with `chat not found` until the next inbound message rewrites the column.

Two comment-only hunks do land in contended files — `schema.ts`'s `last_chat_id` column comment and `MessagingBindingRow.lastChatId`'s doc — because a maintainer reading `last_chat_id TEXT` would otherwise reasonably assume they can `SELECT` it and hand it to the Telegram API. Both now say the string is connector-opaque.

## Only a real forum topic is minted

The Bot API is asymmetric here, and the first round of this PR missed it. `Message.message_thread_id` is documented "for supergroups and private chats only" and is set on an **ordinary reply chain** — a user replying to one of the bot's own messages in a DM, which is the normal way to answer a bot there, and the common case in a non-forum supergroup under Group Privacy, where replies to the bot are most of what it sees. `sendMessage`'s `message_thread_id` is "for forum supergroups only" and refuses anything else with `Bad Request: message thread not found`.

Minting from the inbound field alone therefore persisted a thread id into `last_chat_id` for chats that cannot receive one, and every outbound message from then on cost a rejected send plus a fallback — permanently, until an inbound message with no thread happened to repair the row. That is a regression against one call per message, and it doubles the request rate against Telegram's roughly one-message-a-second per-chat allowance.

`topicIdOf` now requires `chat.is_forum` **and** `is_topic_message`, the pair the Bot API provides for exactly this distinction. `TelegramMessage` gained both fields and the `message_thread_id` doc now states the asymmetry instead of claiming the field is absent outside a forum.

## Reply refs gained the same optional component

A reply normally inherits its target's topic, so the threaded first message was never the bug. But `allow_sending_without_reply: true` degrades a vanished target to a plain send, and a plain send with no topic lands in General — losing the topic precisely when the conversation can least spare it. `replyRefOf`/`parseReplyRef` now carry an optional third component and tolerate the two-component refs already minted (and, after #493, already persisted in `last_inbound_message_id`).

## A deleted topic degrades instead of losing the reply

The Bot API has no `allow_sending_without_thread` to match the reply flag, so a send into a topic that has since been deleted or closed simply fails. `sendWithThreadFallback` retries such a send once without the topic, which puts the message in General.

The retry is keyed on **Telegram's `error_code`**, not on its prose: `call()` now throws a `TelegramApiError` carrying the envelope's `error_code`, and only a `400` with a topic attached is retried. The earlier "any failure, as long as a topic was attached" was too wide in two ways that mattered — a `Too Many Requests: retry after 5` was answered with an immediate second request, lengthening the flood wait and costing two requests per 429; and a transport failure (timeout, reset, a 5xx after delivery) carries no answer from Telegram at all, so re-sending could duplicate the chunk in the chat. Both now surface as `messaging_send_failed`, like any other failure, and a plain `Error` from the fetch path is deliberately never retried.

The degradation is no longer silent. `sendWithThreadFallback` resolves a `MessagingSendNote` — a channel-neutral "the message arrived, somewhere less right than addressed" — and the bridge logs it once per episode rather than once per chunk. It stays off `lastDeliveryError` and off the panel: the message did arrive, and reading it as a delivery failure would be wrong.

## One reply, one place

`deliverReply` snapshotted the binding row once but re-read `entry.lastInboundMessageId` per chunk, with an `await clientFor(...)` between the two. Before this PR a disagreement there produced a wrong quote header; with the topic on the chat ref it could split one reply between two **topics** — chunk 1 threading onto a message that landed mid-delivery in topic 77 while chunks 2..N used the row snapshot naming topic 91. Both halves of the target are now captured together, before the first await. (The between-messages variant is the documented "most recent topic" behaviour and is untouched.)

## Evidence

The repo had no forum fixture; there is one now, and the fixture states the flags Telegram actually sends. Every behavioural change is pinned by a test that fails without it — verified by reverting each source change in place:

| Reverted | Failing test |
| --- | --- |
| `topicIdOf`'s `is_forum`/`is_topic_message` guard | `mints no topic in a private chat, where a reply to the bot carries a thread id`; `mints no topic in a supergroup that is not a forum`; `mints no topic for a reply chain in a forum's General topic` |
| the `error_code === 400` narrowing (back to an unconditional `catch`) | `a flood wait is not retried: one refusal must not cost two requests` |
| the `MessagingSendNote` return | `a topic that has been deleted degrades to a plain send instead of losing the reply` |
| the once-per-episode guard in `noteSend` | same test (three log lines instead of one) |
| the `deliverReply` snapshot | `delivers every chunk of one reply to the chat the row named when the delivery began` |

Against the pre-change source, the three tests that assert the topic reaches sends still fail as they did in the first round:

```
× remembers the topic, and every later message of the run reaches it
  → expected [ undefined, undefined, undefined ] to deeply equal [ 91, 91, 91 ]
× moves to the newest topic the user writes in, like the chat it already remembers
  → expected undefined to be 91
× sends the approval notice and the test message into the remembered topic
  → expected undefined to be 91
```

Call **counts** are asserted now, not just the last send: the fake transport records every `sendMessage` attempt, not only the ones that got through, so "a send that carried no topic is never retried" is actually pinned (one call), the deleted-topic path is pinned at two calls per message, and a threaded send that fails both times is pinned at exactly two attempts with the error still surfacing on the status and in the Project's error records.

## A cross-PR note for the rebase

The legacy-row test writes `last_chat_id` with a direct `UPDATE` rather than through `repo.recordChat`. That is deliberate: #493 adds a fifth parameter to `recordChat`, and a fixture calling it would be a type error on one side of the merge or the other. Writing the column states the stored shape and is immune to both orderings. `pnpm test` was green with the repo call and `pnpm typecheck` was not, which is what caught it.

## Contested files touched

- `packages/server/src/runtime/messaging/telegram-connector.ts`, `telegram-api.ts`, `connector.ts` — the connector/SDK files the media work is rewriting. Additive: `chatRefOf`/`parseChatRef`, an optional third component on the reply ref, `topicIdOf`, `sendWithThreadFallback`, `TelegramApiError`, `is_forum` / `is_topic_message` / `message_thread_id` on `TelegramMessage` and `sendMessage`, `MessagingSendNote` on the two `MessagingClient` send methods (widened to `Promise<MessagingSendNote | void>`, so Feishu's client still satisfies it unchanged), and one doc rewrite on `MessagingInboundMessage.chatId`.
- `packages/server/src/runtime/messaging/bridge.ts` — the target snapshot in `deliverReply`, and `noteSend` plus one `BridgeEntry` field for the degradation log. The first round did not touch this file; the review found two reasons it had to.
- `packages/server/src/db/schema.ts`, `db/repos/messaging-bindings.ts` — **comments only**, no DDL and no query change.

## Verification

Node 24.18.0, Linux. From the worktree root: `npx prettier --check` on all six changed source/test files — clean. `packages/server`: `npx tsc --noEmit -p tsconfig.json` clean, `npx vitest run` **1101 passed (92 files)**. Earlier full CI-parity chain for the first commit: `pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format:check`, `git diff --check` — all clean (web 1363, core 1013 passed / 5 skipped, cli 374, desktop 144, landing 69, docs 53, skills 27; counts include #497, #499 and #505, which this is stacked on).

Eleven tests in `packages/server/test/messaging-telegram.test.ts` cover the forum paths — five of them added in this round, alongside a `failThreadedSendCode` / `failSendCode` knob and a per-client `sendCalls` counter on the fake transport — and one in `packages/server/test/messaging.test.ts` covers the bridge snapshot on the channel-neutral path, using a gate on the fake Feishu SDK's `createClient` to make the race deterministic.


hiyouga:
## Review

Reviewed in a dedicated worktree, against the Bot API reference rather than by inference.

The core idea is sound and the plumbing is clean: packing the topic into the already-opaque `chatId`/`messageId` refs genuinely avoids a schema change, every `lastChatId` reader really is a `!== null` guard or a `sendText` call, and old rows really do parse as themselves. What was missing was the inbound guard.

### 1. `message_thread_id` was minted without checking the chat is a forum

The Bot API is asymmetric here, and the code treated the two sides as one concept. `Message.message_thread_id` is *"for supergroups and private chats only"* — it is populated for **reply chains in non-forum supergroups and in private chats** — whereas the `sendMessage` parameter is *"for forum supergroups only"*. `Message.is_topic_message` exists precisely to separate them and was never read; `Chat.is_forum` was not even in the `TelegramMessage` type. The comment claiming the field is "absent in an ordinary group" is true for `chat.type === "group"` and false for `supergroup` and `private`.

The concrete case is mundane: a user DMs the bot and uses Telegram's *reply* on one of the bot's own messages — the ordinary way to answer a bot in a DM. That update carries `message_thread_id`. `chatRefOf` mints `"123456789:1"`, `recordChat` persists it, and it survives restarts. From then on **every** outbound send — every completed assistant message, every chunk of a long one, the approval notice, the test message — calls `sendMessage` with a thread id into a private chat, Telegram rejects it with `Bad Request: message thread not found`, and the fallback re-sends without it. Two round-trips per message, permanently, until an inbound message with no thread happens to repair the row.

The same shape hits non-forum supergroups, which is the *common* case given Group Privacy: with privacy on, the bot mostly sees replies to its own messages — exactly the ones carrying a reply-chain `message_thread_id`. It is a regression (those chats worked in one call before), and it doubles the call rate against Telegram's ~1 msg/s per-chat allowance, which the design spec already calls 正是招来 429 的形状.

**Fixed** in 2e4ab263 — `topicIdOf` returns a topic only when `chat.is_forum === true && is_topic_message === true`, feeding both `chatRefOf` and `replyRefOf`; the fallback stays as the deleted-topic safety net it was written for; and the doc now states the asymmetry instead of the wrong generalisation. Mutation-verified: reverting `topicIdOf` reddens three new tests (private chat, non-forum supergroup, forum General).

### 2. On-disk state changed meaning with no backward-compatibility entry

`messaging_bindings.last_chat_id` in an existing `web.db` was a bare chat id and is now a connector-opaque string that may carry a topic. The dev skill requires a change touching data already on disk to record the decision in a dedicated `backward-compatibility` pair, naming how long the tolerance stays — and the downgrade leg was unrecorded anywhere: roll back to ≤0.2.7 after chatting in a topic and `last_chat_id = "-100…:91"` reaches `wireChatId`, `Number(…)` is `NaN`, the string goes on the wire as `chat_id`, and every reply and test message fails with `Bad Request: chat not found` until the next inbound message rewrites it.

**Fixed** — a `2026-08-27-backward-compatibility.{md,zh.md}` pair now records it, including the downgrade behaviour and a one-line `UPDATE` to clear it. Worth noting what it says about lifetime: "how long does this stay" is the wrong question here, because a bare id is the only encoding a chat with no topic has, so the dual-format read is permanent semantics rather than compatibility code with a removal date. That is stated at the code site too.

### 3. The thread fallback's `catch` was unconditional and silent

Keyed only on "a thread was attached", so *any* first-send failure triggered an immediate second send, recording nothing. Two real costs: a `Too Many Requests: retry after 5` fired a second request with zero delay, worsening the flood-wait — one 429 costing two requests; and if Telegram accepted the message but the response was lost (the 15s deadline, a reset, a 5xx after delivery), the retry delivered the same chunk twice and the user saw it duplicated. Neither was possible before this PR.

**Fixed** — a new `TelegramApiError` carries `error_code`, the retry is narrowed to `400` with a thread attached, and the fetch-failure path deliberately stays a plain `Error` so a timeout or reset is never retried. The degradation is now observable: the send resolves a note, and the bridge logs `telegram delivered degraded: …` once per episode. It is deliberately kept off `lastDeliveryError` — the message *did* arrive, just in General.

### 4. The reply ref and the stored chat ref could name different topics

`deliverReply` snapshotted `row` once but re-read `entry.lastInboundMessageId` on every iteration, with an `await clientFor(...)` between. Before this PR both pointed at the same chat, so a disagreement only produced a wrong quote header; now they could name different **topics** — a user asks in topic 91, a second message lands in topic 77 during the await, chunk 1 threads onto the topic-77 message and chunks 2..N use the stale snapshot. One reply split across two topics, which is the exact defect this PR set out to remove, in a narrower window. Now snapshotted once beside `row`.

Worth knowing where that test lives: `messaging.test.ts`, not the Telegram file. The race window is the single `await clientFor(...)`, and `TelegramTransport.createClient` is synchronous, so the Telegram fake cannot hold it open deterministically; the Feishu fake's is `async`. The fix is in the channel-neutral bridge, so that is the honest place for it.

### 5. Tests covered only the chats where the encoding is valid

The fixture was hard-wired to a forum supergroup and modelled only "topic present" vs "absent", so the blocker was invisible to the suite. Five cases added, and the fake now counts **attempts** rather than successes — without that, the deleted-topic test could not tell one send from two, so "a send that carried no topic is never retried" was unpinned.

### Still open before merge

**Design.** `05-ARCHITECTURE.md:1277` needs `last_chat_id` described as connector-opaque; `:1281` needs the topic in both the chat ref and the reply ref plus the degrade-to-General retry; `:1280` needs the dedupe identity widened to `chatId:message_id[:topic]`. I am writing those.

### Checks

- Server **1101/1101** (Telegram suite 48, Feishu 36), typecheck clean, prettier clean. Six mutations confirmed red then restored.
- Being restacked: #497 and #499 have merged, so #505 goes onto `main` and this onto the new #505. It stays based on #505 — it genuinely uses that PR's note plumbing.

hiyouga:
**Rebased and merged.** The stack below this collapsed while it was in review — #497, then #499, then #505 all landed on `main` — so this went through three rebases. Two things came out of that worth recording.

**A shared changelog collision.** #508 and this PR had each independently created a `changelog/unreleased/2026-08-27-backward-compatibility.{md,zh.md}` pair. #508's landed first, so this one conflicted on a file neither author knew the other was writing. Merged into one per-batch entry, as the convention intends: both shipping PRs in `PR:`, this PR's downgrade `Breaking:` line kept, six sections mirrored across both languages, and a lead rewritten to describe all three pieces of state (`<root>/install-id`, `penguin.installId`, `messaging_bindings.last_chat_id`) rather than only #508's two.

**A green suite over a red typecheck.** The last rebase passed 1216 server tests while `tsc` failed. #507 widens the connector seam's `sendText` to `Promise<MessagingSendNote | void>`, and #501's QQ test — merged in between — declared its `spendImmediate` helper with a hand-written `{ sendText(...): Promise<void> }` structural type. The tests cannot see that; only the typechecker can. Fixed by taking `Pick<MessagingClient, "sendText">` so the helper tracks the real seam instead of restating it.

All three post-rebase checks confirmed: `topicIdOf` still gates on `is_forum && is_topic_message`; the thread-fallback retry is still `err instanceof TelegramApiError && err.errorCode === 400`, so a 429 or a lost response is never retried; and `deliverReply` still snapshots `entry.lastInboundMessageId` once before the await rather than per chunk.

Server 1216/1216, typecheck clean, prettier clean.

PR #508, opened 2026-08-27
fix(server,web): scope browser-persisted UI state to the data root that produced it

hiyouga:
Deleting the data files and restarting brought the app back with the old Workspace still selected.

## Two mechanisms, one fixed

The Workspace survives a data-root wipe through two independent paths:

1. **Server side — deliberate, unchanged.** `ProjectService.provisionInitialProject` adopts an existing `default_project` directory without overwriting its config. Delete `web.db` but not the project directory and `.project_config.toml`'s `[default_chat].workspace` is adopted as-is. That is intended behaviour and stays.
2. **Browser side — fixed here.** The Workspace also lives in `localStorage`, which has no relationship to the data root at all: in the desktop app it sits in Electron's userData directory, so no amount of deleting `PENGUIN_HOME` touches it. The keys are `penguin.chatDraft.<userId>.<projectId>` (carrying `workspace`, `agentId`, approval mode, model selection and the unsent composer text) and `penguin.sidebarWorkspaces.<projectId>`, among ~26 others — and **both ids are compile-time constants** (`ADMIN_USER_ID = "admin"`, `DEFAULT_PROJECT_ID = "default_project"`). A wipe-and-restart re-provisions the same user and Project, the keys match again, and the old state returns.

Only (2) is addressed. (1) is a separate, intentional seam.

## What this adds

- **`<root>/install-id`** — an opaque `randomUUID` minted the first time a data root is used, next to the existing `api-token`. It is an **identity, not a credential**: it authorizes nothing, contains no host name, user name or timestamp, and gets ordinary file permissions rather than `0600`. Only an *absent* file counts as a new root; any other read failure returns "unknown" so a transient error can never masquerade as a wipe, and a root that cannot be written also returns "unknown" rather than a fresh id on every boot.
- **`GET /api/install`** — public, mounted in the **platform** (`createApp`, just above its auth gate, the shape #502 used for the OAuth callback). A hot push carries platform + cli + web dist as one version and never the runtime, so the route travels with the bundle that calls it; mounted in the runtime shell it would have been missing on exactly the installations that received this web bundle by push. It is deliberately absent from `RUNTIME_PREFIXES` for the same reason — the platform must serve it, not decline it. The data root is still the runtime's; `deps.config.root` reaches the platform either way, so ownership of the file and ownership of the route are different questions. Public is a requirement, not a convenience: the browser must ask before it knows whether anyone is signed in, and a just-wiped root has nobody signed in at all. The id is read per request, so a root deleted under a running server is noticed on the next page load rather than only after a restart.
- **`packages/web/src/lib/install-scope.ts`** — one stored `penguin.installId`, compared on boot; a mismatch sweeps the install-scoped keys.

### Where the sweep runs, and why that is early enough

In `main.tsx`, **before `createRoot`**. The state it may clear is read from `useState` initializers scattered through the tree (sidebar pins and order, the composer's draft), and those run during the first render — a sweep arriving one effect later would let a stale draft be read once and then deleted underneath the component holding it, which is worse than either doing nothing or doing it in time. Before `createRoot` there is provably no component to have read anything.

It costs one same-origin request to the server this page was just served by. It is bounded by a 3s timeout: it can delay first paint, never prevent it. Giving up resolves to "unknown", which changes nothing.

### The sweep alone is not enough: two live copies of what it removed

**Modules evaluated before it ran.** ES module evaluation of the whole static import graph finishes before `main.tsx` runs a statement, so `dock-state.ts` — which parses `penguin.dock.layout` into module state at module scope — is already holding a copy of keys the sweep then deletes, and the first route resolution's `setDockScope` writes the entire pre-wipe map back, reopening a bottom dock pointing at a shell that belonged to the deleted root. The id matches by then, so nothing ever sweeps it again.

A boot that actually swept therefore **reloads instead of mounting**; the second pass evaluates every module against the cleaned store and reconciles to "unchanged". Making `dock-state` lazy would fix `dock-state` and leave the class open for the next module written that way, and a sweep only happens when the data root was actually replaced. A sweep whose marker could not be written back is reported apart (`swept-unrecorded`) and mounts, so the reload cannot repeat forever.

**Other tabs.** The sweep is per page load, so a tab already open against the old root keeps that state in React and module memory and restores it on the next pin, reorder or draft keystroke. The cross-tab `storage` event for `penguin.installId` now sweeps and reloads that tab too — only for an id *replaced* by a different id, since a first recording is another tab's adoption and a removal is site data being cleared.

### One id plus a sweep, not the id in every key name

Key names stay as they are, so no module that builds a key changes and nothing on disk needs migrating. Embedding the id would also create orphans forever — a key carrying a dead install id is something nothing ever reads or deletes, and each wipe would leave another full set behind. Walking the store collects orphans from *every* earlier root in one pass.

## Key classification

`grep -rhoE "penguin\.[A-Za-z]+" packages/web/src` finds 28 storage keys. They are classified in `KEY_RULES`, each with a one-line reason next to it. Two strings that grep finds are deliberately excluded: `penguin.chatRouteApplied.*` is `sessionStorage` (scoped to one tab's history, cannot outlive a data root) and `penguin.ooo` is the product's domain inside an example URL in a doc comment.

**Kept (browser preferences — 12):** `theme`, `fontScale`, `accent`, `currency`, `terminal.theme`, `lang`, `sidebarCollapsed`, `panelWidth`, `sidebarGroupMode`, `sidebarSortMode`, `sidebarNavGroupCollapsed`, `steerMode`. None names anything on the server; a new root can honour every one of them exactly, and wiping a data root is not a request to reset someone's theme.

**Swept (install-scoped — 16):** `chatDraft.*`, `chatDrafts.*`, `sidebarWorkspaces.*`, `pinnedSessions.*`, `sessionOrder.*`, `sessionSeen.*`, `groupOrder.*`, `sidebarCollapsedGroups.*`, `sidebarPinnedGroups.*`, `lastProjectId`, `lastAgentId.*`, `memoryCollapsed.*`, `modelsExpandedGroups.*`, `modelsGroupOrder.*`, `dock.layout`, `terminal.page.id`. Each references Session ids, Agent ids, Workspace paths, Project ids, terminal ids or per-Project provider group names.

Rules are **exact key or dot-terminated family**, never a plain prefix, because two pairs would otherwise be misclassified: `penguin.sidebarCollapsed` (preference) is a prefix of `penguin.sidebarCollapsedGroups.<projectId>` (install-scoped), and `penguin.terminal.` covers both `penguin.terminal.theme` (preference) and `penguin.terminal.page.id` (a terminal id).

An **unclassified** key is left alone. Never deleting something we do not understand is the safe default; the cost is that a new install-scoped key must be added to the table — which is now **enforced rather than asked for**: a test reads `packages/web/src` off disk, collects every `penguin.*` key literal (block comments stripped, since prose names key prefixes and the product's domain) and fails on any the table does not cover, listing the file it came from. The two deliberate exclusions above are on an explicit list, and a second test fails if either stops appearing in the source.

### The arguable ones

- **`penguin.dock.layout` → swept.** The entry is overwhelmingly per-conversation tab arrangement keyed by Session id and holding terminal ids, but it also carries `bottomRatio`, a size preference, in the same JSON. Sweeping loses that number. Splitting the entry would mean a stored-format change and a migration for a value the user re-drags in a second, so the whole key goes.
- **`penguin.modelsExpandedGroups.*` / `penguin.modelsGroupOrder.*` → swept.** Most provider group ids come from the built-in catalog, which ships with the app, so the arrangement would still "work" against a new root. But the keys are per-Project and include user-defined group names that live in that Project's config on disk. Project-scoped state goes when the Project is new; the cost of being wrong here is that the models page returns to catalog order.
- **`penguin.sidebarGroupMode` / `sidebarSortMode` / `sidebarNavGroupCollapsed` → kept.** These are sidebar state, and the sidebar is otherwise install-scoped, which is what makes them arguable. They are also documented in-code as single global user preferences ("switching it anywhere switches it everywhere") and none of them names a server entity — they are the *mode*, not the groups or the order, both of which are separate keys and are swept.
- **`penguin.panelWidth` → kept.** Its own module says it outright: "Width is a layout preference, not session data: it is never reset on a Session switch."

## Upgrade behaviour, and what a user must still do by hand

A browser that holds keys but has never recorded an install id **cannot be told apart** from one whose root was wiped. It therefore adopts the current id and sweeps nothing — destroying legitimate state on upgrade would be a far worse bug than the one being fixed.

Stated plainly: **shipping this does not clean state that is already stale.** A user who wiped their data root before this release still sees the old Workspace, and the only way to clear it today is by hand — clearing site data for the app's origin, which also resets the preferences listed above. Only wipes from this release onward are handled automatically.

That decision, and the fact that the adopt branch is **permanent semantics rather than removable compatibility code** (a browser with no recorded id is also a new profile, a new machine, or anyone who cleared their site data — nobody is on the hook to take it out later), is recorded in `changelog/unreleased/2026-08-27-backward-compatibility.{md,zh.md}`, which the main entry references. Dated 2026-08-27 because the messaging batch already owns the 2026-08-26 pair.

A "reset local UI state" action in settings would cover that gap in one click (it would call the same sweep with a forced mismatch, so it needs no new machinery beyond a button and its copy). Not built here — proposing it rather than adding unasked UI to a fix.

## Notes

- **Two roots, one origin** is the case that matters — the same host and port re-pointed at a different data root; different origins already have separate `localStorage`. A plain server restart on an unchanged root sweeps nothing, which is pinned both in the unit tests and in the manual run below.
- **Storage may throw or be empty.** Every read and write is wrapped and `localStorage` is resolved inside the `try` (the convention the existing storage modules already follow). A store that throws on every access resolves to "adopted" having stored nothing, destroys nothing, and lets boot proceed.
- `GET /api/install` is documented in `packages/docs/content/server-api.{en,zh}.md`.

## Verification

From the worktree root, rebased onto `main` (`8c93148f`):

- `pnpm -r build` — pass
- `pnpm -r typecheck` — pass (8 packages)
- `pnpm -r test` — pass: core 1021 passed/5 skipped, server 1147, web 1449, cli 374, desktop 144, docs 53, landing 69, skills 27. Includes the two files this PR adds: `packages/server/test/install-id.test.ts` (14) and `packages/web/test/install-scope.test.ts` (27).
- `prettier --check .` — "All matched files use Prettier code style!"

By hand, in a real browser (headless Chromium with a persistent profile, the built server serving the built web dist on one `localhost:7371` throughout, isolated root `~/.penguin/dev-data-installid`, since removed):

```
== phase 1: fresh root A ==
  install id A = be3eba7f-10a1-4268-9c6a-9b9a784e4e04
  <root>/install-id = be3eba7f-10a1-4268-9c6a-9b9a784e4e04
  localStorage after using the app on root A:
    penguin.installId = be3eba7f-10a1-4268-9c6a-9b9a784e4e04
    penguin.dock.layout = {"scopes":{},"bottomRatio":0.4}
    penguin.theme = dark
    penguin.chatDraft.admin.default_project = {"text":"","workspace":"","approvalMode":"allow-all","agentId":"default_agent","modelRef":…
== restart on the SAME root (must sweep nothing) ==
  install id after restart = be3eba7f-10a1-4268-9c6a-9b9a784e4e04 (same: true)
  every key survived the restart: true
== delete the data root, restart on the same host:port ==
  install id B = f272227f-6f13-44fe-9d17-dd48b55667ee (different: true)
  localStorage after reload on root B:
    penguin.theme = dark
    penguin.installId = f272227f-6f13-44fe-9d17-dd48b55667ee
```

The draft and the dock arrangement are gone, the theme survived, and the app lands on the login page — which is what a wiped root should look like.

## Since review

- The sweep no longer loses to state read before it ran: a swept boot reloads rather than mounts, and a tab left open across the wipe sweeps and reloads on the cross-tab `storage` event. `swept-unrecorded` keeps the reload from looping.
- `GET /api/install` moved from the runtime shell into the platform, so the fix reaches an install that received this web bundle by hot update instead of 401'ing at the platform's auth gate on every page load. It stays out of `RUNTIME_PREFIXES`, which would make the platform decline the route it now owns.
- The classification table's "add your key here" rule is enforced by a source scan. Also newly covered: the 3s timeout, a non-2xx answer (the `apiFetch` branch that parses the body and calls the global sign-out hook), a sweep that cannot record its marker, overlapping callers on a fresh root, and that the route sits in the platform's own table.
- The changelog entry was rewritten in past tense, `## On upgrading` and `## Not fixed here` folded into `## Details` / `## Compatibility`, and the upgrade decision moved into its own compatibility pair.
- Not done: hoisting `InstallScopeStorage` into one shared storage interface. The other six copies are `{getItem, setItem}` (plus `removeItem` for the draft cache) and none needs enumeration; this module needs `length` + `key` to walk the store. Hoisting would either widen six modules and their test fakes with members they never use, or introduce two shared types — neither is smaller than what is there.


hiyouga:
## Review

Reviewed in a dedicated worktree, with the live route probed and the module evaluation order traced by hand.

The core mechanism is sound and needed no rework: `randomUUID` from `node:crypto`, tmp+rename exactly as `auth/api-token.ts` does it, ENOENT-only minting so a transient error never reads as a wipe, per-request reads so the route follows the file, and a first-sight "adopt" branch meaning no existing user loses a single key on upgrade. I probed the route: `GET /api/install` answers 200 unauthenticated on the app host with no CORS headers, so a foreign page cannot read the id. Two blockers elsewhere.

### 1. The sweep was defeated for `penguin.dock.layout`, permanently

`main.tsx`'s reasoning — "before `createRoot` there is provably no component to have read anything" — is true of *components* and false of *module state*. `dock-state.ts:190` is a top-level `let { scopes, bottomRatio } = loadStored();`, i.e. a `localStorage.getItem` executed during module evaluation. `main.tsx` statically imports `./app` → `./router` → `app-layout` → `dock-state` (all static; the only dynamic imports in the package are xterm and shiki), so that read finishes **before** `syncInstallScope()` is even called.

The sweep then deleted the key while the parsed map stayed live in `scopes`. On the first route resolution, `setDockScope("new")` → `switchScope` → `persist()` wrote the **entire pre-wipe map back**, and `layout = scopes[scope]` reopened the old bottom dock with a terminal tab pointing at a shell from the deleted data root. `penguin.installId` now matched, so it was never swept again — the resurrection was permanent, and it is the same class of bug this PR exists to fix.

**Fixed** in 61896c27, and the remedy chosen is worth recording. Making `dock-state` lazy was the narrower fix, but its module state is touched from ~25 places through three chokepoints — one missed guard reinstates the bug, and it would fix exactly one module while leaving the class open. Instead `bootInstallScope()` returns `"reload" | "mount"` and `main.tsx` calls `location.reload()` on a sweep: the reload re-evaluates the whole graph against the cleaned store and holds however the import graph grows, at the cost of one extra page load only when the data root was actually replaced.

The loop that remedy invites is closed too: `reconcileInstallScope` reads the marker back after writing it and returns a distinct `"swept-unrecorded"` when it did not stick, so a sweep that cannot record its marker mounts rather than sweeping-failing-reloading forever. Both halves mutation-verified.

### 2. Mounted in the runtime shell, so the fix could not reach a hot-updated install

`hmr/host.ts`'s module doc is explicit that a hot push carries platform + cli + **web dist** and never the runtime. Mounting `/api/install` in `createRuntimeApp` meant the new web bundle could arrive by push while the endpoint serving it could not: the request would miss the runtime's table, fall through the seam, not be in `RUNTIME_PREFIXES`, and hit the auth gate → 401. `syncInstallScope` catches it and resolves `"unknown"`, so nothing breaks — the fix is just silently inert, and every page load spends a failed request on it, until a new runtime is installed. #502 had shipped the opposite precedent hours earlier.

**Fixed** — the route now sits in `createApp` immediately above the auth middleware, so it travels with the bundle that calls it. Mutation-verified: moving it back reproduces the 401 exactly.

`RUNTIME_PREFIXES` was deliberately **not** extended, which is the right call now that the mount moved — adding it there would make the platform decline the route it owns. That reasoning is now written at the mount site rather than left to be rediscovered.

### 3. The table's own rule was unenforced, and three failure paths untested

`install-scope.ts:70` says "ADDING A KEY: add it here too… a forgotten install-scoped key silently reintroduces the bug this module fixes" — but the completeness test only walked a hand-written fixture, which a new key never enters. A new install-scoped key would have shipped unclassified with a green suite.

Now enforced by a source scan: it reads `packages/web/src` off disk, strips block comments (prose there names key *prefixes* and families without their dot), collects all 31 distinct `penguin.*` literals, and asserts each is classified or on a two-entry exclusion list with a stated reason — plus a second test that fails if an exclusion stops appearing in the source. Mutation-verified by adding a fake key to an unrelated module.

The three untested failure paths are covered: the 3s timeout under fake timers (not settled at 2999ms, `"unknown"` at 3000ms, store untouched), a non-2xx response — a different `apiFetch` branch from the connection-refused test, and exactly the old-runtime path from finding 2 — and two concurrent mints on one fresh root.

### 4. A tab left open across the wipe re-persisted its pre-sweep state

The sweep is per-page-load, so another tab already open against the old root kept writing its in-memory state back, and was never swept again because the id then matched. `watchInstallScope()` now wires a `storage` listener that sweeps and reloads — firing only for an id genuinely *replaced*, since `oldValue === null` is another tab's adoption and `newValue === null` is site data being cleared.

### 5. Changelog

Present tense throughout, and `## On upgrading` invented a heading where the spec says to reuse `## Compatibility` first. Rewritten past tense; the upgrade decision moved into its own `2026-08-27-backward-compatibility.{md,zh.md}` pair per the dev skill, which names the lifetime honestly: the adopt branch is **permanent semantics, not removable compatibility code**, because a browser with no recorded id is also a new profile, a new machine, a private window, or anyone who cleared site data. Nobody is on the hook to remove it.

Note the slug collides with the messaging batch's `2026-08-26` pair, so this one carries today's date and disambiguates in its H1; the existing pair is untouched.

### Not done, with reasons

`InstallScopeStorage` is the seventh copy of a minimal storage interface — but the other six are `{getItem, setItem}` and none needs enumeration, while this one needs `length` + `key` to walk the store. Hoisting would either widen six modules and their fakes with members they never use, or introduce two shared types. Neither is smaller than what is there.

### Still open before merge

**Design.** This adds a public HTTP route, a new data-root artifact and a defined lifecycle for browser-persisted state — all three categories `CLAUDE.md` names. `05-ARCHITECTURE.md:1209` is where data-root artifacts are spec'd (`initial-admin-password` in exactly this shape) and `06-PROTOTYPE.md:216` already spec's localStorage-vs-origin for the desktop shell. I am writing that.

### Checks

- Rebased onto `main`; only `main.tsx` truly conflicted (#503's katex import kept, the `mount()` restructure re-applied around it).
- Web **1449/1449**, server **1147/1147**, whole monorepo green, both typechecks clean, prettier clean. **CI 22/22.** Every mutation confirmed red then restored.

PR #509, opened 2026-08-27
feat(desktop): probe update download source

laodouuu:
## Summary
- Adds desktop updater source selection that probes the release manifest before choosing GitHub or OSS for update downloads.
- Keeps explicit source overrides intact while making the default path probe-enabled and GitHub-fallback-friendly.
- Retries failed update downloads on the same tag's fallback feed and refreshes source selection before checks so a running app can discover newly published releases.

## Changes
- Adds desktop update probe support in `packages/desktop/src/update-source.ts` and release metadata helpers in `packages/desktop/src/update-support.ts`.
- Wires the selected feed into `packages/desktop/src/updater.ts`, including explicit source modes, manual-check error handling, and download-stage fallback.
- Covers probe/source behavior with focused desktop updater tests.

## Test plan
- `pnpm --filter @prismshadow/penguin-desktop exec vitest run test/update-source.test.ts test/update-support.test.ts test/updater-status.test.ts`
- `pnpm --filter @prismshadow/penguin-desktop typecheck`
- `pnpm --filter @prismshadow/penguin-desktop build`
- `pnpm build`
- `pnpm format:check`
- `pnpm typecheck`
- `git diff --check`
- Fork release/update test through `v0.2.7-fork-desktop-update.7`: verified source selection, normal update downloads, download fallback, and discovering a newer release without restarting the app.

`pnpm test` was also attempted during the full-repo check; the local Windows run failed on existing symlink/EPERM behavior plus an unrelated server memory export ordering assertion, while the focused desktop updater tests passed.

### DIFF
--- a/packages/desktop/src/update-source.ts
+++ b/packages/desktop/src/update-source.ts
@@ -1,0 +1,364 @@
+/**
+ * Update-source selection for the desktop auto-updater.
+ *
+ * This module owns the same source-neutral speed-probe contract the release installers
+ * and download page use: fix the target tag first, read that tag's
+ * release-download-manifest.tsv, prove both sources can serve the shared probe file, and
+ * then compare the large probe using the current installer rule. GitHub is the free
+ * source: it wins outright at or above the minimum speed, and the OSS mirror takes over
+ * only when GitHub misses that minimum and OSS is clearly faster.
+ *
+ * The desktop default feed is still the GitHub Releases feed electron-builder publishes.
+ * An inconclusive test never blocks updates; the caller keeps that default.
+ */
+import { createHash } from "node:crypto";
+
+export const OSS_ORIGIN = "https://penguin-harness-releases.oss-cn-beijing.aliyuncs.com";
+export const OSS_RELEASE_ROOT = `${OSS_ORIGIN}/releases`;
+export const GITHUB_RELEASE_ROOT =
+  "https://github.com/Prism-Shadow/penguin-harness/releases/download";
+export const OSS_LATEST_JSON_URL = `${OSS_ORIGIN}/latest.json`;
+
+/** Same contract as the release installers (install.sh / install.ps1). */
+export const SPEED_PROBE_MANIFEST_TIMEOUT_SECONDS = 5;
+export const SPEED_PROBE_SMALL_TIMEOUT_SECONDS = 5;
+export const SPEED_PROBE_LARGE_TIMEOUT_SECONDS = 8;
+export const SPEED_PROBE_TOTAL_TIMEOUT_SECONDS = 26;
+export const SPEED_PROBE_GITHUB_MIN_BYTES_PER_SECOND = 262144;
+export const SPEED_PROBE_OSS_SWITCH_RATIO = 1.5;
+
+/** Minimal structural slice of fetch so callers can inject Electron's net.fetch or a fake. */
+export interface ProbeFetchInit {
+  headers: Record<string, string>;
+  signal: AbortSignal;
+}
+
+export interface ProbeResponse {
+  ok: boolean;
+  status: number;
+  headers: { get(name: string): string | null };
+  arrayBuffer(): Promise<ArrayBuffer>;
+}
+
+export type FetchFn = (url: string, init: ProbeFetchInit) => Promise<ProbeResponse>;
+
+export interface ReleaseProbe {
+  name: string;
+  size: number;
+  sha256: string;
+}
+
+export interface ReleaseDownloadManifest {
+  smallProbe: ReleaseProbe;
+  largeProbe: ReleaseProbe;
+  /** Integrity check that the manifest belongs to a release carrying this update asset. */
+  assetSize: number;
+}
+
+/** The update asset electron-updater replaces on this platform (dmg is manual-only). */
+export function desktopUpdateAssetName(platform: NodeJS.Platform, arch: string): string | null {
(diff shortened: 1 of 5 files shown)


hiyouga:
## Review

Reviewed in a dedicated worktree against `origin/main`. The module is carefully written — the probe contract genuinely mirrors the installers' (identity encoding demanded, exact byte count, sha256, an absolute deadline threaded through every leg), the manifest parser is strict about every field it reads, and every failure path lands on "keep the default feed" rather than on a guess. One finding I would not ship, and three smaller notes.

### 1. A successful probe narrows update *discovery* to whatever the OSS mirror has published

`selectAutoUpdateFeed` fixes a tag from `resolveOssLatestTag` — OSS's `latest.json`, and nothing else — then hands back a feed URL pointing at that single immutable tag directory. `setGenericFeed` passes it straight to `autoUpdater.setFeedURL({ provider: "generic", url })`, so electron-updater reads `latest.yml` from that one tag and can see no other release.

That is right for an **installer**, which is fetching a version the user already named. It inverts for an **auto-updater**, whose entire job is discovery. The spec draws that line at `05-ARCHITECTURE.md:1192`: the `auto` contract is 先经 `latest.json` 锁定不可变 tag、优先 OSS 并按同 tag 回退 GitHub — *locking* a tag, which is what an install wants and the opposite of what a check wants.

The release workflow guarantees the gap exists. `release.yml:37-39` states it outright: `mirror-oss` downloads the finished GitHub Release assets, verifies them, mirrors them, and *"latest.json is uploaded last"*. So between publishing a release and the mirror job finishing, GitHub has the new version and OSS's `latest.json` does not — and in that window every desktop app pins itself to the **previous** tag, on both sources, and reports no update available. If `mirror-oss` fails (the workflow's own header contemplates re-running it), the window lasts until someone notices.

Two things make this quiet rather than loud:

- The escape hatches are error-only. `switchToFallbackFeed` is called from exactly two places — `updater.ts:333` (feed unavailable) and `:424` (download failed). "The pinned feed says I am up to date" is not an error, so nothing reconsiders it.
- The **inconclusive** path is safer than the successful one. `keep-github` calls `setDefaultGithubFeed`, which restores electron-updater's GitHub provider — the one that discovers the newest release on its own. So a probe that fails keeps working discovery, and a probe that succeeds gives it up. That inversion is the part I would want changed.

`refreshUpdateFeed()` before each check does bound it: once the mirror catches up, the next check moves. So this is a delay, not a permanent stall — but it is a silent one, and it makes desktop update latency a function of a mirror job the user cannot see.

**Suggested shape:** let the probe choose the *source* without also choosing the *version*. Resolve the newest tag from the same place the default feed would — GitHub — and use `latest.json` only to confirm the mirror has that tag before pinning to it; if OSS does not have it yet, keep the default GitHub feed rather than pinning backwards. That keeps every speed benefit and removes the discovery regression.

### 2. No changelog entry

`git diff origin/main...HEAD -- changelog/` is empty. The repo rule is a bilingual pair in `changelog/unreleased/` shipping **inside the PR that makes the change**, and this one changes where every desktop client fetches its updates from — squarely `Type: feature`, `Scope: desktop`.

### 3. No design update, and the spec has a place for it

`05-ARCHITECTURE.md:1192` records the source-selection contract for the installers, and `:1197` extends the same contract to `penguin update`. The desktop auto-updater is explicitly *outside* it today — `06-PROTOTYPE.md:228` says 桌面的更新由 electron-updater 接管. This PR brings it inside, which is exactly the kind of change `CLAUDE.md` wants recorded in the same breath. Whatever shape finding 1 lands in is what the spec should say.

### 4. Two small ones

`probeSource` calls `arrayBuffer()` and only then checks `body.byteLength !== opts.probe.size`, so the whole response is buffered in the main process before the size it was supposed to be is enforced. The declared size is only validated as a positive safe integer, so a wrong or corrupted manifest bounds the download by the 8s timeout rather than by any byte cap. Low likelihood — both sources are ours over HTTPS — but a cap before buffering is cheap.

`ReleaseDownloadManifest.assetSize` is documented as an "integrity check" and is load-bearing only as a *presence* check: the parse fails when the platform's `desktop_asset` row is missing, but the value itself is never compared to anything afterwards. Worth either using it or saying in the comment that presence is the check.

### On the tests

612 test lines for a 364-line module, and the redundancy is structural rather than a matter of count: **the decision rule is pinned twice.** `selectMeasuredSource` — a three-line pure function — gets its own two cases ("keeps GitHub above the minimum speed and on ties below it", "switches to OSS only when GitHub missed the minimum and OSS is more than 1.5x faster"), and then the same rule is exercised again through `selectAutoUpdateFeed` by four more ("keeps GitHub when the GitHub large probe meets the 256 KiB/s gate", "promotes the OSS mirror when GitHub misses the minimum and OSS is clearly faster", "keeps GitHub when both sources miss the minimum and OSS is not enough faster", "uses the OSS mirror when the GitHub large probe fails and OSS can be measured"). The same doubling happens for `latest.json` validation, covered at the unit level and again through the tree.

The decision-tree cases are the ones worth keeping — they pin the module's actual contract, over untrusted-ish remote input, where a wrong branch means the wrong download source. The unit-level duplicates of those same rules can go.

For finding 1, the test I would want is **one**: a probe where OSS's `latest.json` names an older tag than the one GitHub is serving, asserting the app does not pin itself backwards. That is the behaviour at stake; the rest of the tree is already covered.

PR #510, opened 2026-08-27
Red dots for Skill updates, preset-model updates and unexpected errors — and a way to clear them

hiyouga:
The phone-style red dot that already marked a software or Agent-kernel update now also marks three
things the user is expected to act on but might reasonably choose not to: **Skill library**
updates, **model library** preset updates, and **unexpected errors** in the cost center. Each dot
leads down an unbroken path to the control that deals with it, and each one can be put down.

## The three trails

- **Skill library** — the sidebar's Skills entry and the collapsed rail's Skills icon, ending at
  the cards whose "update installs" button reinstalls the library copy on every Agent that is
  behind. The gate rides along on the Agent list (a new `skillUpdates` field), so it costs no
  request of its own; running the update reloads that list and the dot goes down with it.
- **Model library** — the Models entry, ending at the "sync presets" button, which carries the dot
  on the control that acts. Owner-only, like that button: a member has no control at the end of
  that trail, so the badge never appears for one and the probe behind it is not even requested.
- **Cost center** — the Cost Center entry, ending at the errors panel. `unexpected` only (a
  business 4xx is the server saying no on purpose), over a trailing 7-day window.

The mobile menu button still carries one dot for everything; with an unexpected error in the mix
its wording widens from "updates are available" to "something needs attention".

## Why a dismissal, and what it stores

The two existing trails clear themselves — install the update and the badge is gone. These three
can end in a decision *not* to act: a model table deliberately kept off the catalog, an error read
and understood. A dot with no way down stops meaning anything, so all three pages carry the same
notice directly under the page title — a `toneStrip.attention` block holding the red dot the user
followed, the trail's own sentence, and **Mark as handled** (**Mark as read** in the cost center).

What a dismissal stores is not "hidden" but the **signature** of what was waved away — the Skill
names with the library versions on offer, the model references that would change, the newest
error's timestamp. The badge comes back only once what is waiting has grown **beyond** that, which
is containment rather than equality: dismissing a batch and then acting on one of its items leaves
the rest dismissed, and rows leaving the error window (the row cap, an Agent deletion) never raise
the dot again. Dismissing answers "not this one", never "never tell me again".

Markers are stored per user in `ui_prefs` under `todoDismissed`, keyed by Project, beside
`initialPasswordBannerDismissed` — the store is free-form JSON, so this needs no schema or API
change. They follow the account rather than the browser: on a shared workstation one person's
dismissal does not put the dots down for the next, and a second device or tab picks up what the
first one cleared.

## The one thing that had to not drift

The Models badge and the "sync presets" button must never disagree about whether there is anything
to do — a dot leading to a button that answers "already up to date" is exactly the failure this
repo's badge rules forbid. `catalog-sync.ts` gained `catalogDelta`, which asks the sync action's
own question of a **saved** model table rather than the page's row state, and a test replays every
table shape through both `catalogDelta` and `syncRowsWithCatalog(rows.map(toRow))` to pin them
together. `refreshProjectTodos` bumps a per-Project generation so a probe still in flight from
before a sync cannot land on top of the answer from after it and reinstate the pre-sync dot.

## Server

- `AgentSummary.skillUpdates` — installed Skills the bundled library has moved past, each with the
  version on offer. Computed in the pass that already counted them, and bounded to the first 4 KB
  of each `SKILL.md`, which is where the frontmatter block sits: a preinstalled library is roughly
  180 KB of `SKILL.md` per Agent, and the Agent list is reloaded after every Skill update. Skills
  the library does not carry (a zip, a picked directory) are never listed — there is no library
  version for them to be behind.
- `GET /usage/errors` takes an optional `kind` (`unexpected` / `expected`), which is how the badge
  asks for one row instead of pulling the whole dashboard aggregate for two numbers. A `kind`
  outside that set is a 400, not a silently empty page that would read as "no errors".
- `UiPrefs.todoDismissed` declares the markers' shape on the free-form prefs blob. `PUT
  /api/me/prefs` merges only at the top level, so the client holds every Project's markers and
  writes the whole map.

Both routes are platform routes, replaced together with the frontend on a hot update, so there is
no version skew to handle. Nothing on disk or in config changes shape, and the prefs key is new,
so there is no compatibility burden.

## Also in here

The Agents list's "Kernel update needed" capsule is paler, as asked. Its fill is now measured
against the plain gray `Badge` sharing the row — 1.09 : 1 against the card in light and 1.11 : 1
in dark, where that neighbour shows 1.10 : 1 and 1.18 : 1 — instead of sitting a fifth louder than
it. The ink is unchanged and still clears 7.67 : 1 in light and 9.13 : 1 in dark.

## Review round

Fixes from review, in one commit on top:

- **A dismissed batch stays dismissed.** `raisedTodo` compared the whole joined signature, so
  acting on one of two dismissed Skills raised the dot for the other. A `Todo` now carries the
  `items` its signature is joined from, and only an item absent from the dismissed set raises it.
  The errors trail compares ordered instead (`match: "watermark"`): `ErrorsRepo.deleteByAgent` can
  remove the newest row, and a timestamp that moved *down* is not news.
- **One notice shape.** `TodoNotice` moved out of `update-dot.tsx` — whose contract is a
  decorative, absolutely positioned, `aria-hidden` mark — into `components/ui/todo-notice.tsx`,
  restyled to the `toneStrip.attention` block the other attention strips already use, and given
  its own top margin so no call site supplies one. All three pages render it directly under their
  `<h1>`. The model library's separate "mark as done" button is gone; the dot stays on the sync
  button it acts on. The cost center's copy no longer sits between a stats row and a table that
  count a different window from the probe.
- **Markers moved to `ui_prefs`.** The earlier comment claimed a server-side marker "would need
  schema and API changes"; it does not, and per-browser storage let one account clear another's
  dot on a shared machine.
- **`GET /agents` reads a bounded prefix** of each `SKILL.md` rather than the whole file.
- **Copy:** the EN dismiss label is "Mark as handled", matching what the zh already said
  (「标记为已处理」) — "done" asserted an update that was not performed.
- The changelog pair was rewritten in past tense, with the rationale and the two false claims
  removed.

## Verification

First round: `pnpm -r build`, `pnpm typecheck`, `pnpm test` (4401 tests) and `pnpm format` +
`pnpm format:check` all pass on Node 24.

Review round (only `packages/web` and `packages/server` changed): `npx tsc --noEmit` clean for
both, `npx vitest run` green — web 1488 tests across 107 files, server 1196 across 95 —
and `prettier --check` clean on every changed file.

Unit tests: `web/test/todo-badges.test.ts` (the three gates and the dismissal rule — distinct
Skill counting across Agents, order independence, the highest library version winning when two
Agents are behind by different amounts, an Agent list from a server too old to send the field, a
preset delta separating two same-sized changes, an error window whose oldest rows were evicted,
the rest of a dismissed batch staying down after one of it is updated, and the marker parse and
merge); `web/test/catalog-sync.test.ts` (the `catalogDelta` cases plus the agreement test above);
`web/test/update-badges.test.ts` rewritten for the generalized combined note, including the
update-only / not-an-update split.

Integration tests: `server/test/skills.test.ts` (`skillUpdates` empty when current, naming the
library version once behind, cleared by the reinstall the trail ends on, never listing a
non-library Skill, and an unreadable `SKILL.md` counting as neither installed nor an update);
`server/test/errors.test.ts` (the `kind` filter's counts and ordering, no-`kind` unchanged, and an
invalid `kind` rejected).

Checked by hand before the review round, against a dev server on its own data root, with two
Skills aged on disk, four catalog entries removed from the model table and three unexpected errors
seeded: all three dots on the right nav rows and none on Agents; each page's notice reading what
the dot's tooltip said; dismissal clearing its dot and surviving a reload; **Sync presets**
clearing the Models dot with no dismissal at all; updating one of two outdated Skills leaving the
dot up at "1 skill update" and the second clearing it. The review round was not re-driven in a
browser — it is covered by the type checker and the two suites above.


hiyouga:
Placement corrected per review — each mark now sits where its trail ends rather than in a band of its own.

- **Cost center** — the line moved out from under the page title onto the errors panel, directly above the table it counts. It names rows, so it belongs against the rows; the stats line above already answers "how many" and this one answers "and what do I do about it".
- **Skill library** — each outdated card's update button now carries the dot on its top-right corner, straddling the border the way the kernel update button does. The summary line under the page title stays, since the Skills waiting are spread across the cards below it.
- **Model library** — the notice row is gone. The dot rides the **Sync presets** button itself and **Mark as done** sits beside it in the header row that already existed.

Verified by hand on a dev server with two Skills aged on disk, four catalog entries removed and three unexpected errors seeded: all three marks render in the new positions, and dismissing from the model header still clears the sidebar dot. `pnpm typecheck`, `pnpm test` (web: 1491) and `pnpm format:check` pass.

hiyouga:
## Review

Reviewed in a dedicated worktree, with the gate functions run directly rather than reasoned about.

**The architecture is sound.** The signature-instead-of-a-flag idea is right, `catalogDelta` genuinely shares the sync button's own union so the Models dot cannot lead to an "already up to date" button, `error_records` queries stay on `idx_error_project_date`, `SkillUpdateRef` has no optional fields on the wire, and the web guards `skillUpdates ?? []` for older servers. The defects were in the dismissal state machine and the placement.

### 1. Dismissing a batch, then acting on one item, raised the dot again for the rest

`raisedTodo` compared the whole joined signature with `!==`, so **shrinking** the set counted as "something new". Run against the shipped functions:

```
before        : {"signature":"penguin-sdk@2,web-design@3","count":2}
after dismiss : null
after update  : {"signature":"web-design@3","count":1}   ← back, for an update already declined
```

Two Skills behind, Mark as handled, then update one — and the dot returns for the other. Same for Models: dismiss three preset refs, add one by hand, and it raises for the remaining two.

**Fixed** in e7fbe88c — `Todo` now carries the `items` the signature was already joined from, and `raisedTodo` tests set containment rather than string equality. `count` still carries the full waiting count, so the notice keeps agreeing with the number of update buttons on the page.

### 2. The cost-centre notice counted over a different window than the stats and table it sat between

`use-project-todos.ts` used a fixed `presetRange("7d")` with no agent filter, frozen at first fetch, while the stats row above it and the table below it both re-fetch under the page's `preset`/`agentFilter` state. So: open on 7d showing "Unexpected 3" and "3 unexpected errors"; switch to "Last hour"; stats read **0**, the table renders empty, and the notice between them still says **3**. Picking 90d gives the inverse. The changelog asserted the opposite — "so the table always holds the rows the dot is talking about".

**Fixed** by the restyle below: moving the notice to page level removes the mismatch, so no filter plumbing was needed.

### 3. The per-browser marker rested on a false claim, and let one account clear another's dot

`todo-dismissals.ts` justified `localStorage` with "a server-side marker would need schema and API changes". It would not: `PUT /api/me/prefs` already shallow-merges free-form JSON into a per-user blob typed `[key: string]: unknown`, and this app already stores a dismissal there — `initialPasswordBannerDismissed`.

The cost was real: on a shared workstation, the owner clears "5 preset models to sync" and a member logging in afterwards never sees what was waiting; the same user on a second device keeps a dot they already cleared.

**Fixed** — dismissals moved to `ui_prefs.todoDismissed`, read once per session and written fire-and-forget, which also gives cross-tab correctness on reload. Because `PUT /me/prefs` merges only at the top level, the store holds and writes every Project's map. Until the read lands the map is `null` and every gate stays closed, so no dot flashes up and vanishes. The now-dead `penguin.todoDismissed.` rule was removed from `install-scope.ts` — worth noting, since that table's own comment warns that a forgotten key silently reintroduces the bug it exists to fix.

### 4. Three smaller correctness ones

Deleting an Agent re-raised an errors dot already read: the doc claimed rows are "only ever evicted oldest-first", but `deleteByAgent` can remove the newest, so equality against the stored timestamp failed downwards. Now a watermark comparison — raise only when newer — which covers eviction and deletion together.

A late-landing first probe could overwrite a post-sync refresh: `refreshProjectTodos` cleared the cache and started a second fetch, but the first fetch's `.then` still wrote unconditionally, so an out-of-order resolution reinstated the pre-sync answer and the Models dot reappeared pointing at a button that would say "already up to date" — the exact failure this PR set out to prevent. Now a generation guard, on both the `.then` write and the `.catch` eviction (an unguarded catch would have cleared the refresh's own promise).

`GET /agents` had started reading every installed SKILL.md in full, while its comment called it "a read of the same few kilobytes". Measured: 180 KB across 17 files per Agent, of which only the head frontmatter is used — largest block 993 B. Now a bounded 4 KB prefix read, with the comment stating the real cost.

### 5. The changelog

Present tense throughout, describing the current codebase — the shape `changelog/README.md` bans by name — plus a `## Clearing a dot` section that was largely the decision rationale the banned `## Why`/`## Decision` headings exist to keep off disk. Rewritten past tense in both languages, rationale stripped, and the two false claims (the 7-day/table agreement, the schema-and-API justification) removed.

---

## The unified notice

All three now render the same `TodoNotice` (`components/ui/todo-notice.tsx`) directly under their page title, as a `toneStrip.attention` block with the 6 px red dot inside:

```
flex items-center justify-between gap-4 rounded-lg border px-4 py-3 ${toneStrip.attention}
```

`attention` is `tone.ts`'s "unfinished — waiting on time, a queue, or the user", and `toneStrip` is "a bordered notice that owns a row" — so this is the house pattern for exactly this shape, already used verbatim in three other places, rather than a new look. The red stays inside because red is the trail's identity and amber is the notice's shape.

Before this, they were three different things: Skills was a bare flex row, the Model library was **not a notice at all** (a dot on the sync button plus a separate ghost button over in the header control row), and the Cost centre's sat about a screen of charts below its `<h1>`.

Two placement notes worth recording:

- **Models** is placed after the header *row*, not literally after `</h1>` — that `</h1>` sits inside a `flex flex-wrap justify-between` row, so inserting there would have made the notice a third flex item beside the title instead of a block below it. The row's `mb-4` moved onto a wrapper holding row + notice, which gives the intended result and keeps the gap to the table.
- **The Models sync-button dot stayed put.** It marks the control that acts, which is the one mark that should not move; its comment now says so.

`TodoNotice` moved out of `update-dot.tsx`, whose entire contract is "absolutely positioned, layout-neutral, `pointer-events-none`, `aria-hidden`, decorative" — the PR had bolted an apologetic parenthetical onto that file's header conceding this was "the one mark here that is NOT positioned". Once the thing has a ground and a border it is a layout block, not a mark. The red is exported as `UPDATE_DOT_INLINE` so it stays written once, and that file's "not a status tone" paragraph now records that the carve-out is about the mark, not the block around it.

One copy fix: EN said "Mark as done", which asserts the update was performed; the zh 「标记为已处理」 was the accurate one. EN is now "Mark as handled".

### Still open before merge

**Design.** `06-PROTOTYPE.md:42` is currently the only statement of the badge convention in the whole spec and describes exactly one trail (the avatar's 软件更新 case). Nothing there mentions nav-row dots, Skill or preset-model updates, unexpected-error dots, or a dismissal affordance. Four edits are needed — the convention at line 42, plus the Skills, Models and Cost-centre page sections. I am handling this.

### On tests

**Net down, not up.** One case was added — the shrink scenario that pins finding 1 — and the dismissals block went 6 → 3, because finding 3 deleted the `localStorage` plumbing three of those cases existed to cover. The `withDismissal` duplicate collapsed into a single merge case that also asserts another Project passes through untouched.

### Checks

- Web **1488/1488**, server **1196/1196**, both typechecks clean, prettier clean on all 18 changed files. CI: 21 pass, 1 skipped, 0 failures.
- The change has not been driven in a browser — it is one shared component at three call sites, type-checked and suite-green. Worth a look on your side before merging, since the placement is the point.

PR #511, opened 2026-08-27
Messaging stops reporting what is not a problem

hiyouga:
Two failures reported from a live install, both filed on the cost-center dashboard as **unexpected**:

```
[messaging] messaging_image_fetch_failed
  Access denied. One of the following scopes is required:
  [im:message.history:readonly, im:message:readonly, im:message] … (code 99991672)

[messaging] messaging_file_send_failed
  QQ cannot receive "helloworld.md": sending a file to QQ requires a publicly
  reachable URL for it, which this server has no way to provide
```

Neither is a defect. The first is a Feishu app whose scopes have not been granted yet — and the
refusal already reaches the chat carrying the scope names and the console link to grant them. The
second is a permanent shape of the QQ platform: its rich-media path registers bytes against a
publicly reachable https URL, the same requirement that ruled webhook mode out, and this server has
no such address to offer. It will refuse the next file identically.

## The cause

`MessagingBridge.recordError` never passed a `kind`, and `ErrorRecorder`'s default for a non-HTTP
source is `unexpected`. So *every* messaging failure — a scope to grant, a file too large, a
platform limitation, a genuine bug — was filed as "shouldn't happen, needs a human".

That is not cosmetic. The recorder's own stated criterion is **"does a human need to step in"**, and
the cost center highlights the unexpected count. A channel filing its ordinary refusals there points
that highlight at nothing anyone can fix, and buries the failures that genuinely do need looking at.

## The rule

`runtime/messaging/error-kind.ts` classifies a messaging failure, with the reasoning for each case
next to it. It reads **two** things: the error's type, and the `code` of the capture point that
caught it.

Three types are understood:

- `MessagingPermissionError` — a scope the bot's app was never granted.
- `MessagingMediaTooLargeError` — a transfer over the cap; the sender fixes it by sending something
  smaller.
- `MessagingUnsupportedError` (new, in `media.ts`) — the channel structurally cannot carry it and
  says why. No permission to grant, no smaller file, no retry that behaves differently. QQ's
  outbound media refusal now throws this instead of a bare `Error`.

The type alone cannot decide, because it says what went wrong and not whether anyone was told.
`larkFailure` throws the same `MessagingPermissionError` for **every** Feishu call refused with
99991672 — the inbound image download, the outbound upload, *and the plain text send*. Only the
first two are followed by a notice into the chat, so only `messaging_image_fetch_failed` and
`messaging_file_send_failed` file a typed refusal as `expected`.

`messaging_send_failed`, `messaging_inbound_failed` and `messaging_connect_failed` stay `unexpected`
whatever they throw. On a send there is nothing to be told with — the message that would carry the
notice is the message being refused — so an app granted the receive scopes but not
`im:message:send_as_bot` takes every question in the chat and answers none of them, silently. That
is the worst state this feature has, and the record is the only place it surfaces.

Everything else stays `unexpected` too. It is deliberately a small allowlist of **typed** failures at
**named** capture points rather than a message match: anything the connectors have not classified
keeps its place on the dashboard, which is the safe direction — a real fault miscounted as routine is
invisible, while routine noise miscounted as a fault is merely loud. Deciding by type rather than by
wording also keeps the count from depending on how a platform phrases its refusal.

The records are unchanged in every other respect, including still being written: a scope nobody
grants is still worth finding on the dashboard, it is simply not a defect.

### QQ's other refusal

QQ has a second refusal two hundred lines from the media one: `enqueue` throws when the chat has no
message that can still be replied to, which is the ordinary case for a turn driven from the Web App.
It **deliberately stays a bare `Error`**, and `refuseMedia` and the throw site now both say so. It is
the same shape as the Feishu send above: the reply text is dropped and the chat cannot be told,
because the message that would carry the reason is the message being refused. Somebody does have to
notice that, which is what `unexpected` is for — and `MessagingUnsupportedError`'s own contract
("`message` must say what could not be carried and why, because that text reaches the chat") is one
it could not keep there. The media refusal costs nothing by contrast: the reply's text still arrives,
carrying the notice that names the file.

## A mention that names no file is no longer announced in the chat

The reply-file feature scans a reply's prose for path-like tokens and sends the ones that resolve
inside the Workspace. Names that resolved to nothing were announced — *"Named in the reply but not
sent — no such file inside this Session's Workspace: hello-world.md"* — on the reasoning that a
promised file which never arrives, with nothing to say why, reads as broken.

That reasoning holds only when something was promised. The rule reads prose, and prose names files
for reasons that have nothing to do with delivery: a model describing a `hello-world.md` it went on
not to write, or quoting a path out of a document it read. Announcing those puts an error-shaped line
under replies that were entirely correct, and leaves the reader no move to make.
`messagingFilesMissingNotice` became `messagingFilesMissingLog`: the names go to the server log
instead. A file that EXISTS and whose upload failed is still named in the chat — there the reply
promised something that then did not arrive, which is the one case where silence reads as broken.

The design spec is being brought into line with this in the companion design PR; the code change
stands as written.

## Compatibility

Records already written keep the classification they were written with — nothing rewrites history,
which is why the changelog carries a `## Compatibility` section rather than a migration. An existing
"unexpected errors" count therefore does not drop on upgrade; it ages out.

## Changelog

Two entry pairs, one per change: `changelog/unreleased/2026-08-27-messaging-error-kinds.{md,zh.md}`
and `changelog/unreleased/2026-08-27-messaging-mention-no-file.{md,zh.md}`.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm test` and `pnpm format` + `pnpm format:check` all pass on
Node 24 (server suite: 98 files, 1289 tests).

The classification is pinned in two places and no more. `packages/server/test/messaging-error-kind.test.ts`
keeps two cases for the two halves of the rule: a typed refusal is `expected` only at the capture
points that answer into the chat (and the same denial at `messaging_send_failed` /
`messaging_inbound_failed` is not), and anything unclassified stays `unexpected` — including a plain
`Error` repeating a permission error's wording, the case that keeps the rule honest.

`messaging.test.ts` carries the two end-to-end halves through `error_records`: the Feishu scope denial
on the **image download** records as `expected`, and the same typed denial on the **send** (a new
`failSendWith` on the fake, mirroring `failImageFetchWith`) records as `unexpected`. Removing the code
gate from the classifier turns that second assertion red, which is the check that matters here.
`messaging-qq.test.ts` asserts the media refusal is typed rather than bare.


hiyouga:
## Review

Reviewed in a dedicated worktree, with the failure chain proved executably against the real Lark SDK rather than reasoned about.

**The idea is right and the discipline is correct.** `messagingErrorKind` matches on `instanceof`, nothing on wording — which matters, because provider messages are localized and change — and anything unclassified keeps `unexpected`. Keep all of that.

The problem was where the rule was applied.

### 1. A scope denial on the *send* path was filed as `expected`, and the chat is never told

`MessagingPermissionError` is thrown by `larkFailure` for **any** Feishu call refused with 99991672 — including `sendContent` (backing `sendText`) and `replyText`. But `recordError` classified by type alone, with no knowledge of whether a notice reached the chat.

The PR's justification — "the chat has already handed the only person who can act the exact fix" — holds for exactly two paths: the inbound image download, which then sends `messagingImagePermissionNotice`, and the outbound file upload, followed by `noteFileFailure`. It is false on a **text send**, where the channel refused the message itself and there is no notice to carry the fix.

Verified through the existing `larkHttp` wire harness: `sendText` refused with `{code: 99991672, msg: "…[im:message:send_as_bot]…"}` rejects with `MessagingPermissionError`, and `messagingErrorKind(err)` returned `"expected"`.

So a Feishu app granted 接收消息 but not `im:message:send_as_bot` — the ordinary half-configured first run, and equally a revoked scope on a working bot — receives everything, replies to nothing, and the cost centre's unexpected count **stops lighting up**. That is the worst messaging failure mode there is, and the PR muted its alert. The same applied at `messaging_inbound_failed`, the `deliverFiles` outer catch, and two more sites.

The PR's own changelog stated the rule it violated: *"A send the channel refused … stays `unexpected`"*.

**Fixed** in a7f81c54 — the classifier now takes the capture point's `code`, gated on a named `CODES_EXPLAINED_IN_CHAT` set of exactly the two paths that emit a notice; everything else returns `unexpected` before the `instanceof` chain is reached. All three typed errors are gated rather than just `MessagingPermissionError`: it costs nothing today, and it means a typed refusal thrown from a send path later cannot silently mute the send.

Mutation-verified: deleting the one gate line turns the new assertion red with `- "unexpected" / + "expected"`.

### 2. QQ's passive-window refusal — decided, and the decision is now in the code

`qq-connector.ts` threw `MessagingUnsupportedError` for the outbound media refusal and a bare `Error` two hundred lines away for the passive-reply-window refusal, which appeared to meet every criterion the class documents. Since the design spec says the window case is the *common* one — 网页端发起的轮次通常落在窗口之外 — a QQ binding driven from the Web App files one `unexpected` per turn, so the PR's headline complaint survived where it is loudest.

**Resolved by keeping the bare `Error` and writing down why**, which I think is the right call: it is structurally the same failure as finding 1. The reply text is dropped and the chat cannot be told, *because the message that would carry the reason is the message being refused*. Classifying it `expected` would reopen the hole this PR closes, on a different channel. And `MessagingUnsupportedError`'s own contract says its `message` must say what could not be carried and why "because that text reaches the chat" — which a send outside the window cannot keep. The media refusal is the opposite: the reply's text still arrives carrying the notice that names the file. That contrast now leads the doc comment, so the split reads as a decision rather than an oversight.

### 3. An undescribed second commit contradicted the spec

Commit `d26b005a` turned `messagingFilesMissingNotice` into `messagingFilesMissingLog` and stopped sending it — while `05-ARCHITECTURE.md` says, in a sentence merged today, that 「Workspace 内没有这个文件」 is one of the failures explained in the chat. The PR body did not mention the change at all, so neither the spec nor the description told the truth.

The code change stands; I am landing the design correction separately. The PR body now carries a section describing that commit.

### 4. The same three-member allowlist was pinned at two levels

A 61-line unit file testing a six-line pure function, alongside three end-to-end `messagingErrorKinds()` assertions that already proved both the rule and the wiring through `error_records` — with two of the five unit cases restating the same `instanceof` chain from different angles.

Now that the classifier takes a `code`, the unit file is down to two cases and two e2e assertions were dropped. **Net: 1293 → 1289 tests.** The two that remain are the ones that carry the contract — image scope denial → `expected`, send scope denial → `unexpected`.

### 5. Three smaller ones

The changelog described the current codebase in the present tense ("`error-kind.ts` **holds** the rule", "Three failures **are** expected"), which `changelog/README.md` bans by name — rewritten past tense in both languages.

The entry covered two independent changes under one slug and H1, in a repo that deliberately has no index and expects `ls` plus `grep` to answer "why did X change" — the mention-notice half is now its own `2026-08-27-messaging-mention-no-file.{md,zh.md}` pair.

And a permanent comment asserted that "the cost center … raises a to-do badge on their count" — #510 is still open and not on `main`. The highlight half is real; the badge half does not exist. The clause is gone from the module, the test header and the PR body.

### Checks

- Server suite **1289/1289**, typecheck clean, prettier clean on all nine changed files.
- `lastDeliveryError` is untouched throughout, so #505's binding panel is not undone — worth stating explicitly, since that was the risk in a PR about suppressing error reports.

PR #512, opened 2026-08-27
Relay only a run's final reply, per messaging binding

hiyouga:
Every completed assistant message of a run is mirrored to the chat the moment it completes, so a
run that writes working notes between tool calls sends each note as its own message. This adds
`finalReplyOnly`, a per-binding delivery preference that holds those back and delivers the run's
**last** completed assistant text alone, when the run ends.

Off by default, and off reproduces the existing behaviour byte for byte. It mirrors
`linePerMessage` through every layer that field already touches — a column beside `enabled`
rather than a key inside the channel's `config` document, an ordinary saved field the PUT owns,
an omitted value keeping the stored one — so this is a second field along a proven path rather
than a new one.

## What it does

- **Holding.** With the option on, `relay` sends nothing: each completed assistant text replaces
  the one the run was already holding, and the survivor goes out on the run's `task_state` →
  `idle` edge. The preference is read per message off the stored row, like every other delivery
  preference — and no run ever sees it change, since a save on an enabled binding reconnects and
  builds a fresh entry (which also discards whatever was held at that moment).
- **One delivery path.** The held reply is delivered through the same `deliverReply` as any
  relayed message, so chunking, the `linePerMessage` split, the pacing, the group's single
  reply-to and the send chain's ordering are the ones already there. The two options compose:
  `finalReplyOnly` picks which text is sent, `linePerMessage` decides how that text is split, so
  a binding with both on sends the final reply one message per line.
- **Addressed where the run was asked.** This is the one delivery that leaves at the far end of a
  run, so reading the chat at delivery time is not the same thing as reading it at the run's
  start: a second person writing in another Telegram forum topic meanwhile moves both
  `last_chat_id` and the reply anchor, and the *whole* answer to the first question would follow
  them there, quoted onto their unrelated message. The chat ref and the reply anchor are captured
  on the idle → running edge beside `runStartedAt` and passed to the held delivery, and to the
  files that ride behind it so the words and what they name stay together. A message relayed as
  it completes still goes to the chat the row names then, which is where the conversation is.
- **Files follow the words that actually reached the chat.** `deliverFiles` scanned the whole
  run's concatenated reply text for mentions. With the option on it scans the final message
  alone — a file named only in a held-back working note would otherwise land in a chat with
  nothing there naming it, and the mention is the whole reason the feature knows which output
  was the point. Stated in a comment at the `onTaskState` call site and in `deliverFiles`' own
  doc.
- **Untouched.** The `approval_request` notice is not an assistant message and is never held — a
  run blocked on approval is exactly when the chat has to hear something. The fixed notices and
  the test message are unaffected, compaction output is still skipped before any of this, and a
  connection joining mid-run still relays nothing from that run.
- **Not delivered.** A run whose idle edge the connection never sees (the process stops, or a
  state toggle / credential save mid-run rebuilds the entry) delivers nothing of what it held.
  The every-message path loses only the remainder of that run in the same situation; the comment
  on the entry field says so.

## Where it landed

`messaging_bindings.final_reply_only INTEGER NOT NULL DEFAULT 0` (schema + the `ensureColumn`
upgrade list) → the repo row, `mapRow` and `upsert` → the three `*BindingInfo` types and the
three PUT request types → the three PUT handlers and the three info projections → the bridge →
the web form, the binding editor and both string dictionaries → `server-api.en.md` / `.zh.md`.

In the editor the two options now close the form as a pair of `Switch` rows, extracted into one
`DeliveryOptionRow` and fed from the selected channel's sub-state, ordered as they take effect:
which messages are sent, then how each one is split.

On QQ the option is not a shade of the same trade but a different outcome: every send there is a
passive reply anchored to an inbound message, that anchor expires after about five minutes, and
holding the reply until the run ends means a run that outlives the window delivers nothing at
all — where the other two channels only pay in silence. So it is said twice on that channel:
appended to the option's explanation for the reader deciding, and stood under the row as an
attention strip (`toneStrip.attention`) while the switch is actually on, which is the only state
it bites in. QQ's standing replies-only line under its credential fields is untouched by this PR.

### Backward compatibility

Purely additive and the same standing convention as `line_per_message`: the column is ALTERed
into an existing `web.db` on open with a default that reproduces the delivery every existing
binding already had, so no binding changes behaviour, nothing is rewritten and the user does
nothing. A downgrade is invisible in both directions — the effect is the bridge's, so an older
server ignores a column it has never heard of and relays every message again, and the stored
setting is honoured as-is on an upgrade back. This is recorded as its own section in
`changelog/unreleased/2026-08-27-backward-compatibility.md`, this batch's record, which the
feature entry links; the `ensureColumn` line carries the same lifetime as every other entry in
that list — removable only in a release allowed to break existing `web.db` files.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm test` (8 packages, 0 failures) and `pnpm format` /
`pnpm format:check` all passed on the branch as first opened. Every revision since has been
verified per package: `npx prettier --check` on each changed file, `npx tsc --noEmit -p
tsconfig.json` in `packages/server` and `packages/web`, and the full `npx vitest run` in both —
99 files / 1322 tests and 107 files / 1477 tests respectively, all green.

Tests extend the existing suites rather than adding files:

- `packages/server/test/messaging.test.ts` — the default off still mirroring every completed
  message of a run; on, a gated three-message run delivering only its last text, and delivering
  nothing while the run is still running (asserted against the channel's own message count, so
  "nothing was sent" is not a race); on + `linePerMessage` composing; a run whose only message is
  its last one delivering the identical array under both settings; the approval notice arriving
  while the run is held and the held reply landing behind it; the PUT → row → info round trip,
  including that the two preferences are independent fields; and, in the outbound-files section,
  a run whose held-back note names `draft.md` and whose final message names `final.md` sending
  only `final.md`.
- `packages/server/test/messaging-telegram.test.ts` — the flag rides that channel's PUT, defaults
  off, and does not carry the other one along; and, in the forum-topics section, a held reply
  answering into the topic its run was asked in after a second user moved the chat to another
  topic mid-run (that user's own notice still goes where they wrote it).
- `packages/server/test/db-upgrade.test.ts` — a `messaging_bindings` formed before the column
  reads `false`, keeps its other fields (`line_per_message` included), and is writable at once.
  The pre-column seeding helper is now parameterized by column name rather than copied.
- `packages/web/test/messaging-binding-form.test.ts` — per-channel load (each flag set on a
  different fixture, so neither can pass by riding the other), both flags always sent on every
  channel in all four combinations, and either one counting as a dirty edit.
- `packages/web/test/messaging-binding-editor.test.ts` — both rows render on all three channels
  with their help behind the label's "?", in that order, and each switch renders from its own
  form field.


hiyouga:
Follow-up commit `cb8acf2c` adds one thing the implementation surfaced: on QQ the option cuts both ways. It spends the least of that platform’s passive-reply budget a run can spend — one message — but a passive reply is accepted only for a few minutes after the inbound message that funds it (`QQ_PASSIVE_WINDOW_MS`, 5 minutes), and holding the reply until the run ends spends that window on the run. A run that outlives it delivers nothing, where the every-message relay would have sent whatever completed inside the window. Recorded on `QQBindingInfo.finalReplyOnly`, in the reply-only paragraph of `server-api.{en,zh}.md`, and in the changelog entry. No behaviour change and no connector touched — the note is the deliverable.

hiyouga:
## Review

Reviewed in a dedicated worktree, with the run lifecycle traced end to end.

**The bridge logic is sound and needed no rework.** The feature follows `linePerMessage`'s proven path: the column is additive with a default that reproduces today's delivery, the three PUT routes and three info projections are wired symmetrically, and the held reply goes out through the same `deliverReply`, so chunking, the per-line split, pacing and the send chain are unchanged. I traced the cases where "final" is ambiguous and they are all correct — `session-manager.ts` publishes `task_state: idle` from `drive`'s `finally`, so a held reply is never stranded by a failed run; each queued follow-up gets its own idle edge; the approval notice is not a `model_msg` and is never held; and a mid-run join is guarded by `armed: runState === "idle"`.

Two things blocked, and one of them is worth reading even if you skip the rest.

### 1. The PR moved a standing notice *into* a collapsed fold, against an explicit design rule

`design/specs/06-PROTOTYPE.md:71` requires 「QQ 的字段下另有一条**留在屏幕上**、不收进折叠的说明」, keeping only the reply budget in the fold. `08-CHANGELOG.md:15` states the reasoning in bold: 把「只回复你刚发出的消息」**直接写在屏幕上**而不是收进折叠——不知道这条规则的人，会把这个渠道读成坏了.

The PR deleted the standing `<p>{S.qq.repliesOnly}</p>` under the QQ fields, moved the string into `HelpFold` (which opens collapsed), and **deleted the test that pinned it** — a test literally named *"states QQ's replies-only rule on screen, not in a collapsed fold"*, whose comment quoted that rationale.

The failure is the one the design bullet exists to prevent: a user binds QQ, sees a form with no standing caveat, types into the web composer, and nothing arrives — with the one sentence that explains why now two collapsed disclosures away. It was also scope creep, with nothing to do with `finalReplyOnly`.

**Fixed** in da31d906. This finding and the duplicated-assertion one turned out to be the same stray commit (`5c660d43`, "QQ's delivery rules move into the help folds"), so it was reverted wholesale rather than hand-patched — the standing paragraph is back with its original comment, only `replyBudget` remains in the fold, both dictionaries' doc comments are restored, and the test is back with its three assertions. `MessagingBindingBody`'s module doc already described this placement and is true again.

### 2. A held reply was addressed at run *end*, so it followed a chat that moved mid-run

`deliverReply` reads the target when the send begins. With `finalReplyOnly`, that is the run's end rather than its first assistant message — and `onInbound` rewrites `last_chat_id` and `lastInboundMessageId` for every message that arrives meanwhile, including the follow-up a busy Session queues.

Telegram forum group: A asks Q1 in topic A; the run is long. B sends Q2 in topic B while it runs. Run 1's idle edge then delivers the **entire** answer to Q1 into topic B, quoted onto B's unrelated message. With the option off, only messages completing after B's arrival could land there — the option widens the window from "the tail" to "the whole reply".

**Fixed rather than documented**, which was the right call: the failure is silent and total, the option is what widens it, and the capture point already existed — `runStartedAt` is taken on the same edge, so this is three lines of state rather than a new mechanism. A `ReplyTarget` is captured on the idle→running edge; `deliverReply` takes an optional `at` that wins over the live target; `onTaskState`'s idle branch reads it synchronously with `held`/`since`, before the next run's `running` edge can replace it. The `row` is still read fresh, so **where** a reply goes is the run's and **how** it is split is the binding's.

The fix was extended to `deliverFiles` for the held path only — with the text addressed to the run's chat, leaving the files on the live chat would split the bundle, which is the exact defect `deliverFiles` exists to prevent. The every-message path keeps the live chat, correctly: its texts left as they completed and no single target fits all of them.

Verified to fail without the fix (`replyTo: 202, threadId: 77` instead of `201`/`91`).

### 3. On QQ, the warning that prevents total silence was disclosure-only

`qq-connector.ts` throws when the passive window has closed, and the throw is *before* `withhold`, so an undeliverable final reply is not even buffered for the next inbound message's budget. The window is 5 minutes.

So: QQ binding, switch on, user asks at T=0, run finishes at T=6min → one `messaging_send_failed` row and the chat receives nothing, ever, for a run whose answer sits complete in the Trace. A user who flips the switch without opening the "?" never sees the warning that says so. Given the design insists QQ's replies-only rule cannot wait for a disclosure, the same argument applies here with more force.

**Fixed** — the warning now also renders as a `toneStrip.attention` strip under the row, gated on `channel === "qq" && delivery.finalReplyOnly`, so it appears only in the state where it bites. The popover paragraph stays; both dictionaries record that it is said twice and why.

### 4. Two smaller ones

`relay`'s doc claimed "what is already held is delivered by whichever rule the run ends under" — but each PUT handler calls `sync` on an enabled binding, `sync` reconnects, and `connect` opens with a `disconnect` and a fresh entry, so a mid-run save **discards** what was held. That is what `heldReply`'s own comment says 580 lines earlier. Now consistent.

The new column was missing from this batch's `2026-08-27-backward-compatibility` record — which states its own purpose as "so a reader looking for 'does my install need anything?' finds every answer in one place" — while the feature entry linked the **08-26** record, whose added-column section is about `line_per_message`. Precedent was explicit: 08-26 carries a `## The added line_per_message column` section for exactly this. Now added, with `#512` in the `PR:` list and the feature entry repointed.

### Still open before merge

**Design.** `05-ARCHITECTURE.md:1280` enumerates the `messaging_bindings` columns exhaustively and has no `final_reply_only`; `:1286` carries the whole delivery contract and has no `finalReplyOnly`; `06-PROTOTYPE.md:71` names only the 「每行一条消息」 switch. One complication for whoever writes it: **`:1286` still describes the pre-per-message model** — 累计…已完成的 assistant 文本…翻转 idle 时把拼接文本发到最近来信会话 — which has not matched the code since the bridge shipped. So the bullet cannot simply be appended to; the default has to be corrected before the option can be described against it. I am handling this.

### On tests

**No change requested, and none needed.** The additions are proportionate — 7 server and 5 web cases, all extending existing suites, none in a new file. The PUT round-trip being pinned on both Feishu and Telegram is not duplication: those are independently wired route handlers, and it matches how `linePerMessage` was already covered. Two tests were added in the fix: the one pinning finding 2, and the restored assertion from finding 1.

### Checks

- Server **1322/1322**, web **1477/1477**, both typechecks clean, prettier clean on all ten changed files.

PR #513, opened 2026-08-27
Inbound files from Feishu and Telegram, as composer attachments

hiyouga:
A document sent to the bot on Feishu or Telegram reached the "not supported" branch: the connector seam had no notion of a non-image file at all. It now arrives as the web composer's own file attachment.

## What the Agent receives

An inbound file goes down the **composer's attachment path**, not the conversation: `attachFilesToInput` writes it into `<agent>/scratchpad/<sessionId>/` under a filesystem-safe name, and the message text gains one `[attached file: <absolute path>]` line per file, which the model opens with its ordinary file tools. That is the same service, the same placement rule and the same caps an authenticated composer upload answers to — the alternative, a second attachment path with its own numbers, is exactly what `services/attachment-limits.ts` exists to prevent. Nothing downstream scales with the file's size, which is what makes the far larger ceiling (100MB per file by default, against 20MB for an inline image) reasonable at all.

The seam mirrors the image one: `MessagingInboundMessage.files` is a list of **handles, not bytes**, so a redelivered message is dropped before anything is transferred, and the cap rides into `fetch(maxBytes)` so an oversized attachment is refused at the byte that crosses it and is never resident in the process. It is optional, so a channel that delivers no files never populates it.

A caption is that message's text, as it already is for a photo, so text, images and files compose one input: text part, then the image parts, with the file lines folded onto the text by core's shared placement rule.

## Caps, and one number that had to travel

Two byte budgets apply, both the admin-settable attachment limits read fresh per message: the per-file cap rides into each fetch, and the per-message total is spent as the batch is collected, each later file's cap shrinking to what is left. The per-message file *count* is deliberately not enforced — `attachment-limits.ts` calls it a composer-usability bound rather than a resource one, and a chat message carries one file today on both channels.

There is no rolling per-binding budget like the image one, because the cost that budget exists to prevent is not paid here: an inline image is written into the Trace and re-read whole on every history page and every resume, while a file is bytes on disk that nothing re-reads and that are deleted with the Session.

`MessagingMediaTooLargeError` now carries its `maxBytes`. Without it the notice would name the ceiling the *bridge* asked for, while the ceiling that actually refused the transfer can be tighter — Telegram serves a bot no file over 20MB whatever the server would accept — and the sender would be sent to shrink a file against a limit that was never in the way.

## Refusals stay distinct

Four bilingual notices, each naming the file: over the per-file cap, a batch over the per-message total, a permission the bot lacks, and a transfer that failed. They are separate for the reason `media.ts` already records — a message that names two causes names neither, and a permission refusal once reached us reported as "the image is too large or the download failed". The per-file and per-message size notices are also separate: no single file is at fault in a batch overrun, so naming one would send the sender to shrink a file that was never too big. None of the four starts a Task, and none leaves a half-written batch in the scratchpad — the whole batch is downloaded and refused before anything is written. A size refusal is nobody's fault and stays out of the error log; a failure lands under `messaging_file_fetch_failed` with the channel's own reason.

A write that fails after the downloads succeed is deliberately *not* answered in the chat: that is a fault of this server (a full disk, an unwritable scratchpad), the same class as a Session that cannot load, and it reaches the user through `recordDeliveryFailure` and the binding's status line, like a `startTask` that throws already does.

## Per channel: what is supported, and what honestly is not

**Feishu** — the `file` message type, downloaded through `im.v1.messageResource.get` with `type: "file"`. Same resource endpoint as the images, same scope: an app refused an image is refused a file by the same `99991672` denial, so `MessagingPermissionError` and the notice naming the scopes and the console link are reused unchanged. The event names the file, so the sender's own name reaches the model; an event carrying none gets the placeholder `file` rather than an invented extension.

**Telegram** — the `document` field. A picture dragged in uncompressed is a `document` too and now arrives as an attachment, which is what the sender asked for by sending the original bytes.

**Deliberately not supported, on both:** Feishu's `audio` (a voice recording) and `media` (a video), and Telegram's `audio` / `video` / `voice` / `video_note`. They carry a downloadable key like a file does, so this is a choice rather than a limitation, and the reasoning is in a comment at each connector: those are streams the platform re-encodes for playback, named by nothing the sender chose (a `voice` is a nameless opus blob), and nothing downstream decodes or transcribes them — handing one over as an `[attached file: …]` path would offer a capability that is not there. What a sender genuinely wants the Agent to have arrives the moment they attach it as a file, which is one menu item away, and until then the not-supported notice at least says so.

**Telegram's own ceiling** is capped in the connector at 20MB (`TELEGRAM_MAX_DOWNLOAD_BYTES`) rather than left to the Bot API to refuse. Past it `getFile` answers a 400 whose only signal is prose Telegram is free to reword — and branching on a `description` is the one thing `telegram-api.ts` forbids — so a size the sender can act on would otherwise have reached them as an unexplained download failure.

The not-supported notice now names what is supported; its constant is `MESSAGING_UNSUPPORTED_NOTICE`.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm format` + `pnpm format:check` — clean.
- `pnpm test` — 9 packages, 4495 tests, all passing (server: 99 files / 1332 tests).
- New coverage, wire-level wherever a wire can produce the case:
  - `messaging.test.ts` (Feishu, through the fake SDK behind the real connector): a file-only message landing in the scratchpad with the right bytes and the right path line, under the per-file attachment cap; a size refusal naming the file and staying out of the error log; a permission refusal naming the scopes and the grant link; a transfer failure carrying the channel's own reason plus an error record; a redelivery downloading nothing and writing no second copy; a nameless file taking the placeholder; a sticker still getting the not-supported notice and asking for no download.
  - `messaging-telegram.test.ts`: a document landing in the scratchpad under Telegram's 20MB cap; its caption as the message text, `@mention` stripped; a document past the ceiling refused as a size problem naming 20MB; a failed download degrading to a notice with the caption not running alone; a nameless document; a redelivery; and a captioned media message with no `document` still getting the not-supported notice.
  - `messaging-wire.test.ts` (the real Lark SDK over a stubbed axios): the `type: "file"` query parameter on the resource endpoint — the whole of the wire difference from an image, and the layer a seam fake stubs out; a stream-borne scope denial arriving as `MessagingPermissionError` with scopes and grant link; a size refusal carrying its `maxBytes`.
  - One bridge-seam suite, `a message carrying several files`, for what no wire here can produce: neither channel delivers two files in one message, while the seam carries a list. It drives the bridge's loop — ordering, composing with a caption and an image, and the per-message total refusing a batch and leaving nothing on disk.


hiyouga:
## Review — inbound files from Feishu and Telegram

Reviewed against `origin/main`, on a worktree of `feat/messaging-inbound-files`. The design of the seam is right: a handle rather than bytes, the cap riding into `fetch`, all-or-nothing across both attachment kinds, the four notices split by cause, and `MessagingMediaTooLargeError.maxBytes` so a channel that narrows the transfer reports its own number. What follows is one blocker (fixed), one latent wrong notice (fixed), a design-doc gap (not fixed — reporting only), and a suite trim.

Fixes pushed as `fix(messaging): keep a queued inbound file recallable, and trim the added suite`.

---

## Blocker — a queued inbound file is unrecallable, and leaks

`startAttachedTask` calls `startTask(..., { queueIfBusy: true })` with no `recall`. On a busy Session the manager then builds one itself:

```ts
// session-manager.ts
recall: opts.recall ?? recallStoreOf(input),
```

and `recallStoreOf` was written under an assumption this PR invalidates — its own comment said so:

> `files` is empty because **only the HTTP route writes file attachments to the Session scratchpad**, and only it knows the paths they landed on.

The bridge now writes them too. Worse, the `input` it derives from is the **post-fold** one, so `text` already contains the `[attached file: …]` line.

Reproduced end to end (busy Session, one inbound file, then `DELETE /api/sessions/:id/follow-ups/:id`):

```
PENDING:  [{"id":"…","text":"look at this\n\n[attached file: /tmp/…/scratchpad/session-…/notes.txt]",
            "images":0,"files":0}]
A files count:      0          ← queued line says the message carries no files
B marker in text:   true       ← composer gets a raw scratchpad path back
C files returned:   0          ← the attachment is silently lost
D dir after recall: ['notes.txt']  ← the copy stays on disk forever
```

All four are one cause. `recalledResponse` deletes exactly `recall.files`, so an empty store means nothing is cleaned up; and the text that comes back is what the user would resend, producing a second marker line pointing at a file the recall was supposed to take back.

**Fixed** by having the inbound path supply its own store, built on the *pre-fold* input plus the written paths:

```ts
function inboundRecallStore(input, attachments, written): RecallStore {
  return {
    ...recallStoreOf(input),
    files: written.map((filePath, i) => ({
      fileName: attachments[i]!.fileName, path: filePath, mime: attachments[i]!.mime,
    })),
  };
}
```

`MessagingTaskRunner.startTask` gains an optional `recall`; `recallStoreOf` is exported and its contract restated. After the fix the same probe yields `{"text":"look at this","images":0,"files":1}` and `{"files":[{"fileName":"notes.txt","dataUrl":"data:application/octet-stream;base64,aGk="}]}`, with the directory empty. Mutation-tested: disable the `recall:` spread and the new test goes red, restore it and it passes.

One test added for this (`hands a queued message's files back when the follow-up is withdrawn`) — it is the only defect here that needed one.

---

## Wrong notice — the size discriminator reads the cap it offered, not the one that bit

```ts
notice: cap < limits.maxBytes
  ? messagingInboundFilesTooLargeNotice(limits.totalBytes)   // "these files add up to too much"
  : messagingInboundFileTooLargeNotice(file.fileName, err.maxBytes)
```

`cap` is what the bridge *offered*. A connector may narrow it further — which is the whole point of `downloadCap()` and of the new `err.maxBytes`. So on any file after the first, whose `cap` the message's remaining budget has already shrunk below `maxBytes`, a refusal at Telegram's 20MB ceiling is reported as the *message total* being blown.

With the shipped defaults (100MB / 120MB) and a message carrying two files, first 30MB, second a 25MB document: `remaining = 90MB`, `cap = 90MB < 100MB`, Telegram refuses at 20MB → the chat is told "the files in that message add up to more than the 120MB one message may carry". They add up to 55MB. Reproduced against the scripted seam (40/60-byte limits, a connector narrowing to 10):

```
- Expected  "big.bin" is larger than the 0MB limit, so it was not sent to the Agent. …
+ Received  The files in that message add up to more than the 0MB one message may carry, …
```

**Fixed** — the test is now the ceiling the transfer *reports* refusing at:

```ts
const spentTheMessage = err.maxBytes === remaining && remaining < limits.maxBytes;
```

Both existing arms still hold: the batch-overrun case (`err.maxBytes === remaining === 25 < 40`) still gets the batch notice, and Telegram's 20MB refusal (`20MB !== 120MB`) still names the file and the 20MB. No test added — the case cannot occur on any shipped channel today (Feishu's `file` carries one `file_key`, a Telegram media group is one update per document), and the two arms are already pinned.

---

## Blocker (process) — the design docs contradict this PR

`design/specs/05-ARCHITECTURE.md` line 1285, the 入站（渠道 → Session）bullet, documents the behaviour this PR reverses, on three counts:

> **说明文字只在真的带图时才算作该消息的文本**——Telegram 的文件、视频、音频同样带说明文字，而它们的字节并不送达（把图片按原图拖进去就是 `document`）…

> 其余非文本消息（表情、语音、视频、**文件**、飞书富文本）…回一条双语「暂不支持」提示

> 连接器接缝上带的是**延迟拉取的句柄**而非字节 … 上限有两道：单张沿用 `INLINE_IMAGE_MAX_BYTES` … 以及每份绑定一个**滚动窗口的字节预算**

Per `CLAUDE.md` the specs are ground truth and must land in sync. I did not touch them (out of scope for this review); what needs writing, all inside that one bullet:

- the caption sentence — a `document` / `file` caption **is** the message's text now;
- the not-supported list — drop 文件;
- a new passage for the file seam: scratchpad attachment rather than conversation bytes, the admin-settable per-file / per-message budgets (and why there is deliberately **no** rolling window here — nothing re-reads a file), the four notices, Telegram's 20MB `getFile` ceiling applied by us rather than left to the Bot API, and the media kinds deliberately excluded (`audio`/`media` on Feishu, `audio`/`video`/`voice`/`video_note` on Telegram);
- the handle sentence should mention `files` alongside the image handles.

Plus the usual dated line in `specs/08-CHANGELOG.md`.

---

## Verified, no action

- **Filename sanitisation / path traversal.** Probed `scratchpadName` directly with chat-shaped names — nothing escapes the session directory: `../../../etc/passwd → ..-..-..-etc-passwd`, `..\..\windows\system32\x.dll → ..-..-windows-system32-x.dll`, `a/b/c.txt → a-b-c.txt`, `\0evil.txt → -evil.txt`, `.. → file`, `con.txt → _con.txt`, `‮gnp.exe → -gnp.exe` (the RTL override is stripped as category C), `报告 2026.pdf → 报告-2026.pdf`. `path.join(dir, out)` stays inside `dir` in every case. The `"wx"` create also refuses to follow a planted symlink. Not passing a chat-supplied name through `parseAttachmentPart` is fine — that validator rejects rather than sanitizes, which is right for an API caller and wrong for a chat.
- **No new multipart path**, so the undici `FormData`/`fetch` pairing in `telegram-api.ts` is untouched and still imports both from `undici`.
- **Marker parsing stays render-layer.** Nothing in the file path reads message text to decide anything; `startAttachedTask` composes and core's `appendAttachmentLines` places.
- **`sendChain` untouched.** Inbound work never enqueues onto it; notices go through `replyInbound` as before, so a slow download delays only that binding's own inbound handling.
- **Cleanup on a failed start** is right: `removeAttachments(written)` on a `startTask` throw, matching the route's half of `attachFilesToInput`'s contract. A write failure (full disk, unwritable scratchpad) correctly does **not** answer in the chat — it is not the sender's to fix — and reaches `recordDeliveryFailure`.
- `pnpm typecheck` — **exit 0, 0 errors** (an earlier run reported 2 in `packages/cli` for `@prismshadow/penguin-server/{secret-file,auth-token}`; that is a stale `dist/` from before #443 landed, clean after a rebuild).
- `packages/server` vitest, solo: **103 files, 1374 tests, all pass.**

---

## Nits, not fixed

- **`scratchpadName` does not cap the extension.** The stem is truncated to 80 UTF-8 bytes but `ext = sanitizeSegment(path.extname(fileName))` is not, so `a.` + 400 chars yields a 402-char name and `ENAMETOOLONG`. Pre-existing in `task-attachments.ts`, but this PR makes it reachable by anyone who can message the bot. It degrades safely (the throw lands in `recordDeliveryFailure`, nothing is written), so it is a nit — but a one-line `truncateBytes` on `ext` would close it.
- **The image budget is spent even when the message is later refused for a file reason.** `startAttachedTask` downloads images first and `budget.spend()` has already run when `inboundFileAttachments` returns a notice. The user pays window budget for a message that never ran. Small, and the alternative (deferring the spend) is worse.
- **`maxCount` is deliberately unenforced inbound.** Documented, and the byte budgets bound memory and disk — but they do not bound *count*: 1000 one-byte files would be 1000 writes and 1000 marker lines. No channel produces that today; worth a sentence if a batching channel ever arrives.
- **`messagingInboundFilesTooLargeNotice` floors to `0MB`** for sub-MB limits. Only visible in tests.

---

## Test restraint — 677 lines of test for ~433 of source

Six of the added cases pin ground already pinned. Cut:

| Cut | Why |
|---|---|
| wire: `carries a file download's own refusal reason and grant link` | the refactor put both downloads through one `fetchResource`; the existing image test `reports the API's own refusal reason, which arrives as a STREAM` covers that function |
| wire: `refuses a file download past the cap as a size error carrying the ceiling that bit` | `collectUnderCap` past-cap is covered by the image cap test, and `err.maxBytes` end-to-end by Telegram's 20MB notice test |
| telegram: `strips this bot's own @mention off a document's caption` | literally the same `stripBotMention(msg.caption, msg.caption_entities, me)` call the photo-caption test already drives — the ternary's new arm is covered by the caption test next to it |
| telegram: `a document that cannot be downloaded degrades to a notice, caption and all` | channel-agnostic bridge branch, covered by Feishu's `a file the channel will not hand over reports the channel's OWN reason` |
| telegram: `a redelivered document is dropped before anything is downloaded` | `isRedelivery` returns before any kind dispatch; covered for text, stickers, images, the whole `a redelivered inbound message` describe, and — with the file-specific "no second copy on disk" — by the Feishu version |
| scripted: `puts the caption first and the files' path lines behind it` | covered against a **real wire** by telegram's `a document's caption rides along as the message's text, ahead of its path line` |

Kept, because each pins something nothing else does: the two happy paths (they differ — Feishu asserts the fetch cap is the server's, Telegram that it is `TELEGRAM_MAX_DOWNLOAD_BYTES`), the two name-fallbacks (`file` vs `document`, different code), the three Feishu notice arms, Telegram's 20MB size refusal, `video, audio and voice keep the not-supported notice` (successor to the deleted captioned-document test, and the connector's `text:` ternary), the wire's `type: "file"` param, and the scripted multi-file ordering + budget-overrun pair.

Net: **−100 test lines, +1 case.** The scripted `describe` still earns its ~110-line harness on the ordering and budget cases; `composes text, an image and a file` is marginal (no wire delivers both) but pins `startAttachedTask`'s composition order, so I left it.

One more thing on the scripted block: its `LIMITS = { maxBytes: 40, totalBytes: 60 }` inverts the real relation only in magnitude, not in order, which is correct — the admin PUT enforces `attachmentTotalMb >= attachmentMaxMb`, so `cap === maxBytes` for the first file of every message is a real invariant and the tests rely on it.

PR #517, opened 2026-08-27
Qwen 3.8 Flash in the TokenDance group

hiyouga:
Adds `qwen3.8-flash` to the TokenDance group at that gateway's own rates: **CNY 0.8 input, 2.7
output, 0.1 cache hit** per million tokens, over a 1M-token context window, with image input.

## The pair, and why both rows stay

The same upstream id already sits in the Qwen pay-as-you-go group at Qwen's direct list price
(CNY 1 / 3 / 0.1). That is not a duplicate — it is one model reached two ways, priced by whoever is
selling it, and the gateway is cheaper on input and output while matching the cache-hit rate. The
test asserts that relation rather than the numbers alone, so a later edit that quietly makes the
two rows identical fails instead of passing.

## Where the facts come from

- **Prices**: the gateway's own published CNY rates, supplied by the maintainer. Stored through
  `cny()` like every other row, so switching the cost center to CNY shows exactly those figures
  back.
- **Context window and vision flag**: `GET https://tokendance.space/gateway/v1/models` — the public,
  credential-free catalog API the rest of this group is already verified against. It reports
  `context_length: 1000000` and a natively multimodal description.
- **Not discounted**: its catalog `description` carries no bracketed 限时 tag, unlike `qwen3.8-max`
  in the same group (currently 八折). So its list price and its billed rate coincide, which is what
  the group's block comment requires a row to state either way.

## Two things the entry says on purpose

- **`cache_write` carries the input price.** The whole TokenDance block runs on this: the gateway
  publishes an input price and a cache-hit price with no separate cache-write fee.
- **The `openai-chat` pin is a convention here, not a constraint.** This id advertises the widest
  protocol set in the group — `openai:chat-completions`, `openai:responses` *and*
  `anthropic:messages` — where `glm-5.3-flash` above is pinned because its list is
  `openai:chat-completions` alone. Recording the difference keeps the next reader from taking the
  pin as forced.

Row position follows the group's dictionary order by upstream id, which puts it before
`qwen3.8-max`; the file header's dated verification log gains the row's own date.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm test` and `pnpm format` + `pnpm format:check` all pass on
Node 24 — 8 packages, 0 failures (core 1021, server 1341, cli 383, web 1478, desktop 144).

`packages/core/test/model-catalog.test.ts` gained the row to the group's ordered id/context/vision
list, its exact stored USD triple (`0.014286 / 0.114286 / 0.385714`, i.e. the CNY figures at the
catalog's 7:1 convention), and the comparison against the pay-as-you-go row above. The suite's
standing checks — three-bucket pricing present and positive, positive integer context window,
`openai-chat` + the TokenDance base URL for every row in the group — cover the new entry
automatically.

### DIFF
--- a/changelog/unreleased/2026-08-27-catalog-tokendance-qwen38-flash.md
+++ b/changelog/unreleased/2026-08-27-catalog-tokendance-qwen38-flash.md
@@ -1,0 +1,28 @@
+# Qwen 3.8 Flash joins the TokenDance group
+
+- **Date:** 2026-08-27
+- **Type:** feature
+- **Scope:** `core`
+
+[中文版](2026-08-27-catalog-tokendance-qwen38-flash.zh.md)
+
+The built-in catalog now carries `qwen3.8-flash` under TokenDance, at that gateway's own rates:
+CNY 0.8 input, 2.7 output, 0.1 cache hit per million tokens, over a 1M-token context window, with
+image input.
+
+The same upstream id already sat in the Qwen pay-as-you-go group at Qwen's direct list price
+(CNY 1 / 3 / 0.1). Both rows stay: one model reached two ways, priced by whoever is selling it,
+and the gateway is the cheaper of the two on input and output.
+
+## Details
+
+- Context window and the vision flag come from TokenDance's public catalog API
+  (`GET https://tokendance.space/gateway/v1/models`, no credential required), the same source
+  the rest of the group is verified against.
+- The row is not discounted: its catalog `description` carries no bracketed 限时 tag, unlike
+  `qwen3.8-max` in the same group, so its list price and its billed rate coincide.
+- It advertises the widest protocol set in the group — `openai:chat-completions`,
+  `openai:responses` and `anthropic:messages`. The `openai-chat` pin is this group's convention
+  rather than a constraint of the id, which the entry says so nobody later reads it as forced.
+- `cache_write` carries the input price, the convention the whole TokenDance block already runs
+  on: the gateway publishes an input price and a cache-hit price with no separate cache-write fee.
--- a/changelog/unreleased/2026-08-27-catalog-tokendance-qwen38-flash.md
+++ b/changelog/unreleased/2026-08-27-catalog-tokendance-qwen38-flash.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-08-27
 - **Type:** feature
 - **Scope:** `core`
+- **PR:** [#517](https://github.com/Prism-Shadow/penguin-harness/pull/517)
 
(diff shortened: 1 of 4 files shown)


hiyouga:
## Review — approved, one nit fixed on the branch

Verified the row against its sibling rather than against the description.

### Pricing is right, and the claim it makes is right

`cny(cacheRead, cacheWrite, output)` divides by 7, so `cny(0.1, 0.8, 2.7)` lands at `0.014286 / 0.114286 / 0.385714`, which is exactly what the test asserts. Against the `qwen-pay-as-you-go` row's `cny(0.1, 1, 3)` → `0.014286 / 0.142857 / 0.428571`:

| | TokenDance | direct | |
| --- | --- | --- | --- |
| cache read | 0.014286 | 0.014286 | equal, as the comment says |
| input | 0.114286 | 0.142857 | gateway undercuts |
| output | 0.385714 | 0.428571 | gateway undercuts |

`contextWindow` (1000000), `supportsVision` (true), `clientType` (`openai-chat`) and `baseUrl` all match the sibling, so the two rows differ only in provider and price — which is the stated point of having both.

### The test is proportional

~14 lines inside the existing `describe`, no new file. It pins the three prices and then the invariant that actually matters — that the pair must not silently converge — with `toBeGreaterThan` on the two that differ and `toEqual` on the one that doesn't. That is the right shape: it survives a future price edit on either side without needing an update, and fails if someone copies one row's pricing onto the other.

The ordering assertion (`kimi-k3` → `qwen3.8-flash` → `qwen3.8-max`) also pins insertion position, which is why the new entry had to go where it did.

### Nit — fixed in 000bb3e8

The header ledger's rewrap left a stub line, and more importantly it ended up dating `qwen3.8-flash` **twice**:

```
 * ... its glm-5.3-flash row: 2026-08-26 and its
 * qwen3.8-flash row: 2026-08-27; the
 * GLM-5.3 Flash rows (direct + OpenRouter) and qwen3.8-flash: 2026-08-26 — per each
```

The tail's `qwen3.8-flash: 2026-08-26` is the **direct** row; the new clause's is TokenDance's. Only the word "its" separated them. Reflowed and named the second one:

```
 * ... its glm-5.3-flash row: 2026-08-26 and its
 * qwen3.8-flash row: 2026-08-27; the GLM-5.3 Flash rows (direct + OpenRouter) and the
 * direct qwen3.8-flash: 2026-08-26 — per each provider's docs).
```

Comment-only; prettier clean.

### Changelog

Bilingual pair present in `changelog/unreleased/`, metadata block in the required order, past tense. No index touched.

Merging once CI comes back green on the new head.

PR #528, opened 2026-08-28
A retry ladder is one line in the transcript

hiyouga:
Every retryable request end added its own line, so a request that retried four times left four of them stacked in the conversation:

```
[重试]连接超时或网络中断，已发起第 1 次重试
[重试]连接超时或网络中断，已发起第 2 次重试
[重试]网络或服务暂时不可用，已发起第 3 次重试
[重试]连接超时或网络中断，已发起第 4 次重试
```

A later attempt now replaces the one before it, so the ladder is one line counting up to `已发起第 4 次重试`.

## Scope: the transcript only

The change is in the chat page's view model (`lib/omni/stream-model.ts`). **The engine's retry behaviour is untouched, and so is the Trace** — every attempt is still recorded as its own event, and the Trace panel builds from those raw events rather than from this model (`rg -l 'omni/stream-model'` returns `features/chat/*` and nothing else), so the full ladder stays readable there.

## The part that is not obvious

"The last item is a reconnect" is necessary but **not sufficient**, and getting this wrong would have merged two unrelated failures into one line.

A ladder's items are adjacent because nothing is pushed between them — the `request_begin` that resends only marks the waiting item. But a request that **succeeds** also pushes nothing. So after `fail → retry → succeed → fail`, the new failure's item sits directly next to the previous ladder's last one while belonging to a different incident. The existing test pinned exactly that sequence.

What separates them is the attempt ordinal: a continuation climbs, a new incident restarts at 1. `continuedLadder` requires `next.attempt > last.attempt`, so the two cases stay apart.

Two more rules fall out of the same reasoning:

- **A ladder that gave up keeps its line.** It carries the final failure and its `errorMessage` — the one retry line actually worth keeping on screen — so a later incident opens its own rather than overwriting it.
- **Legacy Traces never collapse.** Written before the attempt ordinal existed, they read as attempt 1 throughout, so the ordinal never climbs and they replay exactly as they always did. That is the safe direction for a record this cannot re-derive.

The superseded item's `id` is kept through the replacement, so the rendered row updates in place instead of remounting and replaying its entry animation — which would read as a fresh failure rather than the same one progressing.

## What it costs

A superseded attempt's **cause** leaves the transcript with it. A ladder that failed for different reasons — a timeout, then a network error — now shows only the reason it is currently on. The attempt count survives in the line, and the per-attempt causes remain in the Trace. This is stated in the changelog rather than left to be discovered.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm format` + `pnpm format:check` clean; `packages/web` 111 files / 1534 tests pass.

`stream-model.test.ts` gained a case for a gave-up ladder not being overwritten, and two existing cases were rewritten rather than deleted: the ladder case now asserts one line advancing and the id being preserved, and the legacy mixed-spelling case now asserts the line carries the latest attempt *and its spelling* — the finer pre-convergence statuses still render, one at a time — with the trade written into the test's own comment.


hiyouga:
## Review — reviewed in a worktree at `6936d06c`, and reproduced in a browser

The collapse predicate is sound: I could not construct a wrong merge from any sequence the current core can emit, and `let reconnects = 0` living inside the per-turn loop (`context-engine.ts:745`) is what guarantees it — every new incident restarts at 1, so `next.attempt > last.attempt` is false against any earlier ladder.

I also ran it for real: a model pointed at a dead endpoint, one message sent, the transcript sampled once a second.

```
t=1   [Retry] … retry #1 in 2s…
t=4   [Retry] … retry #2 in 4s…
t=6   [Retry] … retry #3 in 8s…
t=15  [Retry] … retry #4 in 16s…
t=32  [Retry] … retry #5 in 30s…
```

One line throughout, counting up. The feature does what it says.

Four things to fix before it ships.

### 1. blocker — one click on "立即重试" disables the controls for the rest of the ladder

`message-item.tsx:113` — `const [acted, setActed] = useState(false)`, and `message-stream.tsx:96` keys the row on `seg.item.id`. Preserving the id is exactly what keeps the React instance alive, so `acted` survives into every later attempt of the same ladder. Before this change each attempt was a new id, the instance remounted, and `acted` reset.

Reproduced in a browser against a live server, clicking "Retry now" exactly once during attempt #2's countdown:

```
t=4   retry #2 in 4s…   button enabled   <- clicked here, once
t=5   retry #2 sent
t=6   retry #3 in 8s…   button DISABLED
t=15  retry #4 in 16s…  button DISABLED
t=32  retry #5 in 30s…  button DISABLED
```

Both inline buttons stay dead for every subsequent countdown. 放弃 has the composer's Stop as an escape; 立即重试 has none — it is only reachable from this line, and `design/specs/06-PROTOTYPE.md:82` requires it on every live waiting state.

`useEffect(() => setActed(false), [item.attempt])` in `ReconnectLine` fixes it. Keying on `${id}-${attempt}` would too, but that remounts, which is the thing this PR is avoiding.

### 2. should-fix — the same instance reuse carries a stale `now` into the first paint

`message-item.tsx:112` initialises `now` with `useState(() => Date.now())`, which runs at mount only. The ticker's effect re-arms on `[target]`, but the first commit after a replacement renders `target - now` against a `now` last written when the previous countdown stopped. A failed attempt that sat in the 120s idle timeout paints `第 2 次重试，126 秒后发起…` for one frame before the 250ms interval corrects it.

Clamping with `Math.min(target - now, item.plannedDelayMs ?? 0)` or resetting `now` in the same attempt-change effect both work.

### 3. should-fix — the stated adjacency invariant is false, and the case it misses is the common one

`stream-model.ts:1803` and the changelog both say a ladder's items are adjacent "because nothing else is pushed in between". Any attempt that emitted thinking, text, or tool-call arguments before the cut pushes an item between two rungs. Executed against the PR head:

```
request_begin → partial_text "Let me " → cut → request_end(retryable, attempt 1)
… ×3
→ ['assistant_text','reconnect#1','assistant_text','reconnect#2','assistant_text','reconnect#3']
```

Three stacked lines, i.e. the pre-PR rendering. Same with `partial_thinking`. So the collapse does not reach the `network` and `malformed` classes — the two retryable causes that by definition leave content behind, and the reason `[turn_retried]` exists at all (`05-ARCHITECTURE.md:441`).

Worth noting the sibling helper disagrees: `findLastWaitingReconnect` (`:1824`) deliberately scans back past intervening items, while `continuedLadder` (`:1818`) only looks at the tail.

The cheap fix is to correct the claim in the doc comment, the changelog and the body — "adjacent when the failed attempts produced no output". Making the collapse actually cover the mid-stream case means scanning back like the sibling and replacing at that index, which is a larger change with its own ordering questions.

### 4. should-fix — two e2e assertions become sub-frame races

`e2e/llm-errors.spec.mjs:64` and `:70` assert `已发起第 1 次重试` then `已发起第 2 次重试`. Under the old model attempt 1's line stayed on screen for the rest of the session; under the collapse it exists only between retry #1's `request_begin` and attempt 2's `request_end` — one local-mock round trip, and `use-session-stream.ts:41` throttles commits at `BUMP_MIN_INTERVAL_MS = 120`, so both events routinely land in one commit and that text is never painted.

That suite is not CI-gated (`ci.yml:118`'s `test:e2e` resolves to the core live-LLM suite; the web Playwright suite is `bash e2e/run.sh`), so nothing turns red today. This PR is still what invalidates those two lines.

### Smaller

- **Changelog has no `PR:` field.** `changelog/README.md` marks it required when one exists; 265 of 276 released English entries carry it.
- **The second `## Details` bullet is decision rationale** — "is necessary but not sufficient… which is what separates the two cases". The spec routes that to the PR description ("Reasoning does not go on disk"), and it is verbatim the code comment, so nothing is lost by dropping it.
- **The English entry's sample block is zh-only.** Quoting zh UI strings has precedent, but the precedent pairs them: `0.2.2/2026-08-11-web-ui-polish.md:13` writes 「输入 /model 切换模型 / "Type /model to switch models"」. An English reader currently gets four untranslated lines plus `已发起第 4 次重试` in the lead, where their UI says `retry #4 sent`. Same for the new code comment at `stream-model.ts:1729` — the four existing zh quotes in this package (`compaction-banner.tsx:7`, `session-order.ts:176`, `chat-page.tsx:154`, `format.ts:180`) all carry an English gloss.
- **`stream-model.test.ts:706-712` does not test what its comment claims.** `requestEnd("retryable", { attempt: 1 })` carries no `retry_in_ms`, so `:1726` sets `gaveUp = true` and the new line comes from the `last.gaveUp` branch, never the ordinal one. The genuine legacy shape — finer spellings with **no** `attempt` at all — appears nowhere; the fixture at `:785` pairs legacy spellings *with* climbing ordinals, which the core cannot produce. I ran the real legacy shape (`timeout → failed → malformed`, no `attempt`, closed by `abort`) and it is unchanged from before the PR, so the safety claim holds — it is just untested.

### Design docs

`design/specs/06-PROTOTYPE.md:82` reads 「显示重试提示行并标注第几次重试 … CLI 同样打印重试行（无倒计时）」. After this change the Web collapses a ladder to one line while `packages/cli/src/render.ts:800` still writes one per attempt, so 「同样」 now understates a real divergence, and the collapse rule plus its two exceptions (a gave-up line is never overwritten; no ordinal means no collapse) are recorded nowhere. Per `CLAUDE.md` that wants a design-repo change alongside this one — say the word and I will open it.

### Checked and clean

- No wrong merge is reachable: gave-up-then-new-at-1, new-at-1-after-a-waiting-N, gave-up-then-gave-up, and an immediate give-up after a waiting ladder each open their own line.
- Interleavings: compaction request events are gated out at `:1665` before the item is built; subagent messages route into a nested model; `abort` flips `gaveUp` before pushing its marker; an SSE gap raises `resync_required` and rebuilds from scratch rather than appending into stale state.
- Keys and animation: `MessageItems` keys singles on `seg.item.id` and the turn container on the first segment's id — both stable through the replacement, no collision, no remount. `anim-msg` is mount-only, so reuse genuinely suppresses the replay.
- Nothing downstream indexes items by position; `work-summary.ts`, `outline-model.ts`, `input-history.ts` and `agent-topology.ts` never mention `reconnect`. Ids are burned but ordering holds.
- The Trace panel is independent — `features/traces/*` imports nothing from `omni/`.
- 1494 web tests pass at the PR head; merged together with #529 and #530 on top of `origin/main`, `pnpm -r build`, `pnpm typecheck`, 1568 web tests and 1503 server tests all pass.

hiyouga:
Fixed in `8a3e252f`.

**The blocker.** `ReconnectLine` resets its per-attempt state when `item.attempt` changes, and the countdown is clamped to the wait the engine announced so the frame before that reset lands cannot paint a longer one. Re-ran the same live ladder against the built app, clicking "Retry now" once during attempt #2:

```
t=4   retry #2 in 4s…   enabled   <- clicked here
t=6   retry #3 in 8s…   enabled
t=15  retry #4 in 16s…  enabled
```

One line throughout, controls live on every countdown.

**The rest.** The adjacency claim is corrected in the doc comment and the changelog, and pinned by a test that streams text between two rungs and asserts the ladder does not collapse. `stream-model.test.ts`'s legacy case was reaching the gave-up branch rather than the ordinal one it described, and the genuine no-`attempt` shape now has its own case. `llm-errors.spec.mjs` asserts the single line advancing instead of two ordinals that are sub-frame races. The changelog gained its `PR:` field, lost the bullet that was decision rationale, and shows the English rendering.

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` clean; 111 files / 1536 web tests pass.

Left alone: the `design/specs/06-PROTOTYPE.md:82` update, which is a design-repo change — say the word and I will open it.

hiyouga:
Design counterpart opened: [penguin-harness-design#77](https://github.com/Prism-Shadow/penguin-harness-design/pull/77) — `06-PROTOTYPE.md` §「对话页」's retry-line bullet rewritten (one line per ladder, the two exceptions, the adjacency limit, the per-attempt controls, and the CLI sentence corrected), plus its `08-CHANGELOG.md` entry.

PR #529, opened 2026-08-28
One notice block on every page with updates waiting, with an update-all button

### LINKED ISSUE #527: 求【智能体】和【技能库】优化
### The problem

【智能体】和【技能库】的更新一个个点击太麻烦了~
Image
可参考QwenPaw
Image

求新增【智能体】和【技能库】一键更新

### What you do instead today

_No response_

hiyouga:
Closes [#527](https://github.com/Prism-Shadow/penguin-harness/issues/527).

Updating objects one at a time is tedious, so the page notice that says what a nav dot was
pointing at now carries a button that updates all of them — and, because a bulk overwrite needs
consent, a confirmation that states the consequence in each page's own existing words.

## What was already there, and what I did not fork

`TodoNotice` had already been extracted into `components/ui/todo-notice.tsx` as a shared,
`toneStrip.attention` (pale amber) block used on Skills, Models and the Cost Center — the brief
described it as still living inside `update-dot.tsx`, which is stale; `update-dot.tsx`'s own
header already points at the split. I grew that component rather than moving it back or building
a second one. Likewise `忽略` is the existing `dismissTodo`, the Models counts are
`catalogDelta`'s own, and the Skills bulk action is the existing per-Skill update over every
outdated Skill.

## The notice inventory, and what belongs on the shared block

Unified (the four page-level "there is an update waiting" blocks):

| Site | Before | Now |
| --- | --- | --- |
| `features/agents/agents-page.tsx` | no page notice at all | the shared block, new |
| `features/skills/skills-page.tsx` | shared block, dismiss only | shared block + bulk update |
| `features/models/models-page.tsx` | shared block, dismiss only | shared block + bulk sync |
| `features/usage/usage-page.tsx` | shared block, dismiss only | unchanged; one button |

Left alone because they report a **failure** — repainting these yellow would drop the severity
the colour is carrying:

- `features/traces/trace-panel.tsx:145` — `toneStrip.danger`, a trace that failed to load.
- `features/machines/machines-page.tsx:157` — `toneStrip.danger`, an install/load error.
- `components/ui/field.tsx:99` — `role="alert"`, a field's validation error. Red inline text,
  not a strip.

Left alone because they are **a different thing wearing similar clothes**. All five share
`toneStrip.attention` correctly — every one of them is "unfinished, waiting on the user" — but
none is an update trail: no dot, no signature, no dismissal, no count, and each is scoped to one
object rather than a page-full of them.

- `features/agents/prompt-injection-controls.tsx:199` and `features/agents/memory-tab.tsx:588` —
  these two are the closest call, because they use the shared block's exact class string. They
  are per-Agent config repairs inside a settings tab ("this prompt template has no placeholder"),
  fixed by one button that edits that one Agent. Folding them in would give them a count of one
  and a dismissal that means nothing.
- `features/machines/machines-page.tsx:170` — "this server has no installable image". A
  capability statement with no action at all; on the shared block it would grow an update button
  with nothing to update.
- `features/models/models-page.tsx:2308` — `autoRouteMiss`, inside the model add/edit **form**:
  the typed model id matches no auto-route rule, and the button edits the form. Moving it under
  the page title would separate it from the field it is about.
- `components/layout/app-layout.tsx:292` — the initial-password banner. App chrome above the
  router, spanning the shell, with its own `ui_prefs` dismissal — the precedent
  `todo-dismissals.ts` was built on. It has to survive navigation, which a page notice must not.

The three ad-hoc bordered `<p>` notices (`features/chat/{scheduled,skills,goal}-banner.tsx`) are
gray in-conversation banners in the message stream, sized `w-fit` and animated in with a message.
Not notices about pending work.

## Judgement calls

**The three counts do not mean the same thing, so the pages do not claim they do.**
`检测到变更：x 个新增，x 个可升级` is honest only on Models, where the catalog is a list of entries
and one the table lacks is genuinely new. An Agent's kernel is never new — the Agent exists and
its config is behind a defaults generation — and a Skill nobody has installed is not waiting for
anyone. Those two pages state the upgradable count alone rather than padding the sentence with a
zero. `Todo.breakdown` is optional for exactly this reason, and its absence is the assertion;
`todo-badges.ts` carries the argument and `todo-notice.test.ts` pins it.

**The Agents trail had to become dismissible**, since the brief puts a `忽略` on it and
`dismissTodo` is the only dismissal. Its gate moved from `update-badges.ts`'s boolean
`anyKernelOutdated` to `todo-badges.ts`'s `kernelUpdateTodo`, because a gate that can be waved
away has to name *what* is waiting. **Its signature is the set of outdated Agent ids, not
id-plus-generation**, which makes it coarser than the other three: a later defaults generation
does not re-raise the dot for an Agent already dismissed. That is a deliberate limit — the
generation stamp a sharper signature needs (`KERNEL_VERSION`) sits in core's `state/` beside
`node:crypto`, and pulling that into the browser bundle to sharpen a dismissal is a bad trade.
What makes the coarse form safe is that dismissing silences only the block and the nav dot: every
outdated Agent keeps its `UpdatePill` on the list card and its update button in settings, so
nothing waved away becomes unreachable. A newly outdated Agent still raises it.

**A partial failure names the targets.** The Skills page's `Promise.allSettled` + first-error
toast is the precedent and the shape is kept, but on a control whose whole point is "all of them
at once" a count with no names leaves the user re-checking every row by hand. `bulkOutcome` pairs
the settled results back onto the labels they were launched from and the toast reads
`已更新 N 个；以下未成功: alpha、beta — <first error>`. Both pages reload the Agent list even after a
partial failure, because some Agent moved and the gate has to see it.

**Placement.** All four blocks sit directly under the `<h1>`, which is the one-shape rule
`todo-notice.tsx` already documents. The brief said the Cost Center's line sits inside the errors
panel above the table; it does not — it is under the title, and the code comment there states
that as deliberate ("not down against the errors table") because it counts the probe's own
trailing window rather than whatever the filters are showing. I left it. `errors-panel.tsx` is
untouched.

**Button order.** `忽略` sits left of `现在升级`, the reverse of the order they are listed in the
request. Every `ConfirmModal` in the app puts Cancel left and Confirm right, and this button opens
one of those dialogs a moment later; leading with the affirmative action would put it in the slot
the whole app uses for Cancel. Easy to swap if you would rather have the literal order.

## The consequence warning, per page

Each reuses the page's existing per-item wording rather than inventing copy, and each lists
exactly what the batch would write to.

- **Agents** — `S.agent.kernelUpdateConfirmBody`, verbatim: a smart merge that advances the
  settings tabs the user has not customized and leaves the edited ones whole, listing them in the
  result; name, description, State version and MCP servers unaffected. Lists the Agent names.
- **Skills** — the per-Skill `updateConfirmWarning`'s sentence with the single subject dropped
  (`S.todo.skillsConfirmBody`): an overwriting reinstall of the library copy over each Agent's
  installed files, local edits lost, export a backup first. Lists the Skill names.
- **Models** — `S.models.syncCatalogHint`, verbatim: missing entries added, differing ones reset
  to the catalog's fields, locally added models and API keys untouched. Lists the
  `provider/modelId` refs.
- **Cost Center** — no action and no dialog. A past error cannot be updated, so the block renders
  `标记为已读` alone.

## Scope kept out

`packages/core/src/state/model-catalog.ts`, the models page's provider-group ordering,
`runtime/messaging/**`, `features/messaging/**` and `features/usage/errors-panel.tsx` are all
untouched. The two string dictionaries gain one contiguous block each inside `todo:`.

## Compatibility

`ui_prefs.todoDismissed` gains a fourth key, `agents`. A map written before this change has no
such key, which `parseTodoDismissMap` already reads as "nothing dismissed" — the Agents block
appears once and can be dismissed again. No migration, no stored marker lost, and no compatibility
shim to remove later. Recorded in the changelog entry's `## Compatibility` section.

## Verification

Full CI-parity chain on Node v24, from a clean worktree off `origin/main`:

- `pnpm -r build` — pass
- `pnpm typecheck` — pass (core, server, cli, web, desktop)
- `pnpm test` — pass

New and extended coverage in `packages/web/test/`:

- `todo-notice.test.ts` (new) — the block is on exactly the four trails and nowhere else; every
  site carries a way down; a bulk action never appears without its label; only the Cost Center
  omits it; **every `onAction` handler opens a confirmation instead of calling the API** (parsed
  out of the real JSX with the TypeScript parser, the `disclosure-anchor.test.ts` technique) and
  every page offering one renders a `ConfirmModal`; `noticeCounts` never reports more than the
  gate raised the dot for; `bulkOutcome` names the failed targets and pairs them by position.
- `todo-badges.test.ts` — `kernelUpdateTodo`'s gate, ordering, dismissal containment, and that it
  reports no added/upgradable split; the same for `skillUpdateTodo`; `presetUpdateTodo`'s
  breakdown always summing to its count.
- `catalog-sync.test.ts` — the whole chain the Models page walks, delta → raised to-do → notice
  counts, asserted against `syncRowsWithCatalog`'s own numbers, so the block and the button cannot
  drift apart.
- `skill-outdated.test.ts` — `skillUpdatePlan`: one request per Agent carrying every Skill it is
  behind on, distinct Skills matching what the notice counts.
- `update-badges.test.ts` — `anyKernelOutdated`'s cases moved out with the gate they covered.

`disclosure-anchor.test.ts` and `icon-scale.test.ts` both pass: no `InfoPopover` and no new
hand-written `<svg>` were added.

Not manually exercised in a browser — the change is behind a badge that needs a Project with
outdated objects to raise it.


hiyouga:
## Review — reviewed in a worktree at `1cd1ff1d`, and clicked through in a browser

I stood the built server up on a scratch data root, seeded a Project with two agents behind the defaults generation, three agents behind on two Skills, and a model table three entries short of the catalog with two more drifted. All three notices render correctly and all three bulk actions land:

| page | notice | confirm | after |
| --- | --- | --- | --- |
| Agents | `Changes detected: 2 to upgrade` | lists Beta Agent, Gamma Agent | notice gone, dot down |
| Skills | `Changes detected: 2 to upgrade` | lists `data-analysis`, `web-design` | notice gone, dot down |
| Models | `Changes detected: 3 new, 2 to upgrade` | lists all five refs | notice gone, dot down |

No console errors, no unhandled rejections. The architecture is right and the shared block is the correct place for this. Seven things to fix.

### 1. should-fix — a rejecting `reloadAgents()` wedges the page permanently busy

`agents-page.tsx:385-387` and `skills-page.tsx:310-312`. `reloadAgents` is `try { … } finally { … }` with no `catch` (`state/project.tsx:110`), so it rejects on a failed `GET /agents` — a restart, a 401 after the writes, a network blip. The call site is `void runKernelUpdates(...)`, so the rejection is swallowed and `setKernelRunning(false)` / `setKernelConfirmOpen(false)` never run.

`ConfirmModal` disables both Cancel and Confirm on `busy` (`confirm-modal.tsx:97`, `:105`) and `TodoNotice` disables both of its buttons (`todo-notice.tsx:89`, `:99`). Esc still closes the dialog, but the notice's two buttons stay dead until the page remounts, with nothing on screen saying why.

The Models page already does this right — `void syncPresets().finally(() => { setSyncing(false); setSyncConfirmOpen(false); })` (`models-page.tsx:1295`). The other two want the same shape. Note the pre-existing per-Skill update deliberately uses fire-and-forget `void reloadAgents()` (`skills-page.tsx:274`) and cannot wedge; the new code changed that to `await`.

### 2. should-fix — the Skills flow reports three numbers in two units

Three agents each behind on the same one skill produces:

- notice: `Changes detected: 1 to upgrade`
- confirm: `Update 1 skill(s)`, listing `web-design`
- toast: `3 updated`

And on partial failure, `1 updated; these did not: Beta, Gamma` — agent display names, under a dialog that just listed skill names. `bulkDone: (ok) => "${ok} updated"` names no unit, which is what lets the collision stay silent; Agents and Models happen to agree, Skills does not. Either report in skill units on success, or give the Skills page a pair of strings that say "agents".

### 3. should-fix — the Models notice's own button can dead-end

`persist` (`models-page.tsx:642`) writes and refreshes `rows` but never calls `refreshProjectTodos`; the only call is inside `syncPresets` at `:718`, after the write. So: the dot is up for one drifted entry, the user fixes that row by hand and saves, the probe cache is now stale, they press **Update now**, and `syncPresets` takes its early return at `:705` — `toastInfo(syncUpToDate)`, no refresh. The notice still says a change is waiting and pressing it again does the same. Only Dismiss or a reload gets out.

The staleness is pre-existing; what this PR adds is a button that walks the user into it. One line at `:706` closes it.

### 4. should-fix — the Chinese changelog's headings are untranslated

`2026-08-28-bulk-update-notice.zh.md:14,30` ship `## Details` and `## Compatibility`. `changelog/README.md` § "Chinese counterpart" gives the required renderings — `## 细节` and `## 兼容性` — and every other entry in `changelog/unreleased/` follows them (34 use `## 细节`, 7 use `## 兼容性`).

Everything else in the pair checks out: metadata order, no `Breaking`, counterpart links both ways, sections mirroring one for one, and the Compatibility claim matching `parseTodoDismissMap`.

### 5. should-fix — two "three trails" comments were missed

`strings.ts:261` and `strings-en.ts:257` still describe *three* dismissible trails and claim the notice carries the dot's tooltip sentence down "unchanged". This PR makes it four, moves the kernel trail into this block, and replaces `skillUpdates(n)` with `changesUpgradable(n)`. Every other header in the diff was updated three→four; these two are the only stale hits left in the package.

### 6. should-fix — `todo-notice.test.ts`'s headline assertion cannot fail

`:117-121` asserts each page's source `includes("<ConfirmModal")`. At the base commit `agents-page.tsx` already had one, `skills-page.tsx` two, `models-page.tsx` three — so it was true before the feature existed and stays true if the bulk confirm is deleted. `:107-115` only checks the *attribute text* for `api.` and `await`, so `onAction={runBulkNow}` where `runBulkNow` writes directly passes both. The rule the file header names as rule 2 is therefore not defended anywhere. Separately, `:79-84` asserts `dismissLabel` and `onDismiss` are present, but both are non-optional props — `tsc` already rejects a site without them.

About 96 of the 199 lines are the source scan and four of its six assertions are inventory or tautology, while the two real defects above (the unit mismatch, the busy-lock) are exactly what nothing pins. Worth trading some of that scan for an assertion that the confirm's count and the toast's count come off the same collection.

### Smaller

- `bulk-update.ts:36` — `BulkTarget` has no reference anywhere in the repo.
- `models-page.tsx:953` — `syncCounts.added ?? 0` throws away the `number | null` distinction `noticeCounts` exists to express, and would silently print the padded zero the design notes argue against. Branch on `added === null` instead.
- `skills-page.tsx:479` — the confirm's `<ul>` has no `max-h-60 overflow-y-auto`, unlike both siblings. `Modal` caps the body at `max-h-[70vh]` and the button row is inside that scroll container, so with ~30 outdated skills Cancel and Confirm go below the fold.
- The changelog does not record that `dismiss` changed from "Mark as handled"/「标记为已处理」 to "Dismiss"/「忽略」 on every page carrying the block. And "A partial failure now names the targets" reads as global; the per-Skill `updateOutdated` still toasts the first error alone.
- `outcome.failed.join(listSeparator)` is uncapped. Twenty agents is a wall of text in a 6-second toast. Text is React-escaped, so length is the only issue.

### Design docs

`design/specs/06-PROTOTYPE.md` documents these four pages at exactly this granularity — L121 covers the per-Skill update button *and its confirmation wording*, L131 covers 「同步预置」 down to its toast — but says nothing about the badge trails, the page notice, the dismissal model, or the bulk buttons. The gap starts at the update-badges change and this PR widens it with a fourth notice, a new persisted `ui_prefs` key and three bulk-write controls. Per `CLAUDE.md` that wants a design-repo change; say the word and I will open it.

### Checked and clean

- `bulkOutcome` index pairing is sound: both callers build `labels` and the promise array from the same array in the same `.map`, with no filter or sort between them, and `allSettled` preserves input order. Zero targets, one target, duplicate labels, an empty label and a non-`Error` rejection reason all behave.
- Double-submit is impossible: each `setRunning(true)` is synchronous before the first `await` inside a discrete click handler, and `Modal`'s backdrop covers the page.
- `kernelUpdateTodo`'s signature is order-stable via `.sort()`; deleting or renaming an agent behaves; `SEMANTIC_ID_PATTERN` forbids commas so the `join(",")` round-trip is safe.
- `parseTodoDismissMap` tolerates a map with no `agents` key, a stale key, a wrong-typed value, and `null`/array at either level.
- `added + updated === refs.length === todo.count` by construction in `catalogDelta`. `raisedTodo` returns the whole todo or null, never a subset, so the notice cannot overstate the dot. Skills: `skillUpdateTodo.count` and `skillUpdatePlan().skills.length` are the same set.
- Permissions: the kernel-update route is member-level and skill installs are explicitly any-member, so the missing `isOwner` gate on Agents and Skills is right, and Models' is right for its owner-only sync.
- Dismissing genuinely silences only the notice and the nav dot — the per-card `UpdatePill` is gated on `a.kernelOutdated`.
- `anyKernelOutdated` has no remaining caller; the now-unused `AgentSummary` import went with it; `syncNote` is still used at `:927`.
- i18n: both `todo:` blocks have identical keys, order and arity. No Chinese outside the zh dictionary. English copy uses lowercase `agent`, and the `agent(s)` plural form matches four existing strings.
- `toneStrip.attention`, `UPDATE_DOT_INLINE` and `ICON_GAP.menu` throughout; no hand-written `<svg>`. `disclosure-anchor.test.ts` and `icon-scale.test.ts` pass.
- Merged with #528 and #530 on top of `origin/main`: `pnpm -r build`, `pnpm typecheck`, 1568 web tests and 1503 server tests all pass.

hiyouga:
Fixed in `a3d21901`.

- Both bulk handlers reset in a `finally` and surface a failed `reloadAgents`, so a reload that fails after the writes land no longer freezes the page behind the busy flag.
- Both bulk toasts name agents, which is what both pages send one request per and what the partial-failure line already listed. The Skills flow no longer reports Skills, Skills, then Agents.
- `syncPresets`'s "already up to date" branch re-probes the to-do cache, so the notice's button cannot dead-end on a stale probe.
- `todo-notice.test.ts` now asserts the link — the state the button sets is the state the `ConfirmModal` is rendered behind — instead of that a `ConfirmModal` exists somewhere in the file, which was true at the base commit. The dismiss-prop assertion the types already enforce is gone.
- The failure list is capped at eight names with the overflow as `+N`.
- zh changelog headings are `## 细节` / `## 兼容性`; the "Mark as handled" → "Dismiss" relabel is recorded; the partial-failure bullet no longer reads as global.
- Dead `BulkTarget`, the `?? 0` that defeated the added/upgradable distinction, and the Skills confirm's unbounded list.

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` clean; 112 files / 1561 web tests pass. Also clicked through in a browser on a seeded Project: all three notices, all three confirmations, all three bulk writes, dots down afterwards, no console errors.

hiyouga:
Design counterpart opened: [penguin-harness-design#78](https://github.com/Prism-Shadow/penguin-harness-design/pull/78) — `06-PROTOTYPE.md` §「导航结构」 widened to four dismissible trails with the bulk action and the Agents signature's deliberate coarseness, a new notice bullet on §「Agents 列表页」, and the Skills / Models badge bullets updated, plus its `08-CHANGELOG.md` entry.

PR #530, opened 2026-08-28
Classify recoverable gateway closes, and let an owner clear the error table

hiyouga:
Two changes to the cost center's error panel: a QQ gateway connection the platform closed and the connector brought straight back stops counting as a defect, and a Project owner can empty the table the panel is showing.

## 1. A dropped gateway connection is not a defect

`[messaging] messaging_connect_failed / gateway connection closed (code 4009)` was reaching the recorder as a plain `Error`, so `messagingErrorKind` had nothing to read and the recorder's non-HTTP default filed it `unexpected`.

The fix keeps the rule typed rather than textual, as `error-kind.ts` insists. `GatewaySession` now reports a close as `MessagingConnectionClosedError`, carrying the platform's close code **and the verdict** — `recovers`, meaning "the next handshake brings this back with nothing changed". The verdict is computed where the reconnect decision already lives; the file's own doc says that state is held nowhere above it, and a boolean the classifier reads is the only thing that has to cross the boundary. `error-kind.ts` reads `recovers` and nothing else.

**Classified routine (`expected`), on the platform's documented handling plus what this connector actually does:**

- **4009 — connection expired.** The platform cycling a long-lived socket. It is the one drop that stays *resumable*, which is why `closeCodeInvalidatesSession` deliberately excludes it: the session keeps its handle and `seq`, RESUMEs on the next HELLO, and no inbound event is missed. Nothing broke and there is no operator action.
- **4900–4913 — the platform's internal-error band.** Documented handling is re-identify, which the session performs. Tencent's own hiccup: nothing an operator holds could have prevented it or can speed it up.

**Left `unexpected`, deliberately — not every 4-thousand code is benign:**

- **4004 (auth rejected), 4014 (intent never granted)** and their neighbours. These *do* reconnect — that is the trap. The backoff retries to a 60-second ceiling forever and meets the identical refusal every time, so the binding is down until a person changes a credential or a console setting. "Reconnects" is therefore not the criterion; "reconnecting fixes it" is.
- **4914 / 4915 (delisted / banned).** The session stops for good. The clearest case of needing a human, and now typed with `recovers: false` rather than left untyped, so the type means "the platform closed us", not "this was fine".
- **4006 / 4007 (session handle refused, sequence error).** These reconnect, but only by identifying fresh — whatever arrived in the gap is lost, and a persistent one points at this client's own sequence bookkeeping. The module's stated safe direction is to leave the unclassified alone, and these are close enough to a real fault to keep there.
- **Not platform closes at all**: a handshake that never completed, a gateway that stopped acknowledging heartbeats, network failures. Untyped, so unchanged.

**Other channels have no equivalent.** Feishu's long connection is the vendor SDK's own reconnect loop and surfaces a plain failure with no close code; Telegram is a long poll with no connection to close. The class therefore lives beside the QQ gateway — the only connection here that learns *why* it was closed — with the classifier importing it, the same direction `media.ts`'s three error types already point.

## 2. A clear action on the errors table

`DELETE /api/projects/:projectId/usage/errors`, offered in the table's footer behind the standard `ConfirmModal`.

**Scope — the filter on screen, not the Project's whole history.** The panel's stats and rows *are* that set; a clear that removed more than the reader ever inspected would destroy data they never saw. It is also the recoverable direction of the two: someone who wants everything gone widens the range and clears again, while someone who wanted one noisy week gone cannot un-delete a year. The confirmation states the range, the Agent when one is selected, and the row count, so what it names is what goes.

**Authorization — Project owner.** This matches the routes around it: agent deletion is gated `requireProjectOwner` with the comment "Deletion is a Project-level management operation: owner only", and it already cascade-deletes error rows via `deleteByAgent`. Membership is enough to *read* the panel, but the rows are the Project's shared history and one member emptying them takes them from everyone. A member gets 403 `owner_required`; a non-member 404, which does not leak the Project's existence. The panel hides the action from a non-owner rather than offering one that can only 403.

**Unattributed rows are outside every clear, admin included.** `ErrorFilter.includeGlobal` and the route's `includeGlobalErrors: c.var.user.isAdmin` govern *reading*; the delete has no counterpart, by design. Those rows carry no `project_id` and are surfaced in **every** Project's admin view, so a Project-scoped clear that took them would empty them out of every other Project's panel too — the same line `deleteByProject` and `deleteByAgent` already draw ("unattributed errors belong to no Project and are unaffected"). The result is stronger than "you may not delete what you cannot see": the delete's reach is *strictly narrower* than any caller's read, so the question of trusting an `isAdmin` flag at a delete site never arises.

**The delete cannot outrun the read.** `ErrorsRepo.deleteFiltered` builds its `WHERE` from the same private `conds()` the SELECTs use, and its parameter type is `Omit<ErrorFilter, "includeGlobal">`, so narrowing a filter narrows both halves together and there is no second expression to drift.

`kind` is not accepted by the delete: the panel offers no such control, so a clear can advertise no narrowing its reader could have seen.

## Files touched under `runtime/messaging/`

Only `error-kind.ts` and `qq-api.ts` (the connector raising the close). No other file in that directory is modified.

## Verification

Full CI chain on Node v24, all green:

- `pnpm -r build`
- `pnpm typecheck` — 8 packages
- `pnpm --filter @prismshadow/penguin-server test` — 108 files, 1503 tests
- `pnpm --filter @prismshadow/penguin-web test` — 112 files, 1540 tests

New coverage:

- `test/messaging-qq-transport.test.ts` — the verdict asserted at the socket, not on a hand-built error: 4009 / 4900 / 4913 report `recovers: true`; 4004 / 4014 / 4006 / 4007 / 4914 report `false`; every reported close is a `MessagingConnectionClosedError` carrying its own code.
- `test/messaging-error-kind.test.ts` — the typed rule, including that a plain `Error` merely quoting "code 4009" stays `unexpected`.
- `test/errors.test.ts` — `deleteFiltered` honouring range and Agent and sparing unattributed / other-tenant rows; the route clearing exactly its filter; a member refused 403 and a non-member 404, with the seeded rows intact; an admin's clear leaving the unattributed row it can see; a routine close persisting as `expected` and an auth failure as `unexpected` through the recorder.
- `test/errors-panel.test.ts` (new) — the action shown only to an owner and only with rows, the pager unaffected, and the confirm sentence naming the range, the Agent and the count in both dictionaries.

## Notes

Records already on disk keep the `kind` they were written with — nothing rewrites stored rows, and the classification applies from here on. The clear itself is a user-initiated delete of their own log, not a migration, and it is irreversible: an error record has no other copy.


hiyouga:
## Review — reviewed in a worktree at `ff25eeb7`, and reproduced in a browser

The close-code bands are exactly what the body claims. I swept every boundary at the real socket through `createQQTransport`, not on a hand-built error: 4008 false, **4009 true**, 4010 false, 4899 false, **4900–4913 true**, 4914 false, 4915 false, 4916 false, 4004/4006/4007/4014 false, 1000/1001/1006 false. A close with no code yields `recovers: false`. A close during the handshake never reaches the new class — `dropSocket` detaches the socket first and every listener is gated on `this.ws === ws`.

The SQL is sound too: `conds` always anchors `project_id = :pid` when `includeGlobal` is absent, parameters are named-bound (`agentId: "x' OR '1'='1"` deletes 0), `from > to` deletes 0, and `EXPLAIN QUERY PLAN` shows the delete using `idx_error_project_date`.

One blocker, though, and it is on the irreversible half.

### 1. blocker — under "Last hour" / "Last 24 hours" the clear deletes far more than the panel ever showed

`load()` computes the trailing presets' window at load time — `presetTsWindow(preset, new Date())`, which returns a narrow `from`/`to` alongside `fromTs`/`toTs` (`usage-page.tsx:182`). But `errorFilters`, the object handed to `ErrorsPanel` and from there straight into `api.clearUsageErrors`, is built from the `from`/`to` **state** (`:209`), and switching to a trailing preset deliberately leaves that state alone (`applyRange(next, from, to)` at `:302`, "custom keeps the dates already picked").

So the two halves are narrowed by different filters. Reproduced against a live server, with rows seeded 70, 40, 10 and 3 days back plus eight from the last hour:

```
Range = Last 90 days
  table: 10 rows, oldest 2026-06-19
  confirm: "Deletes this Project's 12 error records between 2026-05-31 and 2026-08-28."

Range = Last hour            <- switched, nothing else touched
  table: 8 rows, all 2026-08-28
  confirm: "Deletes this Project's 8 error records between 2026-05-31 and 2026-08-28."
```

The count comes from the one-hour read; the range comes from the stale 90-day state. Confirming would take all 12 rows, including four the reader had no way to see, under a sentence promising 8 and "records outside that range are kept". From the page default the same switch costs 7 days.

This falsifies the PR's central claim — "narrowing a filter narrows both halves together". The same divergence already exists in the pager (page ≥ 1 fetches with the wide filter while page 0 comes from the narrow read), but a read mismatch is cosmetic; attaching an irreversible delete to it is not.

Hoisting the range out of `load` and memoizing `errorFilters` on it fixes both.

### 2. should-fix — `Omit<ErrorFilter, "includeGlobal">` is not the guard the body says it is

TypeScript's excess-property check fires only on a fresh object literal. A value already typed `ErrorFilter` is structurally assignable to the `Omit`, and `conds()` reads `f.includeGlobal === true` regardless of the declared parameter type. I checked both ways: `tsc --strict` accepts a variable, an opaque `ErrorFilter` and even `{ ...wide }` — only the literal errors — and at runtime `deleteFiltered("p1", wide)` with `includeGlobal: true` deleted **2** rows including the unattributed one.

No caller does this today, so it is latent rather than live. Strip it in the method instead of in the type: `this.conds(projectId, { ...f, includeGlobal: false })`.

### 3. should-fix — a routine close now consumes the outage's only report slot

`fail()` reports at most once per outage (`qq-api.ts:833`) and `reported` resets only on a successful handshake. The PR does not touch that, but it changes what the single report means.

The token is revoked in the console while the socket is healthy. The platform then expires the long-lived socket with 4009 on its normal cycle → record #1, now `expected`. Every retry after that is refused 4004, forever, and `fail()` sees `reported === true`, so nothing further is recorded: no second row, no `setStatus`, no `lastConnectionError` update. The binding is permanently down and the cost center's unexpected count stays at **0**, with `lastConnectionError` still reading "gateway connection closed (code 4009)". Before this PR that first record was `unexpected`, so the outage at least raised the dashboard and the nav badge.

Verified by driving the real transport over a fake socket — handshake, `drop(4009)`, then 4004 on every retry: `reports for the whole outage: 1 [[4009, 'expected']]`.

This is the risk `error-kind.ts:28` names in its own words: "a real fault miscounted as routine is invisible." Letting a later, worse failure in the same outage claim a second report would close it; the backoff plus the recorder's dedup window already bound the volume.

### 4. should-fix — the binding panel still calls a routine close a failure

Related, and the reason the classification only half-lands. `recordConnectionFailure` (`bridge.ts:1002`) sets `state: "error"` and stamps `entry.lastConnectionError` for **every** close, 4009 included. `onReady` resets the live state on the next handshake, but `lastConnectionError` is only cleared when the entry is recreated (`:930`) and is surfaced whenever it is non-null (`:847`). So the exact noise this PR removes from the error table stays pinned to the binding's own panel until a re-sync or a restart. Gating those two writes on the `recovers` verdict the PR just introduced is the natural fix.

### 5. should-fix — for an admin the confirm counts rows the clear cannot take

`pagedTotal` is seeded from `errors.total`, computed with `includeGlobalErrors: c.var.user.isAdmin` (`usage.ts:82`). The delete has no counterpart, by design. In the common single-operator deployment the admin *is* the Project owner, so this is the default path.

A 7-day window holding 1 project error and 1 unattributed row shows two rows, the confirm says "2 error records", one is deleted, and the toast says "Deleted 1 error records" with the other still on screen. In the degenerate case where the only rows in range are unattributed, the button still shows, the confirm promises N, and 0 come back. Verified: `summary(…, {includeGlobal: true}).total === 2` while `deleteFiltered(…) === 1`. This PR's own test already shows the second half — `before.items` is `["uncaught_exception"]` and `deleted` is 0.

### 6. should-fix — the badge and its notice go stale after a clear

`onCleared={() => void load()}` refetches the dashboard but not the todo probe, which `use-project-todos.ts` caches per Project at module level for the browser session. Clear every row and the "N unexpected errors" notice and the sidebar dot both stay up until a reload — the exact state that module documents itself as existing to prevent. `refreshProjectTodos(projectId)` alongside `load()`.

### Smaller

- `errors-panel.tsx:36` — a cleared date `<Input type="date">` gives `""`, which `optionalDateParam` reads as absent, so the delete becomes unbounded below while the sentence reads `between  and 2026-08-27. Records outside that range are kept.` The second sentence is then untrue.
- `strings-en.ts:2144,2148` — `"1 error records"` and `"Deleted 1 error records"`. This dictionary pluralizes counted nouns everywhere, including `unexpectedErrors` at `:265` in this exact domain.
- `errors-panel.test.ts:98-103` asserts only that two strings are non-empty and differ; key presence and arity are already compiler-enforced by `en: Strings`. `errors.test.ts:298-324` builds the error by hand and calls `messagingErrorKind` itself, re-proving that `record({kind})` persists `kind`. The seam neither half pins is the one that matters: nothing asserts the object the socket raised is the object the classifier receives, which is what would survive someone later wrapping the error in `recordConnectionFailure`.
- The clear button's hover palette is hand-rolled (`red-200` / `red-50` / `red-600`) rather than a tone token, and the change from `text-rose-500` to `text-red-600` on `pageError` is unrelated to this PR.
- `packages/docs/content/server-api.{en,zh}.md:231` carries the route inventory and lists `GET /usage/errors`. The table is complete down to "API completeness only" routes — twelve DELETEs are documented — so the new one belongs there.

### Design docs

`design/specs/05-ARCHITECTURE.md` § 「异常落库」 states the messaging rule as 类型 + 捕获点 with a closed allowlist, and then explicitly: 「其余捕获点无论抛出什么都留在 unexpected」. This PR makes `messaging_connect_failed` — a capture point that sentence excludes — return `expected`. That is a contradiction, not an extension, and `CLAUDE.md` puts the design change first. The same section's cascade-cleanup sentence wants the owner-only clear beside it, 「API 原型」 wants the new route, and `06-PROTOTYPE.md` § 「成本与统计中心」 wants the footer action. Say the word and I will open the design PR.

Worth knowing separately: the local `design/` checkout is 23 commits behind `origin/main` and does not contain that paragraph at all, so anyone reading the symlink from a worktree sees a spec that never mentions the rule.

### Checked and clean

- The typed error reaches the classifier intact: `onError` → `qq-connector.ts:374` → `bridge.ts:956` → `recordConnectionFailure` → `recordError` → `messagingErrorKind`. Same object, in-process, no serialization, no re-throw, no wrap.
- No import cycle — `qq-api.ts` imports only `undici`.
- No control-flow change in the reconnect path; the only edit at both call sites is the constructor. All 17 existing transport tests pass.
- `deleteFiltered` cannot reach `project_id IS NULL` or another tenant with any route-reachable filter, confirmed against seeded unattributed, other-tenant and out-of-range rows. A single DELETE is atomic; there is no aggregate table or cursor to keep in step, and the id gaps are already handled by `pruneOverflow`'s exact COUNT.
- Route: correct path and method, `requireProjectOwner` gives member 403 `owner_required` and non-member 404, `optionalDateParam` is strict, `jsonOnlyWrites` does not reject a bodyless DELETE, unknown query params ignored like the two GETs.
- Web: double-submit blocked twice over; `page` does reset after a clear via `useEffect(() => setPage(0), [recent])`; the button is hidden on an empty later page; `canClear` is owner-only and the route enforces it independently.
- i18n keys, placeholders and arity are compiler-enforced; no CJK outside the zh dictionary and the `.zh.md` files; "Project" capitalized and "agent" lowercase, correctly.
- Changelog pair matches `changelog/README.md` and states the irreversibility plainly in both languages. It does inherit finding 1: "the date range and Agent currently selected… Rows outside it are kept" is not true under the trailing presets.
- Merged with #528 and #529 on top of `origin/main`: `pnpm -r build`, `pnpm typecheck`, 1568 web tests and 1503 server tests all pass.

### Uncertain

Whether Tencent's own close-code table supports the 4009 and 4900–4913 semantics. I verified the code implements exactly the bands claimed and that they are internally consistent with `closeCodeInvalidatesSession` and `closeCodeIsFatal`, but I could not check the platform docs from here.

hiyouga:
Fixed in `3dd9aebb`.

**The blocker.** The window the dashboard actually read is now held alongside the response, and the panel pages and clears against it rather than the picker's state. Re-ran the same reproduction on the rebuilt app:

```
Range = Last 90 days
  confirm: "Deletes this Project's 21 error records between 2026-05-31 and 2026-08-28."

Range = Last hour
  confirm: "Deletes this Project's 17 error records between 2026-08-28 and 2026-08-28."
```

The pager was reading through the same stale pair, so that is fixed with it.

**The rest.**

- `deleteFiltered` passes `includeGlobal: false` explicitly. The `Omit` stays as documentation, but it is the query that enforces it now.
- The errors payload carries `clearable` — `total` minus the unattributed rows no Project-scoped clear removes. The confirmation states that number, and the action is hidden when nothing in the window is deletable. It is also hidden while a date bound is blank, since an open-ended range is wider than "records outside that range are kept" can promise.
- A routine close no longer keeps the outage's one report slot: `reported` tracks the verdict, and the first non-routine failure in the same outage gets its own record. Test drives the real transport — 4009, then 4004 forever — and asserts two reports and no more.
- Clearing re-probes the to-do cache, so the badge and the notice go down with the rows.
- `DELETE /usage/errors` is in the server API reference in both languages.
- English plurals on the three clear strings.
- New test for the seam neither half pinned: the object the socket raised, handed straight to `messagingErrorKind` the way `bridge.ts` hands it over.

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` clean; 108 files / 1505 server tests and 112 files / 1541 web tests pass.

**Left alone, on purpose.** `recordConnectionFailure` still stamps `state: "error"` and the sticky `lastConnectionError` for a routine 4009, so the binding panel keeps calling it a failure until a re-sync. Gating those two writes on `recovers` is the natural fix, but it means touching `bridge.ts`, which your scope note excludes, and the field's own comment says it is kept across recovery deliberately. Your call.

The design-spec updates are also still open — `05-ARCHITECTURE.md` § 「异常落库」 says 「其余捕获点无论抛出什么都留在 unexpected」, which this contradicts. Say the word and I will open the design PR.

hiyouga:
Design counterpart opened: [penguin-harness-design#79](https://github.com/Prism-Shadow/penguin-harness-design/pull/79) — `05-ARCHITECTURE.md` §「异常落库」 gains the second form of `expected` (the sentence that said the opposite is narrowed) and the owner-only clear with its read-and-delete-share-one-window rule, the QQ bullet gains the routine-versus-defect axis, and `06-PROTOTYPE.md` §「成本与统计中心」 gains the footer action, plus its `08-CHANGELOG.md` entry.

PR #531, opened 2026-08-28
TokenDance leads the models page, and its promoted rows show what they cost

hiyouga:
Three model-catalog changes: Tencent's Hy4 preview joins through both of its sellers, six
TokenDance rows gain a declared discount the UI shows and the cost center bills at, and the
Chinese UI renames the three price buckets.

## Hy4 preview, twice

`tencent/hy4-preview` on OpenRouter: 1,048,576-token context, text in and text out, $0.834
input / $2.501 output / $0.042 `input_cache_read` per million tokens — read from the model
page and confirmed against `GET https://openrouter.ai/api/v1/models`, which the block comment
names as authoritative where the two disagree.

`hy4-preview` on TokenDance: the same upstream model reached through a second seller, at CNY
6 / 18 / 0.3 (input / output / cache hit) over its own 1,024,000-token window. Specs come from
the group's usual source, the credential-free catalog API. Its `description` carries **no**
bracketed 限时 tag, so it is not on a promotion and list price and billed rate coincide; its
`supported_protocols` are `openai:chat-completions` **and** `openai:responses`, so the
`openai-chat` pin is this group's convention rather than a constraint — the entry's comment
says so, the way `qwen3.8-flash`'s does and unlike `glm-5.3-flash`, where the pin is forced.

The pair follows the rule the two `qwen3.8-flash` rows already set: each row records what its
own seller charges, and the test pins that they stay apart rather than converging.

## How the discount is modelled

The catalog stores list prices. So a promoted row keeps its list `pricing` and declares
`discount` — a fraction off, applied to every bucket — and the new `effectivePricing()`
derives the billed rate. A lapsed promotion is then one field to delete, with the rate to
return to still on the row.

**The billed price is the discounted one**, and the cost center prices exclusively against a
Project's stored `[[models]].pricing` (`ProjectConfigService.getPricing` → `UsageService`; the
catalog is never consulted at query time). So `presetModelEntries()` — the one seam between
catalog knowledge and what a Project stores — writes the *effective* rate. List price and
promotion stay in the catalog, where they can be read and restored; the number a Project holds
is the number the gateway charges.

The models page reads both: the row's stored price is the live figure, and the catalog entry
supplies the struck-through list price and a `-50%`-style badge. `discountedPrice()` returns
nothing unless the row's stored price still equals the catalog's effective price, so an edited
price (or a Project that has not synced presets yet) is never decorated with a saving nobody
is getting.

## The two DeepSeek list prices

For four of the six rows, stored × discount already reproduced the rate TokenDance bills. For
`deepseek-v4-flash-0731` and `deepseek-v4-pro-0813` it did not — the stored value *was* the
billed rate, i.e. the promotional price had been recorded as the list price. Per the product
owner, both were doubled, to `cny(0.1, 3, 9)` and `cny(0.3, 9, 27)`, so 50% off reproduces
CNY 1.5 / 4.5 / 0.05 and CNY 4.5 / 13.5 / 0.15 and the six rows are consistent. This is
effectively price-neutral for what a Project stores (the effective price is what was there
before, to within a 1e-6 USD rounding step on one bucket); it makes the catalog's own list
price honest and lets the badge show.

The test asserts the CNY prices **recovered from `effectivePricing()`**, not pasted constants,
so changing a list price or a discount without the other fails.

## A user with a custom group order

Unaffected. `orderModelGroups` places the groups a stored order names in that order and trails
the rest, and `commitModelGroupOrder` materialises **every** group key on the first drop — so
any Project that has ever dragged a group header holds a complete arrangement, and re-curating
`MODEL_PROVIDERS` cannot move a single group for it. Only a Project with nothing stored
follows the new default. `model-group-order.test.ts` now pins that.

The same applies to which group opens on a first visit: `defaultExpandedProviders()` was
hard-coded to `deepseek` while its own doc called it "the first group in MODEL_PROVIDERS", so
it now reads the leading group from the catalog. A Project with a stored expanded set keeps it.

## Chinese price labels

`cache_read` / `cache_write` / `output` read 缓存命中 / 缓存未命中 / 输出 wherever a Chinese
user sees them: the models page's price fields, the cost center's bucket labels (which drive
the token donut, the trend chart and the legend), and the CLI's `--price-*` option help.
English labels, field names, types and wire keys are untouched. 缓存未命中 is right for
`cache_write` because this catalog uses that bucket to carry the plain input price for vendors
with no separate cache-write fee.

Two neighbours were deliberately left alone: `traces.cacheHit` ("命中缓存") labels a Trace
turn's cache-hit token count, not a price bucket, and `usage.legendHitRate` ("缓存命中率") is a
rate.

## One thing to flag

The TokenDance catalog API's bracketed 限时 tags do not agree with the confirmed discounts on
three rows: it shows no tag for `glm-5.3` or `deepseek-v4-flash-0731` (both discounted), and
八折 for `qwen3.8-max` where the confirmed rate is 10%. The confirmed rates are what shipped —
they are arithmetically consistent with the stored list prices and the target prices — and the
block comment no longer claims the tag is exhaustive or authoritative on the rate, only that it
is a cheap anonymous first check on whether a promotion is still live.

## Verification

Full CI-parity chain on Node v24.18.0, all green:

- `pnpm -r build`
- `pnpm typecheck` — 9 packages
- `pnpm test` — 4761 passed, 5 skipped, 0 failed (core 1021, server 1495, web 1538, cli 383,
  desktop 175, landing 69, docs 53, skills 27)
- `git diff --check` — clean

Tests updated per the catalog fan-out: `packages/core/test/model-catalog.test.ts` (provider
order, the OpenRouter and TokenDance id lists, the derived discount prices, the two-seller
pairing, preset entries carrying the effective price), `packages/web/test/model-grouping.test.ts`
(provider order + `discountedPrice`), `model-group-order.test.ts` (a stored order survives a
re-curation), `model-group-expansion.test.ts` (the default follows the leading group),
`protocol-path.test.ts` (the gateway loop was missing `tokendance`), and
`packages/server/test/models.test.ts` (a new Project is preset at the billed rate, not list).

Not manually clicked through — the discount decoration and the group caption are covered by
unit tests over the pure helpers, and the JSX change is confined to the model card's meta line
and one caption above the group section.


hiyouga:
## Second pass on the card

- **Tags are one shape now** — pale wash, a border one step stronger in the same hue, text darker again — spelled once in a local `TAG_CLASS` map. They stay out of `lib/tone.ts` because they are identities, not judgements: "default", "vision", "free" say what a model *is*, and none ranks above its neighbour (same reason `category-colors.ts` and `update-dot.tsx` keep their own colours).
- **The upstream id has left the title row.** A detail you look up, not one you scan by, and one click away in the config dialog.
- **The freed width carries lifetime spend** — `17.7B tokens`, grey, a size below the price line. It is unfiltered by date or Agent, so it comes from a new `GET /usage/model-totals` rather than the cost center's filtered aggregate, and from its own request so a stats failure costs the figure and not the page. A model that never ran shows nothing, not a zero. `humanizeTokens` gained a billions tier, web and CLI together.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm test` (4766 tests across nine packages), `pnpm format:check` — all clean. Three new server cases pin the totals (all-time regardless of date or Agent, keyed by the paired reference, scoped to the Project) and one new web case pins the B tier including the M/B boundary.

Rendered against a dev server with synthetic totals spanning `940` to `17.7B`, in both themes and at 1400px and 720px: three rows hold, the empty mark row keeps unmarked cards the same height, and the narrow header still does not overlap its actions.

hiyouga:
## Third pass — the label, plus a bug that turned out to be three

`29.2k toks` on the card, matching how the page already abbreviates the unit (`tok/s`, `/M tok`).

**Clearing a model's display name did nothing.** Reproduced exactly as reported. For a model the built-in catalog names, an absent `display_name` on disk means *inherit the catalog's*, so clearing the field wrote nothing and the read handed the deleted name straight back — the PUT returned 200 while the name never changed. The empty string now records the deletion, and the read tells it apart from an absent field. A model the catalog does not name still writes nothing; absence already says "no name" there.

**The confirmation read `确定保存对「」的配置修改？`** because the fallback to the id only triggered on `undefined`, and clearing a field leaves `""`. The dialog header had the same gap — empty title, then the id repeated underneath. One exported `modelLabelOf` answers it in both places.

**Re-saving a model was editing its price.** Found while verifying the above: the form rounds the stored USD to four decimals so it is typeable, and submit re-encoded that rounding as the new price. Opening a promoted row and pressing confirm without touching anything moved it off the figure its seller bills and silently dropped its discount mark. A bucket still holding exactly what was loaded into it is now written back byte for byte. This one was not reported — flagging it as scope I added.

## On-disk configuration

Clearing a name writes `display_name = ""` into `.project_config.toml`, a value nothing wrote before. No migration, and nothing to do to an existing Project — a config without the field behaves exactly as it did. An older build reading a newer config treats the empty string as absent and shows the catalog name again; the cleared name reappears, nothing else changes.

## Verification

Browser, Chinese locale, against a dev server, full reload between steps:

| | confirmation | name after reload | untouched re-save |
| --- | --- | --- | --- |
| before | `「」` | catalog name returns | `-50%` mark disappears |
| after | `「deepseek-v4-flash-0731」` | model id, persisted | card byte-identical |

Two new server cases (clearing sticks across a second full-table PUT; a re-named model drops the marker, and a non-catalog model never writes one) fail on the old service. Seven new web cases pin `modelLabelOf` and `priceToSubmit`, including the CNY round trip.

`pnpm -r build`, `pnpm typecheck`, `pnpm test` (4773 across nine packages), `pnpm format:check` clean.

hiyouga:
## DeepSeek peak/off-peak pricing

The rows recorded DeepSeek's **off-peak** tier and costed peak usage at it, halving what peak actually cost. They now record the **peak** price — CNY 0.1 / 3 / 9 for the two V4 Flash rows, CNY 0.3 / 9 / 27 for V4 Pro — and declare `offPeakDiscount`, a weekly schedule that halves every bucket outside Beijing weekday 09:00–12:00 and 14:00–18:00.

**What is on disk stays one stable number.** A flat promotion is a single rate until it lapses, so `sync presets` bakes it into the Project. A scheduled one changes twice a day, so nothing is baked in: the peak price is what goes in `.project_config.toml`, and the reduction is applied when a price is read. A value that meant something different an hour later would be unreadable, and re-syncing would rewrite prices by the clock.

Applied in two places, on the same condition:

- the card prints the reduced figure and marks it `-50%` off-peak; at peak it prints the stored price and carries no mark at all;
- `getPricing` halves what the cost center bills at.

Both bail once the stored price is no longer the catalog's, so a hand-typed number is never halved — nothing here knows whether it is a peak rate. `offPeakAt` is UTC arithmetic, so a server in Los Angeles reaches the same answer as one in Shanghai. The page re-aims a timer at each clock hour, since every window a schedule can express is hour-aligned — a card left open at 08:59 would otherwise still promise half price at 09:05.

Resellers are untouched: OpenRouter, SiliconFlow and Fireworks bill their own rates, and the TokenDance DeepSeek rows keep their own flat promotion.

## Also in this push

- **Marks lose their fill** — a hairline border, no background, hue in the text only. The group header's recommendation pill follows.
- **The config dialog's header is three lines**: name, upstream id, marks. They had shared the name's line, where a long name wrapped underneath them.

## Verification

Browser, clock pinned, after `sync presets`:

| Beijing time | card |
| --- | --- |
| Mon 09:30 (peak) | `$0.0429 / $1.2857 / $3.8571`, no mark |
| Sat 10:00 (off-peak) | `-50%`, `$0.0214 / $0.6429 / $1.9286` — DeepSeek's published CNY 0.15 / 4.5 / 13.5 |

Five new core cases pin the clock rule, including the exclusive window ends (12:00 and 18:00 are already off-peak), Sunday as ISO day 7 rather than `getUTCDay`'s 0, and the same instant judged from a `Date` rather than local time. Two new server cases pin the cost side: one price on disk, halved or not by the hour, and never halved once edited. Two new web cases pin the badge at both hours.

`pnpm -r build`, `pnpm typecheck`, `pnpm test` (4783 across nine packages), `pnpm format:check` clean. Bilingual `models` docs updated.

hiyouga:
## Review — reviewed on this branch rebased onto current `main`, and measured against a running server

The PR was six merges behind and conflicted, so I merged `origin/main` in first and reviewed the result. Two conflicts, both "keep both sides": the route list in `usage.ts`'s header, and two adjacent exports in `endpoints.ts`. One real integration break came with it — `model-oauth.test.ts` pinned `applied: 8` for the TokenDance group and this PR adds a ninth model, so three assertions that are not about the count fail. Deriving the count from the catalog fixes it.

Much of this holds up well. The two doubled DeepSeek list prices are arithmetically exact — `deepseek-v4-flash-0731`'s effective price is byte-identical to what was stored before, and `deepseek-v4-pro-0813` differs on one bucket by 1e-6. `presetModelEntries` writes the effective rate for a fixed discount and the peak rate for a scheduled one, as designed. `discountedPrice` bails on an edited price, an un-synced Project, an unpriced bucket, a non-catalog row and a scheduled row inside its peak window. The schedule's boundaries are right: `[from, to)` end-exclusive, weekends off-peak all day, fixed +480 offset so no DST. Server and browser use the same guard and the same rounding, so the card and the bill agree.

Two blockers.

### 1. blocker — after a sync, the cost center reports a different historical cost depending on the hour it is read

`app.ts:775` feeds the current instant into `getPricing`, `applyOffPeak` reduces the stored rate at read time, and `UsageService.query` prices **every** aggregate from that one lookup — `summary.today`, `last7d`, `total`, the groups and the series. So the whole usage table is re-priced at whichever tier is in force when you open the page.

Measured end to end. I seeded one million tokens in each bucket on a data root written by a pre-#531 build, then opened it with this branch:

| | stored price | reported cost |
| --- | --- | --- |
| before this PR | 0.007143 / 0.214286 / 0.642857 | 0.864286 |
| this PR, not yet synced | unchanged | 0.864286 |
| this PR, after syncing that row | 0.014286 / 0.428571 / 1.285714 | 0.864286 read at Beijing 00:07 |

The last row is measured off-peak. At Beijing 10:00 the same stored config yields `1.728571` — the peak rate, unreduced. Same rows, same Project, no user action, and the number doubles at 09:00 and halves at 12:00 every weekday. Neither figure is right for usage that ran at the other tier.

`applyOffPeak`'s comment argues this is "the same contract, not a new one" because costs are already priced at current pricing rather than the rate in force when each request ran. The difference is that the existing contract only moves when a user edits a price, which is a deliberate act; this one moves twice a day on its own.

Three ways out, in descending order of correctness: price each record by its own `usage_records.ts`, which the column already carries; or keep `offPeakDiscount` display-only and bill a single stored rate, which is what the catalog did before; or accept the oscillation and say so in the changelog and in `models.{en,zh}.md`, neither of which mentions it today.

Related, same seam: syncing also retroactively moves the reported cost of already-incurred usage on five TokenDance rows. Per 1000 Mtok of output, `kimi-k3` goes $14,285.71 → $11,428.57 and `glm-5.3` $4000 → $3600. The changelog says the sync "lowers the six promoted TokenDance rows to the rate they are billed at" but never that the cost center's history moves with them. This is data already on disk changing meaning, which the dev skill routes to the user rather than to the author.

### 2. blocker — three spec sentences now say the opposite of what ships, and there is no design PR

On the design repo's `origin/main`, `06-PROTOTYPE.md` §「模型配置页」 still reads:

- `:150` 「DeepSeek 分组置顶，OpenRouter、Fireworks AI、SiliconFlow、Qwen Token Plan 与 Qwen Pay-As-You-Go 五个网关分组紧随其后」 — this PR puts TokenDance first.
- `:153` 「默认仅 DeepSeek 分组展开 … 无记录时回落仅 DeepSeek 展开的默认态」 — it now follows the leading group.
- `:153` 「卡上为展示名、model_id、状态徽标…与一行小字」 — the card drops `model_id` and gains a spend figure and a discount pill.

Nothing in `specs/` describes a discount mark, an off-peak schedule or a per-model spend figure. Every other PR in this batch shipped its design counterpart (#77–#82); this one has none. Say the word and I will open it.

### Should-fix

**The docs shipped in this PR contradict the card it ships.** `models.en.md:83` and `models.zh.md:83` say the card shows the discounted price "with the list price struck through beside it"; a later commit removed that, and `line-through` now appears zero times in `models-page.tsx`. The changelog agrees with the code ("shows the **billed** price alone"), so the two documents in this PR disagree with each other.

**`tokendance/deepseek-v4-flash-vision-exp` was left at what this PR's own argument calls a promo price.** Its list is `cny(0.05, 1.5, 4.5)` — numerically identical to `deepseek-v4-flash-0731`'s *effective* price — with no `discount` and no comment saying why it is exempt, sitting between two rows that were doubled for exactly that reason. If the seller runs the same promotion on it, its card shows no badge and its list price is wrong by 2×.

**Every Project gets a "10 preset models to sync" badge, and one of the ten is a phantom.** Running `catalogDelta` against `origin/main`'s catalog gives added 2, updated 8. One of the eight is `tokendance/deepseek-v4-pro-0813`, output `1.928571 → 1.928572`, because `3.857143 * 0.5 * 1e6` is exactly `1928571.5` and rounds up. A millionth of a dollar per Mtok. Either pin a list price whose half is exact, or round half to even.

**Two stranded doc comments.** In `usage.ts:93` the new `/model-totals` handler was inserted between the paragraph describing the error-page route and the `app.get("/errors")` it describes, so `/model-totals` now carries "One page of the error detail table" and `/errors` has none. In `model-catalog.ts:305` `effectivePricing`'s block sits above `interface OffPeakDiscount`, and it states something now false — "every caller can use this in place of `.pricing`" — while `presetModelEntries` deliberately does the opposite for scheduled rows. `strings.ts:810` has the same residue: a comment about "the struck-out price" above `visionModelBadge`, left when the struck price was removed.

**The mutual exclusion of `discount` and `offPeakDiscount` is prose only.** A future row declaring both would silently bill the list price at peak. One line in the catalog test closes it.

**`GET /usage/model-totals` has no route test** — no 403 for a non-member, no 200 for a member, no invalid-id case — while one of its three service tests re-asserts the repo's own project scoping. Swapping one for the other is the same number of tests aimed better.

**A Project that dragged a group before TokenDance existed gets it last, under Custom.** `allGroupKeys` is materialised at drop time, TokenDance only entered `MODEL_PROVIDERS` in #465, and `orderModelGroups` trails unplaced groups. The body's "unaffected" is true about stability and false about discoverability. Product call.

### Nits

- `discountedPrice` bails only on `undefined`, so `discount: 0` renders `-0%`, `1` renders `-100%` and also trips the free-model badge, and `-0.2` renders `--20%`. `discount: 0` is a plausible authoring slip precisely because the field's doc says a lapsed promotion is one field to delete.
- `useHourTick` re-aims at the next *local* hour, so a viewer at UTC+5:30 or +5:45 ticks up to 45 minutes off a boundary the server has already crossed — the card shows the peak price with no badge while the bill is already halved.
- Two discount conventions now coexist in one file: OpenRouter's promoted rows still bake the discounted rate into `pricing`, so an equally 50%-off OpenRouter row wears no pill. Worth one line saying that group stays on the old convention deliberately, or migrating it.
- `render.ts` gained the `B` tier but its doc comment and its test did not, while the web side got both — and `format.ts:5` claims the two match exactly.
- `Scope` lists six areas; the changelog spec says one to five. Several bullets are present-tense descriptions of the codebase ("The model card **is** three fixed rows"), which the spec routes to the PR description, and about 40 of the 147 lines are design argument.
- The dev skill's catalog contract still describes pricing as three buckets with a "base tier recorded, knowingly under-costed" convention. Neither `discount` nor `offPeakDiscount` is mentioned, and that tiering bullet now reads as advice this PR departed from.

### Uncertain, and worth settling before this ships

**Whether DeepSeek's peak window really is weekdays only.** `peakDays: [1,2,3,4,5]` is a new factual claim with no cited source, and the file header dates every other price read but not this one. If the windows apply every day, all weekend usage is billed at half the real rate.

Also unsettled: whether `tokendance/deepseek-v4-flash-vision-exp` is on the promotion, and whether the oscillating cost report is intended.

### Checked and clean

- The merge resolution: both route-list lines present, `clearUsageErrors` and `getUsageModelTotals` both intact with the right imports.
- The auto-merged `models-page.tsx` still carries #529's `TodoNotice` with its counts and #532's OAuth dialog with its `done` phase, coexisting with the new recommendation pill.
- Boundary hours empirically: Mon 08:59 off-peak, 09:00 peak, 11:59 peak, 12:00 off-peak, 13:59 off-peak, 14:00 peak, 17:59 peak, 18:00 off-peak; Saturday and Sunday off-peak all day; Sunday's `getUTCDay() === 0` mapped to 7.
- `cache_write` = 缓存未命中 is right — `generative-model.ts:248` defines that bucket as `prompt_tokens`, the input price on a miss. Both dictionaries type-check against one shape, every new key exists on both sides, English is untouched, and no Chinese leaked outside the zh dictionary, the CLI's own catalog and `*.zh.md`.
- `commitModelGroupOrder` materialises every key on the first drop, so a stored order survives the re-curation; a stored expanded set survives; corrupted storage falls back to the leading group.
- The new route is project-scoped, behind `requireProjectAccess`, one grouped scan on an existing index prefix, and the page falls back to an empty map so a stats failure costs the figure and not the page.
- The TokenDance price test recovers its CNY figures *through* `effectivePricing` rather than pasting them, so changing a list price or a discount alone fails. Every time-dependent test uses an explicit `Date`.
- Rebased onto current `main`: build, typecheck and format clean; core 1026, web 1607, server 1610, cli 383 all pass once the `applied` count is derived rather than pinned.

hiyouga:
Correction to my review: the peak-window claim **is** sourced and dated, and I was wrong to list it as unsettled.

`model-catalog.ts:393` carries it directly above the rows that use it — "Re-read 2026-08-18 from api-docs.deepseek.com/quick_start/pricing after the official price increase introduced time-based tiers", followed by the windows themselves. The file header at `:4` dates the same read. The product owner has separately confirmed the schedule is Beijing Mon–Fri 09:00–12:00 and 14:00–18:00, everything else off-peak, against https://api-docs.deepseek.com/zh-cn/quick_start/pricing/ — which is exactly what `DEEPSEEK_OFF_PEAK` encodes and what I verified the boundaries of empirically.

So `peakDays`, `peakHours` and the `[from, to)` end-exclusive handling are all correct, and the weekend-billed-at-half concern does not exist. Nothing else in the review changes: the two blockers and the should-fix list stand as written.

hiyouga:
Merged current `main` in and fixed the review findings — `3dadbdc7` and `35120211`.

**The billing blocker.** Which tier a request ran in is a fact about when it ran, so it is decided from each record's own `ts` rather than from the hour the page is opened. The three priced aggregations now group by `(provider, model_id, peak)` and each half is billed at its own rate. Measured on a running server, one 1M-token record at 09:30 Beijing and one at 20:00:

| | peak record | off-peak record | reported total |
| --- | --- | --- | --- |
| now | 1.285714 | 0.642857 | **1.928571** |
| before, read off-peak | halved | halved | 1.285714 |

The tier expression is built from the schedule rather than hardcoded, so a second vendor's windows cost one catalog entry and no more SQL. `getPricing` stops choosing a tier and returns both, which took the clock back out of the app wiring. A Project that typed its own price gets the same number in both tiers, so the split costs a row and changes nothing — pinned by a test at three different clocks that expects one number.

**The merge.** Two conflicts, both "keep both sides": the route list in `usage.ts`'s header and two adjacent exports in `endpoints.ts`. One real integration break came with it — `model-oauth.test.ts` pinned `applied: 8` for the TokenDance group and this PR adds a ninth model, so three assertions that are not about the count failed. The count is derived from the catalog now.

**The rest of the review.** The docs sentence promising a struck-through list price the card no longer draws; the two stranded doc comments and the `strings.ts` orphan; a guard so `discount: 0` or `1` cannot render `-0%` or a `-100%` "Free" badge; a catalog test pinning that no row declares both a flat discount and a schedule; route tests for `/model-totals` (non-member 404, member 200); the CLI formatter's `B` tier in its doc and its test; `Scope` trimmed to five; and the changelog now says both things it was missing — the preset badge appears for every Project, and accepting the sync moves the cost reported for usage that already happened.

Also corrected in the dev skill: the catalog contract now states the promotion and schedule rules, and there is a note that polling CI on a conflicting PR waits for a merge commit GitHub never builds.

`pnpm -r build`, `pnpm typecheck` and `format:check` clean; core 1027 + 5 skipped, server 1612, web 1607, cli 383.

**Still yours to decide**, unchanged from the review: the 1e-6 rounding row that shows up as a phantom sync entry, `useHourTick` on non-whole-hour offsets, the OpenRouter rows still on the baked-in convention, TokenDance leading but landing last for a Project that dragged groups before it existed, and whether `tokendance/deepseek-v4-flash-vision-exp` is on the promotion. The design PR for the three outdated `06-PROTOTYPE.md` sentences is also still open work.

PR #532, opened 2026-08-28
The key-authorization dialog reports its own outcome

hiyouga:
Minting a provider key with **Authorize key** sends the user to the provider's own page in another tab. The key landing was announced with `toastSuccess` fired the moment the poll saw it — **in a window nobody was looking at**, and faded by the time they switched back. The one step a user leaves the app for was the one step with no visible result.

The dialog now stays open and reports the outcome itself, naming the provider and how many models the key was written to.

## Why the dialog rather than a new modal

It is already on screen and already focused when the user returns from the authorization tab — their eye is on it. Closing it to open a second dialog would replace the surface they are looking at with one that has to fight for attention all over again. So the flow gains a `done` phase instead: the controls and the manual/redirect switch leave, and what remains is the sentence and one button.

Both paths a key can land through settle into it — the redirect flow's poll seeing `status: "done"`, and a manually pasted code answering `ok`. Reporting only the one the author happened to test with is exactly the bug being fixed, one flow over.

**The footer is a single dismissal.** Done is an outcome, not a choice: a cancel beside it would offer to undo a key the server has already stored.

The model table still reloads on success — that is what brings the masked key back.

## The test, and what it does not do

`test/model-oauth-outcome.test.ts` pins the rule rather than leaving it to memory, because the regression is a one-line `toastSuccess` added back by someone who never watched this happen from another tab. It asserts both success paths reach `done`, that neither the dialog nor its caller toasts the outcome, that the sentence carries the provider and the count in **both** dictionaries, and that the done footer offers no cancel.

It reads the real source rather than rendering. The dialog is module-private, and exporting it purely so a test could mount it would widen the module's surface to check a rule about its own text. That is a real limit and it is stated in the file: this catches the shape coming back, not a rendering bug.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm format` + `pnpm format:check` clean; `packages/web` 112 files / 1537 tests pass.

Not exercised against a live provider — the flow needs a real TokenDance authorization to complete. The two success paths are covered structurally rather than end to end, which is stated above.


hiyouga:
Fixes the reported bug: finishing the authorization handed the authorize page back, so the key could never be set.

**Cause.** The dialog rendered behind the models table's `rows &&` guard, and `onApplied` reloaded the table. `load()` blanks `rows` on its first line, so the success it was reporting unmounted the dialog reporting it — and the dialog's mount effect calls `startModelOAuth`, opening a *new* authorization.

**Fix.** Ungate the dialog from `rows`, and move the reload to the dismissal. `count` is read only while the flow is still running, so a reload's momentary absence of rows reads as zero and is never seen.

## Verification

Driven through a real browser against a dev server, with `model-oauth/start` and `model-oauth/flow-*` intercepted to play out a full authorization:

| | outcome sentence shown | `model-oauth/start` calls |
| --- | --- | --- |
| before | no | 4 |
| after | **yes** | 2 |

Two calls per mount is StrictMode's double-invoke; four is the extra mount, i.e. the second authorization the user was seeing.

`test/model-oauth-outcome.test.ts` gained a case that fails on the previous render, pinning both invariants — the dialog is not gated on the loaded rows, and the success handler does not reload.

`typecheck` clean, web suite 1538 passed, `format:check` clean.

hiyouga:
## Review — reviewed in a worktree at `3d3eaf06`

The state machine is sound. `done` has exactly two `setPhase` calls and nothing else targets it; the retry and the mode switch are both hidden there; the poll's cleanup sets `stopped` and clears the interval on every exit path, and both branches re-check it after their `await`, so closing mid-flight, switching modes mid-flight and an error arriving after `done` all land on the same guard. Polling is bounded by the flow's 10-minute TTL. Esc, the backdrop and the header X all still close in `done`.

Two behaviour findings, then conventions.

### 1. should-fix — the redirect path reports a guess, the manual path reports the truth

`models-page.tsx:3205` uses `res.applied ?? count` — the server's real write count. `:3170` uses bare `count`, the client's `rows.filter(...)` guess. They diverge whenever `rows` is not a faithful projection of the config, and the page has one path that makes that happen on purpose: `persist`'s catch at `:665` keeps the locally-attempted rows so the user can fix and retry. After a rejected save, `rows` holds N+1 while the server holds N — and the dialog then states as fact "the API key is set on N+1 models", sourced from state the server refused. A concurrent edit from the CLI, the desktop app or a second tab does the same; nothing refreshes `rows` live.

The server has the number and drops it at the boundary. `http/routes/model-oauth.ts:275` reads `state.applied` to decide whether to publish the credential change, then omits it from the response two lines later, because `ModelOAuthStatusResponse` has no such field — unlike `ModelOAuthCodeResponse`, which does.

Purely additive fix: add `applied?: number` to the status response, pass it through, and make `:3170` mirror `:3205`.

### 2. should-fix — cancelling the dialog now wipes the group's speed-test results

`load()` clears them at `:616`, and its own comment scopes the reason narrowly: keyed by `(provider, model_id)` only, so drop them *whenever the active Project changes*. Before this PR `load()` had two callers, the project-change effect and the success handler. This PR adds a third that fires on **any** dismissal — the base `onClose` was `() => setOauthFor(null)` with no reload.

So: run the group speed test (strictly serial real requests against every model in the group, spending quota, results held only in page memory), open **Authorize key**, change your mind, press Esc. Every speed badge is gone and cannot be recovered without re-spending.

The comment at `:1233` says the unconditional reload "costs one request and cannot be wrong". It costs one request *and* the speed results. The `onApplied` seam at `:1240` is an empty function right now — setting a ref there and gating the reload on it is three lines and gives that seam a real job.

### 3. should-fix — the changelog

- **No `PR:` field.** `changelog/README.md` makes it required when one exists; 99 of 101 entries across the three released folders carry it, including the sibling `0.2.8/2026-08-26-model-oauth-desktop-callback.md`.
- **The `## Details` bullets are the decision rationale.** The spec forbids exactly this — "Reasoning does not go on disk" and "Do not describe the current state of the codebase" — and bullet 4 ("the dialog **is** module-private, and exporting it so a test could mount it would widen the module's surface") is the spec's own example of the forbidden form followed by an alternatives-considered argument. The register the repo wants is in that sibling entry, whose bullets all open in past tense with no argument attached: Mounted, Split, Gave, Answered, Held, Deleted, Recorded. All of this reasoning is already in the PR description, which is where it belongs.

### 4. should-fix — `oauthApplied` is now dead, and the test is coupled to it staying dead

`grep -rn oauthApplied` returns its two definitions and one negative assertion. The trap is that `expect(SOURCE).not.toContain("S.models.oauthApplied(")` is the only assertion in the file that could have caught the bug being fixed — so deleting the dead strings, which is the obvious cleanup, makes it vacuously true forever, since TypeScript would reject any reference to a key that no longer exists. Fix the two together: delete the strings, and re-point the guard at the caller, where the existing test already slices the `onApplied` handler out of the JSX.

### Smaller

- `model-oauth-outcome.test.ts:47` — `not.toContain("toastSuccess")` was already true at the base commit, because the old toast lived in the *caller*, above `type OAuthPhase`, outside the slice. I replayed all eight structural assertions against the base: seven fail there for the right reason, this one passes. It is a fair forward-looking guard, just not the one doing the work.
- `model-oauth-outcome.test.ts:52` — both dictionary assertions pass `n = 7`, so the `n === 1 ? "" : "s"` branch is never exercised. Everything else in that block is compile-enforced by `en: Strings`; the interpolation is the only genuine runtime assertion, and the singular is where the two dictionaries differ in shape. One line.
- `model-oauth-outcome.test.ts:26` — the helper's JSDoc names `KeyAuthorizeDialog`, which does not exist; the component is `ModelOAuthDialog`. Its slice also runs to end of file, so anything appended later is swept into the `toastSuccess` and `setPhase("done")` assertions.
- `models-page.tsx:3264` — a `<>…</>` wrapping a single `<p>`, left over from the edit.
- `models-page.tsx:3247` — the sole dismissal is `variant="primary"`, which no other dismissal in the app is. The exact precedent is `components/account/update-dialog.tsx:100`, a `phase === "done"` footer collapsing to one button written with no variant; `messaging-binding-modal.tsx:45` does the same. In this design system `primary` marks the action.
- Focus is dropped when the footer swaps at the phase flip — the focused "Open authorization page" button unmounts and nothing takes its place, so the next Tab restarts from the top. `Modal` has no focus trap anywhere in the repo, so this is pre-existing platform behaviour rather than a regression, but the phase flip is a new place where it bites, mid-flow rather than at dismissal.

### Design docs

The authorize-key flow is absent from `specs/` entirely — searching the design repo's `origin/main` for 授权页 / OAuth / 授权密钥 returns only QQ scan-bind hits. So this PR inherits the gap rather than creating it. What it does do is carve an exception out of a rule §「模型配置页」 states: 「保存结果、连通性测试结果与一般的成功/信息通知用顶部浮层提示」. Worth recording the flow and its in-dialog outcome together. Say the word and I will open the design PR.

### Checked and clean

- Reopening does not show a stale `done`: the dialog unmounts on `oauthFor → null`, and `applied` is only rendered in `done`, which is only reached after a fresh `setApplied`.
- An unknown `status` is impossible at the type level, and would fall through to the TTL anyway.
- The reload happens on both success paths, via `onClose`, and cannot race `done` into a stale count — nothing reloads until after `applied` is captured. A reload failure surfaces as `loadError`.
- `count` is computed over unfiltered `rows`, matching the server's "every model in the group" semantics, and the Authorize button only renders for a group with at least one model, so it is never zero at open.
- i18n parity is compile-enforced by `en: Strings`; both new entries take `(provider, n)`; no Chinese outside the zh dictionary; the plural form matches the 34 existing uses of the same shape.
- The changelog pair mirrors section for section — the +36/+18 gap is CJK not wrapping at 100 columns, not missing content.
- Reading source text in a test is house style here, not an anomaly: 20 web test files do it.
- Merged with #528, #529, #530, #533 and #534 on top of `origin/main`: build, typecheck, 1600 server / 1592 web / 1021 core tests all pass.

hiyouga:
Fixed in `0282ba75`.

- `ModelOAuthStatusResponse` gained `applied`, the count the route already computed and dropped, so both paths report the server's number instead of one of them counting the table in hand.
- The reload on dismissal is gated on a key having landed. `load` also clears the Project's speed-test results, which are real API quota spent and live only in page memory, so cancelling the dialog was wiping them. That also gives the `onApplied` seam a job.
- `oauthApplied` had no reader left. Deleted, and the test's guard moved onto the caller's own handler — a string the code no longer references is exactly what gets deleted as dead, and that deletion would have made the assertion pass forever.
- The test's source slice is bounded at the component rather than running to end of file, its JSDoc names the component that exists, and the plural branch of the new string is exercised.
- The changelog gained its `PR:` field, and its bullets say what was done rather than why.
- The stray fragment, and the `variant="primary"` on the sole dismissal.

Three server tests asserted the status body with `toEqual`, so `applied` broke them — updated, and the later-poll assertion correctly keeps no `applied`, since only the redeeming poll sets it.

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` clean; 108 files / 1495 server tests and 112 files / 1538 web tests pass.

Left open: the design spec still has no record of the authorize-key flow at all, which predates this PR. Say the word and I will open that design PR with the in-dialog outcome recorded alongside it.

PR #533, opened 2026-08-28
WeChat as a fourth messaging channel, bound by scanning a QR code

hiyouga:
A Session can now bind WeChat's official claw bot, behind the same `MessagingChannelConnector` seam Feishu, Telegram and QQ already sit behind. Binding it is a QR code scanned in WeChat and nothing else — there is no console to copy a credential out of — so the scan is the only path here rather than the convenient one.

## What the references actually let us do

Two MIT-licensed implementations describe the same protocol: Tencent's own [openclaw-weixin](https://github.com/Tencent/openclaw-weixin) plugin and [weixin-agent-sdk](https://github.com/wong2/weixin-agent-sdk), which is derived from it. Neither is taken as a dependency, and the reason is shape rather than licence: both own the process's message loop (`login()` then `start(agent)`), carry ONE account, render the login QR to a TTY with `qrcode-terminal`, and persist state under `~/.openclaw`. This product needs many bindings side by side, each with its own stored config document, and a QR that reaches a browser. So the protocol is reimplemented against the same endpoints behind an injectable seam — the call `qq-scan.ts` already made about Tencent's bind-task package, for a different reason.

It is plain JSON over HTTPS against `ilinkai.weixin.qq.com`, no proprietary binary and nothing to register: the `iLink-App-Id` header is the literal `bot` for every client of this kind.

**Inbound is a long poll** (`ilink/bot/getupdates`, an opaque `get_updates_buf` cursor), so no public URL appears anywhere in the flow — the property that made Telegram's `getUpdates` and QQ's gateway usable here, and QQ's webhook mode not.

**Scan-to-connect** is `get_bot_qrcode` for a handle plus a URL, then a long-polled `get_qrcode_status` until it reports confirmed and returns the bot token, the bot id, the id of the person who scanned, and the API host the bot is assigned. The handle is the security model: it is what polls a token out, so it never leaves the server and the browser gets one this server mints in its place.

## Capability matrix

| | inbound | outbound |
| --- | --- | --- |
| text | yes | yes |
| image | yes (CDN + AES-128-ECB) | yes (pre-signed upload, then an IMAGE item) |
| file | yes | yes |
| video | yes, as a file (`video.mp4` — the platform's own type, not a guessed extension) | not implemented; the seam carries files, and a file send is what a video would be |
| voice | as WeChat's OWN transcription | n/a |

Nothing outbound is refused. This is the only channel here that carries pictures and attachments both ways — QQ refuses them outright, and Feishu and Telegram carry them at a permission's mercy.

## What is genuinely not possible, and why

- **Group chats.** The claw bot channel has no group inbound at all. `WeixinMessage.group_id` exists in the protocol's shape and neither reference implementation ever reads it; both hardcode a direct chat. So `chatKind` is always `direct`, and the panel states that under the controls rather than leaving it in a collapsed fold — a user who binds it and then writes in a group would otherwise read the channel as broken.
- **Audio.** The connector seam carries text, images and files, and nothing downstream would transcribe a recording. WeChat transcribes voice *itself* and ships the text with most recordings, so the usual spoken message is answered rather than refused; a recording it could not transcribe reaches the chat as the shared not-supported notice. That is the same shape QQ's non-text messages take, so no channel-specific refusal was invented — it would say no more than the shared one already does.
- **Reply threading.** There is no quote relation on a send, so `replyText` resolves to the same operation `sendText` does (as on QQ). The bridge's threading choice is invisible here rather than wrong.

There is consequently **no `MessagingUnsupportedError` in this connector**, which is the honest answer rather than a gap: nothing outbound is refused.

## Delivery preferences

All three (`linePerMessage`, `finalReplyOnly`, `renderMarkdown`) apply unchanged, and none costs anything different here — there is no reply budget and no expiring reply window, so `WeChatBindingInfo` re-declares nothing. `renderMarkdown` does mean something new: WeChat reads Markdown *itself*, so `wechat-markdown.ts` subtracts rather than translates. Headings, bold, strikethrough, lists, quotes, rules, links, inline code, fenced code and tables go out as written — the widest subset of the four — while what the client will not show keeps its words and loses its markers (headings past the fourth level, emphasis around CJK, and inline images, which become links).

## Using the channel spec

Adding the channel was one `MessagingChannelSpec` entry plus the genuinely bespoke handlers, exactly as intended. Two of those handlers turned out to be shared rather than bespoke: the scan-result save (account guard, upsert, restart-an-enabled-connector) and the disable-before-scan guard are identical on both scanning channels, so they were lifted out of the QQ handlers into `saveScanResult` / `guardScanStart` and QQ now calls them too.

Two differences follow from having no typed credential, and neither fits the table: the PUT saves the delivery preferences alone and refuses to create a binding (400 `wechat_token_required`), and the credential test takes no request body at all.

## What the credential check proves, and what it does not

Fixed after testing against a real account: `Test connection` failed with `GetTypingTicket rpc failed` on a binding whose sends worked.

**No call on this platform can prove a credential without a side condition.** `getconfig` mints a typing ticket for a CONVERSATION, so a bot nobody has messaged yet cannot answer it; `getuploadurl` wants a peer and a file; `sendtyping` wants a ticket; `getupdates` is a long poll that would park for its window *and* advance the cursor the connector is about to read, so a probe using it would eat a user's message.

What every one of them does share is a single **authentication layer**, and that layer has no side conditions. So `WeChatApiError` now carries which of the two layers refused the request, and the probe reads only the first:

| outcome | verdict |
| --- | --- |
| transport error (DNS, TLS, refused) | **fail** — proves nothing either way |
| HTTP 401 / 403 | **fail** — the credential was refused |
| HTTP 200, `errcode`/`ret` = -14 (stale token) | **fail** — the one business code that *is* about the credential |
| HTTP 200, any other non-zero code (incl. `GetTypingTicket rpc failed`) | **pass** — the request authenticated; the call behind it needs state this binding has not got |
| HTTP 200, `ret` 0 | **pass** |

So the check now says **the stored token signs in**, and deliberately does **not** say the bot can deliver anything, name the bot, or confirm a conversation exists. It carries no account label, for the reason QQ's carries none. The user-visible consequence is that it passes on a freshly-scanned bot nobody has messaged — which is correct, because that bot is fine.

## Readiness no longer waits on a poll

`onReady` sat behind the first poll's payload, so enabling showed 连接中 for up to the long-poll window while messages already flowed. A poll is not an event — it is a request held open until one arrives — so it cannot be the readiness signal on an idle bot.

Each poll cycle now opens with the credential probe, which answers at once and proves what `connected` means here (host resolves, transport works, token authenticates), and re-runs ahead of every recovery so a token revoked mid-outage stops the loop with the right reason. Telegram deliberately proves readiness with `getUpdates` instead, and its reason does not apply here: its one-poller-per-token conflict is visible only on that call, and this platform has no such rule (a second connection on one bot is already refused a binding away, 409 `account_enabled_elsewhere`).

**Two further defects on the same path, found while fixing that:**

- A long-poll window closing with nothing to report **threw**, so the loop read every quiet minute as an outage and flapped the connection into `error`. It now resolves empty on an unchanged cursor — the ordinary quiet case, not a failure.
- The drain shared the long-poll window, so on a bot with no backlog it parked and returned the first message a user sent **after** enabling — then discarded it as backlog. It now asks only for what the platform is already holding, on its own short deadline.

## Verification

Run in a dedicated worktree on Node v24, in CI order:

- `pnpm -r build` — clean.
- `pnpm typecheck` — clean across all 8 packages.
- `pnpm test` — all green: server 1590, web 1548, core 1021, cli 383, desktop 175, landing 69, docs 53, skills 27.

New suites, faked at the depth each one is about:

- `messaging-wechat-transport.test.ts` (31) — the production adapter against a stubbed `globalThis.fetch`: the envelope and its headers, the cursor, inbound normalization (bot's own messages dropped, voice transcription read as text, quoted messages, both `aes_key` encodings), the send shapes, and the CDN media path — including a round trip proving the bytes on the wire are ciphertext that the key declared in the same request opens. The regressions live here: the probe passing on `GetTypingTicket rpc failed`, failing on 401/403 and on -14, failing on a transport error, and never reaching for `getupdates`; plus a quiet window resolving empty while a transport fault and a caller's close still throw. None of them waits on a real timer.
- `messaging-wechat-scan.test.ts` (26) — the protocol against a stubbed fetch (every status mapping, the pairing code on the query, the confirmed payload), the service's state machine over a fake transport (overlapping polls, the pairing code held for the next call and cleared once accepted, the IDC redirect, TTL sweep, per-Session eviction), and the routes, which assert the platform's handle appears in nothing the browser is given.
- `messaging-wechat.test.ts` (29) — the connector and routes over a fake transport whose long polls PARK, as production's do: readiness reported with every poll still parked and nothing returned, the drain asked for on the first call and no later one, a refused token stopping the connection with its own reason, an inbound message reaching the Agent and its reply coming back with the right context token, media both ways, an oversized transfer refused as a size problem, the outage and its recovery, and the token forgotten on close.
- `wechat-scan-connect.test.ts` (11), `messaging-wechat-scan.test.ts` (26), and additions to the two web form/editor suites and `messaging-markdown.test.ts`.

Not exercised against the live platform from this machine — no WeChat bot credentials here, so every test stops at the wire. The product owner's run against a real account is what surfaced the two status bugs above; delivery itself worked throughout.

Rebase: `origin/main` is still `0ec03277` (the commit this branched from), so there was nothing to rebase onto.


hiyouga:
## Review — reviewed in a worktree, through `84b93309`

A lot of this is right, and the parts that carry the most risk are the parts that hold up. The QQ refactor is behaviour-identical — I diffed both lifted helpers against `0ec03277` line by line, and `guardScanStart` is byte-for-byte the guard it replaced while `saveScanResult` performs the same four steps in the same order with the same arguments; the response body, the owner gating and the 404/502 paths are untouched, and all 413 tests across the eleven messaging suites pass. The scan handle appears in no response type, no error body and no log; task ids are 128 random bits; cross-Session polling, verify and cancel all 404. Both `aes_key` encodings round-trip against real `createCipheriv`, and a wrong key is distinguishable from an oversize transfer. The credential probe implements the table in the body exactly, and never touches `getupdates`.

Two things in the poll loop are wrong, and both are reachable without a hostile platform.

### 1. blocker — a drain that times out is spent anyway, and the next poll relays the whole backlog

`wechat-connector.ts:333` asks for the drain, `:341` spends it:

```js
const { messages, cursor: next } = await bot.getUpdates({ cursor, signal, ...(drained ? {} : { drain: true }) });
cursor = next;
if (!drained) { drained = true; continue; }
```

`wechat-api.ts:713` returns `{ messages: [], cursor }` with the cursor **unchanged** when the window closes — correct for a quiet long poll, and also what a drain that missed its 5s deadline returns. So on a binding that has been dark for a week: the drain times out, `cursor` is still `""`, `drained` is spent, and the next call is an ordinary poll from the beginning. The platform hands back the week, and every message of it becomes a queued Task — precisely the flood the module doc at `:37` says the drain exists to prevent.

Executed against the real connector with a client that mirrors the adapter's timeout contract:

```
polls: 3   drain flags: [true, false, false]   cursors: ["", "", "c-after-backlog"]
RELAYED TO THE AGENT: [ 'message from a week ago #1', 'message from a week ago #2' ]
```

Nothing in the suite catches it because the fake at `messaging-wechat.test.ts:128` returns `cursor: \`cursor-${this.polls}\`` — it *advances* the cursor on an empty drain, which the real adapter never does on that path. The transport test at `messaging-wechat-transport.test.ts:262` correctly pins the unchanged cursor. The two halves are each right and jointly wrong.

Spend the drain only when it returned something: `if (next !== cursor || messages.length > 0) drained = true;`, or have `getUpdates` report the timeout so the loop can retry it.

Related and unproven: the drain discards exactly one response. Telegram's equivalent (`telegram-connector.ts:541`) acknowledges the whole backlog in one step regardless of size. If `getupdates` pages its batches, a backlog larger than one page is partially replayed with no timeout involved.

### 2. blocker — `failures = 0` sits in the probe branch, so a poll-only failure never backs off

`wechat-connector.ts:327` resets the counter inside `if (!ready)`. When `checkCredentials` keeps answering while `getUpdates` keeps failing — an endpoint-specific 5xx, a rate limit, an unparseable body, or a cursor the platform has started rejecting (and `cursor` is a `poll()` local that is never reset on error, so a poisoned one is permanent) — each cycle probes green, zeroes the counter, fires `onReady`, then throws on the poll with `failures === 1` and fires `onError` again.

Executed, with `checkCredentials` always resolving and `getUpdates` always throwing:

```
probes: 93  polls: 93
onReady fired: 93   onError fired (= error records written): 93
backoff each attempt: [2000, 2000, 2000, …]  — never grew
```

That makes two of the code's own comments false. `:360` says "reported once per outage, not once per attempt: a token revoked overnight would otherwise write one error record per retry" — this writes one every two seconds, about 43,000 overnight, with a `connected`↔`error` flap at the same rate. `:113` says "doubling to a ceiling"; the backoff never leaves its first step. The existing test at `messaging-wechat.test.ts:599` only exercises a transient poll failure, so it passes either way.

`telegram-connector.ts:555` puts the reset **after** a successful `getUpdates` and says why in as many words. This connector says it mirrors telegram's lifecycle and then inverts that one line.

### 3. should-fix — two shared paragraphs in the API docs now contradict rows this PR added a screen above

- `server-api.en.md:350` (zh `:349`) — "**Video, audio and voice are deliberately not delivered** — nothing downstream decodes or transcribes them." The WeChat paragraph two lines below says the opposite for this channel, and the code does exactly that (`wechat-api.ts:484`, `:515`).
- `server-api.en.md:354` (zh `:353`) — this PR rewrote it to cover both scanning channels, and it still says a second concurrent poll of one task "reads as 404". On WeChat it answers `pending` (`wechat-scan.ts:369`), as the WeChat row at `:345` itself states.

Also missing from the shared prose, which enumerates the equivalent for the other three: WeChat's account identity (the bot id, `messaging.ts:846`) and WeChat's Markdown degradation.

### 4. should-fix — `already_bound` tells the user something this flow cannot know

`wechat-scan.ts:161` sends `local_token_list: []`, and its own comment says what that field is for: the platform takes the tokens this client already holds *so it can recognize a rebind of a bot it has already issued one for*. With an empty list the platform cannot distinguish "bound to this server" from "bound to any client anywhere", so `binded_redirect` cannot support "already connected to **this server**; there is nothing to authorize again" (`strings-en.ts:1975`, `strings.ts:1924`, and the docs row at `:345`).

That matters beyond wording. The editor offers `clearBotToken` (`messaging-binding-editor.tsx:934`). After a clear the row keeps its identity with an empty token, cannot be enabled, and the only way back is a scan — which, if `binded_redirect` is what the platform returns for any already-bound bot, reports `already_bound`, saves nothing, and tells the user everything is fine while the binding is unusable. Word it as "already bound, possibly elsewhere; nothing new was issued" and say what to do next.

### Smaller

- `strings-en.ts:2124` / `strings.ts:2068` — the insertion orphaned a JSDoc: `/** The QQ-only failure … */` now sits above `troubleWeChatDirect`, which brought its own, and `troubleQQPassive` below it is left undocumented. Both dictionaries. Also `strings.ts:1944` / `strings-en.ts:1996` still say per-channel strings live under "`feishu` / `telegram` / `qq`".
- `wechat-scan.ts:423` — `cancelSession` has no caller anywhere (`grep -rn cancelSession packages/server/src` finds only its two definitions, WeChat's and QQ's), a JSDoc naming a Session-delete cascade that does not exist, and a test pinning it. Against this repo's test-restraint value that is a method plus a test for a caller that is not there. The design spec does assert the QQ half of that behaviour, so the gap is real — wire both or drop both.
- `strings-en.ts:1932` — `S.wechat.botToken` has no reader (the section renders no credential field, and its own JSDoc concedes there is nowhere to type one), and where it is spelled `"Bot token"` diverges from Telegram's shipped `"Bot Token"` — and from both zh values.
- `wechat-markdown.ts:157` — table cells go out with only `|` escaped, so a cell containing `*x*` or a backtick opens a construct inside the table, and a link in a cell loses its URL entirely. The module doc's escaping rule is not what the table path does. Low impact; prose cells are the common case.

### Design docs

This one is a hard requirement rather than a nicety. `origin/main` of the design repo has no WeChat anywhere in `specs/`, and no branch carries it. The house pattern is explicit: design PR #70 was "messaging — QQ as a third channel and its QR scan (impl #490 … #507)", merged alongside that channel's code.

A fourth channel obliges, precisely:

- `05-ARCHITECTURE.md` §「消息接入（Messaging）」 — 「当前渠道为飞书（Lark）自建应用、Telegram 机器人与 QQ 机器人」, the `channel` column's enumerated values, `account_id`, `config_json`, the `render_markdown` subset, the `replyBudget` sentence, the inbound/outbound paragraphs (media both ways, voice as transcription, video as file, direct chats only), and the QQ scan-flow paragraph, which now describes one of two scanning channels with a pairing-code step QQ has no equivalent of.
- `05-ARCHITECTURE.md` §「API 原型」 — ten new routes beside the three QQ scan routes already listed.
- `06-PROTOTYPE.md` §「对话页」 and the Session-row binding dialog — both spell the selector as three tabs and describe QQ's scan block; there are four tabs now, one with no credential field and no console link, a pairing-code input, and a permanent direct-chats-only line.

Say the word and I will open it.

### Checked and clean

- Resource limits: the bridge's rolling per-binding image budget is channel-agnostic and WeChat honours it, refusing at the crossing byte over the ciphertext; files use the channel-neutral caps; the 60s transfer deadline matches Telegram.
- Persistence: `channel` is `TEXT` with no `CHECK`, so an existing database opens unchanged and the schema diff is comment-only. An older server skips a `wechat` row defensively. The token is masked on every projection, absent when cleared, never round-trips, and gets the same 409 clear ladder.
- The seam: `connector.ts` and `bridge.ts` changes are the discriminator plus doc text. `replyText` is unreachable on WeChat, so the malformed-ref throw cannot fire. No `replyBudget` means the channel-neutral 20-line cap. `onMessage` never rejects, so the cursor-advanced-before-delivery ordering cannot lose a message to a handler throw.
- Markdown: the shared `markdown.ts` diff is entirely its docblock, and the shared test is additive, so Telegram, Feishu and QQ output is untouched. CJK emphasis, H5/H6, images-as-links, nested lists, backtick-run fencing and fences containing markers all behave as documented.
- i18n: full structural comparison — 1570 key paths each side, none unique to either, and all 183 function entries interpolating the same arguments. Every CJK hit in the diff is in a permitted location. No Chinese in any `.tsx`, in `endpoints.ts`, in server source, in English strings or in test names. No `design § "…"` citations.
- All ten new routes are in both language versions of the API reference with matching methods, paths, error codes and ownership.
- The changelog pair mirrors section for section, twelve bullets each, metadata in fixed order with the PR link, no forbidden reasoning headings.
- `84b93309` is clean: `MESSAGING_ICON` still has its reader in the dock panel's tab, the row-menu icons stay distinct, and the changelog and test moved with it.
- Merged with #528, #529, #530, #532 and #534 on top of `origin/main`: build, typecheck, 1600 server / 1592 web / 1021 core tests all pass.

### Uncertain

- **What `binded_redirect` means with an empty `local_token_list`** — whether a bot this server has already bound can ever be re-scanned. That decides whether finding 4 is a copy fix or a functional dead end after `clearBotToken`. One scan of an already-bound bot on a real account settles it.
- **Whether `image_item.aeskey` is really hex.** `mediaRefOf` (`wechat-api.ts:452`) prefers it over `media.aes_key` and reads it as hex, and `Buffer.from(x, "hex")` truncates silently at the first invalid pair — so a base64 value there would make every inbound picture report "arrived but could not be decrypted". The test fixes the encoding by fiat. One real inbound image settles it, or the reference implementation's image branch.
- **The `strong` / `delete` asymmetry at `wechat-markdown.ts:99`.** The module doc says emphasis needs a word boundary, which CJK runs lack, but only `emphasis` strips its markers — `**中文**` and `~~中文~~` keep theirs. Either the renderer treats them differently or two of the three are wrong.
- `wechat-scan.ts:391` builds `https://${redirect_host}` from an upstream field with no shape check, and the handle is then sent there. Only reachable from a compromised or misbehaving `ilinkai.weixin.qq.com`, so I did not rank it, but a hostname allowlist would close it cheaply.

hiyouga:
Fixed in `acf93179`.

**Both blockers.** `getUpdates` now reports whether the window closed or the platform answered, and the drain is spent only on an answer — retried otherwise, bounded, since an idle bot's poll may park every time. The fake that hid this returned an *advancing* cursor on an empty drain, which the real adapter never does; it now mirrors production, which is what made the hole visible.

`failures = 0` moved out of the probe branch to after a poll that returned, as `telegram-connector` does. Pinned by a test that fails on the old placement with `[1,1,1,1]` where the counter should climb, and that also asserts the outage is one error record rather than one per attempt.

**The rest.**

- The two shared API-doc paragraphs that contradicted the WeChat rows: the blanket "video, audio and voice are not delivered" now names the exception and why (the platform does the decoding), and the "second concurrent poll reads as 404" claim is attributed to QQ, with WeChat's `pending` beside it. WeChat also joins the account-identity and Markdown-degradation lists. Both languages.
- `already_bound` no longer claims the bot is bound to *this* server. The scan sends an empty token list, so the platform cannot tell this server from anywhere else; the copy and the docs row say what it does know and what to do next.
- A table cell's literal markers are escaped like any other text. Worth noting the hazard is narrower than it first looks: a cell's own emphasis is already flattened by the parser, so what this closes is a lone `*` in one cell pairing with one in another.
- The orphaned JSDoc, the per-channel comment still naming three channels, and the unread `wechat.botToken` whose English casing diverged from Telegram's.

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` clean; 111 files / 1593 server tests and 112 files / 1548 web tests pass. (`task-attachments` and one other failed once under full parallel load and pass on a rerun — the known suite flake, unrelated to this branch.)

**Left for you, deliberately.**

`cancelSession` still has no caller. The design spec asserts the QQ half of that behaviour (「DELETE 顺带取消该 Session 在途的扫码任务」), so dropping the method would leave a spec-asserted behaviour unimplemented and wiring it is a change with its own shape — your call which way it goes.

The design specs are the bigger one, and a fourth channel obliges quite a lot of §「消息接入（Messaging）」, §「API 原型」 and §「对话页」. Say the word and I will open it.

Two of the uncertainties from the review still need a real account rather than a fix here: whether `binded_redirect` can ever be cleared for a bot this server has bound (which decides whether the `already_bound` copy is enough, or whether `clearBotToken` is a dead end), and whether `image_item.aeskey` is really hex.

PR #534, opened 2026-08-28
A new Project's default model can read a picture

hiyouga:
A new Project defaulted to `deepseek/deepseek-v4-flash`, which is **text-only** — so the first screenshot pasted into a fresh install was declined for a reason nothing on screen explains. The default is now `deepseek/deepseek-v4-flash-vision-exp`.

## Why this one costs nothing

The two catalog rows are otherwise identical: 1M context, CNY 0.05 / 1.5 / 4.5, same provider, same protocol. The vision revision is a strict superset, so the capability is not traded against price, context or speed. It is DeepSeek's own vision model, not a gateway's.

## Scope

**New Projects only.** A Project's default is copied in at creation and owned by it from then on; nothing rewrites an existing one, and "sync presets" explicitly never touches the stored default. An existing Project switches by picking the model, as it always could. That is the repo's standing convention for catalog changes, not a limitation introduced here.

## The test asserts the property, not the id

`state.test.ts` now checks that whatever `default_model` names resolves to a catalog entry with `supportsVision: true`, in addition to pinning the id. A future change of default cannot quietly land back on a model that declines the first image it is shown — which is exactly the bug being fixed, and the sort that reappears when someone swaps a default for an unrelated reason.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm test`, `pnpm format` + `pnpm format:check` all pass on Node 24 — 8 packages, 4761 tests, 0 failures.

Six assertions across `core/test/{agent,state}.test.ts` and `server/test/{auth,models}.test.ts` moved with the default; each was pinning the old id, and finding them is how you learn where a default is observable — the seeded admin's adopted `default_project`, a fresh Project's preset table, and a Session created with no model reference.


hiyouga:
## Review — reviewed in a worktree at `a7f76568`, and exercised against a running install

The one-line change is right and the two-row identity claim holds exactly. I diffed every field of both catalog entries: they differ in `modelId`, `displayName` and `supportsVision` and in nothing else — same provider, same 1M context, same three prices, and neither carries a `clientType` or `baseUrl`, so both auto-route identically. `ModelCatalogEntry` has no max-output or fast-mode field at all, so there is nothing hidden. Strict superset, confirmed.

I also ran the upgrade. A data root written by a pre-change build, then opened by this one:

| | default model |
| --- | --- |
| existing Project, root written by the old build | `deepseek-v4-flash` |
| Project created by the new build on that same root | `deepseek-v4-flash-vision-exp` |
| existing Project after a full model-table write | `deepseek-v4-flash` |

New Projects only, exactly as the body says.

### 1. blocker — the documented first-run command undoes the fix

`README.md:213`, `README.zh.md:213`, `quickstart-cli.{en,zh}.md:37` and `quickstart-sdk.{en,zh}.md:24` all carry the same line, byte for byte:

```
penguin config model add --provider deepseek --model-id deepseek-v4-flash --api-key sk-... --set-default
```

`--set-default` runs `project-config.ts:563` — `cfg.default_model = { provider, model_id: entry.model_id }` — so a user who follows the README pins the text-only model back. Run against the built CLI on a fresh data root:

```
$ penguin config model list
  deepseek  deepseek-v4-flash             vision=-
* deepseek  deepseek-v4-flash-vision-exp  vision=Y      <- the new default

$ penguin config model add --provider deepseek --model-id deepseek-v4-flash --api-key sk-... --set-default
Updated model (provider=deepseek, model_id=deepseek-v4-flash). Default model: (provider=deepseek, model_id=deepseek-v4-flash)

$ head -1 .project_config.toml
default_model = { provider = "deepseek", model_id = "deepseek-v4-flash" }
```

Their first pasted screenshot fails exactly as before. Before this PR the command was a no-op on the default, since the ids matched; this change is what turns it into a downgrade — and it is the one command every entry path tells a new user to run.

### 2. should-fix — the docs state the old default as fact, and one sample illustrates the bug

- `models.en.md:33` — "A fresh Project defaults to deepseek-v4-flash." `models.zh.md:33` says the same. That sentence introduces the `vision_model` proxy, so it now reads as if the default still needs one to see a picture.
- `models.{en,zh}.md:38` and `configuration.{en,zh}.md:76` — the sample TOML pins the old id, and `configuration.{en,zh}.md:80` gives its `[[models]]` row `vision = false`. The shipped sample therefore illustrates an out-of-the-box default that cannot take images.
- `models.{en,zh}.md:54` uses the DeepSeek family as the text-only exemplar, right under that sample.

Worth noting for the fix: changing only the `default_model` line yields a sample the server rejects, since the default must name an entry in `models` (`project-config-service.ts:388`). Both halves move together.

`.agents/skills/penguin-harness-dev/SKILL.md:97` names the bilingual `models` / `configuration` docs as what a catalog change touches.

### 3. should-fix — a shipped skill hardcodes the old id

`packages/skills/skills/penguin-sdk/SKILL.md:59`: "model selection still comes from the project config, whose preset default is `deepseek-v4-flash`". An agent that loads this skill will state the wrong id confidently. The surrounding claim about `DEEPSEEK_API_KEY` stays true; only the id is stale. Per the house convention the fix wants `version: 22` → `23` and a new `updated:` stamp.

### 4. should-fix — the changelog metadata

- **No `PR:` field.** `changelog/README.md:47` makes it required when one exists, and `:88` documents `grep 'PR:'` as the way to answer which PR shipped an entry. Ten prior entries I sampled all carry it.
- **`Scope: core` omits `model-catalog`.** `changelog/README.md:90` documents `grep -rl 'Scope:.*model-catalog'` as the catalog audit, and this entry is invisible to it. Every prior default/catalog entry tags it. Two of the six moved assertions are in `packages/server/test/`, so `server` belongs there too.
- The lead paragraph says the screenshot "was declined for a reason nothing on screen explains". It was not declined — it uploaded and was handed over as a scratchpad file path — and the reason was on screen twice: `chat-input.tsx:2148` renders `S.chat.imagesAsPathHint` whenever images are attached to a `vision = false` model, and `describe-image.ts:57` ends fatal naming `vision_model`. The real defect is narrower and still worth fixing: a fresh install has no vision model *and* no vision proxy, so there is no path to reading the image at all.
- The "why this one costs nothing" bullet is a price argument in present tense, which `changelog/README.md:60` and `:62` both route to the PR description. The scope bullet stays — `SKILL.md:99` mandates it.

### 5. Worth a sentence in the PR description

The repo recorded the opposite decision seven days ago, twice, in shipped release records: `changelog/0.2.4/2026-08-21-agenthub-0.4.6.md:15` and `changelog/0.2.4/RELEASE.md:29` both close with "A fresh Project still defaults to `deepseek-v4-flash`." Those folders are frozen and correctly untouched, but reversing a documented choice reads better acknowledged than silent.

Separately: nothing in the app treats an experimental revision differently. I grepped `packages/web/src`, `packages/core/src/state` and `packages/server/src` for stability markers and found none — no badge, no warning, no note. The only signal a user gets is the display name "DeepSeek V4 Flash Vision Exp" in the picker. That is a judgement call, not a defect, but it is what shipping `-exp` as the out-of-the-box default means.

### Design docs

`design/specs/06-PROTOTYPE.md:143` (on the design repo's `origin/main`) reads 「默认模型为 DeepSeek V4 Flash」. That is not merely imprecise — "DeepSeek V4 Flash" is still a distinct catalog row that is no longer the default. Nothing else in `specs/` moves: the `vision_model` / `describe_image` passages at `05-ARCHITECTURE.md:292` are all conditioned on the *session* model's flag, not on the Project default. Say the word and I will open the design PR.

### Checked and clean

- The test's new property assertion cannot pass vacuously. `supportsVision` is the real field and is required on `ModelCatalogEntry` — 0 of 107 rows omit it — and a bogus id resolves to `undefined`, so `expect(undefined).toBe(true)` fails. Verified by running the lookup against the built catalog.
- Every `default_model` write path: `writeInitialConfig` (creation only), `setDefaultModelRef` (explicit user PUT), `updateModels` (whole-table PUT, follows renames), `removeModel` (clears when it names the removed entry), snapshot import (Agent State only). `ensurePresetModels` does write a preset default, but only when the config has zero models, which no supported path can produce — worth one word in the changelog's "nothing rewrites an existing one", which is very slightly broader than the code.
- The vision interaction has exactly one path and no gap: a vision-capable default keeps `read_image` and filters out `describe_image` (`agent-state.ts:792`), `visionDescriber` is not injected (`agent.ts:845`), and `vision_model` correctly stays unset — coming back into play only when a user picks a text-only model.
- No fixture, golden file, installer script, workflow or Playwright spec pins the old id. The CLI tests pass it as an explicit argument, and `agent.test.ts:230` keeps the base id deliberately to prove a bare id is never resolved.
- The new code comment is pure English and cites no design doc. The changelog pair mirrors section for section with `## 细节` per the spec's heading table.
- Merged with #528, #529, #530, #532 and #533 on top of `origin/main`: build, typecheck, 1600 server / 1592 web / 1021 core tests all pass.

### Uncertain

`context-engine.ts:307` records that `deepseek-v4-flash` writes a summary body outside `[/summary]` (issue #170), which is why the tolerance ladder exists. The ladder is model-agnostic so I expect nothing to break, but the out-of-the-box compaction path now runs on a revision whose marker behaviour nobody has observed. A live compaction against the exp id would settle it.

hiyouga:
Fixed in `ced33c54`.

All six documented first-run commands now name the new default. Re-ran the check against the built CLI on a fresh data root: the config starts on `deepseek-v4-flash-vision-exp`, and the README's command no longer moves it.

- The `models` and `configuration` samples named the old id, and `configuration`'s `[[models]]` row carried `vision = false`. Both halves moved together, since the default must name an entry in `models`. The "e.g. the DeepSeek series" exemplar beside them now names the text-only sibling rather than the family the default belongs to.
- `penguin-sdk`'s SKILL.md hardcoded the old id; updated with a version bump.
- The changelog gained its `PR:` field, `model-catalog` plus the packages it actually touches on `Scope`, and a lead paragraph without the false premise. The image was never declined — it was saved as a scratchpad path and the reason was on screen twice. The real defect is narrower: a fresh install had no vision model *and* no vision proxy, so nothing could read it.

`pnpm -r build`, `pnpm typecheck`, `pnpm format:check` clean; 1021 core tests pass.

Two things left for you. The 0.2.4 release records say "A fresh Project still defaults to `deepseek-v4-flash`" twice, so this reverses a documented choice — worth a sentence in the PR description rather than leaving it silent. And `design/specs/06-PROTOTYPE.md:143` still reads 「默认模型为 DeepSeek V4 Flash」; say the word and I will open the design PR.

PR #535, opened 2026-08-28
feat(db): the machines tables as migration 3 (after #524, #448)

Myriad-Dreamin:
Stacked on #448 and #524. Not to be merged on its own.

#448 adds three tables to `schema.ts` (`machine`, `machines`, `machine_project`) through the declarative `CREATE TABLE IF NOT EXISTS` track only. #524 introduces ordered, stamped migrations. This branch is the meeting point: migration **3 `machines`**, frozen DDL, `swapSafe: true` (purely additive), with a `down` that drops the three tables and names what that loses.

Order of landing:

1. #524 merges.
2. #448 merges.
3. Rebase this branch onto `main`; it should reduce to one commit touching `db/migrations.ts` and `test/db-migrations.test.ts`. If #524's version numbering moved by then, renumber — the version must stay contiguous.

This draft exists so the migration is not forgotten between the two.


Myriad-Dreamin:
Folded into #448 — that branch is now stacked on #524 and carries migration 3 itself (`b0f234a0`). A hot push onto an existing database needed it before #448 could be tested at all: `SCHEMA_SQL` only runs in `openDatabase`, which is the runtime's, so a pushed platform found no `machine` table and failed to boot. #524's `migrate(db, { swapPath: true })` on the platform path is what fixes it.

PR #539, opened 2026-08-28
feat(core): per-rotation context assembly under a three-tier rule — strict request prefix, soft thinking level, live permissions and policy

hiyouga:
## Summary

Until now the post-compaction rebuild reused the runtime the Session was created with: the same system prompt, the same toolset, the same vault, the same compaction settings. An `AGENTS.md` edit, an installed Skill, an added MCP Server or a new vault key made during a Session only took effect in the next Session — and the spec said as much ("system prompt 与工具定义不属于被压缩的历史，新对象原样保留").

This PR makes the **whole runtime configuration a per-model-context artifact**, assembled from the Agent State on disk at the moment the context opens — exactly as a new Session's first context is — under a three-tier rule for runtime parameters: **the request prefix of a Trace file never changes while it runs, and per-tool permissions and the command policy are read once per context (strict tier); the thinking level may move mid-context with a compact-first reminder (soft tier); the approval mode applies per decision (unrestricted tier).**

- **Three openers, one procedure** (`Agent.assembleContext`): `createSession`, the context a completed compaction opens (summarize and discard alike), and a resume that finds its latest Trace file closed by a completed compaction. Each reads `system_config.yaml` in full (the prompt template with its section prompts and toggles, the builtin tool entries and MCP Servers, `compaction`, `max_turns`, `model.*` defaults), `AGENTS.md`, the vault, the installed Skills' metadata, the Memory indexes, the schedule roster and the date. Nothing under the Agent State is fixed per Session any more; nothing changes inside a running context.
- **Runtime parameters form three explicit tiers.** *Strict* — the system prompt, the toolset (MCP included), the compaction settings and the model reference are the request prefix: byte-fixed across a whole Trace file, changed only at rotation (the vault rotates on the same schedule though it never reaches the model). *Soft* — the thinking level: a context opens at the Session's pinned level (Web picker, CLI `--thinking` / `/thinking`, SDK: assigning `Session.thinkingLevel`) or the Agent config's base read at context open — a pure per-request parameter, recorded nowhere — and a re-pin applies from the very next LLM request, mid-context — the pin becomes engine-owned state (`ContextEngine.setThinkingLevel`), applied to every subsequent turn request (compaction requests keep the context's own level, their prefix must not move) — and every change point reminds the user beforehand that compacting first is recommended, since the change invalidates the provider's cached context (the web picker's menu note, the CLI `/thinking` reply). *Unrestricted* — the approval mode alone, re-read from the DB per decision. A tool's `r`/`rw` and the command policy are strict-tier: `Session.toolPermission` answers from the running context's toolset (rebuilt at rotation, MCP `permission` overrides included) and the command policy is read from `.project_config.toml` once per context open — no approval decision reads a file. `RunOptions.thinkingLevel`, the subagent seams' per-round level, `TaskCreateRequest.thinkingLevel`, the subagent message's level and the follow-up recall's level are gone; `PATCH /sessions { thinkingLevel }` pins the loaded runtime and the loader restores the row's pin at load.
- **Session-fixed** (`SessionSpec`): the Session id, Workspace, model entry (credentials, window, per-model annotations), origin, subagent depth, the Project command-policy snapshot, and the Environment's process host — background commands, child sessions and their listeners survive the rotation.
- **Environment re-equipped** (`Environment.reconfigure`): the new toolset replaces the old; the vault's values go straight into the command environment of every command spawned from then on (running processes keep the environment they started with). **MCP connections are cached by config** (`McpToolProvider.reconfigure`): an entry whose resolved config is unchanged keeps its live connection and discovered tools — a Session that compacts every turn does not respawn its servers — a removed or changed one is closed, and only new, changed or previously failed servers connect, bracketed by the same `mcp_connect_begin` / `mcp_connect_end` pair the first run streams (naming just those servers), then the new `tool_list_ready`. The engine yields these live (a merge queue over the opener's `emit` callback).
- **Recorded**: the rotated Trace file opens with the context's `session_meta`, then its connect pair (if any) and its `tool_list_ready`; `Session.metaMessage` follows the running context. The engine adopts the opened context's `maxTurns` and compaction settings.
- **No silent fallback**: an Agent State that cannot be assembled (a `system_config.yaml` that no longer parses) fails the run with that error; the engine stays on the old context with no rotation pending, so the next trigger compacts again from a consistent state. It is the same error a new Session would hit.
- **One timing for Web and CLI**: the server no longer rebuilds an Agent's cached runtimes on a vault update (that re-resume gave the Web "next task" semantics the CLI never had). Agent State changes reach a running Session at its next compaction and a new Session immediately — and every entry point says so: the Agent settings pages' save toasts (prompt, runtime, tools/MCP, memory, schedules, vault, skills), the skills library toasts, `penguin config vault set/remove`, the in-chat thinking picker's footnote and `/thinking`.
- **Open-context resume** keeps the prompt and opening level its file recorded (the host's pin, restored at load, keeps riding requests); tools, Environment, vault and run settings come from the current Agent State as before — the Trace records no executable configuration.
- `createSession` / `resumeSession` load the Agent State from disk (new `loadAgentState`, which `loadOrInitAgentState` now builds on), so a long-lived Agent object — a self-spawned subagent's — no longer starts Sessions on its load-time snapshot.

- **One loader, one opener.** Agent State loading is a single function: `loadAgentState` — with `init: {}` (or `init: { preset }`) it is the create-or-load entry (`createAgent`, project provisioning; `loadOrInitAgentState` is gone), without `init` a missing Agent throws, which is exactly what a context opening mid-session must see. And opening a context is a single procedure: the first run's bootstrap and the post-compaction `openNextContext` share one implementation (`openAssembled`: connect what is pending, publish the connect pair and the toolset record through `emit`, build the LLM), pumped live by the same merge queue on both paths — so `SessionConfig.bootstrap` now takes `{ emit }` and returns `{ tools, llm }`, and `SessionConfig.mcpServers` is gone (the bootstrap knows what it is connecting). `createSession` and `resumeSession` also converge on one Session assembly (`newSession`).

### SDK seam

`SessionConfig.createLLM` / `ContextEngineDeps.createLLM` → `openNextContext({ emit }) => OpenedContext | Promise<OpenedContext>`, `OpenedContext = { llm, sessionMeta?, maxTurns?, compaction? }`. Records passed to `emit` are yielded on the run stream and written at the rotated file's head after the meta, and the engine authors `token_usage.session` itself — the LLM reports per-request usage only; every optional field absent means "the previous context's stays". `SessionConfig.bootstrap` speaks the same `{ emit }` shape and returns `{ tools, llm }` (`mcpServers` is gone from SessionConfig); `loadOrInitAgentState` is folded into `loadAgentState({ init })`. Assigning `Session.thinkingLevel` is the host's way to change the level — plain state forwarded to `ContextEngine.setThinkingLevel`, live from the next request; the pin functions are gone; `Session.toolPermission` stays sync, answering from the running context's toolset; `SessionConfig.commandPolicy` is a source answering with the running context's policy. The engine stays feature-free — it never sees a prompt, a toolset, a vault or a level, only "an LLM, the meta describing it, its settings, and the records it published". Test fakes return `{ llm }` where they returned the LLM.

### Notes

- No on-disk format change and nothing to migrate: `session_meta` gains `thinking_level` (a file written before it exists resolves the level as a new context would); `session_meta.system_prompt` has always meant "the prompt this file's context ran with" and now differs between files of one Session when the Agent State changed in between; a rotated file's head carries the connect pair before `tool_list_ready`. Readers already do "newest wins" (`context-breakdown`, the web stream model).
- Design spec counterpart: penguin-harness-design [#84](https://github.com/Prism-Shadow/penguin-harness-design/pull/84) — 「上下文轮换与运行配置」 in 05 (the prefix invariant, MCP caching, the reminder rule), the Vault 生效语义, PRN-001, Trace 分文件, 恢复流程, the API prototype, 06's picker semantics.

## Verification

- `pnpm build` (all packages), `pnpm typecheck` — clean
- `pnpm --filter @prismshadow/penguin-core test` — 46 files, 1043 tests passed (1 file skipped: the LLM e2e)
- `pnpm --filter @prismshadow/penguin-server test` — 111 files, 1610 tests passed
- `pnpm --filter @prismshadow/penguin-web test` — 115 files, 1607 tests passed
- `pnpm --filter @prismshadow/penguin-cli test` — 29 files, 383 tests passed
- `pnpm format` + `pnpm format:check`, `git diff --check` — clean
- New tests: engine — records published by the opener are yielded live (the begin reaches the consumer before the end exists) and written at the rotated head, a context that publishes none keeps the previous records, an opened context's `maxTurns` / compaction settings take effect, an opener that throws fails the run and leaves the engine on the old context with no rotation pending; Agent — `createSession` assembles from disk, the post-compaction context is rebuilt whole (prompt, toolset, vault into the Environment, thinking default, `max_turns`, compaction mode) and records it, `session_meta` records no level, `Session.thinkingLevel` is engine state riding requests while contexts keep their base, MCP Servers of the new context connect with the event pair, an unreadable `system_config.yaml` fails the open; Environment — `reconfigure` replaces tools and vault (a command spawned afterwards sees the new values), MCP reconfigure keeps an unchanged server's connection (no connect phase, no outcome), reconnects a changed one, closes a removed one; resume — closed context rebuilt, open context keeps the recorded prompt and level (a pin before the first run does not re-stamp it; a legacy recorded level is ignored); server — `startTask` carries no level and `setThinkingLevel` assigns the loaded runtime's state, a vault PUT leaves cached runtimes alone; CLI — `--thinking` / `/thinking` / caller inheritance / `/clear` all pin through PATCH and task bodies carry no level.

https://claude.ai/code/session_01JNzs9pcQ1yUQAuhVTMp9N4


packages/core/src/engine/context-engine.ts
```diff
@@ -1373,13 +1388,23 @@ export class ContextEngine {
...
   private observeTokenUsage(msg: OmniMessage): boolean {
     if (msg.type !== "event_msg") return false;
     const payload = msg.payload as Partial<TokenUsagePayload>;
     if (payload.type !== "token_usage") return false;
-    if (payload.request) this.lastRequestTotal = payload.request.total;
-    if (payload.session) this.lastSessionTokens = payload.session;
+    if (payload.request) {
+      this.lastRequestTotal = payload.request.total;
+      this.lastSessionTokens = addTokenCounts(this.lastSessionTokens, payload.request);
+      (payload as TokenUsagePayload).session = this.lastSessionTokens;
```
hiyouga, line 1471: **Bug (fixed in 5faee270):** the Session series was stamped here *after* the message had already been pushed to the merge queue and awaited-written to the Trace (`runTurn`: `queue.push(msg); await this.write(msg); observeTokenUsage(msg)` — same order on the compaction path). `GenerativeModel` now yields request counts in the `session` slot, so every on-disk `token_usage` (and whatever a consumer serialized at yield time — the server's SSE publish, `GET /messages`) carried `session == request`. Concretely: three turns of 100/60/40 record `session.total` 100, 60, 40 instead of 100, 160, 200, and `trace/resume.ts lastSessionTokens()` reseeds the resumed engine with the *last request's* counts — the cumulative series collapsed on every resume. The PR's tests passed because they hold the object reference the later in-place mutation fixes up.

Fix: stamp before push/write on both paths, plus an engine test that snapshots at yield time and reads the Trace back from disk (a seam fake hides exactly this class of bug).

packages/core/src/engine/context-engine.ts
```diff
@@ -1749,15 +1774,48 @@ export class ContextEngine {
...
+      } finally {
+        queue.removeProducer();
+      }
+    })();
+    // A rejection is re-thrown by the await below; this branch only keeps it from being
+    // reported as unhandled while the queue is still being drained.
+    void opening.catch(() => {});
+    const records: OmniMessage[] = [];
+    for (;;) {
+      const msg = await queue.next();
```
hiyouga, line 1859: **Not fixed — design call.** Two related gaps around this pump, both new with this PR (the old `createLLM` swap was synchronous):

1. It never consults the run's `AbortSignal` and nothing wires `Environment.cancelMcpConnect` into a rotation (only the first-run `ensureReady` pump races the signal and cancels). A post-compaction context with a new/changed/previously-failed MCP server that is unreachable holds Stop hostage for up to `connectTimeoutMs` per server, at every compaction.
2. The rotation isn't transactional: `openNextContext` in agent.ts commits `current = next` and `environment.reconfigure(...)` *before* `openAssembled` finishes (see the note on agent.ts:1086), and this generator has no `try/finally`, so a consumer that stops driving it mid-open leaves the Environment on the new toolset/vault/policy while the engine keeps the old LLM/meta.

In practice (2) only bites if the open throws *after* the reconfigure — `listTools` doesn't throw for failed servers and the credentials were validated at creation — so the realistic exposure is (1). If you take it on: thread `signal` into `startNewContext`, race it in this loop (`raceAbort`, as `ensureReady` does), and give the engine a `cancelOpenNextContext` dep wired to the same `cancelMcpConnect`; on abort, the opener's rejection should be converted into the run's abort outcome, not an error. Since the vault/toolset swap has already happened at that point, that's the moment to decide whether the open must become commit-at-the-end (assemble → connect → then swap `current`/Environment/engine together). I left it as a comment rather than guessing the contract.

packages/core/src/engine/context-engine.ts
```diff
@@ -1507,7 +1532,7 @@ export class ContextEngine {
         if (summaryText !== "" && attempt.toolCalls.length === 0) {
           const summary = userText(buildContextSummaryText(summaryText));
           yield* this.emitCompactionEnd(reason, "summarize", "completed", { attempt: attempts });
-          await this.startNewContext();
+          yield* this.startNewContext();
```
hiyouga, line 1600: **Not fixed — worth a decision.** `compaction_end(status: completed)` is written and durable, then `startNewContext` runs — which can now throw (the opener re-reads the Agent State). If it does, the process keeps the full old context ("stays on the old context with no rotation pending", per the comment) and continues writing to the same file *after* a completed `compaction_end`, while a restart-resume of that file takes the context-closed path. Live process and disk disagree from that point on. Pre-PR the rebuild couldn't fail here. Options: assemble the new context *before* emitting the completed marker (the summary is already in hand), or mark the rotation pending on failure so the next successful open rotates. Either way the invariant deserves a test.

hiyouga (commented):
Review of the PR at edba2fd6, with fixes pushed on top as 5faee270 (after a merge of `main` — `3a765d30` — which resolved the conflicts with #571's retry ladder in `context-engine.ts` and the `timeoutMs` rows in the configuration docs).

**Fixed in 5faee270** — see the inline notes for the mechanism of each:
- `token_usage.session` stamped after the message was already streamed and written (Trace and wire carried per-request counts; resume reseeded from them).
- The Project half of the thinking-level chain read from the Agent object's snapshot instead of the config the context is assembled from.
- `PATCH { thinkingLevel }` lost when it races a runtime load.
- The subagent panel's level pick never reaching a live child (new host seam through `SubagentHandle.setThinkingLevel`).
- `Environment.reconfigure`: the throwaway `McpToolProvider` from reusing `equip()`, and an omitted `vault` silently clearing the vault (now required).
- The opener pump duplicated in `Session.ensureReady` and `ContextEngine.startNewContext` (`pumpOpener`).
- Dead surface: `bootstrap`'s returned `tools`, `THINKING_LEVEL_NAMES` with zero consumers (the server's two copies now import it), `Environment.mcpServerNames` / `McpToolProvider.serverNames` (their only caller left with `SessionConfig.mcpServers`), `McpConnection.tools` (kept whole tool arrays alive to seed a count that was overwritten), `AssembledContext.state` (unread).
- Stale docs: `launchTask` / goal-path comments, `ManagedSubagentSession.startRun`, the `SessionMetaPayload` JSDoc, the changelog's "prompt and level", the CLI's "subagents follow it", `SUBAGENT_THINKING_LEVELS`' JSDoc reattached.

**Left for you to decide** (inline): the rotation's un-abortable MCP connect and its non-transactional swap (context-engine.ts:1804, agent.ts:1086), the durable `compaction_end` before an open that can now fail (context-engine.ts:1535), whether a live pin should be inherited by `run_subagent` children (agent.ts:838), the thinking-level mirror vs a getter source (session.ts:841), and the fire-and-forget close of replaced MCP servers (provider.ts:236).

Not raised inline, optional follow-ups: `assembleContext` awaits six independent disk loads sequentially and `listInstalledSkills` reads every Skill's `icon.svg` on each context open (only name/description reach the prompt) — this chain now runs at every rotation, not just at creation.

Verification on 5faee270: `pnpm -r build`, `pnpm typecheck`, `pnpm format:check`, `git diff --check` clean; core 46 files / 1048 tests, server 113 / 1633, cli 29 / 383, web 116 / 1622, docs 7 / 53 — all passing.

packages/core/src/agent.ts
```diff
@@ -918,68 +1012,88 @@ export class Agent {
...
+    // as it is now — an edit the model (or the user) made during the old context to
+    // AGENTS.md, system_config.yaml, the vault, the Skills or the MCP Servers lands here.
+    // The Environment is re-equipped with the new toolset and vault, then the context is
+    // opened by the same procedure as the first one — the engine yields the published
+    // records live and writes them at the head of the rotated Trace file. An Agent State
+    // that cannot be assembled (a config that no longer parses) throws: the run fails with
+    // that error and the engine keeps the old context.
+    const openNextContext = async ({ emit }: OpenContextOptions): Promise<OpenedContext> => {
+      const next = await this.assembleContext(spec);
+      current = next;
```
hiyouga, line 1089: Side effects before the open commits: `current = next` and the Environment re-equip happen here, `openAssembled` may still reject (cancel, or `buildLLM`) and the engine then keeps the old LLM while `commandPolicy()` / `toolPermission` / the vault already answer for the new context — the "engine keeps the old context" invariant in the comment above only holds for `assembleContext` failures. See the note on context-engine.ts's rotation pump for the abort side of the same issue. Not changed: the realistic throwing paths after this line are narrow, and the right fix depends on how you want rotation aborts to behave.

packages/core/src/agent.ts
```diff
@@ -711,14 +825,17 @@ export class Agent {
...
+        // none, so the child never falls back to its own Agent config (which would apply on
+        // a cross-agent spawn). A mid-session `Session.thinkingLevel` assignment is engine
+        // state and is not inherited — the spawn argument exists for explicit control.
+        const parentLevel =
+          spec.thinkingLevel === undefined ? current.thinkingLevel : spec.thinkingLevel;
         const childSession = await childAgent.createSession({
           workspaceDir,
           ...childModel,
-          thinkingLevel: spawnThinkingLevel ?? thinkingLevel ?? null,
+          thinkingLevel: spawnThinkingLevel ?? parentLevel ?? null,
```
hiyouga, line 840: **Partially fixed — a design question remains.** With the CLI's `--thinking` now a PATCH, the pin lands only as engine state (`Session.thinkingLevel`) and never in `spec.thinkingLevel`, so a `run_subagent` child spawned without an explicit level runs at the *Agent-config* level — not the parent's pin — while the same row pin, replayed by the server's self-heal path (`createSession({ thinkingLevel: row.thinkingLevel })`), *is* inherited. And `SessionCreateRequest.thinkingLevel` still exists, so the API exposes both flavours. The comment here documents non-inheritance as deliberate, so in 5faee270 I only corrected the CLI comment (and test name) that claimed "spawned subagent sessions follow it". If children *should* follow the live pin (`penguin chat --thinking max` used to give the whole tree max), the cleanest shape is one Session-level accessor (live pin, else the context base) read both here and at context assembly — your call.

packages/core/src/agent.ts
```diff
@@ -190,19 +295,165 @@ export class Agent {
...
+   * `effectiveThinkingLevel`): the picker shows this effective value, and a pick writes
+   * through to the AGENT config — the project default is only ever a fallback, never
+   * overwritten from there.
    */
-  private configuredThinkingLevel(): ThinkingLevelName {
+  private configuredThinkingLevel(state: AgentState): ThinkingLevelName {
     return (
-      this.state.systemConfig.model?.thinking_level ??
+      state.systemConfig.model?.thinking_level ??
       this.projectConfig.default_chat?.thinking_level ??
```
hiyouga: **Fixed in 5faee270:** `configuredThinkingLevel` read the Project half of the chain from `this.projectConfig` — the Agent object's load-time snapshot — while `assembleContext` re-reads `.project_config.toml` from disk two lines later for `command_policy`. A loaded server runtime (or a CLI chat) therefore opened new contexts at the *old* `default_chat.thinking_level` until idle eviction, while a `command_policy` edit in the same file landed at that same rotation. Now the freshly loaded config feeds both; `agent.test.ts` covers the same Agent object seeing the file change.

packages/core/src/environment/environment.ts
```diff
@@ -184,24 +186,66 @@ export class Environment implements EnvironmentInterface {
...
+    if (this.mcp && servers.length > 0) {
+      this.mcp.reconfigure(servers);
+    } else {
+      this.mcp?.closeQuietly();
+      this.mcp =
+        servers.length > 0
+          ? new McpToolProvider(servers, { workspaceDir: this.workspaceDir })
+          : null;
+    }
+    const mcp = this.mcp;
```
hiyouga: **Fixed in 5faee270** (two things on this block):

- `reconfigure` reused `equip()` by letting it construct a brand-new `McpToolProvider` that the save/clobber/restore of `this.mcp` immediately discarded — `resolveMCPServers` ran twice per rotation and reordering these three lines would silently drop every cached MCP connection with no type error. `equip()` now assembles the builtin half only; `newMcpProvider()` builds the MCP half at construction, `reconfigure` reconciles the live one.
- `vault?` with `config.vault ?? {}`: omitting the field silently *cleared* the vault (the test even asserted it). The in-repo caller always passes it, so the type now requires it — "replaced as a whole" is the documented contract, and the caller says so explicitly.

packages/core/src/environment/mcp/provider.ts
```diff
@@ -184,22 +188,59 @@ export class McpToolProvider {
...
+    const resolved = resolveMCPServers(entries);
+    this.servers = resolved.servers;
+    this.configWarnings = resolved.warnings;
+    // The resolved server is plain data produced by one resolver, so its JSON is a stable
+    // identity for "the same entry".
+    const wanted = new Map(this.servers.map((s) => [s.name, JSON.stringify(s)]));
+    const kept: McpConnection[] = [];
+    for (const conn of this.connections) {
+      if (wanted.get(conn.server.name) === JSON.stringify(conn.server)) kept.push(conn);
+      else void conn.client.close().catch(() => {});
```
hiyouga, line 230: Low, not fixed: a changed/removed server is closed fire-and-forget (`void conn.client.close()`), and the replacement connect that `openAssembled` triggers right after can spawn while the old stdio process is still exiting. A server that holds an exclusive resource (a port, a SQLite lock) then fails its reconnect with EADDRINUSE / "database is locked", is recorded `fatal`, and the new context runs without it until the next rotation. Cheap hardening if you want it: collect the close promises here and have the next connect phase await them first.

packages/core/src/interfaces/environment.ts
```diff
@@ -144,6 +142,16 @@ export interface SubagentHandle {
  * argument inherits the parent Session's effective level, so a parent genuinely running
  * without a level still passes that down.
  */
+/** Every thinking level name, the `ThinkingLevelName` vocabulary as a runtime list (validating a recorded or wire value). */
+export const THINKING_LEVEL_NAMES: readonly ThinkingLevelName[] = [
```
hiyouga, line 140: **Fixed in 5faee270:** `THINKING_LEVEL_NAMES` had zero consumers while the server kept two hand-written copies of the same six names (`routes/sessions.ts`, `services/agent-config-service.ts`) doing exactly the wire validation this comment promises — both now import it. Also: the JSDoc above this const belonged to `SUBAGENT_THINKING_LEVELS` (the new export was inserted between a comment and its declaration); reattached.

packages/core/src/session.ts
```diff
@@ -483,77 +503,83 @@ export class Session {
...
-      // surfacing as an unhandled rejection.
-      work.catch(() => {});
-      // Stashed (not written here) so the caller can put the run's INPUT first — the
-      // Trace stays chronological: input, pair, abort.
-      this.abortedBootstrapRecords = records.length > 0 ? [...records] : [];
-      if (this.mcpServers.length > 0) {
-        const end = mcpConnectEnd({ status: "aborted", results: [] });
-        yield end;
-        this.abortedBootstrapRecords.push(end);
+    const queue = new MergeQueue();
```
hiyouga: **Fixed in 5faee270:** this MergeQueue opener pump (addProducer / try-finally removeProducer IIFE / `void work.catch` / drain) was copy-pasted in `ContextEngine.startNewContext`, comment included, and the copies had already diverged (this one honours the abort signal, the rotation one doesn't). Both now go through `pumpOpener()` beside `MergeQueue` in `internal/merge-queue.ts`; each site keeps its own drain loop, so the abort semantics stay where they are.

packages/core/src/session.ts
```diff
@@ -44,57 +37,70 @@ import type {
...
+   * `tool_list_ready` carrying the toolset — are published through `opts.emit` and stream
+   * live from `run`: the same shape as `openNextContext`, because opening the first context
+   * and opening a later one are one procedure in the composition layer. Kept out of
+   * Session construction so creating a Session is instant and the connect wait streams as
+   * visible events instead. An aborted attempt is cancelled via `cancelBootstrap` — the
+   * next run calls this again and reconnects from scratch.
    */
-  bootstrap: () => Promise<{
+  bootstrap: (opts: OpenContextOptions) => Promise<{
     tools: ToolDefinition[];
```
hiyouga: **Fixed in 5faee270:** the returned `tools` was read by nobody — this destructures `{ llm }`, `openNextContext` does the same, and the resume wrapper only touches `r.llm`; the toolset already travels as the emitted `tool_list_ready` record. A dead field in a new SDK contract is the kind of thing that never gets removed later, so `bootstrap` now returns `{ llm }` (13 test fakes and the changelog updated).

packages/core/src/session.ts
```diff
@@ -795,12 +822,37 @@ export class Session {
...
+   * compaction requests and as the fallback while nothing was ever assigned. The cost of
+   * the softness is the provider's cached context — changing the level invalidates the
+   * cached messages — so a host lets the user know at the picker that compacting first is
+   * recommended (the Web menu note, the CLI `/thinking` reply).
+   * Callers: the Web App's in-chat picker (PATCH), the CLI's `--thinking` / `/thinking`.
+   */
+  get thinkingLevel(): ThinkingLevelName | undefined {
+    return this.level;
+  }
+  set thinkingLevel(level: ThinkingLevelName) {
```
hiyouga, line 827: Suggestion, not applied: the level lives in two fields (`Session.level` mirrored into `ContextEngine.thinkingLevel` via `setThinkingLevel`, plus the lazy handoff in `ensureReady` because the engine may not exist yet), while the same PR gives the engine `commandPolicy` as a *source* it reads at the consumption point. A `thinkingLevel: () => this.level` dep read at the one request-assembly site would delete the setter forwarding, the engine field and the handoff — and would double as the accessor a child spawn could read (see the agent.ts:838 note). Left alone because the PR text and docs name `ContextEngine.setThinkingLevel` explicitly.

packages/server/src/http/routes/sessions.ts
```diff
@@ -599,9 +591,11 @@ export function sessionsRoutes(deps: AppDeps): Hono<AppEnv> {
       updated = { ...updated, approvalMode };
     }
     if (thinkingLevel !== undefined) {
-      // Takes effect from the next run on: launchTask/startGoal read the pinned level for
-      // any run that carries none of its own (a run already in flight keeps its level).
+      // The row is what the loader applies at load; a runtime already loaded is assigned
+      // here. Soft-limited: core applies it from the Session's very next LLM request (the
+      // picker advises compacting first — a change invalidates the provider's cached context).
       deps.sessionsRepo.updateThinkingLevel(row.sessionId, thinkingLevel);
+      deps.manager.setThinkingLevel(row.sessionId, thinkingLevel);
```
hiyouga, line 593: **Fixed in 5faee270:** this races a concurrent runtime load. `ensureEntry` snapshots the row, then awaits `loader.load(row)`; a PATCH in that window updates the row and calls `manager.setThinkingLevel`, which no-ops because the entry is inserted only after the load resolves — and the loader applied the *pre-PATCH* row. The pin silently never reached the runtime while `SessionInfo.thinkingLevel` showed it (first task after idle eviction + a picker click is exactly this). `ensureEntry` now re-reads the row's pin after the load; test added.

packages/server/src/runtime/session-manager.ts
```diff
@@ -602,12 +614,24 @@ export class SessionManager {
...
+   * Sets a Session's thinking level on its loaded runtime (core `Session.thinkingLevel`,
+   * plain state): the row's value is what the loader applies at load, so this only needs to
+   * reach a runtime that is already in the active table. The level is soft-limited: it
+   * applies from the Session's very next LLM request (mid-context — the UI advises
+   * compacting first, since the change invalidates the provider's cached context). No-op
+   * when nothing is loaded.
+   */
+  setThinkingLevel(sessionId: string, level: ThinkingLevelName): void {
+    const session = this.entries.get(sessionId)?.session;
+    if (session) session.thinkingLevel = level;
```
hiyouga: **Regression, fixed in 5faee270:** the web subagent panel's picker was replaced by `pinChildLevel` → `PATCH /sessions/:child`, but this only looks in `entries` — child sessions get an index row (`registerChildSession`) yet never become entries, they live inside the parent's core Session. So the PATCH returned 200, the row updated, and the live child kept its spawn-time level while the UI showed the pick (pre-PR the per-round level rode `messageSubagent` and took effect). Now: `SubagentHandle.setThinkingLevel?` → `ManagedSubagentSession` → `Environment.setBackgroundSubagentThinkingLevel` → `Session.setBackgroundSubagentThinkingLevel`, and this method falls back to the loaded parents' live children. Tests on both sides.

packages/server/src/runtime/session-manager.ts
```diff
@@ -1043,19 +1058,7 @@ export class SessionManager {
    * a queued follow-up that carried no level of its own picks up the pin as it stands when
```
hiyouga: **Fixed in 5faee270:** two stacked doc comments described removed behaviour — the one-liner still said `launchTask` "drives the run with the per-turn thinking level", and this block is the deleted `runThinkingLevel()`'s orphaned fallback-chain doc. Same for the goal path's "Same fallback chain as a task: the goal's own level…" comment (goals carry no level any more).

packages/core/src/environment/tools/subagent/session.ts
```diff
@@ -162,18 +157,15 @@ export class ManagedSubagentSession {
    * parent must not receive a completion notice for it. Throws if already disposed or still
```
hiyouga, line 161: **Fixed in 5faee270:** the doc still said "`opts.thinkingLevel` pins this round only (a host follow-up's per-turn picker)" — the option is gone from `startRun` / `SubagentMessageOptions` in this PR.

packages/core/src/omnimessage/types.ts
```diff
@@ -100,10 +100,15 @@ export interface ToolDefinition {
...
- * parameters (e.g. the thinking level, passed with each run) never belong here; a
- * `thinking_level` field still present in a legacy Trace's meta is ignored on resume.
+ * Session metadata: the runtime configuration of one **model context**. One `session_meta`
+ * opens every Trace file — the Session's first, and each file a compaction's rotation starts.
+ * The model reference, the paths and the origin are fixed for the Session's lifetime; the
+ * assembled system prompt is fixed per context — a context is assembled from the Agent State
+ * as it is when it opens (the template, `AGENTS.md` and the other placeholders re-read), so
+ * each file's meta carries the prompt its context actually ran with; the context's toolset
+ * follows as the `tool_list_ready` record. The thinking level is fixed per context as well
+ * and recorded here: everything that shapes the request prefix — model, prompt, toolset,
```
hiyouga: **Fixed in 5faee270:** this paragraph said the thinking level is "fixed per context … and recorded here", contradicting the field-level note twelve lines down ("deliberately NOT here … displayed, never read"), the actual payload (no writer sets it) and this PR's own `omni-message.md` ("nothing records it").

changelog/unreleased/2026-08-28-context-assembled-per-rotation.md
```diff
@@ -0,0 +1,115 @@
...
+- The Environment is re-equipped for the new context (`Environment.reconfigure`): the vault's
+  values go straight into the command environment of every command spawned from then on
+  (processes already running keep the environment they started with). MCP connections are cached
+  by config: a server whose entry is unchanged keeps its live connection and discovered tools —
+  a Session that compacts every turn does not respawn its servers — a removed or changed one is
+  closed, and only new, changed or previously failed servers connect, bracketed by the same
+  `mcp_connect_begin` / `mcp_connect_end` pair the first run streams (naming just those servers),
+  followed by the new `tool_list_ready`; the engine yields them live.
+- Each Trace file a compaction's rotation opens starts with the `session_meta` recording the
+  prompt and level its context runs with, then the connect pair (if any) and the toolset record;
```
hiyouga: **Fixed in 5faee270** (both languages): "recording the prompt **and level** its context runs with" records something the code deliberately does not do (`SessionMetaPayload` has no level; line 38 of this same entry says "nothing is recorded in the Trace"). Also `{ tools, llm }` → `{ llm }` for the bootstrap contract.

packages/core/src/engine/context-engine.ts
hiyouga: Decision: known, deferred — tracked in #572 (the analysis, the bounded impact and a fix sketch live there). Not blocking this PR.

packages/core/src/agent.ts
hiyouga: Deferred together with the abort question — see #572; the two share one fix (a commit-at-the-end open).

PR #540, opened 2026-08-28
feat: software update runs through one modal — confirm, download progress, restart

hiyouga:
## Summary

Both update entries — the **"New version available"** superscript on the new-chat page's version line and the **user-menu row** under System settings — now open one **Software Update** dialog that behaves like a modern app updater:

1. **Check** (opening with nothing known runs one) → the release is offered with its release-notes link and a **Download and update** button. Nothing is fetched before that button.
2. **Download** with a progress bar. **Continue in background** closes the dialog without cancelling: the row keeps showing the percentage, the avatar dot appears when the release is ready, and the outcome toasts.
3. **Restart and update** once ready — a real restart on both backends.

Before: the superscript opened GitHub; the server dialog ran the whole install inside one 10-minute request with no progress and ended in "restart the service by hand"; in the desktop app "Check for updates" started the download on its own.

## What changed, per package

- **web** — `lib/update-flow.ts` is the pure state machine (`unknown / checking / disabled / up-to-date / available / downloading / ready / restarting / error / unsupported`) over both backends; `lib/use-update-flow.ts` owns the actions, the polling and the toasts (zustand vanilla store, one owner mounted by the modal in `AppLayout`). `update-modal.tsx` + `update-row.tsx` replace `update-dialog.tsx`, `server-update-row.tsx`, `desktop-update-row.tsx` and the install `ConfirmModal`. Badges read the flow (`available` / `ready`).
- **server** — the self-update became a background job (`services/update-job.ts`): `POST /api/version/update` starts (or joins) it and answers at once, `GET` reports `phase` (`resolving / downloading / installing`) and a real `percent` parsed from the installer's `curl --progress-bar` output; a finished run stays readable and can be started again (retry). New `POST /api/version/restart`: with a supervisor (`PENGUIN_SUPERVISED=1`, published to the platform as the new `lifecycle` runtime capability) it exits with core's `SERVER_RESTART_EXIT_CODE` (75) after the graceful shutdown; otherwise `{restarting:false, reason:"no_supervisor"}` and the dialog shows the manual hint. Desktop relay: `POST /api/desktop/update/download`, snapshot state `available`.
- **cli** — `penguin server` / `penguin web` run the service as a child process and supervise it: signals forwarded, exit code passed through, relaunch on the restart code so the relaunch runs the newly installed release. A `tsx` dev run cannot be re-spawned by plain node and keeps running in-process (the server then reports `no_supervisor`).
- **desktop** — `autoDownload` was already off, but the shell called `downloadUpdate()` the moment a release was found; now a check ends in `available` and the download starts only on the page's `download` command or the native offer dialog (**Download / Later**). The native "restart now" prompt follows only a natively started download (no double prompt over the web modal). The fallback-feed retry after a failed package download is preserved.
- **docs** — Web App (Version and Updates), CLI (supervision), Server API (job / restart endpoints), both languages.
- **changelog** — `changelog/unreleased/2026-08-28-update-modal.{md,zh.md}`, marked Breaking for the `POST /api/version/update` shape (the old `{status, output, needsRestart}` lives on as `result`). Nothing on disk changed.

Design counterpart: penguin-harness-design [#85](https://github.com/Prism-Shadow/penguin-harness-design/pull/85) — 06 user-menu / desktop auto-update sections, 05 版本识别与在线升级 + API table.

## Notes for review

- Runtime/platform split respected: the restart trigger is a runtime capability (`services/lifecycle-service.ts`, claimed through `hmr/capabilities.ts`, wired in `index.ts` like the desktop shutdown hook); the job and the routes are platform-side. `RUNTIME_INTERFACES.config` gained `supervised`.
- Progress in server mode is best-effort parsing of the installer's output; a run that prints no bar (npm install) shows an indeterminate bar with the phase caption. The outcome never depends on it.
- The restart wait on the page reloads once `/api/version` answers after having seen the server gone (or with a different version), and gives up with a hint after 90s.

## Verification

- `pnpm build` (all packages) + `pnpm typecheck` — clean
- web: 116 files / 1609 tests (full suite; the one sidebar-shape test updated), plus new `update-flow.test.ts`
- server: `version`, `desktop-update`, `update-job` (new), `hmr-resources`, `desktop` — 72 passed
- desktop: 14 files / 176 passed (status fold rewritten for `available` / `download-started`)
- cli: 29 files / 385 passed (`supervisorDecision` tests added)
- docs: 53 passed

https://claude.ai/code/session_013TY4x3XNMLeX6D47pgopR2


hiyouga:
## Review

Read the whole diff, then drove the modal through every state in a real browser against a built server with a faked update backend (the recipe is at the end — it doubles as the answer to "how do I look at this UI when there is no update to be had").

The shape is right. One state machine over both backends, one owner mounted once, the pure/imperative split held to, and the two entry points collapsed into a single explained flow instead of a link to GitHub and a ten-minute blocking POST. The restart-as-a-runtime-capability split is the correct seam, and `SERVER_RESTART_EXIT_CODE` living in core is the right home for a process contract between the CLI and the server.

Four things were broken or missing. All four are fixed in `245771b2`; each was reproduced before and after in the browser.

---

### Fixed

**1. A run that is already over when the POST answers leaves the modal spinning forever.**

`downloadUpdate()` sets `downloadRequested: true`, and the only thing that clears it is the job poll — which runs only while `job.state === "running"` (`use-update-flow.ts:298`). But `UpdateJobService.start()` answers `done` **immediately** when there is no CLI to run (`update-job.ts:174`, `cliEntry === null` → `unsupported`), and `releaseFlow` reads `local.downloadRequested` *before* the finished job (`update-flow.ts:108` vs `:113`). So the one server that most needs to be told "this install cannot update itself" is the one that never gets told:

```
before: "Downloading v9.9.9… / Resolving the release…"   (forever, with the answer already in hand)
after:  "This service was not started through penguin web or penguin server,
         so it cannot be updated from here"  + Open the Releases page
```

The same path swallows an instant `failed`. Fix: the status the POST answers with clears the local flag (`use-update-flow.ts:146`) — it *is* the truth from that moment on.

**2. The restart wait's 90-second timeout only fires on a poll that FAILS.**

The deadline was checked inside `.catch()`, so it can only end a wait that has already seen the server go away. The wait that actually hangs is the one where every poll **succeeds**: the down window fell between two 1s ticks and the relaunched server reports the same version — which is exactly what happens when `penguin update` reinstalled a release the install already had (`classifyUpdateRun` deliberately calls that `updated`). `sawDown` stays false, the version never changes, and the modal sits on "Restarting…" with *no footer buttons at all*, forever. Same for a restart that never happens (`requestRestart()` returning false after the route already answered `restarting: true`).

Verified: with a server that answers `{restarting: true}` and then stays up, the modal now gives up at 90s with the manual hint and the toast, instead of spinning. Deadline is checked on every tick (`use-update-flow.ts:340`).

**3. The avatar's update dot appears where there is nothing behind it (`mode: "none"`).**

`currentFlow()` special-cased `client` and let everything else fall through to `releaseFlow`, and `use-update-badges.ts:72` reads the flow with no mode gate of its own (correctly — it was told the flow has one). A browser signed into a desktop-mode server therefore got:

```
avatar title:      "New version v9.9.9 available"
avatar aria-label: "admin · New version v9.9.9 available"
dot:               present
update row in the menu behind it:  none
```

That is precisely the dead end `update-badges.ts`'s own comment forbids — and its claim that "its flow never offers anything" was not true of the code. Now it is: `mode === "none"` → `{kind: "unknown"}` (`use-update-flow.ts:82`), and the same three checks come back empty.

**4. (gap, not a bug) "Continue in background" did not survive a page reload.**

Nothing ever GETs `/api/version/update` except the poll a POST starts, so a reload during a background download showed "New version available" again while the run was still going — the one promise the feature is named after. The owner now reads the job once on mount (admin only, release mode only).

Deliberately **only a running job is adopted**. My first version adopted any non-idle job, and driving it showed why that is wrong: a finished `unsupported`/`failed` run then answered every later page load with a stale failure and swallowed the badge for the release still on offer. A finished run's outcome was already reported where it was started.

```
row before reload:  "Downloading v9.9.9 66%"
row after  reload:  "Downloading v9.9.9 81%"
```

---

### Not fixed — your call

- **`ready` after a timeout says something false.** The 90s timeout lands on `restart: "manual"`, whose body reads "This service is not supervised by `penguin web` or `penguin server`" — which the timeout does not know, and which is usually wrong (it *was* supervised; the page just lost track of it). The toast says the accurate thing. A third `restart` variant with its own sentence would fix it; I left it alone rather than widen the state machine and both dictionaries mid-review.
- **`supervise()` registers no `child.on("error")`** (`serve.ts:141`). A spawn that fails emits `error` on the ChildProcess, and unhandled that is an uncaught exception in the supervisor. Nearly unreachable (`process.execPath` always exists) — but `spawnUpdateRunner` handles exactly this case eleven files away, so the asymmetry is worth a line.
- **Nothing bounds the respawn loop.** A child that exits 75 on startup is respawned instantly and forever. Only reachable through the restart route, so not a real risk today; a "more than N restarts in M seconds → stop" guard is cheap insurance for a supervisor that is now the thing keeping the service alive.
- **The 1s job poll ships the whole output tail every time** — up to 64 KB per second for up to ten minutes, and `UpdateJobStatus.output` is never rendered while `state === "running"` (the modal shows only the bar; the tail is read from `result.output`). Dropping `output` from the running snapshot would make the poll a few hundred bytes.
- **`Modal` exposes no `role="dialog"` unless `headerless`** (`modal.tsx:103`) — pre-existing, not this PR, but this dialog is the one that now owns a destructive action ("Restart and update"), so it is a good moment to notice.
- Nit: `desktop-update-port.ts:7` now runs to ~110 columns where the rest of the block wraps at ~95.

### Checked and fine

- The spawned argv is a fixed literal list and nothing from a request reaches the command, its args or its env; `PENGUIN_LANG=en` pinning the refusal-marker language is sound because the CLI and the server ship from one install.
- Signal forwarding, exit-code passthrough and the `PENGUIN_SERVE_CHILD` guard against a child supervising in turn.
- `advanceUpdateProgress` — `matchAll` clones the `/g` regex, so no `lastIndex` leak between chunks; the verified line ends the download phase even when a bar redraw follows in the same chunk.
- The lifecycle capability survives an HMR swap (same instance re-claimed, so `index.ts`'s handler stays registered).
- The `Breaking:` marker and the `## Compatibility` section are both there, in both languages.

---

### Testing the update UI when there is no update

Nothing on the server can be told to invent a release — `UpdateCheckService` hits `api.github.com` with a hardcoded URL and the only knob is `PENGUIN_UPDATE_CHECK=off`. So fake it in the page, where all three endpoints the flow uses live behind one `fetch`.

1. Build and start a server on a scratch data root:

```bash
pnpm -r build
PENGUIN_HOME=/tmp/penguin-updtest PORT=8977 HOST=127.0.0.1 \
  PENGUIN_WEB_DIST="$PWD/packages/web/dist" \
  PENGUIN_SEED_ADMIN_PASSWORD=penguin-2026 \
  node packages/server/dist/index.js
```

2. Open `http://localhost:8977`, sign in as `admin`, and paste the snippet below into the DevTools console (a DevTools *snippet* is better — it survives reloads with one Ctrl+Enter). It intercepts `/api/version/update-check`, `GET|POST /api/version/update` and `POST /api/version/restart`, keeps its fake state in `sessionStorage` so it survives a reload, and exposes `window.penguinUpdate`.

<details><summary><b>update-stub.js</b> — paste into the DevTools console</summary>

```js
/**
 * Fake update backend for the Web App's "Software Update" modal — paste into the DevTools
 * console (or add it as a DevTools snippet) and drive it with window.penguinUpdate.
 *
 *   penguinUpdate.downloading(42)      // a run in flight at 42%
 *   penguinUpdate.downloading(null, "installing")
 *   penguinUpdate.installed()          // finished: the modal offers "Restart and update"
 *   penguinUpdate.failed()             // finished: the output tail + Retry
 *   penguinUpdate.unsupported()        // this install cannot update itself
 *   penguinUpdate.startsAs = "failed"  // what POST /api/version/update turns into
 *   penguinUpdate.restart = { restarting: false, reason: "no_supervisor" }
 *   penguinUpdate.reset()
 *
 * The fake state lives in sessionStorage, so it survives a page reload (which is how the
 * "continue in background" path is exercised). Close the tab to forget it.
 */
(() => {
  const KEY = "penguin.updateStub";
  const real = window.fetch.bind(window);
  const json = (body) =>
    new Response(JSON.stringify(body), {
      status: 200,
      headers: { "content-type": "application/json" },
    });

  const DEFAULTS = {
    latest: "9.9.9",
    /** What GET/POST /api/version/update answers. null = idle (nothing has run). */
    job: null,
    /** What POST /api/version/restart answers. */
    restart: { restarting: true },
    /** What POST /api/version/update turns `job` into. */
    startsAs: "running",
  };
  let saved = {};
  try {
    saved = JSON.parse(sessionStorage.getItem(KEY) ?? "null") ?? {};
  } catch {}
  const s = { ...DEFAULTS, ...saved };
  const save = () => {
    try {
      sessionStorage.setItem(KEY, JSON.stringify(s));
    } catch {}
  };

  const running = (percent = null, phase = "resolving") => ({
    state: "running",
    targetVersion: s.latest,
    phase,
    percent,
    output: "",
  });
  const done = (result) => ({
    state: "done",
    targetVersion: s.latest,
    output: result.output,
    result,
  });

  const api = {
    downloading(percent = null, phase = "downloading") {
      s.job = running(percent, phase);
      save();
      return s.job;
    },
    installed() {
      s.job = done({ status: "updated", output: "install ok", needsRestart: true });
      save();
      return s.job;
    },
    failed(output = "curl: (22) 404 Not Found\ninstall.sh: download failed") {
      s.job = done({ status: "failed", output, needsRestart: false });
      save();
      return s.job;
    },
    unsupported(reason = "not_launched_via_cli", output = "") {
      s.job = done({ status: "unsupported", reason, output, needsRestart: false });
      save();
      return s.job;
    },
    reset() {
      Object.assign(s, DEFAULTS);
      save();
    },
  };
  window.penguinUpdate = new Proxy(s, {
    get: (t, k) => (k in api ? api[k] : t[k]),
    set: (t, k, v) => ((t[k] = v), save(), true),
  });

  const start = () =>
    s.startsAs === "failed"
      ? api.failed()
      : s.startsAs === "unsupported"
        ? api.unsupported()
        : s.startsAs === "installed"
          ? api.installed()
          : api.downloading(null, "resolving");

  window.fetch = async (input, init) => {
    const path = new URL(typeof input === "string" ? input : input.url, location.origin).pathname;
    const method = (init?.method ?? input?.method ?? "GET").toString().toUpperCase();

    if (path === "/api/version/update-check") {
      return json({
        currentVersion: "0.0.1",
        buildDate: "2026-08-28",
        latestVersion: s.latest,
        updateAvailable: true,
        releaseUrl: `https://github.com/Prism-Shadow/penguin-harness/releases/tag/v${s.latest}`,
        publishedAt: null,
        checkedAt: new Date().toISOString(),
      });
    }
    if (path === "/api/version/update") {
      if (method === "POST") start();
      return json(s.job ?? { state: "idle", targetVersion: null, output: "" });
    }
    if (path === "/api/version/restart" && method === "POST") return json(s.restart);
    return real(input, init);
  };

  console.log("update stub installed — drive it with window.penguinUpdate");
})();
```

</details>

3. Drive it:

```js
penguinUpdate.latest = "9.9.9";        // what the check offers
penguinUpdate.downloading(42);          // a run at 42%
penguinUpdate.downloading(null, "installing");   // indeterminate bar + phase caption
penguinUpdate.installed();              // -> "Restart and update"
penguinUpdate.failed();                 // -> the output tail + Retry
penguinUpdate.unsupported();            // -> "cannot be updated from here"
penguinUpdate.startsAs = "unsupported"; // what POST /update answers with
penguinUpdate.restart = { restarting: false, reason: "no_supervisor" };  // manual hint
```

`Download and update` → `Continue in background` → reload the page exercises the background path; `startsAs` is what covers the two states the POST can answer with directly (which is where defect 1 lived).

For the **desktop client** half, the same trick works on `/api/desktop/update` (return a `DesktopUpdateStatus` and bump `seq` on every change — the settle watch keys on it). For the **real** server path end to end, the honest way is still an install whose `VERSION` is genuinely older than the newest GitHub release; there is no shortcut for that today. If this becomes a routine chore, a dev-only `PENGUIN_UPDATE_CHECK_FAKE=<version>` in `UpdateCheckService` would be four lines and would let the whole flow run against the real job.

### Verification (with the fixes)

- `pnpm -r build`, `pnpm typecheck` — clean
- `pnpm test` — green across all packages
- Manual: available → download → progress (determinate and indeterminate) → ready → restart → reload, plus the `unsupported`, `failed`, `no_supervisor`, restart-timeout and desktop-mode-browser paths, each before and after the fix

PR #542, opened 2026-08-29
Hooks as a core capability, goal mode and continual learning as plugins, and a plugin library replacing the skill library

### LINKED ISSUE #578: UI margin in agent list page
### What happened

Image
缺少间距

### Steps to reproduce

打开 agent 列表

### Version

v0.2.9

### How it is installed

Desktop app

### Operating system

macOS

### Does it reproduce on a fresh data root?

Yes — an empty data root still shows it

### Relevant output

```text

```

hiyouga:
## Summary

Hooks are now a generic core capability with nothing goal-specific left in core, and the skill library became a plugin library that ships hooks and skills alike:

- **Core codes the hook *points* only** — `stop`, after every Task of a `run` call; `pre_tool_use`, before each tool call's approval; and `user_prompt`, prompt expansion (goal start is the shipped use: `start.mjs` answers `{ context }` sent harness-stamped behind the user's message). Hooks run in core and nowhere else — hosts trigger them through Session APIs (`Session.runUserPromptHook`); the server never spawns hook scripts — plus a subprocess runner. A pre_tool_use script gets the call inline on stdin (`tool_name`, `tool_call_id`, raw `arguments` JSON) and answers `deny` (refuse without consulting approval; the model reads the hook's name and reason as the tool's output) or `allow` (approve without asking) — the project command policy still outranks an allow, since hook packages live in agent-writable state. No built-in plugin ships one; the point exists for custom guards. Hooks come from plugins as **hook packages**: plain Node scripts installed into `agent_state/hooks/<plugin>/` (a first-class part of Agent State beside `agent_state/skills/`), run with `{ hook, session_id, trace_path }` on stdin and a JSON answer on stdout (`decision` / `input` / `reason` / `output` / `subagent`). Every answer is one generic `hook` event; the first `continue` drives the next Task inside the same run; a `subagent` answer spawns a detached child Session.
- **Goal mode is the `goal` plugin** (preinstalled): `start.mjs` writes `GOAL.json` and composes round 1 (the server runs it for `goal: { budget }`), `stop.mjs` reads the round's usage off the Trace and decides — same decision order as before. Core knows nothing about goals; the server 409s `goal_plugin_not_installed` on an Agent without the package and the Web App says so. UI, `goal_*` events, `/goal`, `penguin run --goal` unchanged.
- **`continual-learning` plugin** (not preinstalled; renamed from skill-summary): when the Task that just ended ran more than 30 completed turns, its stop script answers with a `subagent` request carrying that Task's condensed excerpt; the child folds findings into the Agent's `SKILL.md` files. The window is the Task itself, so a Task triggers at most once, at its end.
- **The `[goal]` marker is gone.** Round messages are plain user text; origin rides the structural `sender: "harness"` stamp — Session stamps every stop-hook continue, the server sends the user's goal message verbatim with the plugin's protocol message stamped behind it, and `goal_round` / the CLI round line / the plugin's own usage window count harness inputs (background completion notices share the stamp but are reports, not boundaries — everywhere excluded by their block). The Web App renders injected harness messages as a compact collapsed card (the background notices' form); input history and the outline skip them by the stamp. Fixed along the way: `startGoal` never published the seeded round-1 messages to the session channel, so a second goal on the same session showed no user/protocol message (and no stats row) until a reload; `goal_round` for round 1 is now actually emitted. There is no engine-side special-casing: a harness input held in carry-over is re-sent as-is like any user message (the terminal goal file is what keeps a dead goal dead). `markers/goal-block.ts` (`parseGoalMessage` / `isGoalRoundInput` / `downgradeGoalInput`) is deleted.
- **One npm package per plugin** — `@penguinharness/<name>`, repo root `plugins/<name>/` (`plugin.json` + `icon.svg` + `skills/` + `hooks/`); the **loader lives in `@prismshadow/penguin-core`** (replacing the `packages/skills`/`penguin-plugins` package). It reads the plugin names off the host package's `dependencies` and resolves each package through Node from its own module location — the workspace, an npm install and the packed desktop app all land on their own copy. **There is no bundled `official/` directory any more**: the desktop app declares the plugin packages as `dependencies` (the one manifest field electron-builder carries into the packed app — a `devDependencies` listing shipped an empty library), electron-builder collects them into the app's `node_modules`, and `preflight.mjs` / `verify-packed-cli.mjs` check they are there, since nothing else fails without them. The release chain publishes the plugin packages ahead of core — **each new package name needs a one-time maintainer bootstrap publish before OIDC can publish it; `@penguinharness/continual-learning` is such a name.** Four merged plugins: `software-development`, `model-development`, `agent-development` (with `unified-llm-api` ← agenthub-models and `penguin-config` ← penguin-cli), `agent-tuning`; preinstall unchanged. **A plugin built around someone else's product carries a `use-` prefix** — `use-bento-slides`, `use-firecrawl`, `use-claude-code` — so the package name says what it is for rather than claiming the product; the skills inside keep their names, so existing installs still match the library. (Three more new npm names for the bootstrap publish.) **Three categories**: Office Productivity (goal and continual-learning included — the hook packages sit with the audience they serve), Software Development, AI App Development (agent-tuning included); the Agent Tuning and Session Hooks categories are gone. Versions `YYYY-MM-DD.N` everywhere. `plugin.json` is the single metadata holder — a library SKILL.md's frontmatter is just `name` + `description`, and the loader stamps the plugin's short descriptions and version into the installable copy.
- **The icon belongs to the plugin, and everything it ships inherits it.** `icon.svg` beside `plugin.json` (all 12 built-in plugins ship one) is stamped onto each skill by the loader, written beside an installed skill's `SKILL.md` and beside a hook package's `hooks.json`, and carried by `SkillMetadataItem.icon` (installed lists) and `HookItem.icon`. The settings Skills and Hooks tabs, the composer's skill picker and the chat's skill chips show it; a subject with no icon (a user-authored or zip-imported skill, an install older than the icon copy) falls back to one uniform mark for its kind — the book for a skill, the hook for a hook package, the puzzle piece for a plugin — never a letter. The same puzzle piece is the plugin library's mark in the navigation now; the open book stays the mark of skills.
- **API**: `GET /api/plugins`, `GET /api/plugins/:plugin/files` (everything a plugin ships as text keyed by path — each skill's installable SKILL.md and reference files, the hook scripts), `POST …/agents/:a/plugins`, `GET|DELETE …/agents/:a/hooks`; agent creation seeds `plugins`; `AgentSummary.pluginUpdates` + `hookCount`. **Web**: the library page is `/plugins`; a card's detail Modal is a **two-pane file browser** in the benchmark case browser's shape — a tree on the left with one collapsible group per skill and one for the hook scripts (any number open at once) and a preview on the right (markdown rendered with its frontmatter stripped, everything else as a code block) — so opening a file never replaces the summary or the tree; Agent settings gained a Hooks tab. CLI: `--plugins`.
- **Fixes [#578](https://github.com/Prism-Shadow/penguin-harness/issues/578)**: the Agents page's update notice sat flush against the first card; the title row and the notice now share one header block that carries the gap to the list (the models page's shape).

Design: [penguin-harness-design #86](https://github.com/Prism-Shadow/penguin-harness-design/pull/86) — 05「插件库」「Hook」「Goal 模式」, the API prototype, 06's plugin library page / Agent settings page, 08.

This PR supersedes its own first iteration (the in-core goal hook + `hooks.skill_summary` config), which never merged; the changelog pair describes the final shape only.

## Decisions

- **Subprocess scripts, `.mjs`, builtins only.** Installed packages sit under `agent_state/` with no `package.json` or `node_modules`, so `.js` would be parsed as CJS there and imports of core would fail; `.mjs` with Node builtins runs identically in the library (tested there against fake Traces) and in an Agent's state. Crash / non-JSON / timeout are recorded as failed answers, never fatal.
- **Round 1 is composed by the plugin too** (`start.mjs`), not by the server: the whole `[goal]` protocol lives in one place. The server only knows the file name it reads for the banner and the script it runs to start — both documented as the goal plugin's contract.
- **`GOAL.json` carries `ended`**: the model only ever writes `status`, so that flag is how the stop script tells a verdict this run must act on from a goal an earlier run already ended (it stays silent for those).
- **Token accounting moved into the script** and counts the main session's `token_usage` since the round's `[goal]` input (whole file after a mid-round compaction rotates it); subagent Traces are separate and no longer counted — documented.
- **Versions**: one scheme, `YYYY-MM-DD.N`, compared by date then numeric sequence; an unparseable installed version sorts before every library version, so pre-existing numeric installs read as updatable once.
- Hook package identity = plugin directory name (the event `name` is `continual-learning`).
- **Plugins are resolved, never scanned.** The loader takes the plugin *names* from the host package's `dependencies` and lets Node resolve each one; a scan of some directory (the old `official/`) would have made "which plugins exist" a property of a build layout rather than a declared dependency. The desktop app therefore declares them like any host, and electron-builder — not a copy step — puts them in the packed app.
- **Icons inherit at install time**, not at read time: the installed copy is self-describing (the same reason the frontmatter and `hooks.json` are generated), so the settings tabs read one directory and never consult the library.

## Verification

- Full chain locally: `pnpm build`, `pnpm typecheck` (all packages), `pnpm test` (core 1063 / server 1642 / web 1624 / cli 385 / desktop 180 / docs 53 / landing 69), `pnpm format:check`, `git diff --check`.
- `electron-builder --linux --dir` + `verify-packed-cli.mjs`, then the packed `dist/server.js` run from `resources/app`: `GET /api/plugins` lists the three groups and all 12 plugins (every one with an icon), `GET /api/plugins/goal/files` lists the three scripts, and a project created against it gives `default_agent` its 17 preinstalled skills plus the `goal` hook package — each with the plugin's `icon.svg` beside it. (The earlier `devDependencies` listing produced `groups: []` from the same packed server, which is what the preflight and packed-tree checks now guard.)
- Screenshots against the built web app: the library page (three groups), the detail browser for a multi-skill plugin, a plugin with reference files, and a hook-only plugin; the settings Skills and Hooks tabs showing inherited plugin icons.
- `packages/core` tests run the goal and continual-learning scripts as real subprocesses against fake Traces.

## Compatibility — your call

- Existing Agents get **no hook packages auto-installed**; goal mode on them 409s until `goal` is installed from the plugin library (new `default_agent`s have it). Say the word if you want a boot-time install of preinstall plugins into existing agents instead.
- **Skills and hook packages installed before this change carry no icon** (nothing wrote one beside them), so their rows show the book or hook glyph until the plugin is reinstalled or updated from the library.
- **Old Traces with `[goal]`-blocked rounds** render those messages as plain text (block visible, no round notice), open outline entries and enter input history like any user message — accepted with the marker's removal.
- Installed skills with numeric versions read as updatable once; reinstalling brings the dated version.
- `@prismshadow/penguin-skills` is deprecated (the release chain now publishes `@penguinharness/*`); the old npm package is left as is.
- `hooks.skill_summary` in `system_config.yaml` is no longer read — installing the `continual-learning` plugin is the switch.

https://claude.ai/code/session_01SuuqWTA3XHFS8WVx8gtBnG


hiyouga:
## Review

Read the whole diff against `origin/main` (261 files) in a dedicated worktree, with the core hooks / plugin loader / plugin scripts read line by line and the server, web and CLI/desktop/packaging slices reviewed in parallel, then cross-checked against the design (penguin-harness-design #86). The shape is right: hooks as points in core, plugins as packages, the goal protocol in one place. What follows is what did not hold up — ordered by severity — and what was done about it. Everything marked **fixed** is in the follow-up commit on this branch (`f0fccd49`); the design spec was updated in the same round (penguin-harness-design `docs/stop-hook-goal` @ `dfa401d`).

### Bugs

1. **A second goal POST hijacks a running goal** (`packages/server/src/http/routes/sessions.ts`, `runtime/session-manager.ts`). `runUserPromptHook` took the lock only to resolve the entry and never checked idle, so `start.mjs` overwrote `GOAL.json` (objective B, round 1, budget reset) *before* `startGoal` threw its 409; the running goal's next `stop.mjs` then continued round 2 on objective B. A regression against main, where the file was written inside `session.run` after the idle check. **Fixed**: the hook now runs inside `startGoal`, under the lock, after `assertIdle`; the manager-level `runUserPromptHook` method is gone and the route passes the user's messages. Test added (`goals.test.ts`: the second goal is refused before its start hook runs; 409 `goal_plugin_not_installed` when the Session has no hook).
2. **Installing or uninstalling a hook package never reached a cached runtime.** Hooks are bound when the core Session is built (`Agent.sessionHooks`), and a runtime entry stays cached up to 30 min — install `goal` on an Agent with an open session and goal mode keeps answering 409; uninstall `continual-learning` and every Task end spawns a deleted script. `invalidateAgentRuntimes` exists for exactly this. **Fixed** in the plugins route (install and delete).
3. **A hook-spawned subagent could not be opened anywhere.** `runDetached` drops the child's whole stream, including its origin-stamped `session_meta` — the one message the server registers a child session from (`registerChildSession`) — so the child had no session row, no `session_created` event and no title; the `session_id` on the hook event led nowhere. **Fixed** in core: the spawner returns the child's upfront meta (`SubagentHandle.takeMeta`), `runStopHooks` streams it right behind the hook event, and the server's existing registration path does the rest (a session row, listed with the other subagent sessions). No `subagent` pointer is written to the parent Trace: 05 says the child stays out of the parent's nested view, and this keeps it there.
4. **`max_turns` was not a cutoff for the hook loop.** 05 lists the cap among the cutoffs after which a `continue` must not run, and the session/stop-hook comments say so, but the engine returned `null` at the cap — a `continue` after the max-turns notice ran another Task. No shipped plugin hit it (the goal script detects the notice itself), so this is a contract fix: `RunCutoff.kind` gains `"max_turns"` and the engine returns it. Knock-on, stated in the changelog: a subagent round that hits its cap now reports `[subagent failed: max_turns]` instead of handing the notice text over as its answer. 05's `RunCutoff` block updated. **Fixed.**
5. **Idle-delivered background completion notices rendered as the generic harness card** (`message-item.tsx`). Core stamps every `[background_task_done]` notice `sender: "harness"` (already on main) and the new early return ran before `parseBackgroundTaskDoneMessage`; `e2e/background-notify.spec.mjs` would fail. **Fixed** by checking the notice first — the same order `outline-model.ts` already used.
6. **`pluginInstalled` demanded every part** (`plugins-page.tsx`): an Agent holding an older version that shipped fewer skills, or one a skill was removed from in the Skills tab, showed "Install" and no Update button, while the notice and nav dot (fed by the server's `pluginUpdates`) said it was updatable. **Fixed**: any part present counts; `installedPluginVersion` reads the first installed skill. Tests updated.
7. **`installedPlugins` reloaded the whole plugin library once per installed skill** (`agent-service.ts`): `librarySkill(name)` is `loadLibraryPlugins().find(...)`, so `GET /agents` with default_agent's 17 skills meant 17 full library reads per Agent — which the function's own comment about `SKILL_HEAD_BYTES` was written to prevent. **Fixed**: the library is read once per listing and passed in.
8. **`e2e/skills.spec.mjs:244`** still asserted the old quick-start skill name (`agent-initialization`) after the target became `data-analysis`. **Fixed.**
9. **Docs out of sync with the code**: `goal-mode.en/zh.md` described a `start.mjs` stdin (`objective`, `body`) that does not exist — a hook author following it gets "scratchpad_dir and a non-empty prompt are required"; `skills.en/zh.md`'s `hooks.json` example omitted `user_prompt` and named the field the generator did not write; the changelog said one hook point exists. **Fixed** (and the generator now writes `description_zh`, the key 05 §322 names, instead of `descriptionZh`).
10. **`goal_state` was dropped from `schema.ts` with no migration** — every installed DB (the table shipped in v0.1.3) keeps the dead table and its index forever, and fresh DBs diverge from upgraded ones. Per the repo's backward-compat rule this is your call, so **not fixed**. Options: (a) migration 3 `drop-goal-state` (`DROP INDEX IF EXISTS idx_goal_session; DROP TABLE IF EXISTS goal_state;`, `swapSafe: false`, a `down` recreating the frozen DDL) plus the `-backward-compatibility.md` pair; (b) leave the table and say so in the compat entry. I'd take (a).

### Simplifications and abstractions (all applied)

- `script-hook.ts`: the three adapters shared nothing but built the same stdin object and options three times; now one `scriptRunner(dir, command, timeout)` behind three thin adapters, one `objectAnswer` / `scalarRecord` pair behind the three parsers (the scalar-output filter was duplicated verbatim), and `trace_path: input.tracePath` instead of conditional spreads — JSON drops undefined fields by itself. `failedAnswer(err)` replaces the two private `errorMessage` copies in `stop-hook.ts` / `tool-hook.ts`.
- `listInstalledHooks` re-validated, field by field, a `hooks.json` core itself generates from a typed `HookManifest` — with a second copy of the loader's command-list normalizer. It now reads the file back as the manifest it wrote (the directory name still wins). `HookManifest` carries one required list per hook point instead of `stop` required and the other two optional "for older installs" — hook packages never shipped, there are no older installs.
- The loader hid a broken install: an unreadable host `package.json` gave an empty library, an unresolvable declared `@penguinharness/*` package was skipped, an unparseable `plugin.json` vanished — exactly the failure mode the desktop `devDependencies` mix-up produced. All three throw now with the package or path in the message; `preflight.mjs` checks every declared plugin instead of `goal` alone. `description` is a required manifest field and the "falls back to the first skill's" chains are gone (every `plugin.json` has it; every library SKILL.md frontmatter is `name` + `description` only, as the PR's own rule says — the test named for the fallback exercised a plugin that has its own description). The `PLUGIN_NAME_PATTERN` guards in `libraryPlugin` / `librarySkill` went too (names are Map keys now, never joined into a path), as did the `SKILL_NAME_PATTERN` alias.
- "Harness input that is not a background notice" existed four times (server `goal-events.ts`, `server-task.ts` inline, `message-window.ts`, the plugin's own copy). Core exports `isHookInput` beside the notice parser; the server and the CLI use it (the plugin keeps its standalone copy by design).
- `GET /goal`: the `ended` branch was dead (branches 1 and 3 returned `status` either way), the outcome list duplicated `GOAL_OUTCOMES`, and `fs.stat` fed a field nothing reads. Now `isGoalOutcome` from `goal-events.ts`, one read, no `updatedAt`.
- CLI: `GoalOutcome` re-declared the server's `goal_finished` union with `??` fallbacks on required fields; it now derives from `GoalServerEvent`.
- Web: `trace-event-row.tsx` drew its own hook glyph beside `HOOK_ICON`; `goal_plugin_not_installed` had a hand-written branch beside the `errors.byCode` map; unused `Badge` import; a redundant `installs === undefined` guard; `current?.` inside the branch where it is known.
- `ContextEngine.turns` getter (no reader — a leftover of the superseded in-core summary hook) and core's `uncachedTokens` export (no caller — the scripts carry their own) removed.

### Fields that nothing read (removed)

`GoalStateView.updatedAt` (web `GoalBannerState` has no such field, the CLI never calls `/goal`), `PluginItem.preinstall` (0 readers in web/cli/desktop), `GoalProgress.budget` (the manager publishes `args.budget`), `SkillTile.className` (0 of 4 callers pass it), and the `pluginUpdates ?? []` fallbacks with their three "older runtime" tests — the web bundle is served by the same build as the API, that skew cannot occur. The phantom-compat comments went with them (`HookItem.icon` "installed before icons travelled with it", the changelog's hook-package icon note).

### Not applied

- `RuntimeSession.runUserPromptHook?` is optional only so test fakes can omit it; consistent with the interface's other optional members, so left as is.
- A plugin's `description` is duplicated between `package.json` and `plugin.json`; npm needs the former, so left — worth a test pinning equality if it keeps drifting (the `use-*` packages already differ in escaping).
- The CLI's round line could come from the server's `goal_round` events instead of the stream, but round 1's event lands after the protocol message is rendered; the message-side predicate (now the shared one) keeps the line first.
- electron-builder drops top-level `README*`, `test*`, `example*` from collected modules; no plugin ships any today. A one-line assertion in `plugins.test.ts` would pin that.

### Verification

`pnpm -r build`, `pnpm typecheck` (all packages), `pnpm test` — core 1063 / server 1643 / web 1621 / cli 385 / desktop 180 / docs 53 / landing 69, all green — `pnpm format` + `format:check`, `git diff --check`; all on the review worktree at `f0fccd49`.

hiyouga:
Follow-up on finding 10 (`goal_state`), pushed as `89676526`:

- **Migration 3 `drop-goal-state`** in `db/migrations.ts` drops the table and `idx_goal_session` at the runtime's own open (`IF EXISTS`, so a database this build created migrates the same); `down` recreates the 0.2.9 table empty. It is `swapSafe: false` — **the first restart-only migration**: a hot push of this platform onto a running runtime is refused whole with `RestartRequiredError` before any DDL runs (a rolled-back predecessor would prepare its goal statements against a table that is gone, and its declarative track only runs at a full open). A normal restart, an installer or a package update applies it at open. Downgrading to 0.2.9 works: its `openDatabase` re-declares the table empty; only finished goals' banners fail to restore.
- Tests: the 0.2.4 fixture now carries `goal_state` as 0.2.4 did; a 0.2.9 fixture pins the drop (shape equals a fresh database, `IF EXISTS` on a database without the table, `down` round-trips with the index); the swap-path suite pins the real restart-only migration instead of a fake one. Server suite 1646 green.
- Per the repo's compat rule the batch now has `changelog/unreleased/2026-09-02-backward-compatibility.md` + `.zh.md`, recording the migration and everything else this batch leaves behind (no hook packages on existing Agents, numeric skill versions and old icons, `goal_finished` / `[goal]` in old Traces, the deprecated skills package, `skills → plugins`), and the batch entry's Compatibility section only points at it. The `hooks.skill_summary` note is gone from it: that key never shipped on main.

PR #546, opened 2026-08-30
refactor: extract drafted mechanisms

Myriad-Dreamin:
> **Stack:** `main` ← #599 ← **this** ← #656 ← #383 ← #674 ← #672 ← #673 ← #551 ← #718 ← #719 ← #544 ← #448 ← #450 ← #577 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #637 ← #628 ← #642 ← #639 ← #645 ← #641 ← #526 — one linear chain, in this order since 2026-09-14. Review the diff against `feat/server-module-tree`.

> **Draft.** Stacked on `feat/module-tree` (#543). Design of record: https://gist.github.com/Myriad-Dreamin/a93afcb895591ceaea514f6741c06b22#file-mechanisms-md

## Mechanisms, not implementations

`RuntimeModule` is split into one node per capability, plus `Paths`, `Clock`, `Log` as mechanisms; `BuildDepsOverrides` is deleted — every double a test used to smuggle through it is now an interface with a production implementation (`SystemClock`, `ScryptHasher`, `DefaultMessagingTuning`, the four transport providers, `CoreSessionLoaders`, `DefaultTitleGenerators`, `GlobalFetch`) and a test stands in a **replacement** for the node (`bootAppDeps(config, [[SystemClock, { now }]])`), checked like the node itself. Components declare what they implement (`class SystemClock implements Clock`, recorded by the generator), and a provider declaring the interface wins the wiring over one that merely has the shape. Step B: `src/mechanisms/*.ts` declares every mechanism as an interface class with full signatures (`Users`, `Access`, `SessionIndex`, `ProjectConfigStore`, `Errors`, `Traces`, …); each implementation `implements` its own; every `@Use` field, route dependency and service dependency names the mechanism — **0 of 192** `@Use` fields name an implementation now, and `gen-ifaces` makes any such field an error. `AuthService` requires `Users`, `AuthSessions`, `AuthState`, `Clock`, `PasswordHasher`, `InitialProjectProvisioner` — nothing about SQLite or the runtime.

Step C: the tree is grouped — the root has 14 children, each group a `@Module({ children, exports })` whose exports are forwarded from the child declaring each interface; a group's other nodes are private to it (lexical visibility, ancestors excluded). A whole group is replaceable; so is any node. Ordering follows the exported child, not the group (so groups may need each other's children without a cycle), and the generator records each requirement's providing module as `from` — the table and the page now show which module depends on which.

## Opaque identities

An opaque host type is `<declaring package>#<name>`, never a bare name (`@types/node#Request`, `@prismshadow/penguin-core#Resources`); mapped types (`Pick<…>`, `Record<…>`) project as the shapes they are, an alias with methods as an interface under its own name. The manifest records each node's kind; the page marks components, modules and exporting groups.

## Hot replacement by extensions

`Extension { modules?, replaces? }`: `package.json#penguin.replaces` lists manifests under the names of the nodes they stand in for — a component, a module, a whole group with children and exports of its own. `moduleDefOf(root, { replace })` puts the stand-in where the class would have been without looking at the class; `bootModules` then checks the assembled tree before any node runs (requirements at signature level, exports resolved to a child, provisions present on the instance) and refuses a stand-in that offers less, by name. Two extensions replacing one node is a load-time conflict. Production lookups and the test helpers read the tree through the groups' exports, never an implementation node by name.

Tests (`test/hot-replace.test.ts`): a memory `Settings` replaces `ServerSettingsRepo` and the admin routes serve its numbers; a stand-in missing `getAttachmentLimitsMb` is refused naming the method; one providing nothing is refused at the group's export; one naming an unknown interface is refused; a whole `SettingsModule` is replaced by a two-child subtree of its own; two extensions on one node conflict. Kernel: `replace` stands in by name, is checked as the tree, and must keep the node's name.

## Verification

Server 1506/1506, core 1061, root typecheck, four extension typechecks, docs, format.

---
**Base note (2026-09-04):** this PR's base is `feat/module-tree`, which is where it actually branches from. It was labelled `feat/server-module-tree` (#599), but the two are **siblings**, not a chain: #599 puts each interface beside the thing it describes, this one collects them in `src/mechanisms/` and replaces `BuildDepsOverrides` with node replacements. Both were rebased onto `main` (`a64d08a2`) on 2026-09-04.


Myriad-Dreamin:
Pushed 77cf214d onto this branch: the module-tree boot had dropped `buildAppDeps`'s `migrate(db, { swapPath: true })`, so a pushed platform on this line applied no schema migration at all (found through #617 — a machines table adopted from the earlier line never gained `session_pid`). The call is back in the platform node's `create()`, after the capability claim and before any node is created. Branches above this one pick it up on their next rebase; `feat/server-module-tree` (#599) carries its own copy of the tree boot and lacks the call as well.

Myriad-Dreamin:
A second fix on this branch: the tree's HTTP assembly gated auth on `/api/*` once and nothing else, so the machine proxy — a protected group at `/server/` — ran with no user on the context and threw on `c.var.user.isAdmin`. Every call the page routes to a machine (history, session list, Trace) answered 500 as "Failed to load history: Internal server error". `buildAppDeps` on main gates that prefix explicitly; the tree now gates every protected group outside `/api` on its own prefix. No test covered the proxy on this line; one is added (401 / 403 / the proxy's own 503). Verified on the Windows installation: `/server/<id>/api/version` and the proxied session list answer 200 for the admin, 401 without a session.

hiyouga (approved):

hiyouga (approved):

PR #551, opened 2026-08-30
The command palette and the harness history

Myriad-Dreamin:
> **Stack:** `main` ← #599 ← #546 ← #656 ← #383 ← #674 ← #672 ← **this** ← #718 ← #719 ← #544 ← #448 ← #450 ← #577 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #637 ← #628 ← #642 ← #639 ← #645 ← #641 ← #526 ← #726 ← #727 — one linear chain, in this order since 2026-09-14. Review the diff against `feat/plugin-npm-install`.

> **Draft.** Stacked on `feat/plugin-npm-install` (#672); #526 moved to the top of the chain on 2026-09-14. This PR is now the command palette and the harness history only. What it used to carry besides moved out, each to its own PR: the plugin platform to #383 (2026-09-09), workflows as module trees and Agent packages to #718 and #719, stacked directly on this one (2026-09-14). The final tree of the three is byte-identical to what this branch held before the split.

**Ctrl+P command palette** (⌘P on macOS) — a VS Code-style palette, mounted once in `AppLayout`. Mechanism only in `lib/command-palette.ts` (shortcut matcher, token filter, unit-tested); commands register in `features/palette/app-palette.tsx`.

**Harness history**, the first command — an overlay over the current page (Escape, close, or a click on the margin returns you where you were): the versions this server's data root committed through hot updates, newest first — provenance, commit time, the content-addressed bundles, the current mark — and **what each push changed**. `deploy.mjs` sends the platform's `ifaces.json` with the push; the host stores it by its own sha256, records the summary in the history (`<root>/hmr/history.json`, newest 100), and `hmr/ifaces-diff.ts` diffs two tables node by node and member by member. Routes: `GET /api/version/history`, `/history/ifaces/:hash`, `/history/diff?from=&to=`. **Rollback**: the platform keeps the last five versions it booted and pushes one back through the runtime's own `/api/hmr/upgrade`; the overlay offers *Roll back to this version* (admin, two clicks). A version's whole module tree is shown beside its changes.

## A pushed platform boots on an older runtime

A push lands on whatever runtime is already there, and that runtime publishes the config and the hosts *its* build knew. Two small commits make the platform tolerate that instead of refusing the push: `cliEntry` is optional on `ServerConfig` (absent reads as null; `pluginIndexUrl` gets the same treatment up in #526), the replacements slot accepts the older overrides bag, and `pluginHostFrom` reads a host of an older shape as "no plugins". `supervised` stays required: the seam claims it by name (`HMR_INTERFACES.config`) and it arrived together with the lifecycle capability, so there is no older runtime to tolerate — making it optional (an earlier commit here) refused every push instead, which the review caught. The plugin-host tolerance touches `plugin/host.ts` only to read what an older runtime published; it adds nothing to the plugin platform.

## Review fixes (2026-09-14)

The ten findings of the review on this PR are fixed on the branch: the harness history records once the runtime's commit has landed rather than at boot, rolls back through the runtime's upgrade channel in-process (no `http://host:port`, no token, no HTTPS gate), serializes its writes through core's `atomicWriteFile`, keeps a version as the same upgrade body a hand-over forwards (`hmr/pushed-build.ts`, shared with the Machines page), and reports a refused rollback as `lastRollback` instead of a timeout; the palette and the history overlay share Modal's focus containment through `useDialogLayer`. See the review threads for the one-to-one mapping.

## What this branch does NOT carry

The four pushed-skill-library commits are dropped: `packages/skills` no longer exists on main (#542 replaced the skill library with the plugin library), so the fix has nothing to attach to here. #548 carried that change on its own line; its changelog entry still rides here for now.

## Verification

On 57e1ba3c4 (over `feat/sandbox-settings` 23dab7b76): `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` clean; server 1913 tests green (six timed out once under load on the shared box and passed on rerun), web 1884, cli 394, core, hmr, docs, landing and the plugin suites green.

Tests added along the way: palette shortcut + filter; history append / cap / tolerance, the table diff and its three routes including a path-traversal hash; the older-runtime config and plugin-host shapes; the deferred boot record, the refused rollback, the serialized write, and the in-process push body.


Myriad-Dreamin:
Workflows landed on this branch (956e7d9e, 0530f9a6): an Agent's `workflows/<id>/` is an extension package (`package.json#penguin.modules` + `index.mjs` + `ui/`) booted as its own module tree with the server's interface check, published `WorkflowHost` / provided `WorkflowMain`; watcher reload keyed by content hash, versions under `workflows-history/` with rollback (state.json kept), tabs beside Chat in the web app, `workflow_updated` on the event stream, a prompt section so the Agent can write/fix its own workflow, and the contract in the penguin-sdk skill. Tests: `packages/server/test/workflows.test.ts`, `packages/web/test/workflow-tabs.test.ts`.

Myriad-Dreamin (commented):
Review of the palette + harness-history split (10 findings, most severe first). The first two are the red CI on this tip; verdicts: CONFIRMED = reproduced locally, PLAUSIBLE = traced in code but not executed.

**1. [CONFIRMED]** `HMR_INTERFACES.config` still lists `supervised` as a required member while config.ts makes it optional, so the older-runtime tolerance this PR adds is refused at the claim step, and the PR's own `older-runtime-config.test.ts` fails (test (server) is red on this tip).

A runtime whose ServerConfig predates `supervised` publishes its config without it → `lacksMembers` (`value[m] === undefined`) reports `runtime config lacks supervised` → `claimHmrCapabilities` throws → `doUpgradeAll` rolls the push back and the machine keeps its old platform, exactly what the three optional fields were meant to prevent.

Suggested direction: keep `ServerConfig` fully required and fill the absent fields once at the claim seam (the way `authState.firstLoginToken ??= null` already is) instead of making them optional for every consumer. That also fixes the next item.

_(`packages/server/src/hmr/capabilities.ts:114` is outside this PR's diff, so it could not be attached inline.)_

---

**2. [CONFIRMED]** `new LifecycleService(config.supervised)` passes `boolean | undefined` to a constructor typed `boolean` (lifecycle-service.ts:19) once `supervised` became optional — TS2345; the `typecheck` job is red on this tip. At runtime with the field absent `LifecycleService.supervised()` returns `undefined` rather than the `false` the field's doc promises.

_(`packages/server/src/app.ts:299` is outside this PR's diff, so it could not be attached inline.)_

---

The remaining findings are inline comments.

packages/server/src/services/harness-history.ts
```diff
@@ -0,0 +1,371 @@
...
+        }),
+      ),
+    );
+    const token = readApiToken(this.paths.root);
+    if (token === null) throw new Error("no local api token to push with");
+    // The runtime's own door, from inside: this platform is what gets replaced, so the
+    // route that calls this answers before starting it.
+    // Addressed to the App host: on a loopback bind the two loopback names play different
+    // roles (services/preview-token.ts), and only the App one serves the API.
+    const host = loopbackHostRoles(this.config.host)?.app ?? this.config.host;
```
Myriad-Dreamin: **3. [PLAUSIBLE]** Rollback pushes to `http://${config.host}:${config.port}/api/hmr/upgrade`, but the runtime's HMR gate (hmr/routes.ts:43-58) demands `https` on any non-loopback bind and `loopbackHostRoles` only maps 127.0.0.1/localhost. On `HOST=0.0.0.0` (the Dockerfile default) the self-push is answered `403 hmr_disabled`; on `HOST=::` the URL is invalid — after the route already answered 202. The overlay then polls 60 s and shows `rollbackTimeout` with no reason.

The platform already claims `hmrControl` (`upgrade`/`current`/`endpoint`, capabilities.ts:122); calling `hmrControl.upgrade(target)` directly needs no token, host, port or proxy.

packages/server/src/services/harness-history.ts
```diff
@@ -0,0 +1,371 @@
...
+      id: versionId(current?.bundles ?? { platform: null, cli: null, web: null }, summary.hash),
+      rollbackable: false,
+      source: current?.source ?? null,
+      pushedAt: current?.pushedAt ?? this.clock.now().toISOString(),
+      bundles: current?.bundles ?? { platform: null, cli: null, web: null },
+      ifaces: summary,
+    };
+    const entries = await this.entries();
+    const at = entries.findIndex((e) => sameVersion(e, entry));
+    if (at !== -1 && entries[at]!.ifaces?.hash !== summary.hash) return;
```
Myriad-Dreamin, line 191: **4. [PLAUSIBLE]** `record()` runs at boot while harness.json still names the PREVIOUS version, and the 'leave another platform's line alone' guard only fires when a line for those bundles already exists. So on the first push after this feature lands (or any commit the 1.5 s timer missed) the previous version's line is written with the NEW platform's table, and on a first push over a packaged install a phantom `packaged-<newhash>` line is added.

Root has V1 committed and no history yet. Push V2: V2's `setup()` → `record()` sees harness.json = V1, `entries() = []` → writes `{bundles: V1, ifaces: T2}`; 1.5 s later `{V2, T2}` is added. V1's line permanently carries T2: the overlay's 'changes since V1' for V2 is empty and V1's Tree view shows V2's tree. The tests pre-seed the previous line, so neither case is covered. Recording once the commit has actually landed (`hmrControl.current()` resolves after the swap's persistVersion) removes both.

packages/server/src/services/harness-history.ts
```diff
@@ -0,0 +1,371 @@
...
+    } catch {
+      return null;
+    }
+  }
+}
+
+/** The current commit as the history should show it: what `readHarnessInfo` says, or null. */
+export type { HarnessInfo };
+
+async function writeAtomic(file: string, text: string): Promise<void> {
```
Myriad-Dreamin: **5. [PLAUSIBLE]** `writeAtomic` (and `keepArtifacts` at line 195) use a tmp name keyed only by `process.pid`, yet `record()` runs unserialized from `setup()`, the COMMIT_SETTLE_MS timer and every `list()` call. Two overlapping writers truncate each other's tmp; a torn history.json parses as `[]`; the next `record()` writes `[entry]` and `keepArtifacts`' prune then `rm -rf`s every other `versions/<id>` — the whole history and its rollback copies gone. Same for `versions/<id>.<pid>.tmp`: B's `rm(tmp, {recursive, force})` deletes A's half-copied tree.

Core's `atomicWriteFile` (packages/core/src/internal/atomic-write.ts: random suffix, flush, rm on failure, already used by four server services) exists for exactly this.

packages/web/src/features/harness/harness-history-overlay.tsx
```diff
@@ -0,0 +1,490 @@
...
+      .getVersionHistory()
+      .then((data) => setHistory(data))
+      .catch((err: unknown) => setError(err instanceof Error ? err.message : String(err)));
+
+  const startRollback = async (id: string) => {
+    setRollback({ state: "pushing", id });
+    try {
+      await api.rollbackVersion(id);
+      // The swap replaces this platform; the runtime answers /history again once the new one is up.
+      for (let i = 0; i < 40; i++) {
```
Myriad-Dreamin, line 181: **6. [PLAUSIBLE]** The rollback outcome is never reported. The `done` branch is unreachable (it renders inside `!isCurrent(entry, history.current) ? … : null` at line 342, and `done` is only set after `setHistory(h)` makes that entry current); the landed push broadcasts `web_updated`, which `sessions.tsx:704` turns into `window.location.reload()` before any poll iteration matters; and a failed push only ever surfaces as the generic `rollbackTimeout` because the route discards the runtime's error (`void deps.history.rollback(id).catch(() => undefined)`, version.ts:104). `S.harnessHistory.rollbackDone` is dead code in both dictionaries.

Rolling back to a version that predates this PR is also offered without a guard, and that platform cannot answer `/api/version/history` at all. Suggest: drop the poll, key success off the existing `web_updated` reload, and let the route await/propagate the push's failure.

packages/server/src/services/harness-history.ts
```diff
@@ -0,0 +1,371 @@
...
+    const web = str(manifest.web?.manifest)!;
+    await fsp.copyFile(path.join(hmrDir, platform), path.join(tmp, "platform.mjs"));
+    await fsp.copyFile(path.join(hmrDir, cli), path.join(tmp, "cli.mjs"));
+    await fsp.copyFile(path.join(hmrDir, web), path.join(tmp, "web.webz"));
+    const exec: string[] = [];
+    const assetsDir = str(manifest.assets?.dir);
+    if (assetsDir !== null) {
+      const from = path.join(hmrDir, assetsDir);
+      const to = path.join(tmp, "assets");
+      for (const entry of await fsp.readdir(from, { recursive: true, withFileTypes: true })) {
```
Myriad-Dreamin: **7. [PLAUSIBLE]** `copyVersion`/`push()` re-implement `readPushedBuild`/`readAssets` from machines/upgrade.ts (same manifest walk, webz gunzip, base64 asset map, exec-bit rule, gzip body) and already diverge: the copy does not skip the HMR host's own bookkeeping files (`.materialized` marker and `.manifest.json`, hmr/host.ts:124/609-611), so a rollback ships them back as ordinary asset files. `materializeAssets` then digests a map that includes them → the identical set lands under a NEW `store/assets/<sha>` instead of reusing the existing one, and the marker is written from the payload partway through the copy with the OLD digest as content — a crash between that write and host.ts:611 leaves a directory whose marker says complete while files are missing, which the next push trusts.

Reusing `readPushedBuild(dataRoot)` (or its two helpers parameterised on a manifest + dir) removes the third copy of the upgrade-body format.

packages/server/src/services/harness-history.ts
```diff
@@ -0,0 +1,371 @@
...
+    try {
+      parsed = JSON.parse(await fsp.readFile(path.join(this.dir, "history.json"), "utf8"));
+    } catch {
+      return [];
+    }
+    if (!Array.isArray(parsed)) return [];
+    return parsed.map(entryOf).filter((e): e is HarnessHistoryEntry => e !== null);
+  }
+
+  async list(): Promise<HarnessHistory> {
```
Myriad-Dreamin, line 309: **8. [PLAUSIBLE]** `list()` calls `record()` on every `GET /api/version/history` (an `auth: "user"` route), which unconditionally rewrites history.json via tmp+rename even when the line is unchanged, may copy a whole version (~63 MB / ~1,770 serial mkdir+copyFile+stat on a real data root) and re-runs the `versions/` prune; `setup()` awaits the same call, so boot blocks on that copy whenever the committed version has no `versions/<id>/version.json` yet. Any signed-in non-admin can trigger disk writes and prunes by GET, and the overlay's own poll does it 40 times per rollback.

Suggest: skip the write when the upserted line deep-equals the existing one, move `keepArtifacts` off the boot path (after `hmrControl.current()`), and hardlink or reference the runtime's content-addressed store files instead of copying.

changelog/unreleased/2026-08-30-pushed-skills-root.md
```diff
@@ -0,0 +1,15 @@
+# A hot-pushed bundle finds the skill library through the installation that loaded it
+
+- **Date:** 2026-08-30
+- **Type:** fix
+- **Scope:** `skills`
+- **PR:** [#548](https://github.com/Prism-Shadow/penguin-harness/pull/548)
```
Myriad-Dreamin: **9. [CONFIRMED]** This entry and its .zh.md twin describe a `packages/skills` fix (`resolveSkillsRoot`, `PENGUIN_SKILLS_DIR`) that no hunk in this diff makes: `packages/skills` does not exist at the PR head, and the linked #548 is CLOSED, never merged. The changelog rule is that an entry ships with the PR that makes the change, so the pair should be dropped from this PR.

packages/web/src/features/harness/harness-history-overlay.tsx
```diff
@@ -0,0 +1,490 @@
...
+    return () => {
+      cancelled = true;
+    };
+  }, [entry, previous]);
+
+  if (!open) return null;
+
+  return createPortal(
+    <div
+      className="anim-fade fixed inset-0 z-[70] bg-black/45"
```
Myriad-Dreamin: **10. [PLAUSIBLE]** The overlay (and command-palette.tsx line 87) hand-roll `createPortal` + `fixed inset-0 bg-black/45` + the pushEscLayer/isTopEscLayer keydown effect that is the body of `Modal` (components/ui/modal.tsx, which takes `bare` and `widthClass` for dialogs that own their layout), and pick `z-[70]` above the documented layering where portaled popups sit at `z-[60]` (dropdown.tsx:4, 305). Both copies skip what Modal adds: Tab/Shift+Tab containment and focus restore on close, so keyboard users Tab out of the overlay into the page behind it; and any portaled Dropdown/Select inside the overlay paints beneath the z-[70] backdrop. `<Modal bare widthClass=…>` inherits both for free.

packages/server/src/services/harness-history.ts
Myriad-Dreamin: Fixed in 80ae75904: `push()` hands the kept body to `hmrControl.endpoint()` in-process — a Request built here, never a fetch to `http://host:port` — so the bind address, the HTTPS gate and the api token are out of the picture. The outcome's `web.rev` is broadcast as `web_updated` the way hmr/routes.ts does for a push from outside.

packages/server/src/services/harness-history.ts
Myriad-Dreamin: Fixed in 80ae75904: `setup()` no longer records; it chains the record onto `hmrControl.current()`, which resolves once the swap that booted this platform is over, commit included. A generation the runtime put back is disposed before that resolves and never writes. The 1.5 s timer is gone. Test: *records a boot only once the runtime's commit has landed, and never after being put back*.

packages/server/src/services/harness-history.ts
Myriad-Dreamin: Fixed in 80ae75904: `record()` is serialized through a promise chain, history.json and the tables go through core's `atomicWriteFile`, the kept version's tmp dir gets a pid+random suffix and the prune skips `.tmp-` names. Test: *writes one record at a time, and rewrites nothing when the line is unchanged*.

packages/web/src/features/harness/harness-history-overlay.tsx
Myriad-Dreamin: Fixed in e58a52c3e + 80ae75904: the `done` state and `rollbackDone` (both dictionaries) are gone — a push that lands reloads the tab via `web_updated`. A refusal is now kept by the platform that asked (it is the one still running) as `lastRollback` on `GET /history`, and the overlay's poll shows that message instead of timing out. The route still answers 202 before the swap.

packages/server/src/services/harness-history.ts
Myriad-Dreamin: Fixed in 80ae75904: `readPushedBuild`/`readAssets` moved to `server/src/hmr/pushed-build.ts` (machines/upgrade.ts re-exports the former), and a kept version is now that body byte for byte (`versions/<id>/build.gz` + a one-field version.json). `copyVersion` and the second push encoder are deleted, so the `.materialized` marker is skipped the way the hand-over skips it.

packages/server/src/services/harness-history.ts
Myriad-Dreamin: Fixed in 80ae75904: `list()` writes nothing — it awaits the boot's record (so a read that arrives first waits instead of racing it) and reads. The record itself skips the write when the upserted line is unchanged, and `keepArtifacts` runs after `hmrControl.current()` resolves, off the boot path.

changelog/unreleased/2026-08-30-pushed-skills-root.md
Myriad-Dreamin: Dropped in 57e1ba3c4 (both files).

packages/web/src/features/harness/harness-history-overlay.tsx
Myriad-Dreamin: Fixed in e58a52c3e: Modal's esc-layer + focus-enter/restore + Tab containment moved into `useDialogLayer` (modal.tsx), which Modal, the palette and the history overlay now share; both overlays carry `ref`, `tabIndex={-1}`, `aria-modal` and the panel keydown, and sit at z-50 like Modal so portaled menus (z-[60]) paint above them. The two keep their own layout (`inset-3` / top-aligned), which Modal's panel classes do not express — hence the hook rather than `<Modal bare>`.

Myriad-Dreamin:
Review follow-up: all ten findings fixed on this branch, in four commits on top of the restacked tip.

- **1 + 2** (`supervised` optional → claim refuses, TS2345): 64f691593 — `ServerConfig.supervised` is required again. `HMR_INTERFACES.config` names it, and a runtime that publishes the lifecycle capability publishes it with it (both came in #540), so there was no older runtime to tolerate; `cliEntry` (and `pluginIndexUrl`, up in #526) stay optional. typecheck and `older-runtime-config.test.ts` are green.
- **3, 4, 5, 7, 8**: 80ae75904 — harness-history.ts (see the inline replies).
- **6, 10**: e58a52c3e — web.
- **9**: 57e1ba3c4 — the pushed-skills-root entry is gone.

Verification on 57e1ba3c4 (over `feat/sandbox-settings` 23dab7b76): `pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` clean; server 1913 tests green (six timed out once under load on the shared box and passed on rerun), web 1884, cli 394, core, hmr, docs, landing, plugins green. The chain above was replayed onto the new tip (tail #526 = cb4b8f052).

hiyouga (approved):

Myriad-Dreamin:
Continued in #768: GitHub closed this PR as merged into a stack branch during a restack, so it can no longer be merged into `main`. Nothing from it reached `main`. #768 carries the same branch and content against `main`.

PR #577, opened 2026-09-02
feat(web): the machines page goes live

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← #808 ← #718 ← #544 ← #448 ← #450 ← **#577** ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

Three lines, and the feature the stack before it built becomes reachable:

- `machines` leaves `UNRELEASED_NAV_KEYS` in `nav-group-collapse.ts`, so the sidebar shows the row to admins — the gate `main` already held for exactly this.
- `/machines` gets a route, so the Machines page can be opened.
- The sidebar's **new workspace** picker offers the machine row, so a workspace can be created on a machine rather than only here.

Everything either one shows was already built and tested by the PRs below it; what was missing was the way in. Today on `main` the page component exists with no route and its nav row is gated off, so it is unreachable — this opens that entry, and nothing else.

Kept as a change of its own so that turning the feature on is a decision with a date, separate from the code that makes it possible — and so that turning it back off is a revert of a few lines rather than of a feature.

Both surfaces are admin-only in effect: the list they read answers `403` to everyone else, and a non-admin who types the URL gets the page's own refusal.

## Verification

Web `typecheck` and `format:check` clean; web suite **119 files / 1673 tests** passing.


Myriad-Dreamin:
Reopened on the rebased #450 (`feat/machines-run-agent` @ 3479623b3): this branch is the one commit that opens the gate, and it merges last, after #448 and #450 have landed. The chain above it (#614 and the closed branches up to #641) is restacked on it.

Myriad-Dreamin:
GitHub cannot reopen a PR whose branch was force-pushed, so the gate continues as #655 — same branch `feat/machines-enable`, same single commit, now on the rebased #450.

Myriad-Dreamin:
Restored on this number: the branch was pointed back at 53e7a4af5 long enough for GitHub to reopen it, then moved to the rebased head 34372996d. #655 is closed and superseded by this.

hiyouga (approved):

PR #582, opened 2026-09-02
The compaction row shows the request's thinking beside its result, both streaming, with the wall time

hiyouga:
## Summary

The conversation's compaction row now follows the finished work group's shape: the header's title names the mode and doubles as its status — 「压缩中」/ "Compacting" while it runs, 「压缩完毕」/ "Compacted" once done — and carries the wall time (ticking while the compaction runs, settled once done), and its collapsed body holds two sections — the compaction request's **thinking** and its **result** — both streaming while the request runs.

- **Core** forwards the compaction request's thinking on the output stream between the paired compaction events, the same way it already forwarded the summary text: `partial_thinking` fragments verbatim, or the complete `thinking` when no fragment carried content, never both. The request keeps the thinking level its model context already has — nothing switches thinking on for it (`runCompactionRequest` passes no `thinkingLevel`, as before). The Trace is unchanged.
- **Web** — the stream model accumulates the span's thinking onto the running compaction item (`thinkingText`, `thinkingStreaming`) beside `summaryText`, rebuilds it identically from the span's complete thinking on history replay, and drops both drafts when a compaction does not complete. `compaction-banner.tsx` keeps the StepBanner shell; its body is two `DisclosureRow`s collapsed by default like a thinking block — 「思考」/ "Thinking" (present only once any thinking arrived) and 「压缩结果」/ "Result" — each rendered with the thinking block's `md-body` + streaming `Md`. The header's title doubles as its status, the work-group header's own Running/Done idiom: 「压缩中」/ "Compacting" while it runs, 「压缩完毕」/ "Compacted" once done (「清空中」→「清空完毕」, "Clearing" → "Cleared" for a discard), with no detail line; the failed state keeps the bare mode word over its single-line reason; `discard` stays chevron-less.
- **CLI** stays quiet on the streamed thinking inside the span, as it already did for the summary. The Trace page (raw events) and the server are untouched — the live tail already tracks `partial_thinking` fragments generically, so a join mid-compaction is seeded with the thinking prefix too.
- **Docs**: `agent-loop` and `web-app` pages (en + zh); changelog pair `2026-09-02-compaction-banner-thinking`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/90

## Verification

```
pnpm --config.verify-deps-before-run=false -r build                                   # EXIT=0
pnpm typecheck                                                                        # all packages Done
pnpm --filter @prismshadow/penguin-core exec vitest run test/compaction.test.ts       # 47 passed
pnpm --filter @prismshadow/penguin-cli exec vitest run test/render.test.ts            # 63 passed
pnpm --filter @prismshadow/penguin-web test                                           # 118 files, 1623 passed
pnpm --filter @prismshadow/penguin-docs test                                          # 53 passed
pnpm --filter @prismshadow/penguin-server exec vitest run test/abort-compaction.test.ts test/session-manager.test.ts test/messaging.test.ts test/messages-page.test.ts test/trace-service.test.ts
pnpm format && pnpm format:check                                                      # clean
git diff --check                                                                      # clean
```

Rework (title doubles as status), verified on the remote test host (Node 24.20): `pnpm typecheck` all packages Done; `pnpm --filter @prismshadow/penguin-web test` 119 files, 1671 passed; `pnpm --filter @prismshadow/penguin-docs test` 53 passed; `pnpm format:check` clean.

## Notes

- No backward-compatibility handling: nothing on disk changes shape (Traces already record the compaction span's thinking; only what is forwarded on the live stream and what the reducer keeps changed).
- The thinking section shows no per-section duration; the row's wall time lives in the header, as in the work-group header. Attempts of a retried compaction concatenate into one thinking draft, exactly as the summary draft already does.
- Conflict hotspots: `packages/web/src/lib/omni/stream-model.ts` (compaction branch only), `strings.ts` / `strings-en.ts` (two keys beside `compactionTitle`), `packages/docs/content/{agent-loop,web-app}.{en,zh}.md`.


hiyouga:
## Review

No defects found. I traced the three sides that had to agree and they do:

- `sawPartialText` / `sawPartialThinking` are per-attempt, because `runCompactionRequest` is
  invoked once per attempt inside `summarizeContext`'s retry loop (`context-engine.ts:1572`).
  A model that streams attempt 1's thinking as fragments and delivers attempt 2's whole
  therefore still gets attempt 2 forwarded exactly once, which is what the new
  "tracked per kind" comment claims.
- The CLI cannot leak the compaction's thinking through the complete-message path either:
  `renderNow` drops every complete `model_msg` (`render.ts:740`, "Complete (non-streaming)
  model_msg is never rendered"), so the new `compactionActive` early return in
  `handlePartialThinking` closes the only door core opened.
- Live and replay converge: live sends fragments (complete suppressed once a fragment carried
  content), replay reads the span's complete `thinking` messages, and both land in
  `appendCompactionThinking`. A content-free `start`/`stop` phase only moves the flag, so a
  request that never thought leaves `thinkingText` undefined and the section unrendered
  rather than an empty disclosure row.

### Notes

- **The `streaming` / 「生成中」 detail is the third thing on that row saying "running".**
  `StepBanner` already gives a running row a spinning `StatusIcon`, a `toneInk.busy` title and
  a ticking `LiveDuration` (`step-banner.tsx:49-73`); the new detail adds a fourth mono token
  between them. I take the intent — collapsed, the chevron alone does not say *content is
  arriving now* — but the row is dense enough that this reads as chrome rather than
  information, and this repo's house preference is an icon carrying the state over a textual
  marker beside it. Worth a second look; I left it alone because it is a deliberate,
  documented design decision rather than a defect, and per CLAUDE.md that is a conversation
  before it is an edit.
- While the request is still thinking, the **Result** section renders with a running status
  icon over an empty body. Correct by the "empty like a thinking block" rule, but it does mean
  a reader who expands early sees a live-looking section with nothing in it.
- Multiple attempts concatenate into one thinking draft with no separator between them, as the
  summary draft already does. Called out in the PR notes, so this is only a restatement: with
  thinking added, the concatenation is now long-form prose from two different attempts running
  together, which reads worse than two summaries did.

Verified: read-only review, no changes pushed. I did not re-run the suite — the PR records a
green `-r build` / `typecheck` / core, cli, web and docs tests, and I changed nothing that
could invalidate it.

hiyouga:
Manual verification feedback on the compaction row:

The compressed row should behave like the running row above it. For the **thinking** block and the **compression result** block:
- Auto-expand the two headers (thinking and compression result) so it is clear they are present, but do **not** expand their specific content.
- Both should also show a **time** (wall time), just like the running rows.
- After completion, these two also collapse again, leaving only the compressed summary.

For both compressed and running rows, guarantee:
1. The first layer starts **expanded**, the second layer stays **closed**, appended one by one (逐个 append).
2. Once the whole second layer completes, the first layer **auto-closes**.

So that whether the row is running or compressed, the expand/collapse behaviour is consistent.

hiyouga:
## Rework of the expand/collapse behaviour — 1eacac1e

Point by point against your spec.

### 1. The two body headers are auto-expanded, their content is not

`StepBanner` held `useState(false)` — the row was collapsed for its whole life. It now takes the
work group's shape verbatim: `useState(running)` plus an effect that follows `running` only while
`userToggled.current` is false. So a running compaction row is open on its two section rows, and
each section is a `DisclosureRow` whose `defaultOpen` is (and already was) `false`, so their
contents stay closed. I confirmed that rather than assumed it — `disclosure-row.tsx:88`,
`useState(defaultOpen)` with `defaultOpen = false` in the signature.

A manual toggle pins the row: a reader who opens a finished compaction to read it keeps it open,
and one who closes it mid-run keeps it closed. `StepBanner`'s other caller — the MCP connect row —
passes no children while it runs, so `expandable` is false there and nothing about it changes.

### 2. Each of the two rows shows its own wall time

`DisclosureRow`'s `trailing` slot, rendered exactly as `thinking-block.tsx` renders it:
`LiveDuration` while streaming, `humanizeDuration` once settled, same mono/gray classes, same slot
order. Header, section and thinking block now read identically.

Before the first summary token the result section shows `LiveDuration`'s `…`. That happens to
answer the second note in my earlier review — the section that used to read as "content is arriving
now" over an empty body now says "waiting", which is what that moment actually is.

### 3. Both collapse again after completion

`running` flips false at `compaction_end`, the effect closes the banner, and the sections go with
it. Screenshotted below.

### 4. The two-layer invariant, for the work group too

It did **not** already hold. Layer 1 opening and layer 2 staying closed were both fine, but the
auto-close was driven by `isLast` — "later messages pushed this group away from the end". In every
ordinary turn that fires at the right moment, because the `task_stats` row lands as the Task ends;
but a turn that ends **on the group itself** left it open under a "Done" header, and so did a
replayed history whose last segment is a work group.

The policy now follows `active` — the same value the header already uses to print Running/Done.
The no-flicker property the old comment protects is kept: `active` is `(isLast && taskRunning) ||
itemsRunning`, so the `taskRunning` half bridges the gaps between two steps within a turn, which is
what `itemsRunning` alone would have flickered on. In the common flow both flip together, so this
is a strictly narrower behaviour change than it looks. One side effect worth naming: a tool still
executing after body text already streamed now keeps its group open instead of closing it.

### Timing data — the boundaries I chose

`CompactionItem` gains `thinkingStartedAtMs` / `thinkingDurationMs` and `summaryStartedAtMs` /
`summaryDurationMs`, accumulated in the reducer from the span's own messages live and rebuilt from
the span's complete messages on replay. The rule, written into the doc comment on
`thinkingStartedAtMs`:

> **A section is timed over exactly the window in which it shows itself running**, so the number
> and the spinner beside it can never disagree.

- **Thinking**: starts at the first fragment carrying thinking content (a content-free
  `start`/`stop` still leaves no draft and now no stray time either), settles at that fragment's
  stop — the moment its spinner stops. Not "the first summary text": the stop always arrives on a
  completed request, and a fragment left open by an aborted one has its drafts discarded anyway.
  The one case where the stop never lands on a completed compaction already had a branch here, and
  the duration now settles in it rather than leaving a settled row with no number.
- **Result**: starts at the first summary token, settles at `compaction_end`. Deliberately **not**
  its own `partial_text` stop. The section has no per-section streaming flag — it renders from
  `item.running`, so it spins until the row does. Settling at its own stop would freeze a number
  smaller than the ticker was already showing, and the value would visibly snap backwards, which is
  the exact failure `work-group.tsx:153-162` warns about. The cost is that the result's span
  includes the adoption tail; the header still carries begin-to-end, so the two sections decompose
  it and the remainder is the request setup plus that tail.
- **Replay**: no fragments in the Trace, so each start is approximated with the previous message's
  time and the complete message's own time settles it — the `ThinkingItem` convention
  (`settleThinkingDuration`). Measured on a real reload: live `思考 3.6s` → replayed `思考 4.5s`
  (replay counts from `request_begin` rather than the first thinking token), `压缩结果 2.7s` in
  both.
- **A retried compaction** concatenates attempts into one draft today. Each start is therefore
  recorded **once** and each later stop only extends the settled value, so the wall time covers the
  whole span that produced the prose on screen — the failed attempt and the gap after it included.
  Timing the last attempt alone would print a short number against text it did not write. Pinned in
  `stream-model.test.ts` ("a retried compaction times the whole span its concatenated draft
  covers"): 7s of thinking across two attempts, not the 3s of the second.

### Verified

Unit, and then by hand — you found this by hand, so unit tests alone would not have closed it.

```
pnpm --filter @prismshadow/penguin-web typecheck        # clean
pnpm --filter @prismshadow/penguin-web test             # 118 files, 1626 passed
pnpm --filter @prismshadow/penguin-docs test            # 53 passed
pnpm format && pnpm format:check                        # clean
```

Four new/extended reducer tests in `stream-model.test.ts`: live section timings, the retry span,
replay parity, and the aborted case dropping the timings with the drafts.

**Visual, against a real server.** Built core/server/web, ran the `e2e/` mock-LLM harness on a
scratch `PENGUIN_HOME` and a probed-free port, with a scratch copy of `mock-llm.mjs` that answers
the compaction request with thinking and then the summary, streamed at ~0.9s per delta so the
running state lasts long enough to see. Screenshots taken at each stage; every disclosure button on
the page dumped as `(aria-expanded, text)` at each stage. What I actually saw:

| stage | observed |
| --- | --- |
| work group running | `运行中 0s` expanded=true, child `思考 0s` expanded=false |
| compaction running, before the first token | `压缩 生成中 1s` expanded=true, `压缩结果 …` expanded=false |
| compaction running, thinking arrived | `压缩 生成中 1s` open over `思考 0s` and `压缩结果 …`, both closed, both ticking |
| settled | only `压缩 7.2s` remains — the row closed itself, both sections gone; the work groups above are closed too |
| reopened by hand | `思考 3.6s` and `压缩结果 2.7s`, both still closed |
| after reload | `压缩 7.2s`, `思考 4.5s`, `压缩结果 2.7s` |

All four of your points hold, and the work group shows the same two-layer shape. The instance was
on a scratch data root and is shut down.

`e2e/compaction.spec.mjs` and `e2e/compact-abort.spec.mjs` fail 3 tests here — I baselined them by
stashing the four source files, rebuilding web, and rerunning: **the same three failures at the
same assertions**, so they are pre-existing and unrelated (`compaction.spec.mjs` dies at the
`—/240` context-ring assertion on line 105, i.e. before it ever reaches its collapse assertions on
either branch).

### Not changed, needs your call

The 「生成中」 / "streaming" detail in the running header — the thing I flagged last round as a
fourth "running" marker. With the row now auto-open on two visibly streaming sections that carry
their own tickers, it has less left to say than it did. I left it alone: it is a documented design
decision and per CLAUDE.md that is a conversation, not an edit. Say the word and it comes out in a
line.

Design parity: this changes documented behaviour, so `penguin-harness-design#90` needs the same
edit. I did not touch `design/` (separate repo, separate PR). The impl-side docs
(`agent-loop.{en,zh}.md`, `web-app.{en,zh}.md`) and the changelog pair are updated in place — the
entry described the old collapse behaviour.

PR #583, opened 2026-09-02
feat(web): the shared "Create with AI" kit

hiyouga:
## Summary

The shared "Create with AI" kit: every object the Web App creates from a form gets a second path — describe it to an agent — and this PR ships the reusable pieces the creation surfaces (agents, model groups, Benchmarks, scheduled tasks, vault secrets, hook imports, apps, agent imports) are wired to in sibling PRs.

- `packages/web/src/features/ai-create/` (barrel `index.ts`):
  - `useAiBridge().openAiChat(req)` — parks the active draft, writes the prompt into the draft cache (`buildAiDraft`, pure), pins the agent, navigates to `/chat/new` with `{ agentId, workspace?, autoSend? }` in route state.
  - `pickDefaultAgent(agents)` — `default_agent` (carries the preinstalled plugin library the canned prompts rely on), else the first agent, else `null`.
  - `composeAiPrompt(draft, tail?)` — trimmed draft + one blank line + the surface's fixed instruction tail.
  - `AiCreatePanel` — "done by «agent» in a new chat" line (optional compact agent `Select`), auto-growing prompt box, clickable examples (label + muted description; a click replaces the draft and the chip stays selected while the text still equals it), and a folded **Full prompt** preview (`HelpFold`) with a copy button when a `tail` is given.
  - `AiCreateModal` — `Modal` + panel + Cancel · **Edit in a new conversation** (wand): the one exit, disabled while the draft is blank or no agent resolves; it prefills the new conversation's composer and sending stays the user's move; mounts its body fresh per open.
  - `CreateButtons` — two separate buttons, never a split control or a mode switch: the accented "Create with AI" (wand) and the plain "Create manually" (hand); `onManual` is optional (a member who may not write the form's target still gets the AI path), labels override for objects with their own verb.
- `MAGIC_WAND_ICON` in `components/ui/icons.tsx`; `S.aiCreate` in both dictionaries.
- Nothing is auto-sent: the composed prompt is written into the draft cache as `aiPrefill` (it dies with the draft it seeds unless the user edits it) and the draft page simply opens with it in the composer.
- `skills-tab.tsx` / `memory-tab.tsx` now jump to chat through the bridge.
- Docs (`web-app.{en,zh}.md`) and a bilingual changelog entry.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/91 (`### 「用 AI 创建」模式` in `06-PROTOTYPE.md`).

## Verification

```
pnpm --config.verify-deps-before-run=false -r build   # exit 0
pnpm typecheck                                        # exit 0
pnpm --filter @prismshadow/penguin-web test           # 122 files, 1633 tests passed
pnpm --filter @prismshadow/penguin-docs test          # 7 files, 53 tests passed
pnpm format && pnpm format:check                      # clean
git diff --check                                      # clean
```

New tests: `ai-bridge.test.ts`, `default-agent.test.ts`, `ai-create-prompt.test.ts`, `ai-create-panel.test.ts` (static render: selected example, agent line, composed prompt in the fold).

## Notes

- Rework (2026-09-10): the split "Create ▾" button, the wand-only button, `primaryExit` and the auto-send route state were replaced by the pair of buttons and the single prefill exit; the prompt box uses the small text size and the full-prompt preview lines up with the examples (`HelpFold` gained `flush`).
- The submit is armed via state and fired one render later, so the composer's own pruning of a preselected skill list against the installed skills has committed before the `[use_skills]` block is built.
- Behavior change (intentional, minor): the Memory tab's jump-to-chat now parks typed-but-unsent draft text instead of overwriting it, as the Skills tab already did. The Skills tab keeps its empty-source branch (plain "open a new chat" with the agent, no prefill).
- No backward-compatibility handling: no on-disk data or configuration shape changes.
- Conflict hotspots for sibling branches: `strings.ts` / `strings-en.ts` (new `aiCreate` section right before `agent`), `icons.tsx` (new export before `NAV_ICONS`), `draft-view.tsx`, `chat-input.tsx`, `06-PROTOTYPE.md`, `web-app.{en,zh}.md`.


hiyouga:
## Review

No defects found. The bridge, the prompt composer and the default-agent pick are pure and
tested; the auto-send path is the only risky piece and its guards hold up:
`skillsLoaded` settles through `.finally()` so a failed skills fetch cannot wedge it, the
arm/fire split really is needed (the fire effect's deps are `[autoSendArmed]`, so without
`autoSendFired` StrictMode's doubled effect would send twice), and the sessionStorage
marker keys on `location.key`, so back-navigation and reload both stay quiet.

### Findings not fixed

- **`draft-view.tsx:715` — `const wanted = stateAgentId ?? agentId` makes the guard a
  tautology in the branch it is written for.** `autoSend` reaches the route state only
  through `aiChatRouteState`, and `AiChatRouteState.agentId` is required, so `stateAgentId`
  is always present here; if it ever were not, `wanted` becomes `agentId` and
  `agentId === wanted` degrades to `agentId === agentId`. The fallback cannot fire, and if
  it could it would disable the check rather than perform it. `agentId === stateAgentId`
  says the same thing without the dead arm. Left alone because it is behaviour-neutral —
  say the word and I will take it.
- **`chat-input.tsx:1838` — `submit()`'s boolean return has no reader.** The one call site,
  `draft-view.tsx:744`, is `composerRef.current?.submit()`. So the documented "false means
  nothing is sendable right now" is unobservable: if `canSend` is false at that instant, the
  one-shot marker has already been written and the auto-send is silently gone — the user
  pressed **Send to agent**, landed on a prefilled draft and nothing happened. Reachable
  only through the acknowledged degraded path in `ai-bridge.ts:79` (no `userId`/`projectId`
  ⇒ no draft written ⇒ empty draft ⇒ `canSend` false), which needs a null `currentProject`
  while a create dialog is open. Either drop the return type or let the caller surface the
  failure; keeping an unread signal is the worst of the three.
- **`ai-bridge.ts:54` — `delete draft.handoffAgentId` cannot fire.** The handoff chip is
  only ever persisted by `use-session-draft.ts:71`, which writes to `sessionDraftKey(userId,
  sessionId)`. The slot this touches is `draftKey(userId, projectId)`, and its only writer,
  `draft-view.tsx`'s `persistNow`, builds a fresh object that never carries the field. The
  comment above it asserts a hazard that the key split already rules out.
  `plugins-page.tsx:441` carries the same vestige, so this is a pre-existing convention
  rather than something this PR introduced.

### Notes

- `AiCreatePanel` shows `S.aiCreate.noAgent` ("This Project has no agent yet") whenever
  `agents` is empty, which is also what an unloaded list looks like. If any caller can open
  the dialog before `useProject()` has agents, it asserts something false for a frame.
- `AiCreateModal` gates on `props.open` and then passes that same always-true `open` into
  `Modal`, which gates again. Harmless, but it means the dialog can never animate out.
- Textarea auto-grow is correct: `rows={MIN_ROWS}` survives because `scrollHeight` is floored
  at `clientHeight`, and `Modal` returns `null` rather than mounting hidden, so the
  `useLayoutEffect` never measures a `display:none` box (that would have pinned the height
  at 0 until the next keystroke).
- Docs, both dictionaries and the changelog pair are all present and symmetric.

Verified on `15f872c4`: `pnpm -r build` exit 0; `pnpm typecheck` all nine packages Done;
`pnpm --filter @prismshadow/penguin-web test` 122 files / 1633 tests passed.

hiyouga:
## Follow-up — kit-level copy, from a cross-PR audit

I ran a consistency audit across all six consumers (#591–#596) reading them against each
other rather than one at a time. Three findings landed here in the kit rather than in any
one consumer; pushed in `f4175c27`.

- **`S.aiCreate.byAgent` / `editInChat` said "a new chat" while everything else said "a new
  conversation".** The two words met inside a single dialog — #593's `schedule.aiCreateDesc`,
  #595's `apps.aiDesc` and #596's `benchmark.targetAgentHint` all say "in a new conversation"
  and render directly above the by-line that said "in a new chat". zh was uniform (新对话)
  and the changelog entry already said "conversation", so English was the side that had
  drifted.
- **`S.aiCreate.chooseAgent` was just "Agent" in English** against zh's "执行的智能体". It is
  the picker's only accessible name on the three surfaces that pass `allowAgentChoice`
  (#592, #594, #596), and "Agent" does not say what the control does. Now "Agent that does
  the work". Compare #596's own `optimizerAgent`, "Optimizer agent".
- **The docs paragraph described a shape four surfaces do not have.** It promised the
  **Create** button "in those headers is split" and that **Set up manually** "opens the form
  as before". Not so for the Vault (`vault-tab.tsx:228`, a wand beside a plain Add), Models
  (`models-page.tsx:1041`, a labelled AI button with a separate Add group), Hooks
  (`hooks-tab.tsx:253`, one primary Import hook) or agent import (#594, a separate button) —
  and for the four Segmented dialogs the manual form is a segment of the same dialog, not a
  form that opens as before. #591, #594, #595 and #596 all link readers to this anchor.
  Rewritten in both languages around what is actually constant: the wand, taking whichever
  shape the header has room for.

### One thing I did not change, because it is yours to decide

**`AiCreateModal` hardcodes its footer, so the Vault's default action contradicts its own
warning.** `vault.aiAddIntro` tells the user a value typed there reaches the model provider,
the Trace and the agent's command line, and that "the safer way is to let AI create only the
key names" — and then the emphasised right-most button is **Send to agent**, which sends it
immediately. #591 cannot demote that without leaving the kit, because the three-button footer
is fixed in `ai-create-modal.tsx:82-93`.

A `primaryExit?: "send" | "edit"` prop, defaulted to `"send"` and set to `"edit"` on the Vault
surface, would let the warning and the default agree. I left it alone because it changes the
kit's API and the Vault's behaviour, and #591's reviewer separately found that the flow can
carry a live secret into `localStorage` that is cleared only on a *successful* send — so
**Edit in a new conversation** currently leaves it there indefinitely. The two are worth
deciding together. The expensive surfaces (#595's app build, #596's optimize) are fine as they
are: expense is visible and reversible, a leaked secret is not.

### Also worth knowing before this merges

The audit found real duplication that belongs here rather than in the consumers: `go(autoSend)`
plus the three-button footer exists in five copies (#591, #592, #593, #594, #596), the agent-pick
expression `picked ?? pickDefaultAgent(agents)?.agentId ?? null` in five, and the tail-contract
test in three files under three names covering 3 of the 10 tails. A single kit-owned test
iterating every tail × both dictionaries — asserting each names a skill or a fully-flagged
command, that every `penguin config` line carries `--root`, and that every command targeting an
agent carries `--agent-id` — would have caught both live defects this batch turned up
(#591's wrong data root, #593's wrong agent) before either shipped.

Verified on `f4175c27`: `penguin-web` and `penguin-docs` typecheck clean; web 122 files / 1633
tests passed; docs 7 files / 53 tests passed; `pnpm format` + `format:check` clean.

hiyouga:
## Review

Two changes landed on the branch, both requested: `primaryExit` on `AiCreateModal`, and stopping a composed prompt from outliving the draft it seeds.

### Fixed in 4f0826d4

- **`ai-create-modal.tsx:82-93` — the footer's emphasis was hardcoded, so a surface could not lead with the exit its own copy recommends.** `variant="primary"` sat on **Send to agent** unconditionally, and it was also the rightmost button. A surface that warns the user that a secret typed into the prompt reaches the model provider, the Trace on disk and the agent's command line — and that the safer path is to let AI write only the key names — then put the send-it-unread button under the pointer. `AiCreateModalProps` now takes `primaryExit?: "send" | "edit"`, defaulting to `"send"`, so every existing caller renders identically. `"edit"` moves the accent **and** the rightmost slot together onto **Edit in a new conversation**: `Cancel | [wand] Send to agent | **Edit in a new conversation**`. Moving only one of the two would emphasise one button and put another under the pointer. The wand stays welded to Send in both orders — it names that action, and a label that gains and loses its icon by position reads as two different buttons. Nothing sets the prop in this PR; the Vault surface in #591 is where it gets set.

- **`ai-bridge.ts:79` + `draft-view.tsx:596` — the composed prompt stayed in `localStorage` indefinitely and could end up parked in the sidebar.** `openAiChat` writes the prompt into `draftKey(userId, projectId)` before navigating; the draft page clears that slot only on a successful send. Concretely: open the Memory tab's edit-via-chat modal, take **Edit in a new conversation**, then navigate anywhere without sending — the prompt sits in the cache forever. Do it a second time from any surface and `parkActiveDraft` moves the first one into the persisted draft-conversation list, where a prompt nobody typed shows up as a sidebar row carrying whatever the user put in it. Fixed by marking the prompt as composed rather than typed (`DraftCache.aiPrefill`, written by `buildAiDraft`), which lets the two owners of that slot treat it as what it is:
  - `parkActiveDraft` drops it instead of parking it, leaving the slot vacated (model carry-over only) so the new-chat draft the caller lands on does not read the stale prompt back;
  - `DraftView` clears the slot on unmount when the prefill was neither edited nor sent, exactly as a successful send does;
  - the mark rides through every cache write, so a reload of the draft page still restores the prefill (the constraint the route-key `sessionStorage` marker already exists for);
  - the first `onTextChange` drops the mark — from then on the text is the user's own and is cached, parked and kept like any other draft. Ordinary draft caching is untouched.

  The unmount cleanup touches storage only, never the component refs: StrictMode runs it once right after mount, and the persist effect must be able to rewrite the slot from unchanged component state. That is load-bearing, not incidental.

### Findings not fixed

- **The mechanism does not survive past the page's own lifetime.** A tab closed (or crashed) on an untouched prefill leaves it in `localStorage`, because no effect cleanup runs on unload. It is bounded — it is still marked, so it cannot be parked, and it is cleared the next time that draft page is left — but it will be sitting in the composer on the next visit to `/chat/new` in that browser. Chasing it would need an unload handler, which is exactly the speculative "clear everything on unload" this change was asked to avoid.
- **A prefill written by an earlier build of this branch is unmarked** and keeps today's behaviour (persists, parkable). Additive optional field, no migration, no on-disk data involved — the feature is unreleased, so this only affects anyone running the branch locally.
- **`design/specs/` describes neither the "Create with AI" kit nor `autoSend`.** `06-PROTOTYPE.md` documents the draft page and parked draft conversations in detail but has nothing on this pattern, so there is no spec statement to check the kit against. Worth a design PR before this merges — not something to fix here. The one adjacent spec sentence is compatible: it says the park happens when the draft holds *typed* body text (`已输入正文`), which is precisely the distinction this change now enforces.

### Notes

- The footer is split into an exported `AiCreateFooter` so its order can be tested — `AiCreateDialog` itself needs a router and a portal and cannot be statically rendered, which is why `ai-create-panel.test.ts`'s idiom could not reach it directly.
- "Keep only the model carry-over" is now spelled in three places (`discardDraft`, `parkActiveDraft`, `dropAiPrefill`). Left alone: collapsing it would edit code this PR did not otherwise touch.
- Docs and the changelog entry updated in both languages.

Verified: `pnpm --filter @prismshadow/penguin-web typecheck` (clean); `packages/web` vitest full unit run — 123 files, 1640 tests, all passing; `packages/docs` `test/content.test.ts` (7 passing); `pnpm format` + `pnpm format:check` (clean). Not run: `pnpm -r build` and the Playwright e2e suite (shared box; no selectors or flows moved).

PR #584, opened 2026-09-02
Context panel: rank the top files beside the top tools

hiyouga:
## Summary

- `GET /api/sessions/:sessionId/context` gains `topFiles`: at most five files ranked by the context their `read_file` / `edit_file` / `write_file` calls and results occupy, each as `{ path, tokens, ops: { read, edit, write } }` — the same character heuristic and the same call↔result pairing as `topTools`. A call is keyed by the file its `file_path` resolves to against the Session's Workspace (the file tools' own resolution), so `a.ts`, `./a.ts` and the absolute spelling are one row; a file inside the Workspace is shown relative to it, any other file absolute with `~` for the home directory. A call with a missing or invalid `file_path` stays out of the file ranking (the tool refuses it too) but still counts as tool traffic. The six parts, `contextClosed` and `compactionThreshold` are unchanged.
- The context ring's panel gets a Tools / Files `Segmented` switch above its Top 5. The Files view shows the basename bold with the directory muted (full path as `title`), tiny per-op counts with the file / pencil / page-plus glyphs, and the same share columns and hover highlight the Tools view has; a context whose tools never touched a file says "No file traffic in this context" / 「本轮上下文没有文件读写」. The view picked last is kept in module state for the tab session.
- The three file glyphs move from the file-summary and memory-changes cards into `components/ui/icons.tsx`, so all three consumers draw the same marks.
- Changelog pair `changelog/unreleased/2026-09-02-context-top-files.{md,zh.md}`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/92 — `05-ARCHITECTURE.md` gains the `/context` endpoint line and a 「上下文构成」 subsection (six parts, tool and file Top 5, share-only semantics, `contextClosed`); `06-PROTOTYPE.md`'s ring paragraph now describes the panel and the Tools / Files switch, and drops the `已用 / 上限` figures that moved into the panel header in #442.

## Verification

- `pnpm typecheck` — every package passes
- `pnpm --filter @prismshadow/penguin-server exec vitest run test/context-breakdown.test.ts` — 16 passed (4 new: ranking with every spelling of one path merged and op counts, relative / `~` / absolute display, skipped bad arguments and other tools' `file_path`, cap at five)
- `pnpm --filter @prismshadow/penguin-server test` — 114 files, 1650 tests passed
- `pnpm --filter @prismshadow/penguin-web test` — 118 files, 1623 tests passed (new: file-ranking rescale, `splitFilePath`)
- Visual check against the built app (scratch `PENGUIN_HOME`, a hand-written Trace shard with file-tool traffic, Playwright at 2x): `/context` returned the five expected rows (`src/state/config.ts` with read 1 / edit 1, `README.md`, `notes/CHANGES.md` with write 1, `~/notes/todo.md`, `/etc/hosts`); the panel renders in both languages with the switch, the file rows and op counts, and the remembered view survives closing and reopening the panel.

## Notes

- The brief spelled the per-file figure `chars`; the field is `tokens`, in the same estimated-token unit as `ContextToolShare.tokens` and `total`, so the web computes both rankings' shares against one basis with one helper.
- Paths keep the platform's own separators (the server's `path` module); the web splits basename / directory on either separator. The server tests derive their expectations with `path.join` / `path.sep` / `path.resolve`, so they hold on the Windows runner.
- `packages/docs` does not describe the ring panel or the `/context` endpoint, so no docs page changed.
- No backward-compatibility handling: the response gains a field, and web and server ship together.
- Conflict hotspots: `packages/server/src/api/types.ts` (one interface and one field, inside the context block only), `strings.ts` / `strings-en.ts` (nine keys right after `contextTopToolsHint`), `components/ui/icons.tsx` (three exported constants after `INFO_ICON`), and in the design repo `05-ARCHITECTURE.md` (the session API list and one new subsection before 「Session 恢复」) and `06-PROTOTYPE.md` (one sentence in 「对话页」's composer paragraph).


hiyouga:
## Review

Two fixes pushed; the ranking logic itself is sound (the call↔result pairing through
`callFile` mirrors `callName` exactly, and every spelling of one path really does collapse
to one row because `path.resolve` runs before the key is taken).

### Fixed in `a5fd8d65`

- **`context-breakdown.ts:43` — `FILE_TOOL_OPS[tc.name]` reaches through `Object.prototype`.**
  `tc.name` is a tool name off the wire, and the Trace records a call whether or not the tool
  exists — a model that hallucinates `read_file`'s neighbour can equally hallucinate
  `toString`. Indexing an object literal typed `Record<string, …>` with that returns
  `Object.prototype.toString`, which is `!== undefined`, so the call is admitted as a file op
  and `row.ops[op] += 1` coerces the function to a string key: the row gains a
  `"function toString() { [native code] }": NaN` member that goes out over the API. Needs a
  `file_path` in the hallucinated arguments to land, which is why it is a corruption rather
  than a crash. Now a `Map`, which has no prototype chain to fall through.
- **`context-gauge.tsx:189` — the panel can be taller than the space it was placed in, with
  nothing to scroll.** `usePortalPanel` picks a direction against
  `Math.min(estimatedHeight, innerHeight * 0.7)` but the panel itself carries no height cap,
  so anything past the viewport edge is simply cut off and unreachable. This PR takes
  `PANEL_HEIGHT` from 264 to 370, which is what made it worth fixing rather than noting:
  a short window (or a window whose composer sits high) now clips the Files rows the switch
  was added to reveal. Added `maxHeight: 70vh` — the cap the hook already assumed — plus
  `overflow-y-auto`; the hook's close-on-scroll explicitly exempts the panel's own scroll
  (`use-portal-panel.ts:98`), so this does not make the panel self-dismiss.

### Notes

- `share()` collapsing the two `Math.round((x / data.total) * …)` pairs into one helper is the
  right call and I would not want it undone; note it inherits the pre-existing `data.total === 0`
  division, which is safe today only because an empty breakdown also empties both rankings.
- `splitFilePath` handles the cases that matter (`/hosts` → `/` + `hosts`, a bare name → empty
  dir, both separators). On a Windows drive root it yields `C:` rather than `C:\`, which is
  cosmetic in a muted suffix.
- `shortenHome`'s `if (home === "")` guard earns its place — `path.relative("", abs)` would
  resolve against the cwd and produce a nonsense display path — so I left it.
- The label truncates from the right, so a deep file shows its name and the *start* of its
  directory. Defensible, but the nearest parent is usually the more identifying half.
- Both dictionaries, the changelog pair and the design PR are all in place.

Verified: `pnpm -r build` exit 0; `pnpm typecheck` clean across all packages; `penguin-web test` 118 files / 1623 tests passed; `penguin-server exec vitest run test/context-breakdown.test.ts` 16/16; `pnpm format` + `format:check` clean.

PR #585, opened 2026-09-02
Session list marks conversations that still have background tasks

hiyouga:
## Summary

When a conversation has started background tools or background agents, the sidebar's Session list now says so, and keeps saying so live.

- **Server** — `SessionInfo.backgroundTasks?: { processes, subagents }`: background command sessions still running (soft-yield promotions and `run_in_background` launches) and subagent sessions that hold a `subagent_id` and are mid-round, read live from the loaded runtime's in-memory registries (no Trace scan). Present only while a count is non-zero; an unloaded Session omits it. Single GET and list rows carry it. A new user-level event `session_background { sessionId, processes, subagents }` is published to the Project's owner and members whenever the counts move (zeros included), deduped against the counts last read.
- **Core** — the seam the server needs: `BackgroundRegistry.onChange` (register / remove / eviction / reap / killAll), `ManagedSession.setExitListener` (a persistent exit listener a deliberate kill does not disarm), `CommandSessionManager.onChange` / `SubagentSessionManager.onChange`, and the `Environment` fans those plus the subagent run-state pings into one host listener: `EnvironmentInterface.setBackgroundStateListener` → `Session.onBackgroundState`. Payload-free ping; the host re-reads the registries. `setSubagentStateListener` keeps its contract — the manager's single listener slot is now installed at construction and fans out.
- **Web** — `session-activity.ts` gains a separate facet `sessionBackgroundTasks(row)` (a Session can be idle, read, and still own a dev server). The session row draws a layered-stack mark (`BACKGROUND_TASKS_ICON`, `busy` tone, title / aria-label `S.chat.backgroundTasks(n)` — "N 个后台任务" / "N background tasks") beside the running / compacting / unread glyph. The sessions store applies `session_background` in place (`setBackgroundTasks`; zero stored as absence, equal counts are a no-op). The chat header's former "running services" chip becomes the background-task pill: same count and glyph as the row, read off the row instead of the process poll; a count change also refetches the process list at once.
- **Docs** — `web-app.{en,zh}.md` (row marks, header pill), `server-api.{en,zh}.md` (`session_background`), changelog pair `2026-09-02-session-background-indicator`.

## Design

Prism-Shadow/penguin-harness-design#93 — https://github.com/Prism-Shadow/penguin-harness-design/pull/93 (05「后台运行与访问」/「服务端事件」, 06「对话页」).

## Verification

```
pnpm --config.verify-deps-before-run=false -r build                                  # exit 0
pnpm typecheck                                                                       # all packages Done
pnpm --filter @prismshadow/penguin-core exec vitest run test/background-registry.test.ts test/background-execution.test.ts   # 32 passed
pnpm --filter @prismshadow/penguin-core exec vitest run test/subagent.test.ts test/environment.test.ts                       # 68 passed
pnpm --filter @prismshadow/penguin-server test                                       # 115 files, 1650 passed
pnpm --filter @prismshadow/penguin-web test                                          # 118 files, 1630 passed
pnpm --filter @prismshadow/penguin-docs test                                         # 53 passed
pnpm format && pnpm format:check && git diff --check                                 # clean
```

New tests: core `background-registry.test.ts` (membership listener), `background-execution.test.ts` (Environment pings on a background launch / exit / kill, and on a subagent promotion / settle); server `session-background-tasks.test.ts` (field on GET and list rows, omitted at zero and when unloaded; `session_background` published only on change, zeros included, owner-only audience); web `session-activity.test.ts` (`sessionBackgroundTasks`, `BackgroundTasksMark`), `session-status-events.test.ts` (store routing of `session_background`).

## Notes

- Subagent counting: a child still inside a foreground collect window has no `subagent_id` yet and is the running Task's own work — it is not counted; a child counts once promoted (yield window expired, or `run_in_background`) and mid-round. This mirrors core's own `hasRunningBackgroundSubagents`.
- The header's "running services" count (processes only, polled) was folded into the new pill rather than kept beside it — two overlapping counts in one header would have been confusing. `S.chat.runningServices` and `STAT_ICONS.services` are removed with it; 06 was updated accordingly.
- The event rides the user channel only (like `session_state`); the open conversation reads the same store row, so no Session-channel duplicate was added.
- No backward-compatibility handling: an additive optional field and an additive event type; older clients ignore unknown event types.
- Conflict hotspots: `packages/server/src/api/types.ts` (SessionInfo + ServerEvent), `packages/web/src/state/sessions.tsx` (store + `applyUserEvent`), `packages/web/src/components/layout/sidebar.tsx` (SessionRow props + glyph slot), `packages/web/src/features/chat/chat-page.tsx` (header chips), `strings.ts` / `strings-en.ts` (`chat` section), `packages/server/src/runtime/session-manager.ts`, 05/06 specs, `web-app` and `server-api` docs.


hiyouga:
## Review

### Fixed in 86944d46

- **`packages/server/src/runtime/session-manager.ts:1599` — a discarded runtime leaves the row's mark stuck on its last count forever.** `ensureEntry` drops an idle entry *without* disposing it once the Agent's config generation has moved, and the replacement resumes with empty registries. The counts therefore fall to zero with no core ping left to carry them, so `session_background` is never published: every already-open client keeps drawing the mark (and the header pill) on a count nothing can move again, while a plain list fetch already reports none. Concrete sequence — Session idle with a `run_in_background` dev server, mark reads 1; `POST /api/projects/:p/agents/:a/plugins` (or `DELETE …/hooks/:name`, or `PUT /models`, or a completed OAuth mint) bumps the generation via `invalidateAgentRuntimes` / `invalidateProjectRuntimes`; send any message on that Session; the mark stays at 1 until the tab refetches the list. The header chip this PR replaced did not have this failure mode: it polled `GET …/processes` and self-corrected. The fix baselines the replacement entry's dedup counters on the discarded entry's and reads them once after the listener is wired, so whatever the fresh runtime actually owns goes out — self-correcting rather than a hardcoded zero. Regression test in `packages/server/test/session-manager.test.ts` (verified failing with the publish removed).

### Findings not fixed

- **`packages/server/src/runtime/session-manager.ts:1599` — that same discard orphans the background processes themselves.** The entry is deleted without `disposeRemoved`, so a `pnpm dev` promoted past its yield window keeps running with no runtime holding it: the process list empties, the Stop control goes blind, and the OS process survives until the server exits. Pre-existing (it predates this PR; the mark just made it visible), and the obvious repair — disposing on that path — would kill a user's dev server as a side effect of installing a plugin. Needs your call rather than a reviewer's.

### Notes

- **Staleness audit came back clean everywhere else.** `sweepIdle` pins on a superset of exactly what is counted (`listBackgroundCommands().some(p => p.running)` covers `processes`; `hasRunningBackgroundSubagents()` is `registry.list().some(t => t.running)`, which is precisely `running && subagentId !== null`), so idle eviction can never strand a non-zero mark. The delete/abort paths remove the row itself. A server restart drops the SSE connection and the client's `Last-Event-ID` misses the new buffer, so `sse.ts:72` sends `resync_required` and `applyUserEvent` reloads the list. `Environment.dispose` sets `bgDisposed` before `killAll`, so a Session ending is silent by construction. `replace` / `setStatus` / `setTitle` in the store all preserve the field.
- **Cost.** `sessionBackgroundTasks(row)` is O(1) per row, not O(rows × tasks). Server-side `backgroundTasksOf` is a Map hit plus two filters over registries capped at `MAX_SESSIONS`/`maxTasks`, run once per *page* row (slicing precedes `toInfo`, `session-service.ts:219`), and adds no DB query.
- **Frontend house rules hold.** Tone is picked by meaning and never spelled as a palette class (`toneInk.busy` on the row, `toneSurface.busy` on the header pill). Colour is not the only carrier: the count is in `aria-label` and `title`. The mark shares its emerald with the unread `toneDot.success` dot that can sit beside it on the same row, and they are separated by shape (a 12px layered-stack line glyph against a 6px filled dot) and by motion (neither animates; the amber hourglass and compress marks own the motion) — which is the rule as written. Both dictionaries carry `chat.backgroundTasks`, `placeholders-parity` passes, and `S.chat.runningServices` / `STAT_ICONS.services` have no remaining references anywhere in the tree.
- `BackgroundTasksMark`'s `size` prop is never passed by its only call site. Left as-is because `SessionActivityIcon` immediately below it has the same unused prop and the two must agree on the 12px box; worth removing from both at once, not from one.
- The header pill's count reaches pointer users only: it sits inside the details `<button>`, whose `aria-label={S.chat.infoPanel}` overrides all descendant text, so the `title` never becomes an accessible name. Identical to the chip it replaced and to the neighbouring `StatChip`s, so not a regression — and the sidebar mark does announce the count.
- **Design parity.** `design/specs/05` and `06` on the design repo's `main` carry neither `session_background` nor `backgroundTasks`; the design change is still open as Prism-Shadow/penguin-harness-design#93. Per CLAUDE.md design lands first, so this should not merge ahead of it.

Verified: `pnpm --config.verify-deps-before-run=false -r build` (exit 0); `pnpm -r --workspace-concurrency=2 typecheck` (19 projects, all Done); `pnpm --filter @prismshadow/penguin-core exec vitest run test/background-registry.test.ts test/background-execution.test.ts test/environment.test.ts test/subagent.test.ts` (100 passed); `pnpm --filter @prismshadow/penguin-server exec vitest run test/session-manager.test.ts test/session-background-tasks.test.ts` (45 passed); `pnpm --filter @prismshadow/penguin-web exec vitest run test/session-activity.test.ts test/session-status-events.test.ts test/placeholders-parity.test.ts test/icon-scale.test.ts` (59 passed); `pnpm format && pnpm format:check` (clean). Not verified: no browser run, so dark mode and the zh rendering of the mark were read from the tone tokens and the dictionaries rather than seen; no Playwright e2e.

hiyouga:
Manual verification feedback:

For tools invoked with `"run_in_background": true`, the toolbar should also carry a marker, matching the one on the **Session title** in the left panel. The current marker isn't very nice-looking, and the background-task icon could be easier to understand (更易读/更直观).

Suggestion: align the run_in_background toolbar indicator with the Session-title marker, and reconsider the background-task icon so it clearly communicates "running in the background".

hiyouga:
## Rework of the background-task mark (2116ae5c)

Both points from the manual-verification comment, in order.

### 1. The tool row now carries the same mark

`tool-call-card.tsx` — a call whose arguments carry `run_in_background: true` (`exec_command`
and `run_subagent` both take the flag) draws `BackgroundTasksMark` **right of its duration**,
before the flex spacer and the expand chevron. It sits there rather than beside the tool name so
it never competes with the truncating subtitle, and the row's `StatusIcon` keeps saying what the
call itself did — a backgrounded call returns at once, so its row otherwise looks like any other
settled step. Detection reuses the row's own argument seam (`isBackgroundCall`, exported next to
`previewArguments` / `headerSubtitle` / `pendingFilePayload`); the flag is last in both schemas,
so a still-streaming or malformed argument string reads as "not background" and the mark appears
with the closing brace.

**The two placements do not say the same sentence.** The sidebar's mark names a count; the tool
row's names one call. Reusing `S.chat.backgroundTasks(1)` there would have announced "1
background task", which reads as the conversation's total. New key in both dictionaries:

- `S.chat.backgroundCall` — en `"Runs in the background"`, zh `「在后台运行」`.

### 2. New glyph — an activity trace, not a layered stack

`BACKGROUND_TASKS_ICON`: `M12 3 3 8l9 5 9-5-9-5zM3 16l9 5 9-5` → `M2 12h4l3 9 6-18 3 9h4`.

A flat line with one tall beat in it — the Activity-Monitor reading of "a process is still
ticking". Why it beats the stack:

- **The stack is a noun where the state is an event.** Two layers say "layers / copies"; a trace
  says "something is happening". The mark's job is the second one.
- **It survives 12px.** The stack's two parallelograms sit ~2.5px apart at the row's size and
  merge into a smudge; a single continuous stroke with one tall beat keeps its shape.
- **It is nobody else's shape in either row.** Not the hourglass (`attention` = waiting on time
  or a person — session activity, and a pending approval on a tool row), not the compress
  chevrons, not the spinner ring (`busy` in the foreground), not the circled check / cross /
  square of a tool row's own status, not the unread dot, not the row's expand chevron. Static, so
  the two animated session marks (turning hourglass, squeezing compress) keep motion as their own
  separator.

Colour is still not the carrier: `role="img"` plus `aria-label`/`title` on both placements.

### Dead `size` prop, resolved

`BackgroundTasksMark`'s `size` was never passed, and neither was `SessionActivityIcon`'s. With a
second call site that genuinely wants a different rung:

- `BackgroundTasksMark({ label, size })` — both required, both passed at both sites. Session row:
  `ICON_SIZE.rowMark`. Tool row: `ICON_SIZE.inlineGlyph` (13), matching the `StatusIcon` already
  on that row.
- `SessionActivityIcon` loses `size` outright — both of its call sites wanted the row rung.
- `ICON_SIZE` gains `rowMark: 12`, "a trailing status mark on a list row". That role existed
  already and was being typed as a bare `12` by the session row's pin and relay arrow; those two
  now name it as well.
- `icon-scale.test.ts` gains a one-home assertion for the trace — three surfaces draw it now, and
  that is how the other asserted paths ended up hand-typed five times each.

Docs (`web-app.{en,zh}.md`) and the changelog pair updated to describe the trace and the tool-row
marker.

### Verified

`pnpm --config.verify-deps-before-run=false -r build` (exit 0) · `pnpm typecheck` (all packages
Done) · `pnpm --filter @prismshadow/penguin-web test` (118 files, 1635 passed) ·
`pnpm --filter @prismshadow/penguin-docs test` (53 passed) · `pnpm format && pnpm format:check`
(clean) · `git diff --check` (clean).

**Looked at, not just tested — but statically, not in the running app.** Per instruction the dev
server was not started (the local slot is in use and the admin claim link must not be consumed).
Instead the real components (`BackgroundTasksMark`, `SessionActivityIcon`, `StatusIcon`,
`GlyphIcon`, `Chevron`) were rendered to static markup, screenshotted with Chromium at the real
rungs — 12px on session rows, 13px on tool rows — in light and dark, and inspected pixel by pixel
at 5× nearest-neighbour zoom beside every neighbour: hourglass, compress bar, unread dot, relay
arrow, spinner ring, circled check, expand chevron. Seven candidate paths were compared that way
before this one was picked. What that does **not** confirm: crowding on a real tool row under
live data. Reasoned from the row's actual content instead — the mark is `shrink-0` and lands
after the `shrink-0` duration, so the only element that gives way under pressure is the subtitle,
which already truncates; the static render includes a long-subtitle `run_subagent` row showing
exactly that.

PR #586, opened 2026-09-02
feat(web): a floating launcher for the hidden right dock

hiyouga:
## Summary

The buttons that open the chat page's right dock live in the toolbar and are easy to never notice. This adds an AssistiveTouch-style floating launcher: while the right dock is hidden on a desktop-width window, a translucent round button rides the right edge of the conversation body; a click fans out one round entry per dock panel (every `PANEL_KINDS` kind via the shared panel meta, so a kind added later appears by itself) plus a terminal entry, each opening its panel in the right dock — the dock becomes visible and the launcher unmounts.

- `features/dock/dock-launcher.tsx`: the ball; the fan (staggered spring-ish entrance, exit animation, hover/focus name pills, `aria-label`s, ArrowUp/ArrowDown/Home/End navigation; Esc refocuses the ball; an outside press or the user scrolling folds it); the drag (pointer drag with a 4px threshold, rubberband off the edge and past the body's ends, spring snap-back on release through `lib/spring.ts`); the pending-approval amber dot (same rule as `PanelsToolbar`, named in the accessible name); reduced motion (no spring, instant show/hide — the fan's CSS animations are also cut by the global rule).
- `features/dock/dock-launcher-state.ts`: the pure decisions — visibility, bounds/clamping, ratio ⇄ top round trip, tolerant parse of the stored ratio, injectable storage, drag position, fan direction — covered by `test/dock-launcher-state.test.ts` (23 tests).
- The position is one global preference, `penguin.dock.launcherY` (a ratio of the chat body's height), classified as a browser preference in `install-scope.ts` (kept across data-root switches, like the side-panel width).
- Mount: one delimited change in `chat-page.tsx` — the message-area div becomes `relative` and hosts `<DockLauncher agentsPending={anySubagentPending} />`. That body is the region between the toolbar and the composer, so clamping to it keeps the launcher off both.
- Strings `dock.launcher*` (zh + en), fan keyframes in `styles.css`, a Web App docs subsection (en + zh), the changelog pair.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/94 — 06「对话页」gains the 右侧浮动球 paragraph.

## Verification

- `pnpm typecheck` — exit 0
- `pnpm --filter @prismshadow/penguin-web test` — 119 files, 1644 tests passed (the first run failed exactly one: `install-scope.test.ts` requires a `KEY_RULES` entry for the new key; added, rerun green)
- Visual and behavioural self-check with the built app, Playwright and the e2e mock LLM on a scratch data root (1400×860, 2×, light and dark): resting / hover / expanded / dragging captures; Esc refocuses the ball; ArrowUp from the ball reaches the nearest entry; an outside click folds the fan; choosing Trace opens the right dock and unmounts the launcher; hiding the dock brings it back; a 160px drag snaps back to the edge, stores `penguin.dock.launcherY` = 0.2456 and survives a reload; a 900px viewport shows no launcher. The driver script was deleted, not committed.

## Notes

- Mounted inside the chat body rather than against `data-dock-row`: the right edge is the same while the dock is hidden, and the vertical clamp (off the toolbar and the composer) comes from the container instead of a measurement.
- The fan folds on `wheel` / `touchmove` (the user scrolling) and on a `scroll` that moved the launcher itself (`scrollMovesAnchor`), not on any scroll — a stream auto-following a reply must not fold it.
- The launcher appears the moment the dock's `open` flag flips, riding the widening edge during the 200ms collapse, rather than waiting out the linger.
- `PANEL_RIGHT_ICON` is a third copy of the path in `panels-toolbar.tsx` and `dock-panel.tsx`, and `usePrefersReducedMotion` duplicates the private hook in `sheet.tsx`; both are candidates for a small consolidation, left alone here to keep the diff focused.
- No backward-compatibility handling: the new key is read tolerantly and nothing already on disk changes shape.
- Conflict hotspots: `chat-page.tsx` (one import, a `relative` on the message-area div and a 3-line mount), `strings.ts` / `strings-en.ts` (tail of the `dock` section), `install-scope.ts` (`KEY_RULES`), `styles.css` (after `.anim-fade`), `web-app.{en,zh}.md` (new subsection before Files Panel), `06-PROTOTYPE.md` (a paragraph after the dock paragraph).


hiyouga:
## Review

No defects. The state module is pure and covered, the fan's dismissal set is complete (outside
press, capture-phase Esc that stops propagation, `wheel`/`touchmove` for a real user scroll,
and `scrollMovesAnchor` for a scroll that actually moved the ball), and the arrow-key column
is built in visual order on both fan directions — `[...entries, ball]` when it opens up,
`[ball, ...entries]` when it opens down — which is easy to get backwards and is not.

### Fixed in `2c4dfdd8`

This is the consolidation the PR's own Notes deferred; both items were flagged there as
follow-ups, and neither is a behaviour change.

- **`PANEL_RIGHT_ICON` was a third hand-typed copy of the same path,** and `PANEL_BOTTOM_ICON`
  a second. Both now live in `components/ui/icons.tsx` and are imported by `panels-toolbar.tsx`,
  `dock-panel.tsx` and `dock-launcher.tsx`. `test/icon-scale.test.ts` exists precisely for this
  decay — its header says the three paths it guards "each used to exist in five or six
  hand-typed copies" — so I added the two matching assertions rather than leaving the new
  single home unenforced.
- **`usePrefersReducedMotion` was duplicated verbatim** in `sheet.tsx` and `dock-launcher.tsx`.
  Extracted to `components/ui/use-reduced-motion.ts` (beside `use-portal-panel.ts`, the one
  existing hook-module precedent) and imported by both. The doc comment says why the hook has
  to exist at all when `styles.css` already cuts CSS animation: it is for the springs driven
  frame-by-frame in JS, which that rule cannot reach.

### Notes

- `dragPosition` rubberbands **both** directions, so a rightward pull moves the ball up to
  ~40px past the chat body's right edge. Its container is `relative` with visible overflow, so
  where that lands depends on which ancestor clips; I could not confirm a horizontal scrollbar
  by reading alone, and it self-corrects on release. Worth one drag to the right on a narrow
  desktop window before merge. Clamping `x` to `≤ 0` would also match the module's own
  coordinate doc ("0 on the edge, negative when pulled into the conversation").
- The body is found as `rootRef.current?.parentElement`. It is correct today and the comment
  says so, but it makes the measurement silently wrong for anyone who later wraps the mount —
  a `data-` marker on the body would be sturdier than a parent hop.
- The launcher answers the right question (the toolbar toggles are genuinely easy to miss) and
  the accessible surface is complete: `aria-label` on the ball and every entry, `aria-expanded`,
  a labelled `role="group"` fan, the pending state folded into the ball's name rather than
  carried by the amber dot alone.

Verified: `pnpm -r build` exit 0; `pnpm typecheck` clean across all packages; `penguin-web test` 119 files / 1646 tests passed (1644 before, +2 for the new one-home assertions); `pnpm format` + `format:check` clean.

PR #588, opened 2026-09-02
Fold the image tools into read_file and align input_command's timeout with exec_command

hiyouga:
## Summary

- `read_file` now reads images too, and the `read_image` / `describe_image` tools are removed. A png/jpeg/gif/webp file (magic number, then extension) or an http(s) URL in `file_path` takes the image branch: a Session model that accepts images gets the image back via `tool_call_output.images` with a one-line `image/png, 123.4 kB` note (exactly what `read_image` returned); a text-only model gets the Project's `vision_model` answering the new optional `prompt` argument (default: a detailed description), streamed as the tool output, nothing of that one-off request leaking into the parent stream (exactly what `describe_image` did). No `vision_model` configured still ends `fatal` with the same explanation asking the model to have the user pick one. Text files are unchanged (numbered window, offset/limit, scan cap, secret-store refusal); non-image binaries are still rejected with advice.
- The seam: `EnvironmentServices.visionDescriber`, which `agent.ts` already injects exactly when the session model is text-only (with `modelId: null` when no vision model is configured). Its presence is what tells `read_file` to describe instead of return; absent, the model views images. No new fields, no global state. The shared loader moved from `read-image.ts` into `environment/tools/image-source.ts`; `read-image.ts` / `describe-image.ts` and their tests are deleted.
- Default config: one `read_file` entry (no `forModel`), image-aware description, `prompt` parameter, `timeoutMs: 60000`. The two image entries and their registry factories are gone. `forModel` stays a config feature; no built-in entry uses it (said so in the comments and docs).
- Timeouts: `input_command` `timeoutMs` 130000 → 120000 (aligned with `exec_command`); `DEFAULT_EMPTY_POLL_YIELD_MS` 120000 → 110000 so the default empty poll stays under the timeout minus the clamp margin, with the tool description and docs updated to match.
- Kernel version → `2026-09-02` (tools tab re-pinned, old hash appended to the superseded list, generation note added).
- Consumers: web tool card previews old Traces' `read_image` / `describe_image` calls by their `source` argument (small `LEGACY_IMAGE_TOOLS` mapping, display-only); web strings (zh/en), chat-input / models-page / trace-row comments, CLI `--vision` help (zh/en), server comments, bilingual docs (`tools`, `configuration`, `models`, `web-app`, `agent-loop`, `interfaces`, `omni-message`).
- Changelog: `2026-09-02-read-file-images.{md,zh.md}` plus the dedicated compat pair `2026-09-02-backward-compatibility-read-file-images.{md,zh.md}`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/95 — 05「文件工具」absorbs the former「读图工具（多模态）」section (the input-image / attachment paragraphs move under「对话输入的图片与文件附件」), the command-session paragraph states the 120 s timeouts and the 110 s empty-poll default, 06「模型配置页」and 08 are updated.

## Verification

- `pnpm --filter @prismshadow/penguin-core build` — ok.
- `pnpm --filter @prismshadow/penguin-core test` — 48 files passed, 1 skipped; 1063 tests passed, 5 skipped.
- `pnpm -r build` && `pnpm typecheck` — both exit 0.
- `pnpm --filter @prismshadow/penguin-web test` — 118 files, 1622 tests passed.
- `pnpm --filter @prismshadow/penguin-server test` — 114 files, 1646 tests passed (run alone).
- `pnpm --filter @prismshadow/penguin-cli test` — 29 files, 385 tests passed.
- `pnpm --filter @prismshadow/penguin-docs test` — 7 files, 53 tests passed.
- `pnpm format` + `pnpm format:check` + `git diff --check` — all clean.

## Notes

- **Backward compatibility (please review):** existing Agents' `system_config.yaml` carry `read_image` / `describe_image` entries and the old `read_file` entry. Chosen: the `kill_command` precedent — stored entries whose name has no factory are skipped at assembly (already the registry's behavior; a new Environment test pins it), a model calling the old name gets the unknown-tool reply, and nothing on disk is rewritten. The stored `read_file` entry assembles onto the new implementation, so image reading already works there; the model is only told the old description (no `prompt`, 30 s timeout) until the kernel update, which delivers the new entry to an unedited tools tab (customized tabs are kept, as always — restore defaults or hand-edit). Alternatives not taken: a load-time rewrite of `tools.builtin`, or keeping the old names as aliases for a few releases. The only compat code is the web card's `LEGACY_IMAGE_TOOLS` preview mapping (display-only, no removal scheduled). Details in the compat changelog pair.
- Decision: the "does the session model accept images" fact reaches the tool through the describer's presence rather than a new boolean — `agent.ts` already injects the describer iff `vision === false`, including the `modelId: null` placeholder for the unconfigured case, so the seam already encodes all three states. A direct SDK embedder that builds an `Environment` for a text-only model without injecting a describer gets image content returned (the same situation such an embedder was in before with a `read_image` entry).
- Blog posts under `packages/landing/content/blog/` still mention the old tools; they are dated records and were left as-is.
- Conflict hotspots: `packages/core/src/state/kernel-history.ts` (any sibling that changes defaults bumps the same version/hash), `strings.ts` / `strings-en.ts`, `default-config.ts`, `05-ARCHITECTURE.md` / `06-PROTOTYPE.md` / `08-CHANGELOG.md`, `packages/docs/content/tools.{en,zh}.md`.


hiyouga:
## Review

### Fixed in cb6e1b90

- **`packages/docs/content/tools.en.md:127` / `tools.zh.md:125` — the docs still document the empty-poll default as 120000 after this PR moved it to 110000.** `DEFAULT_EMPTY_POLL_YIELD_MS` is now `110_000` and the `yield_time_ms` description rewritten in `default-config.ts:774` says 110000, but the `input_command` argument block on both language pages was missed. A reader deriving the yield ceiling from the docs gets the pre-PR number against the new 120000 timeout. Both pages now say 110000.
- **`packages/cli/src/tool-render.ts:344` and `packages/web/src/features/chat/tool-call-card.tsx:166` — the new `prompt` argument never reaches the approval payload.** Both renderers enumerate `read_file`'s arguments explicitly (`file_path`, `offset`, `limit`); `prompt` is new in this PR's schema and was not added. Under always-ask approval, `read_file {"file_path":"shot.png","prompt":"…"}` shows the user only the path, while the string actually sent to the vision model is invisible — and `renderFileToolApprovalPayload`'s own contract is "the user must see what they are approving, not just the file path". Before this PR that text travelled as `describe_image`'s `prompt`, which is not a file tool, so the raw-arguments fallback displayed it; folding the tool in silently dropped it. Both now push `prompt`, with an assertion added to each of the two existing tests.
- **`plugins/agent-development/skills/penguin-sdk/SKILL.md:115` — a shipped skill still describes "the built-in image tools" reading an `[attached image: <path>]` line through `vision_model`.** There are no image tools any more. It now names `read_file`. This was the only live prompt/skill/plugin artifact the removal left behind.
- **`packages/core/src/environment/tools/image-source.ts:38,58` — `sniffImageMime` / `imageMimeFromExt` gained `export` in the move with no consumer.** They were module-private `sniffMime` / `mimeFromExt` in `read-image.ts`; nothing outside the file imports them and `packages/core` publishes no subpath export for it. Un-exported.

### Findings not fixed

- **`packages/core/src/environment/tools/image-source.ts:139-152` — the URL branch buffers the whole response before the size check.** `bytes = Buffer.from(await res.arrayBuffer())` is unconditional, and the early `content-length` guard cannot fire when the header is absent or wrong: `Number(res.headers.get("content-length") ?? "")` is `0` for a missing header, which is finite and under the cap. `read_file("https://host/x.png")` against a chunked response of arbitrary length buffers it whole into the process before `bytes.length > MAX_IMAGE_BYTES` rejects it. Pre-existing — moved verbatim from `read-image.ts` — and the fix is a streamed, byte-capped body read, i.e. hardening beyond this PR's scope. Left for your call on whether it belongs here or in its own change.
- **`packages/core/src/state/default-config.ts:641` — `read_file`'s 60000 ms timeout is below the 90000 ms `describe_image` carried**, and the text-only image branch is exactly the operation that budget was widened for (the removed entry's comment: "Includes one vision-model request, so the timeout is slightly wider than plain image reading"). A vision model that answers in 70 s now ends `[tool timeout: exceeded 60000ms]` where it previously completed. Left it: the value is a kernel default, so changing it re-hashes the tools tab and moves `KERNEL_VERSION` again — your call, not a review edit.
- **Backward compatibility, needs your decision: a stored `read_file` entry keeps `timeoutMs: 30000`, and that is now the budget for a whole vision-model request.** For a text-only Agent that never runs the kernel update, an image read gets 30 s where `describe_image` gave it 90 s. The compat entry records the 30000 ms figure but not that it now has to cover a round trip to the vision model. Per CLAUDE.md I am not proposing a strategy for stored config — flagging it as part of the compat decision you already asked for review on.
- **`packages/web/src/features/chat/tool-call-card.tsx:55` — `LEGACY_IMAGE_TOOLS` has no removal owner.** The compat entry says "nobody is scheduled to remove it"; CLAUDE.md asks that every compat carry a stated lifetime and who clears it.

### Notes

- Repo-wide sweep for `read_image` / `describe_image` / `read-image` / `describe-image` / `readImage` / `describeImage` (all text files, excluding `node_modules`/`dist`/`design`): 27 files, all accounted for. Deliberate — the `LEGACY_IMAGE_TOOLS` display mapping and its tests, the registry/kernel-history/docs notes explaining the skip, the new `environment.test.ts` case, this PR's own changelog. Frozen — `changelog/0.2.3` and `0.2.4`. Dated records — the eight `packages/landing/content/blog/*` files you called out. The three hits in `read-file.ts` are the internal `readImageSource` function, not the tool name. After the SKILL.md fix, no prompt, skill, plugin, built-in agent, example or fixture names a tool that no longer exists.
- Old Traces do not break, on either path. `trace/resume.ts` pairs on `tool_call_id` and never looks a tool name up, so a Trace carrying `read_image` calls replays structurally unchanged; rendering goes through the generic `tool_call` / `tool_call_output` rows plus your `source` preview mapping; and a *new* call by an old name gets `Unknown tool: read_image` from `environment.ts:518`, which the added test pins. No compat decision needed here.
- `input_command`'s timeout checks out end to end: milliseconds throughout (`timeoutMs`, `yield_time_ms`, `DEFAULT_EMPTY_POLL_YIELD_MS`), default 120000 matching `exec_command`, `clampYield` ceiling `timeoutMs - 1000 = 119000` which sits above the 110000 default so a default-length poll returns on its own. On timeout, `environment.ts:729`'s `!timedOut &&` guard stops the tool's self-reported `aborted` from being read as a user interruption, so it finalizes `fatal` with `[tool timeout: exceeded 120000ms]` and keeps the output already streamed. The child is deliberately not killed and stays in `CommandSessionManager`, so the model can poll the same `process_id` again; `collect`'s abort listener is removed in a `finally`, so no listener leak.
- Detection order behaves as documented (magic bytes -> content-type for URLs -> extension). Two edges worth knowing, neither a regression: a non-image with an image extension still takes the image branch and goes out under that mime, same as `read_image` did; and a 0-byte `foo.png` reports `is an empty file (0 lines)` from the text branch's `size === 0` check rather than the loader's `Image … is empty.`
- The compat entry is a second `2026-09-02-*backward-compatibility*` pair alongside the one already in `unreleased/` from #542. CLAUDE.md describes one file per batch; keeping them separate avoids a conflict, so this reads deliberate — confirm that is what you intended.

Verified: `pnpm -r build` (exit 0), `pnpm typecheck` (exit 0), `pnpm format` + `pnpm format:check` (clean), `vitest run test/read-file-images.test.ts test/file-tools.test.ts test/environment.test.ts test/state.test.ts test/background-execution.test.ts` in `packages/core` (213 passed, 2 skipped), `test/tool-render.test.ts` in `packages/cli` (19 passed, `FORCE_COLOR=0`), `test/tool-call-preview.test.ts test/stream-model.test.ts` in `packages/web` (129 passed), `pnpm --filter @prismshadow/penguin-docs test` (53 passed). Not run: the full server and web suites, and the web e2e specs.

hiyouga:
## Changelog: the stored timeout, and one field still missing

### Fixed in 95ce33c8

- **`changelog/unreleased/2026-09-02-backward-compatibility-read-file-images.{md,zh.md}` — the entry records the stored `timeoutMs: 30000` but not what it now costs.** An Agent from before this release keeps `read_file` at the old text-only 30000 ms until its kernel is updated (`packages/core/src/state/default-config.ts:641` sets the new default to 60000; the removed `describe_image` had 90000). With image reading folded in, those 30 seconds are now the budget for a whole vision request — a third of what the dedicated tool had. The `## Compatibility` section said only "no action is required to keep working", which is true of the tool name and false of a slow image read. It now states the symptom (the read fails on timeout, which reads as a model fault rather than a configuration one) and the remedy (raise the `read_file` entry's `timeoutMs` in the Agent's tools tab). Both halves, one sentence each, no new sections.

No code or stored config was changed: this is the "state the consequence" option, not a migration.

### Finding not fixed — needs your input

- **`## Old Traces` says the `LEGACY_IMAGE_TOOLS` mapping in `tool-call-card.tsx` "can go once Traces from before 2026-09-02 no longer need rendering; nobody is scheduled to remove it".** `CLAUDE.md` (向后兼容) requires every compatibility burden to say how long it stays and who clears it when, precisely so it does not default to permanent. The entry currently records the opposite. It needs a removal version and an owner — I do not have either and will not invent them, so the sentence is unchanged.

Verified: `pnpm format` + `pnpm format:check` clean in the worktree; the pair stays section-for-section symmetric (4 headings each, same order). Markdown only — no build or test run.

PR #96, opened 2026-07-27
docs(blog): 0.1.4 — Windows support, goal mode, and the agents panel

hiyouga:
## Summary

- New bilingual blog post `penguinharness-0-1-4` (`.en.md` / `.zh.md`, `date: 2026-07-27`) announcing 0.1.4: **Windows support** (the `irm https://penguin.ooo/install.ps1 | iex` one-liner, shell selection Git-Bash → pwsh → powershell with `PENGUIN_SHELL` override and the `Shell:` line, the win-x64 bundled-Node package, the `ci-windows` job, honest caveats), **goal mode** (the round loop, the GOAL.yaml protocol, budgets, all four surfaces), and the **agents panel** (bar rows, the call graph with live elapsed, task-scoped visibility, historical topology) — plus an "Also in 0.1.4" paragraph (live-tail refresh survival, live header stats, Trace import/export, in-app version/update) linking the v0.1.4 release notes, and closing install commands.
- 0.1.4 ships 0.1.3's feature set and is the first of the two to reach npm, so the install section names it as the version to install — one clause, not a story.
- **Eight real product screenshots**, hosted in the community repo like every other blog image (`blog-assets/{agents-panel,goal-mode}-{zh,en}-{light,dark}.webp`, 1920×1350, all verified serving). Nothing binary lands in this repo: the posts keep the portable `/blog-assets/<name>.webp` paths that #98's `MdImage` adapter resolves at render time — the theme-swapped pairs use exactly the raw `` form #98's renderer test pins.
- The capture pipeline that produced them is committed as `packages/landing/scripts/capture-blog-013-shots.mjs` (`pnpm --filter @prismshadow/penguin-landing blog-013-shots`), reusing the capture-shots.mjs machinery: a scripted mock LLM (Anthropic SSE + OpenAI chat-completions SSE), the built Web server on a temp data root, and real tool execution in staged Workspaces — the analyst's scoring script and the goal loop's check suite genuinely run, so every tool card streams real output. Both themes are captured concurrently inside the ~60s exec/subagent yield window so every call-graph node is genuinely live in frame. Output goes to the gitignored `packages/landing/.blog-assets/` staging dir for upload to the community repo, like the other capture scripts after #98.
- `packages/landing/test/blog.test.ts` updated for the new post (12 posts; news-category ordering).

## Verification

[...]

PR #589, opened 2026-09-02
Files panel: directory tree beside the preview, in-place text editing, drop to upload

hiyouga:
## Summary

The Workspace files panel becomes a two-pane browser: a directory tree on the left, the selected file's preview on the right, with plain-text editing of text files and drag-and-drop upload from the desktop.

- **Tree + preview.** Directories list their contents the first time they open (`listWorkspaceFiles` per directory); expand/collapse chevrons, folder/file glyphs, the selected file highlighted, arrow-key navigation (up/down move, right opens or steps in, left closes or steps out). Breadcrumbs of the current directory sit in the toolbar above both panes. A toolbar toggle hides/shows the tree — shown by default, one global localStorage key `penguin.files.treeVisible` (tolerant parse, classified `browser` in `install-scope.ts`). The panel measures itself with a ResizeObserver and below 480px falls back to the single-column drill-down (tree until a file is chosen, then the preview with Back). Message file cards still land on their file: the tree opens down to it.
- **Preview.** Markdown/HTML rendered + source toggle, text highlighted, PDF embedded as before; images now zoom on click (`ZoomableImage`); a file whose extension says nothing (`Makefile`, `LICENSE`) is sniffed client-side and shown as text when its first bytes look like text (no NUL, few control bytes, valid UTF-8). Text reads stream and stop at the 256KB cap instead of downloading the whole file.
- **Plain-text editing.** `Edit` swaps the preview for a monospace `<textarea>` (no rich editor); `Save` (also Ctrl+S / Cmd+S) confirms with the app's standard save card, then `PUT /api/sessions/:id/files/content` with the UTF-8 text as base64, refreshing the row's size/mtime; `Cancel` returns to the preview. A truncated preview stays read-only (saving it would overwrite the whole file with the cut text); a save over the 14MB write limit is refused client-side with a message. The settled-turn re-read skips a file that is being edited.
- **Unsaved-changes guard.** A confirm modal before switching file (tree, breadcrumb in narrow mode, message-card open request, Markdown link) and before the panel closes: a small `features/dock/close-guard.ts` registry that `dock-panel.tsx` (tab ×, hide ×) and `panels-toolbar.tsx` (dock toggle) consult before removing/hiding. `beforeunload` covers reload/tab close. Unmount paths no dialog can intercept (a tab dragged to the other edge, a Session switched from the sidebar — the app uses `BrowserRouter`, so there is no route blocker) keep the draft in a module-level stash and reopen the editor when the same file is opened again.
- **Drag-and-drop upload.** The panel is its own drop region, reusing `isFileDrag` / `dropRegionAction` from `lib/file-drop.ts`; a dashed frame names the target directory, a hovered folder row is highlighted and receives the files, anywhere else means the current directory. Picks and drops share one path: `splitBySize` against the 14MB per-file ceiling (mirrored from the server's `MAX_UPLOAD_BYTES`) before anything is read, overwrite confirmation against the target directory's real listing (loaded on demand), per-file progress in the toolbar, success/error toasts, and the first uploaded file opens afterwards. The app-shell guard keeps cancelling drops elsewhere.
- **Tests.** `test/workspace-tree.test.ts` (tree rows, expand/insert/sort, keyboard step, drop target, narrow layout, text detection, preference parse, unsaved decision) and `test/close-guard.test.ts`.
- **Docs.** `packages/docs/content/web-app.{en,zh}.md` Files Panel section rewritten; the chat-area drop paragraph names the Files panel as the other drop target. Changelog pair `changelog/unreleased/2026-09-02-workspace-tree-editor.{md,zh.md}`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/96 (`docs/workspace-tree-editor`: 05「Workspace 文件预览」gains the panel shape and the editing/upload mechanics; the 06 chat-page files-panel bullet is rewritten).

## Verification

- `pnpm --config.verify-deps-before-run=false --filter @prismshadow/penguin-web... build` — core, server and web built (web bundle OK).
- `pnpm --filter @prismshadow/penguin-web typecheck` — exit 0.
- `pnpm --filter @prismshadow/penguin-web test` — 120 files, 1646 tests passed (25 new).
- `pnpm --filter @prismshadow/penguin-docs test` — 7 files, 53 tests passed.
- `pnpm format && pnpm format:check` — clean; `git diff --check` — clean.
- Not run: Playwright e2e (`chat.spec.mjs` no longer clicks "返回列表" in the wide layout — the tree stays beside the preview; `mobile-dock.spec.mjs` and `files-switch.spec.mjs` flows still hold: the merged bottom surface is narrow, so it drills down).

## Notes

- Save confirms first, following the design's "every server-side file write confirms" rule; a plain Ctrl+S opens the same card.
- Tab inside the editor is not intercepted (it keeps moving keyboard focus); the brief marked two-space Tab as optional.
- The server does not sniff content types, so the "content sniffed as text" part of the brief is done client-side on the first 8KB.
- `workspace-select.tsx` has no open-file entry point (it picks a directory for a draft); the only external open request is the message file card, which is covered.
- No on-disk data or config is touched; the one new localStorage key is a browser preference, so there is no backward-compatibility file.
- Conflict hotspots: `strings.ts` / `strings-en.ts` (`files` section), `dock-panel.tsx`, `panels-toolbar.tsx`, `install-scope.ts` (`KEY_RULES`), `message-files-card.tsx` (exports `FILE_ICON`), docs `web-app.{en,zh}.md`, specs 05/06.


hiyouga:
## Review

Reviewed on `31dbe9d1`. Line numbers below are that commit's.

### Fixed in 31dbe9d1

- **`workspace-browser.tsx:352-361` — a Session switch strands the upload and save chrome, and re-points a staged overwrite at the new Session.** `doUpload` and `save` clear `uploading` / `saving` only `if (sessionIdRef.current === sid)` (`:865`, `:829`), and the render-phase reset did not clear them either. Drop three files, then click another Session in the sidebar before they land: `uploading` stays non-null forever, so the toolbar reads "Uploading 1/3…" against the wrong Session, the picker stays `disabled`, and `stageUpload`'s `if (uploadingRef.current !== null) return` (`:882`) silently drops every later pick or drop for the life of the mount. Same shape for `saving` — switch Session mid-save and `requestSave` returns early on every later save. `pendingUpload` was worse than stranded: it survived the reset holding the *old* Session's `files` and `dir`, while `doUpload` reads `sessionId` from the render it is called in, so answering the overwrite dialog after the switch writes the old Session's files into the new Session's Workspace. The reset block now clears all three. (`main`'s version cleared `uploading` unconditionally, so the stranded flags came in with the sid guards.)
- **`workspace-browser.tsx:159-163` — `fetchTextPreview` read the whole file into memory whenever `res.body` was absent.** The `reader === undefined` branch called `res.arrayBuffer()`, which defeats the 256KB cap the function exists to enforce: the bytes are already resident by the time `truncated` is computed, so a 2GB log previewed by downloading 2GB. It now throws — `previewPath`'s `catch` already renders a failed read as an "unsupported" preview. Dropping that branch also removes the post-loop `looksLikeText(chunks[0])`, which re-sniffed bytes the streaming path had sniffed one iteration earlier.
- **`workspace-browser.tsx:213-233, 948` — dropping a folder on the panel failed with "Unknown error".** `dataTransfer.files` carries a dropped directory as a pseudo-file, so it passed `splitBySize` (size 0), reached the overwrite confirmation if its name clashed with an existing directory, and then died in `blobToBase64`'s `reader.onerror` with a plain `Error` — which `apiErrorText` maps to `S.common.unknownError`, since it is not an `ApiError`. `pickDroppedFiles` reads `DataTransferItem.webkitGetAsEntry()` and names the folders in a skip toast (`files.folderDropSkipped`, added to both dictionaries), still uploading whatever real files came with them.
- **`workspace-tree.ts` `flattenTree` / `workspace-tree-view.tsx:130-136` — the tree is rendered flat but stated only `aria-level`.** With no per-level `role="group"` in the DOM, `aria-posinset` and `aria-setsize` are the only things that let a screen reader say "3 of 7 at this level"; without them the whole tree announces as one undifferentiated list. `flattenTree` now emits `posInSet` / `setSize` per directory and the rows carry them (assertion added in `workspace-tree.test.ts`). The wrapper `<div>` between `role="tree"` and `role="treeitem"` became a `Fragment` — a generic box there breaks the tree's ownership of its items.
- **`workspace-browser.tsx:1314-1318` — the tree toggle said its state twice, in conflict.** It carried both `aria-pressed={treeVisible}` and an `aria-label` that swapped Show/Hide, which announces as "Hide file tree, pressed". `sidebar.tsx`'s `GroupPinButton` spells the rule out: the name stays static, `aria-pressed` alone carries the state, only `title` may swap. Also added `aria-pressed` to the Rendered/Source segmented control (`:999`), whose selection was carried by background colour alone.

### Findings not fixed

- **`workspace-browser.tsx:792-812` (`save`) — a lost-update window with no precondition.** `PUT /api/sessions/:id/files/content` takes only the bytes: there is no `If-Match`, mtime or size check on either side (`grep` for `ifMatch|If-Match|expectedMtime|etag|unmodified` across `workspace-files-service.ts`, `routes/sessions.ts` and `api/endpoints.ts` returns nothing). Sequence: open `notes.md`, click **Edit**, the Agent rewrites `notes.md` during the turn, click **Save** — the Agent's version is gone with no warning, and the settled-turn re-read deliberately skips the file being edited (`:656`), so nothing on screen ever hinted the file had moved underneath. The material for a check exists already (`POST /files/stat` returns `mtimeMs`, and `WorkspaceFileEntry.mtime` rides every listing), but a correct fix belongs on the write endpoint — an expected-mtime parameter that 409s — which is a server change plus a `design/specs/` change, so it is your call, not mine to invent.
- **`workspace-browser.tsx:843-849` (`doUpload`) — a Session switch abandons the rest of a batch silently.** The `if (sessionIdRef.current !== sid) return` inside the loop (`:847`) bails before `toastSuccess`, so a five-file upload interrupted at file two leaves three files unsent and says nothing — even though every request was already addressed to `sid` and would have succeeded. Letting the loop finish and skipping only the UI updates is a behaviour choice (which Session gets the toast), so I left it.
- **`workspace-browser.tsx:882` (`stageUpload`) — a drop during an upload does nothing at all.** The picker is disabled and shows progress, but the panel still accepts drops and `if (uploadingRef.current !== null) return` swallows them with no toast and no change to the overlay. One line and a string, but whether a second batch should queue instead is your call.
- **`workspace-tree-view.tsx:28` — the indent is unclamped.** `paddingLeft: 6 + depth * INDENT_PX` reaches the tree pane's 168px floor at depth 12, past which the name is squeezed to zero width behind a horizontal scrollbar. `Math.min(depth, N)` is the usual answer.
- **Design PR still open.** `design/specs/05-ARCHITECTURE.md` §「Workspace 文件预览」 describes preview and its origin isolation only — no tree, no in-place editing, no drop upload — so on `main` today the code would contradict the spec. Prism-Shadow/penguin-harness-design#96 covers it and should land first.

### Notes

- Server-side path confinement for everything this panel calls is sound and untouched by this PR: `resolveRead` realpaths the whole path and does IO on the canonical form (no check-then-use gap), `resolveWriteParent` probes up to the nearest existing ancestor and `assertInside`s its canonical form before `mkdir -p`, `write` opens the final segment with `O_NOFOLLOW` (plus an `lstat` symlink refusal on win32, where the flag is erased), and `lexicalTarget` rejects NUL bytes and anything resolving outside. `WORKSPACE_UPLOAD_LIMIT_MB = 14` matches `MAX_UPLOAD_BYTES` exactly, and `splitBySize`'s `>` matches the server's own comparison. Worth flagging only that the client mirror is a hardcoded constant while `upload-limits.ts` argues against exactly that — it is right today because `MAX_UPLOAD_BYTES` is fixed, unlike the admin-settable attachment ceilings, and would silently drift the day it becomes configurable.
- Every open directory dims to `opacity-60` on each settled turn: `refreshAll` re-adds it to `loadingDirs` even when the listing is only being re-read, and the row's transition is `transition-colors`, so the opacity change is instant. A visible flicker at every turn end for a tree that usually did not change.
- i18n is complete and symmetric (both dictionaries; `files.uploaded` removed from both with no remaining reference), the changelog pair is symmetric with the fixed metadata block, and `penguin.files.treeVisible` is classified in `KEY_RULES`, which `install-scope.test.ts` enforces against the source.

Verified on 31dbe9d1: `pnpm --filter "@prismshadow/penguin-web..." build` (core + server + web, exit 0), `pnpm --filter @prismshadow/penguin-web typecheck` (exit 0), `pnpm --filter @prismshadow/penguin-web test` (120 files, 1647 tests passed — 1646 plus the one added here), `pnpm format && pnpm format:check` (clean). Playwright e2e not run.

hiyouga:
## Lost-update guard for in-place editing — f97cbb60

`PUT …/files/content` had no write precondition on either side. Open a file in the panel,
edit it, let the Agent rewrite the same path during its turn, hit Save: the Agent's version
was gone with no warning — and `workspace-browser.tsx:656` deliberately skipped re-reading
the file being edited, so nothing on screen had hinted it moved.

### Server

- **`GET …/files/content` returns `ETag: W/"<size>-<mtime>"`** — the same weak validator
  `app.ts:754` already mints for static assets. Not a cache validator (`Cache-Control:
  no-store` means nothing ever revalidates); it is the version of the bytes in that response.
  `WorkspaceFileContent` gains `version`, minted from the stat `read()` already does. That
  stat runs *before* the bytes, and the order is now load-bearing: a rewrite in between tags
  fresh bytes with the older marker, costing at worst one needless conflict. Read-then-stat
  would fail the other way — old bytes under a new marker — which is the exact write this
  exists to refuse.
- **`PUT …/files/content` takes `ifVersion`** in the body (`FilesWriteRequest`, validated with
  the existing `optionalString`), compared **on the open write handle**: the conditional open
  withholds `O_TRUNC`, `fstat`s the handle, compares, then truncates. A handle names an inode,
  so what is stat'd is exactly what is about to be truncated — no window in between.
  Mismatch → `409 file_changed` through the route's ordinary `HttpError` path, nothing
  written.
- **New file vs. changed file:** the *absence of the field* is the distinction, not a sentinel
  value. No marker means the caller read no version, so the write is the unconditional
  create-or-overwrite uploads and drops already relied on — untouched. A marker present means
  the file must exist and match, so `O_CREAT` is left out of that open: a deleted file arrives
  as `ENOENT` and folds into the same 409 (deletion is a change), and a refused write cannot
  leave an empty file behind.

### Web

- The editor carries the version it opened with (`EditorState.version`, read off the same
  response as the text — `fetchTextPreview` now throws if that header is missing rather than
  quietly saving with no precondition) and sends it on save.
- **409 is a question, not an error toast:** a confirm dialog names the file, says it was
  rewritten after it was opened, and offers **Overwrite** (the same write with no marker) or
  Cancel. The draft survives both answers; nothing is retried automatically. Both dictionaries.
- **`workspace-browser.tsx:656` — the skip stays, the silence does not.** Re-reading the
  edited file would still put the Agent's text under the user's hands, so its text is left
  alone; the settled turn now fetches just the version (body cancelled at the headers) and
  marks the editor **Changed on disk** when it moved. The precondition makes Save safe; this
  is what makes the user aware before they get there. A failed probe says nothing — the save's
  own precondition is the guarantee.
- Once known-stale, Save goes straight to the conflict dialog: the plain "the file will be
  overwritten" confirmation understates it, and sending the stale marker first only earns the
  same dialog a round trip later.
- After a successful save the preview drops its `version` instead of keeping the pre-save
  value — the write's resulting version is unknown to the client, and keeping the old one
  would make an immediate second Edit + Save 409 for no reason. `startEdit` re-reads when it
  finds none.

### Design

Spec change landed first, per CLAUDE.md: **penguin-harness-design `a2fd0e4`** on
`docs/workspace-tree-editor` (PR #96) — the route table line and a new paragraph in
`05-ARCHITECTURE.md`'s Files panel section covering the marker, the 409, the field-absence
rule and how the panel surfaces the conflict. This PR depends on that commit.

### Verification

- `pnpm typecheck` — all 7 packages pass.
- `pnpm -r build` — pass (run because `api/types.ts` is a cross-package type surface).
- `packages/server` `test/workspace-files.test.ts` — 18 pass, including 3 new cases (service:
  stale marker refused with the file untouched, then the fresh marker writes and truncates;
  no-marker create plus a marker on a deleted file refused leaving nothing behind; route:
  ETag shape, 409 body, an accepted current marker, an unconditional upload, and a non-string
  `ifVersion` as 400). Stubbing the precondition out (`conditional = false`) fails exactly
  those 3 and nothing else — checked, then restored.
- `packages/web` `test/workspace-tree.test.ts` + `test/close-guard.test.ts` — 26 pass.

Not verified: no browser run — the conflict dialog, the "Changed on disk" marker and the
`fetchFileVersion` probe are covered by types and reasoning only, since `packages/web` has no
component tests near this panel (its suite is pure helpers).

PR #590, opened 2026-09-02
Five small UX fixes: Doubao Seed rows, discount badge wording, cost-center clear range, Copy Session ID, Ctrl+W closes a terminal

hiyouga:
## Summary

Five small UX fixes in one batch:

1. **Doubao Seed models on TokenDance** — `seed-2.1-pro`, `seed-2.1-turbo` and `seed-evolving` join the TokenDance group on the gateway's 50%-off promotion (list price in `pricing`, rate in `discount`, like the group's other promoted rows). Context windows (256K) and protocols come from TokenDance's anonymous catalog API; the detail pages need a signed-in session, so the prices are the seller's quoted post-discount rates doubled back to list (billed CNY 1.5 / 7.5 / 0.3 for Turbo, 3 / 15 / 0.6 for Pro and Evolving). Existing Projects pick the rows up via the models page's **sync presets**.
2. **Discount badge wording** — `-50%` becomes `50% off` / `50% 优惠`.
3. **Cost-center clear** — the confirmation names the picker's preset ("in the last 7 days") and spells out dates only for a custom range; and a clear now deletes exactly what the panel listed. Two gaps closed: the trailing presets narrow the error reads and the delete to their instant window (`fromTs`/`toTs`, inclusive, on both error routes and the dashboard's error statistics), and an admin's clear takes the unattributed rows an admin's read includes — the rows a Project-scoped clear used to leave behind, which is the "sometimes not everything" report. With the delete's reach equal to the read's, `UsageErrors.clearable` is removed (`total` is what a clear takes).
4. **Copy Session ID** — label `Copy Session ID` / `复制 Session ID`; the row menu order becomes pin, rename, remote control, archive, copy Session ID, delete.
5. **Ctrl+W closes the focused terminal** — handled inside xterm's key handler only (the shell never sees it, nothing window-wide): the dock runs the tab's confirm-then-kill; the `/terminal` page confirms, kills, then closes its own window where the browser allows. The tab's × tooltip says `(Ctrl+W)`. Browsers may close their own tab first; the desktop app delivers the key.

Docs: `models`, `web-app` and `server-api` pages updated; five bilingual changelog pairs under `changelog/unreleased/`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/97

## Verification

- `pnpm typecheck` — exit 0
- `pnpm --filter @prismshadow/penguin-web test` — 118 files, 1622 tests passed
- `pnpm --filter @prismshadow/penguin-server exec vitest run test/errors.test.ts test/usage.test.ts test/models.test.ts` — 3 files, 126 tests passed
- `pnpm --filter @prismshadow/penguin-core exec vitest run test/model-catalog.test.ts` — 36 tests passed
- `pnpm --filter @prismshadow/penguin-docs test` — 53 tests passed
- `pnpm format` + `pnpm format:check` clean; `git diff --check` clean

## Notes

- TokenDance's catalog API tags `seed-2.1-pro` / `seed-2.1-turbo` with a limited-time 20%-off line (「指定服务商限时调用八折」); the rows record the 50% the seller quoted, as every promoted TokenDance row records the seller-confirmed rate. Worth a glance at the gateway's price pages from a signed-in session.
- Admin clears now remove unattributed rows (login failures, process crashes). This reverses the earlier "delete strictly narrower than any read" rule: those rows show identically in every Project's admin panel and could never be cleared from anywhere, so the one person who can see them clears instance-level noise. Members are unaffected.
- No on-disk data or config format changes; `UsageErrors.clearable` was a response field only. An older web bundle talking to this server simply stops offering the clear button (it gated on `clearable > 0`).
- The design specs do not describe the terminal dock or Ctrl+` at all, so the shortcut is documented in `web-app.{en,zh}.md` only.
- Conflict hotspots: `strings.ts` / `strings-en.ts` (three small hunks each), `packages/server/src/api/types.ts` (`UsageErrors`), `packages/web/src/api/endpoints.ts` (the two usage-errors endpoints), `web-app.{en,zh}.md`.


hiyouga:
## Review

Five unrelated fixes, reviewed one at a time.

### Fixed in 0384117e

- **`packages/server/src/http/routes/usage.ts:37` — `optionalTsParam`'s comment named only half the tables it now guards.** The parser is shared by the dashboard and the two error routes, and its normalization is what makes the `ts >= :fromTs` string comparison in `ErrorsRepo.conds` valid — but the comment still said "string comparison against `usage_records.ts`", which is the one table the error routes do not touch. Now names `error_records.ts` too.
- **`packages/web/src/features/usage/usage-page.tsx:193` — `loaded` re-spelled the default preset as a literal.** `useState<LoadedRange>(() => ({ preset: "7d", … }))` sits two lines under `useState<RangePreset>("7d")`. Changing the page's default range would have left the clear confirmation naming the old preset until the first load returned. Now reads `preset`.

### Findings not fixed

- **`packages/server/src/http/routes/usage.ts:165` — `DELETE /usage/errors` enforces no date bounds; the UI does.** `clearableFilter` (`errors-panel.tsx:74`) hides the button unless both `from` and `to` are set, precisely because an open bound is wider than the confirmation's "records outside that range are kept". The route reads both through `optionalDateParam`, so a bare `DELETE /api/projects/<p>/usage/errors` from an owner takes that Project's entire error history. This PR widens the consequence: with `includeGlobalErrors: c.var.user.isAdmin` now forwarded to the delete, an admin owner's bare DELETE additionally takes **every unattributed row in the instance, for all time** — rows that belong to no Project and show in every other Project's admin panel. `test/errors.test.ts:1397` exercises exactly that unbounded shape and asserts `{ deleted: 1 }`. Pre-existing gap on `main`; left alone because closing it (400 when a bound is missing) changes three tests including two of this PR's own — your call.
- **`design/specs/05-ARCHITECTURE.md:1517` currently says the opposite of this PR, on both halves of fix 3.** The spec states 「无归属异常不在任何一次清空之内，管理员也不例外」and 「确认弹窗报的是**可删条数**而非屏幕上的条数」, and line 1453 records the route as 「两端日期必填」. Design PR #97 amends the first two and is still open. CLAUDE.md's rule is 设计先落地，代码再跟上, so #97 wants merging before this one; the third (required date bounds) is not in #97 at all and still contradicts the route either way.
- **`errors-panel.tsx:53` — a trailing preset's confirmation goes stale in a way the old wording did not.** The two trailing presets pin their window at load time and never recompute. Open the cost center, pick "Last hour", leave the tab two hours, click Clear: the dialog says "in the last hour" while the delete takes `[T0−1h, T0]`, now three hours old, and the rows from the actual last hour survive. The previous wording printed the two dates, which were equally stale but checkable against the panel. The invariant the PR is built on ("takes what the panel showed") still holds — the panel is stale too — so this is a wording judgement, not a bug in the delete. Left alone because the only fixes are re-adding instants to the sentence (which undoes the change) or reloading on open.
- **`packages/desktop/src/menu.ts:70` — "the desktop app delivers the key" holds only while the terminal has focus.** `{ role: "windowMenu" }` on Windows/Linux expands to Electron's default submenu including `{ role: "close" }`, whose accelerator is `CommandOrControl+W`. The handler's `event.preventDefault()` keeps that accelerator from firing, because Electron only runs menu accelerators for key events the renderer did not consume — so inside the terminal the shortcut is safe, and anywhere else in the app Ctrl+W closes the whole desktop window. Not a regression (that was already true), but the new shortcut is what trains the habit. On macOS the template binds no Cmd+W anywhere and the handler correctly requires `!event.metaKey`, so Cmd+W stays inert there.

### Notes

- **Fix 1 (Doubao Seed rows) — arithmetic and placement check out; the prices themselves I could not verify.** `cny()` is `(cacheRead, cacheWrite, output)`, so `cny(1.2, 6, 30)` at `discount: 0.5` bills 3 / 15 / 0.6 input/output/cache-hit, which is what the test, the changelog and the PR body all say; Turbo's `cny(0.6, 3, 15)` likewise gives 1.5 / 7.5 / 0.3. The three rows land in the group's alphabetical order, and `contextWindow: 256000` has precedent in the file (`stepfun/step-3.7-flash`, line 835) alongside the 262144 rows. The `clientType: "openai-chat"` pin is load-bearing and correct: agenthub's `_client_class_for_model` (`src_py/agenthub/auto_client.py:114`) falls through to `OpenaiChatClient` on the pinned type, while an unpinned `seed-2.1-pro` would match no branch and raise. agenthub holds no TokenDance or Doubao entry at all, so nothing there contradicts these rows. The list prices are the seller's quoted post-discount rates doubled back, from a signed-in session — not reproducible from here, and the PR's own note flags the catalog API's competing 20% line. That is the one number in this batch I cannot confirm.
- **Fix 2 (discount badge) — clean.** Both dictionaries changed, the hover titles left alone, and the badge row is `flex-wrap` (`models-page.tsx:2081`), so the wider label wraps rather than overflowing.
- **Fix 4 (Copy Session ID) — clean.** Label in both dictionaries, one `STAT_ICONS.copy` glyph shared with the details card, order pinned by the unit test, e2e assertion follows, and the copy affordance's accessible name stays the label while only the tooltip flashes the copied state.
- **Fix 5 — the absence of focus management is what makes the shortcut safe, which is worth knowing before someone "fixes" it.** `Modal` (`components/ui/modal.tsx`) neither moves nor traps focus, so after Ctrl+W the confirmation opens with focus still in xterm's textarea. The cost is that keys typed at the dialog reach the live shell behind it — Enter runs whatever is at the prompt. The benefit is that a second Ctrl+W is still consumed by the terminal instead of reaching the browser's close-tab or Electron's Close accelerator. Adding a focus trap to `ConfirmModal` without also handling Ctrl+W would turn the second press into a closed window; I nearly landed a `term.blur()` here before working that out.
- `event.type !== "keydown"` (`terminal-view.tsx:364`) already guards the handler, so the close request fires once per press rather than again on keyup; the dock's `!open` guard is correct too — `useDockMount` renders a collapsing dock with `open: false`, so a lingering view cannot answer the same request a second time.
- Both new strings exist in both dictionaries and slot into the existing sentences without regrammaring them; the preset phrases match the picker's own labels ("Last hour" → "in the last hour", 「近 7 天」→「近 7 天内」). `clearable` has no remaining reference anywhere in the repo.
- Removing `UsageErrors.clearable` changes an HTTP response shape. `changelog/README.md:50` reserves the `Breaking` field for exactly that, and sibling entries use it for comparable changes (`2026-08-28-update-modal.md:7` for `POST /api/version/update`). The cost-center entry records the removal in Details but carries no `Breaking` field, so `grep -rl 'Breaking:' changelog/` will not list it.

Verified: `pnpm typecheck` — exit 0, all 8 packages. `pnpm --filter @prismshadow/penguin-core exec vitest run test/model-catalog.test.ts` — 36 passed. `pnpm --filter @prismshadow/penguin-server exec vitest run test/errors.test.ts` — 68 passed. `pnpm --filter @prismshadow/penguin-web exec vitest run test/errors-panel.test.ts test/session-row-menu.test.ts` — 23 passed. `pnpm format:check` — clean. `pnpm --config.verify-deps-before-run=false -r build` got through core, server, cli and web before my session was interrupted and SIGTERMed it inside `packages/desktop`; I did not re-run it, since neither of my two edits changes a type any package compiles against. No e2e run.

hiyouga:
## Follow-up: the unbounded clear is now refused (9efac803)

Revisiting the first item under "Findings not fixed". I had left it on the grounds that requiring the bounds was a contract change to put to you, but the contract is already written down and this PR is not the thing that set it:

```
DELETE /api/projects/:projectId/usage/errors               # 清空异常表（仅 owner，范围 = 面板当前筛选，两端日期必填）
```

`design/specs/05-ARCHITECTURE.md:1453`. 「两端日期必填」 sits in the DELETE's own constraint list beside 「仅 owner」, which the route does enforce (`requireProjectOwner`) — and design PR #97's one-line edit to that file lands in the 异常落库 prose (~1477), not in this table, so the line reads the same on both sides of it. The route disagreeing with it made this a correctness fix rather than a product decision.

Severity is why it could not wait: `optionalDateParam` reads a missing bound as unbounded on that side, so a bare `DELETE /api/projects/<p>/usage/errors` took the Project's whole error history — and with this PR forwarding `includeGlobalErrors` to the delete, an admin owner's bare DELETE additionally took **every unattributed row in the instance, for all time**: rows belonging to no Project, visible in every other Project's admin panel, with no copy anywhere else.

**`usage.ts:175`** now rejects a missing `from` or `to` with `badRequest`, the same 400 path `optionalDateParam` and `tsWindowQuery` already throw on, so the error code and shape match its siblings. The check sits after `requireProjectOwner`, so a member still gets 403 and a non-member 404 without learning anything from validation. `from`/`to` then pass unconditionally, which drops two conditional spreads. **The two read routes are untouched** — the spec line covers the clear, and an unbounded read is a wide page rather than a deletion.

The UI never issued the request this refuses: `clearableFilter` has always withheld the button while either date input is blank. The gap was that the promise held only for callers who went through the panel.

### Tests whose expectations moved

- **`test/errors.test.ts:1283` "clearing errors takes the filter on screen and nothing outside it"** — its last block asserted `{ deleted: 2 }` for a no-filter clear, commented "No filter clears the Project's whole history, which is what an unfiltered panel showed". It now asserts 400 for each of the three incomplete shapes (neither bound, `from` only, `to` only) and that both seeded rows survive. Those refusals are themselves recorded by `onError` — this is the error-persistence suite — so the surviving-rows assertion filters out the resulting `bad_request` row; that is the suite's nature, not a side effect of the change.
- **`test/errors.test.ts:1374` "an admin's clear takes the unattributed rows an admin's read showed"** — the DELETE gains `?from=2026-07-06&to=2026-07-06`, the date its seeded row already carries. The assertion is unchanged: `{ deleted: 1 }`, no unattributed rows left. The point it makes is the admin/member split, which the bounds do not affect.
- `:1360` (member → 403) and `:1370` (non-member → 404) still send no dates and still pass unchanged, because authorization runs first.

Also updated so the shipped contract is described consistently: the route's file header and block comment, `UsageErrorsClearResponse`'s docblock, the `DELETE` row of `server-api.{en,zh}.md`, and one clause in both halves of the cost-center changelog entry.

Unchanged from the earlier review: the `Breaking` field question on `UsageErrors.clearable`, the merge-order dependency on design PR #97, and the Doubao list prices being the one number in this batch I could not confirm.

Verified: `pnpm typecheck` — exit 0, 8 packages. `pnpm --filter @prismshadow/penguin-server exec vitest run test/errors.test.ts` — 68 passed. `pnpm format` + `pnpm format:check` — clean, nothing rewritten. No full suites and no e2e.

hiyouga:
## Changelog: `Breaking` on the cost-center entry

### Fixed in ae7609c6

- **`changelog/unreleased/2026-09-02-cost-center-clear-range.{md,zh.md}` — the entry removes a released response field but carried no `Breaking` field.** `UsageErrors.clearable` shipped in `changelog/0.2.9/2026-08-28-error-noise-and-clear.md` and this PR drops it from `GET /api/projects/:projectId/usage` (`errors.clearable`). The removal was noted only in `## Details` prose, so `grep -rl 'Breaking:' changelog/` did not list the entry — the exact query `changelog/README.md` says the field exists to answer. Added `- **Breaking:** yes — …` plus the `## Compatibility` section the field requires, to both halves, matching the shape `2026-08-28-update-modal.md` uses in this same batch for its response-shape change (what the route answered before, what to read or send now). Field name and `yes` stay English in both files; only the reason is translated.

### The DELETE 400: named it, in the same entry

Commit `9efac803` makes `from`/`to` required on `DELETE /api/projects/:projectId/usage/errors`, so a bare call that used to clear the Project's entire error history now answers 400. I judged that breaking and put it in the same `Breaking` line and a second `## Compatibility` bullet, rather than leaving it out:

- The route is a released, documented surface (the server-api table on `main` lists `from`/`to` as "the same pair the reads take", and the reads treat them as optional). A request that succeeded before now fails hard, with no fallback — an external caller finds out at runtime.
- This repository's bar for the field is already lower than that: `0.2.3/2026-08-19-compact-after-restart.md` marks a *refusal moving between error codes* as breaking, and `0.2.4/2026-08-20-usage-charts.md` marks removed `GET /usage` fields as breaking. Silence here would make `grep -rl 'Breaking:'` under-report.
- The counter-argument — the panel never emitted an unbounded clear, and `design/specs/05-ARCHITECTURE.md:1453` already required both bounds, so the old behaviour was a defect — explains why the change is right, not why a caller is unaffected. The migration step is one line ("send both bounds"), which is what the section now says.

The two are kept as separate clauses: one is a response-shape change, the other a request-contract change.

Nothing else in the entry was restructured; the `## Details` bullet that mentions `clearable` is untouched.

Verified: `pnpm format` + `pnpm format:check` clean in the worktree; both files are section-for-section symmetric (metadata 5/5 fields, `## Details` / `## 细节` 4/4 bullets, `## Compatibility` / `## 兼容性` 2/2 bullets). Markdown only — no build or test run.

PR #591, opened 2026-09-02
feat(web): "Create with AI" on the Agents page, the Models page and the Vault tab

hiyouga:
## Summary

"Create with AI" on three surfaces, built on the kit from #583 (this PR's base branch is `feat/ai-create-kit`; it is to be rebased onto `main` once the kit merges):

- **Agents page** — the header's "Create agent" is the kit's split `CreateMenuButton`; the create dialog opens on a `Segmented` 手动配置 / 让 AI 创建 switch. The manual form is unchanged; the AI side is `AiCreatePanel` with five examples (jotting agent, financial Copilot, document RAG agent, deep-research report agent, and the `report-writer` agent the onboarding chain starts from) and a fixed tail that has the Project's default agent run the `agent-initialization` skill (new agent under this Project, id given or semantic, AGENTS.md + name/description, only the skills it needs from the plugin library, no other agent touched, id and how to start a conversation reported at the end). The last mode is remembered in module state; before any choice, a Project whose only agent is `default_agent` starts on the AI path and every other Project on the manual form. While the Project has no agent of its own, the list ends in a "no agent of your own yet" call to action that opens the AI path. The AI footer is the kit's standard "Edit in a new chat" / "Send to agent"; no agent picker.
- **Models page** — owner-only "Add models with AI" (`AiCreateButton`) beside "Sync presets" opens an `AiCreateModal` with four examples (OpenRouter listing page, self-hosted vLLM, local Ollama, DeepSeek's official model into `deepseek` as default). Its tail follows the `penguin-config` skill: one `penguin config model add --provider … --model-id … --project-id <current Project> …` per model, `--client-type openai --base-url` for OpenAI-compatible endpoints, web pages fetched first (named models, else the most popular, about ten), a missing API key asked for once or left empty for the Models page, `.project_config.toml` never touched, `penguin config model list` at the end. The dialog's lead says when the existing "Add group → Import models" path is faster. The page loads its table on every mount, so the group is there on return from the conversation.
- **Vault tab** — owner-only `AiWandButton` beside "Add" opens an `AiCreateModal` whose intro is an `attention` strip warning that values typed into the prompt land in the Trace (suggesting key names only, values by hand). Tail: `penguin config vault set --key … --value … --agent-id <this agent> --project-id <this Project>` per secret, never echo values, never read `.vault.toml`, `penguin config vault list` at the end. Sent to the Project's default agent (it carries `penguin-config`); the tab refetches on every mount.
- Prompt material lives in pure modules — `agent-ai-prompts.ts`, `models-ai-prompts.ts`, `vault-ai-prompts.ts` — plus `agent-create-mode.ts` for the remembered mode; strings sit in the `agent` / `models` / `vault` sections of both dictionaries. Bilingual docs (`web-app`, `models`) and changelog entry included.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/98

## Verification

- `pnpm --config.verify-deps-before-run=false --filter @prismshadow/penguin-server... build` — ok (builds core and server)
- `pnpm --filter @prismshadow/penguin-web typecheck` — ok
- `pnpm --filter @prismshadow/penguin-web test` — 124 files, 1643 tests passed (new: `create-with-ai-surfaces.test.ts`, `agent-create-mode.test.ts`)
- `pnpm --filter @prismshadow/penguin-web build` — ok (Vite)
- `pnpm format` + `pnpm format:check` — ok; `git diff --check` — clean

## Notes

- The models and vault tails carry the current Project id (and the vault tail the target agent id) because the CLI's `--project-id` / `--agent-id` default to `default_project` / `default_agent`; a non-admin user's Project would otherwise be configured in the wrong place.
- `report-writer` carries a hyphen on purpose (the sibling Benchmark examples reference that id): core's on-disk id rule and the agent-initialization skill both allow it; only the Web form's and API's semantic-id rule forbids hyphens, and the AI path does not go through them.
- The create dialog is `sm:max-w-xl` in both modes (the kit's width) so switching paths never changes its width; the manual form's fields and behavior are untouched.
- Returning surfaces: the chat page reloads the agent list after every completed turn, so a newly created agent appears in the list and sidebar; the Models page and the Vault tab load on mount, which the AI flow always triggers by leaving the route.
- No backward-compatibility handling: nothing on disk changes shape.
- Conflict hotspots: `strings.ts` / `strings-en.ts` (`agent`, `models`, `vault` sections), `agents-page.tsx`, `models-page.tsx`, `vault-tab.tsx`, `web-app.{en,zh}.md`, `models.{en,zh}.md`.

## Follow-up 2026-09-12

The Agents page rendered its two create buttons at the kit default size while the models library and the vault tab use `size="sm"`; both the header pair and the empty state now take `size="sm"`, so the buttons match the models library. Verified on the remote test host: web typecheck clean, `pnpm format:check` clean, `web/test/create-with-ai-surfaces.test.ts` 10 passed.


hiyouga:
## Review

Scoped to this PR's own diff against `feat/ai-create-kit`.

### Fixed in 45c88cc4

- **`strings.ts` / `strings-en.ts` `models.aiAddTail`, `vault.aiAddTail` — the two CLI tails configure the wrong data root.** They name `--project-id` but not `--root`. `core/src/environment/tools/command/session-manager.ts` strips every `PENGUIN_*` variable from a command's environment (`HARNESS_ENV_PREFIX`, alongside `STRIPPED_ENV_KEYS`), so `resolveRootOption(undefined)` in the agent's shell never sees the server's `PENGUIN_HOME` and falls back to `~/.penguin/data`. Sequence: run the server on `PENGUIN_HOME=~/.penguin/dev-data` (the documented dev entry point, and any container/server deployment), click **Add models with AI**, let the agent run exactly what the tail asks — the group lands in `~/.penguin/data/<projectId>/.project_config.toml`, `penguin config model list --project-id <id>` reads that same wrong file and reports success, and the Models page (server root `~/.penguin/dev-data`) shows nothing, with no error anywhere. Identical for `penguin config vault set`. Both tails now carry `--root <data root>` on every `penguin config` invocation plus a bullet saying where to find it: the parent of the Environment section's **App Data Dir**, which `core/src/state/paths.ts` defines as `<root>/<projectId>` and `default-config.ts:486` injects as `{{PROJECT_DIR}}`. The agent tail was already correct — `agent-initialization` works off `$APP_DATA_DIR`, not an ambient default.

- **`strings.ts` / `strings-en.ts` `vault.aiAddIntro` — the warning named only the Trace.** A value typed into that box is also sent to the model provider, and appears a second time inside the `penguin config vault set --value <secret>` command line the agent runs — which is why the tail's "never repeat a value back in your reply" cannot contain it. The strip now names all three destinations.

- **`agent-ai-prompts.ts`, `models-ai-prompts.ts`, `vault-ai-prompts.ts` — deleted, inlined at their single call sites.** Each was a pass-through to `S.<section>.<key>` with one caller; `*AiExamples()` additionally did `.map((e) => ({ ...e }))`, a shallow copy of objects the panel only reads. `S` is a live binding (`export let S`), so reading `S.agent.aiExamples` in the JSX is identical at render time. `create-with-ai-surfaces.test.ts` now asserts against both dictionaries directly instead of through the wrappers — `expect(agentAiExamples()).toEqual(dict.agent.aiExamples)` was a tautology — and gained a regression assertion that every `penguin config` line in both tails carries `--root`. Net −50 lines.

- **Docs and changelog** updated in both languages to match the two prompt changes (`models.{en,zh}.md`, `web-app.{en,zh}.md`, the paired changelog entry).

### Findings not fixed

- **`vault-tab.tsx:284` — the flow can carry a live secret into localStorage, and the first example asks for one.** `useAiBridge().openAiChat` writes the composed prompt through `saveDraft` into `localStorage` (`penguin.chatDraft.<userId>.<projectId>`, `draft-cache.ts:120`) before navigating. `draft-view.tsx:596` clears that key only on a *successful send*, so **Edit in a new chat** leaves the secret sitting in browser storage indefinitely, and a later AI-create jump parks it into the persisted draft-session list (`parkActiveDraft` → `draftSessionsKey`). Meanwhile the `set` example ("the value is `<paste key>`") and `rotate` ("Replace the value of GH_TOKEN with `<new token>`") both actively direct the user to paste a live secret, and `set` is the first card — directly against what the warning strip immediately above recommends. I widened the warning but did not touch the examples or their order: which examples ship is your and the design PR's call, and reordering or reworking them is a product decision, not a defect fix. My read is that the key-names-only (`audit`) example should lead, and that the two value-carrying examples are the ones worth reconsidering.

- **`agents-page.tsx:213-223, 725-735` — the AI footer and send path duplicate `AiCreateModal`'s.** `sendAiCreate` is `AiCreateDialog`'s `go()` (same `pickDefaultAgent` → `composeAiPrompt` → `openAiChat` → close), and the three-button footer is that dialog's footer. Not a defect and not avoidable with the kit as it stands — the Agents dialog has to host the `Segmented` switch and the manual form in the same modal, which `AiCreateModal` cannot do, and `ai-create-panel.tsx` explicitly documents the embed. Worth noting only because a fourth surface with a mode switch would copy it a second time; the kit is the place to fix that, not here.

### Notes

- The three surfaces do reuse the kit rather than reimplement it: `pickDefaultAgent`, `composeAiPrompt`, `AiCreatePanel`/`AiCreateModal`, `CreateMenuButton`/`AiCreateButton`/`AiWandButton` and `useAiBridge` are all taken from it, and there is no duplicated navigation or agent-picking logic anywhere in the diff.
- Tails check out against what actually exists: `plugins/agent-tuning/skills/agent-initialization` and `plugins/agent-development/skills/penguin-config` are both real and preinstalled on `default_agent`; `penguin config model add` really does accept `--provider --model-id --project-id --base-url --client-type --api-key --context-window --price-cache-{read,write} --price-output`, and `model list` accepts `--project-id` (the skill's own doc omits that one — the CLI has it, `packages/cli/src/commands/config.ts:259`). `--client-type openai` is the deprecated alias of `openai-chat`, normalized at `model-catalog.ts:1679`; it works, and it matches the skill's wording, so I left it.
- The deepseek example asks to "make it the default model" but the tail's command template never mentions `--set-default` (or `penguin config model default`). The skill covers both, so an agent following the skill will get there; flagging only because the tail is otherwise exhaustive about flags.
- `agents.every(...)` is vacuously true on an empty list, so removing the `agents.length === 0` branch in favour of the `firstAgentTitle` call to action is correct — the CTA covers the empty case too, and `agentsLoading` still gates it behind skeletons.
- The three dead-code items from the kit PR (#583) — `submit()`'s unread boolean return, the `stateAgentId ?? agentId` tautology, the unreachable `delete draft.handoffAgentId` — are **not** replicated in any of these three surfaces. `sendAiCreate(autoSend: boolean)` takes a boolean, does not return one, and both call sites pass it explicitly.
- i18n is clean: no hardcoded user-facing literals in the three components (the only quoted English in the added lines is inside comments), and `Strings = typeof zh` structurally forces both dictionaries to stay in step. The `<AiWandButton>` and split-button caret carry `title`+`aria-label` from the kit. The vault `intro` correctly overrides the panel's muted colour via `toneStrip.attention`.
- Docs anchors `#create-with-ai` / `#让-ai-创建` both resolve; `packages/docs` `content.test.ts` enforces that and passes.
- No on-disk format changes, so nothing here raises a backward-compatibility question.

Verified: `pnpm --filter @prismshadow/penguin-web typecheck` — clean. `pnpm --filter @prismshadow/penguin-web test` — 124 files, 1643 tests passed. `pnpm --filter @prismshadow/penguin-docs test` — 7 files, 53 tests passed. `pnpm format` + `pnpm format:check` — clean; `git diff --check` — clean. `pnpm -r build` ran green before these edits and I did not re-run it: nothing outside `packages/web/src`, `packages/docs/content` and `changelog/` changed, and no other package compiles against those. Not verified by hand: I did not stand the app up to click through the three dialogs, so the data-root defect and its fix are established by reading `session-manager.ts`, `root-option.ts` and `paths.ts`, not by reproducing it in a running instance.

hiyouga:
## Vault surface: the copy and the footer now agree with the warning

Pushed in 3df85c56 (on top of a merge of `feat/ai-create-kit`, 35f49699, which was clean).

`vault.aiAddIntro` tells the user that a secret typed into this prompt reaches the model provider, the conversation's Trace and the agent's own command line, and that the safer path is to have AI create only the key names. Two things on the surface said the opposite.

### The examples

The first card handed the user a placeholder to paste a secret into (`the value is <paste key>`), and `rotate` did the same (`Replace the value of GH_TOKEN with <new token>`). Both dictionaries now read, in order:

1. **Find the keys this agent needs** — *Key names now, values filled in by hand.* Asks for the key names an agent's installed skills need, what each is for and where to apply for it; the values go in by hand. This is the shape the intro recommends, so it leads.
2. **Reset an expired token** — *Clears the value; you paste the new one.* An expired `GH_TOKEN` is reset to a placeholder and the agent reports where to issue a replacement; the new token never enters the prompt.
3. **Connect an internal service** — *Address set now, token left for you.* Sets `GITEA_BASE_URL` to a literal address and creates `GITEA_TOKEN` with a placeholder. This is the one card that still sets a value directly, it is not the lead, and the value it sets is an address rather than a secret.

No card carries a `<…>` placeholder any more, and each keeps its `description`, so the two-line card shape matches the Agents and Models surfaces. Every prompt still works against the existing tail, which already covers both writes (`--value <value>`, or the placeholder `TODO` "when only the key name is wanted").

`create-with-ai-surfaces.test.ts` pins it per dictionary: the first vault example is the key-names-only ask, no example prompt matches `/<[^>]+>/`, and every example has a description.

### `primaryExit`

The Vault dialog now passes `primaryExit="edit"`, so its footer is `Cancel | [wand] Send to agent | **Edit in a new conversation**` with the accent and the rightmost slot both on Edit. The Agents and Models surfaces are untouched and keep the `"send"` default — a model group and an agent carry no secret, and their expense is visible and reversible. `ai-create-footer.test.ts` already covers both arrangements.

The Web App docs' Vault row and this PR's changelog entry were updated in both languages to match.

### Not changed

`vault.aiAddPlaceholder` still opens with "Which key to store and its value…" before offering the key-names ask. It is not an instruction to paste a secret, so I left it in place rather than widening this change; say the word if it should lead with the safe path too.

Verified on 3df85c56: `pnpm --filter @prismshadow/penguin-web typecheck` clean; `vitest run test/create-with-ai-surfaces.test.ts test/ai-create-footer.test.ts test/ai-create-panel.test.ts test/placeholders-parity.test.ts` 19/19 pass, and the full `packages/web` unit suite 1652/1652; `pnpm format` produced no changes and `pnpm format:check` passes. No build or e2e run — nothing outside `packages/web` strings, one prop on the Vault surface, docs and changelog changed.

hiyouga:
Reviewed at `95fd373a4`, against the paired design PR [#98](https://github.com/Prism-Shadow/penguin-harness-design/pull/98).

The `--root` mechanism the prompt tails depend on is correct, and I verified it end to end rather than taking it on trust: `session-manager.ts:69-80` strips every `PENGUIN_*` key — `PENGUIN_HOME` included — from an Agent-spawned command's environment, and `paths.ts:27-29` makes `projectDir = <root>/<projectId>`, so "the parent directory of the App Data Dir" really is the data root. Every CLI flag both tails use exists with those exact names. Owner gating matches the spec exactly (Models and Vault, not Agents). Lifecycle is clean by construction — the dialogs never submit, they compose text and navigate — so there is no stale closure, no double-submit window, no unmount-after-response hazard, and persistence is real rather than optimistic.

### 1. The Agents page re-implements `AiCreateModal` instead of using the kit

`packages/web/src/features/agents/agents-page.tsx:154-232, 745-790`

`createMode`, `aiDraft`, `aiTarget`, `aiReady`, `sendAiCreate`, `useAiBridge()`, the branched footer and `widthClass="sm:max-w-xl"` reproduce `ai-create-modal.tsx:44-112` line for line — same default-agent resolution, same `ready` rule, same Cancel + wand / `S.aiCreate.editInChat` footer, same width, same `composeAiPrompt(value, tail)`. There is no behavioural difference that needs it: the dialog's mode is fixed at open time and never switches, so one modal is not required.

The other two surfaces got this right. `models-page.tsx:1526-1536` and `vault-tab.tsx:288-301` are ~10 lines each that hand `AiCreateModal` a title, a lead, a placeholder, examples, a tail and `agents` — exactly the right shape, and I would not change them. The Agents page should collapse into the same component: add `const [aiOpen, setAiOpen] = useState(false)` and one `<AiCreateModal …/>`, delete the six duplicated members, and restore the manual dialog untouched. It costs one new string pair (`agent.aiCreateTitle`) and reverts the re-indent that makes most of this file's +415/-179 unreadable.

That collapse also removes findings 2 and 9 below.

### 2. The manual Create-agent dialog silently got 29% wider

`agents-page.tsx:748` — `widthClass="sm:max-w-xl"` was added to the shared `<Modal>`, whose default is `sm:max-w-md` (`modal.tsx:196`). So the existing form widened from 28rem to 36rem purely so the AI panel would fit beside it. The changelog says "The manual form is unchanged", `web-app.en.md:110` says "**Create manually** opens the form, unchanged", and design #98 says 手动创建打开原表单. Open Create manually on this branch and it is visibly wider than on `main`, with nothing anywhere saying so. Fixing 1 removes this — `AiCreateModal` carries its own `sm:max-w-xl`.

### 3. The spec and the code disagree about the Vault examples, and the spec is the wrong one

Design #98 says the Vault tab ships 示例三个（**存入一个 API key**、盘点已装 Skill 需要的 key 并先建键名、更换一个 token）. The code ships `audit` / `rotate` / `endpoint` — there is deliberately no "store an API key" card, and `create-with-ai-surfaces.test.ts:210-222` pins that no example asks the user to paste a secret, which is the entire point of the `toneStrip.attention` warning above the cards.

Per `CLAUDE.md` the spec is the 准绳, so these two must not merge disagreeing — and here the code is right. Amend #98's vault bullet to the three examples that actually ship, and say the leading card is the key-names-only ask.

### 4. Both prompt tails in the spec omit `--root`, which the code makes mandatory

Design #98 writes `penguin config vault set --key … --value … --agent-id … --project-id …` and `penguin config model add --provider … --model-id … --project-id … [--base-url …]`. The shipped tails put `--root <data root>` on every invocation plus a bullet explaining why — and per the mechanism above, they are right. Add `--root` to both templates in #98 with the one-line reason.

### 5. The skill those tails name tells the agent the opposite

`plugins/agent-development/skills/penguin-config/SKILL.md:31` carries a hard rule: *"**Penguin's own model** (self-configuration …): the default root without `--root` is correct."* The new tails say a command without `--root` "configures a different one and nothing reaches this Project". Both land in the same context — the tail in the user turn, the skill on invocation — and the tail is the accurate one.

On any server running a non-default `PENGUIN_HOME` (every dev instance, every `--root`-launched server) an agent that follows the skill drops `--root`, writes to `~/.penguin/data`, reports success, and the Models page stays empty. Amend that bullet: an agent running inside PenguinHarness has no `PENGUIN_HOME` in its command environment, so self-configuration must pass `--root` too.

### 6. Dead relative link in both changelog files

`changelog/unreleased/2026-09-02-create-with-ai-agents-models-vault.md:10` and `.zh.md:10` link `2026-09-02-create-with-ai-kit.md`, but that entry moved to `changelog/0.2.10/` when the directory was renamed at release while this branch was open. Should be `../0.2.10/2026-09-02-create-with-ai-kit.md`.

### 7. The first-agent call to action renders on top of search results

`agents-page.tsx:727` — the guard is `agents.every(...)` (the unfiltered list) but the block sits inside the `shownAgents.length > 0` branch. In a Project with only `default_agent`, typing "def" in the header search shows the matching card and then a full-height centred "No agent of your own yet" empty state with two buttons underneath the hit. The sibling surface already decided this: `models-page.tsx:1355` gates its dashed "Add group" on `query.trim() === ""`. Add the same term.

### 8. A failed model-table load permanently disables the AI path, which does not need that table

`models-page.tsx:1068` — `load()` leaves `rows === null` and sets `loadError` on failure, so `disabled={rows === null}` greys both buttons until a later successful load. The manual button genuinely needs it (`{rows && addGroupOpen && <AddGroupDialog/>}` at `:1539` would otherwise be a dead click), but "Create with AI" only composes a prompt and navigates. So one failed `GET /api/projects/:id/models` shows an error and takes away the AI path that could have fixed the configuration. Give `CreateButtons` separate `disabled`/`manualDisabled`, or drop the gate and render `AddGroupDialog` without the `rows &&` guard.

### 9. Drift between the three surfaces

`models-page.tsx:1531` passes its lead as `description` → `<p class="mb-3 text-sm text-gray-600">`; `vault-tab.tsx:292-296` and `agents-page.tsx:781` pass theirs as `intro` → `<div class="text-xs text-gray-500">`. Same semantic role, two rungs, two inks. Pick one. Separately the Agents AI dialog reuses `S.agent.createTitle` ("Create agent") for both paths while the other two title theirs distinctly — which finding 1 fixes anyway.

### 10. Two controls, two labels, one dialog

`models-page.tsx:1070` reads "Create manually" / 手动创建; the pre-existing dashed footer button at `:1355` reads "+ Add group" / ＋新增分组; both call `setAddGroupOpen(true)`. Neither the changelog, the docs nor #98 mentions the dashed one staying. Either pass `manualLabel={S.models.addGroup}` — the prop exists in `create-buttons.tsx:19` for exactly this — or drop the dashed button and say so.

### 11. Test gaps

`create-with-ai-surfaces.test.ts` is not vacuous — I traced every assertion against both dictionaries and they all bite, notably the `--root`-on-every-line loop, which `penguin-config` with a hyphen correctly escapes. But:

- It runs `describe.each` per dictionary and never compares the two, so a 6th zh-only example passes and `Strings` cannot see it either. One line ×3 closes it: `expect(en.agent.aiExamples.map(e => e.key)).toEqual(zh.agent.aiExamples.map(e => e.key))`.
- Nothing pins the Agents page's hand-rolled AI branch — that it composes with `S.agent.aiCreateTail`, that the footer is disabled on an empty draft, that `openCreate` clears the draft. Both idioms exist in the repo already (`ai-create-panel.test.ts` renders with `renderToStaticMarkup`; `example-fill.test.ts` asserts JSX structure with the TypeScript parser). Moot if finding 1 is taken.
- Despite the PR description, **no e2e spec was added** — `agent-create-skills.spec.mjs` only has its selector updated from 创建 Agent to 手动创建. The three new flows have no e2e coverage.

PR #592, opened 2026-09-02
Hooks tab: bare hook-point chips beside the name, an enable switch, zip import/export and AI import

hiyouga:
## Summary

Hook packages installed on an Agent gain an enable switch, and the Agent settings Hooks tab gains what the Skills tab already had: export as a zip, and an import dialog shaped like the Skills tab's: a recommended chat import over a zip upload. The hook-point chips now read the bare point name (`stop`, `user_prompt`, `pre_tool_use`) and sit right beside the package name.

- **core** — `hooks.json` accepts `enabled: false`; an absent field means enabled and `hookPackageEnabled` is the one reader. `Agent.sessionHooks` leaves a switched-off package out (the one place a Session's hooks are assembled), `setHookEnabled` writes the flag and removes it again on enable, and `installHook` keeps a switched-off package off across a library reinstall (an update replaces content, not the user's switch).
- **server** — the hooks routes move from `routes/plugins.ts` to `routes/hooks.ts` and gain `PATCH …/hooks/:name { enabled }` (Project owner only), `POST …/hooks/archive` (hooks.json and the scripts at the zip root or in one top-level directory; the manifest's name, display fields, `enabled` and every hook-point command are validated — a command must be a relative path to a file inside the archive; the written manifest keeps the upload's fields and defaults the point lists; 409 `hook_exists` unless `overwrite`) and `GET …/hooks/:name/archive` (the installed directory as a zip, re-importable byte-compatibly, `-v<version>` in the filename when the manifest carries a real version). Every mutation invalidates the Agent's cached runtimes, as install and uninstall did. `HookItem` carries `enabled`; the zip-slip guard and the decoded-size cap are shared with the skills routes.
- **web** — Hooks tab row: icon · name with the bare chips right after it (wrapping) · description below · version · `Switch` (owners) or a "Disabled" badge (members, switched-off rows read dimmed for everyone) · export · uninstall. "Import hook" opens the same dialog the Skills tab has: a recommended chat-import section (a multi-line source field for a URL / repository / local path / description / pasted hooks config, a live preview of the prompt — a URL, repo or path gets an import lead sentence, free text is the lead verbatim, then the fixed tail stating the review step, the package format, the script contract and the install target — and "Copy prompt" / "Open a new chat" on this Agent, both disabled while the source is empty) over the zip upload. The Skills dialog's "Open a new chat" is now disabled together with "Copy prompt" while its source is empty. The plugin detail Modal's chips lost their "hook" suffix too, and `S.plugins.hookBadge` is gone. The archive download the Skills tab used became `archive-download.ts`, shared by both tabs.
- **docs** — `skills.{en,zh}.md` (hook packages section), `web-app.{en,zh}.md` (Hooks row), `server-api.{en,zh}.md` (three rows). Changelog pair `changelog/unreleased/2026-09-02-hooks-enable-import-export.{md,zh.md}`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/99 — 05 §插件库 / §Hook / API 原型 and 06 §Agent 设置页 / §插件库页.

## Verification

- `pnpm --filter @prismshadow/penguin-core build` and `pnpm --filter @prismshadow/penguin-server build` — success
- `pnpm typecheck` — all seven packages pass
- core: `npx vitest run test/hook-packages.test.ts test/plugins.test.ts test/hooks.test.ts test/goal-hooks.test.ts test/agent.test.ts test/state.test.ts` — 6 files, 156 tests passed (the new `hook-packages.test.ts` covers `hookPackageEnabled`, `setHookEnabled` incl. the reinstall, and a real `createSession` that finds / does not find the package's `user_prompt` hook as the switch flips)
- server: `npx vitest run test/hooks.test.ts test/skills.test.ts test/directory-skills.test.ts` — 3 files, 41 tests passed (the new `hooks.test.ts`: list + switch semantics and permissions, both archive layouts, 13 manifest-validation rejections, zip-slip / malformed bodies, 409 + overwrite, byte-identical export incl. the switch travelling with the package)
- web: `pnpm --filter @prismshadow/penguin-web test` — 123 files, 1637 tests passed (new `hook-import.test.ts`: the tail names the target and states the contract in both dictionaries)
- docs: `pnpm --filter @prismshadow/penguin-docs test` — 53 passed
- `pnpm format` then `pnpm format:check` — clean; `git diff --check` — clean

## Notes

- Rework (2026-09-10): the import dialog was first built on the `feat/ai-create-kit` panel with a segmented upload/AI switch; it now mirrors the Skills tab dialog instead and no longer depends on the kit's panel (only `composeAiPrompt` and the AI bridge). Merged with `main` at 20ae50ec.
- Authorization follows the brief: the switch is owner-only; import, export and uninstall stay member-level like the skills archive routes.
- The chat import opens the conversation on this Agent (the one the package is for), like the skill import; the tail still names the Project and Agent id as the install target.
- A switched-off `goal` package makes a goal start answer the existing 409 `goal_plugin_not_installed` (the Session simply has no such hook); its message says "not installed" — a dedicated code was left out of scope.
- Backward compatibility: none needed. Existing `hooks.json` files carry no `enabled` field and read as enabled; the installer never writes the field, so "absent means enabled" is the permanent contract rather than a shim with an expiry.
- Conflict hotspots: `packages/web/src/lib/strings.ts` / `strings-en.ts` (`plugins.hookBadge` removed, the `hooks` section extended), `packages/server/src/api/types.ts`, `packages/web/src/api/endpoints.ts`, `packages/web/src/features/agents/skills-tab.tsx` (`exportSkill` body, `UPLOAD_LABEL_CLASS` exported), `packages/server/src/app.ts` (two import lines), the three docs page pairs.
- Visual pass done against the built app (Playwright, zh locale, 2x): the row layout (chips beside the name, description below, version · switch · export · uninstall), the switched-off row dimming with its toast, the import modal's upload mode, and its AI mode (agent line with picker, example selected, full-prompt fold expanded, Cancel · Copy prompt · Edit in a new chat · Send to agent footer) all render as designed.
- Not touched: `plugins-page.tsx`'s header comment still describes card badges the cards no longer render — pre-existing.


hiyouga:
## Review

Scoped to this PR's own diff against `feat/ai-create-kit`.

### Fixed in 904ec022

- **`packages/server/src/http/routes/hooks.ts:159` — the archive import's size caps ran after inflation, so they bounded nothing.** `unzipSync` allocates each entry's declared uncompressed size up front, then `parseHookArchive` checked `data.byteLength`. Measured: a zip holding 256MB of zeros is **262,338 bytes** (349,784 base64 chars) — three orders of magnitude under the 14MB decoded cap and the 20MB `dataBase64` cap — and `unzipSync` on it grows RSS by 263.2MB before the first check runs. Scaled to the cap, a 14MB upload inflates to ~14GB. `POST …/hooks/archive` is member-level, so any Project member can OOM the server with a ~350KB request. The caps (per-file, total, and the 200-file count) now run inside `unzipSync`'s `filter`, off the central directory, before a byte is inflated; the zip-slip check moved there too, so no path is inflated ahead of its own validation. A lied-about `originalSize` is not an escape: fflate inflates into exactly that buffer and truncates — verified, a 64MB deflate stream declared as 1024 bytes returns 1024 bytes with a 0.2MB RSS delta. New test `archive: a compression bomb is refused off the declared sizes, without inflating` asserts the RSS delta; against the pre-fix code it fails with `expected 131.20703125 to be less than 48`.
- **`packages/server/src/http/routes/hooks.ts:197` — a zip entry named exactly like its single top-level directory answered 500, not 400.** Input: a zip containing the file entry `h` alongside `h/hooks.json` and `h/stop.mjs`. `topLevels.size === 1`, so `prefix` becomes `h/`, and `"h".slice(2)` leaves an empty relative path; `replaceSkillDirectory` then writes to the staging directory itself — `EISDIR: illegal operation on a directory, open '…/hooks/.h.incoming'`, logged as `[server] Unhandled exception` and returned as `internal`. Now rejected as a 400 with the other malformed layouts, with a case added to the existing zip-slip test.
- **`packages/server/src/http/routes/hooks.ts:380` — the PATCH re-read the flag it had just written and answered 500 `hook_switch_failed` if it disagreed.** That re-validates the previous statement's own write. The only sequence that reaches it is a concurrent uninstall between the write and the read-back, which is the 404 this route already answers for a package that is not installed — so it answers that instead. Removing it also removes an error code that had no entry in either dictionary, i.e. one that would have surfaced as raw English.
- **`packages/web/src/lib/strings-en.ts:2510`, `strings.ts:2459` — `hook_too_large` had no localized string.** `collectHookArchive` throws 413 `hook_too_large`, `apiErrorText` looks the code up in `S.errors.byCode`, and an unmapped code falls back to the raw English server message — so a Chinese user hitting the export cap got English. Added beside `skill_too_large` in both dictionaries.
- **`packages/web/src/lib/strings-en.ts:1230`, `strings.ts:1201` — the zip-import copy did not say the package arrives switched on and its scripts run.** The modal's only explanatory line described the zip layout. An imported package has no `enabled` field, so `hookPackageEnabled` reads it as on, `invalidateAgentRuntimes` fires, and the next Session spawns its scripts — a file picker with no statement that it installs running code. The AI mode's tail already instructs a malware review; the upload mode said nothing. Both dictionaries now state it.
- **`packages/web/src/lib/strings-en.ts`, `strings.ts` — `goal_plugin_not_installed` became wrong copy.** `startGoal` maps a null `runUserPromptHook("goal", …)` to that code, and after this PR a switched-off `goal` package returns null too. Repro: switch `goal` off on the Hooks tab, then start a goal in chat — the user is told to install a plugin that is installed and visible one tab away. The message now also names the switch.

### Findings not fixed

- **`packages/server/src/http/routes/hooks.ts:371` vs `:290`/`:390` — the owner-only switch is not an authorization boundary; any member re-enables it two ways.** The route comment says the switch "decides what runs on every member's Sessions, the way the Project's policy settings do", but `PATCH` is the only member-gated verb in the file. Two sequences, both verified against the test app:
  1. `DELETE …/hooks/goal` (member, 204) then `POST …/agents/:a/plugins {names:["goal"]}` (member, 201). The reinstall has no previous manifest to read, so the package comes back with `enabled` absent, i.e. on. `installHook`'s switch-preserving branch only covers reinstalling *over* an existing directory.
  2. `GET …/hooks/goal/archive` (member), delete `enabled` from the manifest, `POST …/hooks/archive {overwrite:true}` (member, 201) — back on.

  Which way this should go is a permission-model decision, not a code fix: either the import/uninstall verbs become owner-gated too, or the switch is documented as a convenience rather than a control. Leaving it to you, but the docs and the route comment currently claim the stronger reading.
- **`packages/server/src/http/routes/hooks.ts:290` — a member-level route that installs arbitrary executable code, enabled on arrival.** `POST …/hooks/archive` writes attacker-chosen `.mjs` files, and `runHookScript` spawns them with `spawn(process.execPath, [script], { env: { ...process.env, … } })` — the server's full environment, no approval gate, at every hook point of every new top-level Session. The skills archive route it is modelled on writes Markdown, which is a prompt-injection surface, not an execution one; this is the first member-level path to unsandboxed code execution on the host. Whether an import should land `enabled: false` pending an owner's consent is your call — it would cost the byte-identical round-trip the export currently guarantees. Flagging rather than deciding.
- **`packages/server/src/http/routes/hooks.ts:238` — `collectHookArchive` reads each file fully before checking its size.** `fs.readFile(absChild)` runs, then `data.byteLength > MAX_FILE_BYTES` is tested at :241. A hooks directory holding one very large file (placed out of band — the import now caps at 5MB/file) is read into memory in full before the 413. An `fs.stat` ahead of the read closes it. Left alone because `collectSkillArchive` is byte-for-byte the same shape and I did not want the two to drift; worth doing to both in one pass.
- **`packages/server/src/http/routes/skills.ts:75` and `:106` — both of the bugs I fixed above have exact twins in the file this route was copied from.** `parseSkillArchive` applies its caps after `unzipSync` inflates (same compression-bomb reachability, same member-level route), and its single-top-level-directory branch has the same empty-relative-path crash. Untouched: they are outside this PR's diff and fixing them is a separate change. Since this PR exported `assertSafeEntryPath` and `MAX_ARCHIVE_BYTES` to share them, folding the `filter` into a shared helper would fix both at once.
- **`packages/server/src/http/routes/hooks.ts:300` — the `try/catch` around `Buffer.from(dataBase64, "base64")` cannot fire.** `Buffer.from` ignores non-base64 characters and never throws (`Buffer.from('!!!not base64!!!','base64')` returns 6 bytes). The `badRequest("dataBase64 is not valid base64.")` at :302 is unreachable; the empty/`byteLength` checks below it are what actually catch garbage. Left because the skills route carries the identical block and I did not want them to diverge over a dead branch.

### Notes

- **Backward compatibility, on-disk `hooks.json`:** reading is clean — an absent `enabled` means enabled, so every manifest written before this PR loads unchanged and no migration is implied. The other direction is worth a decision: a build older than this PR does not know the field, so a package a user switched off runs again after a downgrade. No `changelog/unreleased/…-backward-compatibility.md` in the PR; per CLAUDE.md that is the user's call, not mine.
- The AI-import prompt cannot be used to smuggle instructions past the fixed tail: `composeAiPrompt` puts the user's draft **first** and the tail last, `projectId`/`agentId` are id-validated before they reach the template, and the prompt is sent to the user's own agent from their own draft — no second party's input crosses into it.
- `packages/server/src/services/agent-service.ts:234` — `hookCount` on the agents page counts switched-off packages, so a fully switched-off Agent still advertises "2 hook packages". Defensible ("stays installed"), just noting it as a consequence.
- `packages/web/src/features/agents/hooks-tab.tsx:314` — the switch's accessible name is the action phrase `Enable the ${name} hook package`, which reads oddly when it is already on ("…, switch, on"). The two other `Switch` call sites in the app (`models-page.tsx:3223`, `agent-settings-page.tsx:1090`) both use a noun phrase naming the setting. Not changed — it is a valid accessible name and the copy call is yours.
- `packages/web/src/features/agents/hook-import.ts` — `hookImportExamples()` and `hookImportTail()` are one-line reads of `S.hooks.*` with a single call site each; `hookImportExamples` additionally copies the array for no reader that mutates it. The parallel `skill-import-source.ts` earns its file with `classifyImportSource`; this one has no logic to hold. Left as is since a test file pins the module.
- `packages/web/src/features/agents/hooks-tab.tsx:227` — the AI footer's "Copy prompt" button duplicates the `CopyButton` that `AiCreatePanel` already renders inside its full-prompt fold, and makes this footer the only four-control one in the kit (`AiCreateModal` has three). It is defensible because the fold is collapsed by default.
- `packages/web/src/features/agents/hooks-tab.tsx:192` — `onPickFile` is byte-identical to the Skills tab's. Extracting it was not obviously worth a new abstraction, but it is the second copy.
- The rest of the frontend pass came back clean: both dictionaries carry every new key (enforced by `Strings`), the row's chips wrap instead of overflowing, list/empty/skeleton/error states match the Skills tab, the stacked overwrite `ConfirmModal` sits above the import `Modal` in the Esc-layer stack so the first Escape closes only the confirm, and `S.plugins.hookBadge` is gone with no remaining reference.

Verified (Node v24.18.0, `FORCE_COLOR=0`): `pnpm typecheck` — all 8 packages pass. `pnpm format && pnpm format:check` — clean. `packages/server` `vitest run test/hooks.test.ts` — 7 passed (6 existing + the new bomb test). `packages/core` `vitest run test/hook-packages.test.ts` — 3 passed. `packages/web` `vitest run test/hook-import.test.ts` — 4 passed. `pnpm -r build` passed on the branch before my changes; not re-run after, since nothing I touched is compiled against by another package (`api/types.ts` is untouched). Not run: the web Playwright e2e suite, and the full `pnpm test` of any package.

hiyouga:
## Correction to my review: the lied-about `originalSize` analysis was incomplete

In the review above I wrote that a declared size that lies "is not an escape: fflate inflates into exactly that buffer and truncates". The truncation is real, but the conclusion was too strong, and I have corrected both the test and the source comment in `01e0754c`. Credit to the #601 / #594 reviewers, who measured this on the twin routes.

**What I got wrong.** The declared size is allocated whether or not it is honest, and the truncated result is what defeats a cap read after inflation. `unzipSync` calls the `filter` with `originalSize` taken from the central-directory record (`zh()`), then allocates `new u8(su)` on the next line and inflates into it — so what comes back is a *view* of the real length onto a buffer of the declared length. Measured here on the hooks route's own fixture:

```
CDH: method=8  csize=7  usize=5  ->  usize patched to 536870912
archive                    = 227 bytes
h/pad.bin decoded byteLength = 5
  backing ArrayBuffer        = 536870912
  external 2.2MB -> 514.2MB
```

So a 227-byte upload takes 512MB of heap and yields a 5-byte value. Every check the pre-fix route made was on that 5 — which is why, against this PR's head (`95904a0c`), the route **answered 201 and installed the package**. That is a second, independent way past the caps, and my honest-zeros measurement (256KB → 263MB RSS) did not cover it.

**The fix was already correct; the test was not.** My `filter` bounds on `file.originalSize`, i.e. the central directory's declared size, strictly upstream of the allocation — so a 512MB declaration is refused before any memory is taken. No production change was needed. But the test I shipped in `904ec022` asserted an RSS delta on an *honest* zeros archive, and would have passed against a filter that still bounded post-inflation, so it did not pin this vector. It was also flake-prone: vitest shares one process across a package's test files, so an unrelated allocation can move the number.

**What `01e0754c` changes.** Both archive-limit cases now assert status plus the error message's named entry, with no memory assertion:

- `archive: an entry that lies about its uncompressed size is refused, not installed` — the regression pin. Sub-kilobyte archive, 5-byte entry declaring 512MB; expects 400 naming `h/pad.bin` and an empty hooks directory. Against `95904a0c` it fails with `expected 201 to be 400`.
- `archive: an entry over the per-file cap is refused however small the archive is` — 8MB of zeros in a sub-64KB archive. This one **passes** against `95904a0c` too (it refused it, after inflating the 8MB), so I have relabelled it as the per-file cap's behaviour test rather than a pin, which is what it always was.

One trap worth recording, since it would make the fixture prove nothing: the declared size is only allocated on the deflate path — a stored entry (`c_2 == 0`) is sliced at its compressed size and never reads `su`. `declareSize` therefore reports the central record's compression method and the test asserts it is `8`.

The docstring on `parseHookArchive` carried the same wrong claim and now states both mechanisms. Also noted for whoever lands the shared `unzipBounded` from #601: this route's copy bounds on the declared size and returns 400 (`badRequest`) rather than 413, matching the other cap errors here.

Verified: `pnpm typecheck` (all 8 packages) and `packages/server` `vitest run test/hooks.test.ts` (8 passed), `pnpm format` + `format:check` clean. Both pre-fix comparisons above were run against `git show 95904a0c:packages/server/src/http/routes/hooks.ts` checked into the tree, not a hand-reconstructed revert.

hiyouga:
## Decision on the two open permission questions: no change, and why

Both open items from my review — that `POST …/hooks/archive` is member-level and that imported
packages arrive enabled, and that the owner-only switch is re-enablable by a member — are
**closed as accepted**, by the maintainer's call. Recording the reasoning so this is not
re-raised as a finding on a later PR.

**The premise, which I checked rather than assumed: a project member already has arbitrary code
execution on this server.** `POST …/sessions` goes through `requireProjectAccess`
(`sessions.ts:429`), not `requireProjectOwner`, and `approvalMode` is an ordinary body field on
session create (`sessions.ts:483`). So a member can open a session with `approvalMode:
"allow-all"` and call `exec_command` directly. Hook import therefore grants no capability the
member did not already hold, and gating it would move a lock onto a door standing beside an open
one.

What genuinely differs is narrower than "code execution": a hook runs **without passing the
approval path** and **on every new Session** rather than on demand. Both are real properties of
the feature — a hook that asked for approval on each Session would not be a hook — and neither
is reachable by anyone who could not already run the same code by hand.

So, deliberately unchanged:

- imports stay member-level;
- imported packages stay enabled on arrival, keeping export → import a byte-identical round trip;
- the two member-level paths that clear a `false` switch (reinstall via `POST …/plugins` with no
  prior manifest, and re-import with `overwrite: true` after dropping `enabled`) stay open. The
  switch is an operational control, not an authorization boundary.

**The one thing worth keeping accurate** is the route comment claiming the switch works "the way
the Project's policy settings do". Policy settings *are* an owner-level boundary and this is not,
so that sentence invites exactly the wrong inference. Worth a wording pass whenever this file is
next touched — not worth a commit of its own.

The security fix in this PR is untouched by the above and stands on its own: the pre-inflate
`filter` bound off the central directory, and the lying-header regression test at `01e0754c`.

PR #593, opened 2026-09-02
Scheduled tasks in the chat dock, the Session row menu, and the Schedule tab's AI path

hiyouga:
## Summary

Scheduled tasks become reachable from the conversation itself, on top of the "Create with AI" kit (#583):

- **Dock panel kind `schedules`** (`features/schedules/schedule-panel.tsx`): the tasks bound to the current Session (the agent's schedules filtered by `sessionId`), a search box, All / Active / Paused / Completed chips, rows with a state glyph, the task name and a plain-language schedule line (`describeSchedule`: "每天 08:00 · 下次 明天 08:00", "Every Monday at 09:00", "每 30 分钟", "One-off · Sep 3, 10:00"; settled states named first), an enable switch and an edit / delete menu for owners, and a Suggestions list (daily brief / weekly review / follow-up reminder / update monitor). The list refetches when the tab comes to the front, on window focus and visibility, every 30 s while visible, and after every mutation; the draft page says what the first message unlocks.
- **"Create with AI" that sends INTO the current Session** (`schedule-ai-modal.tsx`): `AiCreatePanel` with the in-Session instruction tail; the footer is "Send to this conversation" / "Copy prompt". Delivery is the chat page's new `sendTextToSession`: `POST /steer` while a Task runs, otherwise `POST /task` with `queueIfBusy` (a 409 from the steer falls back to the task post). The composer's own draft is untouched, and approvals / queueing are the server's, as for any task.
- **Session row menu → "Schedule a task"** (after the messaging binding): a two-row chooser at the menu's anchor — "Create with AI" navigates to the conversation with route state `{ schedulePanel: "ai" }` (the chat page opens the panel in the right dock, the panel opens its dialog, and the state is consumed by replacing the history entry so a reload does not reopen it); "Set up manually" opens the form pinned to that Session (owners).
- **Schedule tab**: the kit's split "New" button — the AI path is `AiCreateModal` to a new Session of the Project's default agent, with a tail naming the agent, `penguin schedule add` / the TOML file, and the new-Session mode; the suggestions stand in for an empty table.
- The create / edit form is extracted from the tab into `schedule-form-modal.tsx` (shared by the tab, the panel and the row menu; `lockedSessionId` pins the target and hides the new-Session fields). The alarm clock becomes `SCHEDULE_ICON` in `components/ui/icons.tsx`, shared with the agents page's count. `AiCreatePanel` gains a `byLine` override, since the in-Session dialog must not say "in a new conversation".
- Specs, the Web App docs (both languages) and a bilingual changelog entry.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/100

## Verification

- `pnpm --config.verify-deps-before-run=false --filter "@prismshadow/penguin-web^..." build` and `pnpm --filter @prismshadow/penguin-web build` — both exit 0
- `pnpm --filter @prismshadow/penguin-web test` — 124 files, 1654 tests passed (new: `schedule-describe.test.ts` 11 tests, `schedule-panel-state.test.ts` 7 tests; updated: `session-row-menu.test.ts`, `dock-state.test.ts`)
- Visual check on the built app (server on a scratch data root on a high port; project, Session and four schedules seeded through the API, three bound to the Session and one new-Session): the panel (zh and en) lists exactly the three bound tasks with the expected lines, the in-Session AI dialog, the row menu with its chooser, the pinned manual form and the settings tab's split button render as designed; no page errors.

## Notes

- Based on `feat/ai-create-kit`; to be rebased onto `main` after #583 merges.
- Members (non-owners) get the AI path only — asking the agent is a message, not a write; the switch, edit, delete and the manual form stay owner-only, as on the tab.
- While the Session is compacting, in-Session delivery goes through `queueIfBusy` (held server-side), the same as the composer's queue mode.
- Backward compatibility: none needed. A stored dock layout naming `schedules` is dropped by an older build's sanitizer, and older layouts simply lack the tab.
- Conflict hotspots: `strings.ts` / `strings-en.ts` (`schedule` section), `session-row-menu.tsx` and its test (smallfix moves `copy`), `sidebar.tsx` (SessionRow props and render site), `chat-page.tsx` (imports, jump-command state, `renderPanel`), `dock-state.ts` / `panel-meta.tsx` / `dock-panel.tsx`, `ai-create-panel.tsx` (`byLine`), `web-app.{en,zh}.md`.


hiyouga:
## Review

### Fixed in 32fb0240

- **`schedule-describe.ts:81,108` — the schedule line contradicts the next trigger it prints, after a DST change.** `describePeriod` took its wall clock from `startAt`, but the server steps in fixed milliseconds (`schedule-file.ts` `nextSlotAfter`: `startAtMs + k * periodMs`) and stores no timezone with the task. A daily task created at `2026-03-01T08:00` in `Europe/London` therefore fires at 09:00 local from 2026-03-29 on, and the row renders `每天 08:00 · 下次 明天 09:00` / `Every day at 08:00 · Next: tomorrow 09:00` — the summary asserting a time the task will never fire at again. The anchor is now `item.nextFireAt ?? item.startAt`, so the wording follows the trigger printed beside it; a settled task (no `nextFireAt`) is unchanged, and a `7d` period keeps its weekday because the next slot is a whole number of weeks away. New case in `test/schedule-describe.test.ts`.
- **`schedule-create-chooser.tsx:49` — Escape out of the chooser drops keyboard focus on `<body>`.** The chooser passes `button={null}` and portals its rows, so `Dropdown`'s default Escape target — `ref.current?.querySelector("button")` — resolves to null; the focused row then unmounts and focus resets to the document. Repro: focus a Session row, Shift+F10, arrow to "创建定时任务", Enter, Escape — the user is out of the list with no place to return to. This is the exact case `dropdown.tsx`'s header calls out and that the row menu itself solves with `returnFocus={ctx.returnFocus}`; the chooser now takes the same accessor.
- **`chat-page.tsx:402` — the row menu's AI path yanks the schedules tab out of the bottom dock.** `openPanel("schedules", "right")` names a dock, so a user who had moved the tab to the bottom dock gets it pulled back to the right edge on every "Schedule a task → Create with AI". The three sibling jump commands (`openPanel("workspace")`, `("agents")`, `("memory")`) name none, and `openPanel` already falls back to `"right"` for a tab with no home — dropping the argument keeps the default and respects the placement.
- **`schedule-panel.tsx:309` — a truncated task name has no tooltip.** The name is `truncate`d and its `title` carried `item.prompt`, so a name like `weekly_status_report_for_leadership` is unreadable in a 360px dock and recoverable nowhere in the panel. The settings tab's own rule for the same value is `title={item.name}` (`schedules-tab.tsx`, "Long text columns truncate with the full value on hover"). The name now leads the tooltip with the prompt under it, the `${a}\n${b}` shape already used in `draft-view.tsx` and `shortcuts-folder.tsx`.

### Findings not fixed

- **`session-row-menu.tsx:91-92` — "Schedule a task" is offered on subagent Session rows.** `contextMenuActions(false)` is the folder-row list (archived / subagent / scheduled) and includes `schedule`, so the row menu invites the user to bind a task to a subagent Session. `design/specs/08-CHANGELOG.md` (2026-08-26) names that target as one of the three the built-in Scheduled Tasks guidance tells the model *not* to write `session_id` for, because a subagent's Session is released when idle — "恰是本改动要避开的那种无人查看的 Session". Left alone: gating a menu row by row kind is a product call, and the server accepts the binding either way.
- **Design parity.** `design/specs/` on `main` has no `schedules` dock panel, no "Schedule a task" row-menu entry and no in-Session AI delivery; design PR penguin-harness-design#100 is open but unmerged. Per CLAUDE.md the design lands first — flagging rather than editing, since specs are a separate repo and a separate PR.
- **`schedule-panel-state.ts:34` — a task bound to this Session from another agent's directory never appears in the panel.** The panel lists `listSchedules(projectId, session.agentId)` and filters by `sessionId`, but the server's fire path does no agent check (`scheduler.ts:358`, `sessions.findById(def.sessionId)`), so a TOML under agent B with `session_id` pointing at a Session of agent A fires into that conversation invisibly. Not reachable from this PR's UI (the picker lists only the agent's own Sessions, and the pinned paths use the Session's own `agentId`) — only from a hand-written file or `penguin schedule add --session-id`. Narrowing it is a server/spec decision.

### Notes

- Timing audit, since this is scheduling code: the PR introduces exactly one timer, the panel's 30 s `setInterval`, cleared in its effect cleanup — no `setTimeout` with a computed delay anywhere, so the >2^31-1 ms fires-immediately trap has no surface here. The server side is a single 30 s poll with a `ticking` re-entrancy guard, `unref`'d and cleared in `stop()`; there are no per-task timers to leak on delete or re-arm on restart, run state lives in SQLite so a fired one-shot does not re-fire, downtime skips rather than backfills (`scheduler.ts` header, and `latestSlotAt` advancing `last_slot` to now), and a deleted target Session marks the task invalid rather than leaving it firing. The one real timezone gap was the display one fixed above: `start_at` is an absolute instant and no timezone is stored with the task, so nothing else in the pipeline has a local-time opinion to get wrong.
- The per-Session in-flight-flag hazard from #589 does not apply: `renderPanel` mounts `<SchedulePanel key={selected.sessionId}>`, so a conversation switch remounts the panel and `busy` cannot survive it, and `load`'s closure captures the Session it was built for. `sendTextToSession` likewise captures `selected` at click time, so a mid-send switch still delivers to the conversation the dialog was opened on.
- None of the three dead-code shapes noted on #583 are replicated here: `ScheduleFormDialog.submit` returns `void`, `ScheduleAiDialog` reads its `SessionSendOutcome` (it picks the toast and decides whether to close), and nothing in these files touches `stateAgentId` or `handoffAgentId`.
- `schedule-panel.tsx:249` renders the panel title as an `<h2>` while the dock tab strip already shows `panelLabel("schedules")` — the same string, twice. `MessagingPanel` deliberately renders no heading for that reason. Not changed: the subtitle and the split Create button share that row, so removing the heading is a layout decision, not a cleanup.

Verified: `pnpm -r build` then `pnpm --filter @prismshadow/penguin-web typecheck` — clean (the build is needed first in a fresh worktree, or the server `.d.ts` resolution fails). `npx vitest run test/schedule-describe.test.ts test/schedule-panel-state.test.ts test/dock-state.test.ts test/session-row-menu.test.ts` — 4 files, 62 tests passed (61 before this commit, +1 for the DST anchor). `npx prettier --write` over the six changed files — all reported unchanged, and `pnpm format:check` was clean over the tree before them. Not run: the web e2e suite (no selectors or flows moved, and ~8 specs fail on `main` locally anyway) and the packages this PR does not touch.

hiyouga:
## Follow-up: the Schedule tab's AI path wrote the task to the wrong agent

Fixed in 22f8dc52, on top of 32fb0240.

### The defect, plainly

**Before this commit, "Create with AI" on agent X's Schedule tab created the scheduled task on the Project's default agent instead, and said nothing.** X's tab stayed empty; the task was real, enabled and firing, just in another agent's directory.

The chain, verified end to end:

1. `schedules-tab.tsx` opens `AiCreateModal` with no `agentId`, so the prompt runs in a new conversation with `pickDefaultAgent(agents)` — the Project's default agent, not X.
2. `app.ts` `controlEnv` injects `PENGUIN_AGENT_ID: ctx.agentId` into every command that conversation runs — the *running* agent, i.e. the default one.
3. `cli/src/client.ts:422` `resolveAgentId(flag)` is `flag ?? PENGUIN_AGENT_ID ?? "default_agent"`, and `commands/schedule.ts:167` `add` calls it with `opts.agentId`.
4. The tail said `use \`penguin schedule add …\`` and named X in prose only. With no `--agent-id`, step 3 resolves to the default agent, and `POST .../agents/<default>/schedules` writes the file there.

Silent because every layer succeeds: the CLI prints its "written" line, the server returns 201, and the agent truthfully confirms "I've scheduled it" — the only symptom is an empty table on the tab the user was looking at.

Two further problems in the same string: `--prompt` and `--start-at` are `requiredOption`s and the tail named neither, so the command as written could not run at all; and it handed the model the TOML keys `start_at` / `period` / `end_at` immediately beside a CLI command whose flags are `--start-at` / `--period` / `--end-at`.

### What I changed

I kept the CLI branch and spelled the command out, rather than dropping it. The branch earns its place: `commands/schedule.ts`'s own header is explicit that the API path gives an agent **synchronous validation instead of the reconcile lag hand-editing hits**, which is the difference between the agent learning about a bad `period` now and the file sitting invalid until the next 30 s scan. Deleting it would have removed the better of the two paths to fix a wording bug. The tail now reads:

> `penguin schedule add <name> --agent-id <X> --prompt "<the request>" --start-at <ISO 8601 or now>`, adding `--period <30m | 12h | 7d>` … and `--end-at <ISO 8601>` … — without `--agent-id` the task lands on whichever agent is running this conversation rather than on that one

The TOML keys now appear only in the file branch, so they can't be read as flags of the command beside them. Both dictionaries, with a comment on each recording why `--agent-id` is load-bearing.

**No `--root`**, deliberately, and it would have been a new bug: `penguin schedule` has no `--root` option at all (`grep -c root commands/schedule.ts` → 0) because it goes through the server, and the injected `PENGUIN_API_URL` / `PENGUIN_API_TOKEN` / `PENGUIN_PROJECT_ID` already point `resolveConnection` at the right one. The `--root` fix was correct for #591 only because `penguin config model|vault` writes the data root directly.

### Also fixed

- **`schedule-ai-modal.tsx:79-87` — two copy controls for one prompt.** The footer `<Button>` used `writeClipboard` + `toastSuccess`, while `AiCreatePanel` already renders `<CopyButton text={fullPrompt} label={S.aiCreate.copyPrompt} />` in the same dialog. Footer button dropped; the fold's glyph-flip is the house convention and the only one now.
- **One control, two labels.** `schedules-tab.tsx:166` passed `label={S.schedule.newButton}` ("新建"/"New") while `schedule-panel.tsx:258` fell through to `S.common.create` ("创建"/"Create"). Both now take the kit's default, and `schedule.newButton` — which had no other reader — is deleted from both dictionaries.

### Left for you to decide

- **`primaryAction` / remembering the last-used path.** Neither of my two `CreateMenuButton`s passes `primaryAction`, so the main half is permanently the AI path. #591 remembers the choice in `features/agents/agent-create-mode.ts`. I have deliberately not built a second memory: if that behaviour is wanted it belongs in `CreateMenuButton` itself, decided once for all six surfaces, not re-implemented per call site.
- **`allowAgentChoice`.** For the in-Session dialog it is not applicable at all — delivery goes to the conversation on screen through `sendTextToSession`, so there is no other agent it could reach; that is exactly why it passes `byLine` instead. For the Schedule tab it was, until this commit, actively unsafe: changing the runner changed which agent the task landed on, silently, by the chain above. Now that `--agent-id` pins the target, the runner and the target are decoupled and offering the picker would be harmless — so this is a free choice for the kit rather than something this PR has to settle.

Verified: `pnpm --filter @prismshadow/penguin-web typecheck` — clean. `npx vitest run test/schedule-describe.test.ts test/schedule-panel-state.test.ts test/dock-state.test.ts test/session-row-menu.test.ts test/placeholders-parity.test.ts` — 5 files, 65 passed (parity included because both tails changed; both sides still interpolate `${agentId}`, now twice each). `npx prettier --write` over the four changed files — all unchanged. Not run: full suites, e2e. Not verified by hand: I did not stand the app up and watch a task land in the right directory — the chain above is read from `resolveAgentId`, `controlEnv` and the `add` command's option list.

PR #594, opened 2026-09-02
Agent import/export: portable bundle, penguin agent export|import, Agents page import/export, agent-porting skill

hiyouga:
## Summary

The Evaluation Center is rebuilt around the loop a first-timer needs — create a Benchmark for an agent, read its scores, hand it to an optimizer — with the prompts written for the person.

- **Page**: title, one-line subtitle and a folded three-step guide naming the Skill behind each step; a search box over titles, descriptions and agents; collapsible per-agent groups (avatar, count, a "create for this agent" wand); rows with case/run counts, the last evaluation time, the newest score with its change, and a score sparkline. **View** (or the row) opens the existing detail — chart, evaluation table, case browser, moved unchanged into `benchmark-detail.tsx` — in a right pane on `lg+` and in place of the list below it. Empty states carry the guide and a Create-with-AI call to action. `?agentId=` still expands only that agent; `?benchmark=<agent>/<id>` opens one directly and a selection writes it back.
- **New Benchmark** (the kit's `CreateMenuButton`): the AI mode adds a Test Agent picker in the panel's intro slot, four scenario examples, and a tail matching `benchmark-design`'s contract (`test_agent_id`, `benchmark_id`, `desired_baseline_score` `<70`, `pilot_iteration_limit` 3, `runs = 1`, the directory layout, the Formal Baseline into `scoreboard.yaml`). The manual mode is a form — agent, title, id proposed from the title, description, runs per case, a dynamic case list (directory suffix, title, statement, rubric) with format hints on screen and the semantics (rubric isolation, what makes a rubric discriminating) behind "?" marks — posting to the new create route.
- **Optimize** (every row and the detail header): `Segmented` 手动配置 / 写提示词 over one `agent-optimization` tail (`test_agent_id`, `benchmark_id`, `runs`, `desired_score` `>=T`, `candidate_round_limit`). Manual: optimizer agent (warns when `agent-optimization` is not installed, read through the skills API), the fixed Test Agent, the model of the optimizer's own conversation (Project models, default = Project default, carried through the draft cache), runs (default the Benchmark's `runs`), round limit 3, target score `min(100, ⌈baseline⌉ + 10)`, optional focus, and the full prompt folded. Prompt: free text with three examples. Both end in the kit footer (send / edit in a new chat). A Benchmark without a baseline shows an attention strip before anything is sent.
- **Overflow menu**: copy the directory path (`agents/<agent>/benchmarks/<id>`); owner-only delete with confirmation.
- **Server**: `POST /api/projects/:p/agents/:a/benchmarks` (owner) writes `benchmark_config.toml`, `scoreboard.yaml` (`evaluations: []`) and per case `statement/README.md` (`# <title>` + statement) / `rubric/README.md` (verbatim); validates ids (PRN-005 alphabet, case ids `CASE-…`, unique, non-blank texts, `runs` ≥ 1); 409 `benchmark_exists` on an existing directory (a non-recursive mkdir is the atomic check) and a half-written directory is removed on failure. `DELETE …/benchmarks/:id` (owner) removes a real directory whole (404 otherwise; symlinks are not followed).
- **Kit**: `AiChatRequest.modelRef` (optional) pins the new conversation's model — the only kit change, additive.
- Docs (`web-app`, `server-api`, `self-improvement`, both languages) and the bilingual changelog pair.

## Design

Prism-Shadow/penguin-harness-design#102 — 05 §「Benchmark 评测与优化」(manual creation through the API, the Web entry points and the fields they hand the Skills, the API prototype) and 06 §「评估中心」(the new page).

## Verification

- `pnpm --filter @prismshadow/penguin-core build`, `…penguin-server build`, `…penguin-web build` — exit 0 each
- `pnpm --filter @prismshadow/penguin-web typecheck` and `pnpm --filter @prismshadow/penguin-server typecheck` — clean
- `pnpm --filter @prismshadow/penguin-web test` — 123 files, 1647 tests passed (new `benchmark-prompts.test.ts`; `benchmark-metrics.test.ts` and `ai-bridge.test.ts` extended)
- `pnpm --filter @prismshadow/penguin-server exec vitest run test/benchmarks.test.ts test/validate.test.ts test/authz.test.ts test/builtin-agents.test.ts test/id-validation.test.ts` — 5 files, 45 tests passed (`benchmarks.test.ts`: 6, three new — the written layout and the 409, malformed and non-owner rejections, delete)
- `pnpm format` + `pnpm format:check` clean; `git diff --check` clean
- Visual self-check: the built app on a scratch data root, driven by Playwright at 1440×900 and 390×844 — list with the guide open, detail pane, the AI / manual / Optimize (both modes) dialogs, the row menu, search; the narrow row wraps its actions under the title.
## Summary

The Evaluation Center is rebuilt around the loop a first-timer needs — create a Benchmark for an agent, read its scores, hand it to an optimizer — with the prompts written for the person.

- **Page**: title, one-line subtitle and a folded three-step guide naming the Skill behind each step; a search box over titles, descriptions and agents; collapsible per-agent groups (avatar, count, a "create for this agent" wand); rows with case/run counts, the last evaluation time, the newest score with its change, and a score sparkline. **View** (or the row) opens the existing detail — chart, evaluation table, case browser, moved unchanged into `benchmark-detail.tsx` — in a right pane on `lg+` and in place of the list below it. Empty states carry the guide and a Create-with-AI call to action. `?agentId=` still expands only that agent; `?benchmark=<agent>/<id>` opens one directly and a selection writes it back.
- **New Benchmark** (the kit's `CreateMenuButton`): the AI mode adds a Test Agent picker in the panel's intro slot, four scenario examples, and a tail matching `benchmark-design`'s contract (`test_agent_id`, `benchmark_id`, `desired_baseline_score` `<70`, `pilot_iteration_limit` 3, `runs = 1`, the directory layout, the Formal Baseline into `scoreboard.yaml`). The manual mode is a form — agent, title, id proposed from the title, description, runs per case, a dynamic case list (directory suffix, title, statement, rubric) with format hints on screen and the semantics (rubric isolation, what makes a rubric discriminating) behind "?" marks — posting to the new create route.
- **Optimize** (every row and the detail header): `Segmented` 手动配置 / 写提示词 over one `agent-optimization` tail (`test_agent_id`, `benchmark_id`, `runs`, `desired_score` `>=T`, `candidate_round_limit`). Manual: optimizer agent (warns when `agent-optimization` is not installed, read through the skills API), the fixed Test Agent, the model of the optimizer's own conversation (Project models, default = Project default, carried through the draft cache), runs (default the Benchmark's `runs`), round limit 3, target score `min(100, ⌈baseline⌉ + 10)`, optional focus, and the full prompt folded. Prompt: free text with three examples. Both end in the kit footer (send / edit in a new chat). A Benchmark without a baseline shows an attention strip before anything is sent.
- **Overflow menu**: copy the directory path (`agents/<agent>/benchmarks/<id>`); owner-only delete with confirmation.
- **Server**: `POST /api/projects/:p/agents/:a/benchmarks` (owner) writes `benchmark_config.toml`, `scoreboard.yaml` (`evaluations: []`) and per case `statement/README.md` (`# <title>` + statement) / `rubric/README.md` (verbatim); validates ids (PRN-005 alphabet, case ids `CASE-…`, unique, non-blank texts, `runs` ≥ 1); 409 `benchmark_exists` on an existing directory (a non-recursive mkdir is the atomic check) and a half-written directory is removed on failure. `DELETE …/benchmarks/:id` (owner) removes a real directory whole (404 otherwise; symlinks are not followed).
- **Kit**: `AiChatRequest.modelRef` (optional) pins the new conversation's model — the only kit change, additive.
- Docs (`web-app`, `server-api`, `self-improvement`, both languages) and the bilingual changelog pair.

## Design

Prism-Shadow/penguin-harness-design#102 — 05 §「Benchmark 评测与优化」(manual creation through the API, the Web entry points and the fields they hand the Skills, the API prototype) and 06 §「评估中心」(the new page).

## Verification

- `pnpm --filter @prismshadow/penguin-core build`, `…penguin-server build`, `…penguin-web build` — exit 0 each
- `pnpm --filter @prismshadow/penguin-web typecheck` and `pnpm --filter @prismshadow/penguin-server typecheck` — clean
- `pnpm --filter @prismshadow/penguin-web test` — 123 files, 1647 tests passed (new `benchmark-prompts.test.ts`; `benchmark-metrics.test.ts` and `ai-bridge.test.ts` extended)
- `pnpm --filter @prismshadow/penguin-server exec vitest run test/benchmarks.test.ts test/validate.test.ts test/authz.test.ts test/builtin-agents.test.ts test/id-validation.test.ts` — 5 files, 45 tests passed (`benchmarks.test.ts`: 6, three new — the written layout and the 409, malformed and non-owner rejections, delete)
- `pnpm format` + `pnpm format:check` clean; `git diff --check` clean
- Visual self-check: the built app on a scratch data root, driven by Playwright at 1440×900 and 390×844 — list with the guide open, detail pane, the AI / manual / Optimize (both modes) dialogs, the row menu, search; the narrow row wraps its actions under the title.

- The first CI run failed on three pinned expectations in `packages/server/test/skills.test.ts` (agent-development's four skill names and the counts built on them) and on the landing page's `skills-sync` test (which names every shipped skill); `4eb594ec` derives those counts from `libraryPlugin()` and lists `agent-porting` in both landing dictionaries — `test/skills.test.ts` 22/22 and landing 71/71 locally, and the rerun is green on every job. (Locally the server suite had passed against pnpm's injected copy of the plugin package, which predated the new skill directory; removing `node_modules/.pnpm/@penguinharness+agent-development@file+…` and re-installing reproduced CI.)

## Notes

- Base branch is `feat/ai-create-kit` (#583); the kit is used as-is apart from the additive `modelRef`.
- `benchmark-design` always calibrates with one run per case, so the AI tail says `runs = 1`; the manual form's "runs per case" is what the later optimization uses. `agent-optimization` reads the evaluation runtime from the baseline, so the Optimize dialog's model select only pins the optimizer conversation's model; the thinking level is backed by the agent's config rather than the draft, so it is not offered.
- The example prompts name the scenario, not an agent id, so they cannot contradict the picked Test Agent; the "AGENTS.md only" optimize example drops the brief's "at most 2 rounds" wording so it cannot contradict the round-limit field.
- Not done: the optional "evaluating / optimizing" row hint — no cheap way to detect a running session titled with the Benchmark.
- Backward compatibility: nothing to handle (no on-disk format changes; the create route writes the layout the readers already parse).
- Conflict hotspots: `packages/web/src/lib/strings.ts` / `strings-en.ts` (`benchmark` section replaced wholesale), `packages/server/src/api/types.ts` (Benchmark DTOs), `packages/web/src/api/endpoints.ts` (benchmark block), `packages/web/src/features/ai-create/ai-bridge.ts` (`modelRef`), docs `web-app` / `server-api` / `self-improvement`.

- Base branch is `feat/ai-create-kit` (#583); the kit is used as-is apart from the additive `modelRef`.
- `benchmark-design` always calibrates with one run per case, so the AI tail says `runs = 1`; the manual form's "runs per case" is what the later optimization uses. `agent-optimization` reads the evaluation runtime from the baseline, so the Optimize dialog's model select only pins the optimizer conversation's model; the thinking level is backed by the agent's config rather than the draft, so it is not offered.
- The example prompts name the scenario, not an agent id, so they cannot contradict the picked Test Agent; the "AGENTS.md only" optimize example drops the brief's "at most 2 rounds" wording so it cannot contradict the round-limit field.
- Not done: the optional "evaluating / optimizing" row hint — no cheap way to detect a running session titled with the Benchmark.
- Backward compatibility: nothing to handle (no on-disk format changes; the create route writes the layout the readers already parse).
- Conflict hotspots: `packages/web/src/lib/strings.ts` / `strings-en.ts` (`benchmark` section replaced wholesale), `packages/server/src/api/types.ts` (Benchmark DTOs), `packages/web/src/api/endpoints.ts` (benchmark block), `packages/web/src/features/ai-create/ai-bridge.ts` (`modelRef`), docs `web-app` / `server-api` / `self-improvement`.


hiyouga:
## Review

### Fixed in 026b2794

- **`packages/server/src/services/agent-porting.ts:405` — the import route inflated the whole zip before any cap applied.** `unzipSync` allocates each entry's declared uncompressed size up front, and the per-directory budgets below it only ever look at entries matching `skills/<n>/…` or `hooks/<n>/…` — anything else in the archive was never size-checked at all. A `payload.bin` of 30MB of zeros deflates to a few KB, so it clears `readArchiveBase64`'s 14MB cap and the import returned **201** with 30MB on the heap; at the cap that is ~14GB and the process is gone. The caps now come off the central directory through `unzipSync`'s `filter`, before a byte is inflated — a lying header buys nothing, since the allocation is that same declared size and fflate will not grow it. New test `refuses a zip bomb from its headers, without inflating it`; it returns 201 against the pre-fix code.
- **`packages/cli/src/commands/agent.ts:37` — `penguin agent export` wrote wherever the server's `Content-Disposition` pointed.** `resolveOutPath` joined the server-supplied filename into `--out` unfiltered, so a server answering `filename="../../.ssh/authorized_keys"` escapes it — and `--server <url>` deliberately targets a machine that need not be the user's. Only `path.basename()` of that name is used now; assertion added to the existing `resolveOutPath` test.
- **`agent-porting.ts:128` — a swallowed vault read.** `loadAgentVault(...).catch(() => ({}))` was the only catching call site of that function in the repo, and the function already returns `{}` for a missing file — so the catch could only fire on a corrupt `.vault.toml`, and then the export silently reports "no vault keys", which is precisely the field the importer is told to act on. Dropped.
- **`packages/cli/src/i18n.ts` — `agent.importFile` is dead.** Declared on `Messages` and translated in both dictionaries, referenced nowhere: `.command("import <file>")` takes no argument description, same as `schedule add <name>`. Removed from all three places.
- **`agent-import-modal.tsx:167` — the field label and the button inside it were the same string.** The picker rendered `<FieldLabel>Choose a file</FieldLabel>` directly above a control also reading "Choose a file". The snapshot picker this copies uses two strings ("Initialize from a snapshot" / "Choose a snapshot package"); added `importFileLabel` ("Bundle file" / 「包文件」) to both dictionaries.
- **`agents-page.tsx:661` — the card's Export button had no pending state.** Packing and fetching a bundle is a request whose only visible effect is the browser download starting, so the click reads as dead and repeat clicks fire repeat exports. Disabled while one is in flight.
- **`packages/docs/content/web-app.{en,zh}.md:91` — the colon introducing the tab table ended up on the wrong sentence.** The new prose was appended after "…organized into tabs:", so the table now reads as a list about Agent State snapshots. Moved the "clicking a card opens the settings page, organized into tabs:" clause back to the end of the paragraph, both languages.

### Findings not fixed

- **`packages/server/src/http/routes/skills.ts:75` — the identical zip bomb in the route this one was modelled on.** `parseSkillArchive` also applies its caps after `unzipSync`. Left alone: outside this PR's diff, one change one PR. The fix is the same `filter` argument.
- **`agent-porting.ts:101` — `redactMcpSecrets` only blanks `env` and `headers`.** `MCPServerConfig.config` is `Record<string, unknown>`: a hosted MCP entry's token usually lives in the `url` query string, a stdio entry's in `args` (`--api-key sk-…`), and both travel verbatim. Meanwhile the module header, the bundle README, the changelog and `exportAgentDesc` all say the bundle carries no secrets. What the policy should be — blank a whole `url`, strip matching query parameters, warn at export time — is your call, so I left the code; the absolute wording needs softening either way.
- **`agent-porting.ts:498` — hook packages round-trip through `strFromU8`, skills round-trip as bytes.** Every file in `hooks/<name>/` other than `hooks.json` and `icon.svg` is decoded to a UTF-8 string before `installHook` (core's parameter is `Record<string, string>`), so any byte that is not valid UTF-8 returns as U+FFFD. No shipped hook package carries a binary today, so it is latent, but export→import is lossy for a user-authored one. Fixing it means widening `installHook`'s signature in core.
- **A bundle is executable content and nothing tells the user.** `hooks/<name>/*.mjs` out of the zip are installed and later run with `spawn(process.execPath, [script])`, and `mcpServers` stdio entries carry a `command`/`args` the Session launches. That is what a hook package *is*, so it is not a defect — but "Import agent" takes a file from a stranger and `importAgentDesc` is entirely about porting convenience.
- **`agent-porting.ts:427` — `MAX_BUNDLE_ENTRIES` is still checked after the unzip.** Harmless now that the byte filter bounds memory; the count is available in the same filter if you want it there.
- **Spec parity.** `design/specs/05-ARCHITECTURE.md`'s route inventory lists the Agent State snapshot export/import but not `GET /agents/:agentId/bundle` or `POST /agents/import`, and neither `penguin-agent.json` nor the portable format appears anywhere under `design/specs/`. Per CLAUDE.md's 设计文档即准绳 this wants a design PR; I did not touch the design repo.

### Notes

- Extraction otherwise holds up: `assertSafeEntryPath` rejects absolute paths, drive letters, backslashes and every `..` segment before any use; directory names are held to `PLUGIN_NAME_PATTERN`; `installHook` re-checks with `assertSafeSkillFile`; symlinks are skipped on export (`isFile()`) and can never be recreated on import, since the writers only `writeFile`. The `id`→path question is closed one layer down — `createAgent` gates on `SEMANTIC_ID_PATTERN` before a directory exists, which is why the length-only check in `validateDefinition` is not exploitable. Import never overwrites (409 `agent_exists`, no force), and an unknown `format` is a 400 naming both the found and the expected value.
- Round-trip is otherwise complete: `name`/`description` share the same max lengths on both sides, `prompt`, `systemPrompt`, `model`, `mcpServers` and both directory sets survive. Two asymmetries beyond the hook-bytes one above: `tools` carries names only, so an Agent whose built-in tool `description`/`permission` were edited comes back with the target install's defaults on those rows; and `maxTurns` / `compaction.*` are not in the definition although `model.*` is, though all three sit on the same Runtime tab.
- The PR body is #583's Evaluation Center summary, pasted twice — only the CI paragraph about `skills.test.ts` and `skills-sync` belongs to this change.

Verified on 026b2794: `pnpm -r build` exit 0; `pnpm typecheck` exit 0; `pnpm --filter @prismshadow/penguin-server exec vitest run test/agent-porting.test.ts` 5/5 (and the new case returns 201 with the fix reverted); `pnpm --filter @prismshadow/penguin-cli test` 30 files / 390 tests (`FORCE_COLOR=0`; one file failed on an earlier run while the build was running concurrently, clean when run alone); `pnpm --filter @prismshadow/penguin-web exec vitest run test/agent-bundle-file.test.ts` 2/2; `pnpm --filter @prismshadow/penguin-docs test` 7 files / 53 tests; `pnpm format` + `pnpm format:check` clean, no file changed. Not verified: the server suite in full, web e2e, and any browser check of the dialog.

hiyouga:
## Correction to my review above

Two claims in it were wrong. I read `design/specs/` at its merged state and missed that the paired design PR was still open.

**Retracted: "Spec parity — this needs a design PR."** Wrong. Design PR [#101](https://github.com/Prism-Shadow/penguin-harness-design/pull/101) is open and MERGEABLE and specifies all of it: a new §「Agent 导入导出（移植）」 in `05-ARCHITECTURE.md` with `penguin-agent.json`'s `format: "penguin-agent/1"` field list, `GET …/agents/:agentId/bundle`, `POST …/agents/import` with `{dataBase64, agentId?}` and 409 `agent_exists`, the `penguin agent export|import` command surface, and the two route-inventory lines I reported as missing — added directly beneath the snapshot lines I quoted. `06-PROTOTYPE.md` covers the card's export button and the header's import dialog. The only thing that holds is the merge order: #101 lands first, per CLAUDE.md.

**Narrowed: the MCP redaction finding.** #101 defines the field as 「`mcpServers`（`env` / `headers` 中形似凭据的值置空）」 — those two, nothing more. So `redactMcpSecrets` **matches the spec**; it is not a policy gap for you to decide, and I was wrong to frame it as one. What was actually wrong is the prose, which promised more than the spec or the code delivers, so I have fixed it rather than leaving it with you.

### Fixed in e6aafedd

- **`agent-porting.ts:9`, `api/types.ts:2969`** — "never carries secrets" / "carries no secrets" now state the rule and its edge: `env` / `headers` blanking is the whole of the redaction, and a secret elsewhere in an MCP config (a token in a `url` query string, a key in stdio `args`) travels.
- **`agent-porting-templates.ts:51` — the bundle's own README.** This one mattered most: it is what whoever receives the bundle reads, and it said "vault values (secrets never travel)". The MCP note is now shared by both branches of `vaultSection` (it was duplicated, and only the vault-keys branch would have been the natural place to look) and ends with "read `mcpServers` before passing this bundle on".
- **`SKILL.md:8` and its Export section** — "Neither carries secrets" → "Neither carries vault values", with the check named where the skill tells the agent to hand a bundle over.
- **`packages/docs/content/cli.{en,zh}.md:171`** — "MCP entries with credential values blanked" implied all of them; now `env` / `headers` explicitly, both languages.
- **`changelog/unreleased/2026-09-02-agent-porting.{md,zh.md}`** — one added sentence, identical in both, naming what is blanked and what is not.

One correction to the correction: `S.agent.exportAgentDesc` and the `server-api` / `web-app` docs say "no vault values", which is accurate — I left them alone rather than churn correct copy.

Everything else in the review above stands, in particular the zip-bomb fix in `026b2794`.

hiyouga:
## Second regression test for the bundle extraction fix — `8186c2c0`

The test that shipped with `026b2794` used an honest 30MB-of-zeros archive, which a filter that merely bounded post-inflate totals would also have passed. It did not pin the case that actually requires reading the central directory, so here it is.

**The vector.** A zip's size fields are self-reported. `unzipSync` allocates the declared uncompressed size and then returns a view as long as what really inflated, so an entry of 5 bytes declaring 512MB costs 512MB of heap and still measures 5 bytes afterwards — every cap applied to decoded bytes waves it through. Measured locally on this branch: a **280-byte** archive, `payload.bin` declaring `536870912`, drives `arrayBuffers` 0.8MB → 537.7MB and `external` 2.4MB → 539.3MB while `entries["payload.bin"].byteLength` is `5` and its underlying `ArrayBuffer` is `536870912`. At ~81 bytes of zip overhead per entry that is ~181,000 declarations inside the route's 14MB cap.

**Why the fix already covers it.** fflate hands the filter `originalSize` from `zh()`, the central-directory record parser, at `browser.js:2677`, and allocates `new u8(su)` at `:2686` — the check is strictly upstream of the allocation, so the declaration is refused before any memory is taken. No change to `agent-porting.ts` was needed.

**What the new test adds.** `refuses an entry that lies about its size, which no later check can catch` builds that archive by rewriting the uncompressed size in the central directory record and the local header beside it, then asserts the archive is under a kilobyte, the import is 413 `bundle_too_large`, the message **names `payload.bin`** — so the refusal is pinned to the per-file declared-size branch rather than to any total measured later — and that no Agent was created. Against the pre-filter code at `4eb594ec` it fails with `expected 201 to be 413`: the archive imported and the Agent existed.

Verified: `pnpm --filter @prismshadow/penguin-server exec vitest run test/agent-porting.test.ts` 6/6; `pnpm --filter @prismshadow/penguin-server typecheck` clean; `pnpm format` + `pnpm format:check` clean, no file rewritten.

Note for whoever lands these: [#601](https://github.com/Prism-Shadow/penguin-harness/pull/601) factors an `unzipBounded` into `services/skill-import-limits.ts` on `main` and names this branch's `026b2794` and #592's `904ec022` as copies to collapse onto it once all three are merged. Nothing to do here — `main` has neither `agent-porting.ts` nor `hooks.ts` yet, so there is nothing to import from.

hiyouga:
Extended the export side so import and export are symmetric: both now offer a form path the server answers and a prompt path an agent answers.

**Import** already had its two (**From a file** / **Let AI import**) and is unchanged.

**Export** was a single download; it is now a three-segment dialog mirroring the import one:

| Segment | What it does |
| --- | --- |
| **API call examples** | The portable bundle as before: definition, `skills/`, `hooks/`, integration guide, endpoint reference, runnable curl / Python / TypeScript clients |
| **Docker files** | `Dockerfile`, `docker-compose.yml`, `entrypoint.sh`, `.env.example`, a run README and the endpoint reference |
| **Let AI export it** | The Create-with-AI panel for shapes neither kind covers — an SDK, Kubernetes manifests, an OpenAPI description, a handover document |

Behind it, `GET …/agents/:agentId/bundle` takes `?kind=api|docker` (`api` by default, an unknown value 400s) and the CLI takes `--kind`, rejected before a request goes out. **Both kinds carry the definition, `skills/` and `hooks/`**, so either zip still re-imports; only the documents packed around that core differ. The import dialog's id suggestion learned `-docker.zip` alongside `-export.zip` for the same reason.

The container installs the CLI, and on first boot writes the model configuration from the environment, starts the server, imports the agent and drops a sentinel so restarts skip the import. The data root is a volume, so removing it re-imports. Three things the README states outright rather than leaving to be discovered in production: it pins no version unless you pass `--build-arg PENGUIN_VERSION=<version>`, it terminates no TLS, and it authenticates nobody but the built-in admin.

## Verification

- `packages/server` — 1654 passed. New: `?kind=docker` packs the container files **and** the portable core while omitting the API examples; the entrypoint's shell expansions survive templating (`"$PENGUIN_HOME"`, `${PENGUIN_MODEL_API_KEY:-}`, no `undefined`); every declared vault key is listed in `.env.example` with no value; no secret appears in any entry; an unknown kind 400s.
- `packages/cli` — 390 passed. New: `--kind docker` sends `?kind=docker` and saves under the server's name; an unknown `--kind` is refused with no request sent. The fake server now names its download after the requested kind and records the query, so a client that forgot to send the kind fails the test instead of passing quietly.
- `packages/web` — 1635 passed; `packages/core` plugin loader, `packages/docs`, `packages/landing` green. Full build and `pnpm typecheck` clean; `prettier --check .` and `git diff --check` clean.

Specs (05 export section, API prototype line, command surface; 06 Agents-list page), the `agent-porting` skill (bumped to `2026-09-03.1`), the bilingual `cli` and `web-app` docs and the changelog pair are updated with it.

PR #595, opened 2026-09-02
The App Center: registry, penguin app, the app-center skill and the /apps page

hiyouga:
## Summary

The App Center: apps built in conversations are published to a page below Agents, listed with their live status and a restart / stop button each. An app is bound to the Session that built it — that Session is where restart / stop requests go, since it knows how the app runs.

- **Server** — a Project-level registry of `<project>/apps/<id>.toml` files (the file is the truth, the API a validated writer; nothing in SQLite): `GET|POST /api/projects/:p/apps`, `GET|PUT|DELETE …/apps/:id`, `POST …/apps/:id/actions`. Status is probed from `healthUrl ?? url` (HEAD, then GET; 2 s each; any HTTP response = running, refused / timeout = stopped, no URL = unknown), cached per URL for 10 s, re-probed on `?refresh=1`. An action composes an `[app_center]` origin block + instructions and delivers it to the owning Session as a new Task, a queued follow-up (compacting) or steering (running); a deleted Session answers 409 `app_session_missing`.
- **Core** — the `[app_center]` marker joins the origin blocks (builder, parser, title-noise list).
- **CLI** — `penguin app ls | register | unregister | status`; `register` binds to `PENGUIN_SESSION_ID` by default, lets the server fill agent / Workspace from that Session, derives an id from the name, and updates an existing `--id` in place.
- **Skill** — `app-center` in `software-development` (manifest `2026-09-02.2`): background run on a stable port, verify, register, keep current, carry out `[app_center]` requests.
- **Web** — the `/apps` page (nav + collapsed rail between Agents and the plugin library): search, status segment, rows with kind glyph / meta line / status pill, Open / Restart / Stop, a row menu (go to session, edit, unregister), auto re-probe every 20 s, and the kit's split "New app" button (Create with AI with the app-building tail and the three skills preselected; a manual form against a recent Session). The chat folds the `[app_center]` block into a one-line notice; input history skips it.
- **Docs** — Web App, CLI, Server API and Skills pages (en + zh); changelog pair `2026-09-02-app-center`.

## Design

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/103 (05: `### 应用中心（App Center）`, API lines, `penguin app` in 命令面, 插件库; 06: nav table + rail order, 对话页 folded block, `### 应用中心页`).

## Verification

- `pnpm --filter @prismshadow/penguin-core build`, `…penguin-server build`, `…penguin-web build` — ok
- `pnpm typecheck` (whole graph) — exit 0
- `packages/core`: `npx vitest run` — 49 files passed, 1 skipped; 1064 tests passed, 5 skipped
- `packages/server`: `npx vitest run` — 118 files, 1664 tests passed (new: `app-registry`, `app-probe`, `app-actions`, `apps` route tests)
- `packages/web`: `npx vitest run` — 123 files, 1639 tests passed (new: `apps-model`; `nav-group-collapse` updated)
- `packages/cli`: `npx vitest run` — 30 files, 392 tests passed (new: `app-command`)
- `packages/docs`: `npx vitest run` — 7 files, 53 tests passed
- `packages/landing`: `npx vitest run` — 10 files, 71 tests passed (after adding `app-center` to the Skills section, which its sync test requires)
- `packages/server`: `npx vitest run test/skills.test.ts test/apps.test.ts` — 28 tests passed after the plugin-enumeration expectations moved to three skills
- `pnpm format:check` clean; `git diff --check` clean
- Manual: built server on a scratch data root; registered three apps through the API (statuses probed running / stopped / unknown; TOML carries the Session-derived agent and Workspace); screenshots of the page (zh, en, 390 px), the row menu, the AI dialog and the manual form; a `POST …/actions` restart was delivered as a Task and the `[app_center]` notice rendered in the chat.

## Notes

- Base branch is `feat/ai-create-kit` (#583); the orchestrator rebases onto main after the kit merges.
- Decisions where the brief was open: writes (register / edit / unregister) are **owner-only, mirroring schedules**; reads and restart / stop actions are for any member. Registering an explicit `--id` that exists **updates in place** (POST 409 → PUT in the CLI). `sessionId` must name a Session of the Project (400 `app_session_unknown`); an id derived from a name gets `-2`, `-3`… on collision. The `[app_center]` body (the instructions) renders as the message body under the notice, exactly like a scheduled task's prompt.
- `plugins/software-development/package.json` keeps its version (release prep bumps plugin versions to the repo version); only its description changed.
- Two CI-only failures were fixed in follow-up commits (`918071af`, `9b01a12f`): the landing page's Skills-section sync test and the server `skills.test.ts` plugin enumeration. Both passed locally at first because the local checkout resolves `@penguinharness/software-development` through pnpm's injected `file:` copy, which never receives a *new* skill directory (only hard-linked edits to existing files) — CI's fresh install sees the third skill. Removing `node_modules/.pnpm/@penguinharness+software-development@file+…` makes local resolution fall through to the workspace directory.
- No backward-compatibility handling: everything here is new (a new directory under the Project, a new marker tag that no existing Trace carries).
- Conflict hotspots: `strings.ts` / `strings-en.ts` (`nav.apps`, new `apps` section, `chat.appCenterAction`), `packages/server/src/api/types.ts` (new App Center section before the snapshots one), `packages/server/src/app.ts` (`AppDeps.appProbe`, `BuildDepsOverrides.appProbeFetch`, route mount), `packages/web/src/api/endpoints.ts`, `components/ui/icons.tsx` (`NAV_ICONS.apps`), `router.tsx`, `app-layout.tsx`, `lib/nav-group-collapse.ts` + its test, `features/chat/{message-item.tsx,user-message-body.ts,input-history.ts,agent-handoff.ts}`, core `markers/{tags,origin-blocks}.ts` + `markers.test.ts`, `packages/cli/src/{i18n,index}.ts` + `test/fake-server.ts`, the four docs pages, `plugins/software-development/plugin.json`, and in the design repo `05-ARCHITECTURE.md` / `06-PROTOTYPE.md`.


hiyouga:
## Review

Reviewed the incremental diff against `feat/ai-create-kit` only.

### Fixed in 14aab79f

- **`packages/web/src/features/apps/apps-model.ts:86` — `relativeAge` renders "0 years ago" for an app registered 360–364 days ago.** Every step of the chain is bounded in days except the month one, which is bounded in months: at `days = 361`, `months = Math.floor(361 / 30) = 12`, so `months < 12` is false and the year branch computes `Math.floor(361 / 365) = 0` → the row's meta line reads `Registered 0 years ago` / `注册于 0 年前`. The existing test jumps 45 d → 400 d, straight over the gap. The month step is now bounded in days like the rest (`days < 365`), and `test/apps-model.test.ts` gained a 364-day case (`12 months ago` / `12 个月前`).

- **`packages/web/src/features/apps/apps-page.tsx:87` — a failed list load left the page in a permanent skeleton and toasted the same error every 20 s.** `load()` only called `toastError`, so `data` stayed `null` and the skeleton never resolved; because `load` is also the `setInterval(…, REFRESH_MS)` body, an unreachable server produced one toast per tick for as long as the page stayed open (stop the server with `/apps` open to reproduce). The failure is now held in an `error` state: a first-load failure renders in place of the skeleton (`plugins-page.tsx`'s idiom, `text-sm text-red-600 dark:text-red-400`), and a later poll failure is left silent so the rows already listed keep standing.

- **`packages/web/src/features/apps/apps-page.tsx:175,283,398` — the page offered members three actions the API refuses.** `POST` / `PUT` / `DELETE /apps` are `requireProjectOwner` (and design 05 says so: 「登记、修改与取消登记仅 owner（同定时任务）」), but "New app", the empty state's "Create with AI", "Edit details" and "Unregister" rendered for everyone — a member who fills in the whole manual form gets a 403 at Save. All four are now gated on `currentProject?.role === "owner"`, the way `schedules-tab.tsx` gates its create/edit/delete. `packages/docs/content/server-api.{en,zh}.md` gained the matching "writes are owner-only" sentence that the schedules section directly above it already carries.

- **`packages/web/src/features/apps/app-form-modal.tsx:66` — the manual form's owning-Session default was frozen at mount.** `useState(app?.sessionId ?? candidates[0]?.sessionId ?? "")` evaluates once; open "Set up manually" before the sidebar's Session list resolves (a cold load straight onto `/apps`) and `sessionId` stays `""` after the options arrive — the `<select>` shows a session, `ready` is false, and Save is disabled with nothing explaining why. The first candidate is now the fallback at render time, so the select and the submitted body agree.

- **`packages/server/src/http/routes/apps.ts:103,276` — dead null path in `toItem`.** It took `AppDefinition | null` and returned `AppItem | null` for a `null` no caller could produce: the list already filtered on `e.parsed.ok` before the `e.parsed.ok ? e.parsed.def : null` ternary, and `readItem` asserted the result away with `!`. Collapsed to non-null in and out; the `filter((a): a is AppItem => a !== null)` and the `!` are gone.

### Findings not fixed

- **`packages/server/src/http/routes/apps.ts:244` — a web-form edit silently discards a `--workspace` override.** `write()` fills `workspace` from the owning Session whenever the body omits it, and `AppFormModal` never sends it (the field is not in the form and not displayed), so PUT is a full replacement that resets it. Reproduce: `penguin app register --id x --name X --workspace /srv/x`, then Edit → Save in the page; the file's `workspace` becomes the Session's. Left for you: the fix is a product call (round-trip the field in the form, or have PUT keep the stored value when the body omits it), and `apps.test.ts` currently asserts the re-derivation as intended behaviour.

### Notes

- **No download or extraction surface exists in this PR**, so the archive questions do not apply: the registry is local TOML under `<project>/apps/`, and the diff adds exactly one outbound `fetch` — the health probe. No archive handling, no `child_process`, no remote registry URL.
- **Path traversal on the app id is closed.** `requireAppId` runs `isValidId` (`^[A-Za-z0-9_-]+$`) before any path is built, on the path param and on an explicit body `id`; a derived id comes from `slugifyAppId`, whose output is in the same charset and never empty. `apps.test.ts` covers `id: "../x"` → 400.
- **The probe's SSRF shape is a recorded decision, not an oversight.** `app-probe.ts:84` issues a server-side HEAD/GET at any absolute http(s) URL a registration carries, and the API hands back a running/stopped oracle that `?refresh=1` re-arms — design 05 states the trade-off explicitly (「地址是用户自己的应用，不限主机、只收紧超时」), and `redirect: "manual"` stops a redirect pivoting off the app's origin. Flagged only so the decision stays visible.
- `apps.ts:130` — two concurrent `POST /apps` with the same `name` (or the same explicit `id`) both clear the "is this id free?" read before either writes, so one registration overwrites the other instead of getting `-2` / a 409. Check-then-write with no lock; consistent with the file-is-the-truth, last-write-wins model the whole registry uses, so not worth a lock on its own.
- `S.apps.emptyDesc` still ends "you can also have the AI create one from scratch" for a member who no longer has that button. `schedules-tab.tsx` covers the same case with a `readOnlyHint`; a string pair would close it.
- `relativeAge` repeats the shape of `lib/format.ts`'s `relativeShort` / `daysAgoLabel` / `updatedDaysAgo` — same hardcoded-bilingual convention, longer scale. It reads fine where it is; `lib/format.ts` is where its three siblings live.
- Spec parity is clean against design PR #103: nav and rail order, the status tones (`success` / `muted` / `attention` via `Badge`'s green/gray/amber), the probe semantics and cache, the meta line, the `[app_center]` fold, and the CLI surface all match 05/06 as written there. Changelog pair `2026-09-02-app-center.{md,zh.md}` is present, symmetric and carries the metadata block in the fixed order.

Verified (Node v24.18.0, `FORCE_COLOR=0`): `pnpm -r build` — exit 0. `pnpm typecheck` — exit 0 across all 19 projects. `pnpm format` (no file changed) + `pnpm format:check` — "All matched files use Prettier code style!". `packages/server`: `npx vitest run test/apps.test.ts test/app-registry.test.ts test/app-probe.test.ts test/app-actions.test.ts` — 18 tests passed. `packages/web`: `npx vitest run test/apps-model.test.ts test/nav-group-collapse.test.ts test/placeholders-parity.test.ts test/disclosure-anchor.test.ts` — 23 tests passed. `packages/docs`: `npx vitest run` — 53 tests passed. Not run: the Playwright e2e suite (a known ~8-spec local baseline failure on `main`) and the full `packages/{web,server,cli}` unit suites — the CLI is untouched by these fixes.

hiyouga:
## Follow-up: the two "Create with AI" family items (7edb2c55)

My earlier review stands; this is only what the cross-PR audit of the six kit surfaces turned up.

### Fixed

- **`apps.aiCreateTail` (both dictionaries) — the tail named `penguin app register` but left its flags to prose.** `--name` is a `requiredOption` (`packages/cli/src/commands/app.ts:99`), so "register it with `penguin app register` (with its URL and start / stop commands)" costs the agent a `--help` round trip before it can run anything. The tail now carries the invocation in full — `penguin app register --name "<display name>" --url http://localhost:<port> --start-command "<command that starts it>" --stop-command "<command that stops only this app>" --kind web|api|cli|other` — and says `--name` is required and that the owning Session, its agent and its Workspace come from the calling Session. Placeholders match the ones the `app-center` SKILL.md already uses, so the two agree.
  **No `--root`**, deliberately: `penguin app` talks to the server, `resolveConnection` reads the `PENGUIN_API_URL` / `PENGUIN_API_TOKEN` / `PENGUIN_PROJECT_ID` the server injects, and the command declares no such option — adding it would be a flag that does not exist.

- **`apps.aiExamples` (both dictionaries) — the four examples were the only ones in the family without a `description`.** `AiExample.description` is the muted second line `AiCreatePanel` renders under the label (`ai-create-panel.tsx:24,146`), so the App Center's cards were one line tall where all five sibling surfaces' are two. Each example gained a phrase (not a bare value): "An Express API, a React frontend and SQLite storage", "Personal bookmarks, kept with tags and search", "Upload a CSV file, get charts back", "A submission page that compiles the entries into Markdown", plus the zh equivalents. The `as ReadonlyArray<{…}>` annotation in both dictionaries was widened to keep the field.

### The two left to you

- **A8, `primaryAction` on `CreateMenuButton`.** I did not build a second memory. My read: remembering the last choice belongs in the kit, not in the six call sites — the preference is about the split button, not about apps versus agents versus models, and six copies is six places to drift. But I would key it **per surface** rather than globally: the App Center's manual form registers an app that is already running (a fallback), while the Agents page's manual form is the ordinary path, so one shared "last choice" would flip this page's main half because of something the user did elsewhere. If you'd rather have one global preference, the App Center is fine either way — the point is that the decision is made once, in the kit.

- **A10, `allowAgentChoice`.** I think it should be on here, and it is genuinely one prop: `AiCreateModal` already owns the pick (`picked` state, `onAgentChange={setPicked}` handed to the panel at `ai-create-modal.tsx:104`), so the flag alone turns the picker on. The App Center's agent is only "who does the building" — the app binds to the *Session*, not to the agent, and any agent in the Project can build one, so a user with a coding-specialised agent has a real reason to redirect the work. I left it unset because the audit found no stated rule and three of six surfaces pass it; making this a fourth data point would settle the rule by accident. One prop, whenever you want it.

Verified: `pnpm --filter @prismshadow/penguin-web typecheck` — exit 0. `packages/web`: `npx vitest run test/apps-model.test.ts test/placeholders-parity.test.ts` — 9 passed. `pnpm format` (neither dictionary changed) + `pnpm format:check` — "All matched files use Prettier code style!".

PR #596, opened 2026-09-02
Evaluation Center: Benchmarks become Project-level peers of agents, created with AI or by hand and optimized against

### LINKED ISSUE #564: 评估中心
### The problem

Image

这里好像没有+ 可以手动创建评估

希望增加手动创建评估，agent不同的评估更具轨迹

### What you do instead today

。

hiyouga:
Fixes #564. Reworked on 2026-09-12 on the maintainer's instruction: a Benchmark is a peer of an agent, not something an agent owns.

## The change

- **Project-level Benchmarks.** `agents/<agent>/benchmarks/<id>/` becomes the Project's own `benchmarks/<id>/`, a sibling of `agents/`. `benchmark_config.toml` records no agent; one Benchmark can evaluate several agents and one agent can be evaluated by several Benchmarks. Files stay the single source of truth.
- **The label.** Every `scoreboard.yaml` evaluation carries the tested agent as `agent_id` (right after `time`); with the Agent State `version`, the `(provider, model_id)` pair and `thinking_level` it labels the record. The chart plots time against score and draws one series per label (`<agent_id> · v<version> · <model_id> · <thinking_level>`, grey "未标注" for records missing a part); the evaluation table gains a tested-agent column; a card's score change is measured against the previous evaluation of the same label.
- **The page.** No agent grouping: one flat card list, each card with the agents it has evaluated (avatars), the sparkline, the latest score and change, and ghost-button actions like the scheduled-tasks rows. `?agentId=` filters to Benchmarks that evaluated that agent. A Benchmark lives at `/benchmark/:benchmarkId`. The page-level pairs (create, optimize) are the kit's `CreateButtons` at the size the scheduled-tasks tab uses.
- **Forms.** Manual creation no longer names an agent. The optimize form picks the **tested agent** (default: the agent of the latest evaluation, else the current agent); the AI-optimize tail carries the same default. The AI-create modal keeps its tested-agent picker (a baseline needs one).
- **Server.** Routes move to `/api/projects/:projectId/benchmarks` (`GET`, `POST` owner, `DELETE …/:id` owner, `…/:id/cases`, case file routes); membership and ownership are checked at the Project level. DTOs: `BenchmarkEvaluation.agentId: string | null`, `BenchmarkSummary.agentIds: string[]`.
- **Core.** `benchmarksDir(root, projectId)`; the example Benchmark is provisioned at the Project level when `default_agent` initializes, its three sample evaluations labelled `agent_id: default_agent`.
- **Skills.** `benchmark-design`, `agent-evaluation`, `agent-optimization` read and write `<app_data_dir>/benchmarks/<benchmark_id>`, state the peer relationship, and carry `agent_id` in every scoreboard entry and in the evaluator's result protocol. `agent-tuning` plugin version `2026.09.12.1`.
- Docs (`web-app`, `server-api`, `self-improvement`, both languages) and the changelog pair rewritten.

## Backward compatibility — the maintainer's call

Existing `agents/<agent>/benchmarks/` directories (including every existing data root's `example-benchmark`) are **not read** by this version; nothing migrates or deletes them, and an existing Project gets no Project-level example (it is seeded only when `default_agent` initializes). Options, none implemented here: (1) permanent dual read, labelling the agent from the legacy path; (2) a one-time migration moving each legacy directory into `benchmarks/` and stamping `agent_id` into its evaluations (colliding ids need a rule); (3) accept the break and say so in the release notes. Please pick one; the entry and this PR will follow.

## Design

Prism-Shadow/penguin-harness-design#102 — 05 § Benchmark 评测与优化 (Project-level location, `agent_id`, the label rule, the API), 06 § 评估中心 (rewritten), 01 PRN-005 (the tree).

## Verification

Remote test host (Node 24.20): `pnpm typecheck` clean across every package after building core, server, web and cli; `pnpm format:check` clean.

- core `test/example-benchmark.test.ts`: 5 passed
- server `test/benchmarks.test.ts` + `test/builtin-agents.test.ts`: 12 passed
- web full suite: 135 files, 1873 passed
- docs: 7 files, 53 passed; landing: 10 files, 71 passed; core `test/plugins.test.ts` (the plugin loader over the changed skills): 17 passed

## Notes

- The chart's x-axis is evaluation order labelled with timestamps (as before), not a proportional time scale.
- Older English copy still says "Test Agent" in a few keys while the new keys say "tested agent"; a copy sweep is a separate decision.
- Conflict hotspots: `packages/web/src/lib/strings.ts` / `strings-en.ts` (`benchmark` section), `packages/server/src/api/types.ts`, `packages/web/src/api/endpoints.ts`, `packages/core/src/state/paths.ts`, the three SKILL.md files, docs `web-app` / `server-api` / `self-improvement`.

## Round 5 (2026-09-14)

- The intro block above the step cards (the how-to line and the agent-tuning plugin note) is removed, and so are the grey Skill names beside the step titles. The three cards now carry a number, a title and one sentence each. Strings, docs and the changelog pair follow; design PR #102 updated to match.
- Verified on the remote test host: web `typecheck`, `format:check`, and the full web suite (138 files, 1906 tests) pass.

## Round 6 (2026-09-14)

- The three step cards now say where on the page each step happens: Create with AI / Create manually at the top right; pick a Benchmark and press Use → Evaluate for a score; pick a Benchmark and press Use → Optimize to raise it. Both dictionaries, the docs and the changelog pair follow.
- The Use dialog's conversation-model field on both tabs is the Project settings' own model picker (`ModelSelect`, form variant, the same one the new-chat defaults and the schedule form use), preset to the Project's default model. The native select with its "Project default (…)" row is gone, and so are the two strings that named it. Design PR #102 updated to match.
- Verified on the remote test host: web `typecheck`, `format:check`, and the full web suite (138 files, 1906 tests) pass.

## Round 7 (2026-09-14) — draft / published status

- `benchmark_config.toml` gains `status = "draft" | "published"` (`BenchmarkSummary.status` on the API). `benchmark-design` writes `draft` when it creates the directory and flips it to `published` right after the Formal Baseline is appended and verified; a `calibration_failed` Benchmark stays `draft`. The hand-made create route and the built-in example write `published`. `agent-optimization` refuses a draft. Plugin `agent-tuning` → 2026.09.14.1.
- Evaluation Center: a draft card is masked — dimmed under an overlay reading "构建中 / Being built" with the hint that the agent is still writing and calibrating, the card's main button plus Use and View are disabled, and only the owner's delete icon stays live above the mask so a failed calibration can be cleaned up. A draft reached by its address shows the same notice in place of the cases and the chart, with no Use button.
- **Compatibility (please confirm):** only a literal `status = "draft"` locks a Benchmark; a config without the field (every Benchmark written before this round) or with any other value reads as `published`. This is a permanent default of the schema rather than a temporary shim, so no backward-compatibility changelog file was added. If you would rather have existing Benchmarks treated differently, say so.
- Verified on the remote test host: build + typecheck of every package, `format:check`, core example-benchmark + plugins tests (24), the full server suite (125 files / 1825, one new test `reads status from the config: only a literal draft locks, everything else is published`), the full web suite (138 files / 1906), docs tests (53).

## Round 8 (2026-09-14) — shared file browser, Evaluations folder

- **One file browser.** New `components/ui/file-browser.tsx`: the Workspace files panel's `FileTree` on the left and a read-only preview pane on the right (text via the chat code block, Markdown through the shared plugin pipeline with the chat's code-block chrome and file-relative links / images, image, PDF, download link, truncation note). The Benchmark case dialog is rewritten on it — task materials and rubric are the tree's two top-level folders, listed level by level like the Workspace panel, file sizes beside the names, README auto-previewed, a Markdown link to another case file opens it in place — and the plugin detail Modal uses the same component, so the three browsers share one style. The case-tree row builder is pure and unit-tested.
- **Evaluations folder.** Sessions gain a third origin, `benchmark`, end to end: core `session_meta.source`, the server's `SessionSource` / categories / counts, `POST …/sessions` accepting `source: "benchmark"` (the only value a client may set; `subagent` / `schedule` stay server-written), `penguin run --source benchmark` in the CLI, and `agent-evaluation` launching every Test Session with it (plugin `agent-tuning` → 2026.09.14.2). The Use dialog's conversation is created with the same mark via the AI-create draft. The sidebar files them under a new **评估任务 / Evaluations** folder between Scheduled and Archived (per Agent, per Workspace group and in time mode alike), and a benchmark Session is never auto-opened as the last conversation. Docs: web-app, cli, server-api, self-improvement (en + zh). Design PR #102 updated.
- Verified on the remote test host: build + typecheck of every package, `format:check`, core (1165), server (1826, incl. the new `a Session created with source benchmark lists under the benchmark category`), cli (394), web (1913, incl. the new folder and case-tree tests), docs (53).

## Round 9 (2026-09-14) — failed status, folder-only groups, harder examples, Ask AI default

- **`status = "failed"`.** `benchmark-design` writes it when calibration fails (`calibration_failed`) instead of leaving the Benchmark on `draft`; the card and the page show a creation-failed notice (danger ink) asking to delete and recreate, Use/View stay disabled, delete stays live. `agent-evaluation` returns `benchmark_invalid` on it and `agent-optimization` stops on it. Reader rule: literal `draft` / `failed` are themselves, everything else is published. Plugin `agent-tuning` → 2026.09.14.3.
- **Folder-only groups fold last.** A sidebar group with no active conversation of its own — only rows in its Subagents / Scheduled / Evaluations / Archived folders, which is what an evaluation's per-case Workspaces and an agent that only ran evaluations look like — now collapses by default and sorts after the other groups of its mode (pinned groups excepted), in workspace and agent grouping alike. Its header is dimmed and counts the folded rows; opening it once persists per Project (`penguin.sidebarExpandedFolderGroups.<projectId>`, classified install-scoped). `orderGroups` gains a `demote` option, unit-tested in `group-order.test.ts`.
- **Harder AI-create examples.** The four scenarios are rewritten in the draft page's structured style (Benchmark id, capability, case count, techniques, desired baseline, pilot limit): three cases each, hidden preconditions, vague or conflicting input, strict deliverables, `desired_baseline_score < 50`; the fixed tail now lets the draft override the pilot limit as it already did the baseline.
- **Ask AI default first.** Both Ask AI dialogs list their default question as the first example button, so it can be brought back after trying another; the test pins that the first example equals the default.
- Verified on the remote test host: build + typecheck of every package, `format:check`, core (1179), server (1845), cli (394), web (1967), docs (53). Design PR #102 updated.

## Round 10 (2026-09-15) — agent-evaluation in every phase, plain-text examples

- **The model is made to use `agent-evaluation`.** Each Benchmark flow now opens its conversation with the Skills it rests on preselected through the AI-create bridge (`skills` → the draft → the `[use_skills]` block on send): create with `benchmark-design` + `agent-evaluation`, evaluate with `agent-evaluation`, optimize with `agent-optimization` + `agent-evaluation`. The composer only keeps Skills the chosen agent has installed, and the Use dialog already warns about a missing one. All three fixed tails also say in as many words that every evaluation goes through a `run_subagent` subagent told to use `agent-evaluation`, never scored by the model itself and never bypassed.
- **Examples back to plain prose.** The four create examples are one- or two-sentence scenario descriptions again (what contradicts, what is hidden, what is vague); the structured requirements moved into the fixed tail — about three cases, few and hard; the techniques; `desired_baseline_score < 50`; `pilot_iteration_limit` 4 — each still overridable by the draft.
- Verified on the remote test host: web `typecheck`, `format:check`, the full web suite (143 files / 1967). Design PR #102 updated.

## Round two (2026-09-12, maintainer feedback)

- **Flow block.** The folded guide is a standing block under the title: three steps side by side (出题 / `benchmark-design` → 评估 / `agent-evaluation` → 优化 / `agent-optimization`), a "怎么用" line (create → press 使用 on a Benchmark → compare series by label) and the plugin note.
- **Use dialog.** A card's Optimize is a dark **使用** button (`variant="primary"`); the dialog switches between an **评估** tab — tested agent, evaluator agent (warns when it lacks `agent-evaluation`), the conversation's model, runs per case, an optional note; its tail has the evaluator run the full Case × runs matrix through self-spawned `agent-evaluation` subagents, check every result's label, and append exactly one labelled evaluation without touching the agent or the Benchmark — and the **优化** tab (the previous manual form). The free-text optimize mode is gone; both tabs end in the kit's "在新对话中编辑" exit.
- **Card actions**: 使用 / 查看 / the agent list's delete icon button (owner). The overflow menu is gone; the Benchmark page header shows `benchmarks/<id>` with a `CopyButton` and the same 使用 button.
- **Example Benchmark for every Project.** It is seeded whenever the Project has no `benchmarks/` directory, on `default_agent` initialization or load (directory-level check: deleting the example leaves the directory, so it does not come back). This is the maintainer's answer to the legacy question above: existing data roots get the example at the Project level on their next load; legacy `agents/<agent>/benchmarks/` directories are still not read.
- **Frozen cases.** A Benchmark's statements and rubrics cannot be changed after creation — no edit path exists in the UI or the API, and the docs say so.
- Docs (`web-app`, `self-improvement`, both languages) and the changelog pair updated.

Verification (remote test host, Node 24.20): `pnpm typecheck` clean across every package, `pnpm format:check` clean; core `example-benchmark` 7 passed; server `benchmarks` + `builtin-agents` 12 passed; web full suite 135 files, 1875 passed; docs 53 passed.

## Rounds three and four (2026-09-12, maintainer feedback)

- **Intro block + three step cards.** The how-to line (create with AI or by hand → press 使用 on a Benchmark → compare series by label) and the plugin note sit in one standing intro block; the three steps (出题 / 评估 / 优化) are three separate cards beneath it, three across on a wide screen, nothing repeated below them.
- **One series per agent, model and thinking level.** The Agent State version no longer splits a series: the same tested agent on the same model and thinking level draws one connected line across versions, the version marked on each point's hover bubble and in the table — so the example Benchmark's three evaluations connect. Provider does not split either. A card's score change is measured within that series.
- **Evaluation detail dialog.** Clicking an evaluation row opens a dialog shaped like the case dialog: series label, version and tested agent; Score / cost / duration; the summary title and body; the per-case table with run rows and Session ids.
- **问 AI.** Both the evaluation dialog and the case dialog carry an ask-AI button that opens the kit's dialog with a default question and examples; its fixed tail hands the default agent the record's structured facts (Benchmark id and path, time, label, version, runtime, scores per case, Session ids, summary — or, for a case, both README paths and the latest run results) and asks for an explanation without changing anything. The exit is the kit's 在新对话中编辑.
- Branch merged with `main` (the #699 button rungs; `CreateButtons` defaults to `sm`) — no conflicts.

Verification (remote test host, Node 24.20): `pnpm typecheck` clean across every package; `pnpm format:check` clean; core `example-benchmark` 7; server `benchmarks` + `builtin-agents` 12; web full suite 138 files, 1906 passed (the button-rung guard included); docs 53.


hiyouga:
## Review

### Fixed in 5a523b7d

- **`packages/web/src/features/benchmark/create-benchmark-modal.tsx:238` — per-case validation errors are keyed by array position, so removing a case moves another case's errors onto it.** `validate()` builds `cases: cases.map(() => ({}))` and the render reads `errors.cases[i]`, but `CaseDraft.key` is the stable identity, not `i`. Repro: add a second case, leave case 1's rubric blank, fill case 2 completely, press Create — the error lands on case 1's rubric. Now delete case 1: case 2 slides to index 0 and inherits that stale "required" message on a rubric that is filled in, while nothing marks the case that actually failed. `FormErrors.cases` is now a `Record<CaseDraft["key"], CaseErrors>`, so an error stays attached to the case it was raised for.

- **`packages/server/src/http/routes/benchmarks.ts:163` — `runs` had a floor but no ceiling, and the Optimize dialog only accepts 1–1000.** Nothing in the manual form or the route stopped `runs: 5000`; the server wrote it into `benchmark_config.toml` and `readBenchmark` reads back any integer ≥ 1. Open Optimize on that Benchmark afterwards and `intIn(runs, 1, 1000)` returns null, so the runs field opens already in error and both **Send** and **Edit in chat** stay disabled until the number is retyped — a Benchmark the product created cannot be optimized through the product. `runs` also multiplies the Case × runs matrix the optimizer evaluates every round, so an unbounded value on a write route is a cost footgun in its own right. The route now rejects `runs > 1000` with a 400, the form validates against the same bound (`isValidRuns`), and the dialog clamps its initial value so a hand-edited `benchmark_config.toml` still opens usable. Copy in both dictionaries names the range. Tests: `runs: 1001` added to the route's malformed-request list, plus an `isValidRuns` case in `benchmark-prompts.test.ts`.

- **`packages/web/src/features/benchmark/optimize-modal.tsx:104` — the skills-cache effect threw away a response it still wanted.** The result is stored under the agent id it was fetched for, so it is never stale; the `cancelled` flag only meant that picking optimizer A, switching to B, then back to A dropped A's in-flight response and refetched it. Guard removed.

- **`optimize-modal.tsx` — `FullPromptFold` was a verbatim copy of the fold in `ai-create-panel.tsx`.** Exported once from the kit as `PromptFold` and used in both.

- **Three error lines used raw palette classes instead of a tone token** (`benchmark-page.tsx:503`, `benchmark-detail.tsx:360` — `text-red-500`, which is not in the palette at all; `create-benchmark-modal.tsx:323` — `text-red-600 dark:text-red-400`). All now `toneInk.danger`.

### Findings not fixed

- **`packages/server/src/services/benchmark-service.ts:311` — `remove` deletes the directory an evaluation may be writing into.** `fs.rm(benchDir, { recursive: true, force: true })` runs with no check that a Session is using the Benchmark; an `agent-evaluation` run in flight keeps writing to the same paths and leaves a partial directory behind, which `list` then shows with a fallback title. There is no Session-to-Benchmark link to test against — the PR's own "Not done" note says the same about the row hint — so this needs your call on whether a delete should be refused, warned about, or left as is.

- **`packages/web/src/features/benchmark/benchmark-page.tsx:298` — one `listBenchmarks` request per agent, fired unbounded on mount.** Each one is a `readdir` plus a TOML + YAML parse per Benchmark on the server. A Project with many agents issues that many requests on every visit to the page. Bounding it means a Project-wide list route, which is a design change rather than a review fix.

- **Spec parity.** `design/specs/06-PROTOTYPE.md` §评估中心 still describes the page this PR replaces: 「曲线支持在**分数 / 成本 / 耗时**三个指标间切换」, series split by `(provider, model_id)`, and 「run 的 session 链接**直达该 Session 的轨迹观测**」. The code is score-only, keys series by `(model_id, thinking_level)` with provider deliberately excluded (`benchmark-metrics.ts:72`), and renders the session id as plain text (`benchmark-detail.tsx:257`). All three predate this PR, but this PR is where `packages/docs/content/web-app.*` drops the metric switcher, so the docs and the spec now say different things. `05-ARCHITECTURE.md:1445` also still lists only the GET route. penguin-harness-design#102 is open, not merged — this should not land ahead of it.

### Notes

- Create/delete on the server holds up: the non-recursive `mkdir` as the atomic existence check is correct, the half-written directory is cleaned up, `lstat().isDirectory()` before `rm` refuses to follow a symlink placed at `benchmarks/<id>`, and `isValidId` / `CASE_ID_PATTERN` leave no way to escape the agent's directory. No path traversal found.
- Nothing in the scoring path substitutes zero for an error — `scoreValues`, `sparklineSeries` and `latestScore` all drop non-finite values rather than counting them.
- The validation on the create route covers only the manual path. "Create with AI" never reaches the server: the `benchmark-design` Skill writes the directory itself, so a model-written layout is bounded only by `readBenchmark`'s tolerant parsing. That is the existing architecture, not something this PR changed, but it is worth being explicit about since the PR adds the validated route beside it.
- i18n is symmetric (the `Strings` type enforces it) and no user-visible literal is hardcoded in the new components.
- I reviewed the UI statically and did not render the page; the visual claims in the PR description are not independently confirmed here.

Verified: `pnpm typecheck` clean; `pnpm --filter @prismshadow/penguin-web test` 123 files / 1648 tests passed; `pnpm --filter @prismshadow/penguin-server exec vitest run test/benchmarks.test.ts` 6 tests passed in 4.4s; `pnpm format` + `pnpm format:check` clean. `pnpm -r build` exited 0 on the branch head before these fixes and was not re-run — nothing here changes a cross-package surface (the route edit is internal to `packages/server`, the rest is `packages/web`), which typecheck covers. Under load the box produced `[vitest-worker]` RPC timeouts and a 10s `beforeEach` hook timeout in the server suite, with the cascading `database is not open` / `user_exists` teardown errors that follow one; zero assertions failed, and the same file passes 6/6 in 4.4s on the idle box.

hiyouga:
## Half-deleted Benchmarks no longer list (ec38621c)

`BenchmarkService.list` now skips a `benchmarks/<id>/` directory that has no `benchmark_config.toml`.

**What gets hidden.** Deleting a Benchmark while an `agent-evaluation` run is still executing against it does not stop the run — it keeps writing to the paths it was just deleted from, and recreates a partial directory. `list` presented that as a Benchmark row, using the directory name as a title because no config was there to supply one. That row is gone.

**What still lists.** Everything that has a config. In particular a Benchmark created and never evaluated: no `scoreboard.yaml`, no evaluations, zero cases if you like — it keeps its config and lists exactly as before. Absence of results is not absence of a Benchmark.

**Why the config file is the marker.** It is the file the evaluation Skills require (`agent-evaluation/SKILL.md`: "Require `agent_state/system_config.yaml`, `benchmark_config.toml`, …"), and it is the only source of the title and the per-case run count — a directory without one has nothing to display and nothing to run. It is the same criterion `listInstalledSkills` already applies to `skills/<name>/SKILL.md` ("a directory with no SKILL.md doesn't count as a Skill"). Only a missing config skips: the check distinguishes `ENOENT` and rethrows any other error rather than swallowing it.

**The race is untouched, deliberately.** This is a mitigation, not a fix. `remove()` still deletes a Benchmark that a Session is running against, and the half-written directory is still created on disk — it just no longer appears in the Evaluation Center. Refusing deletion while a run is in flight needs the Session→Benchmark association this PR's own "Not done" note calls out as missing, and the decision was explicitly not to improvise one here. The window remains open.

**No log line.** A skip would be logged on every Evaluation Center mount, for every agent, for as long as the directory sits on disk, and `BenchmarkService` has no logger — none of `packages/server/src/services/*` logs, so it would mean threading `deps.log` in for that. Documentation carries it instead: `/docs/self-improvement` § "Benchmark storage" and the `GET …/benchmarks` row in the server API reference now say a directory without a config is not listed, where it comes from, and that it is safe to delete by hand. Both languages.

Also in this commit: a bullet on the existing `changelog/unreleased/2026-09-02-evaluation-center{,.zh}.md` entry rather than a new entry, since this refines the same unreleased feature in the same PR.

Verified: `pnpm typecheck` (clean); `packages/server` `vitest run test/benchmarks.test.ts test/builtin-agents.test.ts` (12 passed), and the full `packages/server` suite incidentally (114 files / 1650 tests, all passed); `packages/docs` content/frontmatter/search tests (33 passed); `pnpm format` + `pnpm format:check` (clean). The new test was confirmed non-vacuous — with the service change stashed it fails, listing `half-deleted` alongside `never-run`.

hiyouga:
Reviewed at `228559cb0`, in four slices: backend (server/core/cli/DB), the two big list surfaces, the detail views and modals, and the app shell / lib / tests / docs.

Good news first, because it is the part most likely to be wrong in a diff this size and it is not: **the two dictionaries are in lockstep.** I diffed flattened key sets by parsing both files with the repo's own TypeScript rather than by eye — 153 keys added to each side, 3 removed from each (`selectBenchmark`, `emptyAgent`, `legendUnlabeled`, all with zero remaining references), no key missing from either side, no existing value changed, no CJK in the added `en` entries, every function pair with identical parameters, every `${…}` placeholder present on both sides. Also clean: the sidebar contract is honoured (every group visible in every mode, the Evaluations folder rendered from `FOLDER_CATEGORIES.map` at both call sites so it appears in all three modes without a fourth branch, folder rows still paging ten at a time), `sessionCategory` stays total and single-valued, the new `install`-scoped key family is enforced by `install-scope.test.ts` rather than trusted, and all four new modals go through the shared `Modal` — Esc layering, focus trap and stacking are inherited and correct, including both stacking pairs.

---

## The thing that needs a decision before anything else

`packages/core/src/state/paths.ts:154` moves `benchmarksDir` from `<project>/agents/<agent>/benchmarks` to `<project>/benchmarks` and drops the `agentId` parameter. I checked: nothing reads the old location, nothing migrates, nothing deletes. Per `CLAUDE.md`'s 向后兼容 clause that is not a call the implementation gets to make on its own, and the entry's single line — "A `benchmarks/` directory left under an agent by an earlier version is not read by this version; nothing migrates or deletes it" — understates it in two ways:

**1. The upgraded install does not look empty. It looks populated with fake data.** `agent-state.ts:186-197` now calls `provisionExampleBenchmark` on the *load* path — the one behind Session creation, compaction and resume — whenever `agentId === DEFAULT_AGENT_ID` and `<project>/benchmarks/` does not exist. An install with `agents/default_agent/benchmarks/{example-benchmark, my-bench}` has no Project-level directory, so the first message after the upgrade writes a fresh example Benchmark with three fabricated 2026-07 evaluations. The Evaluation Center then shows exactly one Benchmark — the sample — while the user's real cases and score history sit unreachable in the old directory. That is the worst shape this break could take.

**2. The installed Skill still points at the old path.** `plugins/agent-tuning` went `2026.09.02.1 → 2026.09.14.3` and its SKILL.md changed `BENCHMARK_DIR` to the Project-level path, but installed Skills are file copies and are never auto-upgraded — `installPlugin` runs only from `POST …/plugins` and from agent creation; `agent-service.ts:185-205` merely computes an "update available" hint on the card. An upgraded user who clicks Use → Evaluate without noticing that hint hands the evaluator a prompt naming `benchmarks/<id>` while its own Skill names `agents/<agent>/benchmarks/<id>`. Either the run dies on a missing `benchmark_config.toml`, or — on a legacy root where the old directory still exists — it evaluates the stale cases. Either way the user pays for a full Case × Run matrix and the chart does not move. The dialogs only warn when the Skill is *absent*, never when it is stale.

Also: `changelog/README.md` requires `Breaking: yes — <one line>` plus a `## Compatibility` section stating what breaks and the migration step. The entry has neither. The sibling `2026-08-22-sandbox.md` does exactly this for the same class of change.

My recommendation if the break is accepted: name it in the entry with the recovery (`mv <root>/<project>/agents/<agent>/benchmarks/* <root>/<project>/benchmarks/`) and the "update `agent-tuning` on every agent that has it" instruction, and suppress the example when `agents/*/benchmarks/` exists so an upgraded install does not get a sample planted on top of a silently orphaned one.

A related design question, separable from the above: `loadAgentState` now writes eight files on a path whose own docstring says it "must never re-create a deleted Agent as a side effect". Deleting `<project>/benchmarks/` to be rid of the sample re-creates it on the next message. Deleting *through the UI* leaves the directory behind so the example stays gone — but that rule is invisible from the filesystem, and `rm -rf benchmarks` is the obvious way to mean "no thanks". A marker file (`benchmarks/.provisioned`) would state the intent instead of inferring it from directory existence.

---

## Correctness

**1. The Optimize baseline is not label-scoped, so the optimizer can be handed a bar it already clears.** `benchmark-metrics.ts:190-196` — `latestScoreOfAgent` filters only by `agentId`, then takes the array-last record whatever model and thinking level it ran on. Score `report-writer` at 85 on `deepseek-v4-pro/xhigh`, then run a cheap sanity check at 42 on `deepseek-v4-lite/low`; Use → Optimize prints "baseline 42 → target 52" and ships `desired_score: >=52` — which the agent already beats by 33 on the runtime the optimizer will actually reuse. The loop exits at round 1 having changed nothing. `latestWithDelta`'s own doc states the rule this path skips: "Only same-label scores are comparable — another agent, another model or another thinking level is a different measurement." Not caught by the tests because every fixture gives each agent one label. Fix: narrow to the newest record's label after filtering by agent, or return the label so `baselineLine` can name the runtime it came from.

**2. "Create manually" is offered to project members, and the server refuses them.** `benchmark-page.tsx:397` and `:442` — `POST …/benchmarks` is `requireProjectOwner` (`routes/benchmarks.ts:145`). `isOwner` is computed at `:284` and used only for `canDelete`. A member fills in id, title and cases, submits, and gets a permission error after all the typing. The repo already has the pattern — `schedule-panel.tsx:252` and `schedules-tab.tsx:168` both do `{...(isOwner ? { onManual: … } : {})}`, and `create-buttons.tsx`'s own docblock names this exact case.

**3. A create that fails after the dialog is dismissed reports nothing.** `create-benchmark-modal.tsx:109-141` — Cancel is `disabled={busy}`, but `Modal` closes on Esc and overlay mousedown unconditionally and the component returns `null` the instant `open` goes false, so `setSubmitError` lands on a dead component. Press Create, press Esc inside the request window, server 500s: no toast, no error, no benchmark — indistinguishable from a create that succeeded but has not refreshed. The success path already uses `toastSuccess`; `toastError` exists.

**4. Every preview failure tells the user to retry something that will never succeed.** `benchmark-case-browser.tsx:228`+`:238` — `throw new Error(String(response.status))` is caught and passed to `apiErrorText`, which returns `S.common.unknownError` for anything that is not an `ApiError`. A 404 on a file a Markdown link points at renders "Request failed, please try again later". `workspace-browser.tsx:887` does not make this mistake.

**5. Field errors never clear while typing.** `create-benchmark-modal.tsx:166-181` — `errors` is rewritten only by `validate()` on the next submit, so a blank-title error (and the 409 `idExists`) stays red under a field the user has already fixed. `vault-tab.tsx:65` solves this one file over with `clearAddErrors` called from every `onChange`.

**6. The previewed Markdown remounts its links and images on every render.** `benchmark-case-browser.tsx:286` — `FileBrowser` memoizes `markdownComponents` on `[resolveRef, onOpenFile]` with the comment "a fresh map every render would remount the whole body", but `openAndReveal` is a plain arrow in the component body, so its identity changes every render, `components.img`/`components.a` become new element *types*, and React tears down and rebuilds every `` and `<a>`. Read a case README, tab to a link in it, let an in-flight listing land — focus drops to `<body>`. Expanding a folder re-decodes every image. One fix in `file-browser.tsx` covers both callers: hold the two callbacks in refs and build the map with `useMemo(…, [])`, which is what the comment already claims.

**7. One drag freezes the folder-only demotion into persisted order.** `sidebar.tsx:657-661` — `agentGroupSequence`/`workspaceGroupSequence` map over the *post-demotion* order, and `onDrop` commits that full sequence. Drag any group once and every folder-only group is written to the tail of `penguin.groupOrder.<projectId>.<mode>`; when one later gains a real conversation it is no longer demoted but its stored place is the tail, so it stays at the bottom. `group-order.ts:136-137` asserts the opposite ("a group that gains an active row simply rejoins the others where it was all along") and `group-order.test.ts:427` pins a pre-demotion array production can no longer produce. Committing a pre-demotion sequence would desynchronise the splice indices from what is on screen, so the honest fix is probably to correct the docstring.

**8. A group that stops being folder-only falls back to a stale collapsed entry.** `sidebar.tsx:2191-2196` (and the agent twin at `:2102-2107`) — `collapsed` reads `collapsedGroups` when `foldedOnly === undefined` and `expandedFolderOnlyGroups` otherwise, and neither set is rewritten on transition. Collapse group G long ago; G later empties and becomes folder-only; open it and start a chat in that Workspace; the new active row flips `isFolderOnly` false, classification returns to `collapsedGroups`, and G collapses on the spot — hiding the conversation just created. Deleting the key from `collapsedGroups` when `toggleGroup` opens a folder-only group closes it.

**9. A benchmark draft's `source` outlives the prefill it came with.** `draft-cache.ts:54-59` — `aiPrefillRef` is dropped by the first keystroke and on send, but `cached.source` is read from the mount-time snapshot on every persist and never dropped by an edit. Open Use → Evaluate, edit the prompt, navigate away, come back, delete everything, type an unrelated question, Send: `draft-view.tsx:689-690` sets `body.source = "benchmark"` and the sidebar files that conversation under Evaluations. Survives reload, nothing shows or clears the mark. Either drop `source` alongside `aiPrefill` in `onTextChange`, or reword the comment to say it deliberately outlives it.

**10. Smaller ones.** `benchmark-service.ts:247-252` rethrows any non-ENOENT `fs.access` failure, so one unreadable benchmark directory turns the whole Evaluation Center into a 500 where the file header promises graceful degradation. `:311-319`'s rollback lets a failing `fs.rm` replace the real error, so the client is told about the cleanup rather than the write (`await fs.rm(...).catch(() => {})` before the rethrow). `benchmark-case-browser.tsx:186` shares one `listError` slot across materials and every subdirectory, so one directory's success clears another's error. `:262`'s README auto-preview can overwrite a file the user opened first when the `statement` listing is slow — setting `readmeOpened.current = true` at the top of `openAndReveal` closes it. `benchmark-page.tsx:459` prints the raw agent id in the filter chip while computing `nameOf(filterAgentId)` on the two adjacent lines for the avatar and tooltip. `use-benchmark-modal.tsx:117-127` fetches the default agent's skills twice on open and three times for two distinct agents, because the in-flight bookkeeping is the same state the effect depends on.

---

## Simplification, and guards against states the server cannot produce

`benchmark-service.ts` already validates and *drops* malformed records before they reach the client — `scoreOr` requires a finite number in `[0,100]` or the evaluation goes, `sessionId` must be non-empty or the whole case goes, `version` an integer ≥1. Against that contract:

- `benchmark-metrics.ts:171` (new here) — `typeof e.score === "number" && Number.isFinite(e.score)`: `Number.isFinite` alone subsumes the `typeof`, and neither can fire.
- `evaluation-detail-modal.tsx:92, 108` guard `durationMs` on a required `number`, while `:230` calls `humanizeDuration(evaluation.durationMs)` unguarded — the file disagrees with itself. Same for `version` at `benchmark-detail.tsx:91, 222, 240` vs `evaluation-detail-modal.tsx:221`.
- `evaluation-detail-modal.tsx:29-35, 111, 161` stacks three guards on one `sessionId`: an optional prop, a conditional spread at the call site, and a `.filter(id => id !== "")`.

The cost is not just dead code — `scoreValues`'s `(number | null)[]` drags null-handling through `seriesValues`, `lineSegments`, `sparklineSeries` and two display sites. And it cuts the other way: if the premise really held, `formatScore(evaluation.score)` at `benchmark-detail.tsx:234` is unguarded and would throw. Pick one side.

Also worth collapsing:

- `create-benchmark-modal.tsx:58-61` and `use-benchmark-modal.tsx:50-52` carry byte-identical `digits` and `errorProp`. `errorProp`'s premise is false here: `tsconfig.base.json` sets `strict` but not `exactOptionalPropertyTypes`, and `Input`/`Textarea` consume `error` as `Boolean(error)`, so `error={undefined}` is indistinguishable from omitting it. The comment "an explicit undefined is not an absent prop" does not hold in this codebase — delete both copies and write `error={errors.title}`.
- The `[1, MAX_RUNS]` bound is spelled twice, once as a boolean (`isValidRuns`) and once as a parser (`intIn`, which also has a `Number.isSafeInteger` check the other lacks). One parser returning `number | null` does both.
- `use-benchmark-modal.tsx:99-101, 129-138` solves the late-default problem with state plus a sync effect, where `ai-create-modal.tsx:59-62` — same PR, same author — derives it during render. Two effects and two renders disappear by adopting the version that already exists.
- `benchmark-case-browser.tsx:244-255`'s "different case" reset effect can never fire: the only mount site derives `openCase` from `openCaseId`, and the only effect that changes `bm.id`/`projectId` nulls `openCaseId` first, so the component always unmounts instead. On mount it re-sets six state slots to values they already hold with fresh `Map`/`Set` identities, costing an extra render. If it ever *did* fire it would be incomplete anyway — `FileTree`'s internal `seenSerial` is not reset, and the caller restarts its toggle serial at 1, so a case switched in place after exactly one toggle would silently skip the next collapse animation. Delete it; add `key={openCase.id}` at `benchmark-detail.tsx:448` if the invariant is worth stating.
- `benchmark-detail.tsx:337-342` resets five state slots the parent has already replaced with a Skeleton, and `:359` copies a mutable array nothing mutates, with a comment guarding a sort that exists nowhere in the file.
- `session-grouping.ts:234-241`'s `foldedShare` NaN guard is dead at its only production call site: `groupShares` hands it a locally built object whose five keys are initialised to `0` and `Math.max`-ed. The justification it borrows from `aggregateWorkspaceCounts` (which does consume raw server JSON) does not transfer.
- `use-benchmark-modal.tsx:86-90` recomputes `[...benchmark.evaluations].reverse().find(...)` on every keystroke to seed state read once — `useState(() => …)`.
- Dead parameters: `initialTab` has one value at both call sites; `ask-ai-modal.tsx`'s `open` is always `true` (both callers gate the mount); `md.tsx:99-101`'s `SETTLED_MD_COMPONENTS` is an alias of `SETTLED_COMPONENTS`.
- The material identity is spelled out in five places in `benchmark-case-browser.tsx` (`MATERIAL_PATHS`, `splitTreePath`'s `head === "rubric"`, the `materials` array, `headerPath`, `rowTrailing`); one `{material, label}` table carries all five.

On the four modals: **far less should collapse than ~1000 lines suggests.** `ask-ai-modal.tsx` is already the right shape — a thin adapter over `AiCreateModal`. `evaluation-detail-modal.tsx` is read-only and shares nothing with the forms. The real shared surface is ~30 lines between the create and use dialogs (the items above), not a new abstraction. What has genuinely drifted between them: two different mount strategies (internal `open` gate vs external `key` + `open`), only one of them can fail and it is the one that loses its error on dismissal, the use dialog handles an empty *model* list but not an empty *agent* list (reachable — a Benchmark outlives the agents it scored, and `AiCreatePanel:120-122` already has the line), and `evaluation-detail-modal` guards required fields in two places and trusts them in a third.

`file-browser.tsx` earns its generality, for the record — two real callers shipping today, and the plugin-detail refactor onto it is a net deletion that removed a duplicate `stripFrontmatter` and an extension→language table that had drifted from `code-languages.ts`. Its weakest part is the four sizing props that exist only so the case dialog can be a few vh taller than the plugin modal.

---

## Tests

- **`benchmark-prompts.test.ts` never exercises the English dictionary.** `S` is a live binding initialised to `zh` and the test never calls `setActiveStrings`, so all ~30 assertions run against the Chinese tails only — even though the regexes are written bilingually (`/runs[：:] ?\`2\`/`), so both were clearly intended. Drop `${p.version}` from `strings-en.ts`'s `evaluateTail`, or leak `agent-optimization` where line 206 asserts its absence, and it ships silently; `Strings` cannot catch it because both sides are `(p) => string`. `memory-chat-prompts.test.ts:8-15` has the pattern: loop the describes over `[zh, en]`.
- **The tested composition helpers are production-dead.** `use-benchmark-modal.tsx:162-163` inlines `composeAiPrompt(note, evaluateTail(...))` rather than calling `buildEvaluatePrompt`; the only callers of both helpers are the test. So the test proves a composition the app never performs and the app's actual composition is untested. Calling the helpers from the modal — one line each — makes the existing test mean something.
- **`groupShares` is where folder-only classification is actually decided, and nothing tests it.** The new tests cover `foldedShare` and `isFolderOnly` (two near-trivial predicates) and `orderGroups`'s `demote` with a synthetic key prefix. Drop the `Math.max` against `totals.active` and a group whose active rows simply have not been paged in yet would fold up and sink to the bottom — no test would notice. Exporting it from `session-grouping.ts`, where its two collaborators already live, plus one case ("server says active: 30, nothing loaded → not folder-only") closes it.
- Worth pinning cheaply: `resolveRelative`'s `..` clamp at the material root is the traversal guard and is only correct because `[].pop()` is a no-op (`resolveRelative("a/b", "../../../x") === "x"`); `splitTreePath`; `previewKindOf`'s two documented mappings, which are exported and have no test at all. No test covers `create`'s rollback, `remove` against a symlink, `runs` defaulting to 1, or the CLI's new `--source`.

The rest of the suite is good and I checked it rather than assuming: `benchmarks.test.ts`'s status table distinguishes `draft`/`failed`/missing/unknown, the config-less directory case encodes the real "deleted mid-evaluation" scenario, the create test reads bytes back off disk, the 409 test verifies the first benchmark survived, and `benchmark-case-tree.test.ts` and the rewritten `benchmark-metrics.test.ts` would each fail if the behaviour they name were broken.

---

## Copy and docs

- `strings-en.ts:2951` vs `:3057` — `benchmark.targetAgent` and `benchmark.testedAgent` are both 被测智能体 in zh but "Test Agent" and "Tested agent" in en, and both are `label=` on an agent picker in the same feature. "Test Agent" reads as the agent *doing* the testing, while `agentColumn`, `testedAgents` and the docs all use the other form. Set `targetAgent: "Tested agent"`; leave the prompt tails alone, where "Test Agent" matches the Skills' own `test_agent_id`.
- `use-benchmark-modal.tsx:148, 333` — Round limit is bounded by `MAX_RUNS` (the runs-per-case cap) and renders `S.benchmark.invalidRuns` — "Must be an integer from 1 to 1000" — under a field whose hint never mentions a range. It is the one field of three reusing another's message; the neighbouring Target score has its own.
- `strings.ts:2855` — `askEvaluationTail` backticks every structured value but interpolates the free-text `p.summary` bare into a bullet, so a multi-line model-written summary breaks out of its bullet and reads as top-level prompt text. Mitigated by the user seeing the assembled prompt in `PromptFold` before sending, so this is low, but indenting continuation lines would close it.
- `sidebar.tsx` lines 392, 1563 and 1749 carry Chinese inside English comments (智能体 → 评估中心). Pre-existing on `main`, but `CLAUDE.md` says to translate such residue when editing nearby, and this PR edits within a few lines of 392.
- `benchmark-detail-page.tsx:47-67` calls `listBenchmarks(projectId)` and picks one out, so opening one Benchmark downloads every Benchmark's complete scoreboard — every case, every run, every session id — which `benchmark-page.tsx` had just fetched. The comment names the cause honestly. Worth a follow-up `GET /benchmarks/:id` rather than a change here.
- Docs check out otherwise: five of six pairs match exactly on headings, bullets and table rows (the `web-app` bullet-count difference is pre-existing on `main`), the `status` claim matches `benchmark-service.ts:460` verbatim in both languages, and the architecture and sessions-and-traces trees now place `benchmarks/` beside `agents/`, matching the new `benchmarksDir`.

PR #611, opened 2026-09-04
feat(machines): run an agent — and its Sessions, Traces and terminal — on a machine, over the held connection

Myriad-Dreamin:
> **Stack:** `main` ← #448 ← **this**. Review the diff against `feat/machines-identity`. Re-derives #450 over the held-connection design that landed with #574–#576; #450 should be closed in favour of this.

Pick a workspace that lives on another machine and the conversation runs there: that machine's server drives the agent, holds the messages, writes the trace, serves the terminal. The window never moves — it stays on this server and names the machine on the calls that concern it.

## What is different from #450, and why

#450 was written before the stack: its server half was the pre-stack machines subsystem, and its page carried a connect-on-first-need loop — a widening schedule on the page that raised a forward whenever a call about a machine found none. That contradicts the rule the stack settled on in #574: **a connection someone asked for is held by the server**, never idles out, is re-established by the transport when it drops, and is restored after a restart or a hot push. So the page no longer connects anything. Reachability is the server's word (`MachineInfo.connection`), read from the machine list and re-read on a slow tick while any machine is out of reach; `raiseForward`, `setMachineConnector`, the `not_connected` retry in the fetch wrapper, `meOnMachine` probing and the autoconnect module are gone. Everything else in #450 that was about *Sessions living elsewhere* stands, and is re-implemented here on the current base.

## A Session's calls follow the machine that owns it

A Session lives on the server whose filesystem its workspace is on, so **every** call about it has to reach that machine — two dozen endpoints. Routing is a rule over the **path**: a request to `/api/sessions/<id>/…` goes wherever `<id>` was last seen, and the output stream with it (`lib/session-machines.ts`). Call sites stay unchanged; a Session's machine is recorded when a list hands one back and when one is created. An explicit `server` still wins for the calls that are not Session-scoped and know their target.

Two shapes escape a rule over the path and carry the machine by hand: the Trace endpoints, which bury a Session id in an Agent-level path, and bare URLs — a Workspace file's content, an attachment's scratchpad image, a Trace download — which never pass through the fetch wrapper. `trace-routing.test.ts` scans `endpoints.ts` so the next endpoint of the first shape is caught as it is written.

The mapping is in memory on purpose: rebuilt by the very list that displays the Sessions. Two caches that DO survive a reload sit beside it, per machine: the Sessions and Agents each was last seen holding, shown while the server is still re-holding its connection to that machine after a restart or a push, replaced wholesale by the machine's next answer (`lib/machine-cache.ts`).

## One list, several machines

The sidebar is every connected machine's Sessions, merged. Each server pages its own rows with its own cursor, so page state is keyed per source; counts are summed across the sources that answered; the merged list is ordered by a total order, so equal timestamps do not reshuffle between refreshes.

The line the list rests on: **a source that could not answer is not a source that answered nothing.** Only a 404 — this server has not got that Agent — is an answer. This server failing mid-swap abandons the reload and leaves the rows standing; a machine going quiet keeps its cached rows on screen and its cache intact; a machine with no connection contributes what it last held. `loading` is raised only when there is nothing on screen, because the chat page draws a skeleton in place of the open conversation while it is true. The chat page holds the routed Session through a refetch that momentarily stops naming it, and a failed direct lookup settles nothing while the machine that could hold the Session is out of reach.

## The list stays true without a reload

A Session created anywhere — the CLI, another tab, a schedule, an agent's child — is announced on the user channel (`session_created`, to the Project's owner and members) and the list fetches the row; a title set through `PATCH` is announced as `session_title`. Each connected machine's own event stream is opened through the proxy, since a Session there changes state on that machine's server and nothing else knows. A machine's `web_updated` is ignored: that is its web, not this window's.

## The composer follows the machine

The draft carries the workspace's machine with the path — in state, route state and the saved draft — offers that machine's Agents (last seen ones first, while it is asked), creates the Session on that machine, and re-records the routing under the id `postTask` hands back when the Session self-healed. Starting a chat from a Workspace group keeps the group's machine. A machine that cannot be asked for its Agents says so in the row and points at the Machines page; nothing here connects.

## Terminals

A shell opens on the machine its Workspace is on, including the fall-back to home, which is home *there*. The stream needs no runtime change — the seam it uses (`terminals()` / `attachStream`) is already on `main`: the platform answers `get` for `<terminalId>@<machineId>@<userId>` with a reference to the remote pty and relays the socket to that machine's own stream **through the held connection's SOCKS port**, like every other request to a machine's API (`machines/terminal-relay.ts`). The user id is what the runtime's owner check reads, so a client can only name itself; the relay then checks the user is an admin (`AuthService.isAdmin`, new).

The terminal list asks every connected machine, prunes a conversation's stored tabs only once every source answered and the machine set is published, adopts a shell already running on the conversation's own machine before starting another, and a detached pane carries its machine in the URL.

## Verification

`build`, `typecheck` (0 errors), `lint`, `format:check` clean. Server **123 files / 1748 tests**, web **128 / 1704**, cli **30 / 391**, docs **53** — all passing. New: `session-created-events` (3), `terminal-relay` (3) on the server; `session-machines` (6), `session-merge` (3), `machine-cache` (5), `terminal-machines` (5), `trace-routing` (3), `route-session-hold` (5), `session-aggregation` (5) on the web — the store's multi-source reload is exercised directly against a mocked API, including the three "could not answer" cases.


Myriad-Dreamin:
Opened by mistake as a separate PR; the same commits now live on #450's branch (feat/machines-run-agent), which carries this work.

PR #612, opened 2026-09-05
feat(core): add x-opencode-session header support

paskaa:
## Summary
OpenCode Go gateway requires a stable `x-opencode-session` header per conversation to optimize backend routing.

- Add optional `sessionId` to `GenerativeModelConfig` and thread it from Session spec
- `attributionHeaders` now returns `x-opencode-session` for `opencode.ai` and `freqtrade.1unlock.top`
- Forward `sessionId` from `buildRuntime` (bootstrap, openNextContext, vision describer, bare LLM)

## Test plan
- [ ] New session against `opencode.ai/zen/go` logs show `x-opencode-session: <sessionId>`
- [ ] Session resume reuses same header value
- [ ] Existing OpenRouter/TokenDance attribution headers unchanged

### DIFF
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -983,6 +983,7 @@ export class Agent {
               ...(visionEntry.context_window !== undefined
                 ? { contextWindow: visionEntry.context_window }
                 : {}),
+              ...(sessionId !== undefined ? { sessionId } : {}),
               requestTimeoutMs: 60_000,
             }),
         };
@@ -1057,6 +1058,7 @@ export class Agent {
         // model that rejects it.
         ...(modelEntry.fast_mode === true ? { fastMode: true } : {}),
         ...(context.thinkingLevel !== undefined ? { thinkingLevel: context.thinkingLevel } : {}),
+        ...(sessionId !== undefined ? { sessionId } : {}),
         ...(context.requestTimeoutMs !== undefined
           ? { requestTimeoutMs: context.requestTimeoutMs }
           : {}),
@@ -1126,6 +1128,7 @@ export class Agent {
         ...(modelEntry.context_window !== undefined
           ? { contextWindow: modelEntry.context_window }
           : {}),
+        ...(sessionId !== undefined ? { sessionId } : {}),
         requestTimeoutMs: 30_000,
       });
 
--- a/packages/core/src/engine/context-engine.ts
+++ b/packages/core/src/engine/context-engine.ts
@@ -309,6 +309,30 @@ export interface ContextEngineDeps {
 const isImageMessage = (m: OmniMessage): boolean =>
   (m.payload as { type?: string }).type === "image_url";
 
+/**
+ * Simple non-cyclic hash for a string — used to compare tool-call argument JSON.
+ * Not cryptographic; just needs to be stable and fast for short strings.
+ */
+function simpleHash(str: string): string {
+  let h = 0;
+  for (let i = 0; i < str.length; i++) {
+    h = (h * 31 + str.charCodeAt(i)) | 0;
+  }
+  return h.toString(36);
+}
+
+/**
+ * Tool-call loop detection: when the model repeatedly issues identical tool calls across
+ * turns, the ReAct loop runs forever (maxTurns defaults to -1). We track a rolling window
+ * of per-turn "signatures" (sorted name+args digests). When the last REPEAT_WINDOW
+ * signatures are identical AND match the one immediately before them, we fire a tool_loop
+ * cutoff. A window of 3 means 3+1 = 4 consecutive identical turns triggers the cutoff —
+ * enough to avoid false positives on legitimate retries but tight enough to stop a stuck
+ * model within a few seconds.
+ */
+const TOOL_LOOP_WINDOW = 8;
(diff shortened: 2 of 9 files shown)


hiyouga:
Thanks — I'd like to take the `x-opencode-session` idea, but this branch has to be cut down to that one thing first. About 18 of the 315 lines here are the header; the rest is unrelated work that needs its own PR or is already on `main`.

## Keep

The feature itself, four files:

- `interfaces/llm.ts` — `sessionId?: string` on `GenerativeModelConfig`
- `agent.ts` — thread `spec.sessionId` at the three construction sites
- `llm/generative-model.ts` — pass it to `attributionHeaders`
- `state/model-catalog.ts` — emit `x-opencode-session` for `opencode.ai`

## Drop

**1. The tool-loop detector** — `engine/context-engine.ts` (+101), `test/engine.test.ts` (+103), and the `tool_loop` kind in `interfaces/shared.ts`.

Three independent reasons, any one of which stands on its own:

- It has nothing to do with a request header. A reviewer who wants to reason about one cannot review the other.
- `context_engine` is deliberately kept feature-free — it moves OmniMessages between the Human / LLM / Environment interfaces and decides the next action from `stop_reason`, and that is all. Policy that vetoes or cuts off a run is layered outside it (the precedent is the command policy, which decorates the injected `ApproveFn` at the `Session.run` boundary rather than teaching the engine a rule).
- `RunCutoff.kind` is pinned to exactly four values in the design docs (`specs/05-ARCHITECTURE.md`). A fifth is a design change, and in this repo the design change lands first, in the design repo, before the code.

The underlying problem is real — an unbounded ReAct loop with `max_turns: -1` is worth solving. Please open it as its own PR starting from the design side. Worth thinking through there: four identical turns is not always a stuck model (a poll loop, or re-running the same command while waiting on something, both look exactly like this), so the false-positive cost needs an answer before the cutoff is worth having.

**2. The `exec_command` `command` alias** — `environment/tools/exec-command.ts` and its cases in `environment.test.ts`. This already shipped in v0.2.11 via #661, which reads the alias through one shared list so the command policy screens the same text the tool actually runs. The version here reads the alias inside the tool only, so a `command`-keyed call would have run without being screened. Same note as on #620.

**3. `freqtrade.1unlock.top`** — a third-party mirror does not belong in the built-in catalog. We hard-code hosts we operate or have an attribution agreement with. `opencode.ai` only.

## On the header itself

`sessionId || "penguin-harness"` should be "no header at all when there is no session id". A constant makes every conversation look like one session to the gateway, which is precisely the opposite of what a stable per-conversation id is for — that is worse than sending nothing. Return `undefined` and let the request go out without the header.

## Before it merges

The branch conflicts with `main` and needs a rebase, and per `CLAUDE.md` this wants a `changelog/unreleased/` entry as an English/Chinese pair.

Once it's just the four files above, it's an easy merge.

hiyouga:
I've trimmed this down to the session header myself and pushed it to your branch (`983ca90a7`, a fast-forward — nothing of yours was rewritten). It is now `MERGEABLE`, +99/−2 across 7 files, and CI is running.

**What survived** — the feature, rebuilt against current `main`:

- `interfaces/llm.ts` — `sessionId?: string` on `GenerativeModelConfig`
- `agent.ts` — threaded at the three construction sites
- `llm/generative-model.ts` — passed to `attributionHeaders`
- `state/model-catalog.ts` — the `opencode.ai` branch

**What came out**, for the reasons in my earlier comment: the tool-loop detector and its `tool_loop` `RunCutoff` kind, the `exec_command` `command` alias (already on `main` via #661), and `freqtrade.1unlock.top`. Those five files are now byte-identical to `main` — I checked blob hashes both ways, not just an empty diff.

**Three changes to your patch you should know about:**

1. **No constant fallback.** The branch now reads `if (hostMatches(host, "opencode.ai") && sessionId) return { "x-opencode-session": sessionId };` and otherwise falls through to `return undefined`. `sessionId || "penguin-harness"` would have made every conversation look like one session to the gateway, which is the opposite of what a stable per-conversation id is for.

2. **`agent.ts` passes `sessionId` plainly**, not behind `...(sessionId !== undefined ? { sessionId } : {})`. `SessionSpec.sessionId` is a required `string` (`agent.ts:228`, destructured at `:917`), so that condition can never be false. The neighbouring conditional spreads are there because those fields genuinely are optional.

3. **Added a test and a changelog pair.** `packages/core/test/model-catalog.test.ts` pins the header with a session id, no header without one, no header for the removed mirror, and that OpenRouter and TokenDance are unaffected by a session id. The changelog entry is `changelog/unreleased/2026-09-14-opencode-session-header.{md,zh.md}` (`Type: feature`, per `changelog/README.md`'s enumerated values).

Verified on our test box against current `main`: core 1179 passed / 5 skipped, typecheck clean, lint clean, `format:check` clean. No spec change is owed — `attributionHeaders` and its header list are not described anywhere in `specs/`.

Thanks for the report and the original patch; the header itself was a good catch. If you want to pursue the tool-loop detection, please open it separately and start on the design side — the unbounded ReAct loop under `max_turns: -1` is a real problem, it just needs the `RunCutoff` contract changed first.

hiyouga (approved):

PR #613, opened 2026-09-05
feat(web): drive a terminal using fingers

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← #808 ← #718 ← #544 ← #448 ← #450 ← #577 ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← **#613** ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

The touch terminal, on its own. This PR used to carry the phone dashboard too; that half is now #815, directly below, and #613 took it in by a merge commit — so the diff here is the terminal alone.

## What

A soft keyboard has no Esc, no Tab, no arrows and no Ctrl, which left the terminal readable on a phone and not usable on one. A key bar mounted inside `TerminalView` gives both hosts **Esc, Tab, Ctrl, arrows, ^C, paste, a keyboard toggle** and **Alt last**. Two things make it behave like a keyboard rather than like buttons: a press never moves focus (each cap cancels its own `mousedown`, which is the event that does it for a tap as much as for a click), and Ctrl/Alt are sticky one-shot, because a phone cannot hold a modifier down. The caps compose their own byte sequences, so they honour what a real keypress would have consulted — the terminal's cursor-key mode and xterm's modifier parameter.

Everything is behind `(pointer: coarse)`: a device with a real keyboard renders exactly what it did before.

## Also here: a finger scrolls a program that has taken over the terminal (from #645)

Folded in from #645 (rebased onto this branch, squash-merged as e6d9333f7). xterm turns its own touch scrolling off as soon as a program takes the mouse — exactly the programs that need it, like Claude Code on the alternate screen, which moves its transcript only when a wheel reports. The drag is now that wheel: `terminal-touch.ts` turns travel into lines (slop threshold, carried remainder), and the view dispatches a `WheelEvent` per line through xterm's own element, so the report carries whatever mouse protocol the program asked for. A tap stays a tap; where the mouse is free, xterm's own touch scrolling still scrolls the scrollback; the host carries `touch-action: none` on touch devices.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` and the full `pnpm -r test` green on the test machine (server 2179, web 2462) at f5d04bfc0, and again with #645 folded in (web 2470, `terminal-touch.test.ts` included); the split added only a merge and changelog text on top, no code. `terminal-keys.test.ts` pins the wire format: arrow sequences in both cursor-key modes with xterm's modifier parameter, control codes, and that an armed Ctrl or Alt leaves a multi-character chunk (a paste, an IME commit) alone. `e2e/terminal-touch.spec.mjs` drives the key bar under a coarse pointer.


Myriad-Dreamin (commented):
Review of this branch (the diff includes the merged-in #618 dashboard work). Nine findings, inline below.

The two worth acting on first: the dashboard counts subagent / scheduled / evaluation Sessions as human work, so the board contradicts the sidebar that hides them; and the machine filter tests `selectable` (ssh listing) where API reachability needs `connection`, so disconnected machines are polled every 15 s and then reported silent.

packages/web/src/features/dashboard/dashboard-view.ts
```diff
@@ -0,0 +1,99 @@
...
+export function dashboardRows(
+  sources: readonly DashboardSource[],
+  seen: SessionSeenState,
+): DashboardRow[] {
+  const rows: DashboardRow[] = [];
+  for (const source of sources) {
+    const machineLabel = source.local ? null : source.machineLabel;
+    const prefix = `${source.machineId ?? ""}\0`;
+    const byKey = new Map<string, DashboardRow>();
+    for (const s of source.sessions) {
```
Myriad-Dreamin, line 115: **The dashboard counts subagent, scheduled and benchmark Sessions as human work.**

`SessionActivityInfo` carries no `source` (the server deliberately does "no source classification", `session-service.ts:233`), so `dashboardRows` counts every non-archived row. Subagent rows are registered in `sessions` with `has_trace = 1` and keep `lastActiveAt = createdAt` (`repos/sessions.ts:47-55`), and nobody ever opens one to stamp a seen marker — so an Agent that spawned 5 subagents in `/home/u/proj` reads "6 running" while one conversation runs, and "5 to review" forever afterwards. An evaluation run adds one Test Session per Case × Run into the merged "Temporary workspaces" row the same way.

The sidebar hides all three kinds in collapsed folders (`sessionCategory`, `session-grouping.ts:231`), so the board and the list disagree — exactly what the PR body says it should not do ("subagents are left out").

packages/web/src/features/dashboard/dashboard-page.tsx
```diff
@@ -0,0 +1,178 @@
...
+    [sources, seen],
+  );
+  const [silent, setSilent] = useState(0);
+  const [error, setError] = useState<string | null>(null);
+
+  const load = useCallback(async () => {
+    if (projectId === null) return;
+    let servers: Server[] = [THIS_SERVER];
+    try {
+      const machines = workspaceMachines(await api.getMachines(projectId)).filter(
```
Myriad-Dreamin: **The machine filter tests the wrong reachability fact, so unconnected machines are polled and then reported as silent.**

`selectable` is documented as "a tunnel is NOT required" (`workspace-machines.ts:57-62`), because it governs ssh directory listing. Reaching that machine's *API* needs `MachineInfo.connection` ("Present means `/server/<id>/api/…` has somewhere to go", `api/types.ts:3973-3980`).

A Project with two installed-but-not-currently-connected machines shows the amber "2 machines did not answer" strip permanently, refreshed every 15 s, and each tick spends two failing proxy attempts — whose ssh timeouts can outlive the 15 s interval.

packages/web/src/lib/session-seen.ts
```diff
@@ -209,7 +210,41 @@ export function forgetSession(
...
+// The bridge from other tabs: `storage` fires in every OTHER same-origin document when one
+// writes, so a marker stamped over there invalidates the copy held here. Coming back to a
+// backgrounded tab re-reads everything as well — a window that was hidden may have missed
+// nothing, but the re-read costs one parse per Project and never shows a stale dot.
+if (typeof window !== "undefined") {
+  window.addEventListener("storage", (event) => {
+    if (event.key !== null && event.key.startsWith(SEEN_KEY_PREFIX))
+      dropSessionSeenCache(event.key);
+  });
+  document.addEventListener("visibilitychange", () => {
```
Myriad-Dreamin, line 269: **The new `visibilitychange` reset throws away markers that only exist in memory.**

`write()` catches a failing `setItem` on purpose — "Quota / private mode: the in-memory copy still serves this tab". With the module-level listener, backgrounding and re-showing the tab calls `resetSessionSeenCache()`, which drops that copy; the next read re-parses storage, which never got the write.

In a private window (or past the quota) every Session the user read lights up as "to review" again after a single tab switch — in the sidebar and in the new dashboard. Dropping only keys that storage actually holds, or skipping the reset when the last write failed, avoids it.

packages/web/src/features/dashboard/dashboard-page.tsx
```diff
@@ -0,0 +1,178 @@
...
+          machineId: server.machineId,
+          machineLabel: server.label,
+          local: server.local,
+          sessions,
+        };
+      }),
+    );
+    const sources = answers.flatMap((a) => (a.status === "fulfilled" ? [a.value] : []));
+    const failed = answers.flatMap((a) => (a.status === "rejected" ? [a.reason as unknown] : []));
+    if (sources.length === 0) {
```
Myriad-Dreamin: **When every server fails, `silent` and `rows` are left stale.**

The early `return` after `setError` skips `setSilent`/`setSources`. One machine of two down → strip says "1 machine did not answer"; the next poll fails for *both* (this server restarts) → the page shows the old rows, the old "1 machine" strip, plus an error line — i.e. it still asserts counts it no longer has any answer for.

packages/server/src/services/session-service.ts
```diff
@@ -226,6 +227,26 @@ export class SessionService {
...
   }
 
+  /**
+   * The dashboard's read: every non-archived Session of the Project over every Agent, as the
+   * facts it counts from. Off the index and the manager and nothing else — no Trace
+   * discovery, no source classification: `hasTrace` is the row's own flag, and the counts are
+   * about this moment. Read versus unread is the client's (a per-browser marker), which is
+   * why this hands over facts rather than totals.
+   */
+  sessionsOverview(projectId: string): SessionActivityInfo[] {
```
Myriad-Dreamin: **The overview is an unpaginated full-Project dump polled every 15 s.**

`listByProject` returns every non-archived row with no limit, while the list route paginates for exactly this reason. A Project that has run evaluations (one Session per Case × Run) or many subagents answers with thousands of objects every 15 s, per server asked, for a page whose entire output is two numbers per Workspace.

packages/web/src/components/ui/icons.tsx
```diff
@@ -162,6 +162,9 @@ export function CloseButton({
 }
 
 /** Info circle: the app's 9-radius status circle with a bar and a dot inside it. */
```
Myriad-Dreamin: The new constant was inserted between `INFO_ICON` and its doc comment: `/** Info circle: … */` now documents `ADD_ICON`, and `INFO_ICON` (line 168) is left undocumented. Editor hover on `ADD_ICON` reads "Info circle: the app's 9-radius status circle…".

packages/web/src/features/terminal/terminal-keys.ts
```diff
@@ -0,0 +1,86 @@
...
+ * nothing surprising and it sending one stray ^V.
+ *
+ * A pairing with no control code (Ctrl+1) sends the plain character, matching what a
+ * physical keyboard does.
+ */
+export function applyModifiers(data: string, mods: TerminalModifiers): string {
+  if (!hasModifier(mods)) return data;
+  let out = data;
+  if (mods.ctrl) out = controlCode(data) ?? data;
+  if (mods.alt) out = `\x1b${out}`;
```
Myriad-Dreamin, line 86: **An armed Alt prefixes a whole paste with ESC, contrary to the function's own contract.**

The docstring promises "Anything longer — a paste, an IME commit … passes through untouched", but the length guard only covers `ctrl`; `out = `\x1b${out}`` runs for any length. Tap Alt (or tap it by mistake), then the paste cap: `term.paste()` fires `onData` with the whole clipboard, which arrives as `ESC ls -la`, and readline consumes `ESC l` as `downcase-word` — the first character of the pasted command is eaten. The unit tests cover the Ctrl case only.

packages/web/src/features/terminal/terminal-keybar.tsx
```diff
@@ -0,0 +1,180 @@
...
+  // it is what makes the row overflow (and scroll) on the narrowest phones instead of
+  // squeezing the caps down to slivers.
+  // Shape here, colour from the chrome — and the two tones REPLACE each other rather than
+  // stacking: this app has no class merger, so `border-gray-300` and `border-amber-600` on
+  // one element resolve by stylesheet order, not by which was written last.
+  const capShape =
+    "flex h-9 min-w-8 flex-1 basis-0 items-center justify-center rounded border px-1 text-xs font-medium";
+
+  /** Sends a sequence the bar composed itself, then spends any armed modifier. */
+  const send = (data: string): void => {
```
Myriad-Dreamin, line 75: **Caps that cannot carry a modifier still spend one.**

`send()` clears `ctrl`/`alt` for Esc, Tab and `^C`, which compose nothing — arming Alt and tapping Tab silently discards the modifier instead of sending `ESC \t`, and Alt+Esc / Ctrl+Tab are unreachable from the bar. Conversely the paste cap goes through `control.current?.paste()` and never clears, so an armed Ctrl is consumed by the pasted data instead: a one-character clipboard ("c") pasted with Ctrl armed becomes `\x03` and interrupts the foreground command.

packages/web/src/features/dashboard/dashboard-page.tsx
```diff
@@ -0,0 +1,178 @@
...
+      setError(apiErrorText(failed[0]));
+      return;
+    }
+    setError(null);
+    setSilent(failed.length);
+    setSources(sources);
+  }, [projectId]);
+
+  useEffect(() => {
```
Myriad-Dreamin, line 198: **No in-flight guard on `load`.**

The 15 s interval, the visibility handler and the mount call can all run concurrently (a proxied machine request easily outlives 15 s — see the `selectable` comment above), and `Promise.allSettled` results are applied unconditionally, so an older answer can land after a newer one and set `sources`/`silent` backwards until the next tick. A request generation counter, or skipping a tick while one is in flight, would pin it.

packages/web/src/features/dashboard/dashboard-view.ts
Myriad-Dreamin: Correction to the above: this was already fixed once. #643 ("the counts unfold… and leave subagents out") was merged into this branch on 2026-09-08, and a later restack left its commit out of here — it sat in #639 instead, while this PR's description still claimed it as step 3. Restored here as 651da270a (+ 1730a573f for the changelog link): `dashboardRows` skips `source === "subagent"`. Scheduled and evaluation Sessions stay counted, which is what #643 decided.

packages/web/src/features/dashboard/dashboard-page.tsx
Myriad-Dreamin: Correction to the above: two claims were wrong. A machine with no connection does not cost an ssh timeout — the proxy answers `503 not_connected` at once — and naming it in the silent strip is the designed answer (the strip says to connect it on the Machines page), not a defect. What remained true: `selectable` is an ssh fact, and such a machine was sent a request on every poll that could only fail. a90fd5f7e (`dashboardServers`) counts it as silent without asking; only machines with a `connection` are requested.

packages/web/src/lib/session-seen.ts
Myriad-Dreamin: Fixed in 0473d94a3: keys whose last write never reached storage are kept across the refresh (and across a `storage` event); everything else re-reads. The visibility listener now calls `refreshSessionSeenCache`, and `resetSessionSeenCache` stays the full wipe tests use. Covered by two new cases in `session-seen.test.ts`.

packages/web/src/features/dashboard/dashboard-page.tsx
Myriad-Dreamin: Fixed in a90fd5f7e: when no server answers, the page drops the rows and the silent count along with showing the error.

packages/server/src/services/session-service.ts
Myriad-Dreamin: Withdrawn — leaving this as it is. The row is a handful of short fields, and the read is documented as index-only by design. The server also cannot bound the set: unread is decided per browser, and a Session with no marker falls back to that browser's baseline, so any settled row may count for someone.

packages/web/src/components/ui/icons.tsx
Myriad-Dreamin: Correction to the above: hover on `ADD_ICON` reads "Plus…", not "Info circle…" — TypeScript takes the JSDoc block nearest the declaration. The real defect was only that `INFO_ICON` lost its comment. Fixed in f5d04bfc0.

packages/web/src/features/terminal/terminal-keys.ts
Myriad-Dreamin: Fixed in cc5aedbda: only a lone character (one code point) carries a modifier, under Alt as under Ctrl. Tests added for a plain paste, a bracketed paste, Ctrl+Alt, and a character outside the BMP.

packages/web/src/features/terminal/terminal-keybar.tsx
Myriad-Dreamin: Withdrawn — this one was wrong. Spending the armed modifier on Esc / Tab / `^C` is the documented behaviour of `send` ("then spends any armed modifier"), a design choice rather than a defect. And the paste cap does clear: `term.paste()` arrives through `onData`, which applies the modifiers and then calls `setModifiers(NO_MODIFIERS)`. The one-character-clipboard case is real but not worth a rule of its own.

packages/web/src/features/dashboard/dashboard-page.tsx
Myriad-Dreamin: Fixed in a90fd5f7e: a poll for the same Project waits for the one in flight, and an answer for a Project that is no longer the one being polled is dropped. (The premise that a forward to an unconnected machine outlives the interval was wrong — see the correction on the machine-filter thread — but a slow connected machine still could.)

PR #614, opened 2026-09-05
fix(core,server): move the plugin host-package lookup out of import time, and the restart step out of the runtime's capabilities

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← #808 ← #718 ← #544 ← #448 ← #450 ← #577 ← #803 ← #807 ← #809 ← #810 ← **#614** ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

Two things stopped a hot push from reaching an installed release: where the plugin library looks for its host package, and what the restart step claims from the runtime.

## 1. A hot push loads where there is no host package

A push to a Windows installation was refused with `No package.json above the plugin loader at …/hmr/store/platform`. The plugin library computed its package root at module load by walking up from `import.meta.url`, and the bundle `deploy.mjs` pushes sits where nothing above it is a package. The host package is found on the **first library call** now, and looked for above the running program as well as above this module — which is where a pushed platform's plugins are installed. A package.json that cannot be read on the way up is stepped over, so somebody else's file in the data root's ancestry cannot hide the host package above the program.

## 2. The restart step claims nothing from the runtime

`runtime:lifecycle` was a *required* capability, so no released installation could take a push at all. It carried one fact (`supervised`, which the platform reads from its own environment) and one trigger (the runtime's own graceful shutdown). Both are the platform's; the capability is gone, and a platform carrying this boots on every runtime.

Leaving, though, has to survive the runtime it was pushed onto. Presetting `process.exitCode` works on the runtime in this branch, but **every runtime built before it ends its shutdown with an explicit `process.exit(0)`** — including every release — and a hot push cannot replace that code. Without more, "Restart and update" on a pushed platform would stop the service with the supervisor standing down. The restart code is therefore also forced from an `exit` listener, the last word Node gives anyone; both runtimes leave with 75. That listener is the compatibility half and can be dropped one release from here, which the changelog entry records.

## Verification

`pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` and `pnpm -r test` (6688 tests) clean on this branch. The push to the Windows installation that produced the original error went through (`[deploy] ok in 12.0s`): the bundle loads from the store and the library resolves the plugins beside the program.


Myriad-Dreamin:
With this fix and #615 applied, the push to the Windows installation that produced the error went through (`[deploy] ok in 12.0s`); the bundle loads from the store and the library resolves the plugins beside the program.

PR #615, opened 2026-09-05
fix(server): the restart step claims nothing from the runtime

Myriad-Dreamin:
> **Stack:** `main` ← #551 ← #448 ← #450 ← #544 ← #577 ← #614 ← **this** ← #617 ← #613 ← #618 ← #619 ← #623 ← #622 ← #624 ← #625. Review the diff against `fix/plugin-library-host-package`.

Pushing the current platform to an installed release is refused at boot:

```
/api/hmr/upgrade → 400: the pushed platform failed to boot: this runtime publishes no business capabilities this platform can claim (config: missing supervised) — update the installation itself; a push replaces the platform, never the runtime
```

#540 (the software-update modal) added `runtime:lifecycle` as a **required** capability and `supervised` as a required member of the runtime's published `config`. Every published release predates it (v0.2.9 is the latest tag; #540 landed after), so no existing installation can take a hot push. That floor was never called out.

## The capability is unnecessary

`LifecycleService` carried one fact and one trigger, and the platform needs neither from the runtime:

- **The fact** — `supervised` — is `PENGUIN_SUPERVISED === "1"`, the supervisor's announcement in the environment of the very process the platform runs in. The platform reads it itself.
- **The trigger** is the runtime's own graceful shutdown, the one it registers on SIGTERM, exiting with `SERVER_RESTART_EXIT_CODE` instead of 0. The platform presets that code on `process.exitCode` and raises the SIGTERM event in-process (Windows delivers no signals, so it is the event, not a signal); the runtime's shutdown honours a preset code instead of forcing 0.

An old runtime never sets `PENGUIN_SUPERVISED`, because the supervisor that sets it arrived with #540 too, so on such a runtime the route answers `no_supervisor` and never exits the process. On a new runtime the supervisor, the variable and the exit-code handling ship together. No combination exits a process nobody relaunches.

## What changed

- New `services/process-restart.ts`: `processRestart()` → `{ supervised(), request() }`, injectable for tests; the version routes take it as `restart`.
- Removed: `LifecycleService`, `runtime:lifecycle`, the descriptor's `lifecycle` entry and its `config.supervised` member, `ServerConfig.supervised`, and the trigger registration in index.ts. `shutdown()` takes an optional exit code and leaves an undefined one to `process.exitCode`.
- The CLI supervisor (`penguin server|web`) is unchanged.
- No compatibility code and nothing to remove later: a platform carrying this boots on every runtime.

One thing to know: a runtime built from `main` between #540 and this change forces exit code 0 in its shutdown, so a restart requested through it stops the service without a relaunch. No released runtime is in that window; a rebuild puts a dev install past it. Recorded in the changelog entry.

## Verification

- New `process-restart.test.ts` (the environment read; the preset code and the raised event). `version.test.ts` (`no_supervisor` unsupervised), `hmr-resources.test.ts`, `extension.test.ts`, `park-quiesce.test.ts`, `hmr-host.test.ts`; then the whole server suite: 122 files / 1744 tests passing. Server typecheck, prettier, `git diff --check` clean.
- Pushed with #614 to a Windows installation whose runtime predates #540: see the follow-up comment.


Myriad-Dreamin:
Pushed to a Windows installation whose runtime predates #540 (a source build of feat/workspace-pr-chip, v0.2.9-61), with #614 applied as well and the machines stack (#577's tip) as the platform: `[deploy] ok in 12.0s`. `POST /api/version/restart` there answers `{"restarting":false,"reason":"no_supervisor"}`, as expected of a runtime nobody relaunches.

One addition after the first round: on the module-tree stack the kernel checks the runtime's config node against the `Config` interface, which promises `supervised`, so the compat branch also fills `config.supervised = false` in place when the runtime published no lifecycle — the same in-place treatment the auth state gets. Folded into the commit; the test pins it.

Myriad-Dreamin:
Pushed to a Windows installation whose runtime predates #540 (a source build of feat/workspace-pr-chip, v0.2.9-61), with #614 applied as well and the machines stack (#577's tip) as the platform: `[deploy] ok in 13.6s`. `POST /api/version/restart` there answers `{"restarting":false,"reason":"no_supervisor"}` — read off the environment, nothing claimed.

PR #617, opened 2026-09-05
fix(server): a machines table from before its migration gains its columns, and a failed job always offers the way out

Myriad-Dreamin:
> **Stack:** `main` ← #551 ← #448 ← #450 ← #544 ← #577 ← #614 ← #615 ← **this** ← #613 ← #618 ← #619 ← #623 ← #622 ← #624 ← #625. Review the diff against `fix/lifecycle-capability-optional`.

Two regressions seen on a Windows installation running the machines stack, whose data root had run the machines line before it was released.

## 1. `table machines has no column named session_pid`

A connect failed at "Opening the connection…" on the row it writes. Migration 4 creates the machines tables with `IF NOT EXISTS`, so a table already there — created by the earlier line's own migration, with `forward_port`/`forward_pid` and no `session_pid`/`platform` — was adopted as it stood and stamped to 4.

**Migration 5, `machines-columns`**: `ensureColumn` for `session_pid` and `platform`; a complete table is left alone. Additive and `swapSafe`, so a hot push applies it. Its `down` does nothing: what it adds, 4 already declares, and a column it found in place was never its to take away. Test: an adopted table at version 3 migrates through 4 and 5, gains both columns, takes the row the connect writes, and a second run applies nothing.

## 2. No way on after a failure

The install-anyway button appears only when the result carries `canReplaceProgram`, and only three failures set it. Every other failure — this one included — left the page with nothing to offer.

Now the job wrapper (`#startJob`) adds the offer to **every failed install or connect**, except a run that was itself that install, and a result that withholds it with `canReplaceProgram: false` (this server has no build of its own to send — installing would not mend that). The three explicit offers are gone; the rule is in one place. The API field widens from `?: true` to `?: boolean`; the page's `=== true` check is unchanged. Still never taken on the page's own initiative.

The page copy for the button (on the machines stack, #448) described the hand-over case; it will be reworded to say what the button does when the stack is rebased over this.

## Verification

- `db-migrations.test.ts` (new adopted-table case; the applied/reverted lists), `machines-api.test.ts`, `machines.test.ts`; then the whole server suite: 121 files / 1743 tests passing. Server typecheck, prettier, `git diff --check` clean.
- Pushed to the installation that produced the error, on top of #577 with #614/#615 applied. That first push changed nothing: the module-tree line had dropped the `migrate(db, { swapPath: true })` call `buildAppDeps` makes (main still has it), so a pushed platform there applied no migration at all — restored on `feat/mechanisms` (#546) as 77cf214d. With that, migration 5 applied on the push and the connect that failed went through; see the follow-up comment.


Myriad-Dreamin:
Verified on the Windows installation that produced the error (runtime from before the machines line was released; platform = #577's tip with #614, #615, this PR and the #546 swap-path fix applied):

- Before the swap-path fix, the connect to `kamiya-hypoxic` failed the same way but now offered the way out: `{"ok":false,"step":"connect","message":"table machines has no column named session_pid","canReplaceProgram":true}`.
- With the pushed platform migrating again, migration 5 added the columns and the same connect answered `{"ok":true,"connected":true}` — log: "Connected; its server is on port 7364 over there. Models synced: default_project."

PR #620, opened 2026-09-05
fix(core): accept 'command' as an alias for exec_command's 'cmd'

paskaa:
## Problem

Some agentic models (observed: dots3-note-prev, a Claude-Code-era model) emit the shell command under the parameter name `command` even when the exposed tool schema says `cmd`. `exec_command` then fails validation with:

```
Missing required argument "cmd" for exec_command.
```

The model sees the schema, believes it *did* send `cmd`, and retries the exact same malformed call — producing a fatal retry loop that yields no usable output.

Observed in the wild: one 22-minute agent session where **124 of 155** `exec_command` calls failed this way (first 2 calls used `cmd` and succeeded; then the model switched to `command` and never recovered). Same harness, same prompt on other models: zero failures.

## Fix

In `createExecCommandTool`, accept `command` as a legacy/foreign alias for `cmd`:

- `cmd` stays authoritative when both keys are present (schema is unchanged; models that send `cmd` are unaffected).
- If only `command` arrives, it is honoured, so a slightly-off-schema call executes instead of dying.
- The validation error now names the received key when `command` was the only key supplied, so a model in this loop gets actionable feedback instead of an opaque repeat.

## Tests

Added 3 cases in `packages/core/test/environment.test.ts`:

1. `command` alone executes successfully and writes the file.
2. `cmd` wins when both `cmd` and `command` are present.
3. Neither key → still fails with the `Missing required argument "cmd"` message.

## Verification

- `vitest run test/environment.test.ts` — 41 passed (7 in the `command`-filtered run include the 3 new cases).
- `tsc --noEmit` clean; `oxlint` + `prettier` clean on changed files.
- Two pre-existing `exec-session.test.ts` proxy-env failures are environmental (host `NO_PROXY` injection) and reproduce on the untouched base commit.

### DIFF
--- a/packages/core/src/agent.ts
+++ b/packages/core/src/agent.ts
@@ -983,6 +983,7 @@ export class Agent {
               ...(visionEntry.context_window !== undefined
                 ? { contextWindow: visionEntry.context_window }
                 : {}),
+              ...(sessionId !== undefined ? { sessionId } : {}),
               requestTimeoutMs: 60_000,
             }),
         };
@@ -1057,6 +1058,7 @@ export class Agent {
         // model that rejects it.
         ...(modelEntry.fast_mode === true ? { fastMode: true } : {}),
         ...(context.thinkingLevel !== undefined ? { thinkingLevel: context.thinkingLevel } : {}),
+        ...(sessionId !== undefined ? { sessionId } : {}),
         ...(context.requestTimeoutMs !== undefined
           ? { requestTimeoutMs: context.requestTimeoutMs }
           : {}),
@@ -1126,6 +1128,7 @@ export class Agent {
         ...(modelEntry.context_window !== undefined
           ? { contextWindow: modelEntry.context_window }
           : {}),
+        ...(sessionId !== undefined ? { sessionId } : {}),
         requestTimeoutMs: 30_000,
       });
 
--- a/packages/core/src/environment/tools/exec-command.ts
+++ b/packages/core/src/environment/tools/exec-command.ts
@@ -61,9 +61,16 @@ export function createExecCommandTool(
         return { stopReason: "fatal" };
       }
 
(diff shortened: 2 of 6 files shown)


hiyouga:
Thanks for this, and for the session data — 124 of 155 calls lost to the same rejected argument is exactly what made the case.

The alias has since landed on `main` in #661 and shipped in v0.2.11: `exec_command` runs a call that carries the shell text as `command`, the schema still declares `cmd` alone, and `cmd` wins when both are present — the same three behaviours your tests pin, and they're covered by tests on `main`.

Two things the merged version adds, which is why it went in as its own change rather than as a merge of this branch:

- **The alias is read through one exported list** (`EXEC_COMMAND_TEXT_ARGUMENTS` in `packages/core/src/environment/tools/exec-command.ts`), shared by the tool, the command-policy gate (`packages/core/src/internal/command-policy.ts`) and the Web/CLI call previews. That part matters: the policy mapped `exec_command` to the single name `cmd`, so with the alias living inside the tool only, a `command`-keyed call would have been executed without being screened by the policy, and the approval preview would have fallen back to raw argument JSON instead of the `$ …` line.
- **A rejected call now answers with a full correction guide** instead of one line — the fault, the argument names received with the undeclared ones called out, the tool's parameters restated from the schema the model was handed, and the shape of a correct call. It applies to every built-in tool's argument validation, not just `exec_command`.

The other half of this PR — threading `sessionId` through `GenerativeModelConfig` and emitting `x-opencode-session` for `opencode.ai` / `freqtrade.1unlock.top` — is unrelated to the problem reported here and was not taken. It's a separate decision; if you'd still like it, please open it as its own PR and we'll look at it there.

Closing as superseded (this branch also conflicts with `main` now). Thanks again — the diagnosis was right, and the bug is fixed because you reported it.

PR #622, opened 2026-09-05
feat(server,web): the Machines page is two verbs — use, stop using — and can declare an ssh host

Myriad-Dreamin:
> **Stack:** `main` ← #354 ← #543 ← #546 ← #383 ← #526 ← #551 ← #448 ← #450 ← #544 ← #577 ← #614 ← #613 ← #619 ← #623 ← **this** ← #625 (a sibling fork off #613: #626 ← #627 ← #628 ← #629 ← #630 ← #633 ← #634 ← #637). Review the diff against `feat/outline-index`.

The Machines page reduced to what a person actually asks of a machine, plus the one step that had to happen outside the page before any of it worked. #624 was a separate PR on this branch and is folded in here.

### What

The Machines page becomes two verbs and one sentence per machine.

- **Use** does everything a machine needs to run agents — install or bring the build forward when the record says so, start its server, connect, hand over the Model config — as one job. **Stop using** drops the connection and the Project membership and leaves the install alone.
- Rows carry checkboxes; every machine in use is ticked by default, so "make them all work" and "let them all go" are one tap. A batch is one POST; the server queues the jobs one behind another and each row says whether it waits, works, is ready, or what went wrong.
- Machines not yet in use live behind **Add machines…**, a search over the ssh config (it can declare hundreds of hosts). Picking several and confirming uses them all.
- Held is a standing intent: every minute the service re-holds machines whose record says held and that are not, with a per-machine backoff doubling from one minute to fifteen. A restart or a dropped host no longer waits for someone to visit the page.
- Version, install date, last check and the job's output fold under each row. The forced install stays where it was: on a failed row only, with its warning.

### Why

The old page exposed the pipeline's steps as separate controls — install, update, restart, connect, disconnect — and left a person to order them by hand and to interpret "running · port 7364 · not connected". For someone who is not a computer person, the only questions are "can I use this machine?" and "how do I make it usable?", and the answer to the second is always the same button. The batch is what a fleet needs: reconnecting five machines after a server restart was five separate connect clicks, and now it is none (the standing intent) or one.

### Server

- `MachinesService.startUse(projectId, addresses, replaceProgram?)` queues one `use` job per machine and answers refusals decidable without ssh (`unknown-machine`, `self`, `no-image`) by id. The job body is `#installWork` (the old install closure, extracted) when the record says the build is missing or different, then `#connect`. Windows stays at installed, as `startConnect` already refuses it.
- `#startJob` is now a queue with one runner (`#drain`): jobs still run one at a time, and `jobs()` lists the queued, running and last finished job per machine. `job()` is unchanged. A machine queued twice keeps the later ask.
- `stopUsing(projectId, addresses)` = disconnect + release, per machine.
- `start()` arms an unref'd interval (`KEEP_HELD_MS`) calling `autoConnect()`; `stop()` clears it. `autoConnect()` skips machines inside their backoff window and doubles it on each failed re-hold.
- Routes: `POST /api/projects/:projectId/machines/use` (`{ machines, replaceProgram? }`, 202, `refused` in the body), `POST …/stop-using` (`{ machines }`). The per-machine routes stay for anything that still calls them.

### Web

- `machines-view.ts` is rewritten around `readMachine(machine, job, imageVersion)`: a fixed precedence (queued → working → held-connection-is-ready → failed → installed-only → behind → probe) so a machine a re-hold brought back after a failed job reads as ready. `defaultSelection` ticks every machine in use.
- `machines-page.tsx` is rewritten (batch bar, rows, add picker). The bilingual `machines` strings are rewritten; every unused key is gone.

### Tests

- Server: a new `using machines` block (queue order and per-row outcomes, skip-when-current, bring-forward-when-behind, a failing row does not stop the batch, refusals by id, stop using, the re-hold backoff). 72 in the file, 170 across the machines suites.
- Web: `machines-view.test.ts` rewritten for the reading's precedence and the default selection; full web package 1730 passing.

### Screenshots

**The page at rest — one card per machine, this server first, verbs in a fixed slot above the cards**

The page at rest — one card per machine, this server first, verbs in a fixed slot above the cards

**A card unfolded: the record, the verbs, then the output**

A card unfolded: the record, the verbs, then the output

**Dark, two cards selected**

Dark, two cards selected

**Phone, light**

Phone, light

**Phone, dark**

Phone, dark

### What

A machine that is not in the server's ssh config yet can be declared from the Machines page.

- The **+** at the foot of the **Add machines…** panel, beside the chevron that unfolds the rest of the config, opens a short form: alias (`Host`), address (`HostName`), and optionally `User`, `Port` and `IdentityFile`. Same `Modal` + `Input` shape as the vault's add-secret form; format rules stay visible as field hints, and the first bad value is reported under its field before anything is sent.
- The server appends the matching block to its own `~/.ssh/config`, led by `# Added by PenguinHarness on <time>`, so a person reading the file later knows which lines are not theirs. It creates `~/.ssh` (0700) and `config` (0600) when missing and separates the block from a file that did not end in a newline.
- The new host then appears under **Add machines…** like any other and can be enabled at once.
- A host this app wrote can be reconfigured: in a card's details the verbs now sit above the output, and a gear beside **Disable** reads the block back (`GET …/ssh-hosts/:alias`) and opens the same form on it; saving rewrites it in place (`PUT …/ssh-hosts/:alias`). A hand-written block is shown but not saved — it may carry options the form does not know — and the form points at the file.

### Why

Until now a machine had to exist in the ssh config before the page could do anything with it, and writing that file is exactly the step a non-technical person cannot take. The page already promises "enable and it works"; this closes the one gap before enable.

### Server

- `ssh-config.ts` gains the pure half: `SshHostEntry`, `validateHostEntry` (one word per value, no `#`, alias not a pattern, port 1–65535), `renderHostBlock`, and `findHostBlock` / `replaceHostBlock` for reading a block back and rewriting it (a block is ours when it carries the marker comment; a `Host a b` line is never matched). `transport/targets.ts` gains the writes: `appendHostBlock`, `readSshConfig`, `writeSshConfig`. The service's `addSshHost` validates, refuses an alias the config already declares (ssh takes the first block; a second would be silently ignored), and appends.
- `POST /api/projects/:projectId/machines/ssh-hosts` → 201 with the machines list; 400 `ssh_host_invalid` naming the field; 409 `ssh_host_exists`. Admin only, like the rest of the group.

### Tests

- `ssh-config-hosts.test.ts`: what is refused, the exact lines written, and finding and replacing a block without touching its neighbours.
- `machines-api.test.ts`: the route appends and the list names the host at once; a declared alias is refused with nothing written; a bad field is named.

### Screenshots

**The Add machines panel: matches, then a foot with Expand and New**

The Add machines panel: matches, then a foot with Expand and New

**The new-host form**

The new-host form

**A card's details: Disable beside Configure, which opens the same form on the host's block**

A card's details: Disable beside Configure, which opens the same form on the host's block

### DIFF
--- a/packages/web/src/features/machines/machines-page.tsx
+++ b/packages/web/src/features/machines/machines-page.tsx
@@ -494,9 +494,9 @@ export function MachinesPage() {
         {/* The selection bar: a fixed slot between the title and the cards, so the cards
             never move when a selection appears or goes. The count is the slot's label; on the
             right, select all and none, then the two verbs, each dimmed when it would do nothing. */}
-        <div className="mt-3 flex h-10 items-center gap-3 px-1 text-xs text-gray-500">
+        <div className="mt-3 flex min-h-10 flex-wrap items-center gap-x-3 gap-y-2 px-1 text-xs text-gray-500">
           <span className="tabular-nums">{S.machines.selectedCount(selectedIds.length)}</span>
-          <span className="ml-auto flex items-center gap-1">
+          <span className="ml-auto flex flex-wrap items-center gap-1">
             <Verb
               label={S.machines.pickAll}
               d={SELECT_ALL_PATH}
@@ -599,6 +599,7 @@ function Verb({
       disabled={disabled}
       title={title ?? label}
       aria-expanded={ariaExpanded}
+      className="whitespace-nowrap"
       onClick={(event) => {
         event.stopPropagation();
         onClick();


Myriad-Dreamin:
Folded into #448 as real commits (cherry-picked onto `feat/machines-identity`, with #624 and #617 alongside): the Machines page lands as one PR, and #577 stays the gate on top of #450. The changelog entry is now `2026-09-09-machines-page` on #448; this branch stays as it was.

PR #627, opened 2026-09-06
feat: one plugin list per Project, applied without a restart, kept the same on its machines

Myriad-Dreamin:
> **Stack:** `main` ← #354 ← #543 ← #546 ← #383 ← #526 ← #551 ← #544 ← #448 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← **this** ← #628 ← #630 ← #634. Review the diff against `feat/claude-code-plugin`.

## What

Plugins stop being a per-deployment file that only takes effect at process start. A Project owns its list, the process runs the closure over its Projects, a write applies without a restart, and the list is kept the same on the machines that Project uses.

Design: **PRFC-0010** (penguin-harness-design#124). Stacked on #626.

## Why

Raised by the operator: "we manage plugins through the remote machines" and "a plugin must be enabled on all machines at once". There is no path in the product to log into each machine's server and enable a plugin there — the Machines page is how remote is managed. Session surfaces (#626) make it concrete: a Session on a machine opens its surface there, so that machine needs the plugin, and so does this server or the New chat picker never offers it.

## Shape

- **core** — `plugins` in `.project_config.toml`, parsed leniently.
- **server** — the loader reads the closure (union over a root's Projects) from files, before the platform exists. Routes move under the Project. A write re-reads the closure into a fresh plugin host and re-assembles the App from the same bundle, the swap `upgrade()` already performs.
- **server** — `machines/plugins-sync.ts`, mirroring `models-sync.ts`, on the same trip and triggers. Strict parity.
- **web** — the plugin surfaces act on the Project in view.

## The detail that matters most

`Hmr.reload` is an optional **field**, not a method. The kernel's signature check tolerates a field a runtime does not declare and refuses a platform whose required **method** is missing — which is exactly how `config.supervised` once made every push fail on the Windows runtimes. An older runtime keeps taking pushes and reports `restartPending` instead.

## Breaking

The data root's `plugins.json` is no longer read, and nothing migrates it — the operator chose this over a migration or a dual read, so that no compatibility code has to be carried or cleaned up later. A deployment that had one starts with **no plugins** until each Project asks again. The old file is left on disk untouched, so a rollback finds it.

Strict parity has its own cost, recorded in the changelog and the RFC: a platform-specific sandbox backend (`sandbox-mxc` on Windows, `sandbox-bwrap` on Linux) is not in the other's list and is therefore removed. Keeping one means listing it fleet-wide.

## Verification

Server suite 1891 (including six sync cases and the rewritten loader/route tests), core 1111, web 1737, docs 53. Full `pnpm build`, `pnpm typecheck`, `pnpm format:check` clean.

### DIFF
--- a/packages/plugin-test/README.md
+++ b/packages/plugin-test/README.md
@@ -2,7 +2,7 @@
 
 Integration testing for PenguinHarness plugins, the way `@vscode/test-electron` tests a VS Code extension: start the **real** host with the plugin under development installed, then run the tests against it.
 
-The host is the server (`@prismshadow/penguin-server`'s built entry), started as a child process on a scratch data root whose `plugins.json` lists the plugin. What a test exercises is the path a deployment takes — the loader resolves the package, reads its manifests, pairs them with its code, and boots its modules into the platform tree. Nothing is assembled in-process, nothing in the host is mocked.
+The host is the server (`@prismshadow/penguin-server`'s built entry), started as a child process on a scratch data root whose Project config lists the plugin. What a test exercises is the path a deployment takes — the loader resolves the package, reads its manifests, pairs them with its code, and boots its modules into the platform tree. Nothing is assembled in-process, nothing in the host is mocked.
 
 ```ts
 import { startHarness, waitFor } from "@prismshadow/penguin-plugin-test";
@@ -13,7 +13,7 @@ const harness = await startHarness({
 });
 try {
   const api = await harness.login();                    // the seeded admin
-  const { plugins } = await api.get("/api/plugins/installed");
+  const { plugins } = await api.get(`/api/projects/${harness.projectId}/plugins/installed`);
   const { session } = await api.post("/api/projects/default_project/agents/default_agent/sessions", {
     surface: "claude-code",
   });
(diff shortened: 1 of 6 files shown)


Myriad-Dreamin:
Split three ways rather than merged into one place, so closing rather than claiming a merge that would show the wrong diff.

| where it went | what |
|---|---|
| **#614** | the plugin list belongs to a Project and the process runs the closure; the sync's strict parity and its 404 report; the plugin apply that puts the previous host back; **and the whole HMR-layer work** — the registry's two halves, the platform loading its own plugins, the layer renamed to HMR, the mechanism extracted as `@prismshadow/penguin-hmr`, the two push/plugin fixes, and the CI filter they exposed |
| **#625** | the two *below the hot seam* moves — `/api/command` and `/api/desktop/update` leaving the layer. They belong beside the PR that created `/api/command` in the first place |
| **#626** | `fix(plugin-test): seed the Project's plugin list, not the removed root file` — a change to `packages/plugin-test` and `plugins/claude-code`, which is where those live |

Nothing was rewritten in the move: the tree at the top of the rebuilt chain is byte-identical to this branch's, which is how the split was checked.

Myriad-Dreamin:
Folded on 2026-09-09: the per-Project plugin list, the restart-free apply and its fix live in #383 (the plugin platform), the machine parity in #448 (the Machines page). The chain's top tree is unchanged; this branch stays as it was.

PR #629, opened 2026-09-07
Machines: a glyph-only selection bar, and the server root on a card

Myriad-Dreamin:
> **Stack:** `main` ← #354 ← #543 ← #546 ← #383 ← #526 ← #551 ← #448 ← #450 ← #544 ← #577 ← #614 ← #613 ← #626 ← #627 ← #628 ← **this** ← #630 ← #633 ← #634 ← #637 (a sibling fork off #613: #619 ← #623 ← #622 ← #625). Review the diff against `feat/devtools-command`.

Two small things on the Machines page.

## The selection bar goes back to glyphs

The four buttons in the bar — select all, select none, use, stop using — drop their words. They got them when every verb on the page was unified into one shape (#624); in the bar that reads as a sentence nobody wrote, and each glyph is already the plainest thing about its button: the box, the empty box, the plug, the unplugged plug.

The word survives as the tooltip and as the accessible name, so a screen reader and a waiting pointer lose nothing. Verbs that stand alone — configure, new, force install — keep their word: `wordless` is opt-in, and the bar is the only caller.

## A card names its server root

`MachineInfo.root` — the data root the server there runs on (`PENGUIN_HOME`), shown in the unfolded details.

This is not decoration: the **profile** decides which installation a machine's row is about (`machines/layout.ts`), so a dev instance and a release instance can show the same alias and mean two different servers on that host. The page never said which. Written in that machine's own spelling once its platform is known (POSIX before that, corrected by an install); the local entry names this process's own root, already resolved.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` — clean.
- `pnpm --filter @prismshadow/penguin-server --filter @prismshadow/penguin-web test` — 1894 + 1739 pass. The list's expectation now pins both roots: the local entry's is the process's own, a never-installed host's is the layout's POSIX spelling.


Myriad-Dreamin:
Folded into #448 as real commits (cherry-picked onto `feat/machines-identity`, with #624 and #617 alongside): the Machines page lands as one PR, and #577 stays the gate on top of #450. The changelog entry is now `2026-09-09-machines-page` on #448; this branch stays as it was.

PR #646, opened 2026-09-08
A Workspace group is a directory on a machine, not a path

Myriad-Dreamin:
`/srv/app` on this server and `/srv/app` on a machine are two different directories. The sidebar keyed its Workspace groups on the path alone, so once the Session list started merging several machines (#450), two of them arrived as **one folder** — and every action that folder offers went somewhere wrong:

- its **"+"** created the chat on this server, in whatever this machine happens to have at that path — silently the wrong directory when the path exists on both (a repo checked out on two machines is the common case);
- its **badge** was both machines' counts added together, contradicting the rows under it;
- its **pages** were asked of every machine, so `load more` could offer rows that belong to another machine's folder and never arrive in this one, and the folder stayed "not loaded" until a machine it is not on had answered.

`lib/workspace-registry.ts` had the rule right in its header — *"a path is only meaningful together with its machine … the pair is the identity"* — and stored `machineId` per entry. Every lookup over those entries still matched on the path alone.

## What changed

- **`workspaceGroupKey(path, machineId)`** — one group per (machine, path). A machine of `null` keys exactly as before, so every collapse state, pin and manual order already persisted in a browser still resolves; only groups that could not exist before get new keys.
- **The group is named for its machine**: `app [SSH: prod-1]` (`nameOnMachine`, bilingual `S.chat.onMachine`), on the group label, its tooltip, and the Workspace line of a Session's detail panel. This server's own names stay bare — marking them would mark everything. An unlabelled machine (the machine list is admin-only, and a host can drop out of `~/.ssh/config`) falls back to its own id rather than inventing a name.
- **Counts fold per group**, not per path string, in the store (`workspaceCountsByAgent` is keyed by group key) and in `aggregateWorkspaceCounts`.
- **A group's fetches go to its own machine only** (`sourcesFor`): `loadMoreFor`, `isLoadedFor` and `hasMoreFor` consult the machine the group is on; the Agent's whole stream still fans out over every source.
- **The registry matches the pair** in `loadWorkspaceRegistry` (which dropped one of two same-path entries for good), `setWorkspaceAlias` (which renamed the other machine's entry, and rewrote it *without* its machine — moving it here), `unregisterWorkspace` (which removed both) and `mergeRegisteredWorkspaces`.
- The sidebar's `machineOfWorkspace` is gone: it read the registry by path and returned "this machine" for every session-derived group, which stopped being true when the list started merging machines.

Seeing the naming needs a machine with a workspace on it, so there are no screenshots here — the behaviour is covered by the tests below.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format` + `pnpm format:check` — clean (Node 24).
- `pnpm test` — web 141 files / 1802 tests and server 139 files / 1916 tests pass (run per package with `--maxWorkers=2`: the whole suite at once is OOM-killed on this box, which is a machine limit, not a failure). New cases:
  - `session-grouping.test.ts`: a machine's key and its two halves, the query dropping the machine half, one path across machines splitting into two groups, each machine's temp workspaces staying its own group, and counts for one path on two machines staying two groups.
  - `workspace-registry.test.ts`: both machines' entries surviving a load, renaming and removing acting on one machine's entry only (and keeping its machine), and the merge giving one group per machine with the alias landing on its own.
  - `session-aggregation.test.ts`: the store keeping two machines' counts for one path apart, and a group's page being asked of exactly the machine it is on, by bare path.

Stacked on #645 (`feat/terminal-touch-scroll`), the tail of the machines chain.


Myriad-Dreamin:

PR #648, opened 2026-09-08
The Evaluation Center reads every machine the Agent runs on

Myriad-Dreamin:
Reported from the deployed build: the Evaluation Center shows no Benchmark from a machine.

It never could. `benchmark-service.ts` reads `benchmarksDir(root, projectId, agentId)` — a directory on the server that answers — and `listBenchmarks` had no machine to be asked about. The only cross-machine routing the web app has is a rule over the PATH (`lib/session-machines.ts`), and it deliberately matches `/api/sessions/<id>/…` only, so a project/agent-scoped path like `/benchmarks` always went to this server. The Session list got its fan-out (#450); this page did not.

But an Agent is **one identity** across the machines it runs on, so this is not a "group by machine" problem: `benchmarks/` is simply written wherever the evaluation ran, and one Agent's library and scoreboard end up spread over several disks. A Benchmark built on a machine was invisible; one evaluated in both places drew a chart with the rounds that ran elsewhere missing.

## What changed

- **`lib/benchmark-merge.ts`** (pure, unit-tested): folds every source's answer by benchmark id — one entry, one history — with the machine carried on each evaluation as attribution. First source (this server) wins the title/description/`runs`; `caseCount` takes the max, which is the closest true statement the list endpoint alone supports.
- **Order.** A scoreboard's append order *is* its evaluation sequence and the page trusts it over the timestamps. Two scoreboards on two disks share no append order, so a **joined** history is ordered by time and a **single-source** one is left exactly as its file had it.
- **The page fans out** over this server plus the machines this server holds a connection to (`useSessions().machineIds`, the set the sidebar already polls — no second admin-only `getMachines` call). `Promise.allSettled`: a machine that cannot answer is left out of the merge, everything this server holds still renders, and only when no source answered is there an error.
- **Attribution in the UI**: an evaluation row carries `[SSH: prod-1]` when its Benchmark was evaluated on more than one machine, and a Benchmark only one machine has is titled `name [SSH: prod-1]` in the tree. Local stays bare, per the convention from #646.
- **Cases** are the union of what those machines hold (first source wins a shared id), and `BenchmarkCaseBrowser` takes the machine its listing came from, so file listings, previews and downloads (`benchmarkCaseFileUrl` now goes through `apiUrl`) are read off the disk that actually has them.
- `listBenchmarks` / `listBenchmarkCases` / `listBenchmarkCaseFiles` / `benchmarkCaseFileUrl` take an optional machine; `SessionsContext` exposes `machineIds`.

Not in this PR: the Agent's own state (AGENTS.md, skills, memory, vault, schedules) is still a separate copy per machine, and `usage_records` is still per-server, so machine spend is missing from the cost center. Both need a design decision first, not a fan-out.

## Verification

- `pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format` + `pnpm format:check` — clean (Node 24).
- `pnpm --filter @prismshadow/penguin-web test` — 142 files / 1809 tests pass, including 7 new cases in `test/benchmark-merge.test.ts`: one id folded across machines with every holder recorded, a joined history ordered by time, a single scoreboard left in its written order despite out-of-order timestamps, first-source description with the largest case count, a machine-only Benchmark listed after this server's, no mutation of the answers, and the Case union with its per-case machine.
- Hot-pushed to a running instance (`scripts/deploy.mjs`) to check it renders.

Stacked on #646.


Myriad-Dreamin:

PR #650, opened 2026-09-09
Switching the model keeps the conversation on its machine

Myriad-Dreamin:
Reported: switching the model reports `Workspace does not exist`.

`onSwitchModel` (chat-page.tsx) opens a NEW Session for the same Agent on the picked model and deliberately carries the source Session's Workspace, so the files the conversation refers to stay reachable. It called `api.createSession(...)` **without a machine** — so the request went to this server with a path that is a directory on the machine the conversation runs on. `assertWorkspaceAllowed` realpaths it, cannot find it here, and answers `400 workspace_not_found`: "Workspace does not exist or is inaccessible: …" — about a directory that exists and is reachable where the Session actually lives.

The creation now goes to `machineForSession(selected.sessionId)`. `createSession` records the new Session against that machine, so the follow-up `postTask` (the `[model_switch_from]` block plus the user's remainder) reaches the same server, and the trace the new model reads for context is on the disk that serves it.

## The same class, found while looking

- **`/agent` handoff** (same file, just below) also creates on this server. It passes no Workspace, so it cannot fail this way — but a handoff from a conversation on a machine silently starts the new chat *here*, with a temporary workspace, instead of beside the work. Left alone: the target Agent is picked from this server's Agent list, and creating it over there would 404 for an Agent that machine has never heard of. It becomes a one-line change once the Agent list is the union across machines (the follow-up to #648).
- **Schedules** carry a `workspace` and are read and written on this server only, so a schedule for an Agent that lives on a machine has the same shape of problem. That one is part of the larger gap — an Agent's state (AGENTS.md, skills, memory, vault, schedules) is a separate copy per machine — which needs a design decision, not a routing fix.

## Verification

- `pnpm typecheck`, `pnpm lint`, `pnpm format:check` clean; `pnpm --filter @prismshadow/penguin-web test` — 142 files / 1808 tests pass.

Stacked on #648.

### DIFF
--- a/changelog/unreleased/2026-09-09-model-switch-on-its-machine.md
+++ b/changelog/unreleased/2026-09-09-model-switch-on-its-machine.md
@@ -1,0 +1,11 @@
+# Switching the model keeps the conversation on its machine
+
+- **Date:** 2026-09-09
+- **Type:** fix
+- **Scope:** `web`
+
+[中文版](2026-09-09-model-switch-on-its-machine.zh.md)
+
+Switching the model on a conversation that runs on a machine answered `Workspace does not exist or is inaccessible`. The switch opens a NEW Session for the same Agent on the picked model, deliberately carrying the source Session's Workspace so the files it refers to stay reachable — but it asked THIS server to create it. The path it carried is a directory on the machine, and a server refuses a Workspace it does not have, so the switch failed with a message about a directory that exists and is perfectly reachable where the conversation actually lives.
+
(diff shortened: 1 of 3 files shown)


Myriad-Dreamin:
Folded into #450 as real commits (`d8e48ffd` + `3b6a728b`, cherry-picked onto `feat/machines-run-agent`); the changelog entry now names #450. The branch stays as it was.

PR #652, opened 2026-09-09
The server's test shards build the server's own dependencies

Myriad-Dreamin:
Every server test file fails to load, on all three platforms:

```
Cannot find package '@prismshadow/penguin-hmr/manifest'
  imported from packages/server/src/hmr/manifest.ts
Failed to resolve entry for package "@prismshadow/penguin-hmr"
```

The shards that run the server's tests build `@prismshadow/penguin-core...` — core and core's dependencies. That was the whole closure while the hot-update layer lived inside the server package. Since it became `@prismshadow/penguin-hmr` (a workspace package the server depends on), nothing in those shards builds it, so every test file dies on its first import. It is not a test failure at all: `Test Files 144 failed` with zero tests run.

`server` (Linux, macOS) and `server-1..3` (Windows) now build `@prismshadow/penguin-server...`.

The Windows `rest` shard failed for a different reason — `plugins/claude-code`'s surface test asserted the searched paths as one joined string, separator included. Cherry-picked the fix that asserts the message instead (`test(plugins): the searched-paths assertion is about the message, not the separator`).

## Verification

Reproduced the shard locally: with `packages/hmr/dist` removed, `pnpm --filter "@prismshadow/penguin-server..." build` rebuilds it, and then

- `pnpm --filter @prismshadow/penguin-server test` — 144 files / 1971 tests pass (the `agent-kernel-update` failure that showed locally was the same missing build, and is gone).
- `pnpm --filter @prismshadow/penguin-plugin-claude-code test` — 2 files / 15 tests pass.

This sits at the end of the machines/socket chain. It touches only `.github/workflows/ci.yml`, one test file and a changelog pair, so it can also be cherry-picked wherever the CI work belongs — #599 carries the other CI change on this line.


Myriad-Dreamin:
Moved into the branches the fixes belong to: the shard filter is now a commit on #599 (feat/server-module-tree, where the CI work lives) and the Windows path assertion rides with #637 (fix/claude-code-resolve-binary). The chain above both was replayed so every PR inherits them.

PR #653, opened 2026-09-09
The Agents page lists every machine's Agents

Myriad-Dreamin:
The same gap #648 hit one level up. An Agent belongs to the Project, but its state directory is created on whichever machine it has run on: `revival_worker`, which has only ever run on one machine, exists only over there. The Agents page was built from this server's list alone, so it had no row for it — nothing to click, nothing saying it exists, and (as #648 found) nowhere to hang its Benchmarks.

The page now asks this server plus every machine it holds a connection to (`useSessions().machineIds` — the set the sidebar already polls) and merges by agent id with `mergeAgents`, the fold the Evaluation Center uses. `Promise.allSettled`: a machine that cannot answer contributes nothing, and this server's list still stands.

- An Agent this server also has is described by this server; one that lives on exactly one machine is named for it — `[SSH: prod-1]` beside the name, in the alias's own casing (an Agent name renders uppercase; an alias is an identifier).
- **New chat** on such a card opens the draft against that machine with an empty Workspace — the temporary workspace ON it — so the conversation starts where the Agent's state is. (The draft applies a machine only together with a path, so the empty path is what carries it.)
- Settings and its six stat shortcuts, usage and delete all read or write a state directory this server does not have. They are `disabled` with the reason in the tooltip (`This Agent lives on prod-1; manage it there`) rather than answering 404 after the click.

Not in scope: making the settings page itself work against a machine. That is the larger gap — an Agent's state (AGENTS.md, skills, memory, vault, schedules) is a separate copy per machine — which needs a design decision, not a routing change.

## Verification

`pnpm typecheck`, `pnpm lint`, `pnpm format:check` clean; `pnpm --filter @prismshadow/penguin-web test` — 143 files / 1831 tests pass.

Sits at the end of the chain, on #641.


Myriad-Dreamin:

PR #655, opened 2026-09-09
feat(web): the machines page goes live

Myriad-Dreamin:
> Supersedes #577, which GitHub cannot reopen after its branch was force-pushed onto the rebased #450; same branch, same one commit, same body.

> **Stack:** `main` ← #354 ← #543 ← #599 ← #546 ← #383 ← #526 ← #551 ← #544 ← #448 ← #450 ← **this**. Review the diff against `feat/machines-run-agent`. Above it sit #614 and the closed branches of the terminal / chat / plugin / messaging lines (#613 … #641), restacked on 2026-09-09 and waiting to be reopened.
>
> **Draft on purpose — this is the switch.** Merge it when the feature has settled, not with the stack.

Three lines, and the feature the stack before it built becomes reachable:

- `machines` leaves `UNRELEASED_NAV_KEYS` in `nav-group-collapse.ts`, so the sidebar shows the row to admins — the gate `main` already held for exactly this.
- `/machines` gets a route, so the Machines page can be opened.
- The sidebar's **new workspace** picker offers the machine row, so a workspace can be created on a machine rather than only here.

Everything either one shows was already built and tested by the PRs below it; what was missing was the way in. Today on `main` the page component exists with no route and its nav row is gated off, so it is unreachable — this opens that entry, and nothing else.

Kept as a change of its own so that turning the feature on is a decision with a date, separate from the code that makes it possible — and so that turning it back off is a revert of a few lines rather than of a feature.

Both surfaces are admin-only in effect: the list they read answers `403` to everyone else, and a non-admin who types the URL gets the page's own refusal.

## Verification

Web `typecheck` and `format:check` clean; web suite **119 files / 1673 tests** passing.


Myriad-Dreamin:
Closing to retry reopening #577; if #577 reopens, this one is redundant.

Myriad-Dreamin:
Reopened: #577 is refused with GitHub's own words — "state cannot be changed. There are no new commits on the feat/machines-enable branch." — so the gate continues here.

Myriad-Dreamin:
Closed for good: #577 is restored on its own number.

PR #657, opened 2026-09-10
test(web): stacked sticky rows across a long expanded thinking body

Myriad-Dreamin:
## Context

A user report (screenshot, relayed second-hand) shows a finished work group's `DONE` header and its `Thinking` row painted in the middle of that thinking's own prose, with text both above and below the two bars, while a later `RUNNING 2 steps` group sits further down. The reasoning text itself is intact and in order, so the defect is in the Web App's rendering, not in the event stream.

The suspected mechanism is the stacked sticky pair (`sticky -top-4 z-[5]` on the group header, `sticky top-4 z-[4]` on the thinking row) over a body that, unlike a tool output's `max-h` box, has no internal scroll.

## What this PR does

- Adds a mock LLM scenario, `long thinking test`: a 120-paragraph thinking block followed by an `exec_command`.
- Adds `sticky-thinking.spec.mjs`, which expands the thinking row and walks the scrollport through the body in 150px steps, asserting at every position that the header is flush with the scrollport top, the row is stacked right below it, and the element painted just above the row is the header, never prose. Covered: a settled group; a still-running group (parked approval, ticking header) in dark mode; phone width.

## Status

**The reported misplacement does not reproduce on main** under any of the covered conditions, and none of the sticky-related files changed since v0.2.9. This PR is a draft carrying the regression coverage; the fix waits for the original screenshot (which scroll container the bars are pinned against, or whether the body grew after the bars were placed). No changelog entry: test-only.

## Verification

- `packages/web`: `bash e2e/run.sh` equivalent with `SKIP_BUILD=1` for `-g "thinking body|phone width"`: 2 passed.
- `prettier --check` on the two files: clean. `oxlint`: clean.

### DIFF
--- a/packages/web/e2e/mock-llm.mjs
+++ b/packages/web/e2e/mock-llm.mjs
@@ -20,6 +20,8 @@
  *    every 200ms for ~8s (reload-midstream.spec reloads while its output streams)
  *  - "slow text test" -> a long text streamed one delta every 200ms for ~8s
  *    (reload-midstream.spec reloads while the TEXT streams)
+ *  - "long thinking test" -> a thinking block several viewports tall + tool_use(exec_command)
+ *    (sticky-thinking.spec scrolls through the expanded thinking body)
  *  - last message has tool_result -> final text (turn 2)
  *  - otherwise (first user turn) -> thinking + text + tool_use(exec_command)
  */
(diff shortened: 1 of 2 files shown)


Myriad-Dreamin:
Superseded by #667: the misplacement is the engine segmenting a chunk that carries both the reasoning tail and the answer head, not a sticky-row issue in the Web App. The e2e coverage here did not reproduce the report.

PR #664, opened 2026-09-10
feat(web,server): a conversation opens on its newest Trace file and loads earlier ones per click

hiyouga:
## Summary

Opening a conversation loaded every Trace file of the Session at once. The Web App requested a tail window of 200 Task units, and since almost no Session has 200 Tasks, that window covered the whole transcript: every shard was read on the server, shipped, reduced and rendered. On long agentic Sessions that is the whole performance problem the user reported.

The conversation page now loads **Trace files, not the whole transcript**. One Trace file is one model context (files rotate at compaction, never by size), so the newest file is exactly what the model itself still sees. On open the page takes the newest file and, while the newest files together hold fewer than **50 turns**, the server keeps adding the file before them in the same response (`minTurns=50`), so a Session whose contexts each ended after a few turns does not open on a near-empty page. Older files then load one per click of a **load-earlier button** at the top of the stream; nothing loads on scroll, because a file can be a whole context's worth of tool output and whether to pay for it is the reader's call.

## What changed

**Server**
- `GET /api/sessions/:id/messages` gains `unit=task|file` for the existing `tailLimit` / `before`+`limit` window forms. `file` windows are whole shards: a tail of `n` files is the newest `n` shards from their own `session_meta` header; a `before` page is the `n` shards preceding the cursor's shard (plus that shard's head when a Task-unit cursor sits mid-shard). File cursors are `<fileIndex>:0`; `before` without `limit` defaults to one file.
- `minTurns=<n>` (file unit only) extends a window backwards by whole files until it holds at least `n` outline turns or reaches the beginning. The turns inside a window are the cumulative count at its end minus the cached count before its start, so growing the window reads nothing extra; only the end shard is scanned, and only when the window takes records from it.
- The `page` envelope gains `earlierFiles` (Trace files that begin before the window), the count the button shows.
- A file boundary can fall inside a Task (a compaction mid-run rotates the file). The priors at such a cut fold the still-open Task in (`totalsWithOpenTask`), so the header chips add up to the same totals as a full load. The parameterless full read and the Task-unit windows are unchanged; the page-stats cache format is untouched.

**Web**
- The controller opens on `{ tail, limit: 1, unit: "file", minTurns: 50 }` (`TAIL_MIN_TURNS`) and loads earlier files with `{ before, limit: 1, unit: "file" }`; `OlderHistoryState` carries `earlierFiles`.
- The reducer opens a round at a window's first `[context_summary]` or `request_begin` when no Task has ever opened in that model, so a file that begins mid-Task renders the continuation as its own round with a stats footer. A cut Task therefore shows as one round per file with the compaction banner between them. Full transcripts are unaffected: the guard only fires before any Task has opened.
- Resync with earlier files loaded: the refetched newest file is spliced when its start cursor is unchanged (true until a rotation); if the file rotated, the newest file alone is adopted and the loaded earlier files are dropped, reloadable by a click. The legacy full read now only serves a server without windowing.
- The message stream replaces the scroll-near-top trigger with a `Button` (`加载更早的对话（还有 N 段）` / `Load earlier messages (N segments left)`); spinner, retry and beginning-of-conversation rows stay.

**CLI** is unchanged: `logs`, `--resume` and `input` keep the parameterless full read.

**Docs and specs**: the server API and Web App pages describe the file-by-file loading; the specs change is in the design repo PR (see below).

## Design

- Why whole files rather than smaller Task windows: the user's mental model and the on-disk unit coincide (one file = one context), the tail page is what the model sees, and the cut is at a compaction, which the transcript already marks with a banner. Task windows of any fixed size still read every shard for ordinary Sessions.
- Why the cut is the file boundary even mid-Task: gluing the spanning Task to the page would pull in the previous file's tail (and for a Task that compacted several times, several files), defeating the goal exactly on the long-running Tasks where it matters. The continuation-round rule in the reducer plus the server-side fold keep the numbers honest instead.
- Why click and not scroll: a file is bounded by the compaction threshold, not by a small page size; loading it should be deliberate.
- Why the 50-turn floor lives on the server: one response instead of a cascade of load-earlier requests, and the window policy stays in the one place that already knows the per-file turn counts (the cached prefix states).

Design repo PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/136

## Verification

Vitest ran on the remote test host (this machine only compiles):

| Suite (vps3) | Result |
| --- | --- |
| `@prismshadow/penguin-server` test | 123 files, 1784 passed |
| `@prismshadow/penguin-web` test | 122 files, 1771 passed |
| `@prismshadow/penguin-docs` test | 7 files, 53 passed |

New/updated tests: `server/test/messages-page.test.ts` (file-unit windows: tail/walk-back chain, `limit: 2`, a Task cursor mid-file, the mid-Task rotation fold, shard-read discipline), `server/test/messages-live.test.ts` (`unit` param validation), `web/test/stream-controller.test.ts` (`TAIL_PAGE`/`OLDER_FILES`, `earlierFiles`, the rotated-file resync branch), `web/test/stream-model.test.ts` (the continuation round: summary-opened, discard-mode, and the full-transcript guard).

Locally: `pnpm -r build`, `pnpm --filter @prismshadow/penguin-server typecheck`, `pnpm --filter @prismshadow/penguin-web typecheck`, `pnpm format:check`, `git diff --check`.

https://claude.ai/code/session_016ggUKAni35vEFcZPjFj8LC


hiyouga:
Closing: the per-file window is more machinery than the problem needs. #202 already paginates history; only its window sizes (200 / 100 units) and the scroll trigger made every open read the whole transcript. A smaller PR adjusts those constants, replaces the scroll trigger with a click, and keeps resync from falling back to the full read.

PR #673, opened 2026-09-10
feat(sandbox): the confinement agent commands run under is configured on a Settings page

Myriad-Dreamin:
> **Stack:** `main` ← #599 ← #546 ← #656 ← #383 ← #674 ← #672 ← **this** ← #551 ← #718 ← #719 ← #544 ← #448 ← #450 ← #577 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #637 ← #628 ← #642 ← #639 ← #645 ← #641 ← #526 — one linear chain, in this order since 2026-09-14. Review the diff against `feat/plugin-npm-install`.

Split out of #383 on 2026-09-10: this page drives the `Sandbox` service that predates the plugin platform and shares nothing with the marketplace, the installed list or plugin configuration, so it is reviewed as the feature it is.

## What

- `GET|PUT /api/admin/sandbox` (admin, 403 otherwise): the confinement mode (`danger-full-access` / `workspace-write` / `read-only`), whether the network is cut off, and the paths masked from confined commands (trimmed, de-duplicated, at most 64). A rejected body changes nothing; a saved one applies to the next command spawn, and the settings park with the platform so they survive a hot update. The response also lists the mounted backends and the isolation dimensions each implements, awaited past the backends' asynchronous load so a page opened during boot does not read as "none installed".
- The System settings dialog gains a **Sandbox** page in the server group, after Upload limits. When no backend is mounted it says so plainly: a mode chosen without a backend confines nothing, and a security control that implies otherwise is worse than an absent one.

## Verification

At this branch's tip (Node 24): build of the server and web closures, `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, `test/admin-sandbox.test.ts` (4 tests) and the web `settings-sections` suite — clean.


Myriad-Dreamin:
Merged into #674 (2026-09-14): the sandbox settings are now a settings group on the Settings dialog's Plugins page — the policy, the backend notice, and each mounted backend's own options drawn inside the same card — instead of a route and a page of their own. Nothing from this branch is lost; its commit was folded into #674 and reworked there.

PR #675, opened 2026-09-11
fix(web): mint client ids where crypto.randomUUID does not exist

paskaa:
## What this changes

`crypto.randomUUID` is defined only in a **secure context** — HTTPS, `localhost`, `file://`. A harness reached over plain HTTP (the usual `http://<host>:7364` on a LAN) has `crypto` but no `randomUUID`, so every call site died with:

```
Uncaught TypeError: crypto.randomUUID is not a function
    at rne (index-Bm9n7ZV-.js:439:4358)
    at Gu  (index-Bm9n7ZV-.js:439:4642)
```

Two of them sit behind the first click a user makes — the id for a new conversation's parked draft (`features/chat/draft-sessions.ts`, `rne`) and the id for a saved shortcut (`features/chat/user-shortcuts.ts`). The click threw before any request left the browser, so **新建对话 did nothing** on that deployment; nothing in the server logs, because nothing was sent.

This adds `randomUuid()` (`packages/web/src/lib/random-uuid.ts`): a v4 UUID assembled from `crypto.getRandomValues`, which carries no secure-context restriction, falling back to `Math.random` only where Web Crypto is absent altogether. Both callers now take their id from it and slice it exactly as before, so the `draft-<8 hex>` and `sc-<12 hex>` shapes are unchanged. The ids are opaque handles the server stores and hands back, never secrets.

## Tests

`packages/web/test/random-uuid.test.ts` (3 cases):

1. 100 calls return 100 distinct v4 UUIDs.
2. `Crypto.prototype.randomUUID` removed → still returns a v4 UUID (the shape of the reported deployment).
3. No Web Crypto on the global at all → still returns a v4 UUID.

Case 2 is the regression guard: reverting the helper to `crypto.randomUUID()` fails it.

## Verification

- `pnpm --filter @prismshadow/penguin-web test` — 130 files, 1819 tests passed.
- `pnpm format:check` — clean. `pnpm lint` (oxlint) — 0 warnings, 0 errors.
- Mutation check — with `randomUuid()` replaced by a bare `crypto.randomUUID()`, `test/random-uuid.test.ts` fails 2 of 3.
- Reproduced and re-checked in the real app: Chrome (Playwright) on `http://<lan-ip>:7364`, where `isSecureContext === false`, `typeof crypto.randomUUID === "undefined"`, and the two expressions from the bundle above threw. With a UUID polyfill injected into that deployment's `index.html`, the same two expressions return `draft-…` / `sc-…`, which is what this change moves into the source.

## Checklist

- [x] Changelog entry pair under `changelog/unreleased/` (`2026-09-11-web-client-id-insecure-context.md` + `.zh.md`).
- [x] Docs — no documented behaviour changes; the ids keep their prefixes and lengths.
- [x] No credential in the diff, the logs or the screenshots.

### DIFF
--- a/changelog/unreleased/2026-09-11-web-client-id-insecure-context.md
+++ b/changelog/unreleased/2026-09-11-web-client-id-insecure-context.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-09-11
 - **Type:** fix
 - **Scope:** `web`
+- **PR:** [#675](https://github.com/Prism-Shadow/penguin-harness/pull/675)
 
 [中文版](2026-09-11-web-client-id-insecure-context.zh.md)
 
--- a/changelog/unreleased/2026-09-11-web-client-id-insecure-context.md
+++ b/changelog/unreleased/2026-09-11-web-client-id-insecure-context.md
@@ -1,0 +1,15 @@
+# Client ids minted where `crypto.randomUUID` does not exist
+
+- **Date:** 2026-09-11
+- **Type:** fix
+- **Scope:** `web`
+
+[中文版](2026-09-11-web-client-id-insecure-context.zh.md)
+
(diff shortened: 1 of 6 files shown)


hiyouga:
sorry I did not see this PR, the same problem has been resolved in #687

PR #676, opened 2026-09-11
fix(web): the Trace overview's Elapsed shows the total, breakdown on hover

hiyouga:
The Trace file view's overall summary printed the elapsed total and its API / tool breakdown on the
same line, so a row meant to be read at a glance carried three durations. The per-round chip beside
it already had the right shape — total visible, breakdown in `title` — and this makes the global row
match it.

`SummaryRow` gains an optional `detail`. When it is non-empty the row carries `title` and
`aria-label`: the hover text is `${S.chat.statElapsed}${durationSplit(...)}`, exactly the string the
per-round chip hovers, so both surfaces word it the same way. The accessible name keeps the value
beside the breakdown, because `aria-label` replaces the row's own text rather than adding to it —
without that, a screen reader that honours the name would lose the total. An empty `detail` (neither
component measured anything) leaves both attributes off rather than announcing a name that only
repeats the visible label.

`durationSplit` is reused verbatim and the text is unchanged: API time and tool time are two
measurements of the same span, not a partition of it, and nothing here restructures them into
anything that reads as a sum. No new strings — `S.chat.statElapsed`, `statElapsedSplit` and
`statParenOpen/Close` all already exist in both dictionaries. The timeline chart, `trace-panel.tsx`,
`trace-event-row.tsx` and the chat header's twin are untouched.

## Verification

Node v24.18.0, from the worktree.

- `pnpm --config.verify-deps-before-run=false -r build` — exit 0 (needed once so the web typecheck
  can resolve `@prismshadow/penguin-core`).
- `pnpm --filter @prismshadow/penguin-web typecheck` — exit 0, no diagnostics.
- `pnpm format` then `pnpm format:check` — "All matched files use Prettier code style!", exit 0. The
  changelog files are `*.md`, which `.prettierignore` excludes, so they are wrapped by hand.
- `PENGUIN_TEST_HOST=vps3-test .agents/scripts/remote-test.sh <worktree> --build "@prismshadow/penguin-core" -- --filter @prismshadow/penguin-web test`
  — on **vps3-test**: `Test Files 129 passed (129)`, `Tests 1816 passed (1816)`, exit 0. Includes
  `test/trace-refresh.test.ts (17 tests)`, which pattern-matches the source of `trace-file-view.tsx`.
  The first attempt without `--build` failed 31 suites on unresolved `@prismshadow/penguin-core/*`
  imports — a fresh remote tree has no `dist`; building core first fixes it and is unrelated to this
  change.


hiyouga:
## Review

This moves the Trace overview's API / tool breakdown off the Elapsed line into the row's hover text,
matching the per-round chip beside it. The number itself is untouched and the label is honest:
`analysis.elapsedMs` is the server's **sum of per-turn spans**, not a first-to-last wall clock
(`packages/server/src/services/trace-service.ts:1403`, contract at
`packages/server/src/api/types.ts:2679-2685`), which is exactly what the per-round chips below add
up to — and every degenerate case (no turns, a turn whose `startTs` never got a `request_begin`, a
file still being appended to) lands on `0` and drops the tooltip through `durationSplit`'s empty
string. Ship after the accessibility fix below, which is pushed to this branch.

### Findings

**1. `aria-label` on a role-less `<div>` is not announced, so the breakdown reached mouse users
only** — `packages/web/src/features/traces/trace-file-view.tsx:123` (pre-fix) · *should fix*

A `<div>` carrying no role maps to the ARIA `generic` role, whose *Name from* is **prohibited** — an
`aria-label` there is dropped rather than read. So a screen-reader user on the Trace overview got
"Elapsed 1m20s" and no breakdown at all, which is the one reader the attribute was added for. The
premise it was built on ("`aria-label` replaces the row's own text rather than adding to it", the
reason the value was repeated into the label) does not hold for a generic container in either
direction. The surrounding code already applies this rule: every non-interactive `aria-label` in
these features carries a role that permits naming — `features/traces/timeline-chart.tsx:722`
(`role="scrollbar"`), `features/chat/context-gauge.tsx:259` (`role="img"`), `:286` and `:530`
(`role="group"`), `:693` (`role="slider"`).

Fixed by carrying `detail` as `sr-only` text right after the value instead — the pattern already
used at `features/agents/agent-settings-page.tsx:547` and `features/models/models-page.tsx:1051` to
fold a hint into what is read out. It reads once, in order ("Elapsed", the total, the breakdown),
needs no role juggling, and the repeated value goes away. `title` and the visible row are unchanged.
✅ Fixed in `5213fd51`

**2. The breakdown is still unreachable without a mouse** — `trace-file-view.tsx:121-133` · *nit* ·
⬜ Left for the author

Stated plainly rather than fixed: a `title` on a non-focusable `<div>` never appears for a keyboard
user (there is no focus for it to hang on, and browsers do not surface `title` on focus) and never
on touch. After the fix a screen reader gets the breakdown; a sighted keyboard-only or phone user
does not. That is the same deal the per-round `StatChip` (`:146`) has always offered, so this is a
consistent product position rather than a regression introduced here — flagging it because the
changelog presented the hover text as the fallback path. Making it reachable would be a UX decision
(an inline breakdown below `sm`, or a focusable row), not a review fix.

**3. The total has no hour form** — `packages/web/src/lib/format.ts:32` · *nit* · ⬜ Left

`humanizeDuration` tops out at `m`+`s`, so a long file's Elapsed reads `183m12s`. That magnitude is
now the row's only number instead of the first of three, so it is a little more exposed than before
— but both the value and the formatter are byte-identical to `main`, so widening the diff for it did
not seem right for this PR.

### Left for you

- **The paired design PR is still open and does not name this surface.**
  `Prism-Shadow/penguin-harness-design#112` writes the Trace page's rule as
  「全局用时同样带这两个分项，与逐轮一样是各轮之和」 while explicitly saying 悬停 for the chat
  header's chip in the same diff. #676 makes the Trace global row hover-only too, so one clause
  there would keep the still-unmerged spec in step with what ships. I did not touch it — specs live
  in the other repository.
- **Concurrent edit in #681 (`feat/trace-avg-tool-calls`).** It also edits this file and
  `lib/format.ts`. Its hunks are the import list, the comment at ~`:453`, and a new `SummaryRow` in
  the counts column; mine is `SummaryRow`'s own definition (`:106-133`). Disjoint, and its new row
  simply passes no `detail`, so no conflict is expected in either direction.
- **No test, deliberately.** `packages/web/vitest.config.ts` is `environment: "node"` with
  `include: ["test/**/*.test.ts"]` — the suite renders no React and cannot load `.tsx`, so the
  `title` / hidden-text composition is not assertable there. The only pure function involved,
  `durationSplit` (`:193`), is byte-identical to the base commit, so a unit test of it would pass at
  `$BASE` and prove nothing about this change; `test/trace-refresh.test.ts` already pins this file
  by source text and stays green. Playwright would reach the real DOM, but `packages/web/e2e` is not
  runnable on the build box.

### Verification

```
Node v24.18.0, from the worktree; vitest ran on vps3-test via vps3-vitest.sh (never locally).

pnpm --filter penguin-core --filter penguin-server build          exit 0 (fresh worktree has no dist)
pnpm --filter @prismshadow/penguin-web typecheck                  exit 0, no diagnostics
pnpm lint (oxlint)                                                exit 0
pnpm format && pnpm format:check                                  "All matched files use Prettier code style!"
git diff --check                                                  clean

vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web exec vitest run \
    test/trace-refresh.test.ts test/trace-compaction-badge.test.ts test/format.test.ts
                                                                  3 files, 62 tests passed
vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web test   129 files, 1816 tests passed

Changelog pair re-checked against changelog/README.md: H1, metadata order Date/Type/Scope/PR, PR as
a full URL to this PR, counterpart links both ways, no ## Why / ## Verification sections, zh mirrors
the en paragraph for paragraph. Both files reworded in the same commit to match the fix.
```

PR #677, opened 2026-09-11
fix(web): the session row's scheduled-task mark stops flickering and recedes

hiyouga:
The alarm clock a session row wears while a scheduled task is bound to it blinked in and out
while switching between conversations, and it was drawn in the same amber as the running
hourglass and the pending-approval badge. Both come from the same place: the mark was treated as
live work, and the list behind it lived in one global slot that every scope change emptied.

## Why it blinked

`schedule-store.ts` held one `items` list, one `error` and one `inflight` promise for whichever
scope the store last pointed at. Three failures compounded:

- **A scope change blanked the data.** Clicking a row flips `currentAgentId` synchronously, so
  the sidebar re-rendered for the new Agent before its effect ran; the scope check failed, the
  list read as `null`, and `pendingScheduleSessions(agentSchedules ?? [])` turned "not loaded"
  into "nobody has a task". Every alarm in the list went out for at least a frame, and came back
  only after a round trip.
- **A refetch could be dropped.** `refreshSchedules` began with `if (inflight) return inflight`,
  and that promise belonged to the *previous* scope. Switch agents mid-request and the new scope
  was never fetched: the old promise resolved, discarded itself on the scope check, and nothing
  else asked. Those rows stayed bare until a window focus, a `schedule_fired` / `schedule_queued`
  event, or the dock's schedules tab.
- **Two consumers fought over one scope.** The sidebar scopes to the current Agent, the dock's
  panel to the open conversation's. Between a navigation and the chat page's sync effect they
  disagree, and each mismatching `scopeSchedules` nulled the list and notified — so the marks
  blinked even without leaving an Agent.

## What changed

The store keys everything by scope. One `Map` entry per project + agent holds that agent's list,
its error, its in-flight request and a count of its mounted readers; `refreshSchedules` takes the
scope it is refreshing instead of reading a module global, which is what makes the second and
third failures impossible rather than merely unlikely. A scope is answered from its own entry, so
returning to an Agent already loaded shows its marks on the first frame, and a request asked for
on behalf of agent B goes out for B even while A's is still in flight.

Entries are not evicted, and the header says why: growth is bounded by the number of Agents one
sitting opens, and an entry is a list already fetched once.

`retainSchedules` ref-counts the mounted readers, so a regained focus and the schedule events
refresh the scopes actually on screen and no others — the behaviour the old `schedulesScopedTo`
guard approximated with a global. `scopeSchedules` and `schedulesScopedTo` are gone: nothing
outside the module used them, and neither idea survives per-scope entries.

The sidebar stops reading a null list as an empty one: while an Agent has never been answered it
holds the set of marks already on screen. The dock's panel names its conversation's Agent on its
poll and after every create, edit, toggle and delete.

## The tone

`ScheduleMark` moves from `toneInk.attention` to `toneInk.muted`, and in the row it moves from
after the background-tasks mark to beside the pin and the messaging-relay glyph, ahead of
`StatusGlyph`. A standing arrangement is settled, not live work; `muted` is the tone whose
meaning is "the mark should recede", and it is allowed below 3:1 exactly because its meaning is
already carried in text — here the `aria-label` and tooltip (`S.chat.sessionScheduled`).

`toneInk.muted` is `text-gray-400 dark:text-gray-500`, the pair the pin and the relay glyph
already spelled inline, so those two now read the token instead. The cluster is one thing in the
code as well as to the eye, and the row reads: pin, relay, alarm → live status, background tasks,
pending approvals.

## Verification

Run on Node v24.18.0, after building `@prismshadow/penguin-core` so the web package's
`@prismshadow/penguin-core/*` subpath types resolve.

```
$ pnpm --config.verify-deps-before-run=false --filter @prismshadow/penguin-core build
DTS ⚡️ Build success in 4844ms

(no output — clean)

Checking formatting...

$ pnpm lint
$ oxlint
(no findings)
```

Vitest ran on the remote host `vps3-test`, at the user's standing request for this session:

```
$ PENGUIN_TEST_HOST=vps3-test .agents/scripts/remote-test.sh . \
 ✓ test/schedule-store.test.ts (4 tests) 3ms
 ✓ test/session-activity.test.ts (21 tests) 14ms

 Test Files  130 passed (130)
      Tests  1821 passed (1821)
   Duration  2.49s
```


hiyouga:
## Review

The race is real and the fix is structural, not a debounce: one `items` / `error` / `inflight` slot
for whichever scope the store last pointed at meant that `if (inflight) return inflight` handed the
new scope the *previous* scope's promise, which then discarded itself on `scope !== target` — so
after an agent switch mid-request nothing asked again and the rows stayed bare until a focus, an
event or the dock tab. Per-scope entries remove that, and the mid-flight test proves it. Ship after
one decision on the tone half, which reaches further than the title says.

### Findings

**1. `BackgroundTasksMark` moved to `muted`; its hand-drawn twin in the chat header did not** — `packages/web/src/components/ui/session-activity-icon.tsx:91` vs `packages/web/src/features/chat/chat-page.tsx:1902` · *should fix — your call*

A conversation with two background tasks: its sidebar row now draws the trace in gray-400, and the
chat header above it draws the same glyph with the same count in emerald-600. `chat-page.tsx:1895-1898`
says in so many words that it is "in the live-status green, the same figure and glyph as the session
row's mark" — that is now false, and nothing in the diff updates it.

The reclassification itself is worth a second look. `lib/tone.ts:12-16` defines `busy` as "executing
right now" and `muted` as "settled"; the number this mark carries is `sessionBackgroundTasks` =
`processes + subagents` (`lib/session-activity.ts:51-54`), which the header's own comment describes as
"command processes past their yield window, background subagents **mid-round**" — work that is
executing. `muted` also puts the mark at gray-400/gray-500, recorded at 2.60–2.49 : 1 in light
(`tone.ts:49`), under WCAG 1.4.11's 3:1 and permitted only where the meaning is already in text. That
part holds — the label is required — but "settled" does not describe a running subagent.

It is also a UX change to two surfaces the title does not cover (the session row and a
`run_in_background` tool row, `tool-call-card.tsx:382`), so I have not touched the code. Either bring
`chat-page.tsx:1902` to `muted` as well, or revert `BackgroundTasksMark` to `busy` and keep the PR to
the scheduled-task mark. ⬜ Left for the author — the changelog bullet I added has to follow whichever
way you go.

**2. `retainSchedules` / `noteScheduleEvent` shipped with no test** — `packages/web/src/features/schedules/schedule-store.ts:140,153` · *should fix*

The reader count decides whether a `schedule_fired` / `schedule_queued` event costs a request at all.
Get it wrong in the permissive direction and every event in every cached agent refetches; get it wrong
in the strict direction and a task fires while its row is on screen and the row never updates. Neither
direction had cover. ✅ Fixed in 3d3ca43f: three tests — an event dropped for a cached scope nobody
reads, taken for a retained one, and the retain being a *count* rather than a flag (both surfaces
normally hold the same agent, so either one releasing must not silence the other). Each was run against
a deliberately mutated store — the `readers === 0` gate removed, then the count replaced by a
set-to-1/set-to-0 flag — and fails there; my first cut of the count test did *not* catch the flag
mutation, so it is rewritten to read the call count before settling the request.

**3. The module header still documented the tone the mark no longer wears** — `packages/web/src/components/ui/session-activity-icon.tsx:31-32` · *should fix*

It read "an activity trace in the `busy` tone" four lines of code above `toneInk.muted`. ✅ Fixed in
3d3ca43f by removing the tone from the header entirely rather than restating it: each mark names its
own tone at its own definition, so there is one place to change and none to go stale. That also makes
the header survive either resolution of finding 1.

**4. The changelog pair recorded three of the four marks that changed colour** — `changelog/unreleased/2026-09-10-session-row-schedule-mark.md` · *should fix*

"The mark moved next to the pin and the messaging-relay glyph … and all three take `muted`" — but
`BackgroundTasksMark` changed too, visibly, on a session row and on a tool row. ✅ Fixed in 3d3ca43f,
both languages, noting that the chat header's own count is drawn separately and kept its tone.

**5. `lastScheduledRef` argues against the comment ten lines above it** — `packages/web/src/components/layout/sidebar.tsx:391-395` vs `:369-371` · *nit*

Workspace grouping, current Agent A loaded, a row of A wearing the alarm. Click a row of Agent B for
the first time this tab: A's rows keep their marks from the ref, then lose them the moment B's list
lands, because B's list holds no session of A. The data was never wrong, but a mark that stood for a
round trip and then vanished is a different shape of blink from the base's, which removed it at once —
and `:369-371` still says such a row "answered from another Agent's list would not be [honest]", which
is the position the removed `scopeSchedules` doc took too ("worse than no mark at all"). Worth settling
one way in the comments. ⬜ Left for the author.

**6. A mutation's refresh can be answered by a request issued before the write** — `packages/web/src/features/schedules/schedule-store.ts:114`, `features/schedules/schedule-panel.tsx:196` · *nit, pre-existing*

The panel's 30 s poll fires at t=0. The user flips a task's switch at t=+40 ms; `updateSchedule`
resolves at t=+140 ms; `await refreshSchedules(projectId, session.agentId)` finds `entry.inflight` set
and returns the **poll's** promise, which answers with the pre-write list. The toast says "disabled",
the row's switch springs back, and it stays wrong until the next poll. Same shape at the base commit
(one global `inflight`, cleared in `finally`), so this is not a regression — but this PR rewrote the
function, which makes it the moment to decide whether a post-mutation read should be allowed to join an
older request. ⬜ Left for the author.

### Test quality

Replayed against `git show $(git merge-base origin/main HEAD):…` — the base API takes no scope, so the
equivalent is `scopeSchedules(…)` + `refreshSchedules()`:

- *keeps a list per agent* — base held one `items` slot, so a2's answer replaced a1's. **Fails at base.**
- *issues a request for the scope it was asked for, mid-flight* — base returned a1's `inflight` to a2's
  caller, so `listSchedules` was called once, with a1 only. **Fails at base**, and this is the actual
  defect rather than the cosmetic one.
- *confines a failure to the scope that failed* — base had one global `error`. **Fails at base.**
- *shares one request per scope while it is out, and refetches once it has settled* — the global
  `inflight` already deduped and already cleared in `finally`. **Already passed at base**; a guard, not
  a regression test. Worth keeping, worth not counting as evidence.

Still uncovered: `useAgentSchedules` and the sidebar's ref fallback — the first-frame-after-navigation
behaviour the PR is named for. `packages/web/vitest.config.ts` is `environment: "node"` and the package
carries no jsdom or testing-library, so that would cost a dependency; I have not added one, and the
store-level mid-flight test is the closest honest substitute.

### Overlap with #689

No collision. #689's only `sidebar.tsx` hunks are at 2275–2340 (`size="sm"` on the rename dialogs'
`Button`/`Input`); this PR touches 173, 302, 364–395 and 2716–2760. A textual merge either way.

### Verification

```
Node v24.18.0, in ../penguin-harness-wt/rv-677 at 3d3ca43f.

$ pnpm --filter @prismshadow/penguin-core build                 # dts, for the web subpath types
$ pnpm --filter @prismshadow/penguin-web typecheck              # clean
$ pnpm lint                                                     # oxlint, no findings
$ pnpm format && pnpm format:check                              # all files use Prettier style
$ git diff --check                                              # clean

vitest on the build box (vps3-test), per this session's standing rule:

$ vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web exec vitest run \
      test/schedule-store.test.ts test/session-activity.test.ts
  ✓ test/schedule-store.test.ts (7 tests)   ✓ test/session-activity.test.ts (21 tests)

$ vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web test
  Test Files  130 passed (130)      Tests  1824 passed (1824)

Mutation checks on the same host, reverted after each:
  noteScheduleEvent's `readers === 0` gate removed  -> both new gating tests FAIL
  retainSchedules' count replaced by a 1/0 flag     -> "counts readers" FAILS
```

PR #679, opened 2026-09-11
fix(web,server): keep the catalog's display name when syncing presets

hiyouga:
## What was wrong

A model added to a Project by the Models page's **sync presets** button was listed under its raw
`model_id`, and no amount of re-syncing brought its name back. The chain:

1. `presetToRow` (`packages/web/src/features/models/catalog-sync.ts`) built the new row from
   `presetModelEntries()`, which emits the **persisted** `ModelEntry` shape — and `display_name` is
   only persisted when it differs from the built-in catalog, so a preset entry never carries one.
   The row's `displayName` was `undefined`.
2. `rowToEntry` (`models-page.tsx`) submitted the field only when non-blank, so the PUT entry had
   no `displayName` — and the page PUTs the whole table, not just the rows it changed.
3. The route dropped a falsy `displayName`, so the service could not tell an absent name from an
   empty one.
4. The write path read an absent name on a model the catalog *does* name as "the user deliberately
   cleared it" and wrote `display_name = ""`.
5. The read path mapped `""` back to no name, so the row fell back to its id — and the next sync
   repeated step 1.

## What changed

**The direct cause.** `presetToRow` looks the name up with `catalogEntryFor(provider, modelId)` and
carries it on the row, so a synced entry reaches the PUT named and the merged rows the page
renders say what they submit.

**The root cause.** Absent and empty are now different requests, end to end:

- `routes/models.ts` accepts an explicit empty string instead of dropping falsy values.
- The service writes `display_name = ""` only when the client explicitly sent `""`; an absent field
  means "inherit the catalog's name".
- `GET /projects/:id/models` reports a cleared name as `""` rather than as no name. Without this
  half, the next save of any kind — set default, delete a row, reorder a group — would re-send the
  table with no name for a model whose name the user had deliberately cleared, and the new "absent
  means inherit" rule would hand the name back. This is the one piece not in the original brief;
  the two consumers that rendered `displayName ?? modelId` now treat blank as absent.
- `rowToEntry` submits the field whenever the row has one at all, the empty string included, so
  clearing the name in the dialog still clears it.

Either half would stop the symptom on its own — a named entry is never read as a deletion, and an
absent name no longer destroys anything. Both are here because a client should say what it means
and a server should not read silence as an instruction.

**Repairing Projects already damaged.** "Sync presets" fills a blank display name from the catalog
and never overwrites one that is there: unlike pricing or the context window, a stored name may be
the user's own rename. `catalogDelta` — which drives the Models nav badge — counts that repair on
the same rule, so the badge and the button cannot disagree about whether there is anything to sync.

**The trade-off, stated out loud:** a name the user deliberately cleared is filled in again the
next time they press "sync presets". Pressing it is what asks for the catalog's values, and the
alternative — a row stuck showing its model id with no way back — is worse.

**No reinterpretation of data on disk.** `display_name = ""` means "cleared" before and after this
change, and an ordinary save on the Models page still round-trips it. Existing config files are
untouched until the user syncs.

## Verification

All on a fresh build of the worktree (Node v24.18.0). Vitest ran on the remote host `vps3-test`
via `.agents/scripts/remote-test.sh`, per the standing instruction for this session.

```
pnpm --config.verify-deps-before-run=false -r build      # all packages built
pnpm --filter @prismshadow/penguin-web typecheck         # clean
pnpm --filter @prismshadow/penguin-server typecheck      # clean
pnpm format && pnpm format:check                         # All matched files use Prettier code style!
```

```
PENGUIN_TEST_HOST=vps3-test .agents/scripts/remote-test.sh /home/cc/dev/penguin-harness-wt/syncname \
# Test Files  129 passed (129)
# Tests  1818 passed (1818)

PENGUIN_TEST_HOST=vps3-test .agents/scripts/remote-test.sh /home/cc/dev/penguin-harness-wt/syncname \
  --build "@prismshadow/penguin-core" -- --filter @prismshadow/penguin-server test
# Test Files  125 passed (125)
# Tests  1800 passed (1800)
```

Tests added or changed:

- `packages/web/test/catalog-sync.test.ts` — a row `presetToRow` builds carries the catalog's name;
  a sync fills a blank one and leaves a user's own alone; `catalogDelta` agrees with the merge on
  all three, replayed through the existing badge-versus-button invariant.
- `packages/server/test/models.test.ts` — an absent name inherits the catalog's and writes nothing;
  only an explicit `""` records the clear, and the GET reports it as `""` so the round trip holds.


hiyouga:
## Review

The fix is correct and the "absent vs empty" split is the right cut: the route no longer collapses
`""` into absent, the service only records a clear on an explicit `""`, and the GET reports the
clear so the next whole-table PUT does not undo it. I traced every writer of `ModelUpdateEntry` and
every reader of `ModelInfo.displayName` — the web page's single `persist` funnel, `group-import`,
and `machines/models-sync.ts` on both halves — and all of them come out right; the machine sync in
particular stops clearing the name of every catalog model it pushes to a remote host. One
simplification and one test-fixture fragility fixed below; the cleared-name trade-off is a product
call I have left to you.

### Findings

**1. Two helpers, two sentinels, for one question** — `packages/web/src/features/models/catalog-sync.ts:38-59` · *should fix*

`catalogDisplayName` returned `""` for "no name" and `displayNameFill` immediately converted that
back to `undefined` (`catalogDisplayName(p) || undefined`), while `presetToRow` tested the other
sentinel (`...(displayName ? … : {})`). Folded into one helper returning `string | undefined`;
`presetToRow` now asks `displayNameFill(p, undefined)`, which is exactly the question a brand-new
row poses. No behaviour change — `ModelCatalogEntry.displayName` is a required field
(`packages/core/src/state/model-catalog.ts:139`) and no catalog row carries `""`, so the two forms
were already equivalent. ✅ Fixed in ef34804

**2. The doc comment and the test both describe a state that cannot exist** — `catalog-sync.ts:39`, `packages/web/test/catalog-sync.test.ts:112` · *should fix*

`catalogDisplayName`'s JSDoc said `""` meant "the catalog does not name it", and the test's comment
read "A preset entry the catalog does not name gets none". `displayName` is required on
`ModelCatalogEntry`, so there is no such entry. The real reason that assertion passes is different:
`qwen-token-plan/qwen3.8-max-preview` **left the catalog** on 2026-08-03
(`model-catalog.ts:1482`), so the fixture's pair resolves to nothing. Worth stating precisely,
because it is the only place the injected-`preset`-list seam is visible — `displayNameFill` looks
the name up in the global catalog, not in the list the caller passed. That assertion also already
held at `$BASE` (every row's `displayName` was `undefined` there); it is kept as the guard on
`presetToRow`'s "never attach an empty name" branch, with a correct comment and an explicit
`catalogEntryFor(...)` assertion so the reason is visible rather than incidental. ✅ Fixed in ef34804

**3. The test pinned catalog display names by hand** — `packages/web/test/catalog-sync.test.ts:54,111,134,158,162` · *should fix*

`inSyncRow` hard-coded `"DeepSeek V4 Pro"` and the assertions `"GLM-5.2"`. Every other field in
`PRESET` is a hand-made stand-in compared only against `PRESET` itself, so the display name was the
one value reaching into the shipped catalog — meaning a rename in `model-catalog.ts` would break
`catalog-sync.test.ts`, a file about the merge algorithm, and that file is not on the list of
places `reference/model-catalog.md` says a catalog change touches. Now read through a
`catalogName(provider, modelId)` helper. The assertions still fail at `$BASE` (`undefined` vs the
catalog's name), so nothing was weakened. ✅ Fixed in ef34804

**4. The internal contract for "sync presets" no longer matched** — `.agents/skills/penguin-harness-dev/reference/model-catalog.md:58` · *nit*

That paragraph is the sentence a future catalog author reads: sync "appends and updates
catalog-owned fields but never deletes and never touches the stored default". After this PR the
display name is a catalog-owned field that is *filled but not overwritten*, which the sentence
would have them get backwards. Added the carve-out. ✅ Fixed in ef34804

### Checked and clean, for the record

- **Whose name wins.** A hand rename survives a sync in both directions:
  `displayNameFill` returns `undefined` for any non-blank `current`, and `presetFields` — which the
  merge does spread over the row — carries no `displayName`. A catalog-side rename needs no sync at
  all: `getModels` inherits the catalog name at read time for any entry with no stored
  `display_name`.
- **Pair keying.** Every lookup this PR adds or changes is `(provider, model_id)`:
  `catalogEntryFor(p.provider, p.model_id)`, and both `Map`s key on `${provider}\0${modelId}`
  (`catalog-sync.ts:126,160`). No bare-id match anywhere.
- **`ModelInfo.displayName: ""` tolerated by every reader.** `modelLabel`, `modelLabelOf`,
  `toRow`, `model-grouping.ts:64`'s search filter and `models-sync.ts:96` all handle both `""` and
  absent, so an older server or an older stored row still renders as the model id.
- **No docs drift.** `configuration.{en,zh}.md:69` ("persisted only when it differs from the
  built-in catalog") is still exactly true, and `server-api.*.md` documents the models endpoints
  without field-level display-name semantics. No docs change is owed.
- **Test replay against `$BASE`.** Every new assertion except the `qwen3.8-max-preview` one (see
  finding 2) fails at the base commit: `glm.displayName`, the blank/absent/renamed trio, the
  `rows.every(...)` named-row sweep, the `catalogDelta` blank-name case, and all four new server
  assertions. The exact-order `refs` assertions in `catalogDelta` were updated, not loosened —
  `"names every entry it would add or rewrite, in catalog order"` still pins the full array.
- **Changelog pair** mirrors section for section, metadata order is right, `PR:` is a full URL,
  no `## Why` / `## Verification`.

### Left for you

**A name the user deliberately cleared is refilled by the next sync — and the Models nav badge
nags until they let it happen.** The PR states the first half in the entry. The second half is not
stated and is the part worth a decision: `catalogDelta` counts a blank name as "1 model to
upgrade", so a Project where someone cleared a catalog model's name on purpose now carries a
standing update badge until they dismiss it or press the button that reverts their choice.

There is no way to separate the two populations, and that is the crux: the bug wrote exactly
`display_name = ""` into `.project_config.toml`, which is byte for byte what a deliberate clear
writes. Nothing on disk distinguishes "the old sync broke this" from "I wanted the raw id here", so
"repair only the damage" is not implementable without introducing a third marker — more
compatibility surface than the problem is worth, in my reading. That leaves:

1. **Ship as is.** One rule, and the damage the bug already did is repaired by the same pass. Cost:
   clearing a catalog model's name is no longer durable, and the badge says so until they comply.
2. **Ship the refill, keep the badge quiet.** Fill the name when the button is pressed, but drop
   the `displayNameFill(...) !== undefined` term from `catalogDelta` (`catalog-sync.ts:140`) so
   a blank name never raises the dot on its own. The clear still ends when they sync for another
   reason, but nothing pesters them into it. It does break the badge/button agreement the module
   header and the final invariant test assert, so both would need revisiting.
3. **Accept it in the product too.** If a cleared name cannot survive a sync, the dialog arguably
   should stop offering it as a durable setting. Bigger change, out of scope here.

This is on-disk state, so per the repo's rule it is yours to call — I have implemented none of
them. If you keep option 1, the entry is worth one more sentence about the badge, since that is
what a user will actually notice.

### Verification

```
Local (worktree /home/cc2/zyw/penguin-harness-wt/rv-679, node v24.18.0):
  pnpm -r --filter penguin-core --filter penguin-server build   ok
  pnpm --filter @prismshadow/penguin-web typecheck              ok
  pnpm --filter @prismshadow/penguin-server typecheck           ok
  pnpm lint (oxlint)                                            ok, exit 0
  pnpm format && pnpm format:check                              All matched files use Prettier code style!
  git diff --check                                              clean

Vitest on the vps3-test build box (never on this machine), via vps3-vitest.sh:
  web  catalog-sync + models-input + model-grouping
     + group-import + protocol-detect + model-group-expansion   6 files, 128 tests passed
  server test/models.test.ts                                    26 tests passed
  core  test/model-catalog.test.ts                              37 tests passed
  pnpm --filter @prismshadow/penguin-web test                   129 files, 1818 tests passed
  pnpm --filter @prismshadow/penguin-server test                125 files, 1800 tests passed
```

PR #680, opened 2026-09-11
Redraw the dock launcher's shortcut icons, and make its ball answer the pointer

hiyouga:
The floating launcher's shortcuts were glyphs that named their panels only loosely — a spawn tree for subagents, an open book for Memory, three stacked lines for Trace, a chat bubble for remote control, and the dialog close cross for "hide". This redraws them to say what they open, and makes the ball itself report what the pointer is on.

## The ball answers the pointer

Three states, resolved in one place so they cannot contradict each other:

| condition | ball glyph | caption |
| --- | --- | --- |
| a fan entry is hovered or focused | that entry's glyph, at 18px | that entry's name |
| else the ball is hovered or focused | the expand mark | `S.dock.launcherOpen` — 打开 / "Open" |
| else | the workbench tiles | `S.dock.launcherCaption` — 快捷方式 / "Shortcuts" |

A fan merely standing open is deliberately **not** the hover state: with the fan up and nothing pointed at, the ball goes back to its resting mark and its own caption. Keyboard focus drives all of this exactly as the pointer does, so the existing arrow-key walk down the arc now also redraws the ball.

`FanEntry` now carries `glyphAt: (size: number) => ReactNode` rather than a pre-rendered node, because the same mark is drawn at two rungs — 15px in the fan, 18px in the ball — and the launcher tracks the hovered entry's **key** rather than its label, so one piece of state resolves both the name and the glyph. The ball's hover is tracked with plain `onMouseEnter`/`onMouseLeave`/`onFocus`/`onBlur`; `usePointerDrag` binds only `onPointerDown`, so nothing is clobbered.

`S.dock.launcherOpen` was added to both dictionaries.

## The marks

All new paths are on the 24×24 grid at `strokeWidth` 1.7 through `GlyphIcon`, with ~2px of margin, and were drawn against the **13px** dock tab strip rather than the 18px ball — that is the size that decides what survives.

`panel-meta.tsx` is one table feeding four surfaces (the launcher fan, the dock's panel picker rows, its "+" add menu and its tab strip), so each of these moved on all four at once.

- **Subagents** — two robot heads, a large one above-left and a small one below-right. The head is the Agent's own (`AGENT_GROUP_ICON`), reduced to antenna + rounded head + two eye dots; the ears and the smile fall below a pixel at 13px. `AGENT_GROUP_ICON` itself is untouched, so the landing page's byte-identical copy and its sync test are unaffected. The multi-element `AgentsGlyph` component is gone — the pair is expressible as one path, so it goes through `GlyphIcon` like every other line icon.
- **Memory** — a brain with its stem, replacing the open book. The book had **two homes** (the dock's table and `memory-changes-card.tsx`, hand-typed); the brain is exported once from `icons.tsx` and imported by both.
- **Trace observation** — an open eye. `NAV_ICONS.traces` itself changed, so anything drawing the Traces nav mark moves with it.
- **Remote control** — `MESSAGING_RELAY_ICON`, the paper plane a session row already flies while relaying, so one feature wears one mark everywhere. That left `MESSAGING_ICON` (the chat bubble) with no caller anywhere; it was deleted. `plugins-page.tsx`'s `SEND_ICON` is a different plane for a different meaning and is untouched.
- **Hide launcher** — the same eye struck through corner to corner, replacing `CloseIcon` and its 14×14-grid special case. It keeps `toneInk.muted`, so it still reads as a lesser thing than the panels above it — including when the ball mirrors it.
- **Unchanged**: the Workspace folder, the scheduled-tasks alarm clock, the terminal prompt, `AGENT_GROUP_ICON`, and the ball's resting workbench tiles.

The eye outline is composed once into both the Trace mark and the hide mark, rather than typed twice, so the pair cannot drift.

### Two adjacencies worth a look before merging

1. **Trace is an eye and "hide launcher" is a slashed eye, and they sit in the same fan about 15px apart.** The slash is a full corner-to-corner stroke and the hide mark drops the pupil, so the two differ by two features rather than one, and they are separable at 13px. They are still visibly siblings. If that is too close, "hide" is the one to give a different mark — it is the only caller of `HIDDEN_ICON`.
2. **`NAV_ICONS.models` is already a brain** (the Models library in the sidebar), so Memory-as-a-brain puts a second brain in the app. They differ in silhouette — Memory has a full midline and a stem, Models is two closed lobes — but at 13px both read as "a round thing with a vertical line". Raising it rather than deciding it.

## Verification

Node v24.18.0, from the worktree.

```
$ pnpm --config.verify-deps-before-run=false -r build
exit 0

exit 0

Checking formatting...
```

Vitest ran on the remote host `vps3-test`:

```
 Test Files  129 passed (129)
      Tests  1817 passed (1817)

    --build "@prismshadow/penguin-core" -- --filter @prismshadow/penguin-landing test
 Test Files  10 passed (10)
      Tests  71 passed (71)
```

1817 web tests is the 1816 baseline plus the one assertion added here: `test/icon-scale.test.ts`'s "one glyph, one home" suite now pins the memory brain to a single file, which is the rule that did not exist before this change. The landing run covers `agent-glyph-sync.test.ts`, since the agents mark was in scope.

Every new path was rendered through `GlyphIcon`'s exact SVG settings at 13 / 15 / 18px on both grounds, screenshotted with the repo's chromium, and looked at — including a magnified 1× raster to check what actually smears at 13px, and a simulated fan column and tab strip to judge the marks against each other. Several candidates were rejected on that evidence: the first two-robot drawings were too small to carry weight next to their neighbours, a cloud-silhouette brain read as a tree, and `lucide`'s broken-outline eye-off turned to mush at 13px.


hiyouga:
## Follow-up: Models gets the cube, Memory keeps the brain

Resolves the second adjacency flagged above. `NAV_ICONS.models` was a brain of its own; beside the Memory brain the dock now draws, the two reduce to the same picture at the sizes those rows render.

**Picked: a cube** — one model, as a discrete thing picked off a shelf.

Both suggested candidates were drawn and rejected on the render, not on argument:

- **Stacked layers** renders fine at 13px, but `machines` is the **next nav row down** and is already two stacked server units. At 13px and 16px the two marks reduce to the same horizontal bands, in adjacent rows. That is a worse collision than the one being fixed.
- **A chip** puts its interior pins exactly where detail smears at 13px, and it re-litigates a decision already recorded in the table: the chip named the hardware a model runs on, not the model.

Also rejected on the render: a 4-node neural net (collapses into a `>` at 13px), a globe (reads as locale, and this app has a locale switcher), and a database cylinder (reads as storage, and rhymes with `machines` again).

The cube's hexagon is the only silhouette of its kind in `NAV_ICONS`; it is clean against the robot, the puzzle piece, the server stack, the bar chart, the trophy, the folder and the brain.

### One thing found on the way

`agents-page.tsx`'s memory-count stat carried a **third** hand-typed brain — lucide's, with an extra inner fold that does not survive the 13px it renders at. Left alone, the Memory mark would still have had two homes, with `icon-scale.test.ts` unable to catch it because the second copy is a *different* path. It now draws the shared `MEMORY_ICON`. With that, the lucide brain path is gone from the repository entirely.

### Render sites checked

`grep -rn "NAV_ICONS.models"` turns up exactly two: `app-layout.tsx:92` (the collapsed rail, 18px) and `sidebar.tsx:1504` (the nav rows, via `NAV_GROUP_KEYS`, 16px). Both read the path out of `NAV_ICONS`, so the one string covers them, and nothing hard-codes the old brain — `grep -rn "5.997"` across `web`, `landing` and `desktop` is now empty.

### Verification

```
$ pnpm --filter @prismshadow/penguin-web typecheck     → exit 0
```

vitest on `vps3-test`, re-run after the change:

```
web:      Test Files  129 passed (129)      Tests  1817 passed (1817)
landing:  Test Files   10 passed (10)       Tests    71 passed (71)
```

The contact sheet was regenerated from the shipped sources and read, including the two comparisons that decided this: **Models vs Memory** and **Models vs Machines**, both at 13 / 16 / 18px, plus the full nav strip on light and dark.

hiyouga:
## Three reworks: chip, transverse brain, detailed alarm

### 1. Models is a chip — and the pins do hold, at the sizes it actually draws

`M5 5h14v14H5z` body, `M9 9h6v6H9z` die, **three pins a side**.

Honest answer on the pins, with the render as evidence: **at 16px and 18px — the two sizes `NAV_ICONS.models` actually renders at (sidebar nav rows, collapsed rail) — the pins resolve individually on all four sides.** At 13px they do *not* separate; they fuse into a continuous toothed fringe. The mark still reads as a chip there, because the concentric body-and-die is doing the work and the fringe reads as "pinned edge" even when you cannot count the teeth — but I am not going to claim individual pins at 13px, because the pixel zoom says otherwise.

Three was the right number: six a side fuse into a serrated edge even at 16px, and two a side leave the edge too sparse to read as a package.

**The `machines` collision does not happen.** Without the die it would have: a bare body with side ticks and a stack of server units both reduce to "a rectangle with lines", which is exactly what killed the layers candidate. The die makes the chip *concentric*, and pins ring all four sides where a server stack can only ever have horizontal bars. Side-by-side at 13/16/18 in the sheet — no confusion at any size.

### 2. Memory is the brain from above

Two hemispheres, three gyri a side along the outer edge, and the fissure between them. The fissure runs the **full height** rather than stopping short of the outline: it is the feature carrying the whole mark once the gyri stop resolving, and a line that ends inside the shape closes up at these sizes. At 12–13px the mark reduces to an ellipse with an open midline, which is the correct reduction of a transverse section. Three bumps a side survive; four start to wobble at 18px without adding anything.

### 3. The alarm clock gained detail that survives 12px

Domed bells on the shoulders (replacing two bare diagonal ticks), a dial with the hands' right angle, and two splayed feet (it had no base at all). The whole mark was raised ~0.8 units so it sits centred in the box rather than low.

**Rendered at 12px specifically**, because `ScheduleMark` draws at `ICON_SIZE.rowMark` on a session row — bells, feet and the hand angle each hold a whole pixel there, and the notch between the two bells stays open, which is what keeps them reading as two. Rejected on that evidence: a top stem/button between the bells (at 12px it fills the notch and the bells merge into one cap) and 10:10 hands (the shallow V merges; the L-shape is far more legible small).

### Verification

```
$ pnpm --filter @prismshadow/penguin-web typecheck     → exit 0
```

vitest on `vps3-test`, private log paths:

```
web:      Test Files  129 passed (129)      Tests  1817 passed (1817)
landing:  Test Files   10 passed (10)       Tests    71 passed (71)
```

`icon-scale.test.ts`'s "one glyph, one home" needle was moved to the redrawn brain, so it still pins the Memory mark to a single file.

Contact sheet regenerated from the shipped sources, now with a **12px column** for the alarm, and read before pushing: the full table at 12/13/15/16/18/64 on both grounds, the fan column, the dock tab strip, the sidebar nav strip, and the Models-vs-Memory and Models-vs-Machines comparisons at 13/16/18.

hiyouga:
## Review

This redraws six marks in `panel-meta.tsx`'s table plus `NAV_ICONS.models`/`traces`, and makes the
launcher ball resolve one of three states from the pointer or focus. I re-rendered every shipped
path through `GlyphIcon`'s exact settings at the sizes each mark actually draws at and read the
raster — the claims in the two follow-up comments hold, with one wording caveat noted below. The
interaction change is clean: no listener is bound per render, `usePointerDrag` still owns capture
and unmount cleanup, the passive wheel/touchmove listeners still never `preventDefault`, and the
keyboard path gets the same three ball states as the pointer. Ship after the two prose fixes, which
I have pushed; one caption question is left for you.

### Findings

**1. The Web App manual still describes the old ball** — `packages/docs/content/web-app.en.md:89`,
`web-app.zh.md:88` · *should fix*

That section documents this widget down to the 32px inset and the caption's exact wording, and said
the ball "carr[ies] a workbench glyph" with pointing at an entry showing "its name in the caption".
After this PR the ball also swaps its glyph, and a third caption string exists (`launcherOpen`) that
the manual never mentions — so a reader following the manual sees behaviour it does not describe.
Added one sentence to each language recording the three states. ✅ Fixed in `6310ee42`

**2. The changelog entry's lead counts five marks over a list of six** —
`changelog/unreleased/2026-09-10-dock-launcher-icons.md:19` · *nit*

"Five marks were redrawn … so each moved on all four surfaces at once" is followed by six bullets.
The blank line before the sixth also makes the whole list loose in CommonMark, so it renders with
paragraph spacing throughout rather than separating that item. The alarm clock is also the one entry
whose smallest placement (`ICON_SIZE.rowMark`, `sidebar.tsx:2738`) is not one of those four
surfaces, so I moved it out of the list into a paragraph — the count is true again and the list
stays tight. Same edit in the `.zh.md`; `Scope` gains `docs` for finding 1. ✅ Fixed in `6310ee42`

**3. The ball offers to "Open" while the fan is already open** — `dock-launcher.tsx:471` · *nit*

`ballActive` is true and `hovered` is null for the whole period after a click that opens the fan
with the pointer resting on the ball — the commonest state there is — so the caption reads "Open"
and the glyph is the expand mark while a click would fold. `aria-expanded` is correct, so this is
visual only. Which way it should go (a close word, or falling back to `launcherCaption` while
`fanOpen`) is a product call, so I left it. ⬜ Left for the author

### Render evidence

Every path below was extracted from the shipped source with esbuild, drawn through `GlyphIcon`'s
24×24 / `strokeWidth` 1.7 / round-cap geometry at the sizes its own call sites pass, on `#ffffff`
and on the app's `#0d0d0d` dark ground, screenshotted with the repo's chromium at
`deviceScaleFactor: 1` and also pixel-zoomed with smoothing off. Alpha figures are per-pixel
coverage read back off the canvas.

- **Models vs Machines** — holds at every size. `models` at 16px is concentric body + die with a pin
  fringe on four sides (31.8% ink); `machines` is two horizontal bars (28.4%). Nothing in the nav
  table reduces to the other. Render sites confirmed as the two claimed: `sidebar.tsx:1504`
  (`navRow` 16) and `app-layout.tsx:92` (`size={18}`) — nothing draws it at 13.
- **The pins** — the honest claim is the right one. At 16px the top edge reads `....44.44.44....`:
  three separated stubs a side, resolving individually. At 13px they fuse (`...45.8.54...`) into the
  toothed fringe the comment describes.
- **Models vs Memory** — no relationship left. Worth recording why the swap was necessary: the *old*
  `models` brain and the new Memory brain both reduce, at 13px, to a ring with a one-pixel vertical
  midline — I rendered them side by side and they are the same picture.
- **Memory at 13px** — reduces to exactly what the comment says: a ring with a full-height fissure
  (column 6 at alpha 0.89, its neighbours empty, so the midline is crisp rather than smeared across
  two columns). The gyri do not survive at 13/15px; they are visible at 18px and up.
- **Alarm at 12px** — bells read as two: the row above them is empty between the domes, and the row
  through them is 0.22 in the notch against 0.50–0.67 on the domes. The hands' right angle holds.
  One caveat on "bells, feet and the hands' right angle each hold a whole pixel there": the feet
  hold at roughly a third of ink (0.33–0.44) and their outermost tips land at ≈0.0–0.05, so they are
  the faintest thing in the mark at 12px rather than a whole pixel. Still legible, and total ink
  went 15.8% → 19.1%, so the mark is denser but not closed up.
- **Trace eye vs hide eye at 13px** — separable by two features as claimed: the trace eye keeps a
  pupil ring the hide eye drops, and the hide eye's slash is a full corner-to-corner stroke.
- **Subagents** — the eye dots do *not* separate from the head's side walls at 13px (dots at 0.33
  against walls at 0.33–0.55); the pair reads there as a large and a small rounded square on a
  diagonal, and the dots only resolve from about 18px, the ball's rung. Not a weight problem — at
  13px it carries 20.5% ink against Memory's 20.1% and the alarm's 19.0%.

### Left for you

**#682 (`feat/file-browser-context-menu`) is safe to land in either order, and no silhouette
collides.** `git merge-tree --write-tree HEAD origin/feat/file-browser-context-menu` succeeds
(tree `516fa602`). #682 does not touch `panel-meta.tsx`, and its `ADD_TO_CHAT_ICON` is a
self-contained literal rather than a reference to the `MESSAGING_ICON` this PR deletes, so the
merged tree drops that constant with no dangling consumer. Drawn at 15px, no pair across the two
sets reduces to the same silhouette; the only near-pair is #682's own upload/download tray, which
is deliberate there. Worth knowing in the other direction: `NAV_ICONS.traces` was three stacked
lines, and #682's `WRAP_TEXT_ICON` is three lines with a turn at the bottom — at 15px those two
were nearly the same mark, and this PR's eye removes that collision before it arrives.

### Verification

```
Node v24.18.0, worktree ../penguin-harness-wt/rv-680, on 6310ee42.

$ pnpm --config.verify-deps-before-run=false -r build      exit 0
$ pnpm --filter @prismshadow/penguin-web typecheck          exit 0
$ pnpm lint                                                 exit 0 (oxlint, no findings)
$ pnpm format && pnpm format:check                          All matched files use Prettier code style!
$ git diff --check                                          clean

vitest on vps3-test (never on the review box):
$ pnpm --filter @prismshadow/penguin-web exec vitest run test/icon-scale.test.ts
    Test Files  1 passed (1)      Tests  10 passed (10)
$ pnpm --filter @prismshadow/penguin-web test
    Test Files  129 passed (129)  Tests  1817 passed (1817)
$ pnpm --filter @prismshadow/penguin-docs test        (docs content changed)
    Test Files  7 passed (7)      Tests  53 passed (53)

Replays against the merge base 34f17ad3:
- new "one glyph, one home" needle: git grep -F "3.4 3.4 0 0 1-4.4 2.6M12 2.6v18.8" 34f17ad3 --
  packages/web/src → no match, so the assertion fails there; at HEAD it matches exactly
  components/ui/icons.tsx, and MEMORY_ICON really does have one home (three importers).
- git grep AgentsGlyph 34f17ad3 → panel-meta.tsx only (its own definition and use), so the deleted
  export had no other consumer; MESSAGING_ICON likewise had exactly one caller.
- every panelGlyph/panelLabel consumer (chat-page.tsx, dock-panel.tsx ×5, dock-launcher.tsx) still
  resolves; no case of the switch lost its glyph.
```

hiyouga:
## Rework: sagittal brain, and a ball that names the next click

Pushed as `940c3f04`, on top of `origin/main` merged in.

### Memory is a sagittal section, not a transverse one

The mark was two hemispheres with the fissure between them, seen from above. It is now the profile,
cut open, with the fold spiralling inward.

Four shapes were drawn and read at 12 / 13 / 16 / 18 / 48 on both grounds before choosing:

- **outline + an open C** is the most legible small, and at 48px it reads as ©. A ring with a C in
  it is the copyright sign; the lobes are the only thing arguing otherwise, and they lose.
- **outline + two separate folds** reads as a face at 48px.
- **outline + a C with an inner tick** reads as a euro sign.
- **outline + a spiral curling in from the lower left** is what shipped. It reads as a fold rather
  than a letterform, and the curl is the feature that says "cut open".

`MEMORY_ICON`'s smallest real render is 13px (`inlineGlyph`, the memory-changes card and the agent
cards' memory count); `rowLead` 14 and the dock tab are larger. At 13 it holds as a lobed ring with
a resolved interior. I am not claiming the individual turns separate at 12 — they do not, and
nothing draws it that small.

The earlier follow-up's claim about the transverse reduction is void with the shape it described.

### The ball says what the click will do

`launcherOpen` was the caption whenever the pointer sat on the ball, fan open or shut — so a click
about to **fold** the fan was labelled as one that would open it, with the expand glyph beside it.
Caption and glyph now both follow `fanOpen`: expand and "Open" while it is shut, collapse and
"Close" while it stands open. `COLLAPSE_ICON` is `EXPAND_ICON`'s four corner brackets turned
inward, so the two read as one axis rather than two ideas.

This is the caption question raised as finding 3 in the review above.

`icon-scale.test.ts`'s "one glyph, one home" needle moved to the redrawn brain, so it still pins the
Memory mark to a single file. Both the English and Chinese changelog entries and the `web-app`
manual pages were corrected in the same commit.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Memory is the brain the emoji draws

Pushed as `823aee70`.

The two-hemisphere mark said "a brain" without saying "memory" — a smooth pair of lobes with a gap
between them. This is 🧠's brain instead: a gyrified mass in profile, its folds showing.

**Three strokes, and each earns its place at 13px**, which is the smallest size any of the three
surfaces draws it at (the memory-changes card and the agent cards' memory count):

1. the lobed outline — a smooth one is a cloud;
2. a fissure winding down the left of centre;
3. a second fold curling in the right half.

Eleven shapes were drawn and read at 12 / 13 / 16 / 18 / 48 on both grounds before this one. The
rejected ones are worth recording, because each failed for a reason that constrains the next
attempt:

- **One interior stroke** leaves a blob with a line in it. Two is the step from "blob" to "brain".
- **Flat parallel sulci**, which is what a real cortex actually has, read as a hamburger menu
  inside a cloud at every size below 48.
- **A third fold** is scribble by 16px.
- **A single open C** inside the outline is a copyright sign at 48px.
- **Two symmetric arcs plus a lower one** is a face.
- **A brainstem stub** on the outline is a speech-bubble tail.

`icon-scale.test.ts`'s "one glyph, one home" needle follows the redrawn mark, so Memory still has
exactly one picture of itself. Both changelog entries were updated.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Memory is the brain in profile, outline only

Pushed as `2f4a3ddb`.

The emoji brain carried two interior folds. At **13px** — the size the memory-changes card and the
agent cards' memory count actually draw it at — interior strokes are texture rather than features,
so they cost complexity and returned nothing.

This is the lateral view, **one closed path with nothing inside it**. Two features do the work, and
both sit on the silhouette because that is the only place 13px can still see them:

1. three lobes across the crown;
2. the cleft at the lower front, where the temporal lobe hangs below the frontal one.

That cleft is what makes the view lateral rather than generic, and it is the reason the shape is not
a circle.

### What the rejected shapes taught

Twenty were drawn and read at 12 / 13 / 16 / 18 / 48 on both grounds. Each failure constrained the
next attempt, so they are worth recording:

- **A smooth profile is a cloud**, at every proportion tried — wide, elongated, wedge-shaped, with
  or without a flat base. Without lobes there is no brain reading at all.
- **Six small gyri around the perimeter are fuzz below 16px.** The scallops have to be three and
  deep, not many and shallow, or the mark turns into a fluffy disc.
- **A cleft narrower than about a fifth of the icon closes up at 13px.** It has to be big enough to
  look almost like a bite at 48px to still be a cleft at 13.
- **A smooth outline with the Sylvian fissure drawn inside it** reads as a blob with a dash through
  it — the same failure as the earlier open-C, which read as ©.
- **A brainstem stub** on the outline reads as a speech-bubble tail.

Checked against the marks it sits among — the Models chip, the Machines server stack, the usage
bars — with no silhouette collision at any size.

`icon-scale.test.ts`'s "one glyph, one home" needle moves to the cleft: the one reversed arc in the
path, and the feature that makes the view lateral. Both changelog entries were updated.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Memory follows a real brain profile now

Pushed as `c3eec343`.

The hand-drawn lateral outline was a lumpy blob — three arcs across the crown and a bite out of the
lower front, which is not what a brain looks like from the side. This one is **traced** from the
reference rather than guessed at.

### How

1. Close the gaps in the reference's strokes, so a flood fill cannot leak into the interior. It is a
   drawing of separate strokes, not a closed curve; without this step the fill reaches everywhere
   and the "silhouette" is just the ink.
2. Fill from outside, invert, undo the closing → the filled silhouette.
3. Erode by half the reference's stroke width, so the traced line lands on the stroke's **centre**,
   which is where a `strokeWidth` 1.7 path will be drawn.
4. Trace the boundary, simplify with Ramer-Douglas-Peucker, rebuild as Catmull-Rom curves on the
   24-grid.

Tolerances from 4 to 22 were rendered at 12 / 13 / 16 / 18 / 48 on both grounds. **10** is where the
crown lobes and the brainstem still hold and the path stops carrying points nobody can see; below
about 14 points it rounds off into a cloud again, which is the failure every hand-drawn attempt hit.

### What shipped

One closed path, 22 points, 597 characters. That is longer than the geometric marks around it, and
deliberately so: this is a profile rather than a figure built from arcs, and the extra points are
the lobes. Still nothing inside it — interior gyri are what 13px cannot hold, and 13px is the
smallest size the memory-changes card and the agent cards' memory count draw it at.

Checked against the Models chip and the Machines server stack at every size: no silhouette
collision. The `icon-scale.test.ts` needle moved to the new path's opening curve.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## The brain gets its gyri

Pushed as `9cf700a0`.

The traced outline was the right shape and an empty one. The interior is what makes a brain read as
a brain rather than as a lobed cloud, so it is in now: **twelve gyri**, taken from the same
reference by the same pipeline.

### How they were extracted

1. Erode the filled silhouette past the outline band, so only the interior strokes are left.
2. Split what remains into connected components.
3. Thin each to a one-pixel skeleton (Zhang-Suen), then walk the skeleton end to end for that
   stroke's own polyline.
4. Simplify and smooth it on **the same transform the outline uses**, so the parts stay registered.

Sixteen strokes survive the extraction. Twelve ship; the four dropped are the shortest, which at
these sizes are specks rather than marks.

### Why twelve and not four

Counts of 4, 6, 9, 12 and 15 were rendered at 12 / 13 / 16 / 18 / 48 and inspected at six times
magnification. The finding that decided it: **at 13px none of the interiors resolve one by one** —
they are texture at any count. So a sparser interior buys nothing at the smallest size and costs
detail at 18px and above, where each stroke is its own mark.

That also answers the earlier "too complex": the problem then was the shape, not the stroke count.
A bad shape with three interior strokes reads worse than a real profile with twelve.

### The cost, stated plainly

1509 characters — much the longest path in the family, roughly 2.5× the next. It is a traced profile
rather than a figure built from arcs, and the interior is the half that carries it. The `icon-scale`
needle still matches the outline's opening curve, so Memory keeps its one home.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Six gyri, and spread out

Pushed as `84e23371`. 1094 characters, down from 1509.

The count is halved — the average of twelve and the empty outline before it — but the count was not
the real problem. **The crowding was.** The selection kept the longest strokes, and the longest ones
cluster in the same lobe, so one region filled to a smudge while the rest of the shape stayed empty.

So the selection now carries a **minimum separation** as well as a length order: a stroke is kept
only if it sits far enough from every stroke already taken. That is what spreads six marks across
the whole shape instead of stacking them where the reference happened to be busiest. Halving the
count alone would have thinned the empty regions just as much as the crowded one.

Separations of 0, 1.6, 2.2, 2.8 and 3.4 grid units were rendered at 13 and 18px and inspected at six
times magnification:

| separation | strokes |
| --- | --- |
| 0 | 15 |
| 1.6 | 12 |
| 2.2 | 10 |
| 2.8 | 7 |
| 3.4 | 6 |

3.4 is where the last cluster breaks up, and it lands on six.

The outline is untouched, so the `icon-scale` needle still matches its opening curve and Memory
keeps its one home.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## The cortex is on the outline; the interior is empty

Pushed as `7348066f`. One closed path, 1130 characters, nothing inside it.

The folds that were separate strokes in the middle are on the **perimeter** instead, so the edge
carries the reading the interior used to. Each of the profile's long stretches gets a gyrus pushed
out about 0.65 grid units, with the joins between them pulled in to read as the sulci.

**The brainstem and the notch under the temporal lobe stay smooth, deliberately.** They are too
short to ripple without being destroyed, and they are exactly what make the silhouette a brain
rather than a cloud. Rippling everything uniformly is how the earlier attempts failed.

### Two approaches drawn and thrown away first

- **Amplify the reference's own perimeter undulation.** Its folds are already high-frequency, so
  deepening them gives fur below 16px rather than cortex. Tried at amplification 1.8 and 2.6 over
  smoothing windows of 40, 60 and 90.
- **Resample the outline to a low point count and push alternate points in and out.** Clean,
  regular scallops — and no brain. An even resampling cuts the stem and the temporal notch first,
  which leaves a spiky blob. Tried at 14, 18 and 22 points.

What worked is keeping the traced profile at full fidelity and adding folds only where a stretch is
long enough to carry one.

### Picking the fold

Widths of 1.9 to 3.2 units and depths of 0.5 to 0.8 were rendered at 13 and 18px and inspected at
six times magnification. Below about 2.4 the folds stop reading as cortex and become beads on a
string. 3.0 wide by 0.65 deep is what shipped.

The `icon-scale` needle moved to the new opening curve, so Memory keeps its one home.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Cortex sized for 5mm, not for the contact sheet

Pushed as `88ca725e`. 339 characters — down from 1130, and from 1509 for the version with an
interior.

You were right, and the arithmetic says why. At 13 to 18px one grid unit is about **three quarters
of a pixel**. The old fold stood 0.65 units out, which is **half a pixel**: it was not subtle, it
was absent. I had been judging the folds at 48px, where everything looks fine.

Nine gyri now, each spanning 5.5 units and standing 1.3 out — the smallest that still reads at 13px,
checked at six times magnification rather than on the contact sheet.

### The part that mattered more than the size

They are **real outward arcs**, not displaced points. Pulling a point out far enough to be seen
turns the fold into a **tooth**, because the curve fitted through it comes to a corner — I drew that
version first and it looked like a saw blade. An arc of known chord and sagitta stays round however
deep it is drawn, so depth and roundness stop fighting each other.

That is also why the path got shorter: an arc says in one command what a fitted curve needed six
numbers to approximate.

### Unchanged on purpose

The brainstem and the notch under the temporal lobe stay smooth. They are too short to ripple
without being destroyed, and they are what make the silhouette a brain rather than a cloud.

The `icon-scale` needle moved to one of the new arcs, so Memory keeps its one home.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Memory takes your drawing

Pushed as `01898dbe`.

Two adjustments were needed to fit the shared renderer, and neither changes what is drawn:

- **One path string.** `GlyphIcon` emits a single `<path>`, so the outline and the gyri become two
  subpaths of one `d`.
- **The transform is baked in.** The drawing carried its figure on `translate(0 1.25)`; that shift
  is folded into the coordinates, because a transform is an attribute the renderer has nowhere to
  put. Only the absolute commands moved — the relative ones are deltas and are unaffected.

**Verified rather than assumed.** The original markup and the flattened single path were rasterised
at 480px and compared pixel by pixel: 163 of 52581 ink pixels differ, all single-pixel antialiasing
along edges, which is exactly what baking a translate into coordinates does to subpixel rounding.
No geometric difference.

One deliberate change: it is stroked at the family's **1.7**, not the 1.6 it was drawn at. That
weight belongs to the set rather than to any one mark, and `icon-scale.test.ts` enforces it.

The `icon-scale` needle moved to the gyri subpath's opening, so Memory keeps its one home.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

PR #681, opened 2026-09-11
feat(web): Trace summary shows average tool calls per round

hiyouga:
The Trace view's global summary carried a **compaction count** that could only ever read 0 or 1: `context-engine.ts` defers the Trace rotation until the first write after a compaction, so a single Trace file holds at most one compaction round. That is one of nine slots in the summary spent on a figure with no information in it.

It is replaced in place by **average tool calls per round**, computed as `global.toolCalls ÷ analysis.tasks.length` — the two figures printed directly above it in the same column, so a reader can check the division by eye.

## What changed

- `trace-file-view.tsx`: the third row of the summary's first column now renders the average. The denominator is the full round count, compaction rounds included, which is the scope rule the whole column already obeys ("every figure is the sum across rounds and must add up"); the comment there was rewritten to state that constraint, since it used to point at the compaction row for the compaction question.
- `lib/format.ts`: new `formatAverage(total, count)` — one decimal place, kept even at `.0` so an average is never mistaken for the raw count beside it, and `—` when the count is 0. That is the same mark `formatMoney(null)` and `formatTps(null)` already render, so an empty Trace reads consistently across all three columns.
- Both dictionaries: `traces.compactions` became `traces.avgToolCalls` in the same position — zh `每轮平均工具调用` (the wording the request used), en `Avg tools / round`. `SummaryRow`'s label is `shrink-0` against a `truncate` value, so a long label would eat the number; hence the short spelling. The old key had exactly one reader (the row that is gone) — the round card's compaction badge goes through `chat.compactionTitle`, which is untouched.
- `test/format.test.ts`: `formatAverage`'s rounding and its zero-count case.
- `e2e/chat.spec.mjs`: the Trace-summary label list asserted `压缩次数` by name, so it moves with the row.

The server is untouched: `compactionCount` stays on the Trace analysis in the API and simply stops being rendered.

Design side: Prism-Shadow/penguin-harness-design#139 updates `specs/01-PRINCIPLES.md`'s Trace-page stat list to match.

## Heads-up for whoever merges second

[#676](https://github.com/Prism-Shadow/penguin-harness/pull/676) (`fix/trace-duration-total`) edits the same file — `SummaryRow`'s signature and the Elapsed row in the third column. Neither is touched here; the hunks sit ~17 lines apart in different columns and should merge cleanly, but it is worth a look.

## Verification

Node v24.18.0. Tests ran on the remote host `vps3-test` (this session's standing instruction), with `@prismshadow/penguin-core` built there first.

```
  (clean)

Checking formatting...

$ PENGUIN_TEST_HOST=vps3-test .agents/scripts/remote-test.sh . \
 ✓ test/format.test.ts (44 tests) 10ms
 ✓ test/trace-refresh.test.ts (17 tests) 4ms
 ✓ test/trace-compaction-badge.test.ts (3 tests) 2ms
 ✓ test/placeholders-parity.test.ts (3 tests) 2ms

 Test Files  129 passed (129)
      Tests  1818 passed (1818)
```

`e2e/chat.spec.mjs` is not part of that suite — it runs under Playwright with a live key in the
`installer-e2e` job. Its edit is a one-word label swap, checked here with `node --check` and left
to CI.


hiyouga:
## Review

The statistic is sound: the denominator is exactly the "Turns" row printed above it and the
numerator exactly the "Tool calls" row (`trace-file-view.tsx:365` sets `g.toolCalls =
analysis.toolSpans.length`), so the eye-check the comment promises actually holds, and zero rounds
lands on `—` rather than `NaN`. Two things needed fixing — the English label names a unit the page
never uses, and the changelog states as fact something that is false when a compaction fails — both
done in d3aac463. Ship after a look at those.

### Findings

**1. The English label's denominator names a unit the page never uses** — `packages/web/src/lib/strings-en.ts:2615` · *should fix*

The counts column rendered `Turns` / `Tool calls` / `Avg tools / round`. `round` appears nowhere
else in the English Trace dictionary: `tasksLabel` is `"Turns"`, `task(n)` is `` `Turn ${n}` ``,
`taskInput`/`taskOutput` are `"… this turn"`. The row's whole purpose — your comment at
`trace-file-view.tsx:457` and the PR description both say it — is that a reader can check the
division by eye; in English they cannot, because the row does not name either figure it divides.
zh was already internally consistent (`轮次` / `工具调用` / `每轮平均工具调用`).

Changed to `"Avg tools / turn"`. That is one character *shorter* than the original, so your width
constraint (`SummaryRow`'s `shrink-0` label against a `truncate` value) is strictly better off, and
"tools" stays as you spelled it. The zh entry and design PR
[penguin-harness-design#139](https://github.com/Prism-Shadow/penguin-harness-design/pull/139) are
Chinese-only and quote no English label, so nothing else moves. ✅ Fixed in d3aac463

**2. The changelog asserts a bound that a failed compaction breaks** — `changelog/unreleased/2026-09-11-trace-avg-tool-calls.md` · *should fix*

> A compaction defers the Trace rotation to the next write, so one Trace file holds at most one
> compaction round and the count could only ever read 0 or 1.

The rotation is pending only after a compaction that *succeeded*. `pendingTraceRotation = true` is
set once, in `startNewContext()` (`packages/core/src/engine/context-engine.ts:1941`), which is
reached from `discardContext` (`:1540`) and from the one `summarize` branch that extracted a usable
summary (`:1659`). A `summarize` that is aborted or exhausts its retries emits `compaction_begin`
plus a non-`completed` `compaction_end`, keeps the old context and rotates nothing — and the
context is still over threshold, so the next trigger writes a **second** `compaction_begin` into the
same file. Concrete input: one Trace file whose compaction failed once and then succeeded →
`analysis.compactionCount === 2`.

It does not change the verdict (0/1/2 is still a near-empty slot), but it was on disk as fact, and
`changelog/README.md` keeps reasoning off disk anyway — so I dropped the sentence from both files
rather than repairing it. The same claim opens the PR description; worth correcting there, since
that is where the reasoning now lives. ✅ Fixed in d3aac463

**3. A nonzero tool-call count can render as `0.0`** — `packages/web/src/lib/format.ts:114` · *nit*

`formatAverage(1, 21)` → `"0.0"` while the row directly above reads `1`. It takes ≥21 turns with a
single tool call, so it is uncommon but reachable in a long, mostly conversational Trace. Worth
naming only because `formatMoney` two definitions below goes out of its way to avoid exactly this
shape — *"significant digits, not fixed decimals, so a nonzero cost is never rounded down to zero"*
(`format.ts:124`). Whether an average deserves the same treatment (`<0.1`, or two decimals below 1)
is your call, so I left it. ⬜ Left for the author

### Checked and clean

- **Scope of the denominator.** Compaction rounds are in both the visible round count and the
  divisor — consistent, and a compaction round can hold no tool call, so it dilutes rather than
  distorts. Subagent rounds and tool calls are in neither: every push in `trace-service.ts` is
  behind the `hasOrigin` guard (`:1318` for `toolSpans`), and the writer never records
  child-session messages at all (`trace/writer.ts` `isRecordable`). A round in flight would be
  counted, but `trace-refresh` only re-fetches on a settled turn, so the panel never shows one.
- **Division by zero.** `tasks.length === 0` (a file with only `session_meta`) → `—`; `7 / 0` is
  guarded too, not just `0 / 0`.
- **No duplicate helper.** `format.ts`'s other one-decimal helper (`trimZero`) drops a trailing
  `.0` deliberately, which is the opposite requirement; `formatPercent` rounds to a whole percent.
- **Tests are real.** Every assertion in the new `format.test.ts` block fails at
  `$(git merge-base origin/main HEAD)` — `formatAverage` does not exist there, so the import is
  `undefined`. Boundaries covered: `.0` retention, zero numerator, round-down, round-up, both
  zero-count cases.
- **e2e line not loosened.** `getByText("工具调用", { exact: true })` still resolves only to the
  "Tool calls" row — `exact` is whole-string, so the new `每轮平均工具调用` does not collide with it,
  and the swapped label is asserted the same way the removed one was.
- **Nothing left behind.** `rg` finds no remaining reader of `traces.compactions`, no other web
  reader of `compactionCount` or `reconnectCount`, and no docs page naming the removed row. The
  server DTO is untouched, so there is no persisted-shape or compatibility question.
- **#676 merges clean.** `git merge-tree --write-tree HEAD origin/fix/trace-duration-total` returns
  a tree, no conflict. It widens `SummaryRow` with an optional `detail` prop and rewrites the
  Elapsed row; the new `SummaryRow` call here passes no `detail`, which is valid against that
  signature either way round.

### Verification

```
$ export PATH=/home/cc2/.local/node-v24/bin:$PATH   # node v24
$ pnpm --filter @prismshadow/penguin-core build      # fresh worktree, web typecheck needs the dts
$ pnpm --filter @prismshadow/penguin-web typecheck   # clean
$ pnpm lint                                          # oxlint, clean
$ pnpm format && pnpm format:check                   # All matched files use Prettier code style!
$ git diff --check                                   # clean
$ node --check packages/web/e2e/chat.spec.mjs        # syntax ok (Playwright not run — see below)

# vitest on the build box (vps3-test), never locally:
$ vps3-vitest.sh <worktree> rv-681 -- pnpm --filter @prismshadow/penguin-web exec vitest run test/format.test.ts
   Test Files  1 passed (1) · Tests  44 passed (44)
$ vps3-vitest.sh <worktree> rv-681 -- pnpm --filter @prismshadow/penguin-web test
   Test Files  129 passed (129) · Tests  1818 passed (1818)
```

`packages/web/e2e` was not run: Playwright is not installed on the build box and the local suite has
pre-existing failures. The changed line was read and syntax-checked instead.

PR #682, opened 2026-09-11
feat(web): a Files panel with a context menu, wrapping, and an editor that looks like the file

hiyouga:
The dock panel that browses the Workspace is renamed to **Files** (文件浏览), and gains the
per-entry actions it previously had nowhere to put.

## What a secondary click now offers

Right-click, Shift+F10 and touch press-and-hold all open the same menu, on a tree row and on
the preview body:

| | file | folder |
| --- | --- | --- |
| Copy relative path | ✓ | ✓ |
| Add to conversation | ✓ | ✓ |
| Upload here | | ✓ |
| Download | ✓ | |

On the preview the target is the file on screen, and while a non-empty selection sits inside
the preview, one more row offers **that selection** rather than the whole file.

Two gestures deliberately reach nothing. A right-click inside the HTML or PDF preview goes to
the iframe and cannot be heard from this side — no attempt is made to fake one. A right-click
inside the in-place editor keeps the browser's own menu, because that is how text gets pasted
into it. A gesture that lands between tree rows (the search box, the space below the last row)
likewise keeps the native menu.

## Add to conversation

There is no mention mechanism in this app and none was built: the `@` is literal text for the
reader, and nothing parses it. A file or folder inserts `@<workspace-relative path>` (a folder
keeps its trailing slash) at the composer's caret, leaving the rest of the draft untouched. A
preview selection inserts a fenced block instead:

    @src/lib/workspace-tree.ts (L3-L7)
    ```typescript
    …the selection, verbatim…
    ```

The line range is given only where the source view can resolve it from the DOM — the rendered
Markdown and HTML views have no line structure, and a guessed range would be a lie, so those
print the path alone. The fence is opened one backtick longer than the longest run inside the
selection, so a selection that carries fences still nests.

`ComposerControl` gained a second method for this. `fillPrompt` replaces the whole text body,
which is wrong for a contribution to a message someone is still writing; `insertAtCaret`
splices and leaves focus and the caret after what it inserted. Because it goes through the
composer's `onTextChange`, the insertion is cached in the session draft like any keystroke.

## The source view and the editor became one presentation

The source view lost its border, its language label and its header bar: it presents the text the
way the editor does. Copy moved to the preview header, beside Edit / Open in new tab / Download.
Both views number their lines and highlight them, both take the Wrap toggle, and pressing Edit
now reflows nothing — because the editor *is* the source view with a textarea over it.

`CodeBlock` was split: `CodeSurface` is the code with no chrome, `CodeBlock` is that surface
inside the message's bordered box with its language label and copy button. Message code blocks
take neither line numbers nor wrapping — horizontal scroll stays a transcript's answer — and
their CSS rule is restored to the pixel, padding included.

**The editor's two layers.** A transparent-text textarea over the surface, stacked in one grid
cell: the taller one sizes the box, the single scroll container carries both, and nothing is
synchronised in JS, so there is nothing that *can* drift. Measured in Chromium before building on
it: `pre.scrollHeight === textarea.scrollHeight` exactly, wrapped (1076/1076) and unwrapped
(865/865).

**The gutter under wrapping** is the part that needed a real answer. A wrapped logical line takes
several visual rows and one number, so the lines became blocks carrying `padding-left: <gutter>`
with a matching negative `text-indent` — the number hangs in the first row's indent, and the code
sits at exactly one gutter width on *every* row. That is all the textarea has to match, and it
matches it with plain `padding-left`; no per-line rule, no second gutter.

Shiki separates its `<span class="line">` wrappers with a newline, which is what breaks the lines
while they stay inline; blocks would then double-space. A transformer drops those separators when
line numbers are asked for. Verified that a `Range` over the result still serialises the source
text verbatim — tabs, blank lines and all — so drag-select-copy and the context menu's "add
selection" are unaffected.

**Highlighting while typing** re-runs 200ms after the text settles, and only up to 32KB. Measured
on this bundle's JS regex engine over TypeScript: 4KB ~11ms, 16KB ~148ms, 32KB ~367ms, 64KB
~936ms. While a highlight is pending the surface shows the *current* text uncoloured rather than
the previous text's colours — a code layer that lags the keystrokes under it is worse than a
colourless one. Line numbers are CSS and cost nothing, so they stay at any size.

**IME.** Composition candidates reach the textarea before they reach `value`, and the textarea's
text is transparent — so the layers swap for the length of a composition. Driven through CDP
rather than assumed (`Input.imeSetComposition` then `Input.insertText`): the candidate is visible
while composing, `你好` lands in the value on commit, and the colours and numbers come back.

## Wrap

The source view gets the toggle the editor already had, and **the two share one stored answer**.
They are the same file seen two ways and Edit swaps them in place; a second preference would let
pressing Edit reflow the file and carry off the line you were aiming at.

Wrap now defaults **on**. `penguin.files.editorWrap` is unchanged and still read, so anyone who
has ever used the toggle keeps their answer and only a browser with no stored value picks up the
new default; the key keeps its name because renaming it would silently discard those answers. The
changelog carries a `## Compatibility` section saying so.

## Shape

One `useRowContextMenu` hook serves the whole tree rather than one per row: the tree draws
hundreds of rows — twice over while a closing subtree animates out — and each hook carries a
long-press timer. The row a gesture landed on is resolved from its own `data-tree-path`, the
attribute the drop hit test already reads, so `workspace-tree-view.tsx` needed no changes at
all. The whitespace rules around an insertion are pure functions in `lib/workspace-tree.ts`
with their own cases in `test/workspace-tree.test.ts`; nothing about React is tested.

No server change: every endpoint this uses already exists.

Design counterpart: [penguin-harness-design #140](https://github.com/Prism-Shadow/penguin-harness-design/pull/140)

## Verification

Run from the worktree, on Node 24.18.0, rebased on `main` at 1a26d1cd8.

```
$ pnpm --config.verify-deps-before-run=false -r build
(exit 0)

(exit 0, no diagnostics)

$ pnpm lint
$ oxlint
(exit 0, no findings)

Checking formatting...
```

Tests on the remote host `vps3-test` (this session's standing instruction), which syncs sources
and rebuilds `dist` there:

```

 ✓ test/workspace-tree.test.ts (44 tests) 13ms
 Test Files  129 passed (129)
      Tests  1825 passed (1825)
```

Baseline is 1816; the 9 added cases are the caret-insertion rules (inline spacing, the blank line
above a block, a clamped caret, the line-range header, a verbatim selection, fence escalation) and
the flipped soft-wrap default, which asserts that a stored answer still wins over it.

**Looked at, not assumed.** The built server was driven with Playwright over a 420-line, 18.2KB
TypeScript file, screenshotting the panel deep in the file (around lines 165-272, and 260px right
where wrap is off) in both themes: source view wrapped and unwrapped, editor wrapped and
unwrapped, typing at the caret, and an IME composition mid-file. The layers stay in register at
depth, the gutter stays pinned while the code scrolls under it, and a wrapped line's continuation
rows sit under its own code rather than under its number. A chat message's code block was shot in
both themes and is unchanged: border, `typescript` label, copy button, no numbers, no wrapping.

Two mechanics were measured before being built on, rather than after: the two layers' scroll
heights agree exactly (1076/1076 wrapped, 865/865 unwrapped), and a `Range` over the block-level
line spans serialises the source text verbatim.

Playwright's own specs were not run: `e2e/files-switch.spec.mjs`, `e2e/dock.spec.mjs` and
`e2e/mobile-dock.spec.mjs` reach the panel through the `dock-pick-workspace` test id, not through
the renamed label, and no selector they use moved.


hiyouga:
## The panel's actions are icon buttons, and a quoted selection stays selected

Pushed as `1b43678e`.

### Icon buttons

Wrap, Edit, Copy, Download, Refresh and Upload are now the same 24px square as the tree toggle
that already sat beside them. Each names itself twice — on the element's own `aria-label`, and in
a tooltip — because a control with no visible text has nothing else to be read by.

- **Wrap** and **Refresh** needed new glyphs; both are in `components/ui/icons.tsx`, on the app's
  24×24 / 1.7 grid. Refresh reuses the arc idiom `HISTORY_ICON` already draws with, so it lands in
  the same family rather than introducing a second way to draw a turn.
- **Copy** is driven by `useCopied` + `CopyCheckGlyph` rather than `CopyButton`: the tooltip here
  is a `Tooltip` panel, not a `title`, and a trigger may carry only one of the two. The glyph swap
  and the `CopiedStatus` live region are unchanged.
- **Upload** loses its "Uploading 2/5…" label, so the count moves into the tooltip and the glyph
  becomes a spinner. Its name goes on the `<input>`, not the `<label>` — a label with no text names
  nothing.
- `toggleActionClass` had one consumer and is gone. `ghostActionClass` stays: "Open in new tab" is
  still a text link.

**`Tooltip` grew a `placement` prop** to make this work. It only knew how to sit to the right of its
trigger, which is right for the collapsed rail it was written for and wrong for a horizontal
toolbar, where the panel would cover the buttons next to the one being asked about — and a Files
panel docked right has no room on that side at all. `placement="bottom"` anchors to whichever of the
trigger's vertical edges faces the roomier half of the window, so the panel grows inward and its
width never has to be measured. The rail's geometry is byte-identical.

Glyphs were rendered at 15 / 24 / 48px on both grounds before pushing, with the two toolbar rows
mocked at their real sizes.

### The selection survives "add selection to conversation"

`insertAtComposerCaret` focuses the textarea inside a `requestAnimationFrame` so the caret lands
after the insertion. Focusing a text field drops whatever the document had selected, so the preview's
highlight vanished on an action that never touched it.

`readSelection` now clones the live `Range`, and `addSelectionToChat` re-applies it one frame on —
after the composer's own callback, which is scheduled first and therefore runs first. A range whose
ends have left the document is dropped rather than re-applied.

No test: `packages/web/vitest.config.ts` is `environment: "node"` with no DOM, and both halves of
this change are DOM behaviour. The honest home for a regression here is `packages/web/e2e/`.

### Verification

```
```

vitest on `vps3-test` (never locally): `@prismshadow/penguin-web` → **129 files / 1825 tests passed**,
before and after the change.

### One thing left

"Open in new tab" is now the only text action in that row, and it only renders for `.html` files.
Say the word and it becomes an external-link glyph too.

hiyouga:
## "Open in new tab" is an icon too

Pushed as `bff75d1b`. Every action in the panel is now a mark.

- The external-link glyph was a private constant in `features/models/models-page.tsx`. A second
  consumer makes that the second home the icon rule does not allow, so it moved to
  `components/ui/icons.tsx` and both call sites read it from there.
- `ghostActionClass` existed for this one link and is gone with it.
- The sandboxing caveat folds into the link's **name** instead of hanging a `⚠` beside the label
  there is no longer room for. The tooltip shows the same string, so the two cannot disagree, and
  the attention tint is a second carrier rather than the only one.

### Verification

```
```

vitest on `vps3-test`: `@prismshadow/penguin-web` → **129 files / 1825 tests passed**.

hiyouga:
## Review

A context menu, line numbers and a transparent-textarea editor for the Files panel, plus the icon-button pass over the panel's actions and a `placement="bottom"` on `Tooltip`. Two defects found and fixed in `5b495b5b`; the rest holds up, including the three things that were asked about specifically — the rAF ordering behind the selection restore, the `right` placement being unchanged, and the accessible names on the new icon triggers. Ship after a look at the fixes, with one compatibility question left below.

### Findings

**1. Shift+F10 on the preview reuses the previous gesture's selection** — `packages/web/src/features/chat/workspace-browser.tsx:1709` · *should fix*

The preview body handed the chord straight to the hook: `onKeyDown={previewMenu.rowProps.onKeyDown}`, which opens the panel through `setAnchor(rowRect())` (`components/ui/context-menu.tsx:154`) and never touches `menuSelection`. Both pointer paths do refresh it — `openPreviewMenu` and the long-press `onPointerDown` — so the chord was the only opener running on a stale capture.

Reachable: preview a text file, select lines 3–5, right-click → menu opens with the selection row. Press Escape — the Dropdown's Escape handler calls `returnFocus()`, which is `previewMenu.anchorOwner`, i.e. the preview body div, so focus lands on the one element that can receive this chord. Click once in the preview to collapse the selection. Press Shift+F10: the menu reopens still offering **Add selection to conversation**, and taking it inserts `@path (L3-L5)` around text nobody has selected. `restoreSelection` then no-ops on the detached range, so the only symptom is the wrong block in the composer.

Fixed by routing the chord through the same capture, folding the three copies of those two lines into one `captureMenuSelection`. No unit test: `packages/web/vitest.config.ts` is `environment: "node"` and states "no DOM tests", so there is no harness in this package that could exercise it. ✅ Fixed in 5b495b5b

**2. The editor's dark-mode selection colour is dropped from the built CSS** — `packages/web/src/styles.css:704` (pre-fix) · *should fix*

`.code-editor textarea::selection { @apply bg-sky-200 dark:bg-sky-500/40; }`. The project's `dark` variant is `@custom-variant dark (&:where(.dark, .dark *))` (`styles.css:13`), so the dark half compiles to `…::selection:where(.dark, .dark *)` — invalid after a pseudo-element, and dropped silently. Built from this branch before the fix:

```
.code-editor textarea::selection{background-color:var(--color-sky-200)}
```

— one rule, no dark counterpart. So selecting text in the editor in dark mode paints the light `sky-200`, not the translucent `sky-500/40` that was written, over transparent text on a near-black surface. The comment at `styles.css:679` documents exactly this trap and works around it with `html.dark` for `.line::before`; the rule twelve lines further down did not get the same treatment. Split it the same way; the rebuilt bundle now carries `html.dark .code-editor textarea::selection{background-color:color-mix(in oklab,var(--color-sky-500) 40%,transparent)}`. ✅ Fixed in 5b495b5b

**3. `menuSelection` keeps a detached `Range` after the preview navigates** — `workspace-browser.tsx:555` · *nit*

A session switch clears it in the render-phase reset, so nothing survives that. Opening another file does not, so the closed file's rendered `<pre>` subtree stays reachable through the Range's boundary nodes until the next context menu opens. Bounded, self-healing, and not worth a state write on every navigation. ⬜ Left for the author

**4. The zh docs still call the panel 工作区面板** — `packages/docs/content/web-app.zh.md:60` (twice), and the `### 文件面板` heading below it · *nit*

The zh label became 文件浏览 in this PR. The English side already read "Files panel" in `web-app.en.md` and `server-api.en.md`, so only the Chinese copy drifted. Which of the two names is canonical is a naming call, so left alone. ⬜ Left for the author

**5. `belowTrigger`'s left-anchored branch does not clamp the panel's far edge** — `components/ui/tooltip.tsx:78` · *nit*

`left: Math.max(rect.left, VIEWPORT_MARGIN)` keeps the near edge on screen; the panel is `w-max max-w-64` (256px), so a trigger whose centre sits just left of `innerWidth / 2` can run past the right edge once the window is under ~512px. Nothing reaches it today — every `bottom` caller this PR adds sits at the right edge of its row and takes the right-anchored branch — and the panel is `position: fixed`, so it adds no scrollable overflow in either branch. ⬜ Left for the author

**6. The tree toggle is the one toolbar control still on a native `title`** — `workspace-browser.tsx:2110` · *nit*

It sits in the same row as Refresh and Upload, which now use `Tooltip`, so a pointer crossing the toolbar gets two different tooltips with two different delays and only one of them on keyboard focus. A `Tooltip` label can swap Show/Hide the same way the `title` does. Visual call, left alone. ⬜ Left for the author

### Checked and correct

Recording these because they were the parts most likely to be wrong.

- **The single rAF is sufficient.** `addSelectionToChat` calls `onInsertReference` before `restoreSelection`, and `insertAtComposerCaret` registers its rAF synchronously, so the focus callback is registered first and rAF callbacks run in registration order. React 19 gives a click-handler update `SyncLane` and flushes it in a microtask, and microtasks drain before the frame's rAF callbacks, so both callbacks see the committed textarea. `Dropdown` returns focus only on Escape (`dropdown.tsx:205`), never on an item click, so nothing re-steals focus after the restore.
- **The cloned Range survives the re-render.** `CodeSurface` renders the highlighted path through `dangerouslySetInnerHTML` with an unchanged `__html` string and the fallback through the same keyed `.line` spans, so the commits caused by `setMenuSelection` and `setAnchor` mutate no DOM under the range. `isConnected` does cover a replaced node — React removes it from the tree first.
- **`placement="right"` is unchanged**: same `top` clamp, same `left`, same `-translate-y-1/2`, and `app-layout.tsx` (the only other caller) passes no `placement`. A `style` carrying `left: undefined` or `right: undefined` is fine — React drops undefined style values. Esc / scroll / resize are keyed on `position !== null` alone, so dismissal is placement-independent.
- **Accessible names.** The upload control's name is on the `<input>` — `HiddenFileInput` spreads props onto it, and `aria-label` outranks a text-less wrapping `<label>`; `GlyphIcon` is `aria-hidden`; `onFocus` on the Tooltip's span catches the hidden input's focus because React's `onFocus` is `focusin`. No trigger in the panel carries both a `title` and a `Tooltip`.
- **The copy button keeps `copy-button.tsx`'s contract**: glyph swap via `CopyCheckGlyph`, tooltip flips to 已复制 while `aria-label` stays `S.chat.copyCode`, and `CopiedStatus` is a sibling of the `Tooltip` — outside the button, so it joins neither the visible label nor the accessible name, and `sr-only` is absolute so it is not a flex item in the header row.
- **i18n parity**: four new `files.*` keys in both dictionaries with the same shape, and the `editorWrap` → `wrapLines` rename applied on both sides. `EDITOR_WRAP_KEY` is still registered in `lib/install-scope.ts:166`, so the localStorage-scoping registry did not drift.
- **Tests**: every new assertion fails at the base commit — `pathReference`, `insertAtCaret` and `selectionBlock` do not exist there, and the wrap block asserts the inverse of `parseEditorWrap` (`parseWrapLines(null) === true` vs `parseEditorWrap(null) === false`).
- **The `## Compatibility` section is true of the shipped `parseWrapLines`**: only `0` / `false` / `off` / `no` return false, `null` and everything else return true, and the writer has only ever stored `"1"` / `"0"` — so anyone who used the old toggle keeps their answer and only an unset browser picks up the new default.

### Merge outlook

- `git merge-tree --write-tree HEAD origin/refactor/shared-file-tree` (#685 at `c5c10bc4`) → clean, tree `7d5873e9`, and coherent rather than merely textual: the merged `workspace-browser.tsx` takes #685's move of `TreeToggle` to `components/ui/file-tree`, the merged `lib/workspace-tree.ts` keeps this PR's three composer helpers while losing `subtreeEnd` / `treeKeyStep`, and the merged `test/workspace-tree.test.ts` imports neither — no dangling reference in either direction.
- `origin/feat/dock-launcher-icons` (#680) also merges clean (tree `a1d7795`). It rewrites a different region of `icons.tsx` and edits `web-app.{en,zh}.md` around line 85, away from the doc lines in finding 4.

### Left for you

The soft-wrap default flip is a change to data already on disk, which is the user's call rather than the PR's. The decision itself looks sound and is documented — anyone who never touched the toggle now gets wrapping — but two loose ends:

1. This batch *does* carry compatibility handling (`parseWrapLines` keeps reading `penguin.files.editorWrap` under its old name and old spellings), and the repo convention asks such a batch for a dedicated `changelog/unreleased/2026-09-10-backward-compatibility.md` pair that other entries reference. Either add it, or confirm that an in-entry `## Compatibility` section is enough for a rename-avoidance this small.
2. Nothing names how long the tolerant parse stays, who removes it, or what has to be true first. If the answer is "forever, it is the key's format", say so at `workspace-tree.ts:573` so the next reader does not re-open it.

### Verification

```
Node 24, worktree ../penguin-harness-wt/rv-682, base 1a26d1cd.

$ pnpm --filter @prismshadow/penguin-web typecheck    # tsc --noEmit, no diagnostics
$ pnpm lint                                           # oxlint, no findings
$ pnpm format && pnpm format:check                    # All matched files use Prettier code style!
$ git diff --check                                    # clean

$ npx vite build --mode production                    # before and after the CSS fix, to read the
                                                      # emitted ::selection rules out of the bundle
  before: .code-editor textarea::selection{background-color:var(--color-sky-200)}
  after:  .code-editor textarea::selection{background-color:var(--color-sky-200)}
          html.dark .code-editor textarea::selection{background-color:#00a5ef66}
          html.dark .code-editor textarea::selection{background-color:color-mix(in oklab,var(--color-sky-500) 40%,transparent)}

vitest on the build box vps3 (this session's standing rule), before and after the fixes:

$ vps3-vitest.sh ... -- pnpm --filter @prismshadow/penguin-web exec vitest run test/workspace-tree.test.ts
  Test Files  1 passed (1)        Tests  44 passed (44)

$ vps3-vitest.sh ... -- pnpm --filter @prismshadow/penguin-web test
  Test Files  129 passed (129)    Tests  1825 passed (1825)

packages/web/e2e not run — Playwright is not installed on the build box.
```

hiyouga:
## Preview / Source, and the two actions that belong to the body

Pushed as `eba986a7`, on `origin/main` merged in.

### The toggle names what you are looking at

`Rendered` / 渲染视图 named how the view was produced. It now reads **Preview / Source**, 预览 / 源码.

### Copy and Edit float over the file

Both act on the body underneath them — the text on screen. The title row names the *file* and
carries what takes it away: the view toggle, wrap, the external link and download. That row is also
the one the panel can be narrower than, so it is the row worth emptying.

They sit in a small bordered pill at the top-right of the content, and they are **always drawn**
rather than revealed on hover. A hover-only control is no control at all on a touch screen, and
Edit has no other way in from here. The cost is that the pill covers the right end of the first
line or two; say the word if you would rather trade that for a hover reveal.

One thing that is not cosmetic: the pill's containing block is a **new positioned wrapper**, not the
scroller. An absolute box inside a `static` scroller escapes to the initial containing block and
gives the whole shell a second scrollbar — `styles.css` records that having shipped three times.

Rendered at the real type sizes in both themes before pushing, with the header row and the pill in
place over a numbered source view.

### Verification

```
pnpm typecheck (23 projects)   → exit 0
pnpm lint (oxlint)             → exit 0
pnpm format && pnpm format:check → clean
git diff --check               → clean
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## One header row, whole-Workspace search, and rename / move / delete

Pushed as `cedc4f81` and `67e0d0b5`.

### The panel has one header row

The directory breadcrumbs and the filename title were two rows saying one thing. They are now one
strip naming whatever is open, fitted tail first, so the filename is the last thing to go and the
leading directories collapse into a single `…` ahead of it.

What was in those rows moved by **what it acts on**:

| row | holds | because |
| --- | --- | --- |
| path row | view toggle, download | they belong to the file |
| floating pill | wrap, copy, edit, open in a new tab | they act on the text under them |
| tree pane header | search, refresh, upload | they are the panel's own |

The editor carries the pill too — it had no other way to reach wrap once that left the row.

Two smaller ones in the same commit: a right-click inside selected text no longer clears the
selection (the menu already read the range before opening, since focusing the panel can collapse
it; it now puts it back as well), and both context menus size to their own rows instead of a fixed
`w-48`.

### Search covers the whole Workspace

`GET …/files/search?q=` walks it breadth-first on the server. The old box filtered the rows the lazy
tree happened to have loaded, so a match was reachable only if every directory above it was already
open. Results are a flat list naming each hit's full path, shallowest first — a hit can live
somewhere the tree has never listed, and the path is the part the query did not say.

The walk stops at 200 hits or 20000 entries and says so in the list. Breadth-first is what makes
that degrade into "the shallowest matches" rather than "whatever was enumerated first".

### Rename, move and delete, behind the same precondition

Rename and move are one action: both are the same write of a new path. They and delete read the
file's **current** version when the dialog opens, and the server refuses with 409 `file_changed` if
the Agent rewrote the file while the question was on screen. Read at the dialog, not at the preview:
what this guards is the window in which you are deciding, and a marker from five minutes ago would
refuse actions nobody needed warning about. Until the version lands the action is not offered at
all — an unconditional move or delete is the thing this exists to prevent.

A directory has no version marker, so none of the three is offered for one.

**The destination of a move is never replaced**, and that holds against a write landing on it
mid-operation. The move is `link` + `unlink`: the kernel fails it `EEXIST` rather than overwriting,
and the inode is preserved, so what lands is what was checked. `rename` behind an `lstat` only
narrows that window. Across filesystems it is `copyFile` with `COPYFILE_EXCL`, which keeps the same
guarantee; only a mount with no hard links at all (FAT, some network and container mounts) falls
back to `lstat` + `rename`, where the window reopens because nothing atomic is left.

Two limitations worth stating rather than discovering: a **symlink** at the final segment is refused
outright, so the panel cannot rename or delete one, and a directory cannot be moved or deleted at
all.

### Verification

```
pnpm typecheck (23 projects)      → exit 0
pnpm lint (oxlint)                → exit 0
pnpm format && pnpm format:check  → clean
git diff --check                  → clean
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**, with
31 tests in `workspace-files.test.ts` covering the move and delete preconditions, `..` traversal at
both ends, a symlinked destination directory, a final-segment symlink, search confinement and
symlink cycles, the hit cap, and name-not-path matching. Each guard was checked by breaking it and
watching the right tests go red.

hiyouga:
## The right-click was not what cleared the selection

Pushed as `e87846ea`. The earlier restore-the-range fix could not work, and here is why.

### What was actually happening

`CodeSurface` handed React a fresh `{ __html: … }` object on every render. **React compares that
prop by object identity and never looks at the markup inside it** — so an identical string in a new
object is still a change, and every render destroyed and rebuilt every node in the block.

A live `Range` anchored in those nodes collapses when they are replaced. That is why restoring it a
frame later did nothing: by then the range was already collapsed, not merely unselected. It is also
why the guard for detached nodes missed it — when the children go, the boundary moves up to the
parent, and the parent is still connected.

### How it was found

Reasoning about it kept producing wrong answers, so I built the real `CodeSurface`, `Dropdown` and
`useRowContextMenu` into a page and drove a real right-click at it in Chromium. Two earlier guesses
died there first: focusing a menu item does **not** clear a selection, and neither does focusing the
container that holds it. The instrumented run showed the real thing:

```
capture: enter          "nst line3 = \"alpha b"
after menu open         "nst line3 = \"alpha b"
rAF: range.collapsed=true startConnected=true sameLineNode=false htmlChanged=false
FINAL                   (collapsed)
```

`htmlChanged=false` with `sameLineNode=false` is the whole diagnosis: same markup, new nodes.

An unhighlighted block never showed it, because that path renders real React children instead of
`innerHTML` — which is why a plain-DOM reproduction kept passing.

### The fix

Both `dangerouslySetInnerHTML` payloads are memoised, so identical markup keeps its nodes.

This was never only the Files panel. **The message stream's code blocks had it too**, where it also
meant the browser re-parsed every highlighted block on every render of a streaming transcript.

`test/inner-html-stability.test.ts` pins both call sites. Its first assertion fails against the code
this commit fixes, checked by stashing the fix and running it.

### Verification

```
pnpm typecheck (23 projects)      → exit 0
pnpm lint (oxlint)                → exit 0
pnpm format && pnpm format:check  → clean
git diff --check                  → clean
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## Cancel and Save become icons, so the header stops moving

Pushed as `b90d113c`.

They were the last text buttons in the path row, and a text button is taller than the icon buttons
beside it — so pressing Edit pushed the row down, and leaving the editor pulled it back up.

Measured in Chromium against the **built stylesheet**, not from the class names:

| state | row height |
| --- | --- |
| at rest | 37.00px |
| editing, text buttons | 39.00px |
| editing, icon buttons | 37.00px |

`Button size="sm"` is an 18px line box plus 9px of padding plus 2px of border. The icon buttons are
a flat `h-6`, which is 27px at this app's 18px root. Hence the two pixels.

Save keeps no primary tint: it opens a confirmation, and that dialog's own button is where the
weight belongs. While a save is in flight the check becomes a spinner and the tooltip says so,
which is the same pattern the upload control already uses.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo, `FORCE_COLOR=0 pnpm -r test` → **exit 0**.

hiyouga:
## A quoted selection is a chip now, not a paste

Pushed as `59ebbf74`.

Adding a selection dumped the whole fenced block into the composer, so the sentence being written
ended up buried under the thing it pointed at. It is a **chip** now — above the text body, beside
the attachment chips, naming the file and the lines it covers. The same shape `/agent` and `/skill`
already stage a pick in.

The quotation never enters the textarea. The composer holds it as a reference, the chip removes it,
and on send it goes into the message **in front of** the typed text, byte for byte what it was
before. Nothing the model receives has changed.

### Three decisions worth naming

- **`@path` still splices inline.** That one is a word in a sentence, not a block, and weaving it
  into a half-typed line is the whole point of it. Only the selection — the thing that was actually
  pasting 原文 — became a chip. Say the word if you want the whole-file mention to become a chip too.
- **Quotation marks, not the paperclip.** Nothing is attached: the file stays where it is and a few
  of its lines are quoted. Borrowing the attachment glyph would claim otherwise.
- **Held in composer state, not the draft cache** — exactly like attachments. What a reference
  quotes is a file on disk that may have moved on by the time a stale draft is reopened, so it
  belongs to this composer's life rather than to the text's.

A reference also makes a message sendable on its own, the way attachments do, and it rides a
steering message too.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo → **exit 0** (web 1863, server 1820, cli 394,
desktop 183).

hiyouga:
## Files and folders stage like skills, in the composer's own chip row

Pushed as `a703bda0`.

Two changes, and the second is the one you asked for last.

### One channel for everything the panel contributes

Adding a file or a folder spliced `@path` into the draft at the caret, while a selection was staged
as a chip — two ways to hand over the same kind of thing. All three now go through one path: the
entry becomes a reference carrying its own `kind`, and the chip wears a folder, a file or a
quotation mark accordingly.

### The chips moved inside the box

They were in the stack **above** the composer beside the attachment chips. They now sit **inside**
it, in the row `/agent`, `/model`, goal mode and the selected skills already share, wearing that
row's look — `bg-gray-100`, mono, 13px glyph, the same remove button.

That row is for things staged alongside what is being typed, which is exactly what these are. The
attachment row is for payload that rides the message; a reference is a pointer, not a payload.

### What that removed

With nothing splicing into the draft any more, `insertAtCaret` and `InsertLayout` lost their only
callers. They are gone, along with `ComposerControl.insertAtCaret`, the chat page's callback that
reached it, and the panel's `onInsertReference` prop. None of it ever shipped — this branch
introduced it — so there is no compatibility question, just less surface.

`workspace-tree.test.ts` loses the caret-splicing tests with it and gains one pinning that a file
and a directory travel the same way a quotation does.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo → **exit 0**.

hiyouga:
## Smaller chips, `:10-12` ranges, and a filename the row cannot squeeze

Pushed as `2e8d46df`. Three complaints, and the third turned out to be a real bug.

### The chips were at body size

They were `text-sm` — **15.75px** at this app's 18px root, which is body size for something that is
not body text. Every chip in that row is `text-xs` (13.5px) now: goal mode, the `/agent` and
`/model` targets, the skills, and the file references. One row, one size.

### Ranges read as `:10-12`

They were spelled out — `第 10-12 行` in one language, `L10-L12` in the other — and appended with a
space. They are now `:10-12`, joined straight onto the name, which is the form an editor and a stack
trace already use.

Because it needs no translating, **the dictionary key that held it is gone**: a number range was
never language. The quoted block's header follows the same form, so the chip and what the model
receives agree.

### The filename bug

The view toggle really could squeeze the name into an ellipsis, and the markup explains why: the
leading directories were `shrink-0` while the name was the only element that could truncate. So
every pixel `visibleCrumbSegments` estimated wrong — and it estimates, at 7px per character — was
paid by the name.

That is inverted now. The directories shrink and ellipsize first, since the tree beside them says
the same thing. The name does not shrink at all, and when it alone outruns the row it is **split at
its last dot** so the stem ellipsizes and the extension survives.

A leading dot is not an extension, so `.gitignore` stays whole rather than ellipsizing to nothing —
pinned by a test, along with `archive.tar.gz` splitting at the right dot.

### Verification

```
```

vitest on `vps3-test` (never locally): whole repo → **exit 0**.

hiyouga:
## Highlighting moved to a worker; both size caps removed

`ts` highlighting was never broken — the file you were looking at was simply over the cap.
`workspace-browser.tsx` is 119,601 bytes and `HIGHLIGHT_LIMIT` was 64KB, so it rendered plain.

The cap existed for a real reason. Tokenizing is linear in the input and runs to completion once
it starts, measured sequentially on this bundle's JS regex engine over TypeScript:

| Input | Tokenize |
| --- | --- |
| 64KB | 338ms |
| 128KB | 527ms |
| 256KB | 1059ms |
| 384KB | 1606ms |
| 512KB | 2154ms |

~4.2ms/KB. Raising the cap to 500KB would have bought a two-second frozen page, so the cap moved
rather than the number: `highlighter-core.ts` holds the engine, `highlighter.worker.ts` runs it,
and `highlighter.ts` is a client that correlates requests by id.

**Both ceilings are gone** — the source view's 64KB and the editor's 32KB — and `TEXT_PREVIEW_LIMIT`
went 256KB → 1MB. Markdown *rendering* keeps a 64KB ceiling under the new name `MD_RENDER_LIMIT`,
because remark plus the React tree it produces is main-thread work a highlighting worker does
nothing about; past it a file opens in the source view and the reader can switch.

### Measured on a built bundle, real browser

A Playwright probe highlighting generated TypeScript while counting `requestAnimationFrame` ticks:

| Input | Tokenize | Main thread | DOM insert |
| --- | --- | --- | --- |
| 256KB | 2968ms | 60 fps | 71ms |
| 512KB | 5722ms | 60 fps | 133ms |
| 1MB | 11007ms | 60 fps | 395ms |

The page holds a full 60 frames a second for the whole pass. The only main-thread cost left is
inserting the result, which at the new 1MB ceiling is a single 395ms hitch.

### Bundle

`worker: { format: "es" }` was required in `vite.config.ts` — Vite's default worker format is IIFE,
which cannot code-split, and the worker dynamically imports its themes and each grammar. The build
fails outright without it.

The client's fallback uses `await import("./highlighter-core")` rather than a static import, so the
engine stays out of the entry chunk:

```
dist/assets/highlighter-CuXbPCQx.js            1.00 kB   the client
dist/assets/highlighter.worker-CL6f9Fzw.js   166.95 kB   the worker, engine inlined
dist/assets/highlighter-core-BZNoCu9s.js     146.60 kB   fetched only on the fallback path
```

The entry chunk contains no Shiki (`grep -c shiki index-*.js` → 0), and the probe confirmed only
the worker chunk loads when a worker is available. `test/highlighter-worker.test.ts` pins that
arrangement: a static import of the engine outside the worker would silently double-bundle it
without failing a build.

### Verification

```
pnpm --filter @prismshadow/penguin-web typecheck   pass
pnpm lint                                          pass (exit 0)
pnpm format:check                                  pass
pnpm --filter @prismshadow/penguin-web build       pass
git diff --check                                   clean
vitest (on vps3): 136 files, 1862 tests, all pass
```

PR #685, opened 2026-09-11
refactor(web): draw the plugin library's file browser with the Workspace tree

hiyouga:
The plugin library's detail Modal had its own file browser, hand-rolled from plain buttons: a
text triangle (`▾`/`▸`) per group, a middot per file, a border between every row, and a nested
file carrying its whole relative path in its name — `reference/case-stud…`, nine times over for
the humanizer plugin. The Workspace panel next to it already has a proper tree. This makes them
one component.

## What moved

`components/ui/file-tree.tsx` (`FileTree`) is the Workspace tree's view, moved rather than
rewritten. It owns the presentation and the interaction:

- row markup, the indentation scale, the folder / folder-open / page glyphs (`GlyphIcon` over
  the paths already in `group-list.tsx` and `icons.tsx`, sizes from `lib/icon-scale.ts` by role);
- the selected, hover and focus treatment, and the drop-target ring;
- the WAI-ARIA `tree` semantics — `role="tree"` / `role="treeitem"` / `role="group"`,
  `aria-level`, `aria-posinset`, `aria-setsize`, `aria-expanded`, `aria-busy`;
- the roving tab stop and the arrow-key walk;
- the `data-tree-path` / `data-tree-kind` attributes on the row element;
- the retained-subtree pass, so a collapsing directory animates out.

**Data loading deliberately did not move.** The two callers differ there and forcing one model
on both would be the wrong seam: the Workspace lists one directory per level as it is opened,
the plugin browser groups a single flat listing it already holds. Each caller keeps its own
fetching and hands `FileTree` flattened rows.

`lib/file-tree.ts` holds the DOM-free part — the row shape, `subtreeEnd`, `treeKeyStep` — moved
out of `lib/workspace-tree.ts` together with its test cases (`test/file-tree.test.ts`).
`treeKeyStep`'s step out of a row now takes the parent from the nearest row above it at a
shallower depth, instead of slicing the focused row's path. The plugin tree's top-level rows
are whole paths (`skills/humanizer`), and a path-derived parent would have focused `skills`,
which is not a row. The one added test case covers exactly that.

`WorkspaceTreeView` shrank from 335 lines to 77 and keeps only what is Workspace-specific: the
size and modified time in a row's tooltip and after its name, the upload directory standing in
for a selection while no file is open, and the three things an empty row list can mean.

`plugin-detail.tsx` gained `pluginTreeRows`, which turns `groupPluginFiles`'s groups into rows
with real nesting, so a skill's `reference/` is a directory row instead of a slash inside nine
file names. The grouping itself (skill order, SKILL.md first, hooks last) is unchanged.

## Not in scope: the benchmark case browser

`benchmark-case-browser.tsx` was left alone. It is not a tree: `MaterialGroup` drills down one
directory at a time (`setPath`) with a breadcrumb row, refetching per level from a per-material
endpoint and keying rows to the listing they came from. Putting it on `FileTree` would mean
replacing the drill-down with in-place expansion and building a third listings-accumulator to
feed it — a behaviour change, not a presentation unification. Half-migrating it (depth-0 rows
whose `aria-expanded` is a lie) would be worse than leaving it.

## Interaction with open PRs

- **#682** (`feat/file-browser-context-menu`) resolves a right-clicked row by reading
  `data-tree-path` off the event target. Those attributes are still on the same element, spelled
  the same way, alongside `data-tree-kind`; the surrounding markup of `workspace-browser.tsx` is
  untouched (this PR changes one import line there). Verified in the browser: every
  `role="treeitem"` in the Workspace panel carries `data-tree-path`.
- **#684** adds `components/ui/tooltip.tsx`; this adds `components/ui/file-tree.tsx` — same
  directory, different file, no overlap. **#683** (`models-page.tsx`) does not touch anything here.

## One unrelated paragraph: a wrong claim in the frontend skill

`.agents/skills/penguin-harness-frontend/SKILL.md`'s "Every user-facing string is bilingual"
section credited `test/placeholders-parity.test.ts` with checking that both dictionaries
interpolate the same placeholders. It does not: that test pins the Agent Prompt tab's
`agent.placeholders` list against the tokens in core's real `defaultSystemConfig().system_prompt`,
in first-appearance order, for both dictionaries (it was written after `{{SHELL}}` reached the
prompt but not the list). Nothing in `packages/web/test/` compares one dictionary's interpolations
against the other's.

The sentence now describes the guard that exists — the `const en: Strings` type — and names the
gap it leaves: a function-valued string need not use the parameter it is handed, so
`(n: number) => "items"` typechecks while the other side interpolates `n`. One sentence, in a
section this PR is otherwise not touching, and no new test for the gap.

## Verification

Node v24.18.0, from the worktree `../penguin-harness-wt/filetree`.

```
(clean)

$ prettier --check .
Checking formatting...
```

Vitest on the remote host `vps3-test`:

```

 Test Files  130 passed (130)
      Tests  1819 passed (1819)
   Duration  2.49s
```

(129/1816 before this branch; +1 file, +3 cases.)

### Looked at, not just typechecked

Both dists built from this branch and from its merge base, served in turn from the same scratch
data root (`PENGUIN_WEB_DIST`, `PENGUIN_SEED_ADMIN_PASSWORD`, port 8931), driven with Playwright.

| | before | after |
| --- | --- | --- |
| plugin detail, light | `plugin-detail-light-before.png` | `plugin-detail-light-after.png` |
| plugin detail, dark | `plugin-detail-dark-before.png` | `plugin-detail-dark-after.png` |
| Workspace panel, light | `workspace-panel-light-before.png` | `workspace-panel-light-after.png` |
| Workspace panel, dark | `workspace-panel-dark-before.png` | `workspace-panel-dark-after.png` |

The two Workspace pairs are **byte-identical** (`cmp` reports no difference), which is the
check that matters for "the reference comes out no worse than it went in".

25 DOM assertions over the built app, all passing: `role="tree"` with a name; 12 `treeitem`
rows; `data-tree-path` / `data-tree-kind` / `aria-level` / `aria-posinset` / `aria-setsize` on a
row; exactly one tab stop in each tree; ArrowDown / ArrowUp / Home moving focus; ArrowLeft
collapsing an open directory and stepping nowhere from a collapsed top-level row; ArrowRight
reopening it and stepping into it; Enter opening the focused file; `.tree-group-close` appearing
on a collapse and being dropped after the shrink; and, in the Workspace panel, file sizes,
the `path · size · time` tooltip and click-to-open still in place.

No e2e spec selects on `treeitem`, `data-tree-path`, `role="tree"`, or the plugin detail Modal
(`grep` over `packages/web/e2e/`), so nothing there needed changing.

No user-facing string was added: the tree's accessible name reuses `S.files.treeLabel`.

### DIFF
--- a/.agents/skills/penguin-harness-frontend/SKILL.md
+++ b/.agents/skills/penguin-harness-frontend/SKILL.md
@@ -134,9 +134,11 @@ the caret, the close cross or the collapse chevron.
 ## Every user-facing string is bilingual
 
 Two dictionaries: `src/lib/strings.ts` is zh (and defines the `Strings` type), `src/lib/strings-en.ts`
-is en and is typed `const en: Strings`. Adding a key to one and not the other is a **type error**,
-not a runtime surprise — and `test/placeholders-parity.test.ts` checks that both sides interpolate
-the same placeholders. Add both, in the same shape, in the same PR.
+is en and is typed `const en: Strings`. That type is the whole guard: a key added to one and not
+the other, or a signature that changed on one side, is a **type error** rather than a runtime
+surprise. What it cannot see is whether a function-valued string uses the parameter it is handed —
+`(n: number) => "items"` typechecks while the other side interpolates `n`. Add both, in the same
+shape, in the same PR.
 
 `S` is a live binding swapped on locale change, so read it at render time; never hoist `S.x.y` into
 a module-level constant.
--- a/changelog/unreleased/2026-09-11-shared-file-tree.md
+++ b/changelog/unreleased/2026-09-11-shared-file-tree.md
@@ -1,0 +1,32 @@
+# The plugin library's file browser draws the Workspace's tree
+
+- **Date:** 2026-09-11
+- **Type:** refactor
+- **Scope:** `web`
+- **PR:** [#685](https://github.com/Prism-Shadow/penguin-harness/pull/685)
+
+[中文版](2026-09-11-shared-file-tree.zh.md)
+
+The Workspace panel's directory tree moved into `components/ui/file-tree.tsx`, and the plugin
+detail Modal's file browser was put on it. That browser's rows had been plain buttons — a text
+triangle per group, a middot per file, a border between every row, and a nested file carrying
+its whole relative path in its name; they are now the rows the Files panel draws, with a
+chevron, a folder or page glyph, indentation per level and the height animation on opening and
+closing a directory. A skill's `reference/` became a directory row of its own instead of a
+slash inside nine file names.
+
+## Details
+
+- `FileTree` owns the row markup, the indentation scale, the glyphs, the selected, hover and
+  focus treatment, the WAI-ARIA `tree` semantics (`role`, `aria-level`, `aria-posinset`,
+  `aria-setsize`, `aria-expanded`), the roving tab stop and the arrow-key walk, and the
+  `data-tree-path` / `data-tree-kind` attributes a caller resolves a pointer event through.
+  Callers keep their own data loading: the Workspace panel lists one directory per level as it
+  is opened, the plugin browser groups the single listing it already holds.
+- The row shape, `subtreeEnd` and `treeKeyStep` moved from `lib/workspace-tree.ts` to
+  `lib/file-tree.ts`, with their test cases. Stepping out of a row now takes the parent from
+  the row above it rather than from its path, so a tree whose top-level rows are whole paths
+  (`skills/<name>`) steps onto a row that exists.
+- `WorkspaceTreeView` kept what only the Workspace has: the size and modified time in a row's
+  tooltip and after its name, the upload directory standing in for a selection while no file is
+  open, and the three things an empty row list can mean.
--- a/changelog/unreleased/2026-09-11-shared-file-tree.md
+++ b/changelog/unreleased/2026-09-11-shared-file-tree.md
@@ -17,12 +17,12 @@ slash inside nine file names.
 
 ## Details
 
-- `FileTree` owns the row markup, the indentation scale, the glyphs, the selected, hover and
-  focus treatment, the WAI-ARIA `tree` semantics (`role`, `aria-level`, `aria-posinset`,
-  `aria-setsize`, `aria-expanded`), the roving tab stop and the arrow-key walk, and the
-  `data-tree-path` / `data-tree-kind` attributes a caller resolves a pointer event through.
-  Callers keep their own data loading: the Workspace panel lists one directory per level as it
-  is opened, the plugin browser groups the single listing it already holds.
+- The row markup, the indentation scale, the glyphs, the selected, hover and focus treatment,
+  the WAI-ARIA `tree` semantics (`role`, `aria-level`, `aria-posinset`, `aria-setsize`,
+  `aria-expanded`), the roving tab stop, the arrow-key walk and the `data-tree-path` /
+  `data-tree-kind` attributes a caller resolves a pointer event through all went into
+  `FileTree`. Data loading stayed with the callers: the Workspace panel lists one directory
+  per level as it is opened, the plugin browser groups the single listing it already holds.
 - The row shape, `subtreeEnd` and `treeKeyStep` moved from `lib/workspace-tree.ts` to
   `lib/file-tree.ts`, with their test cases. Stepping out of a row now takes the parent from
   the row above it rather than from its path, so a tree whose top-level rows are whole paths
--- a/changelog/unreleased/2026-09-11-shared-file-tree.zh.md
+++ b/changelog/unreleased/2026-09-11-shared-file-tree.zh.md
@@ -1,0 +1,26 @@
+# 插件库的文件浏览器改用 Workspace 的目录树
+
+- **Date:** 2026-09-11
+- **Type:** refactor
+- **Scope:** `web`
+- **PR:** [#685](https://github.com/Prism-Shadow/penguin-harness/pull/685)
+
+[English](2026-09-11-shared-file-tree.md)
+
+Workspace 面板的目录树移入 `components/ui/file-tree.tsx`，插件详情 Modal 的文件浏览器改由它绘制。
+该浏览器此前的行是朴素按钮——每个分组一个文本三角、每个文件一个间隔点、行与行之间一条分隔线，嵌套
+文件还把整段相对路径写进名字里；现在它们就是文件面板画的那种行：折叠箭头、文件夹或文件图形、按层级
+缩进，展开与收起目录时带高度过渡。Skill 的 `reference/` 成为一个独立的目录行，不再是九个文件名里的
+一条斜杠。
+
+## 细节
+
+- `FileTree` 负责行的标记、缩进刻度、图形、选中与 hover、focus 的表现，WAI-ARIA `tree` 语义
+  （`role`、`aria-level`、`aria-posinset`、`aria-setsize`、`aria-expanded`）、轮转式 tab 落点与方向键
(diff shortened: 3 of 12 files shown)


hiyouga:
## Review

I diffed the old `workspace-tree-view.tsx` and `workspace-tree.ts` line by line against
`components/ui/file-tree.tsx` + `lib/file-tree.ts`: the move is faithful — keyboard walk,
retained-subtree animation with its `CLOSE_FALLBACK_MS` timer, render-phase `closing` capture,
roving tab stop, `scrollIntoView`, ARIA attributes, glyphs, indentation, drop ring and the
`data-tree-*` attributes all came across unchanged. Of the 22 assertions deleted from
`workspace-tree.test.ts`, 21 have an exact counterpart in `file-tree.test.ts` and the 22nd — the one
that also pinned `flattenTree`'s row paths — is subsumed by the stronger path+depth assertion at
`packages/web/test/workspace-tree.test.ts:115`. Nothing was silently dropped. The
seam is a good one — fourteen props and not a single boolean flag, so it is one component, not two.
Ship after the fixes below, which are small. `git merge-tree HEAD origin/feat/file-browser-context-menu`
(61afc46c) auto-merges clean; details under Verification.

### Findings

**1. The plugin Modal mounts an empty `role="tree"` while it has nothing to draw** — `packages/web/src/features/plugins/plugin-detail.tsx:259` · *should fix*

`groups` is `[]` until `getPluginFiles` resolves, so `rows` is `[]` and `FileTree` renders
`<div role="tree" aria-label="File tree">` with `children` undefined — an empty tree beside the
`SkeletonList`. The same node stays for the whole session when the request fails (`files` never
leaves `null`) and for a plugin that ships no `skills/` or `hooks/` files. A `tree` with no owned
`treeitem` is ARIA-invalid (`aria-required-children`), and the pre-PR code rendered nothing there —
`groups.map(…)` over an empty array. Gated the mount on `rows.length > 0`.
✅ Fixed in c5c10bc4

**2. `TREE_INDENT_PX` was exported from the DOM-free module for no consumer** — `packages/web/src/lib/file-tree.ts:12` · *nit*

`rg TREE_INDENT_PX` finds the definition and two uses, both in `components/ui/file-tree.tsx`. That
module's own header says it holds "the shape of a row, where a row's subtree ends in the flat row
list, and the keyboard step" — a pixel indent is none of those, and the other two pixel constants
(`TREE_PAD_PX`, `CHEVRON_COL_PX`) are already in the component. It was private to the view file
before this PR. Moved back, import narrowed.
✅ Fixed in c5c10bc4

**3. The changelog's first detail bullet describes the codebase, not the change** — `changelog/unreleased/2026-09-11-shared-file-tree.md:20` · *nit*

"`FileTree` owns the row markup, … the roving tab stop and the arrow-key walk, and the
`data-tree-path` / `data-tree-kind` attributes" is exactly the shape `changelog/README.md` rules out
("**Do not describe the current state of the codebase.** … Write what the change did at the time"),
and it is the bullet most likely to drift — the prop list will grow. Bullets 2 and 3 are already in
past tense. Reworded both languages to match them.
✅ Fixed in c5c10bc4

**4. A 135-column line left by a re-wrap** — `packages/web/test/workspace-tree.test.ts:3` · *nit*

Dropping "the keyboard step over those rows," from the header comment joined two lines into one.
Prettier does not reflow comments, so `format:check` never saw it. Re-wrapped, confined to the two
lines #682 does not also edit so the merge stays clean.
✅ Fixed in c5c10bc4

**5. The roving tab stop's fallback moved, which the screenshot comparison cannot see** — `packages/web/src/features/chat/workspace-tree-view.tsx:52` · *nit*

`WorkspaceTreeView` now folds `currentDir` into the `selectedPath` it hands down, and `FileTree`
uses that same value for both the highlight and the tab-stop fallback. Before, the fallback read the
raw `selectedPath` prop, so with no file open the stop was `rows[0]`; now it is the `currentDir` row.
Reachable: in the narrow layout the tree unmounts when a file opens (`workspace-browser.tsx:1707`),
so going back resets `focused` to null while `currentDir` still points at that file's directory —
Tab then lands on that directory row instead of the first root row. Left alone: it is the better
behaviour, since that row is the one already drawn `aria-selected`, and it is what the component's
own doc comment promises ("the focused one, else the selected one, else the first"). Worth a line in
the PR description, because "the two Workspace pairs are byte-identical" does not cover it.
⬜ Left for the author

**6. Nothing runs `treeKeyStep` over real `flattenTree` rows any more** — `packages/web/test/file-tree.test.ts:44` · *nit*

The moved cases build `FileTreeRow` literals by hand. That is the right shape for a shared helper,
but it lands in the same PR that changed ArrowLeft from `parentDir(row.path)` to "nearest earlier row
at a shallower depth" — a rule that holds only because `flattenTree` emits a parent immediately above
its subtree at depth−1 (and `filterTreeRows` keeps ancestors, which I checked). Nothing now fails if
that stops being true. One assertion in `workspace-tree.test.ts` stepping left over
`flattenTree(LISTINGS, new Set(["a","a/b"]))` would close it.
⬜ Left for the author

**7. `Type: refactor` under-describes the plugin half** — `changelog/unreleased/2026-09-11-shared-file-tree.md:4` · *nit*

`changelog/README.md` defines `refactor` as "restructuring with no capability change" and `feature`
as "new user-facing capability". The Workspace half is a pure move, but the plugin browser came out
of this with arrow-key navigation, ARIA tree semantics and nested directory rows it did not have —
the entry's own lead paragraph describes them. Your call; the field drives nothing automated.
⬜ Left for the author

Also checked and clean: no new dictionary keys (`S.files.treeLabel` is reused and exists in both
`strings.ts` and `strings-en.ts`); `parentDir` still has production consumers after leaving
`treeKeyStep`; nothing in `packages/web/e2e/` selects `treeitem`, `data-tree-path`, `role="tree"` or
the plugin Modal; `.agents/skills/penguin-harness-frontend/SKILL.md` is a tracked file here, not a
design-repo symlink, and its new sentence is accurate — `test/placeholders-parity.test.ts` pins
`agent.placeholders` against `defaultSystemConfig().system_prompt`, and no test in
`packages/web/test/` compares the two dictionaries' interpolations.

### Verification

```
# node v24.18.0, worktree ../penguin-harness-wt/rv-685, base 34f17ad3

pnpm --filter @prismshadow/penguin-core --filter @prismshadow/penguin-server build   # fresh worktree needs this first
pnpm --filter @prismshadow/penguin-web typecheck    -> clean
pnpm lint                                           -> clean (oxlint)
pnpm format && pnpm format:check                    -> "All matched files use Prettier code style!"
git diff --check                                    -> clean

# vitest on the build box vps3-test (never on the dev box), via vps3-vitest.sh:
  before the fixes -> Test Files 130 passed, Tests 1819 passed
  after the fixes  -> Test Files 130 passed, Tests 1819 passed
pnpm --filter @prismshadow/penguin-web exec vitest run \
    test/file-tree.test.ts test/plugin-files.test.ts test/workspace-tree.test.ts
                     -> 3 files, 40 tests passed

# collision check with #682, read-only (no checkout of that branch):
git merge-tree --write-tree --messages HEAD origin/feat/file-browser-context-menu   # 61afc46c
  -> no conflict. #682's workspace-tree.ts hunks are all below line 500 (the wrap
     preference and the composer-reference helpers); this PR's deletions are at 143-260,
     so they never meet. In the merged tree, workspace-browser.tsx keeps this PR's
     `import type { TreeToggle } from "../../components/ui/file-tree"` (#682 does not
     touch that line) and #682's `hitRow` still resolves a row with
     `target.closest("[data-tree-path]")`, which this PR keeps on the same element and
     still withholds from retained rows. #682 does not touch workspace-tree-view.tsx and
     imports neither `subtreeEnd` nor `treeKeyStep`. The only shared text is the
     workspace-tree.test.ts header comment, where #682 edits the last line and this PR
     (after finding 4) the second and third — checked clean after the fix.
```

PR #686, opened 2026-09-11
feat(web): measure the server's reach to the model providers from Proxy options

hiyouga:
Proxy options gains a reachability test, **below the save row**: it lists the four provider targets with the exact URL each one probes, and one button measures them all, reporting a latency for every host that answered.

## What it does

- `GET|POST /api/admin/settings/proxy-probe` (admin only, same guard as the rest of the router). `GET` serves the target list without probing anything, so the page can show what it is about to request; `POST` runs the four probes concurrently.
- Reachable is judged by the fact that an answer arrived at all — **any HTTP status, 401 and 403 included**. A rejected credential still proves DNS resolved, TCP connected, TLS completed and the host replied, which is the whole question here. Unreachable is the transport failing, reported as a short kind: `timeout`, `dns`, `refused`, `tls`, `network`.
- Each target gets 5 s (matching the protocol probes next door). Concurrent, so one black-holed host does not add its timeout to the other three.
- `ms` is the wall time until the answer's headers arrived — DNS, connect, TLS, response — which is what the user waits for. The body is never read.

## Why the server, and why no key

The thing being measured is the *server's* outbound path — the one the proxy settings configure, and one the browser knows nothing about (a cross-origin request to a provider host would not even be sent). The route calls plain `fetch`, which is the undici fetch the startup entry installs and which resolves the process-global dispatcher per call, so a probe travels exactly the configured path. Verified by hand: saving an address changes every outcome (see Verification).

No credential is sent, and — unlike `services/protocol-detect.ts`, which deliberately falls back to `OPENAI_API_KEY` and friends because an authenticated probe identifies a protocol far more reliably — this one does not read the environment either. Nothing about reachability improves with a key, and a probe the admin did not authenticate must not quietly spend one. The route says so at the call site; the test asserts no `authorization` or `x-api-key` header reaches any target.

The target list is a server-side constant and the probe accepts **no request body**, so there is no caller-supplied address for the server to fetch and no SSRF surface. It is served rather than copied into the frontend so the URLs the page advertises cannot drift from the ones actually fetched — a test asserts the listed set equals the fetched set.

## Saved versus typed

The outbound dispatcher only moves on a save, so a typed-but-unsaved address is not in the path. The block therefore sits **below Save**: read top to bottom it is edit, save, then measure, which answers the question by layout instead of by warning. The response carries the proxy configuration the probes travelled and the block names it beside its heading (`出站路径：…` / `Outbound path: …`); a successful save clears results that now describe a superseded configuration. The probe has its own busy flag, so it neither blocks Save nor is blocked by it.

## UI

- **The latency is the result.** A reachable target shows only its number; the worded state is reserved for rows that have no number — unreachable ones, the not-yet-run state and the in-flight state — and travels with the figure in an `sr-only` span, since a bare number does not say "reachable".
- **The graphic** is a horizontal bar per reachable target, scaled against the slowest result of the **same run**, filled from `toneDot.success` (the map for solid marks with no interior to read). Deliberately relative: "slow" for a network hop has no absolute threshold this app could defend, and the models page's `ttftTone` thresholds are calibrated for streaming TTFT, not a TCP+TLS round trip. An unreachable target has no bar at all, which is why the row's words carry the state.
- Reachable is `success`, unreachable is `danger`, through `lib/tone.ts`; the not-yet-run state uses plain secondary text, because an absence is not a judgement about the target.
- The *explanation* — what the button measures, why a 401 counts as reachable, why it measures the saved settings, what the bars do and do not claim — lives behind the pane heading's existing "?" (`proxyInfo`). The **live value** it names stays on screen: that popover takes a static string built in `settings-dialog.tsx`, which has no access to the section's probe state, so only the prose moved.

## Probe URLs, and what each returns with no key

| Provider | URL | Unauthenticated answer |
| --- | --- | --- |
| OpenAI | `https://api.openai.com/v1/models` | 401 |
| Anthropic | `https://api.anthropic.com/v1/models` | 401 |
| Gemini | `https://generativelanguage.googleapis.com/v1beta/models` | 403 |
| DeepSeek | `https://api.deepseek.com/models` | 401 |

`GET`, not `HEAD`: two of the four do not answer `HEAD` honestly (`api.anthropic.com` 405, `generativelanguage.googleapis.com` 404), which would still prove reachability but would measure a path real traffic never takes.

## Tests

The classification (response/exception → outcome) is a pure function, `classifyProxyProbe`, tested in `packages/server/test/admin-settings.test.ts` beside the existing route coverage: any status is reachable; a timeout, a DNS failure, a refusal, a TLS fault and an unknown code each land on their kind. One route test with a mocked `fetch` — no test reaches the network. No React rendering test, no test per provider.

One subtlety the tests pin: a proxy that refuses the `CONNECT` tunnel arrives as an `AbortError` *nested* under `TypeError: fetch failed`, so only an abort on the outermost error may read as a timeout.

## Docs

`packages/docs/content/server-api.{zh,en}.md` gained both routes — the DTO contract names that file as the public reference.

## Verification

Node v24.18.0, from `/home/cc/dev/penguin-harness-wt/proxyprobe`.

```
$ pnpm --filter @prismshadow/penguin-server typecheck     # clean
$ pnpm --filter @prismshadow/penguin-web typecheck        # clean
```

Vitest on the remote host `vps3-test` (`.agents/scripts/remote-test.sh … --build "@prismshadow/penguin-core"`), after the rework:

```
 --filter @prismshadow/penguin-server test
 Test Files  125 passed (125)
      Tests  1803 passed (1803)

 --filter @prismshadow/penguin-web test
 Test Files  129 passed (129)
      Tests  1816 passed (1816)
```

Server baseline was 1800; the three added tests bring it to 1803. Web is unchanged at its 1816 baseline.

Driven by hand against a built server (`PENGUIN_WEB_DIST`, scratch `PENGUIN_HOME`, port 8947) with Playwright. Endpoint behaviour across four saved configurations:

```
# no address (direct / follow the environment)
4x reachable, ms 82/87/76/302, status 401/401/403/401        total 0.316601s

# http://127.0.0.1:8948 — a listener that accepts and never answers
4x timeout, ms ~5001                                          total 5.016146s
      ^ four 5 s timeouts inside 5 s: concurrent, and the proxy really is in the path

# http://127.0.0.1:8949 — nothing listening
4x refused, ms 7-9                                            total 0.016932s

# http://127.0.0.1:8950 — a test CONNECT proxy that refuses one host and delays two
openai   network    20 ms   (proxy answered 403 to CONNECT)
anthropic reachable 170 ms
gemini    reachable 682 ms
deepseek  reachable 2755 ms                                   total 2.768705s
      ^ a 16x spread plus one target with no bar: the case the graphic has to survive
```

Screenshots, all read back — four states x two themes x two languages, 1280x900, under `scratchpad/proxyprobe-logs/shots2/`:

| state | files |
| --- | --- |
| listed, not yet run | `{zh,en}-{light,dark}-1-idle.png` |
| mid-run | `{zh,en}-{light,dark}-2-probing.png` |
| mixed (one unreachable, 20x spread) | `{zh,en}-{light,dark}-3-mixed.png` |
| all reachable, real latencies | `{zh,en}-{light,dark}-4-reachable.png` |
| the "?" | `{zh,en}-{light,dark}-5-info.png` |

Design PR: penguin-harness-design [#142](https://github.com/Prism-Shadow/penguin-harness-design/pull/142)


hiyouga:
## Outbound path, saved settings, and the bars

Pushed as `531a4d40`.

### "Outbound path: the system proxy environment variables" is gone

That line was drawn whenever the switch was on with no address stored. The server does not
publish `HTTP_PROXY` / `HTTPS_PROXY`, and they need not be set at all — with none set, that exact
state connects directly through `EnvHttpProxyAgent`. So the line named a proxy that may not exist.

`pathLabel` now returns `null` for that state and the line is not rendered. The two states the
server can name exactly keep it: **direct, no proxy** when the switch is off, and the stored
address otherwise. `proxyProbeEnvProxy` is removed from both dictionaries, and the disclosed
explanation at the pane heading now says which states get a name and why the third does not.

### The test already follows the saved settings — now it is pinned

No change was needed: `probeProxyReachability` calls the global `fetch` that the startup entry
replaced with undici's, which resolves the process-global dispatcher per call, and
`applyProxySettings` rebuilds that dispatcher from the **stored** rows on every successful PUT.
With the switch off, `buildProxyDispatcher` returns a plain `Agent` — direct connections, proxy
variables ignored — which `proxy.test.ts` already covers.

What was not covered is the case that makes it visible: an address stays stored when the switch is
turned off, so the answer has to report the switch rather than echo the address back. Added to
`admin-settings.test.ts`:

```
the probe follows the saved switch, so a stored address with the proxy off is not the path
```

### The bars are gone

`barPercent` and the `slowestMs` scale are deleted with them, along with the sentence in
`proxyInfo` that had to explain that the bars were only relative. The millisecond figure is the
result; a bar that ranks four numbers against each other adds no reading and invites the absolute
judgement the explanation then had to take back.

### Verification

```
pnpm --filter @prismshadow/penguin-web typecheck     → exit 0
pnpm --filter @prismshadow/penguin-server typecheck  → exit 0
pnpm lint (oxlint)                                   → exit 0
pnpm format && pnpm format:check                     → clean
git diff --check                                     → clean
```

vitest on `vps3-test` (never locally):

```
server  admin-settings.test.ts + proxy.test.ts   38 passed
server  full suite                              125 files / 1804 tests passed
web     full suite                              129 files / 1816 tests passed
```

hiyouga:
## GLM's two hosts, first-come results, and no outbound-path line

Pushed as `0d7401b3`.

### GLM is two targets

`api.z.ai` is the global endpoint the catalog defaults to; `open.bigmodel.cn` is the mainland one a
bigmodel.cn key needs. They resolve and route differently, so a proxy can carry one and not the
other, and choosing between those two keys is exactly what this page's reader is doing. Both answer
an unauthenticated `GET /api/paas/v4/models` with 401, checked from the build box, which is the same
shape the other four targets have.

### Results are first come, first shown

The route now measures **one** target, named by a path segment matched against the same fixed list,
and the page asks for all six at once. A host that black-holes connections holds up its own row and
not the five that already answered. A single response carrying all of them meant the full
five-second timeout with five working results sitting behind it.

Three things that shape needs and now has:

- A run counter discards answers belonging to a run a save has superseded. Save is not blocked by a
  probe, so a mid-flight `adopt` could otherwise refill a row with a path nothing travelled.
- One toast per run, not one per target. A lost session fails all six identically.
- An id outside the list is a 404 that reaches no network. That is what keeps "the route takes no
  URL" true now that it takes something.

`ProxyProbeResponse` is one probe rather than an array, and the `proxyForApp` / `proxyUrl` echo is
gone with the line that read it.

### The outbound-path line is gone

Both strings behind it too. It repeated what the pane's own disclosed explanation says, and in the
one state it could not name it had nothing to put there at all.

### Verification

```
pnpm --filter @prismshadow/penguin-web typecheck     → exit 0
pnpm --filter @prismshadow/penguin-server typecheck  → exit 0
pnpm lint (oxlint)                                   → exit 0
pnpm format && pnpm format:check                     → clean
git diff --check                                     → clean
```

vitest on `vps3-test` (never locally):

```
server  admin-settings.test.ts   21 passed
server  full suite              125 files / 1804 tests passed
web     full suite              129 files / 1816 tests passed
```

The endpoint tests now fire one call per provider, assert each answer names its own target, and
cover an id outside the list reaching no network at all.

hiyouga:
## Review

Adds an admin-only reachability probe over the server's own outbound path: a served target list, a
one-target-per-call `POST`, and a Proxy options block that fires all six at once and fills each row
as its answer lands. The shape is right — the id-against-a-fixed-list route has no SSRF surface, the
run counter closes the stale-answer paths, and the tests reach no network. Four things wanted fixing;
all four are pushed in 599b5b9a. Ship after a look at the two items left for you.

### Findings

**1. The changelog pair claims the probe response carries the proxy configuration it travelled** — `changelog/unreleased/2026-09-11-proxy-reachability-probe.md:38`, `.zh.md:30` · *should fix*

`ProxyProbeResponse` is `{probe: {provider, url, outcome, ms, status?}}`. The `proxyForApp` /
`proxyUrl` fields that sentence describes were dropped in 0d7401b3 along with the outbound-path
line, and its test with them — the sentence survived in both languages. The rest of that bullet is
true, so only the clause went. ✅ Fixed in 599b5b9a

**2. The reachable verdict was written twice, and the two copies could disagree** — `packages/server/src/services/proxy-probe.ts:162` · *should fix*

`probeTarget` hard-coded `outcome: "reachable"` on a successful fetch, while `classifyProxyProbe`'s
`{status: number}` branch — the one the unit test pins — had no production caller anywhere in
`packages/*/src`. Concretely: make `classifyProxyProbe` return `"network"` for `status >= 500` and
`probe classification: any HTTP answer is reachable` goes red while the shipped endpoint keeps
calling a 503 reachable. Two representations of one rule, and a test guarding neither end of the
thing it names. The success branch now goes through the classifier, and `gemini` answers 503 in the
endpoint test so the rule is pinned where it is served, not only where it is declared. Mutation-checked:
that edit now fails both tests; before, only the unit test. ✅ Fixed in 599b5b9a

**3. "Not one credential, not even the process environment's" was asserted against an empty environment** — `packages/server/test/admin-settings.test.ts:405` · *should fix*

The test asserted no `authorization` and no `x-api-key` on any probe request, but never put a
provider key in `process.env` — so an `authorization: Bearer ${process.env.OPENAI_API_KEY}` added to `probeTarget`'s headers,
which is exactly the fallback `protocol-detect.ts` next door does have, passed it.
The five keys the catalog names (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`, `GEMINI_API_KEY`,
`DEEPSEEK_API_KEY`, `ZAI_API_KEY`) are stubbed for the run now, and no request may carry one under
any header name or in its URL. Mutation-checked: that edit fails with
`expected 'Bearer secret-OPENAI_API_KEY' to be null`, and passed before. ✅ Fixed in 599b5b9a

**4. Stale comment: the timeout note still describes four targets running concurrently in this module** — `packages/server/src/services/proxy-probe.ts:63` · *nit*

There are six, and since the route was split a call measures one — the concurrency is the browser's.
Rewritten to say what the five seconds now bounds: one row's wait, not the run's. Also dropped the
`PUT /api/admin/settings` left standing mid-test in `one request per target`; it was setup for the
removed configuration assertion and asserted nothing. ✅ Fixed in 599b5b9a

**5. A probe answer that fails after the section closes still raises a global toast** — `packages/web/src/features/settings/proxy-section.tsx:171` · *nit*

Unmount does not bump `runSeq`, and the ref object survives in the closure, so `runSeq.current === run`
still holds for an in-flight answer after the section is gone. The `setResults` is harmless — React 19
drops a state update on an unmounted component silently, so the premise that it warns and leaks does
not hold here — but the `.catch` branch calls `toastError`, which is a global surface. Whether an
abandoned run should stay silent is a UX call, not a defect; one line in the mount effect's cleanup
(`runSeq.current += 1`) would do it. ⬜ Left for the author

### Left for you

- **`redirect` is left at the default `follow`.** None of the six targets redirects today (checked
  with `curl -L`, `num_redirects=0` on each that answers), and no credential is attached, so nothing
  can leak. But a redirect would fold extra round trips into the reported `ms`, and a hijacked
  provider host would get to name the next hop the server fetches. `redirect: "manual"` returns the
  3xx itself under undici, which the "any HTTP answer is reachable" rule already handles, and would
  measure the first hop only. Behaviour change, so left to you.
- **Anthropic answers 403, not 401,** from a region its Cloudflare front blocks
  (`{"error":{"type":"forbidden","message":"Request not allowed"}}`), which is what
  `proxy-probe.ts:40`'s parenthetical calls 401. The code does not depend on it — the same
  parenthetical says the status is immaterial — so it is only worth touching if you want the note
  to read true from everywhere.

### Verified, no finding

- **The fan-out.** `runSeq` covers every stale-answer path I could construct: `adopt` bumps it on
  both the hydration and the save path and clears `probing` with it; a second click is blocked by
  the `probing` guard and the disabled button; a superseded run's `.finally` cannot clear the
  current run's `probing`, and `Promise.allSettled` with a `.catch` on every leg cannot reject, so
  `probing` cannot stick true. `reported` is declared inside `runProbe`, so it is per-run, not
  shared. `setResults((prev) => new Map(prev ?? []).set(…))` is a functional updater, so concurrent
  answers cannot lose one another under React 19 batching, and a remount gets a fresh ref and fresh
  state, so the old closure writes only to the dead instance.
- **The route's input.** `proxyProbeTarget` is a `===` match against the six-element list, so
  `%2F`-encoded traversal, a different case, a URL and an empty segment are all 404s that reach no
  `fetch` (pinned), and a multi-segment path does not match `:provider` at all. The admin guard is
  `app.use("*")` on the same router, which both routes are registered on, and both are asserted 403
  for a non-admin. The only header sent is `accept`, and the request is issued through global
  `fetch` — the undici one the startup entry installs — which is what makes the measured path the
  one `net/proxy.ts` configures rather than the pushed bundle's private copy.
- **The GLM targets.** `https://api.z.ai/api/paas/v4/models` and
  `https://open.bigmodel.cn/api/paas/v4/models` are `<base>/models` for exactly the two bases
  `settings.providerEnvNotes.zhipu` documents, and both answer an unauthenticated `GET` with 401.
- **No orphans.** `probeProxyReachability`, `OutboundPath`, `pathLabel`, `proxyProbeVia` and
  `proxyProbeDirect` have zero hits across `packages/*/src`, `packages/docs/content`, `plugins` and
  `scripts`. Both `server-api` docs carry both routes with the shipped response shape and the
  `probe_target_not_found` code, in both languages, and the i18n keys are paired in both
  dictionaries (typecheck enforces it; `proxyProbeFailure` is keyed by the non-`reachable` outcomes,
  so adding an outcome server-side breaks the web build rather than rendering blank).

### Verification

```
# vitest on the vps3 build box (never locally), worktree rv-686
pnpm --filter @prismshadow/penguin-server exec vitest run test/admin-settings.test.ts test/proxy.test.ts
    -> 2 files, 38 tests passed
pnpm --filter @prismshadow/penguin-server test   -> 125 files, 1804 tests passed
pnpm --filter @prismshadow/penguin-web test      -> 129 files, 1816 tests passed
    (core + server built remotely by the runner before each)

# mutation checks, same box, reverted after
probeTarget sends `Bearer ${process.env.OPENAI_API_KEY}`
    -> admin-settings.test.ts fails 1/21 ("expected 'Bearer secret-OPENAI_API_KEY' to be null")
classifyProxyProbe maps status >= 500 to "network"
    -> admin-settings.test.ts fails 2/21 (unit rule + the endpoint's 503 target)

# locally
pnpm --filter @prismshadow/penguin-server typecheck   -> clean
pnpm --filter @prismshadow/penguin-web typecheck      -> clean
pnpm lint (oxlint)                                    -> clean
pnpm format && pnpm format:check                      -> clean
git diff --check                                      -> clean

# endpoints, unauthenticated GET from a third host (no credential sent)
api.anthropic.com 403 · api.deepseek.com 401 · api.z.ai 401 · open.bigmodel.cn 401
api.openai.com and generativelanguage.googleapis.com timed out from that host — which is the
state this feature exists to show.
```

PR #687, opened 2026-09-11
fix(web): mint draft and shortcut ids without crypto.randomUUID

hiyouga:
## Summary

Parked-draft and user-shortcut ids are minted from `crypto.getRandomValues` instead of `crypto.randomUUID`, through a small shared helper (`packages/web/src/lib/random-id.ts`). The id shapes are unchanged: `draft-` plus 8 hex characters, `sc-` plus 12.

## Why

`crypto.randomUUID` exists only in a secure context (HTTPS, or `localhost`). A Web App served over plain HTTP from any other address (a LAN IP, a hostname behind a non-TLS reverse proxy, the Docker image reached by IP) has no such function, so the first id it tried to mint threw:

```
Uncaught TypeError: crypto.randomUUID is not a function
    at rne (index-Bm9n7ZV-.js:439:4358)   // randomDraftId
    at Gu  (index-Bm9n7ZV-.js:439:4642)   // parkActiveDraft
    at Ga  (index-Bm9n7ZV-.js:448:115888) // "+" click handler
```

Every "New chat" entry point (sidebar "+", group headers, the Agents and Plugins pages, "Create with AI") calls `parkActiveDraft` first, which mints a draft id whenever the active draft holds typed text. The throw happened before the draft was moved, so the text stayed in the active slot and every later click failed the same way: once a user had typed a draft and navigated elsewhere, "+" was dead until the text was cleared by hand. Saving a user shortcut hit the same function.

`crypto.getRandomValues` is available in every context that has `crypto` at all, so no fallback branch is needed. The desktop shell loads `http://localhost:<port>` and was never affected.

## Verification

Run on the remote test host via `.agents/scripts/remote-test.sh` (core + server built there first):

- `vitest run test/random-id.test.ts test/draft-sessions.test.ts test/user-shortcuts.test.ts` — 3 files, 39 tests passed
- `prettier --check` on the six changed files and `git diff --check` — clean (local)

https://claude.ai/code/session_01NN67fpAQm33xAnKVLpcA4n

### DIFF
--- a/changelog/unreleased/2026-09-11-random-id-insecure-context.md
+++ b/changelog/unreleased/2026-09-11-random-id-insecure-context.md
@@ -1,0 +1,14 @@
+# The "+" button parks a typed draft over plain HTTP
+
+- **Date:** 2026-09-11
+- **Type:** fix
+- **Scope:** `web`
+
+[中文版](2026-09-11-random-id-insecure-context.zh.md)
+
+Parked-draft and user-shortcut ids come from `crypto.getRandomValues` now, not
+`crypto.randomUUID`. The latter exists only in a secure context (HTTPS, or localhost), so a Web
+App opened over plain HTTP from any address but localhost threw
+`crypto.randomUUID is not a function` the moment it had to park a typed draft — a click on "+"
+while the draft held text — and every later click failed the same way while that text remained.
+The ids keep their shape: `draft-` plus 8 hex characters, `sc-` plus 12.
--- a/changelog/unreleased/2026-09-11-random-id-insecure-context.md
+++ b/changelog/unreleased/2026-09-11-random-id-insecure-context.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-09-11
 - **Type:** fix
 - **Scope:** `web`
+- **PR:** [#687](https://github.com/Prism-Shadow/penguin-harness/pull/687)
 
 [中文版](2026-09-11-random-id-insecure-context.zh.md)
 
--- a/changelog/unreleased/2026-09-11-random-id-insecure-context.zh.md
+++ b/changelog/unreleased/2026-09-11-random-id-insecure-context.zh.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-09-11
 - **Type:** fix
 - **Scope:** `web`
+- **PR:** [#687](https://github.com/Prism-Shadow/penguin-harness/pull/687)
 
 [English](2026-09-11-random-id-insecure-context.md)
 
--- a/changelog/unreleased/2026-09-11-random-id-insecure-context.zh.md
+++ b/changelog/unreleased/2026-09-11-random-id-insecure-context.zh.md
@@ -1,0 +1,12 @@
+# 通过明文 HTTP 访问时，「+」按钮也能把已输入的草稿存入列表
+
+- **Date:** 2026-09-11
+- **Type:** fix
+- **Scope:** `web`
+
+[English](2026-09-11-random-id-insecure-context.md)
+
+草稿会话与用户快捷方式的 id 改由 `crypto.getRandomValues` 生成，不再使用 `crypto.randomUUID`。后者只在安全
+上下文（HTTPS 或 localhost）中存在，因此以明文 HTTP 从 localhost 以外的任何地址打开 Web App 时，一旦需要把
+已输入的草稿存入列表——即草稿里有文字时点击「+」——就会抛出 `crypto.randomUUID is not a function`，且只要
+文字还在，之后每次点击都同样失败。id 的形态保持不变：`draft-` 后接 8 位十六进制，`sc-` 后接 12 位。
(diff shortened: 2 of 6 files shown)


hiyouga:
## Review

The fix itself is right: `crypto.getRandomValues` is not secure-context gated (only `randomUUID`
and `subtle` are), so the helper reaches the one API that works over plain HTTP, and it does so
with no `Math.random` fallback. The one real problem was in the test — the insecure-context case
asserted on the helper rather than on the two functions that threw, so reverting either call-site
swap left the whole new file green. Fixed here; ship after the note below.

**Entropy is unchanged, not merely "close".** `randomHex(8)` draws 4 bytes (32 bits) and
`randomHex(12)` draws 6 (48 bits). The old expressions took the first 8 and 12 hex characters of a
v4 UUID, and the version/variant nibbles of a v4 sit at hex indices 12 and 16 — past both slice
points — so the old ids carried exactly 32 and 48 random bits too. Bit-for-bit the same, and
neither id is a capability: `draft-*` is a localStorage array key inside one browser
(`draft-sessions.ts:40`, capped at 50), and `sc-*` addresses a row inside that one user's own
`ui_prefs.draftShortcuts` (`api/types.ts:259`, "Unique per user", max 3, server-enforced). At those
populations the birthday odds are ~2.9e-7 and ~1.1e-14. Both are far below 122 bits and both are
fine here.

**The fix is complete.** `rg randomUUID packages/web/src` returns nothing after this change — those
two were the only browser-side callers. Every remaining hit is in `packages/server/src` or
`packages/core/src` under Node, where secure context does not apply, and none of the five core
subpaths the web bundle imports (`omnimessage`, `markers`, `interfaces`, `context-limits`,
`model-catalog`) contains `randomUUID` in its build output. The other secure-context-only API in
`packages/web/src` is the clipboard, and both call sites already use `navigator.clipboard?.` so
they degrade instead of throwing. No backward-compatibility question either: the id alphabet and
length are identical, so ids already in localStorage and in `ui_prefs` keep parsing.

### Findings

**1. The insecure-context test asserted on the wrong function** — `packages/web/test/random-id.test.ts:24` · *should fix*

```ts
vi.stubGlobal("crypto", { getRandomValues: real.getRandomValues.bind(real) });
expect((globalThis.crypto as { randomUUID?: unknown }).randomUUID).toBeUndefined();
expect(randomHex(8)).toMatch(/^[0-9a-f]{8}$/);
```

Line 27 asserts a property of the object literal line 26 just wrote — it can never fail. Line 28
exercises `randomHex`, the one function in the repo that provably never names `randomUUID`.
Neither `randomDraftId` nor `newShortcutId` — the two frames in the stack trace in the PR
description — is called anywhere in the file. Concretely: restore `packages/web/src/features/chat/draft-sessions.ts:129`
and `user-shortcuts.ts:41` to `crypto.randomUUID()...` while keeping the new helper and test, and
`vitest run test/random-id.test.ts` still reports 2/2 passed with the bug fully live. I verified
exactly that on the remote host before changing anything.

Replaced with two cases that stub the same `randomUUID`-less `crypto` and then call
`parkActiveDraft` and `newShortcutId`. Against the pre-fix sources both fail with
`TypeError: crypto.randomUUID is not a function` at `draft-sessions.ts:129` and
`user-shortcuts.ts:41`, which is the reported stack trace; against this branch both pass.
✅ Fixed in 11ecb362

**2. `does not repeat` has a ~1-in-216,000 flake rate** — `packages/web/test/random-id.test.ts:21` · *nit*

200 draws at 32 bits gives `1 - exp(-200·199 / 2^33)` ≈ 4.6e-6 per run. Left alone: 8 is the real
draft-id length, and testing the length the product actually mints is worth more than the four
extra nines that `randomHex(32)` would buy. Noting it so a one-off red on this line is read as
what it is rather than debugged. ⬜ Left as is

### Verification

```
# vitest on the build box (vps3) via /home/cc2/.local/bin/vps3-vitest.sh rv-687 — never locally

# 1. the new tests against the PRE-FIX sources (git checkout $BASE -- the two src files):
vitest run test/random-id.test.ts
  → 2 failed | 2 passed — both new cases: "crypto.randomUUID is not a function"
    at randomDraftId (src/features/chat/draft-sessions.ts:129:53)
    at newShortcutId (src/features/chat/user-shortcuts.ts:41:23)
  (the PR's two original randomHex cases passed here — the tautology, demonstrated)

# 2. sources restored:
vitest run test/random-id.test.ts test/draft-sessions.test.ts test/user-shortcuts.test.ts
  → 3 files, 40 tests passed
  → 130 files, 1820 tests passed

# local
pnpm --filter @prismshadow/penguin-core --filter @prismshadow/penguin-server build   → ok
pnpm --filter @prismshadow/penguin-web typecheck                                      → clean
pnpm lint (oxlint)                                                                    → clean
pnpm format && pnpm format:check                                                      → clean, no other file touched
git diff --check                                                                      → clean
```

PR #688, opened 2026-09-11
Short display names for the built-in tools

hiyouga:
The chat's tool-call card names each of the seven built-in tools by a short alias instead of the name the model calls it by, and Appearance settings gained a switch (default on) that turns the aliases off.

| wire name | en | zh |
| --- | --- | --- |
| `read_file` | read | 读取 |
| `write_file` | write | 写入 |
| `edit_file` | edit | 编辑 |
| `exec_command` | exec | 执行命令 |
| `input_command` | follow | 跟进命令 |
| `run_subagent` | subagent | 子智能体 |
| `input_subagent` | communicate | 交流 |

## What changed

- `lib/tool-alias.ts` — `toolDisplayName(name, aliased)`, resolving a name against a keyed record in both dictionaries (`S.chat.toolAliases`, shaped like `S.chat.approvalModes`). A name with no entry comes back unchanged, so MCP tools (`mcp__<server>__<tool>`) and the names only older Traces carry (`kill_command`, `read_image`, `describe_image`, `kill_subagent`) render exactly as they are, in either switch position.
- `tool-call-card.tsx` — the alias is applied at the two render expressions that spell a tool name: the collapsed row header and the pending-approval block's name pill. Both carry the tool's own name in `title` whenever an alias is shown, so a reader matching a Trace or writing a permission rule can still get it without turning the switch off.
- `state/theme.tsx` — the preference, persisted per browser under `penguin.toolAliases` alongside the other Appearance choices. Default on: only the explicit off value reads as off, so an absent value (every existing browser) gets the aliases. Registered in `install-scope.ts`'s `KEY_RULES` as `scope: "browser"`, which `install-scope.test.ts` enforces.
- `appearance-section.tsx` — a `PrefRow` + `Switch`, with the explanation in `info` (the "?" beside the label) rather than `hint`, since it is semantics and not a format rule.

## What deliberately keeps the wire name

The alias never reaches logic. `DESCRIBED_TOOLS`, `FILE_TOOLS`, `LEGACY_IMAGE_TOOLS`, `previewArguments`, `headerSubtitle`, `pathArgument`, `pendingFilePayload`, `stream-model.ts`'s `run_subagent` branch and `agent-topology.ts` all keep reading `item.name`. Nor does it reach the surfaces where the literal value is the point: the Trace viewer (timeline lane labels and grouping key, bar tooltips, event payloads and summaries), the Agent settings tools table (where the Name column **is** the config key), and the Cost Center's error codes (`tool_failed:exec_command`). The CLI is untouched — this is a Web App display preference.

## Verification

Node v24.18.0, from the worktree `penguin-harness-wt/toolalias`.

```
$ tsc --noEmit -p tsconfig.json          # clean, no output

Checking formatting...

 ✓ test/tool-alias.test.ts (4 tests) 3ms
 ✓ test/install-scope.test.ts (27 tests) 53ms
 ✓ test/placeholders-parity.test.ts (3 tests) 2ms
 Test Files  130 passed (130)
      Tests  1820 passed (1820)
```

The vitest suite ran on the remote host `vps3-test`, per the standing instruction for this session.

Checked by hand against a built server (scratch `PENGUIN_HOME`, mock Anthropic endpoint emitting one turn that calls all seven built-ins plus `mcp__notion__search_pages` and `read_image`), driven with Playwright at 1180px, deviceScaleFactor 2:

| shot | what it shows |
| --- | --- |
| `01-cards-zh-alias-on.png` | nine cards, zh aliases; the two non-built-ins raw |
| `02-cards-zh-alias-off.png` | same conversation, switch off — every wire name back |
| `03-cards-en-alias-on.png` | en aliases |
| `04-cards-en-alias-off.png` | en, switch off |
| `05-appearance-zh.png` / `06-appearance-en.png` | the new row, last on the Appearance page |
| `07-approval-zh-alias-on.png` / `08-approval-zh-alias-off.png` | the pending-approval pill, both positions |
| `09-appearance-info-zh.png` | the row's "?" disclosure |

A short alias does not empty out the header: the status icon, name, subtitle, duration and chevron keep their order and spacing, and the subtitle simply starts further left (`读取 …/ws/notes.md`, `exec List the workspace`). Nothing truncates that did not truncate before.

## Notes

- `tool-call-card.tsx` is also touched by the open branch `feat/send-tool-to-background`, which adds a button to the same header row. This change adds two attributes and swaps one expression in that row; the card was re-read immediately before editing.
- One design line goes stale and is **not** edited here: `design/specs/06-PROTOTYPE.md:94` says the tool card 流式展示工具名 — with the switch on by default it shows the alias, and the tool's own name moves to the tooltip.


hiyouga:
## Review

The alias really is display-only: `toolDisplayName` has exactly one production call site and its
result reaches only the two JSX expressions, while `DESCRIBED_TOOLS`, `FILE_TOOLS`,
`LEGACY_IMAGE_TOOLS`, `previewArguments`, `headerSubtitle`, `pendingFilePayload`,
`agentIdFromRunSubagentArgs` and the approval POST (keyed by `toolCallId`, never by name) all keep
reading `item.name`. The table covers exactly core's seven `BUILTIN_TOOL_FACTORIES` and anything
unmapped falls through to the raw name. Three things were missing around it — the Playwright suite,
a pin, and the docs — all fixed below; ship after a look at the two decisions at the end.

### Findings

**1. The nine e2e locators that reach a tool card look for `exec_command`, which the card no longer
renders** — `packages/web/e2e/chat.spec.mjs:59` · *should fix*

The specs run in Chinese (`getByPlaceholder(/输入消息/)`, `name: "发送"`), and with the switch on by
default the collapsed row and the approval pill now read 「执行命令」. Every one of these stops
matching — `exec_command` survives on the card only inside `title=`, which is an attribute, not
text: `chat.spec.mjs:59,109,164`, `reload-midstream.spec.mjs:69,86`, `subagent.spec.mjs:324`,
`outline.spec.mjs:214`, `steer-abort.spec.mjs:58`, `steer-reload.spec.mjs:59`. Three of them
(`chat:59`, `reload-midstream:86`, `steer-*`) gate the `允许` click, so those specs die on the
visibility timeout before the approval is ever clicked. Not a CI failure — root `test:e2e` is core's live-LLM suite, and
`packages/web`'s `bash e2e/run.sh` is not wired into a workflow — but it is the suite anyone
touching the chat UI runs. Switched to the short name the app actually paints; `title` stays
untested because Playwright was not run (not installed on the build box). ✅ Fixed in 6546d42c

**2. Nothing pinned the table to core's real tool list** — `packages/web/test/tool-alias.test.ts:74`
· *should fix*

`expect(Object.keys(zh.chat.toolAliases)).toEqual(Object.keys(EN_ALIASES))` compared the table to a
copy of itself in the same file, so registering an eighth factory in
`packages/core/src/environment/tools/registry.ts` would have shipped one card still reading as the
model's wire name with nothing failing anywhere. Now compared against `BUILTIN_TOOL_FACTORIES`
imported from `@prismshadow/penguin-core`, the way `placeholders-parity.test.ts:9` derives its
tokens from core's real default prompt. ✅ Fixed in 6546d42c

**3. A new user-visible preference with no docs entry** — `packages/docs/content/web-app.en.md:58` ·
*should fix*

The Appearance page gained a row that nothing in `packages/docs` mentions. Added to the tool-card
bullet in both `web-app.en.md` and `web-app.zh.md`, which is where the Shortcuts launcher names its
own Appearance switch (`web-app.en.md:89`) rather than in the `### Appearance` list. ✅ Fixed in
6546d42c

**4. Real conflict with #691** — `packages/web/src/features/chat/tool-call-card.tsx` · *coordination*

`git merge-tree --write-tree HEAD origin/feat/send-tool-to-background` reports a content conflict in
`tool-call-card.tsx`; both dictionaries auto-merge clean. #691 turns the collapsed header from a
single `<button>` into a `<div>` wrapping one, rewriting the exact span this PR adds `title` and
`displayName` to. The resolution is mechanical — carry `title={nameTitle}` and
`{displayName || S.chat.unknownTool}` onto #691's restructured span. Its `DETACHABLE_TOOLS` set
keys off `item.name` and stays correct. Whoever merges second also owns the e2e locators in
finding 1 if they re-land the old text. ⬜ Left for the author

**5. The context panel's "Top tools" list still spells the wire name** —
`packages/web/src/features/chat/context-gauge.tsx:556` · *nit*

`S.chat.contextTopTools` ranks `read_file` / `exec_command` by context cost beside a stream whose
cards now read 「读取」/「执行命令」. Probably right — it is a measurement surface, like the Trace
viewer — but the changelog's list of surfaces that keep the real name names the Trace viewer, the
tools table and the Cost Center and not this one. ⬜ Left for the author

**6. `as Record<string, string>` opts the table out of the two-dictionary type check** —
`packages/web/src/lib/strings.ts:2031` · *nit*

`Strings = typeof zh`, so the cast erases the seven literal keys and a key added to one dictionary
alone would not be a type error the way the frontend contract promises. It is the house pattern
(`accentNames`, `statusNames`, `approvalModeNames` and five others do the same), and the lookup
needs a string index, so leaving it — the test from finding 2 now covers the parity the compiler
gave up. ⬜ Left as is

### Left for you

- **On by default.** That default is what rewrote the e2e text and what makes the pending-approval
  pill read "exec" where the wire name now lives only in `title=` — unreachable on a touch screen,
  on the one card the user is being asked to authorise. Options: keep it (the preview below the pill
  already shows the real command, which is what is actually being approved); default the switch off;
  or alias the collapsed row only and leave the approval pill spelling the tool's own name.
- **`penguin.toolAliases` is not a backward-compatibility question.** It is a new key, absent for
  everyone upgrading, and `initialToolAliases()` reads anything but the literal `"0"` as on — so an
  absent, corrupt or hand-edited value lands on the default instead of throwing. Nothing already on
  disk changes meaning. Flagging it only because a storage key usually deserves the check.

### Verification

```
# test replay against the base commit (34f17ad)
git show 34f17ad:packages/web/src/lib/tool-alias.ts   -> does not exist
git show 34f17ad:packages/web/src/lib/strings.ts | grep toolAliases -> no match
  => all four assertions fail at base (missing module, and Object.keys(undefined) in the parity
     test). None already passed there.

# vitest, on the vps3 build box (never on the dev box)
vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web exec vitest run test/tool-alias.test.ts
  1 file, 4 tests passed
  130 files, 1820 tests passed
vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-docs test
  7 files, 53 tests passed   (content/search/toc, for the two doc edits)

# locally
pnpm --filter @prismshadow/penguin-web typecheck   pass (tsconfig includes test/)
pnpm lint                                          pass (oxlint)
pnpm format && pnpm format:check                   pass
git diff --check                                   clean
git merge-tree --write-tree HEAD origin/feat/send-tool-to-background
  CONFLICT (content) in packages/web/src/features/chat/tool-call-card.tsx; strings.ts and
  strings-en.ts auto-merge

# not run: packages/web/e2e (Playwright is not installed on the build box)
```

PR #690, opened 2026-09-11
fix(server,web): a rejected sign-in link lands on the login page, not on raw JSON

hiyouga:
Reusing a one-time sign-in link dumped an error body in the browser:

```
http://localhost:7365/api/auth/claim?token=lZ5Kv2sC…
{"error":{"code":"unauthorized","message":"Invalid or already-used sign-in link."}}
```

`claim` is the one route a browser *navigates* to — its success path has always been a
redirect, because only the server can set the HttpOnly cookie. Its failure path threw an
`HttpError` instead, so the visitor ended up on a page of JSON with no way on.

## What changed

- **Server** (`packages/server/src/http/routes/auth.ts`): the refusal redirects to `/login`
  with `?claimFailed=`, instead of throwing. Nothing calls this endpoint programmatically —
  the desktop shell builds the URL and `loadURL`s it, the docs print it for a human, and no
  `fetch` anywhere in web/desktop/cli touches it — so no 401 was kept for an API caller.
- **The reason value is read from the deployment, never from the token.** With a desktop
  shell attached (`deps.desktop`), restarting the app is the way back in, because the shell
  mints a fresh link on every start; without one, restarting helps nobody and a new link has
  to come from whoever runs the server. Every token a given server rejects yields the same
  value, so the property the route's comment protects still holds: the two token kinds — the
  shell's one-shot token and the first-login link, which expire differently — remain
  byte-identical in failure.
- **Web** (`packages/web/src/pages/login.tsx`): the page reads the value once, raises a
  `Modal` over the ordinary form (portalled, Esc-layer aware, no new styling dependency) with
  copy that says what to *do*, and clears the parameter with a history replace, so a reload
  does not resurrect the dialog and a copied address does not carry it. The form behind it
  stays usable — the link is one way in, the password is the other.
- Copy lives in both dictionaries (`strings.ts` zh / `strings-en.ts` en). The dialog's single
  action is a new shared `common.gotIt`, so it does not repeat the header close button's name.
- `server-api` docs record the failure redirect beside the success one; changelog entry added
  in both languages.

## Verification

Node v24.18.0. Vitest ran on the remote host `vps3-test` (the standing instruction for this
session), typecheck and format locally in the worktree.

```
$ pnpm --filter @prismshadow/penguin-server typecheck     # tsc --noEmit, exit 0
$ pnpm --filter @prismshadow/penguin-web typecheck        # tsc --noEmit, exit 0
$ pnpm format && pnpm format:check                        # All matched files use Prettier code style!

    --build "@prismshadow/penguin-core" -- --filter @prismshadow/penguin-server test
 Test Files  125 passed (125)
      Tests  1801 passed (1801)          # baseline 1800 + the new reason-value test

 Test Files  129 passed (129)
      Tests  1816 passed (1816)          # baseline 1816
```

Seen by hand, not only asserted: built web + server, launched on a scratch `PENGUIN_HOME`
with `PENGUIN_SEED_ADMIN_PASSWORD`, ports probed with `ss -tln` (9147 plain, 9148 with
`PENGUIN_DESKTOP_TOKEN`), then navigated a browser straight at
`/api/auth/claim?token=<garbage>`:

| What | Screenshot |
| --- | --- |
| No shell attached, en | `claim-server-en.png` |
| No shell attached, zh | `claim-server-zh.png` |
| Desktop shell attached, en | `claim-desktop-en.png` |
| Desktop shell attached, zh | `claim-desktop-zh.png` |
| The real one-shot token still signs the window in | `claim-success-desktop.png` |

(Under `/tmp/claude-1017/-home-cc-dev-penguin-harness/fcdda244-9b29-434b-882a-9b2965d8c4b7/scratchpad/claimfail/shots/`.)

The success path was re-checked end to end — a valid desktop token still sets the cookie,
redirects to `/`, and lands on the chat page signed in as admin — and after dismissing the
dialog the address is a bare `/login`, a reload raises nothing, and signing in with
admin / the seeded password still works.

Design PR: penguin-harness-design [#145](https://github.com/Prism-Shadow/penguin-harness-design/pull/145)


hiyouga:
## Review

The route change is sound: the redirect target is a constant, no part of it comes from the
request (no query param, no `Host`/`X-Forwarded-Host`, no path segment), so there is no open
redirect; no token, user id or error text reaches the URL; the refusal consumes nothing it did
not already consume; and no machine client parses this route's body — `rg` finds exactly one
non-browser caller, `packages/desktop/src/util.ts:24`, and the shell only feeds it to
`win.loadURL` (`packages/desktop/src/main.ts:211`), so a redirect is strictly better there than
JSON. Ship after the fixes below, which are pushed; one item is left for you to decide.

### Findings

**1. The copy calls the first-login link "one-time" and says it "has already been used"** — `packages/web/src/lib/strings-en.ts:399`, `packages/web/src/lib/strings.ts:404` · *should fix*

`claimFailedServer` is the only string a non-desktop server can show, and on such a server the
only link `/api/auth/claim` can refuse is the first-login link — which is deliberately *not*
one-shot: `packages/server/src/auth/service.ts:157` ("Not single-use, because a prefetching
browser would burn it before its reader clicked") and `packages/docs/content/security.en.md:40`
("The link is not one-shot — a mail client may fetch it before you do, and a second open must
still work"). Concretely: admin sets a password, then opens the link from the console
scrollback — nothing was "used up" and it was never one-time, but the dialog tells them it was.
It now names the two things that actually retire the link (a password exists; a restart
re-minted it), which is what `security.en.md` already documents and what makes the next step
obvious. The `desktop` string is left alone — there "one-time" is accurate for the shell's
one-shot token. ✅ Fixed in e84c6bf4

**2. `size="sm"` is the only small button in any modal footer in the app** — `packages/web/src/pages/login.tsx:174` · *nit*

`rg 'footer=\{' -A 8 packages/web/src --glob '*.tsx' | grep 'size='` returned exactly one hit —
this one. Every other dialog footer (project-dialogs, change-password, schedules, admin-users,
mcp-servers, shortcuts, messaging-binding) uses the default `md`. Dropped the prop. ✅ Fixed in e84c6bf4

**3. `claimFailureRedirect` was exported only so a test could assert its own string format** — `packages/server/src/http/routes/auth.ts` · *nit*

`rg claimFailureRedirect` found the definition, one call in the same module, and
`desktop.test.ts` — no production consumer for the export. The property the unit test stated
(the advice comes from the deployment, never from the token) is already pinned end to end by
the route tests: three different rejected tokens all answer `claimFailed=desktop` under
`createDesktopApp` (a replayed shell token, a wrong token, a missing token), and
`createTestApp` answers `claimFailed=server`, with `first-login.test.ts` asserting the two
refusals carry an identical `Location`. The template is inlined and your rationale comment
moved onto the route, so the reasoning is where the string is. ✅ Fixed in e84c6bf4

**4. `first-login.test.ts` re-implemented `expectRefusal` inline** — *nit*

The "mints nothing once the server has been claimed" case asserted status and `Location` by
hand while the helper defined at the top of the same `describe` was in scope, and so dropped
its `set-cookie` check. Now calls the helper. ✅ Fixed in e84c6bf4

**5. Dismissing the dialog drops focus to `<body>`** — `packages/web/src/pages/login.tsx:57` · *nit, left alone*

`noticeOpen` starts `true`, so `Modal`'s closed→open edge fires during LoginPage's very first
render, when `document.activeElement` is still `document.body`; `restoreFocusRef` captures
that, and the close path calls `document.body.focus()`, which is a no-op. A keyboard user who
lands here from a dead link and presses "Got it" is left at the top of the document rather than
on the autofocused username field. Left alone — the remedy is either a ref into the form or
opening the dialog from an effect, and both are more machinery than the wrinkle is worth.
⬜ Left for the author

### Left for you

**A rejected claim now tells an unauthenticated caller whether a desktop shell is attached** —
`packages/server/src/http/routes/auth.ts:65`.

Before this PR both deployments answered the same 401, and `desktop.test.ts` said so in words
at the base commit: *"the same answer any wrong value gets, so the reply does not report which
modes are enabled."* After it, `curl -sI 'http://host/api/auth/claim?token=x'` reads the
deployment kind off `Location`. `desktopMode` was previously behind auth only
(`packages/server/src/http/routes/me.ts:45`). The exposure is small — a shell-started server
binds `127.0.0.1` (`packages/desktop/src/server-process.ts:134`), so only local callers can ask
it, and the answer is the same for every token, so it still enumerates nothing about a
credential — and your comment on the route states the trade-off deliberately. Two options, and
it is your call, not mine to make in a review:

- Keep it. The deployment-specific advice is the whole point of the change, and a login page
  has no authenticated way to learn `desktopMode` for itself.
- Emit one `?claimFailed=1` and write copy that covers both deployments, restoring the
  identical-in-every-mode answer at the cost of the "restart the app" instruction.

If you keep it, the base test comment it contradicts is already gone, so nothing is left
claiming the old property.

### Verification

```
# on vps3 (this box does not run vitest)
pnpm --filter @prismshadow/penguin-server exec vitest run test/desktop.test.ts test/first-login.test.ts
  -> 2 files, 23 tests passed (pre-fix, at 206e5263)
pnpm --filter @prismshadow/penguin-server test   -> 125 files, 1800 tests passed
pnpm --filter @prismshadow/penguin-web test      -> 129 files, 1816 tests passed

# locally
pnpm --filter @prismshadow/penguin-server typecheck   -> pass
pnpm --filter @prismshadow/penguin-web typecheck      -> pass
pnpm lint (oxlint)                                    -> pass
pnpm format && pnpm format:check                      -> clean
git diff --check                                      -> clean
```

Also checked and clean: i18n parity (both dictionaries carry the four new keys in the same
order, no placeholders on either side); `server-api.en.md` and `server-api.zh.md` say the same
thing and match the shipped behaviour; the changelog pair mirrors section for section with the
metadata block in the required order and a full PR URL; the 302 carries no cache headers and
302 is not heuristically cacheable, so a rejected credential is not stored; `/login` resolves
through the SPA fallback in `registerStaticRoutes`; and `deps.desktop` is `DesktopService |
null`, never `undefined` (`hmr/capabilities.ts:293` normalises with `?? null`), so the
`!== null` test cannot mis-read a non-desktop server as a desktop one.

PR #691, opened 2026-09-11
feat: move a running tool call to the background from its card

hiyouga:
While an `exec_command` or `run_subagent` call is executing, its row in the conversation now carries a **Send to background** action that hands the work back as a background task. The call closes with a `process_id` / `subagent_id`, nothing is killed, and the turn carries on instead of waiting the command out; the completion arrives later as the usual `[background_task_done]` report.

## Core

`ToolExecutionContext` gains `detachSignal`, a per-call channel distinct from `signal`. It has to be distinct: `Environment` classifies any `signal.aborted` as `stopReason: "aborted"` and appends `[interrupted: tool aborted by user]`, so a detach travelling on that signal would read as the opposite of what happened. `Environment.executeTool` mints the controller alongside the timeout controller and keeps it in a per-call map for the lifetime of the call, which is what lets something outside address one call; the entry is dropped in the same `finally` that clears the timeout.

`EnvironmentInterface.detachToolCall(toolCallId)` fires it and answers `"detached" | "not_running" | "not_detachable"` — three outcomes rather than a boolean, because the host answers them differently and `SessionManager.removeProcess` already establishes that shape here. `BuiltinTool.detachable` is the tool's own declaration; a tool that ignores `detachSignal` behaves exactly as before, which is what keeps the file tools and MCP tools untouched.

`exec_command` and `run_subagent` take the promotion path they already had: the detach ends the collect window (via `collectUntil`, the union of the two signals), and the existing "still running" branch registers the work and returns the handle. Two differences from a yield-deadline promotion, both deliberate:

- the completion report is armed, as on a `run_in_background` launch — the model never chose to poll this one, so nobody would learn how it ended;
- a detached `run_subagent` child also inherits the launching call's approval sink and live message tap. Without them the child parks at the approval queue forever, since no collect window will ever attach another.

The kill listener stays on `signal` alone, so a detach never kills the process; core's `collect` already ends non-destructively.

## Server

`POST /api/sessions/:sessionId/tool-calls/:toolCallId/background`, the sibling of the approvals route — the only other per-`tool_call_id` entry point — with the same `resolveSession` authorisation and the same 204.

- **404 `tool_call_not_found`** when nothing with that id is executing: unknown, already finished, or the runtime is gone. Same shape as `approval_not_found`, and the frontend treats it as the benign race it is.
- **409 `tool_not_detachable`** when the call is real and running but its tool has no background form. 404 would deny a call the user can see on screen and 400 would blame a well-formed request; the conflict is with what the tool *is*, which is the same reading as the process route's 409 `process_running`.

A detached `exec_command` then appears in `GET /:sessionId/processes` with its Stop/Remove actions, and the header and sidebar background counts move on their own through `session_background`, because `backgroundTaskCounts` counts registered-and-running entries and cannot tell a click-detached one from a `run_in_background` one. Both were confirmed on screen, not assumed.

## Web

The action shows only while the call is executing (`item.callComplete && !item.outputComplete && !pending`), only for the two detachable tools, only on a main-session card — a subagent's call lives in the child Session's environment, which the route does not target, the same rule `onRetryNow` already follows — and **not on a call launched with `run_in_background`**: its work is already in the registry, so there is nothing left to hand over, and the row would otherwise briefly carry the background glyph twice, once as the action and once as the mark that call already wears.

It is **inline text, not an icon button**: the app's existing text-action style (`text-brand-600 … hover:underline`), a real `<button>` because it acts and navigates nowhere. Type at the row's own size, no padding of its own, so it adds nothing to the row's box — an action that changed the row's height would break the rhythm of a list of calls. Measured, not eyeballed; see Verification.

The header row became a flex wrapper with the row button, the action and the chevron as siblings: a `<button>` cannot nest another, and this repo already settled that question in `memory-tab.tsx` ("a real `<button>` cannot nest another, so the actions are siblings, with the hover highlight on the whole header"). The hover tint stays on the wrapper, so the row still reads as one line, and the chevron keeps toggling. The chevron is named for what it does (`Collapse` / `Expand`, the naming every other chevron toggle here uses) rather than after the tool, which would give one row two buttons under a single accessible name.

**The chat header's background-task reading** becomes the words ("2 background tasks") instead of the activity glyph plus a number, and takes `toneInk.muted` — the ink the shared `BackgroundTasksMark` is moving to in #677, on the grounds that work parked behind the conversation is a standing arrangement rather than a turn in progress. The same fact should not be two colours on two surfaces. `muted` is allowed to recede below the graphical-mark threshold only when its meaning is already in text, which here it literally is.

**The background mark** is read from the output, not remembered from the click: `isBackgroundCall` parses the arguments JSON for `run_in_background`, which a click-detached call never set. The detach note opens with `DETACHED_TOOL_NOTE_PREFIX`, exported from core so the sentence the tool writes and the string that matches it stay one value, and the card marks the call when its output carries it. That survives a reload — the click is nowhere in the Trace and the note is — and it needs no optimistic state, because the detach ends the call immediately and its output lands within milliseconds.

## Note for reviewers

`packages/web/src/features/chat/tool-call-card.tsx` is also being edited by `feat/tool-name-aliases`, which touches the tool-name expressions in the header; this branch restructures the row around them and touches the duration/mark region. Expect a conflict in that file.

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/146

## Verification

Node v24.18.0. From the worktree `/home/cc/dev/penguin-harness-wt/tobg`:

```
pnpm --config.verify-deps-before-run=false -r build   # exit 0
pnpm typecheck                                        # exit 0
pnpm format && pnpm format:check                      # exit 0 — "All matched files use Prettier code style!"
```

Tests ran on the remote host `vps3-test` (this session's standing instruction), via the repo runner:

```
PENGUIN_TEST_HOST=vps3-test .agents/scripts/remote-test.sh /home/cc/dev/penguin-harness-wt/tobg \
  --build "@prismshadow/penguin-core" -- --filter @prismshadow/penguin-core \
  --filter @prismshadow/penguin-server --filter @prismshadow/penguin-web test
```

```
packages/core test:    Test Files  54 passed | 1 skipped (55)
packages/core test:         Tests  1162 passed | 5 skipped (1167)
packages/server test:  Test Files  125 passed (125)
packages/server test:       Tests  1803 passed (1803)
packages/web test:     Test Files  129 passed (129)
packages/web test:          Tests  1823 passed (1823)
```

Baselines were core 1158 / server 1800 / web 1816; the deltas are exactly the 4 + 3 + 7 tests added here.

### Row height, measured

The acceptance test for an inline action is that it costs the row nothing. A mock turn issues three `exec_command` calls at once — one still executing (it carries the action), one launched with `run_in_background`, one that just finishes — and the driver measures every tool row, then removes the action node from the row that has it and measures that row again (one controlled variable, so any difference could only be the action's own box):

```
MEASURE light-zh: {"heights":[31.48,31.48,31.48],"withAction":0,"before":31.48,"after":31.48}
MEASURE dark-zh:  {"heights":[31.48,31.48,31.48],"withAction":0,"before":31.48,"after":31.48}
MEASURE light-en: {"heights":[31.48,31.48,31.48],"withAction":0,"before":31.48,"after":31.48}
MEASURE dark-en:  {"heights":[31.48,31.48,31.48],"withAction":0,"before":31.48,"after":31.48}
```

Every row in the group is 31.48 px, with or without the action, in both themes and both languages. The same run asserts that exactly one of the three rows offers the action.

### Driven by hand

Built web dist + built server (scratch `PENGUIN_HOME`, `PENGUIN_SEED_ADMIN_PASSWORD`, port 18940 picked after `ss -tln`), mock LLM issuing an `exec_command` of `printf 'starting\n'; sleep 20; printf 'finished\n'` with a 120 s yield window, driven with Playwright from `packages/web`, in all four theme × language combinations:

1. the call executes and the action is on the row, which also carries exactly three buttons under three distinct accessible names — `01-executing-<theme>-<lang>.png`, `01b-action-closeup-<theme>-<lang>.png`;
2. one click: the conversation carries on to the model's next turn while the command still runs, and the header says so in words — `02-detached-*.png`, `02b-toolbar-*.png`;
3. the card reads completed (not interrupted) with the mark, and its output is `starting` + `[moved to the background by the user with process_id …]`, no `[interrupted: …]` — `03-card-output-*.png`;
4. three rows together, showing the even height and the action on only the executing one — `10-three-rows-*.png`.

An earlier run of the same flow also confirmed stage 5, the completion arriving later as the usual `后台命令完成` banner, which the model then acts on.

Screenshots: `/tmp/claude-1017/-home-cc-dev-penguin-harness/fcdda244-9b29-434b-882a-9b2965d8c4b7/scratchpad/tobg/shots/`


hiyouga:
## The marker moves to the row, and the header keeps its glyph

Pushed as `9036df7e`.

`2e2829c4` put the words in the chat header and left the session-list glyph on the tool row. That
is the wrong way round: the header's reading is a **count**, and the row's is a **single call
saying where its work went**.

- **Tool row**, right of the duration: `[Background]` / `[后台任务]`, in the row's own mono type at
  `muted`. That is the bracketed shape this row already uses for an outcome that needs a word.
  `BackgroundTasksMark` is the session list's count-of-many mark; standing in for one call it said
  nothing without a hover, and nothing at all on touch. It keeps its one remaining home in the
  sidebar.
- **Chat header**: the glyph and the number are back, in `busy`, exactly as before `2e2829c4`. A
  glyph beside a figure is how every other chip in that row says what its figure counts.

Everything else from that commit stands: the inline text action, its exclusion on a call already
launched with `run_in_background`, and the chevron named for what it does.

### One thing for you, across PRs

PR #677 moves `BackgroundTasksMark` from `busy` to `muted`. This header chip is a hand-drawn twin
of that mark and stays `busy`, so after both land the same count is emerald in the header and gray
in the session row — and this file's comment claims they are "the same figure and glyph". #677's
review raised the same point from the other side. It is one decision, taken once, for both
surfaces; I did not take it.

### Verification

```
```

vitest on `vps3-test` (never locally): `@prismshadow/penguin-web` → **129 files / 1823 tests passed**.

hiyouga:
## Review

Reviewed the core and server halves — the per-call detach channel, the "a detach is not an
interruption" claim, the detached `run_subagent` child's lifetime, and the route. The mechanism
holds up: the `inFlight` entry is minted before the generator is created and deleted in the
consuming loop's `finally`, which a `for await` also reaches when the engine abandons the
generator, so nothing leaks per call; a second detach fires an already-aborted controller;
a detach racing the tool's own completion falls through to the ordinary exit path because
`session.running` is read after collection ends, and `Environment` still emits exactly one
`tool_call_output`, with `aborted` false and no `TOOL_ABORTED_NOTE`; and the detached child gets
the same registry, eviction pin (`hasRunningBackgroundSubagents`) and dispose-on-shutdown/delete
path as a `run_in_background` one, with the server's entry-lifetime `entryApprove` behind both the
standing sink and the fallback sink, so an orphaned approval has somewhere to go. Ship after the
fixes below — four are pushed in `f53d6191`, two are left for you.

### Findings

**1. A detached `run_subagent` still tells the model to poll for an approval the standing sink is already carrying** — `packages/core/src/environment/tools/run-subagent.ts:233` · *should fix*

Detach a `run_subagent` while its child is parked at an approval. `collectWindow`'s `finally`
drops the window sink, the branch then calls `setPersistentApprovalSink(approve)` — whose
`pumpApprovals()` returns immediately, because `pumpingApprovals` is still `true` from the window
consultation — so `session.pendingApprovals` is still `1` when `approvalHint(session)` runs one
line later. The model gets back
`…(abort: true stops its current run)] [subagent is waiting for approval of 1 tool call(s); poll to review]`
while that exact request is being escalated to the user through the sink just attached. The
`run_in_background` branch, which installs the same sink, deliberately appends no hint
(`run-subagent.ts:186-191`); the deadline promotion at `:241` keeps it because it gets no sink.
Dropped the hint from the detach branch, with a regression test
(`a detached child with a queued approval is not reported as something to poll for`) that fails
against the pre-fix code with `expected '[moved to the background by the user …' not to contain 'poll to review'`.
✅ Fixed in `f53d6191`

**2. The new route is not in the API docs** — `packages/docs/content/server-api.{en,zh}.md` · *should fix*

Every other session route is in that table with its status codes — `/approvals/:toolCallId` and the
three `/processes` routes included — and `POST /tool-calls/:toolCallId/background` was in neither
language. Added a row after `/approvals/:toolCallId` in both, with the 204 / 404
`tool_call_not_found` / 409 `tool_not_detachable` branches. Also amended `tools.{en,zh}.md`'s
"Background completion reports" opening sentence, which named `run_in_background: true` as the only
launch that reports a completion: the detach branches call `armCommandDoneReport` /
`armSubagentDoneReport`, so a moved call reports too, and the page said otherwise.
✅ Fixed in `f53d6191`

**3. A web test still documents the glyph coupling this PR removed** — `packages/web/test/session-activity.test.ts:66-91` · *nit*

The `BackgroundTasksMark` block's doc comment names "a tool row, where it marks the single call
made with `run_in_background`" as one of the glyph's two placements, and `names one call where it
marks one call` renders the glyph with `S.chat.backgroundCall` as its accessible name. After
`9036df7e` the tool row draws a bracketed word and `BackgroundTasksMark` has exactly one caller
left (`sidebar.tsx:2729`). Nothing fails — the string is still a valid label — the test just
asserts a pairing no production code makes. Retargeted: the glyph tests use count labels only, and
the "one call, not a count of one" assertions moved to `tool-call-preview.test.ts` beside the rest
of the row's background logic.
✅ Fixed in `f53d6191`

**4. The Chinese changelog entry uses `## 详情`** — `changelog/unreleased/2026-09-11-send-tool-to-background.zh.md:15` · *nit*

`changelog/README.md` and `README.zh.md` both map `## Details` to `## 细节`, which 179 of the 187
existing `.zh.md` entries use. Renamed.
✅ Fixed in `f53d6191`

**5. The action renders, and can never work, on a subagent's own session page** — `packages/web/src/features/chat/tool-call-card.tsx:199` · *nit*

`showsBackgroundAction` keeps the action off nested cards with `origin.length === 0`, but the
subagents panel's "open as session" (`subagents-view.tsx:182`) navigates to
`/chat/<childSessionId>`, and there the child's own calls render at origin `[]`. The server never
builds a runtime entry for a child session — it gets a `sessions` index row
(`session-manager.ts:2005`) and is driven through the parent — so `resolveSession` succeeds,
`SessionManager.detachToolCall` finds no entry, returns `not_running`, and the 404 is swallowed at
`chat-page.tsx:1353`. On that page any mid-execution call therefore offers an action that does
nothing at all. Both repairs are product calls (hide it on subagent-sourced sessions, or let the
route resolve a child through its parent Session's environment), so I left it.
⬜ Left for the author

**6. The `[Background]` marker is drawn in `muted`, below text contrast, as the sole carrier** — `packages/web/src/features/chat/tool-call-card.tsx:426` · *nit*

`toneInk.muted` is `text-gray-400 dark:text-gray-500`, measured in `tone.ts` at 2.60:1 light and
4.34:1 dark; the frontend skill allows it under 3:1 only "because its meaning is always already in
text". Here the marker *is* the text and the only thing saying where the work went, and it sits
immediately right of a duration drawn in `text-gray-500 dark:text-gray-400` — the row's own
secondary pair, which `tone.ts` classes as typography rather than a tone. That pair would be both
more legible and consistent with the line it joins. Left alone because you set this marker in
`9036df7e`; it is a one-class change if you want it.
⬜ Left for the author

### Left for you

- **The cross-PR tone decision from your comment is now taken as "both gray".** Per the repository
  owner, the chat header's background chip moves from `toneInk.busy` to `toneInk.muted`, keeping
  its glyph and its count exactly as they are, so it agrees with where PR #677 takes
  `BackgroundTasksMark`. The comment above that span no longer claims the live-status green, and it
  now says why `muted` is admissible there (the `title` names the count in words, so colour carries
  nothing alone). A bullet went into both changelog files.
- **PR #688 (`feat/tool-name-aliases`) conflicts.**
  `git merge-tree --write-tree HEAD origin/feat/tool-name-aliases` reports one content conflict, in
  `packages/web/src/features/chat/tool-call-card.tsx`; both dictionaries auto-merge. It is the row
  header: this PR turned the row `<button>` into a `<div>` with an inner button, and #688 replaced
  `{item.name || S.chat.unknownTool}` with `{displayName || S.chat.unknownTool}` plus a
  `title={nameTitle}` on that same `<span>`. Resolution is mechanical — keep this PR's structure and
  take #688's `title`/`displayName` span inside it; #688's second hunk (the pending-approval pill)
  is untouched here and merges cleanly.
- **No backward-compatibility question.** Nothing new is persisted: the `[Background]` marker is
  read out of the note the tool writes into the Trace, so an old Trace simply carries no marker, and
  no config, `web.db` or localStorage shape moves.

### Verification

```
Local:
  pnpm typecheck                                    → all 23 projects Done
  pnpm lint (oxlint)                                → exit 0
  pnpm format && pnpm format:check                  → clean
  git diff --check                                  → clean
  pnpm --filter @prismshadow/penguin-web build      → built (the new value import of
                                                       DETACHED_TOOL_NOTE_PREFIX from
                                                       @prismshadow/penguin-core/interfaces
                                                       resolves and lands in the bundle)

vitest on vps3-test only, never on this box:
  core  test/background-execution.test.ts (pre-fix, hint still present)
        → 1 failed: "expected '[moved to the background by the user …' not to contain 'poll to review'"
  core  test/background-execution.test.ts            → 33 passed
  core  @prismshadow/penguin-core test               → 54 files / 1163 passed, 5 skipped
  server @prismshadow/penguin-server test            → 125 files / 1803 passed
  web   @prismshadow/penguin-web test                → 129 files / 1823 passed
  docs  @prismshadow/penguin-docs test               → 7 files / 53 passed

Test replay against the merge base (34f17ad3):
  core   background-execution.test.ts — every new assertion needs detachToolCall / ctx.detachSignal,
         neither of which exists there; the subagent case would sit out its 60s window instead.
  server session-processes.test.ts — the route does not exist at base, so 204 and 409 both come back
         404 and the tool_call_not_found code assertion reads the generic handler's code instead.
```

PR #694, opened 2026-09-11
Render thinking and compaction bodies as output, not as a quotation

hiyouga:
## What

An expanded thinking block, and both sections of a compaction banner, sat in an inset rounded box on a tinted ground — the shape the transcript uses for a quotation. Both now wear the block an expanded `exec_command` output wears: full-bleed under their row, separated from it by the same divider, inset by the same padding, in the same ink. The class string lives in `DISCLOSURE_BODY_MD_CLASS`, next to the `DISCLOSURE_OUTPUT_PRE_CLASS` it is derived from, so the two stay in step.

## Where it deliberately differs from the output block

- **Prose type, not the output block's.** Aligning to `text-xs` was tried first and was wrong on screen: the output block is monospace, and the app's sans at the same size reads visibly smaller beside it. The body stays at `text-sm`/`leading-relaxed` — within a hair of the 1.7 that `.md-body p` applies, so paragraphs, list items and headings land on one rhythm instead of two. The outermost block's margin is dropped at each end, so the body's own `py-2` is the whole gap, the same one the output block leaves.
- **Markdown keeps rendering.** These bodies are prose — a compaction summary is Markdown — so `whitespace-pre-wrap` and a raw `<pre>` would be a regression, not parity.
- **No height cap.** Both bodies stream. `max-h-72 overflow-auto` would put the live tail inside a nested scrollbox that the transcript's own follow scrolling cannot reach.

## Verification

Run in a dedicated worktree on `origin/main` (`1a26d1cd8`), CI's chain, at both commits:

- `pnpm install --frozen-lockfile` — clean
- `pnpm -r build` — 11 packages, `Done`
- `pnpm typecheck` — every package `Done`, zero `error TS`
- `pnpm format` + `pnpm format:check` — 1280 files checked, 0 rewritten, "All matched files use Prettier code style!"
- `pnpm --filter @prismshadow/penguin-web test` (run on the test host, with `--build "@prismshadow/penguin-core... @prismshadow/penguin-web..."` as CI does) — **Test Files 129 passed (129), Tests 1816 passed (1816)**

No test asserted the old class string; no spec covers the body's styling (`specs/06-PROTOTYPE.md` describes the group's sticky behaviour, not its inset), so no paired design change.


hiyouga:
## Review

The shape change lands: rendered against the built stylesheet at the app's real 18px root, both bodies lose the tinted rounded box and sit full-bleed under their row with the `exec_command` output's divider, inset and ink, in light and dark. One property the new constant documents did not survive contact with the cascade, though: the two arbitrary-variant utilities meant to flush the body's end margins never apply, so the gap at each end stayed roughly double the output block's. Fixed here and pinned by a test; the changelog pair also needed the spec's heading and its reasoning taken back off disk.

### Findings

**1. `[&>*:first-child]:mt-0 [&>*:last-child]:mb-0` never apply — the body's end inset stayed ~2× the output block's** — `packages/web/src/features/chat/disclosure-row.tsx:73` · *blocking*

`packages/web/src/styles.css` writes all of its own rules outside a cascade layer, while Tailwind emits utilities inside `@layer utilities`. An unlayered normal declaration beats a layered one at any specificity, so `.md-body :is(p, ul, ol, pre, blockquote, table) { margin: .5rem 0 }` and `.md-body :is(h1, h2, h3, h4) { margin: .75rem 0 .5rem }` win outright — the two utilities are dead classes for every block Markdown can put first or last.

Measured in headless Chromium against `packages/web/dist/assets/*.css` at root font 18px, a thinking body whose first block is an `<h3>` and whose last is a `<p>`:

| | first child `margin-top` | gap above first glyph | gap below last glyph |
| --- | --- | --- | --- |
| `main` | 13.5px | 22.5px | 18px |
| this PR | 13.5px | 23.5px | 18px |
| the `exec_command` `<pre>` it is matched to | — | 9px | 9px |

A one-paragraph body measures 18px / 18px on `main` and 19px / 18px here, against the same 9px / 9px. So "the outermost block's margin is dropped at each end, so the body's own `py-2` is the whole gap — the same one the output block leaves" did not hold — in the comment, in the PR description, or in the changelog.

Fixed by moving the reset into `styles.css` as `.md-body-flush`, unlayered like the `.md-body` margins it has to beat, and pointing the constant at it. Re-measured: 10px above (the `py-2` plus the 1px divider) and 9px below — the output block's inset exactly. `packages/web/test/disclosure-body.test.ts` pins the rule, the fact that it is unlayered, and that the constant never goes back to a utility. ✅ Fixed in `1d36c2f3`

**2. Changelog: the Chinese heading is not the one the spec names, and the rationale went on disk** — `changelog/unreleased/2026-09-11-thinking-body-output-style.zh.md:11` · *should fix*

`changelog/README.md` maps `## Details` to `## 细节` (178 entries in the tree use it; 7 use `## 详情`), and states outright that reasoning does not go on disk. The entry carried "the output block is monospace, and the app's sans at that size reads visibly smaller beside it" and "a nested scrollbox would strand the tail the transcript's own follow scrolls to" — both already in the PR description, which is where they belong. The English file also had no lead paragraph: its whole body was a single nine-line bullet under `## Details`, where the template and all four sibling entries open with a paragraph. Rewritten as a lead paragraph plus four short bullets, mirrored section for section, heading corrected. ✅ Fixed in `1d36c2f3`

**3. The extraction is a drift guard, not a size win — and the part that is actually duplicated is untouched** — `packages/web/src/features/chat/disclosure-row.tsx:54` · *nit*

Each call site traded a 136-character inline class string for a 37-character interpolation, while the shared piece is two lines of code under a twenty-line comment, so the diff grows. That is a fair trade — two byte-identical strings can no longer drift, and the constant is now readable against the `DISCLOSURE_OUTPUT_PRE_CLASS` it is derived from twenty lines above. But both call sites still spell out the whole body verbatim — ``<div className={`anim-fade ${DISCLOSURE_BODY_MD_CLASS}`}><Md text={…} streaming={…} /></div>`` — so the duplication that remains is the one an extraction would remove. A `DisclosureMdBody({ text, streaming })` next to `DisclosureRow` would take it; keeping `anim-fade` out of the class constant (the output block's callers put it on an outer wrapper) is the only argument against. Your call — left as is. ⬜ Left for the author

### Verification

```
# rendering: both surfaces, light and dark, at the real 18px root
headless Chromium (@playwright/test) over packages/web/dist/assets/index-*.css,
DOM reproduced from work-group.tsx / step-banner.tsx / disclosure-row.tsx:
  thinking block (rich markdown, one paragraph, long body), compaction banner
  (both sections), empty body, before / as-submitted / after — measurements in finding 1

pnpm --filter @prismshadow/penguin-web build   ✅ (confirms Tailwind emits .md-body-flush)
pnpm --filter @prismshadow/penguin-web typecheck   ✅ no errors
pnpm lint                                          ✅ oxlint, exit 0
pnpm format && pnpm format:check                   ✅ "All matched files use Prettier code style!"
git diff --check                                   ✅ clean

# vitest on the build box (vps3), never on this machine
vps3-vitest.sh <worktree> rv-694 -- pnpm --filter @prismshadow/penguin-web test
  Test Files  130 passed (130)
       Tests  1822 passed (1822)   ✅  (the new file's 6 tests included)

# the new test's assertions replayed against the pre-review tree
git show f941208f:packages/web/src/styles.css        | grep -c md-body-flush  -> 0
git show f941208f:.../disclosure-row.tsx             | grep -c md-body-flush  -> 0
git show 1a26d1cd:.../thinking-block.tsx    -> className="md-body anim-fade mx-3 mb-2 rounded-md bg-gray-50 …"
git show 1a26d1cd:.../compaction-banner.tsx -> the same string
```

PR #698, opened 2026-09-11
The web-design skill takes the generic techniques from the Z.ai Code prompt

hiyouga:
Nine small additions to `plugins/software-development/skills/web-design/SKILL.md`, taken from the published [Z.ai Code prompt](https://github.com/x1xhlol/system-prompts-and-models-of-ai-tools/blob/main/Z.ai%20Code/prompt.txt). The brief: bring over the generic techniques, nothing tied to the `z-ai-web-dev-sdk`. Each item is one sentence or one recipe bullet, placed in the section that already owns the topic.

## What came over

| Z.ai rule | Where it landed | Form |
| --- | --- | --- |
| "write the frontend first to let user see the result, then write the backend" | *Before you start* | frontend first against mocked data, then the backend |
| standard stack when the user names none; "local memory caching, no additional middleware (MySQL, Redis)" | *Before you start* | pick a stack and state it: SQLite / `localStorage`, in-process caching, no infrastructure a one-line request did not ask for (framework left to the system prompt's "prefer React") |
| "Mobile-First: design for mobile, then enhance"; Tailwind breakpoints | *Ship complete* | base styles are the phone layout, `min-width` queries add columns from 640px (replaces "single column under 640px", same threshold) |
| "Semantic HTML: main, header, nav, section, article"; "sr-only class for screen reader content" | *Ship complete* | landmarks instead of div soup, an `.sr-only` utility, next to the existing `alt` / `aria-label` line |
| "always read dev log when you finish coding"; "only read the most recent logs" | *Ship complete* | tail the dev server log and the browser console after each coding pass, fix before presenting |
| "Card alignment and padding — consistent padding" | *Components › Card* | cards in one grid share padding and stretch to equal height |
| "Long list handling — max height with scroll overflow and custom scrollbar styling" | *Components › Long list* (new) | `max-height: 24rem; overflow-y: auto;` thin styled scrollbar; the page never scrolls because of it |
| "Loading States: spinners/skeletons" | *Components › Skeleton* (new), *Chat › States* | skeleton blocks for regions, dots for actions |
| "Feedback: toast notifications for user actions" | *Components › Toast* (new) | placement, tokens, `role="status" aria-live="polite"`, auto-dismiss; never `alert()` |
| "Hover Effects: interactive feedback on all clickable elements" | *Motion* | every clickable element has a hover state and `cursor: pointer` (the color-only rule stands) |
| "websocket/socket.io for real-time communication" | *Chat › Messages* | deltas stream over SSE or a WebSocket, never polling |

## Left behind, on purpose

- Everything about `z-ai-web-dev-sdk`, `z-ai-generate`, the auto-started dev server, the single `/` route and `npm run build` being forbidden — product plumbing.
- Next.js 15 / App Router / Prisma / shadcn / Zustand / TanStack as *the* stack, and "redirect users to our built-in alternatives" — the skill is framework-agnostic on purpose and the system prompt already says "prefer React".
- "avoid indigo or blue colors" — the Penguin language is built on a single blue accent; the rules conflict and ours is the product's.
- "do not write any test code" — contradicts the sibling `software-engineering` skill and the repo's verification stance.
- "Minimum 44px touch targets" — the skill says ≥ 40px in two places; raising it is a design decision for the skill's owner, not a port. Flagging it here in case you want 44 (Apple HIG / WCAG 2.5.5).
- Color variables, light/dark theming, typography hierarchy, responsive breakpoints, accessibility basics, animations, error messages — already covered by the skill's tokens, *Ship complete* and *Motion* sections.

## Version

`plugins/software-development/plugin.json` → `2026.09.11.1` (skill content changes bump the plugin's date version; `SKILL.md` carries no version of its own).

## Verification

Run locally on this machine.

- `pnpm --filter @prismshadow/penguin-core exec vitest run test/plugins.test.ts test/agent-skills.test.ts` — 31 passed.
- `pnpm --filter @prismshadow/penguin-docs exec vitest run test/skills-sync.test.ts` — 3 passed; `pnpm --filter @prismshadow/penguin-landing exec vitest run test/skills-sync.test.ts` — 6 passed.
- `pnpm format` + `pnpm format:check`, `git diff --check` — clean.
- No spec change: 05 describes the plugin mechanism, not the skill's content; 08 only records large repositionings of this skill.

https://claude.ai/code/session_01NN67fpAQm33xAnKVLpcA4n


hiyouga:
## Review

Nine techniques ported into `web-design`, plus the plugin's dated version bump and a bilingual
changelog pair. I checked the provenance question independently against the leaked prompt and it
comes out clean — nothing expressive was carried across (finding 1). Two places where the new text
and the file around it disagreed are fixed in `8621a193`; ship after a look at those.

### Findings

**1. Provenance: what landed is technique, not expression** — `plugins/software-development/skills/web-design/SKILL.md` · *verified, no change needed*

I diffed the added text against the published Z.ai Code prompt independently of the PR description
(the `x1xhlol` capture the description links, cross-checked against the mirror at
`systemprompts.org/zai-code-system-prompt/`). **No sentence in this diff is verbatim or
near-verbatim.** Every item is re-expressed in this skill's own vocabulary — its CSS tokens, its
sentence rhythm — and the four I looked at hardest all come out on the "fact, not wording" side:

- `max-height: 24rem` (:85) against Z.ai's *"Set max height with scroll overflow
  (`max-h-96 overflow-y-auto`)"*. `max-h-96` **is** 24rem, so the number carried. But the number is
  the technique, and it had to be rewritten as plain CSS anyway because this skill bans Tailwind.
- The landmark list (:26) against *"Use `main`, `header`, `nav`, `section`, `article`"*. Same five
  elements — but that is the HTML5 landmark set, not someone's phrasing of it; the PR adds `footer`
  and the "instead of div soup" framing.
- *"never Redis, MySQL"* (:14) against *"Local memory caching, no additional middleware (MySQL,
  Redis, etc.)"*. The pairing carried, the sentence did not, and Prisma — the actual vendor-specific
  half of that rule — was correctly dropped.
- `.sr-only` (:26) against *"Use `sr-only` class for screen reader content"*. `.sr-only` is the
  universal class name (Bootstrap, Tailwind, every a11y guide); naming it is unavoidable.

Everything product-shaped stayed out: Next.js 15 / App Router, Prisma, shadcn/ui, Zustand,
TanStack, next-themes, `z-ai-web-dev-sdk`, the hardcoded `/home/z/my-project/dev.log`, and the
"no indigo or blue" rule. That matches the description's "left behind" table exactly.

One thing I checked and am deliberately **not** asking for: this repo does have an in-file
attribution convention — `plugins/use-bento-slides/skills/bento-slides/SKILL.md:16` names its
source with a URL and a licence. That convention is for an adapted derivative of someone else's
*text*. Nothing here reaches it, and adding such a line would overstate the relationship in a file
that ships to users. Source in the PR description is the right place for it.

**2. The Long list scrollbar renders near-white in dark mode** — `SKILL.md:85` · *should fix*

`scrollbar-color: var(--gray-300) transparent`, but the `.dark` block (:47–53) redefines only
`--bg`, `--surface`, `--border`, `--control-border`, the `--fg-*` and the accent/brand tokens — it
never redefines `--gray-300`. So on a page that wires dark mode (which *Ship complete* makes
mandatory) the "thin" thumb paints `#d1d5db` on `--bg: #000000`. The file's convention is to name a dark counterpart
wherever it reaches for a raw gray — including the **Skeleton** recipe added one line below
(`--gray-100`/dark `#1f1f1f`), so the convention was applied within this very diff and missed here.

Fixed by moving to `--control-border`: `#d1d5db` in light — byte-identical rendering — and `#303030`
in dark, the same tone the file already trusts for input borders. ✅ Fixed in `8621a193`

**3. "single column under 640px" survives in the Page layout recap** — `SKILL.md:133` · *nit*

The diff replaced exactly that phrase in *Ship complete* with the mobile-first form, but the same
phrase still closes the Page layout section. It is not a logical contradiction — the rendered result
is the same — but it is the framing that produces `@media (max-width: 640px)`, which is the query
direction the new rule at :23 forbids. An agent reading the file top to bottom gets the rule and
then a worked example pointing the other way.

Reworded to "the single column is the base and a `min-width: 640px` query adds the grid".
✅ Fixed in `8621a193`

**4. The dev-server-log bullet assumes a dev server this skill never establishes** — `SKILL.md:28` · *nit*

Every other bullet in *Ship complete* is a property of the delivered page (`<html lang>`, viewport,
dark mode, keyboard path, `alt` text, zero external requests). This one is a workflow step, and it
presumes a running dev server — which the skill's own defaults often rule out: "Zero external
requests: system fonts, inline or local CSS/JS, inline SVG icons" plus "treat the user's one-line
request as the whole spec" means the common delivery is a single static file with no server and no
log. In the source prompt the equivalent rule was safe because Z.ai's harness always has a dev
server at a fixed path; generalising it kept the assumption.

Left alone: scoping a directive ("…and, when the page runs behind one, the tail of its log") changes
what the shipped skill tells users' agents to do, which is your call rather than a defect fix.
⬜ Left for the author

### Left for you

**Tap targets, 40px vs 44px.** You flagged this in the description and I agree it is not a port
decision. For the record, both sides have weight: 40px is already stated twice and is a stretch
against the skill's own controls (14px text + 6px/12px padding lands near 32px), while 44px is both
WCAG 2.5.5 AAA and Apple HIG. Raising it means touching :26 and :133 together — two lines
either way, but it is a design decision for the skill's owner.

### Verification

```
# vitest on the vps3 build box (vps3-vitest.sh, worktree rv-698)
pnpm --filter @prismshadow/penguin-core exec vitest run test/plugins.test.ts    17 passed
pnpm --filter @prismshadow/penguin-docs exec vitest run test/skills-sync.test.ts 3 passed
pnpm --filter @prismshadow/penguin-landing exec vitest run test/skills-sync.test.ts 6 passed

# locally
pnpm format:check   clean
pnpm lint           clean (oxlint, exit 0)
git diff --check    clean
```

Also checked, no action: `plugin.json` `2026.09.11.1` matches `PLUGIN_VERSION_PATTERN`, the entry
date and the one-bump-per-content-change precedent from
[#666](https://github.com/Prism-Shadow/penguin-harness/pull/666); the package's own `0.2.11` is the
release version and correctly untouched. Nothing else reads that string — the README category
tables (`packages/core/test/plugins.test.ts:248`) and both skills-sync tests key on plugin and skill
*names*, which did not move, so no docs table, manifest or fixture had to follow. `SKILL.md`
frontmatter still carries only `name` + `description` (the loader stamps `version` /
`short_description_zh` from `plugin.json`, asserted at `plugins.test.ts:70`), `## Before you start`
survives, no Chinese outside the `.zh.md`, and the changelog pair mirrors section for section with
`Type: feature` matching the `0.1.5` web-design precedent.

PR #700, opened 2026-09-11
A compaction request runs at the Session's thinking level

hiyouga:
`runCompactionRequest` passes the Session's pinned thinking level, exactly as `runTurn` does.
One argument, the comment above it rewritten, and the test that asserted the old contract inverted.

### The contract that was, and the case for it

The thinking level is a per-request parameter. Every LLM request takes the level the Session has
pinned — the composer's picker, the CLI's `--thinking` / `/thinking`, `session.thinkingLevel` on
the SDK — and falls back to the level the context was opened with only when there is no pin. The
compaction request was the single exception: it always took the fallback.

The reason was prefix caching, and it was not a bad one. Compaction fires when the context is at
its largest, which is exactly when a cache miss is most expensive; keeping the level fixed kept the
request's prefix identical to the context it summarises.

### Why that argument inverts once the pin has moved

The exception is only ever observable when the pin differs from the context's base — that is, only
after the user has moved the pin mid-context. And in exactly that situation, every turn since the
move has already gone out at the new level, so whatever the provider has cached was rebuilt against
the new level some turns ago. The compaction request, alone in still holding the old one, was the
one request whose prefix differed. It broke the consistency it was there to protect.

The cache cost is real but it is paid at the switch, not at the compaction after it — which is what
`/thinking`'s reply has always told the user, advising `/compact` first
(`packages/cli/src/thinking-command.ts:14`). The same file's documented semantics already said core
applies the pin "from the Session's next LLM request"; the code now agrees with it.

### What the split cost

On a provider that reads the level as a thinking mode rather than as a budget, a Session running
its turns at one level and its compaction at another is a thinking mode changing mid-conversation.
A turn produced under the quieter level carries no chain of thought, and DeepSeek rejects a request
whose current tool-call chain replays such a turn back into thinking mode.

The neighbouring rule is untouched: the compaction request's **config** — its tools and system
prompt — still has to be byte-identical to a regular request's, and that is where the prefix the
provider actually caches lives.

### The spec carried this exception; design #149 removes it

`specs/05-ARCHITECTURE.md` § "上下文轮换与运行配置" stated the exception in so many words, so this
was a design change before it was a code change. Design first:
[penguin-harness-design #149](https://github.com/Prism-Shadow/penguin-harness-design/pull/149) is
open with the parenthetical removed, and this PR is the code half.

### The test was inverted, not added to

`engine.test.ts`'s "applies setThinkingLevel to every turn request — reconnects included — while
compaction requests keep the context default" asserted `["high", "high", undefined]`: two turn
requests carrying the live pin and a compaction request deliberately without it. It now asserts
`["high", "high", "high"]`, under a title and an inline comment that say why.

Inverting is the right shape rather than adding a case beside it. The old contract has no remaining
claim on the code, and an assertion that still encoded it would only be a second test disagreeing
with the first.

Because that makes the inverted assertion the sole proof of the change, it was checked against a
deliberate revert: the one-line spread was dropped from the compaction `streamGenerate` call and the
core suite re-run on `vps3-test`. One test failed, and it was that one.

```
 FAIL  test/engine.test.ts > ContextEngine live thinking level (the soft-limited runtime parameter)
       > applies setThinkingLevel to every request of the Session — reconnects and the compaction
         request included

AssertionError: expected [ 'high', 'high', undefined ] to deeply equal [ 'high', 'high', 'high' ]

- Expected
+ Received

  [
    "high",
    "high",
-   "high",
+   undefined,
  ]

 ❯ test/engine.test.ts:1029:20

 Test Files  1 failed | 53 passed | 1 skipped (55)
      Tests  1 failed | 1163 passed | 5 skipped (1169)
```

The source was then restored byte-identically — `sha256 605a4fd4d2198e96f783a11977b2c187d3ebbd037ab331182a9a6ac408d5ed3b`
on `packages/core/src/engine/context-engine.ts` before the revert and after the restore — and the
suite is green again at 1164.

`compaction.test.ts` keeps one test on the other half of the rule: with no pin, neither a turn nor a
compaction names a level at all, and the LLM object's construction default stands. It passed in both
directions (50 tests, both runs), as it has to — it asserts the unpinned case, which this change does
not reach. A positive test there duplicating `engine.test.ts`'s was dropped.

## Verification

Node v24.18.0, from `../penguin-harness-wt/compactthink`. Tests ran on the remote host `vps3-test`.

```
$ pnpm --config.verify-deps-before-run=false -r build     # exit 0
$ pnpm typecheck                                          # exit 0
All matched files use Prettier code style!                # no file changed

$ PENGUIN_TEST_HOST=vps3-test ./.agents/scripts/remote-test.sh <worktree> -- \
    --filter @prismshadow/penguin-core test
 Test Files  54 passed | 1 skipped (55)
      Tests  1164 passed | 5 skipped (1169)

$ PENGUIN_TEST_HOST=vps3-test ./.agents/scripts/remote-test.sh <worktree> \
    --build "@prismshadow/penguin-core..." -- --filter @prismshadow/penguin-server test
 Test Files  125 passed (125)
      Tests  1807 passed (1807)
```

Core ran twice green — once before the deliberate revert and once after the restore — with identical
counts both times. 1169 total is the previous 1170 less the one duplicate test removed from
`compaction.test.ts`; the server suite matches its baseline of 125 / 1807 exactly, with no flake.


hiyouga:
## Review

The change itself is right and minimal: `runCompactionRequest` now reads `this.thinkingLevel` at
request-build time and spreads it exactly as `runTurn` does, so the level is taken live, per
attempt, from the same mutable field the turns read — no closure captured at construction, no
second derivation. That respects the three-tier rule rather than breaking it: the thinking level is
the *soft* tier, explicitly allowed to move inside a context, while the strict tier this request
still holds byte-identical to a turn's (system prompt, toolset, model reference) is untouched.
Design PR [penguin-harness-design #149](https://github.com/Prism-Shadow/penguin-harness-design/pull/149)
is open and removes the spec's exception, so the pairing holds. Ship after the fixes below — the
shipped documentation still states the contract this PR removes.

### Findings

**1. The docs still tell users the opposite of what the code now does** — `packages/docs/content/agent-loop.en.md:201`, `packages/docs/content/agent-loop.zh.md:194` · *should fix*

Both published pages carried "The request runs at the thinking level its model context already
has; nothing turns thinking on for it." / "压缩请求沿用模型上下文本来的思考等级，不会为它单独开启思考。"
Concretely wrong after this PR: a Session whose Agent config resolves to `none` and whose user then
pins `high` in the composer now runs its compaction at `high` — thinking *is* turned on for it. The
three-tier paragraph two sections up (`agent-loop.en.md:193` / `zh.md:186`) never carried the
exception and already says "every LLM request takes the Session's pinned level", so the two
paragraphs contradicted each other in the same file. Rewritten in both languages to say the request
carries the pin, falling back to the level the context opened with. The same sentence appears in
`changelog/0.2.10/2026-09-02-compaction-banner-thinking.md:14`; that is a released folder and the
entry is correct past tense, so it is deliberately left frozen. ✅ Fixed in 996022da

**2. The new comment narrates the change instead of describing the code** — `packages/core/src/engine/context-engine.ts:1814-1827` · *should fix*

Fourteen lines above a one-line spread, arguing the removed design: "This request was excepted on
purpose **until now**", "The exception, though, can only ever fire…", "It broke the very consistency
it was protecting", "What the split **did cost**". `.agents/skills/penguin-harness-dev/reference/authoring.md`
names both shapes as things to delete — *Change narration* ("a comment describes what the code does
now"; past-tense narration belongs in the changelog entry, which already carries it verbatim) and
*Reviewer-addressed justification* ("defensive paragraphs arguing a choice was fine — state the
invariant, or delete"). A reader at HEAD in six months gets the history of a decision they cannot
see instead of the rule. Rewritten to eight lines that state the invariant (one Session never has
two levels in flight), keep the counterfactual-present the rules explicitly allow (the DeepSeek
rejection), and name the boundary that *is* load-bearing here — the strict-tier prefix is the system
prompt and toolset; the level is a request parameter and never was part of it. ✅ Fixed in 996022da

**3. The new `compaction.test.ts` case passes unchanged at the base commit** — `packages/core/test/compaction.test.ts:286` · *nit*

Replayed against `git show $(git merge-base origin/main HEAD):packages/core/src/engine/context-engine.ts`:
the test never calls `setThinkingLevel`, so at the base `this.thinkingLevel` is `undefined`, the turn
takes the `level !== undefined` branch and omits the key, and the compaction request set no key at
all — `llm1.levels` is `[undefined, undefined]` there too. It also cannot catch the mutation it
reads as guarding against: the assertion inspects `params.thinkingLevel`, which is `undefined`
whether the key is absent or written explicitly, so dropping the conditional spread for a plain
`thinkingLevel: level` would pass it as well. The only thing it would catch is the engine inventing
a concrete default when unpinned, which `mapThinkingLevel(override ?? default)` in
`generative-model.ts:1017` already makes inert. `engine.test.ts:1029` is the assertion that carries
this PR: it reads `params.thinkingLevel` on the third request — the compaction one — and was
`undefined` at the base, so it is behaviour, not plumbing. Left alone: a passing guard is not
harmful, and deleting a test the author deliberately kept is your call, not mine. ⬜ Left for you

**4. The Chinese changelog heading is not the documented rendering** — `changelog/unreleased/2026-09-11-compaction-thinking-level.zh.md:10` · *nit*

`changelog/README.md:79` fixes `## Details` → `## 细节`; the entry used `## 详情` (175 entries in
`changelog/` use `细节`, 8 use `详情`). ✅ Fixed in 996022da

**5. Reuse: the level is read in two places, but it is one read, not two derivations** — `packages/core/src/engine/context-engine.ts:1186`, `:1822` · *nit, not changed*

`rg -n "const level = this.thinkingLevel" packages/core/src/engine/context-engine.ts` returns
exactly those two lines, and `rg -n "streamGenerate\(" packages/core/src` shows they are the
engine's only two request sites (`session-title.ts:216` and `read-file.ts:341` run on their own
bare LLM objects, deliberately at `thinkingLevel: "none"`). Both read the same field with the same
conditional spread — one fact, no second derivation to drift — so a shared accessor would buy a
five-line abstraction used twice and force a diff into `runTurn`, which this PR does not touch.
Noted rather than changed. ⬜ Left for you

### Left for you

Making the pin ride the compaction request widens an exposure that `packages/core/src/llm/context-limits.ts:20-24`
already records as "Thinking-level interaction (considered, out of scope)". The arithmetic: the
compaction threshold is `min(max_context_length, window − COMPACTION_HEADROOM)`
(`context-limits.ts:215-219`, headroom 2048), so on any model whose window is the binding term —
every 128k/200k catalog entry under the seeded 256000 threshold — the compaction fires with ≈2048
tokens of window left, and that request's own cap derives to ≈`2048 − OUTPUT_SAFETY_MARGIN` ≈ 1024
(`context-limits.ts:191-200`). Before this PR the level going into that ~1k budget was the Agent
config's; now it is whatever the user pinned mid-session, and the pickers offer `xhigh` and `max`.
The failure shape is the documented one — reasoning consumes the cap before any `[summary]` text,
the attempt is rejected as unusable, and the compaction burns its retry ladder and ends `retryable`
at the moment the context is largest. The Web App's recommended path is not the one exposed:
"Compact, then switch" applies the level only after the compaction settles
(`packages/web/src/features/chat/chat-page.tsx:1458-1470`), so that compaction still runs at the old
pin — it is "Switch anyway", plus whatever automatic compaction fires later, that now carries the
new one. I did not change this: capping the compaction request's level,
or leaving it, is a product decision. Options as I see them: (a) ship as is — consistency is the
point, and the exposure was already reachable through `model.thinking_level`; (b) clamp the
compaction request's level to the lesser of the pin and the context base; (c) leave the code and
note the interaction at the call site. I cannot verify provider-side behaviour from here, so treat
the token arithmetic as the verified part and the provider outcome as the documented expectation.

### Verification

```
Worktree /home/cc2/zyw/penguin-harness-wt/rv-700, branch fix/compaction-thinking-level, Node v24.18.0.
pnpm install --frozen-lockfile && pnpm -r build          exit 0

Vitest ran on the remote build box vps3-test (never locally), via
  /home/cc2/.local/bin/vps3-vitest.sh <worktree> rv-700 -- pnpm --filter @prismshadow/penguin-core test
  Test Files  54 passed | 1 skipped (55)
       Tests  1164 passed | 5 skipped (1169)      exit 0

Docs content changed, so also on vps3-test:
  ... rv-700 -- pnpm --filter @prismshadow/penguin-docs test
  Test Files  7 passed (7)
       Tests  53 passed (53)                      exit 0

Local:
  pnpm typecheck        exit 0
  pnpm lint (oxlint)    exit 0
  pnpm format && pnpm format:check   "All matched files use Prettier code style!" — no file changed
  git diff --check      clean

main sync: not required. git merge-tree --write-tree origin/main HEAD is conflict-free
(single tree, exit 0). #697 moved packages/core/src/state/default-config.ts and kernel-history.ts,
which this PR does not touch, and kernel-version.test.ts hashes defaultSystemConfig() against
KERNEL_HISTORY.current — both taken whole from main by the merge — so the pinned-hash test cannot
go stale from this branch. The compaction prompt lives in the `runtime` tab's `compaction` section
and is likewise untouched.
```

PR #702, opened 2026-09-12
fix(web): ask for notification permission from a settings switch

### LINKED ISSUE #607: macOS: task-completion notifications never show — app never requests notification permission
## Summary

On macOS, the desktop app's task-completion notification **never fires**. macOS has never granted notification permission to the app: the app never requests it.

## Expected behavior

Per `packages/web/src/state/use-completion-notifications.ts`, when a task finishes while the desktop window is hidden/unfocused, a system notification should appear ("任务完成"), and clicking it opens the session.

## Actual behavior

- No system notification has ever been shown by PenguinHarness on macOS.
- The app is **absent from macOS System Settings → Notifications** (the per-app notification registry `~/Library/Preferences/com.apple.ncprefs.plist` contains 147 apps but no `com.prismshadow.penguinharness` entry). The system has never received a permission request from this bundle id.
- No notification permission record exists in the app's user data (`~/Library/Application Support/PenguinHarness`).

## Root cause

The renderer code gates on `Notification.permission === "granted"` (only then does it construct `new Notification(...)`), and the file header states the assumption:

> "there the Notification permission is granted by default"

But nothing ever requests permission:

- No `Notification.requestPermission()` call anywhere in `packages/web`.
- No notification setting/switch in the UI.
- The Electron main process (`packages/desktop/src/main.ts`) has no `setPermissionRequestHandler`/`setPermissionCheckHandler` and no `UNUserNotificationCenter` authorization request.

On modern macOS an app must actively request notification authorization (which triggers the system prompt); otherwise notifications are silently dropped and the app never appears in System Settings → Notifications. The "granted by default" assumption holds for Electron's internal permission check but not for the OS-level authorization, so the feature is dead in practice.

## Suggested fix

1. Add a notification toggle in the app settings (duplicated with the browser-side behavior or desktop-only).
2. When enabled, call `Notification.requestPermission()` from the renderer, and/or request authorization from the main process on first launch.
3. If the user denies the system prompt, show a hint directing them to System Settings → Notifications → PenguinHarness.
4. (Optional) Add a `sound` option to the notification for a completion chime.

## Environment

- PenguinHarness 0.2.9 (Electron, arm64)
- macOS 26 (Darwin 25.6.0), zh-CN locale
- Reproduced on 0.2.7 / 0.2.8 / 0.2.9 (same code path in all three)

hiyouga:
## Root cause

`packages/web/src/state/use-completion-notifications.ts` showed a notification only when
`Notification.permission === "granted"`, and its header stated the premise it was built on:
in the desktop shell "the Notification permission is granted by default". It is not.
Nothing in `packages/web` ever called `Notification.requestPermission()`, so the permission
stayed at `"default"` for the life of every install and the gate never opened once.

On macOS the consequence is worse than a silent feature: a bundle that never requests
authorization is never registered with the notification centre, so the app is absent from
System Settings -> Notifications and the user cannot allow it by hand either. The reporter
confirmed `com.prismshadow.penguinharness` is missing from `~/Library/Preferences/com.apple.ncprefs.plist`
(https://github.com/Prism-Shadow/penguin-harness/issues/607).

## The opt-in, and why the prompt hangs off the toggle

New module `packages/web/src/lib/notification-pref.ts` owns the two facts that decide
whether a notification can be shown, neither of which implies the other: the user asked for
notifications (`penguin.notifications`, **off by default**, registered as a `browser`
preference in `install-scope.ts`), and the platform grants permission right now. It follows
the `dock-launcher-state.ts` idiom — module-level version counter read through
`useSyncExternalStore`, value read separately — so the settings row and the app-wide hook
stay in step.

Permission is requested **at the moment the switch is turned on**, never at launch. A
permission dialog nobody asked for is the one users dismiss reflexively, and on macOS that
dismissal is final: `requestPermission()` never prompts again and the only recovery is
System Settings. Asking on the toggle means the dialog answers a question the user just
asked, and the default-off preference means it is never asked of someone who did not.

The row (System settings > General) settles on what the request answered rather than on
what was clicked:

- `"granted"` — the preference is stored and the switch goes on.
- `"denied"`, or a dismissed prompt that leaves `"default"` — nothing is stored, the switch
  does not latch on, and `PrefRow`'s `hint` points at the system's own notification settings
  (a fact about current state, which is what `hint` is for per `setting-row.tsx`).
- no `Notification` API at all — the row is disabled with its own hint.
- turning it off stores `false`; it does not and cannot revoke the OS permission.

## The desktop gate is removed

`sessionVia === "desktop"` existed for exactly one reason, stated in the old header: plain
browsers "would face a permission prompt". With an explicit opt-in that reason is gone, and
keeping the gate would leave the new switch visible but inert in a browser. The hook now
gates on the stored preference plus a live read of `Notification.permission` — re-read on
every completion, so a permission revoked while the app runs stops it — and its header no
longer carries the false premise.

## Verification

Run in `../penguin-harness-wt/notifyperm` on Node v24.18.0; tests on the shared test host.

| Command | Result |
| --- | --- |
| `pnpm --config.verify-deps-before-run=false -r build` | PASS (exit 0) |
| `pnpm typecheck` | PASS (exit 0, no diagnostics) |
| `pnpm format` then `pnpm format:check` | PASS — "All matched files use Prettier code style!" |
| `pnpm --filter @prismshadow/penguin-web test` (remote) | PASS — 135 files, 1860 tests, 0 failures, 2.71s |

New `packages/web/test/notification-pref.test.ts` (10 tests): the stored preference's
default-off/read/write/subscriber-notification and its behaviour against a throwing store;
the permission outcomes — granted stores the opt-in, denied and dismissed do not, a platform
without the API reports `unsupported` and is asked nothing. `install-scope.test.ts` (27
tests) covers the new key: it walks `packages/web/src` for `penguin.*` literals and fails on
one `KEY_RULES` does not classify. The existing `completion-notify.test.ts` tracker tests are
untouched.

Not covered by CI, and not reproducible on Linux: the macOS system dialog itself, and the
app's appearance in System Settings -> Notifications after the first granted request.

Closes https://github.com/Prism-Shadow/penguin-harness/issues/607


hiyouga:
## Review

The PR moves task-completion notifications from a dead `sessionVia === "desktop"` gate onto an
explicit opt-in that requests the OS permission from the switch itself, which is the right shape:
the permission is asked for on a gesture that already says yes, stored only when the request comes
back granted, and re-read live at the point of notifying rather than trusted from the moment it was
granted. Two defects in how the row reports a request that did not succeed, plus a stale-tracker
ordering bug the dynamic `enabled` introduced, are fixed below; the whole web suite passes. One
product call — what happens to desktop installs whose notifications did work — is left to you.

### Findings

**1. A prompt closed without an answer sprang the switch back with nothing said** — `packages/web/src/features/settings/general-section.tsx:52` · *should fix*

`notificationHint` covered `"unsupported"` and `"denied"` only. Concrete case: Chrome or Firefox at
`Notification.permission === "default"`, the user turns the switch on, the prompt opens, they close
it without choosing (Esc, or the `X`). `requestPermission()` resolves `"default"`, so
`enableNotifications()` stores nothing and `setAccess("default")` selects no hint — the switch
slides back to off with no line under it. That is the one outcome with a recovery the user cannot
guess (ask again, unlike a denial), and the entry's own Details bullet claimed it was covered.
`lib/notification-pref.ts` now exports `notificationHintFor(access, asked)`, which separates a
platform that has simply not been asked (every first visit, no hint) from a request that came back
unanswered, and `S.settings.notificationsDismissed` is the third line. ✅ Fixed in 584c4545

**2. The completion tracker stopped observing while the preference was off** — `packages/web/src/state/use-completion-notifications.ts:45` · *should fix*

`enabled` used to be `sessionVia === "desktop"` — constant for the life of the mount, so returning
before `trackerRef.current.observe(...)` cost nothing. This PR made it a live preference and left
the return where it was. Concrete case: notifications on, Session A running; the user switches them
off; A finishes; later they switch them back on. Nothing observed in between, so the tracker's entry
for A still reads `active: true`, and the first `observe` after re-enabling reports A as a
completion — a "Task completed" for a run that ended minutes ago. In practice the focus gate two
lines down usually swallows it, since the switch lives in the settings dialog; what is unambiguous
is that the early return contradicts the comment directly under it ("The tracker must see every
snapshot"). Now it observes first and gates afterwards. ✅ Fixed in 584c4545

No test: `packages/web` vitest is `environment: "node"` with no DOM, so the hook is not renderable
here. `test/completion-notify.test.ts` already covers the tracker behaviour the fix relies on
("fires when a known running session goes idle", "never fires on first observation").

**3. A revoked permission leaves the switch reading on** — `packages/web/src/features/settings/general-section.tsx:37` · *nit*

`checked={notificationsOn}` is the stored preference alone. Concrete case: the user granted and
switched on; later they set Notifications to Block for this origin in browser site settings.
Reopening System settings › General shows the switch **on**, with `notificationsDenied` under it
ending "…then turn this on again". Nothing misfires — the completion path re-reads the live
permission and stays silent — but the row displays the two facts disagreeing, in copy that assumes
the switch is off. Both ways out are product calls, so this one is yours: render
`checked={notificationsOn && access === "granted"}` (honest, but the stored `true` then cannot be
cleared from the UI), or reword `notificationsDenied` so it reads correctly in either position.
⬜ Left for the author

**4. The entry's lead paragraph named the wrong mechanism** — `changelog/unreleased/2026-09-12-notification-permission.md:11` · *should fix*

It said the permission stayed at `"default"`, so the `permission === "granted"` gate never opened
and nothing was ever shown. That gate only ever ran under `sessionVia === "desktop"`, and the issue
being closed records the opposite there: "The 'granted by default' assumption holds for Electron's
internal permission check but not for the OS-level authorization." So the gate *did* open in the
shell — macOS dropped the notification a layer below it — and a browser never reached the gate at
all. Rewritten in both files. ✅ Fixed in 584c4545

**5. The new browser key was missing from install-scope's fixture** — `packages/web/src/lib/install-scope.ts:200` · *nit*

`KEY_RULES` gained `penguin.notifications`, but `test/install-scope.test.ts`'s `populated()` and
`PREFERENCE_KEYS` — where every one of the other seventeen browser preferences is listed — did not,
so nothing asserted that a data-root sweep keeps it. Added; both assertions fail against the merge
base (see Verification). ✅ Fixed in 584c4545

**6. The rule's `why` line described a recovery that cannot happen** — `packages/web/src/lib/install-scope.ts:204` · *nit*

"clearing it would make the app prompt for permission all over again" — clearing the preference
cannot re-prompt anyone: the OS permission survives it, so the next `requestPermission()` resolves
`"granted"` with no dialog. The `browser` scope is right for the other reason the same line gives;
the clause was replaced with what clearing actually costs. ✅ Fixed in 584c4545

### Left for you

**1. Desktop installs whose notifications did work lose them silently.** The preference defaults to
off and the `sessionVia === "desktop"` gate is gone, so a shell install that was showing completion
toasts stops until someone finds System settings › General. Nothing already on disk is
reinterpreted — `penguin.notifications` is a new key and an absent value has always meant off — so
**no `changelog/unreleased/YYYY-MM-DD-backward-compatibility.md` pair is owed as the PR stands**.
Three ways forward:

- accept it, recorded as-is (the entry now carries a line saying so) — zero compatibility cost;
- default the preference on when `sessionVia === "desktop"`, reading it rather than writing it — no
  stored value, but the switch then starts on for a browser-less shell and off elsewhere;
- write `true` once for desktop sessions on upgrade — this *does* put a value on disk and would
  need the backward-compatibility entry, naming how long the one-shot migration stays and who
  removes it.

**2. Plain-HTTP deployments.** `penguin server` is commonly reached at `http://<lan-ip>:<port>`,
which is not a secure context, and browsers refuse notification permission there outright. The row
would then show `notificationsDenied`, sending the user to their system notification settings where
there is nothing to change. I could not exercise this here (no browser in the harness), so it is
unverified — but if LAN-over-HTTP is a deployment you support, it probably wants either its own
line or a sentence in the docs.

### Verification

```
Node v24.18.0, worktree ../penguin-harness-wt/rv-702 at 584c4545.
vitest ran on the vps3 build box (vps3-vitest.sh); nothing ran vitest locally.

pnpm install --frozen-lockfile && pnpm -r build     PASS (exit 0)
pnpm typecheck                                      PASS (exit 0, all 11 projects)
pnpm lint                                           PASS (oxlint, exit 0, no diagnostics)
pnpm format && pnpm format:check                    PASS ("All matched files use Prettier code style!")
git diff --check                                    PASS (clean)
pnpm --filter @prismshadow/penguin-web build        PASS (exit 0, built in 4.97s)

[vps3] pnpm --filter @prismshadow/penguin-web test
  before the fixes   135 files, 1860 tests, 0 failures
  at 584c4545        135 files, 1864 tests, 0 failures

Replay of the new assertions against the pre-fix source (new tests kept, notification-pref.ts
restored to caf7facd and install-scope.ts to the merge base), on vps3:
  [vps3] vitest run test/notification-pref.test.ts test/install-scope.test.ts
  2 files failed, 5 tests failed | 36 passed — the three notificationHintFor cases and
  install-scope's "classifies every penguin.* key" and "covers every key the populated
  fixture holds". Source restored afterwards; the tree is clean at 584c4545.

packages/web/e2e not run (Playwright is not installed on the build box).

Sync with main: not needed. `git merge-tree origin/main HEAD` merges cleanly. #699 does not
touch general-section.tsx (`git diff <base> origin/main -- …/general-section.tsx` is empty);
it adds that module to control-size.test.ts's DIALOG_BODY_MODULES, but findLooseDialogBodyButtons
only inspects `Button` JSX tags and this module contains none (`grep -c '<Button'` = 0, before
and after these fixes), so that guard passes on the merge result. The only files both this branch
and main changed are strings.ts / strings-en.ts, and the additions do not overlap.
```

PR #703, opened 2026-09-12
fix(core,cli): AgentHub 0.4.14 so a DeepSeek tool chain survives a strict Responses relay and the turns the model thought nothing on

### LINKED ISSUE #404: Bug: DeepSeekV4Client (client_type=deepseek-v4) → 400 invalid_json after any tool call — function_call_output.output sent as array
### Impact

Using a DeepSeek V4 model configured with `client_type=deepseek-v4` (which drives `DeepSeekV4Client` over the **Responses API**), any session that performs **one tool call** will fail on the **very next** model request with:

```
400 {"code":"invalid_json","message":"Invalid JSON"}
```

It retries 5 times and then aborts, leaving the session stuck. This is reproducible across many sessions, always "tool call → next request fails."

### Environment

- PenguinHarness 0.2.4 (desktop / web, same server kernel)
- Model: `deepseek-v4-flash-0731` via a third-party relay that **strictly conforms to the DeepSeek Responses spec** (also supports Chat Completions, but this bug is on the `/responses` path)
- `client_type=deepseek-v4`

### Root cause (suspected kernel regression, please verify)

In `DeepSeekV4Client.transformUniMessageToModelInput`, the `tool_result` branch (around line 158995 in `dist/server.js`) serializes `function_call_output.output` as an **array**:

```json
[{"type": "input_text", "text": "..."}]
```

But the DeepSeek **Responses spec expects `output` to be a string**. A strictly-validating relay returns `invalid_json` because of this shape. (The array form is appropriate for `input_text`-style message items, not for `function_call_output.output`.)

### Why it's a 0.2.3 → 0.2.4 regression

- **0.2.3** (`node_modules/@prismshadow/agenthub/dist/deepseek_v4/client.js`): `deepseek-v4` client used **Chat Completions** (`chat.completions.create`), tool results as `{role:"tool", tool_call_id, content:[{type:"text",text}]}` — no `function_call_output`. Worked fine.
- **0.2.4** (`dist/server.js`): the client was rewritten to use **Responses API** (`responses.create`, with `input`/`function_call`/`function_call_output`), and `function_call_output.output` is now sent as an array. The changelog for 0.2.4 confirms the migration "from Chat Completions to the OpenAI Responses API."

So the bug only appears after upgrading to 0.2.4, and only for `client_type=deepseek-v4` clients (openai-chat / Chat Completions path is unaffected).

### Controlled reproduction & evidence

- Same `/responses` request, only change: `output` **array → string** → relay returns **200** (streaming and non-streaming both work, tool round-trip fine).
- Keep `output` as **array** → relay returns **400 invalid_json** — byte-for-byte the same error seen live.
- 0.2.3 (Chat Completions) works; 0.2.4 (Responses, array output) reproduces.

Thus the relay itself is fine; the harness is emitting the wrong shape for `function_call_output.output`.

### Request

Please serialize `function_call_output.output` as a **string** per the Responses spec (e.g. `item.text`, or `content`'s text). Until an official fix, we are working around it with a local reverse proxy that rewrites that array to a string — so it's not blocking, but it is a real bug that will hit any strictly-validating `/responses` endpoint when tool calls are used.

### Additional context

- The 0.2.4 changelog's note about "entries pointing to Chat-Completions-only endpoints need a `client_type` set" is an orthogonal warning for endpoints that don't support `/responses`; in our case the endpoint **does** support `/responses` — the failure is specifically the `output` being sent as an array.

hiyouga:
Closes #404. `@prismshadow/agenthub` moves from 0.4.12 to 0.4.14 in `packages/core` and `packages/cli`, and `pnpm-workspace.yaml`'s `minimumReleaseAgeExclude` names 0.4.14. Two DeepSeek failures end with it.

## 0.4.14 — agenthub [#215](https://github.com/Prism-Shadow/agenthub/pull/215)

A row pinned to `deepseek-v4` — the direct `deepseek-flash` preset, or a relay entry with `client_type = "deepseek-v4"` as in #404 — died on the request after its first tool call with `400 {"code":"invalid_json"}` behind a strictly validating Responses endpoint: the `deepseek_v4` (and `gpt6`) client wrapped every tool result in a one-part content list, the multimodal form, and the engine classified the rejection as a provider 4xx, `fatal`, so the Task stopped where it stood. Both clients now send a text-only tool result as a plain string and keep the list only for a result that carries images — the rule the generic `openai-responses` / `openai-chat` clients took in 0.4.11 — so every Responses and Chat Completions client AgentHub ships behaves alike.

## 0.4.13 — agenthub [#212](https://github.com/Prism-Shadow/agenthub/pull/212)

A task run against DeepSeek over the `openai-chat` protocol no longer dies part-way through with `400 The reasoning_content in the thinking mode must be passed back to the API. (invalid_request_error)`: the client sends the reasoning field as an empty string on a tool-calling turn the model thought nothing on. DeepSeek requires the chain of thought back only when it cannot recover it from a `tool_call` id it issued itself, so this hits endpoints that reissue ids (relays, custom OpenAI-compatible entries in front of DeepSeek), not a direct `provider = "deepseek"` row.

## Lockfile

Re-resolved with `pnpm install --lockfile-only` on the test host. pnpm's `minimumReleaseAge` policy rejects the lockfile's 0.4.13 entry the moment the exclude list stops naming it, so the resolution ran with both versions excluded and the 0.4.13 line was dropped afterwards; the second `--lockfile-only` pass left the file byte-identical and the policy check passes on 903 entries.

## Verification

Remote test host (Node 24.20): `pnpm install --frozen-lockfile`, `pnpm build` of core / server / cli and their dependencies, `pnpm typecheck` across every package — clean; core `test/model-catalog.test.ts` 37 passed; the installed `@prismshadow/agenthub` under `packages/core/node_modules` reports 0.4.14.


hiyouga:
## Review

A two-release dependency bump: `@prismshadow/agenthub` 0.4.12 → 0.4.14 in `packages/core` and `packages/cli`, with `minimumReleaseAgeExclude` and the lockfile moved to match. I audited the bump for completeness against the upstream releases: every pin site in the tree says 0.4.14, the lockfile re-resolved nothing else, and both behaviour claims match what agenthub actually shipped. Ship after the two changelog-contract fixes below, which are already pushed as `d54d552f`.

### Findings

**1. The second cross-repository reference is not named, and `#215` collides with a live issue here** — `changelog/unreleased/2026-09-12-agenthub-0.4.14.md:11`, `.zh.md:11` · *should fix*

The lead cited the two upstream PRs as `(agenthub [#212](…/agenthub/pull/212) in 0.4.13, [#215](…/agenthub/pull/215) in 0.4.14)` — the repository is named only on the first. `changelog/README.md:54` and `.agents/skills/penguin-harness-dev/reference/changelog.md:37` both require a cross-repo reference to name its repo, because a number in prose reads as this repository's. That is not hypothetical here: `Prism-Shadow/penguin-harness#215` is a live issue ("Trace JSONL corruption caused by concurrent writes of large multimodal messages"), so a reader scanning the rendered text sees "#215" and associates the 0.4.14 fix with an unrelated bug. The house precedent for exactly this shape — one entry citing two agenthub PRs across two releases — names it on both: `changelog/0.2.3/2026-08-19-agenthub-0.4.4.md:12`. ✅ Fixed in `d54d552f`

**2. The `.zh.md` heads its section `## 详情` where the contract specifies `## 细节`** — `changelog/unreleased/2026-09-12-agenthub-0.4.14.zh.md:13` · *nit*

`changelog/README.md`'s heading-rendering table and `.agents/skills/penguin-harness-dev/reference/changelog.md:19` both map `## Details` → `## 细节`, and the skill names an untranslated-or-drifted heading as "the most common way the pair stops mirroring". The tree agrees: 175 `.zh.md` entries use `## 细节` against 9 that drifted to `## 详情`, and the immediately preceding entry in this same series — `changelog/unreleased/2026-09-11-agenthub-0.4.12.zh.md:12` — uses `## 细节`. ✅ Fixed in `d54d552f`

**3. The entry's title and lead scope the 0.4.14 fix to DeepSeek, but it also covers `gpt-6-astra`** — `changelog/unreleased/2026-09-12-agenthub-0.4.14.md:1,11` · *nit*

agenthub 0.4.14 changed both first-party Responses clients, `deepseek_v4` **and** `gpt6`. The `## Details` bullet says so correctly, so nothing is missing — but the H1 and the lead sentence name only `deepseek-v4` rows. `packages/core/src/state/model-catalog.ts:1761` is a direct `gpt-6-astra` row and `model-catalog.ts:2102` routes any `gpt-6` id to the `gpt6` client, so a `gpt-6-astra` row pointed at a strictly validating relay would hit the same `400 {"code":"invalid_json"}` and is equally covered by this bump. (I have no report of one; the wire shape is identical, which is why upstream changed both clients in one PR.) Someone searching the changelog for that symptom on GPT-6 will not find this entry by its title. Widening the lead is a framing call — upstream's own release notes lead with DeepSeek too — so I left it alone. ⬜ Left for the author

### Verification

Completeness audit — all clean, recorded so it need not be redone:

- **Every pin site carries 0.4.14, and only 0.4.14.** `rg` across the tree for `@prismshadow/agenthub` and for every `0.4.x` string finds exactly three live pins: `packages/core/package.json:72` (`^0.4.14`), `packages/cli/package.json:39` (`^0.4.14`), `pnpm-workspace.yaml:64` (`@prismshadow/agenthub@0.4.14`). No Dockerfile, CI workflow, script or test asserts an agenthub version; every other hit is historical prose in frozen changelog folders. No `0.4.12` or `0.4.13` survives anywhere in the tree.
- **The branch name is stale, the tree is not.** The branch is `chore/agenthub-0.4.13` and its first commit did bump to 0.4.13, but the second commit moved every site to 0.4.14. The name is squashed away at merge; there is no second version in the tree.
- **`minimumReleaseAgeExclude` moved.** `pnpm-workspace.yaml:64` names 0.4.14 in place of 0.4.12, so a clean clone can install a release published today. No stale entry left behind.
- **Lockfile.** The diff is 11 added / 11 removed lines, all `@prismshadow/agenthub`: both importers, the package entry with its new integrity hash, both peer-variant snapshots, and the three `file:packages/*` snapshots. Nothing unrelated was re-resolved, and no orphan 0.4.13 entry remains.
- **The claims hold.** agenthub [#212](https://github.com/Prism-Shadow/agenthub/pull/212) (0.4.13) and agenthub [#215](https://github.com/Prism-Shadow/agenthub/pull/215) (0.4.14) touch only client internals plus their own tests and changelogs — no public API or type change — so no harness-side code had to move with the bump. The 0.4.13 fix builds its `replayFields` set per request from the message list it is handed rather than from client-instance state, so it engages regardless of `GenerativeModel` constructing a fresh `AutoLLMClient` (`packages/core/src/llm/generative-model.ts:986`), and core replays `thinking` payloads with `fidelity` verbatim (`generative-model.ts:151`), which is what the fix keys off. `deepseek-flash` is indeed a direct row on the `deepseek-v4` client (`packages/core/src/state/model-catalog.ts:497-505`), so the entry's enumeration of affected rows is accurate, and 0.4.14's "keeps refusing images on the ids whose model reads none" leaves `changelog/unreleased/2026-09-11-deepseek-v4-flash-text-only.md` still correct.
- **No index file was added**, correctly — agenthub's own changelog workflow has one and this repository deliberately does not.
- **No sync needed.** `main` has not touched `pnpm-lock.yaml`, `pnpm-workspace.yaml` or either manifest since this branch's merge base, and `git merge-tree --write-tree origin/main HEAD` is conflict-free.

```
# vitest on the vps3 build box (never locally)
vps3-vitest.sh … rv-703 -- pnpm --filter @prismshadow/penguin-core test
    Test Files  54 passed | 1 skipped (55)
    Tests       1163 passed | 5 skipped (1168)            exit 0
vps3-vitest.sh … rv-703 -- env FORCE_COLOR=0 pnpm --filter @prismshadow/penguin-cli test
    Test Files  30 passed (30)
    Tests       394 passed (394)                           exit 0

# locally, in the worktree
pnpm install --frozen-lockfile                   pass
pnpm -r build                                    pass
pnpm typecheck                                   pass (every package)
pnpm lint                                        pass (oxlint, 0 findings)
pnpm format && pnpm format:check                 pass (nothing reformatted)
git diff --check                                 clean
git merge-tree --write-tree origin/main HEAD     merges cleanly
installed @prismshadow/agenthub version          0.4.14 under packages/core and packages/cli
```

PR #697, opened 2026-09-11
Guardrails and interaction rules in the default system prompt

hiyouga:
Fourteen edits to `DEFAULT_SYSTEM_PROMPT`, each a clause or one bullet and several of them replacing an existing sentence, taken from the published [Claude Code](https://github.com/x1xhlol/system-prompts-and-models-of-ai-tools/blob/main/Anthropic/Claude%20Code/Prompt.txt), [Qoder](https://github.com/x1xhlol/system-prompts-and-models-of-ai-tools/blob/main/Qoder/prompt.txt) and [CodeBuddy Craft](https://github.com/x1xhlol/system-prompts-and-models-of-ai-tools/blob/main/CodeBuddy%20Prompts/Craft%20Prompt.txt) prompts. The brief: make an agent finish the user's task accurately and fast instead of looping or delivering something that does not exist; then, widened: anything meaningful, including what makes the interaction better, replacing existing text rather than only adding; then, trimmed: fewer restrictions; then, structured: the output rules gathered into an `# Output` section in the slot PRN-010 gives it, plus a link rule. Smallest possible change per item, a reason for each.

The three sources were read in full. Most of their text is tool-specific (XML tool formats, `search_replace` line limits, Supabase integrations) or contrary to this project's principles (identity refusal, prompt-leak decoys); what survives are the behavioural rules below. Design counterpart: penguin-harness-design [#148](https://github.com/Prism-Shadow/penguin-harness-design/pull/148).

## A. Accuracy, loops, phantom deliverables

Ordered by expected impact. Each entry: what changed, where it comes from, why it earns a line on every turn.

**1. `# Stop rules` — an observable trigger for "cannot resolve".**
`If you hit an error you cannot resolve` → `… — the same error still there after three different fixes — stop and report the blocker to the user, with what you tried; do not keep trying variants.`
Source: CodeBuddy *Objective* §4 ("a task that has failed multiple times … ask the user"). Why: the old rule had no observable trigger, so a stubborn error was met with an open-ended run of variants — the loop the brief describes. "Same error after three different fixes" is a signal the model can check, and an error that *changes* is progress, so debugging is not cut short. Three matches the goal protocol's `blocked` threshold (05 § Goal 模式). The API-key special case (retry once) is unchanged and still stricter.

**2. `# Tool use` — send independent tool calls together.**
New bullet: `Send independent tool calls together in one turn — several file reads, unrelated checks — they run concurrently; a call that needs an earlier result, or writes the same file as another, waits for the next turn.`
[...]

PR #705, opened 2026-09-12
feat(core,web,cli,docs): every OpenRouter model speaks the Responses API

hiyouga:
Every OpenRouter row in the built-in catalog now pins `client_type = "openai-responses"` (33 rows moved from `openai-chat`; the ten `openai/*` rows already had it), and the OpenRouter provider group pins the same protocol at group level — the mechanism the vLLM group already uses — so a model a user adds to that group takes it too. The models dialog offers gateway groups no protocol picker, so the group default is the only protocol a user-added OpenRouter entry can get; without the group pin the presets and the user's own rows would have spoken different protocols with no way to see or change it.

OpenRouter serves the Responses API at the same `https://openrouter.ai/api/v1` base for every upstream. The earlier limitation to the `openai/*` rows rested on the generic Responses client's reasoning-item round-tripping; since AgentHub 0.4.11 that client sends a text-only tool result as a plain string and replays reasoning items only where the upstream returned them, which is what made the maintainer's decision to move the whole group workable.

Design half: [penguin-harness-design #151](https://github.com/Prism-Shadow/penguin-harness-design/pull/151).

## What changed

- `packages/core/src/state/model-catalog.ts`: the 43 OpenRouter rows, the group pin, and the header / group / OpenRouter block comments.
- Web: the add dialog's hint for a pinned group with a preset base URL (new `addProtocolHintPinnedGateway` in both dictionaries — the vLLM wording "set the base URL to your own server" was wrong for a gateway), plus comments in `protocol-path.ts` and `models-page.tsx`.
- CLI: `config model add` already takes the group pin through `providerClientType`; comments only.
- Docs: the gateway paragraph of `models.{en,zh}.md`.
- Tests: the catalog roster of group pins (`["openrouter", "vllm"]`), the OpenRouter gateway assertions, `server/test/models.test.ts` (the catalog `xiaomi/mimo-v2.5` row), `cli/test/config-model.test.ts` (adding to the OpenRouter group), `web/test/protocol-path.test.ts` and two new lines in `web/test/protocol-detect.test.ts` pinning `clientTypeAfterProviderChange("openrouter", …)`.

## Existing Projects

Presets are copied into `.project_config.toml` at creation and nothing rewrites them: stored OpenRouter rows keep `openai-chat` until the owner runs "sync presets" on the models page, which rewrites the catalog-owned fields (`client_type` is one of them, see `catalog-sync.ts`) and leaves credentials alone. The sync badge counts those rows meanwhile. No compatibility code was added.

## Not verified live

Both OpenRouter keys available to this session answer `401 User not found`, so the Responses path was not exercised against OpenRouter with a non-OpenAI upstream in this PR.

## Verification

On the remote test host (`.agents/scripts/remote-test.sh`, Node 24.20): `pnpm typecheck` clean after building core, server, web and cli; `pnpm format:check` clean.

- core `test/model-catalog.test.ts` + `test/state.test.ts`: 26 passed
- server `test/models.test.ts`: passed
- cli `test/config-model.test.ts`: 20 passed
- web full suite: 134 files, 1850 passed
- docs: 7 files, 53 passed


hiyouga:
## Review

All 43 OpenRouter catalog rows move to `openai-responses` and the group gains a `clientType` pin, so user-added rows inherit it too. The claim in the title holds: I counted the rows at the merge base (33 `openai-chat` + 10 `openai-responses`) and on HEAD (43 `openai-responses`), the new `for (const m of or)` assertion is what stops a later row from being added on the wrong protocol, and every place that decides a protocol already reads `providerClientType` — `protocol-types.ts:114,158` (`protocolForPersist`, `envHintClientType`) picks the pin up without being touched, and `providerClientType` is just `providerInfo(id)?.clientType`, so the server's direct `group.clientType` read at `project-config-service.ts:127` cannot drift from it either. The row pin and the group pin are two statements of one fact, but they have to be (the row pin is what gets copied to disk, the group pin is for ids with no row) and the two new catalog assertions pin them together. One thing the PR's analysis does not cover changes the request shape on the wire and is worth a decision before this ships; two stale-prose misses are fixed below.

### Findings

**1. The Responses client sends a `reasoning` object that the Chat Completions client never sent, and two core paths hard-code the one effort OpenRouter is documented to refuse** — `packages/core/src/state/model-catalog.ts:559` · *needs a decision from the author*

The PR's justification covers reasoning-item *replay* and tool-result shape. The outbound direction is the part that changes and is not addressed:

- `openai_chat/client.js` has **zero** references to `thinking_level` — its `transformUniConfigToModelConfig` sends no effort parameter at all.
- `openai_responses/client.js:110-117` (0.4.12, the version this repo installs) sets `reasoning: { effort }` (plus `summary: "concise"`) whenever `thinking_level` is present.
- Core hard-codes `thinkingLevel: "none"` on two paths: `createBareLLM`, used for meta requests such as `generateTitle` (`packages/core/src/agent.ts:1270`), and the vision-describer proxy that `read_file` uses for text-only session models (`packages/core/src/agent.ts:1125`). `mapThinkingLevel("none")` → `ThinkingLevel.NONE`, and `_convertThinkingLevelToEffort` maps NONE to `"none"` for every model except `gpt-6*`.
- AgentHub 0.4.12's own comment in that client (`openai_responses/client.js:106-109`) records that OpenRouter "reads a reasoning object carrying no effort as 'reasoning disabled' and refuses it on a forced-thinking model — 'Reasoning is mandatory for this endpoint and cannot be disabled' (400, verified live 2026-09-03 with z-ai/glm-5.3)". `z-ai/glm-5.3` is one of the 43 rows this PR moves.

Scenario: a Project on `openrouter` / `z-ai/glm-5.3`, created after this lands or synced. Before, title generation posted to `/chat/completions` with no reasoning field and succeeded. After, it posts `reasoning: {effort: "none", summary: "concise"}` to `/responses`. `title-generator.ts:177` catches the failure, so the visible result is that the session keeps its fallback title and records a `title_llm_failed` error — and because the generator retries while the stored title still equals the fallback, that repeats on every Task start. The vision-describer proxy at `agent.ts:1125` is the same story for `read_file` images, and there the failure is not swallowed.

**The last link is unverified**: AgentHub's note is about a reasoning object carrying *no* effort, and whether OpenRouter treats an explicit `effort: "none"` the same way I could not test — the PR reports both available OpenRouter keys answer 401. I am flagging the shape of the change, not asserting the 400. Options, none of them mine to pick: accept it; have AgentHub degrade NONE to `"low"` on the generic Responses client the way it already does for `gpt-6`; or leave `thinking_level` off the two hard-coded "none" paths. ⬜ Left for the author

**2. The `openai/*` rows are no longer "the only gateway rows on the Responses protocol"** — `packages/core/src/state/model-catalog.ts:826` · *should fix*

The section comment above those rows still claims they are the only OpenRouter rows on Responses, and cites the file's own header block — which this PR rewrote to say the opposite. Every other protocol comment in the file was updated; this one was missed. Dropped the false clause, keeping the context-window sentence. ✅ Fixed in 56977c7

**3. The shipped `unified-llm-api` skill still scopes OpenRouter's Responses support to `openai/*` ids** — `plugins/agent-development/skills/unified-llm-api/SKILL.md:111` · *should fix*

"OpenRouter serves both protocols … so its `openai/*` ids work with `clientType: "openai-responses"` as well as `"openai-chat"`" now contradicts the catalog, which pins `openai-responses` on `z-ai/glm-5.3-flash` and 32 other non-`openai/*` rows. An agent reading the skill would conclude those rows are misconfigured. Generalised the sentence to every id the gateway resells, and named it in the changelog pair's docs bullet. I deliberately did **not** bump `plugins/agent-development/plugin.json`'s `version`, and did not add `skills` to the entry's `Scope` — see "Left for you". ✅ Fixed in 56977c7

**4. `protocol-path.test.ts` dropped the one assertion covering a stored, pre-sync OpenRouter row** — `packages/web/test/protocol-path.test.ts:37` · *nit*

`protocolPathForModel("openrouter", "openai-chat")` still returns `/chat/completions` — an explicit client type wins over group membership, so the removed line would still pass. It was the only assertion pinning the suffix the dialog shows for an OpenRouter row that is *still on `openai-chat`*, which by this PR's own compatibility note is what every existing Project holds until its owner syncs. The branch stays covered by the other gateways, so this is coverage of a meaning rather than of a code path. Not restoring it myself. ⬜ Left for the author

### Left for you

**What happens to `.project_config.toml` rows written before this change.** I verified the PR's account and it is accurate: `presetModelEntries()` reaches disk only through `defaultProjectConfig()`, which `loadProjectConfig` uses on ENOENT and `writeInitialConfig` on creation, and `ensurePresetModels` backfills only a Project with no models at all. Nothing rewrites a stored row. What the PR does not spell out:

- `clientType` is a catalog-owned field in `catalog-sync.ts:30`, so every existing Project carrying the OpenRouter presets now shows the Models badge with 43 rows out of date, and one press of **sync presets** rewrites all 43 to `openai-responses`.
- Gateway groups get no protocol picker: `models-page.tsx:2568` gates `showProtocolSelector` on `customLikeGroup`, which `openrouter` is not. If finding 1 turns out to bite after a sync, the way back is not in the Web App — it is `penguin config model add --provider openrouter --model-id <id> --client-type openai-chat` (an explicit `--client-type` updates an existing entry) or editing the TOML by hand.

Options: ship as is; or mark the entry `Breaking: yes` with a `## Compatibility` section naming that recovery command, the way `changelog/0.2.4/2026-08-21-agenthub-0.4.6.md` did for the DeepSeek protocol move; or give gateway rows the picker so the escape hatch exists in the UI (a product change, out of this PR's scope). For the record, **no `changelog/unreleased/YYYY-MM-DD-backward-compatibility.md` pair is owed** — the repo's rule creates one only for a batch that adds compatibility handling, and this PR adds none.

**The skill edit's shipping half.** `changelog/0.2.10/2026-09-09-gpt-6-astra.md` bumped `plugins/agent-development/plugin.json` to `2026-09-09.1` after editing the same skill "so installed copies are offered the update"; PR [#649](https://github.com/Prism-Shadow/penguin-harness/pull/649) changed that file with no bump. A bump notifies every installed copy, so I left it to you. Adding `skills` to the entry's `Scope` would make it six entries against the README's cap of five, so that needs a swap rather than an append.

### Verification

```
worktree: ../penguin-harness-wt/rv-705 @ feat/catalog-openrouter-responses
base:     b16ce66a (git merge-base origin/main HEAD)
pnpm install --frozen-lockfile && pnpm -r build      # exit 0

vitest on vps3 (never on the review box), via
/home/cc2/.local/bin/vps3-vitest.sh <worktree> rv-705 -- <cmd>
  core     54 files, 1163 passed, 5 skipped   exit 0
  server  125 files, 1807 passed              exit 0
  web     134 files, 1850 passed              exit 0
  cli      30 files,  394 passed              exit 0   (FORCE_COLOR=0; the shell exports 3)
  docs      7 files,   53 passed              exit 0

local:
  pnpm typecheck      PASS (23 projects)
  pnpm lint           PASS (oxlint, exit 0)
  pnpm format         no files changed
  pnpm format:check   PASS
  git diff --check    clean

row counts, by parsing the catalog at each ref:
  merge base  43 openrouter rows: 33 openai-chat + 10 openai-responses (all 10 openai/*)
  HEAD        43 openrouter rows: 43 openai-responses
```

hiyouga:
## Open decision resolved

The repo owner has accepted the protocol change as it stands: moving every OpenRouter row to the Responses API is the intended behaviour, and no `Breaking` metadata or `## Compatibility` section is being added for the preset re-sync. Finding 1 above needs no further action.

Two things still stand between this and a merge, neither of them about the diff:

- **The branch is 6 commits behind `main`** (#680, #682, #694, #697, #698, #699 landed since). It needs a sync before merging; `models-page.tsx` is touched by both this PR and #699.
- **`test-windows (core)` failed, and not because of this PR.** The failure is `EBUSY: resource busy or locked, rmdir` in `hooks.test.ts`, a Windows-only race between the test's teardown and a hook script that has not finished exiting. This diff is the model catalog and does not reach that file. It is also not a rerun flake — CI's built-in retry hit the same race on a fresh temp directory. Fixed separately in [#711](https://github.com/Prism-Shadow/penguin-harness/pull/711); once that lands, syncing this branch picks it up and the check should go green.

PR #706, opened 2026-09-12
fix(web): a scrolling session title no longer also shows a tooltip

### LINKED ISSUE #570: Web/Desktop: Session list overflow — drop the hover tooltip, keep the title scroll
### The problem

In the Session list, an item whose title overflows the row width is handled twice at once: hovering the row makes the title text scroll (marquee), and at the same time a floating tooltip with the full title appears. The two displays carry the same content, so the tooltip is redundant — showing scrolling text and a complete overlay simultaneously is visually noisy.

Drop the floating tooltip and rely on the hover-scroll alone, which already reveals the full title. The scrolling text is sufficient; the extra floating overlay adds nothing.

Where: the Web App / desktop Session list UI.

### What you do instead today

Move the mouse away and back to re-read the scrolling title, or just accept both overlays.

hiyouga:
Closes #570. A sidebar session whose title overflowed its row disclosed the title twice on hover: the scroll reveal slid the clipped tail into view while a native `title` tooltip opened over it with the same text. `Truncated` now attaches `title` only where the scroll cannot reveal the tail — callers without `scrollReveal`, and scroll-reveal rows under the system reduce-motion preference, where `styles.css` disables the keyframes and nothing scrolls. It reads the preference through the existing `usePrefersReducedMotion` hook.

Design half: [penguin-harness-design #152](https://github.com/Prism-Shadow/penguin-harness-design/pull/152).

## Test

The web package runs vitest in a node environment with no DOM harness, so `test/title-reveal.test.ts` gains one source-contract assertion in that file's own style, pinning the new condition (`overflowing && (!scrollReveal || reducedMotion)`) and the hook import. It fails at the base commit.

## Verification

Remote test host (Node 24.20): `pnpm --filter @prismshadow/penguin-web typecheck` clean after building server and core; `pnpm format:check` clean; `web/test/title-reveal.test.ts` 15 passed.


hiyouga:
## Review

The PR drops the native `title` from a sidebar title that scrolls its tail into view and keeps it
exactly where the scroll cannot run — callers without `scrollReveal`, and reveal rows under
`prefers-reduced-motion`. That is what #570 asks for, and it matches the rule this repo already
wrote down for itself in `packages/web/src/components/ui/tooltip.tsx:5-7` ("a trigger must carry one
or the other, never both"). Ship after the fixes below, which I pushed as `7044bf7a`; none of them
changes the behaviour the PR ships.

Blast radius, since `Truncated` is shared — four call sites, two affected:

| Call site | `scrollReveal` | Effect of this PR |
| --- | --- | --- |
| `packages/web/src/components/layout/sidebar.tsx:2405` (draft row) | yes | tooltip dropped |
| `packages/web/src/components/layout/sidebar.tsx:2663` (session row) | yes | tooltip dropped |
| `packages/web/src/features/chat/chat-page.tsx:1833` (conversation header) | no | unchanged |
| `packages/web/src/features/benchmark/benchmark-page.tsx:116` (benchmark row) | no | unchanged |

`.title-scroll` / `.title-scroll-text` / `--title-scroll-*` have no consumer outside
`styles.css:394-405`, and the sidebar's other truncating rows (`sidebar.tsx:1542`, `:1563`, `:2212`,
`:2859`) carry no `title` at all, so the session row loses nothing its neighbours still have.

Two things I checked because removing a reveal affordance usually costs something:

- **The accessible name still carries the full title.** `truncate` clips visually only
  (`overflow: hidden`); the complete string stays a text node inside the row `<button>`
  (`truncated.tsx:93`), and nothing on that path uses `aria-label`, so the button's name is computed
  from that text and is byte-for-byte what it was. `title` was never the name — it only ever added a
  pointer tooltip, and it never surfaced on touch, so the touch path is unchanged too.
- **The overflow measurement is not the usual bug.** It runs in `useLayoutEffect`, not during render;
  the `ResizeObserver` is disconnected on unmount; and the `[text, className]` deps cover both a
  rename by the title generator and the selected-state font-weight switch that `ResizeObserver`
  cannot see. A web-font swap after mount is still never re-measured (the span is `flex-1`, so its
  border box does not move and the observer never fires) — pre-existing, untouched by this PR, and
  not worth widening the diff for.

### Findings

**1. The reduced-motion arm opens a media-query subscription on every truncating row, reversing a property the reveal shipped with** — `packages/web/src/components/ui/truncated.tsx:48` · *should fix*

`usePrefersReducedMotion` is a `useState` + `useEffect` pair that calls `window.matchMedia` and
registers a `change` listener *per component*. It had two callers, both singletons (the sheet, the
dock launcher). `Truncated` makes it one per sidebar session row, per draft row, per benchmark row
and one for the conversation header — and for every caller without `scrollReveal` the value is
provably never read, because `!scrollReveal` short-circuits the condition, yet those components
still hold a listener and still re-render when the OS preference flips. Both `styles.css:388-390`
and the shipped entry `changelog/0.2.3/2026-08-18-sidebar-title-scroll-reveal.md:18` ("no timers, no
per-row media-query subscriptions") record the absence of exactly this as a deliberate property of
#309.

Fixed by giving the hook one module-level `MediaQueryList` served through `useSyncExternalStore`,
following the module-store pattern already in `features/dock/dock-state.ts:333-341`. A subscriber
now costs a `Set` entry, all three callers are behaviourally identical, and the window in which the
old `useState`/`useEffect` pair missed a change landing between first render and effect is gone.
✅ Fixed in `7044bf7a`

**2. Two expressions for one fact, and the older one ignores reduced motion** — `packages/web/src/components/ui/truncated.tsx:75` and `:88` · *should fix*

`scrolling = scrollReveal && overflowing` gates the `title-scroll` class and the two custom
properties; the new `overflowing && (!scrollReveal || reducedMotion)` gates the tooltip. With
`prefers-reduced-motion: reduce` and an overflowing reveal row, both are true: the span gets
`title-scroll`, `--title-scroll-shift`, `--title-scroll-ms` **and** the tooltip, while nothing can
scroll. It renders correctly today only because the one rule keying on `.title-scroll`
(`styles.css:404`) declares nothing but `animation`, which that media query already suppresses — the
second representation is inert by luck, and the comment right above it says the class is attached
"only while there is something to scroll".

Fixed by collapsing both into one `titleDisclosure({ overflowing, scrollReveal, reducedMotion })`
returning `"none" | "scroll" | "tooltip"`, in `lib/title-reveal.ts` beside the other pure rules; the
class, the custom properties and the `title` now all read that single value. ✅ Fixed in `7044bf7a`

**3. The new assertion pins the source text of the condition, not the behaviour** — `packages/web/test/title-reveal.test.ts:143` · *should fix*

`expect(truncatedFlat).toContain("overflowing && (!scrollReveal || reducedMotion)")` fails on any
rewrite that preserves the behaviour — reordering the operands, hoisting the predicate to a named
const, an early return — and passes on a broken one for as long as that substring survives anywhere
in the file, comments included. It does fail at the merge base, so it is not vacuous; what it locks
is the spelling.

Replaced by unit tests of `titleDisclosure` over the four states that matter (fits / scrolling /
reduced motion / reveal not requested), which is precisely the seam `lib/title-reveal.ts` exists for
— its header says the logic is kept out of the component so it is testable under vitest's node
environment — plus one remaining source assertion that the component still routes through the rule
rather than re-deriving it inline. These also fail at the merge base: `titleDisclosure` does not
exist there, and the scrolling case asserts the absence of the tooltip that base attaches.
✅ Fixed in `7044bf7a`

**4. Comments left behind asserting that no JS gates the reduced-motion path** — `packages/web/src/styles.css:388-390`, `packages/web/test/title-reveal.test.ts:110-112` · *nit*

"no JS media query to subscribe to per row, nothing to go stale between render and effect" and "it
does not gate on any JS state" were both true before this PR and false after it. Rewritten to say
what JS now reads the preference for and what it still does not do (start or stop the motion). The
reduced-motion block's own comment at `styles.css:518-521` is still accurate and was left alone.
✅ Fixed in `7044bf7a`

### Left for you

`benchmark-page.tsx:116` renders the same shape of row as the sidebar — same `flex min-w-0 flex-1
text-sm`, same active/inactive ink, same `rounded-md px-2.5 py-1.5` button — and still discloses its
tail with a tooltip. After this PR the app has two different answers for one kind of row. Either is
defensible (the benchmark list has no `data-title-reveal` host, so adopting the scroll is not free),
but the choice is yours: leave it as the deliberate exception, or give it `scrollReveal` too in a
follow-up. I did not change it — it is a product decision, not a defect.

### Verification

```
# vitest ran on the vps3 build box (never locally), whole web package since Truncated is shared:
/home/cc2/.local/bin/vps3-vitest.sh <worktree> rv-706 -- pnpm --filter @prismshadow/penguin-web test
  -> 134 test files passed, 1855 tests passed, exit 0

# local (worktree at 7044bf7a):
pnpm --filter @prismshadow/penguin-web typecheck   -> clean
pnpm lint                                          -> oxlint, exit 0
pnpm format && pnpm format:check                   -> "All matched files use Prettier code style!";
                                                      format touched no file outside the five in this PR
git diff --check                                   -> clean
pnpm --filter @prismshadow/penguin-web build       -> built in 5.16s
packages/web/e2e not run (Playwright is not installed on the build box)

# main sync: not needed. main is 6 commits ahead of the merge base; `git merge-tree --write-tree
# origin/main HEAD` returns a tree with no conflict. The only file both sides touch is
# packages/web/src/styles.css, where main's hunks start at lines 543 and 590 and the reveal block
# edited here is at 366-405. sidebar.tsx, truncated.tsx and title-reveal.test.ts are untouched on
# main, and #680's icons.tsx rewrite / #699's control-size guard reach nothing in this diff.
```

PR #153, opened 2026-08-03
fix(core,cli): strip ANSI from command output

March-77:
## Summary

- strip terminal control sequences at the command-session boundary with a stateful parser, including ANSI sequences split across output chunks
- stop inherited or Vault-provided `FORCE_COLOR` values from reaching child commands while preserving `NO_COLOR=1` and `TERM=dumb`
- make CLI history and streaming colors depend on the actual output TTY, with deterministic coverage for redirected output

## Root cause

Command sessions could inherit forced-color environment variables, and raw PTY/process output was buffered without sanitization. That let ANSI CSI/OSC bytes leak into Web output, CLI output, traces, and model-visible tool results. A chunk-local replacement would still miss escape sequences split by streaming boundaries, so the sanitizer keeps parser state per command session.

## Validation

- `pnpm -r build`
- `pnpm typecheck`
- `pnpm test` (core 603 passed, CLI 207 passed, all workspace unit suites green)
- `pnpm format:check`

Closes #102

PR #707, opened 2026-09-12
feat(desktop): a tray icon keeps the app running when the window closes

### LINKED ISSUE #569: Desktop: add a system tray / menu bar icon
### The problem

The desktop app has no tray presence. Closing the main window quits the app (or it must stay open in the Dock/taskbar), so there is no lightweight way to keep PenguinHarness running in the background while working in other apps. Checking on an agent, opening the models page, or seeing whether a run finished means finding and refocusing the main window.

A system tray (Windows) / menu bar (macOS) icon for the desktop app:

- Closing the window minimizes to the tray instead of quitting (ideally configurable).
- Clicking the tray icon restores / opens the main window.
- The tray menu offers quick actions: open the main window, jump to recent Sessions, open the models page or Settings, and a Quit entry that actually exits the app.
- Optional: a subtle indicator when a run is active or a scheduled task fires.

Where: the desktop app (`packages/desktop`).

### What you do instead today

Keep the window open in the Dock/taskbar, or quit and relaunch whenever I need to check something.

hiyouga:
Closes #569. The desktop shell takes a place in the system tray (Windows notification area, macOS menu bar, Linux tray). By default, closing the main window now hides it there and the embedded server and its background tasks keep running; the icon is the way back.

Design half: [penguin-harness-design #153](https://github.com/Prism-Shadow/penguin-harness-design/pull/153).

## What ships

- **Icon**: rendered from the brand artwork by `scripts/render-icon.mjs` (extended) and committed under `packages/desktop/build/tray/` — a 32px colour image with a 64px `@2x` for Windows and Linux, and a 16px monochrome Template image with a 32px `@2x` that the macOS menu bar inverts itself. `scripts/build-assets.mjs` stages them into `dist/tray/`; the shell resolves them app-path-relative like the window icon. Tooltip is `app.name` (dev suffix included).
- **Clicks**: macOS and Windows — left click shows / restores / recreates the main window, right click (Ctrl+click on macOS) pops the menu. Linux — the menu is attached to the icon (`setContextMenu`), because AppIndicator / StatusNotifierItem trays emit neither `click` nor `right-click`; the desktop decides which click opens it, and "Open PenguinHarness" is the first entry.
- **Menu**: Open PenguinHarness · New Session (`/chat`) · Models (`/models`) · the checkbox "Keep running in the tray when the window closes" · Quit (the existing graceful quit path). Navigation is an in-window URL load; the window stays a plain browser with no IPC channel.
- **Close-to-tray**: on by default; `win.on("close")` hides instead of closing unless the app is quitting. Off restores the previous per-platform semantics. The preference lives in `userData/tray.json` (`{ "closeToTray": boolean }`); missing or malformed reads as on, a failed write is logged. A second launch and a macOS Dock click now also surface a hidden window.
- A platform that cannot host a tray logs and continues without one; the close handler then falls back to the old behaviour.
- Not included, as the spec says: a recent-sessions submenu and a running-activity indicator — both need a shell-to-server query the shell does not have (it holds only the shutdown token and the update relay port).

## Code shape

`tray-prefs.ts` (file read/write, pure) and `tray-menu.ts` (menu as data, pure) are unit-tested; `tray.ts` is the thin Electron layer; `app-icon.ts` gained the tray icon lookup; `main.ts` wires the close handler, `showMainWindow()` / `navigateMainWindow()` and the install at `whenReady`.

## Not verified by running Electron

The tray was not exercised in a running shell in this PR: the Linux `setContextMenu` path, whether a given desktop also emits `right-click`, the `@2x` sibling pickup by `nativeImage.createFromPath`, and that Electron flips a checkbox item's `checked` before its click handler runs (if inverted, `setCloseToTray(!menuItem.checked)` is the one-token fix). A dry-run `pnpm --filter @prismshadow/penguin-desktop start` on a desktop machine is the way to confirm.

## Verification

Remote test host (Node 24.20): `pnpm --filter @prismshadow/penguin-desktop typecheck` clean after building cli, server and core; `pnpm --filter @prismshadow/penguin-desktop build` (tsup + build-assets) clean, `dist/tray/` holds the four PNGs; `pnpm format:check` clean; desktop tests 17 files, 201 passed (new `tray-prefs.test.ts`, `tray-menu.test.ts`, and `app-icon.test.ts` now checks the four tray PNGs' IHDR dimensions).


hiyouga:
## Review

The tray feature is sound: the icon is installed once at `whenReady()` on all three platforms, held in a module-scope reference, destroyed on `before-quit`, and the close-to-tray interception is correctly bypassed while quitting — I could not find a way to strand a running app with no window. One real defect is fixed below. Separately, the repo owner sent a product requirement during this review (an Appearance switch for the icon, on by default, applied without a restart); I implemented it in the same commit and it is described under **Owner requirement**.

I cannot launch Electron in this environment, so everything about window/quit behaviour below is read from the code, not observed in a running app. The asset and build-script claims are executed and measured.

### Findings

**1. A tray navigation entry that has to create the window loads the app root, then aborts it** — `packages/desktop/src/main.ts:190` (pre-fix) · *should fix*

`navigateMainWindow` called `showMainWindow()` first; with `win === null` that ran `createWindow(`${appOrigin}/`)`, which issues `void win.loadURL(origin + "/")`. Control then fell through to `void win.loadURL(origin + target)` on the same WebContents a tick later. Reachable state: macOS, **Keep running in the tray when the window closes** turned off, window closed (app stays in the Dock), then tray → **New Session**. Result: the whole Web App boots on `/` and is cancelled mid-load, and the first `loadURL`'s `ERR_ABORTED` rejection is `void`-ed with no handler. Fixed by creating the window directly on the destination. ✅ Fixed in e0ecec2d

**2. The close/quit rule was inline in an untested module** — `packages/desktop/src/main.ts:111` · *should fix*

`if (!quitting && tray?.closeToTray() === true)` is the single most consequential line in the feature and it sat in `main.ts`, which the PR correctly declares untested. Its three inputs are now `hidesOnClose()` in `util.ts`, pinned in `test/util.test.ts` — including the case the owner requirement makes reachable on purpose, where there is no tray at all and the close must proceed rather than hide. ✅ Fixed in e0ecec2d

**3. Test nit: `TRAY_NAV_PATHS` is asserted by restating its literal** — `packages/desktop/test/tray-menu.test.ts:43` · *nit*

`expect(TRAY_NAV_PATHS).toEqual({ "new-session": "/chat", models: "/models" })` re-types the constant; the `startsWith("/")` line beside it is the part that pins a contract. The claim worth pinning is that both destinations are routes the Web App actually serves — they are (`packages/web/src/router.tsx:69,73`), but nothing here would notice if one were renamed. ⬜ Left for the author

### Checked and correct

- **Quit vs. close.** `window-all-closed` still calls `app.quit()` off macOS, and that is right: it can only fire once every window is *destroyed*, and close-to-tray prevents the destroy, so the handler is unreachable while the feature is on. `quitting` is one-way, but there is no cancelled-quit path to leave it stuck — the only `preventDefault()` is the graceful server stop, which always re-enters `app.quit()` from a `.finally`, and the second `before-quit` pass takes the `stopPromise !== null` branch. `activate` and `second-instance` both route through `showMainWindow`, which shows-then-restores before falling back to recreating.
- **Lifecycle.** The `Tray` is retained by `main.ts`'s module-scope handle and by `installTray`'s own closures, so it cannot be collected; `openTray()` is idempotent and is the only creator, so re-activation cannot produce a second icon.
- **Assets.** Decoded all four PNGs: `trayTemplate.png` is 16×16 and `trayTemplate@2x.png` 32×32, both **RGBA with 0 non-black opaque pixels** (max channel spread 0) — genuine black-and-alpha template images, so the macOS menu bar will invert rather than draw a blob. `tray.png`/`tray@2x.png` are 32×32 and 64×64, exactly 2×. `node packages/desktop/scripts/render-icon.mjs` reproduced all nine committed icons byte-for-byte here, and `build-assets.mjs` fails loudly (exit 1) if `build/tray/` is empty. Caveat, pre-existing rather than new: the script rasterizes through Playwright Chromium, so "byte-identical" holds for a given Chromium build and could drift on a different one.
- **Preferences.** `userData/tray.json` is an entirely new file next to the existing `server-port`, `preferred-port` and `cli-command.json`. It reads and rewrites no existing shape, so there is no on-disk compatibility question. A dev run keeps its own `userData`, so a dev instance cannot overwrite an installed build's preference.

### Owner requirement (not a review finding)

Requested by the repo owner mid-review and implemented in e0ecec2d: the tray icon is a user preference, on by default, controlled from **Settings › Appearance**, applied immediately in both directions.

- **Nothing had to change about when the tray is created** — `installTray` already ran at `whenReady()`, before `boot()`, on all three platforms; it was never lazy on window close. What is new is that it is now gated on the preference and can be torn down and rebuilt at runtime.
- **Channel used: the existing one.** No new IPC surface. The page calls `GET`/`PUT /api/desktop/tray`, gated by the same `desktopMode && sessionVia === "desktop"` rule as the client-update routes; the server forwards over the `process.parentPort` utilityProcess relay that `services/desktop-update-port.ts` already owns; the shell applies, persists and pushes the new state straight back. The window stays a plain browser.
- **Icon off + window closed** is deliberate and tested: with no icon there is nothing to bring a hidden window back, so the close is not intercepted — the app stays in the Dock on macOS and quits on Windows and Linux, exactly the pre-tray behaviour. `hidesOnClose` pins it.
- **Fallbacks.** A missing, unreadable, malformed or ill-typed `tray.json` reads as ON, per field independently. Both writers now read-modify-write the file, so the Appearance switch and the tray menu's own checkbox cannot clobber each other. A failed write is logged and nothing more.
- **i18n.** `settings.trayIcon` / `settings.trayIconInfo` added to both `strings.ts` and `strings-en.ts`; parity is enforced by the type system (`en: Strings`) and typecheck passes. The row uses the same `PrefRow` + `Switch` pair as the two booleans above it, so the dialog-control size rule is not engaged (no new `Button`).
- The bilingual changelog pair is updated, and `Scope` widened to `desktop`, `server`, `web`.

### Left for you

- **The changed default on Windows and Linux.** Closing the window used to quit the app; after this PR it keeps running in the tray. Nothing on disk migrates, but it is a behaviour existing users will meet on first launch with no prompt. If that is meant to be opt-in instead, the flip is one default in `DEFAULT_TRAY_PREFS`.
- **The tray menu is English-only.** So is the app menu (`src/menu.ts`), so this PR is consistent with what is there — but the Web App beside it is bilingual, and the shell has no string catalog at all. Whether the desktop chrome should follow the app's language is a product call, not something to settle inside this PR.

### Verification

```
# in the PR worktree, branch feat/issue-569-desktop-tray
pnpm install --frozen-lockfile                                     ok
pnpm -r build                                                      ok
pnpm typecheck                                                     ok (all packages)
pnpm lint                                                          ok (oxlint, 0)
pnpm format && pnpm format:check                                   ok
git diff --check                                                   clean
pnpm --filter @prismshadow/penguin-desktop build                   ok — "dist/tray/ (4 icons)"
node packages/desktop/scripts/render-icon.mjs                      byte-identical to all 9 committed icons

# vitest, on the build box (never locally)
pnpm --filter @prismshadow/penguin-desktop test                    17 files, 207 tests passed
pnpm --filter @prismshadow/penguin-server exec vitest run \
    test/desktop-tray.test.ts test/desktop-update.test.ts \
    test/desktop.test.ts                                            3 files, 29 tests passed
pnpm --filter @prismshadow/penguin-web test                       134 files, 1851 tests passed

git merge-tree --write-tree origin/main HEAD                       no conflicts — no sync needed
```

Not executed: anything requiring a running Electron process — the click paths, the menu popup, `Tray` creation on a real desktop, and the window hide/show/quit sequence. Those are reviewed by reading.

hiyouga:
## Menu bar icon redrawn, and New Session fixed

Both from the repo owner, not from the review. Pushed in dc2159c8.

### The menu bar image was the brand illustration, shrunk

`trayTemplate.png` was `penguin-logo.svg` with its backdrop stripped, flattened to black and rasterized at 16px. The illustration has gradients, separate flippers, a beak and feet, and none of that survives at that size. Measured on the committed file:

| | old | new |
| --- | --- | --- |
| alpha-weighted coverage | 19.1% | **44.9%** |
| fully opaque | 5.9% | **37.5%** |
| antialiased fringe | 31.6% | 17.2% |

Five of every six painted pixels were fringe rather than mark. On screen that is a grey smear beside the solid marks around it, which is what was reported.

`scripts/menu-bar-glyph.svg` is drawn for the size instead: a head, sloping shoulders, and a notch for the feet. Nothing else. The viewBox is 16 units so one unit is one pixel of the 1x image and every edge lands on the grid deliberately.

Two shapes were tried first and are recorded in the file so they are not tried again. Flipper slots thinner than a pixel never become transparent at 16px — they render as grey lobes that read as headphones. A knocked-out belly turns the figure into a ring.

Only the two template files changed. The other seven icons re-rendered byte-for-byte identical, which also re-confirms the renderer is deterministic. `BACKDROP_PATH` and its guard are gone with the derivation they protected.

### New Session opened the last conversation

`TRAY_NAV_PATHS["new-session"]` was `/chat`. The chat page with no session in the path redirects to the most recent conversation (`chat-page.tsx:735`), so the entry reopened the last session rather than starting one. It is now `/chat/new`, the Web App's draft sentinel.

### Three assertions, each failing against the previous commit

`packages/desktop/test/png-pixels.ts` decodes the committed PNGs so the tests can hold what the files actually look like, not just their dimensions.

```
× opens a draft for New Session, not the last conversation
  → expected '/chat' to be '/chat/new'
× gives trayTemplate.png enough weight to read
  → expected 0.1908394607843138 to be greater than 0.4
× gives trayTemplate@2x.png enough weight to read
  → expected 0.18597579656862726 to be greater than 0.4
```

The route pin reads `DRAFT_SESSION_ID` out of the Web App source rather than restating the literal, because this package cannot import it and a hand-copied constant is what drifted in the first place. The template images are also asserted to carry no colour: a coloured template is not ignored by the menu bar, it renders as a black blob.

### Verification

```
desktop suite on vps3     17 files / 212 tests passed (was 207)
mutation run              3 failed / 18 passed with the old icon and route restored
pnpm --filter @prismshadow/penguin-desktop typecheck   pass
pnpm lint / pnpm format:check / git diff --check       pass, clean
template purity           trayTemplate.png 140 non-transparent px, 0 non-black
                          trayTemplate@2x.png 508 non-transparent px, 0 non-black
```

vitest ran on the remote build host. Electron itself is still not launched here, so the icon is measured rather than seen in a real menu bar.

hiyouga:
## Menu bar icon redrawn from the logo's own outline

Rework requested by the repo owner: keep the logo's outer contour, fill the belly black, drop the two arcs, and carry more weight. Pushed in a94439275, which also merges current `main`.

### What it is now

The penguin out of `penguin-logo.svg`, not a second drawing of one. The body path holds the penguin as two subpaths — the outline, and the belly cut out of it — so taking the first alone gives exactly the outline with the belly solid. The two arcs sweeping around the penguin, the highlights and the eye are separate paths and are simply not drawn.

`scripts/render-icon.mjs` takes that out of the artwork at render time and measures its bounding box with the browser it already launches, so there is no copied path and no hand-copied number to go stale. It fails loudly if the body stops being one gradient-filled path of exactly two subpaths, the way the old `BACKDROP_PATH` guard did for the derivation it replaced. `scripts/menu-bar-glyph.svg` is deleted with the hand-drawn figure it held.

### Framing, and why 35% rather than 40%

Framed to the square its bounding box inscribes, so the mark fills the bar's width instead of sitting inside the artwork's rounded-square padding.

| framing | coverage | mark |
| --- | --- | --- |
| flattened whole illustration (before) | 19.1% | a wisp |
| bounding box, exact | **34.7%** | whole |
| 8 units tighter | 36.2% | beak grazing the edge |
| 30 units tighter | 40.4% | beak and tail cut |
| 60 units tighter | 46.6% | visibly chopped |

Roughly double the weight it had, but it stops short of 40%, and that is the shape rather than the rendering: a 4:3 silhouette in a square canvas leaves a band above and below, and buying the extra coverage means cropping the beak off one side and the tail off the other. I rendered the tighter framings at size to check rather than assuming, and the last two are plainly truncated. The coverage assertion moves from 40% to 30% to match, still far above the 19% it exists to catch.

### Merge

`main` moved 16 commits under this branch. Both string catalogs conflicted, purely additively — the tray icon switch here against the task-completion notification copy from [#702](https://github.com/Prism-Shadow/penguin-harness/pull/702). Both sides kept, and key parity across the two catalogs re-checked afterwards: no key in one and not the other.

### Verification

```
desktop suite on vps3     17 files / 212 tests passed
web suite on vps3         139 files / 1899 tests passed
pnpm -r build / pnpm typecheck / pnpm lint / pnpm format:check   pass
git diff --check                                                 clean
template purity           trayTemplate.png 115 non-transparent px, 0 non-black
                          trayTemplate@2x.png 412 non-transparent px, 0 non-black
coverage                  34.7% at 1x, 35.0% at 2x
```

Only the two template files changed under `build/`; the other seven icons re-rendered byte-identical. vitest ran on the remote build host.

hiyouga:
## The tray menu now reads in the window's language

Requested by the repo owner. Pushed in 07e103af4. The macOS icon and the Windows colour icon are unchanged — Windows was already on `tray.png`, the colour one, and only macOS takes the template.

### Why the page has to report it

The language is a browser preference in `localStorage` (`penguin.lang`, resolved to `zh` or `en` by `state/locale.tsx`). The shell is a separate process and the desktop design keeps that boundary one-way: the window is a plain browser with no IPC bridge into it. So the page reports its language over the same `PUT /api/desktop/tray` the Appearance switch already uses, on load and on every change, and the shell persists what it hears.

That route's body became a **patch** — `showTrayIcon`, `locale`, or both. The switch and the language are written by different parts of the page at different moments, and neither should have to restate the other's value to change its own. A request that asks for nothing is refused rather than relayed.

### What the shell does before anyone reports

`tray.json` gains `locale`, so a fresh launch draws the right menu before any window has finished loading. Until something has been reported it follows the device language, by the same rule the Web App applies to its own "follow the system" default — an untouched install then agrees with itself across the window and the tray. A `tray.json` written before this field existed reads as "nothing reported", which is the same thing.

### The labels are written twice, deliberately

Six strings, in `tray-menu.ts`. The shell starts before any page loads and must draw a menu whether or not one ever does, and nothing in `packages/desktop` depends on `packages/web`, so it cannot read those catalogs. A `Record<TrayLocale, TrayLabels>` makes a missing translation a compile error rather than a blank menu entry, and a test pins that both languages carry the same keys and that none is empty. The app name stays untranslated in both, as a proper noun.

The menu is built per popup, so a language change costs nothing until the next one opens — except on Linux, where the menu is attached to the icon, and the tray handle rebuilds it in place.

### Pinned by tests, not just by types

- every entry drawn in the given language, same entries and same order in both
- both languages carry the same keys, none empty
- the shell's device-language fallback matches the Web App's `resolveSystemLocale`, tag for tag
- the language survives the read-modify-write the other two writers share, so a report cannot revert the switch and a switch cannot revert the report
- a patch frame carrying one field, the other, both, or nothing this build understands
- the route's `invalid_locale` and `empty_tray_patch` refusals
- a status push from a shell too old to send a language reads as English rather than dropping the whole push, so an old shell against a new server keeps its icon switch

A compile-time assertion in `use-tray-locale.ts` also holds the Web App's `Locale` and the route's `DesktopTrayLocale` to the same set, so a divergence is a build error rather than a 400 nobody reads.

### Verification

```
desktop suite on vps3   17 files / 216 tests passed (was 212)
server suite on vps3    126 files / 1829 tests passed
web suite on vps3       139 files / 1899 tests passed
pnpm -r build / pnpm typecheck / pnpm lint / pnpm format:check   pass
git diff --check                                                 clean
```

vitest ran on the remote build host. Electron is still not launched here, so the menu is tested as data and reasoned about as a native menu.

Issue #581, opened 2026-09-02
Subagent: a background-completion notice for one subagent interrupts other still-running subagents (conditional)


hiyouga:
### What happened

When a "后台任务完成" (background task completed) notice appears, it reflects that **one** particular subagent finished — but at the same time, other subagents that are **still running** get interrupted automatically. Not always reproducible; it seems conditional on some state (likely the notice-delivery / run / steering path), but the exact trigger is not yet isolated.

Expected: a background-completion notice for one subagent should not abort other in-flight subagents.

### Steps to reproduce

1. Launch several subagents in parallel (e.g. multiple `run_subagent` calls running in the background).
2. Wait for one of them to finish.
3. The "后台任务完成" notice appears for that finished subagent.
4. Observe that the other subagents that were still running are interrupted/aborted.

Note: this does not reproduce every time — a specific ordering or state seems required, so far undetermined.

### Version

unknown (will update once confirmed)

### How it is installed

Desktop app

### Operating system

macOS

### Does it reproduce on a fresh data root?

Not tested

### Relevant output

No log captured yet — the symptom is visible in the UI: the completion notice for one subagent coincides with the other in-flight subagents stopping.

hiyouga:
Investigated from the report without a reproduction; here is what was ruled out and what would pin the trigger.

**Ruled out (by reading every path that can end a background child's run, and by a test):**

- The completion notice itself, on both delivery paths. Idle: the host takes the notice and runs it as a new Task; mid-run: the engine drains it into the running Task at the next boundary. Neither path touches the other children. PR #708 adds a core test that launches three `run_in_background` subagents, settles one on each path, and checks the other two keep running and end `completed`.
- Registry eviction (`makeRoom` only ever evicts a child that has already finished), the foreground collect window's abort (`finally` only kills a child that was never promoted), and `Environment.dispose` (session/agent/project deletion, shutdown, and idle eviction, which running children and undelivered notices pin).
- A compaction triggered by the notice: context rotation re-equips the same Environment, so children survive it.
- The parent's Task ending: it denies only the main session's pending approvals; a child's escalated approval carries its origin and stays pending for you.
- The web topology: an idle-launched notice is treated as part of the Task that launched the background work, so the panel does not reset.

**What would settle it.** If it happens again, please attach:

[...]

PR #708, opened 2026-09-12
test(core): sibling background subagents survive a completion notice

hiyouga:
For #581 (a background subagent's completion notice reportedly interrupts the other subagents still running). No runtime change: this adds the test that pins the contract, after an audit found no path by which a sibling's notice can end a child's run.

## What the test does

Three `run_in_background` children on one real `Environment` + `Session` (scripted runner, per-child gates). One child settles while the Session is idle — the host takes the notice and runs it as a Task, exactly as the server's idle-arrival signal does; another settles while a Task runs — the engine drains the notice into that Task at its next boundary and the host is not signaled. After each delivery the remaining children are still running (`listBackgroundSubagents`, `hasRunningBackgroundSubagents`), and the last one ends `completed` on its own.

## What the audit covered

Every call site that ends a background child's run: the foreground collect window's abort and its `finally` (foreground only), `input_subagent` `abort: true`, the panel's stop, registry eviction (`makeRoom` never evicts a running child), and `Environment.dispose` (session/agent/project deletion, shutdown, and idle eviction — which running children and pending notices pin). Context rotation after a compaction re-equips the same Environment. A parent Task's end denies only the main session's pending approvals (`denyMain`); a child's escalated approval carries its origin and stays pending. The web topology already treats an idle-launched notice as part of the launching Task.

## Verification

Remote test host (Node 24.20): `packages/core/test/background-execution.test.ts` — 34 passed (the new case included).

### DIFF
--- a/changelog/unreleased/2026-09-12-sibling-subagents-survive-notice.md
+++ b/changelog/unreleased/2026-09-12-sibling-subagents-survive-notice.md
@@ -1,0 +1,15 @@
+# A test pins that a subagent's completion notice ends no sibling
+
+- **Date:** 2026-09-12
+- **Type:** process
+- **Scope:** `core`
+- **Issue:** [#581](https://github.com/Prism-Shadow/penguin-harness/issues/581)
+
+[中文版](2026-09-12-sibling-subagents-survive-notice.zh.md)
+
+The core suite gained one test for the report that a background subagent's completion notice interrupted the other subagents still running: three `run_in_background` children on one Session, one of them settling while the Session is idle and another while a Task runs. On both delivery paths the notice reaches the model as input only, the host is signaled on the idle path alone, and the remaining children keep running until they finish on their own with `completed`.
+
+## Details
+
+- The test drives the real `Environment`, `Session` and subagent tools with a scripted runner whose children answer when released, so the timing of each completion is exact.
+- No runtime code changed: every path that can end a background child's run was read against the report and none is reached by a sibling's notice.
--- a/changelog/unreleased/2026-09-12-sibling-subagents-survive-notice.md
+++ b/changelog/unreleased/2026-09-12-sibling-subagents-survive-notice.md
@@ -3,6 +3,7 @@
 - **Date:** 2026-09-12
 - **Type:** process
 - **Scope:** `core`
+- **PR:** [#708](https://github.com/Prism-Shadow/penguin-harness/pull/708)
 - **Issue:** [#581](https://github.com/Prism-Shadow/penguin-harness/issues/581)
 
 [中文版](2026-09-12-sibling-subagents-survive-notice.zh.md)
--- a/changelog/unreleased/2026-09-12-sibling-subagents-survive-notice.zh.md
+++ b/changelog/unreleased/2026-09-12-sibling-subagents-survive-notice.zh.md
@@ -1,0 +1,15 @@
+# 用测试钉住：子智能体的完成回报不会中断其它子智能体
+
+- **Date:** 2026-09-12
+- **Type:** process
+- **Scope:** `core`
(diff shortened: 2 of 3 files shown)


hiyouga:
## Review

Adds one core test pinning that a background subagent's completion notice ends no sibling, closing out the audit of [#581](https://github.com/Prism-Shadow/penguin-harness/issues/581) with no runtime change. The test is real — I verified it has teeth rather than trusting that it passes — and I found no path in core or in the server by which a sibling's notice ends another child's run. Ship as is; the one thing I would not do is treat this as closing the issue, for the reason below.

### Findings

**1. The test is not vacuous — verified by injecting the reported symptom** — `packages/core/test/background-execution.test.ts:1140` · *no action needed*

A test added alongside "we audited and found no bug" is worth nothing if it passes for the wrong reason, so I checked directly. I patched `Session.takeBackgroundNotices` in `packages/core/src/session.ts` to abort every other running background child as it takes a notice — exactly the symptom #581 describes — and reran the file on the test host:

```
FAIL  test/background-execution.test.ts > sibling background subagents survive a completion notice (issue #581)
Test Files  1 failed (1)
      Tests  1 failed | 33 passed (34)
```

The new case fails and the other 33 keep passing, so the pin is narrow and it is load-bearing. The patch was reverted; the tree is clean.

**2. It fails as a 5-second timeout, not at the assertion that states the contract** — `packages/core/test/background-execution.test.ts:1141` · *nit*

Under the injected fault the failure surfaced as `Test timed out in 5000ms` reported against the `it(...)` line, not as `expected length 2, got 0` at `expect(running()).toHaveLength(2)`. Aborting a child stops its run without settling it, so the run stays listed briefly and the test instead hangs later, in a `waitFor` for a signal that never arrives. It catches the regression either way, so this is only about what the next person sees first: a timeout costs a debugging cycle that a count mismatch would not. Left alone deliberately — tightening it means changing the test's timing, which is how a deterministic test becomes a flaky one.

**3. The issue should stay open; this answers it rather than closing it** — *left for you*

The report is explicit that the failure is conditional: "Not always reproducible; it seems conditional on some state... a specific ordering or state seems required, so far undetermined." The test covers the two deterministic orderings — notice at idle, notice mid-Task — and proves neither ends a sibling. It cannot prove the absence of a race that only appears under an ordering nobody has isolated yet.

The PR body is honest about this and carries no `Closes`/`Fixes` keyword, so merging will not auto-close #581. Worth keeping it that way, and worth asking the reporter for a Trace from a run where it happened.

### Left for you

**The reported deployment is Desktop, and this test is core-only.** `abortBackgroundSubagentRun` is the one API that ends a specific child's run, and its only production caller is `packages/server/src/runtime/session-manager.ts:720`, reached from the panel's stop. The notice path in that layer is `startBackgroundNoticeTask` (`session-manager.ts:1727`), which I read: it takes the queue and launches a Task only when `entry.status === "idle"` with no queued follow-ups, and aborts nothing. So the audit holds at the server layer too — but it holds by inspection, not by test. If you want the contract pinned where the bug was actually seen, the missing test is a server-level one asserting that a notice arriving for one child leaves the others running. Your call whether that is worth it now or only if #581 recurs.

### Verification

```
core suite on vps3            54 files passed | 1 skipped (55);  1164 passed | 5 skipped (1169)
fault-injection rerun         1 failed | 33 passed (34)   <- the new test, with the bug simulated
pnpm --filter @prismshadow/penguin-core typecheck   pass
pnpm lint (oxlint)                                  pass
pnpm format:check                                   pass
git diff --check                                    clean
```

vitest ran on the remote build host, not locally. Changelog pair checked against `changelog/README.md`: both files present and mirroring, metadata in order, `Issue:` correctly points at an issue rather than a PR (confirmed #581 is an issue, not a pull request — they share one numbering space here), no `## Why` / `## Verification` sections.

PR #709, opened 2026-09-12
fix(core): edit_file and write_file serialize per file

hiyouga:
Parallel `edit_file` calls on one file overwrote each other: the tool reads, computes and writes back with no mutual exclusion, and the engine runs a turn's approved tool calls concurrently (subagents in the same process too), so both calls read the original content and the last atomic rename won — one edit lost while both reported success. `write_file` raced an edit the same way.

Design half: [penguin-harness-design #154](https://github.com/Prism-Shadow/penguin-harness-design/pull/154).

## What ships

- `packages/core/src/internal/file-lock.ts`: a process-wide mutex keyed by the file's real path (`fileLockKey`: `realpath`, or the real directory plus the name for a file that does not exist yet, so a symlink and its target, and a file before and after `write_file` creates it, share one key). `withFileLock(key, fn, signal)` serves callers in arrival order; a caller whose signal fires while queued rejects with an `AbortError` and never runs; a rejecting `fn` never blocks the next caller; the per-key entry holds a caller count and is deleted exactly when nobody is inside or waiting.
- `edit_file` and `write_file` run their whole stat → read → compute → atomic write under that lock; the locked section became a plain async function returning an outcome, and the generator yields after the lock is released. Every message text, stop reason and diff output is unchanged. An `old_string` that an earlier edit removed now fails with the existing "old_string not found" message instead of overwriting.
- `read_file` stays unlocked: the atomic rename already gives a reader the old bytes or the new ones. Other processes (a CLI-spawned run, an editor) are outside the lock — no OS-level file lock is taken.
- Docs (`tools`, both languages) state the rule; changelog pair.

## Tests

`file-tools.test.ts`: eight concurrent edits of one file each land; two concurrent edits of the same `old_string` — exactly one succeeds and the other reports not found; `write_file` racing `edit_file` ends in one of the two sequential orders; edits through a symlink and its target both land. `file-lock.test.ts`: arrival order, a rejecting `fn` releases the next caller, abort while queued, the entry is gone once drained, key equality for a missing-then-created file and for a symlink, an unresolvable path keys on itself.

## Verification

Remote test host (Node 24.20): `pnpm --filter @prismshadow/penguin-core typecheck` clean; `pnpm format:check` clean; core `file-tools`, `file-lock`, `atomic-write`, `environment` suites — 4 files, 120 passed, 2 skipped (the existing win32 skips); docs 7 files, 53 passed.


hiyouga:
## Review

Adds a process-wide per-file mutex and runs `edit_file`/`write_file`'s whole stat → read → match →
atomic-write under it, keyed by the file's real path. The mutex itself holds up: I probed arrival
order over 200 callers, a rejecting `fn`, a pre-aborted signal, 50 waiters aborting behind a holder
and the map afterwards — order preserved, every path releases, `pendingFileLocks()` back to 0, no
unhandled rejections. Two things needed fixing before that guarantee is real, both done in
c5763b4a; ship after a look at those.

### Findings

**1. One file could take two different keys while its directory was still missing** — `packages/core/src/internal/file-lock.ts:55` · *should fix*

`fileLockKey` fell back to the immediate parent only, and to `path.resolve(dir)` when that parent
did not resolve either. `write_file` creates missing parents, so "the directory is not there yet" is
its ordinary state, and under a linked ancestor the two branches disagree. Concretely, with
`/tmp/X/ws` a symlink to `/tmp/real`:

```
fileLockKey("/tmp/X/ws/newdir/a.txt")  before mkdir → /tmp/X/ws/newdir/a.txt     (unresolved)
                                       after  mkdir → /tmp/real/newdir/a.txt     (realpath)
```

A `write_file` that keys before its own `mkdir` therefore holds a key no later call on that file can
derive: the second call keys canonically, takes a free lock, and the pair the docs say is serialized
is not. macOS reaches every temporary directory through `/var` → `/private/var`, so this is the
common case there, not an exotic one. Fixed by walking up to the deepest ancestor that does resolve
and re-attaching the segments below it; the link-cycle and symlink-folding behaviour is unchanged.
✅ Fixed in c5763b4a, with `file-lock.test.ts` "gives a file under a directory that does not exist
yet that same key" — it fails against the pre-fix body (`linked-root/created/later.txt` vs
`real-root/created/later.txt`), verified by replaying it with the old body restored.

**2. The `write_file` vs `edit_file` race test passed against the pre-fix tools about half the time** — `packages/core/test/file-tools.test.ts:888` · *should fix*

`expect(["fresh\nedited\nnew-tail\n", "fresh\nshared\nnew-tail\n"]).toContain(final)` accepts an
outcome the unserialized code also produces: with no lock, whichever tool renames last decides the
file, and when that is `write_file` the result is exactly `fresh\nshared\nnew-tail\n` — the same
bytes a real (edit, write) order leaves. Replaying the test with both tool files restored to
`$(git merge-base origin/main HEAD)`: run 1 failed (`keep\nedited\nold-tail\n`), run 2 passed. The
file content alone cannot tell the orders from the race. Fixed by asserting what the *second* caller
read — each tool renders its diff against the bytes it found, so `-edited` in `write_file`'s diff
(or ` fresh` in `edit_file`'s) is what only a serialized order can produce. Three of three base
replays now fail on `expected 'Overwrote "both.txt" …' to contain '\n-edited\n'`.
✅ Fixed in c5763b4a.

**3. The changelog entry carried a test inventory** — `changelog/unreleased/2026-09-12-file-tool-lock.md:33` · *nit*

`changelog/README.md` keeps verification and claims about what the tree currently *is* off disk and
in the PR description, which already carries the same paragraph; it was the only one of the 38
entries in `unreleased/` with one. Removed from both languages. The key bullet was also reworded to
match finding 1. ✅ Fixed in c5763b4a.

**4. The mutex is not reentrant, and says so nowhere** — `packages/core/src/internal/file-lock.ts:106` · *nit, left alone*

`withFileLock(k, () => withFileLock(k, …))` never settles — no timeout, no diagnostic, and the key's
entry is leaked for the life of the process (I confirmed this by running it). Nothing nests today:
`applyEdit`/`applyWrite` only reach `node:fs`, `missingPathHint` and `atomicWriteFile`, none of which
take a lock, so the code as shipped is safe. But `apply_patch`, a multi-file edit, or anything else
later built *on top of* these two tools would self-deadlock silently. One line in the module doc
naming the constraint would be worth it — I left it to you rather than editing prose.

**5. A dangling symlink and the file it names still key apart** — `packages/core/src/internal/file-lock.ts:49` · *nit, left alone*

`write_file` follows the leaf chain and writes the target (`file-tools.test.ts:783` pins that), but
`realpath` fails on a dangling link, so the key stays the link's own path: `fileLockKey(dangling.txt)`
→ `…/dangling.txt` while `fileLockKey(ghost.txt)` → `…/ghost.txt`. Two concurrent `write_file`s, one
through the link and one through the target, are not serialized. I left it: no read-modify-write is
lost by it (`edit_file` through a dangling link fails at `stat`; two creating writes have only their
`Created`/`Overwrote` line to get wrong), and closing it means running `resolveWriteTarget` before
canonicalizing, which is a design call rather than a bug fix.

### Verification

```
Worktree ../penguin-harness-wt/rv-709 @ c5763b4a, Node 24.18.0.

vitest on the build box (vps3), never locally:
  pnpm --filter @prismshadow/penguin-core --filter @prismshadow/penguin-docs test
      core  55 files passed | 1 skipped, 1175 passed | 5 skipped   PASS
      docs   7 files passed,                53 passed             PASS
  (baseline before my commit: core 1174 passed — the delta is the new key test)

Replays against $(git merge-base origin/main HEAD), each on vps3:
  pre-fix fileLockKey body  + new key test        → 1 failed | 7 passed   (expected)
  base edit-file/write-file + all 4 race tests    → 4 failed              (expected)
  base edit-file/write-file + old race assertion  → run 1 FAIL, run 2 PASS (finding 2)
  base edit-file/write-file + new race assertion  → 3 of 3 FAIL

Locally:
  pnpm typecheck        PASS (all packages)
  pnpm lint (oxlint)    PASS
  pnpm format:check     PASS
  git diff --check      clean
```

PR #710, opened 2026-09-12
fix(web): the Trace view loads every page of a file, so late rounds keep their messages

hiyouga:
Bug: in the Trace view of a conversation with many turns, later rounds — the compaction round among them — showed a timeline but "消息（0）". The view fetched only the first 1000 events of a Trace file (`getAgentTraceEvents(…, 0, 1000)`) while the analysis, and the timeline drawn from it, covered the whole file; messages are attributed to rounds by the server's index ranges, so every round past event 1000 rendered empty, and a compaction sits at the end of a long file.

Design half: [penguin-harness-design #155](https://github.com/Prism-Shadow/penguin-harness-design/pull/155) ("大文件分页加载" now means: page through the whole file).

## What ships

- `packages/web/src/features/traces/trace-events-loader.ts`: `loadTraceEventPages(fetchPage, { pageSize, onPage, signal })` walks the file from offset 0 in pages of 1000 (the server's `limit` cap), following the latest page's `total` (a file appended during the walk is followed), stopping on an empty page, on cancellation, or at a 500-page cap; it resolves with the final total.
- `trace-file-view.tsx`: the analysis and the page walk run concurrently; each page is **spliced in at its offset**, so a fresh load reads from the file's start while the rest arrives, and a refresh (the panel re-reads on every settled turn) updates rows in place instead of shrinking a long file back to its first page; the walk's end trims to the final total. The error is cleared only once both the analysis and the first page have landed. The footer note now reads "已载入 N / M 条消息…" while pages are still arriving (`truncatedNote` → `loadingNote` in both dictionaries) and disappears when complete.
- Changelog pair. No docs change: no page described the 1000-event cap.

## Tests

`packages/web/test/trace-events-loader.test.ts` (8): three pages in order for 2500 events; advancing by what was delivered, not by the requested size; a file that grows mid-walk is followed to the new total; a zero-event page ends the walk; cancellation stops before the next page; the page cap; the resolved total (plain, grown, and cancelled-before-first); the page size pinned to the server's cap.

## Verification

Remote test host (Node 24.20): `pnpm --filter @prismshadow/penguin-web typecheck` clean after building server and core; `pnpm format:check` clean; web full suite after building core, server, web and cli: 135 files, 1858 passed.


hiyouga:
## Review

The walk itself is correct: I traced it page by page against the real endpoint and it neither
drops nor repeats an event, it follows a file that grows under it, it terminates on four distinct
conditions with a hard request cap, and the trim on completion cleans up a shorter file's
leftovers. `messageFrom`/`messageTo` really do index the same array `readEvents` slices
(`trace-service.ts:847` and `trace-service.ts:1365` both read `readFileByIndex`), so paging to the
end is the right fix for the empty later rounds. Ship after the test I pushed — the one real gap
was in coverage, not in the code.

### Findings

**1. The cancellation check that matters was not covered by any test** — `packages/web/src/features/traces/trace-events-loader.ts:82` · *should fix*

Of the two checks, line 80 (before the request) and line 82 (after the await, before `onPage`),
only line 82 stops a page fetched for file A from being reported into a view that has already
moved to file B. No test reached it: the existing cancellation case flips the flag *inside*
`onPage`, which is synchronous, so the loop always exits at line 80 on the next iteration.

Proved by mutation: with line 82 deleted, all eight cases still pass.

```
× does not report a page that arrived after the walk was cancelled  (new)
✓ pages a 2500-event file in three requests …
✓ stops without reporting or requesting anything more once it is cancelled
✓ … 6 more                                Test Files 1 failed | Tests 1 failed | 8 passed
```

Added a case that flips the flag while a page is in flight (the real shape — the flag is flipped
by the effect cleanup at `trace-file-view.tsx:325`, never from inside `onPage`) and asserts
nothing was reported, nothing further requested, and the walk resolved with 0.
✅ Fixed in 50408416

**2. Two comments describe an endpoint that does not behave that way** — `trace-events-loader.ts:37`, `:49` · *nit*

- "The server caps `limit`, so a larger value comes back clamped" — it does not clamp. `limitQuery`
  (`packages/server/src/http/validate.ts:74-79`) throws `badRequest` for anything outside 1-1000,
  so `limit=2000` is a 400, not a 1000-event page.
- "or shortened by a line the reader skipped" — cannot happen. `readEvents` slices the *parsed*
  array and reports `total: messages.length` of that same array (`trace-service.ts:845-851`), and
  `readTraceTolerant` drops malformed lines before that (`packages/core/src/trace/resume.ts:156-168`),
  so a skipped line shortens the page and `total` together.

`offset += res.events.length` is still the right rule and I left the code alone; only the stated
reasons were wrong, and the fake's `serverLimit` comment repeated one of them.
✅ Fixed in 50408416

**3. A page belonging to the previous file can still be spliced into the new file's list** — `trace-file-view.tsx:252-257` vs `:306-313` · *nit, pre-existing*

The exact interleaving: the file-switch reset runs during render, so `events: []` is committed for
file B; the effect cleanup that sets `signal.cancelled` runs in React's passive-effect flush, which
is scheduled on a Scheduler macrotask *after* that commit. A `getAgentTraceEvents` response for
file A that settles in the gap runs its continuation as a microtask, reads `cancelled === false`,
and `onPage` splices file A's events into file B's list.

The trim at `:320` heals this in every case I could construct except one: file B whose walk ends on
an empty page while still reporting a larger `total` (say 500 delivered, `total` 2000). Then
`prev.length (1000) > loaded (2000)` is false, indices 500-999 keep file A's events, and any round
of B with `messageFrom >= 500` renders file A's messages.

Not introduced here — the same gap existed at the base commit, where the single-shot `Promise.all`
result could land the same way — and there it healed completely, because the next result replaced
the list outright. Left alone: closing it means either resolving the walk with the offset it
actually covered rather than the last page's `total` (which is your documented contract), or keying
the walk to `fileKey`. ⬜ Left for the author

**4. `expect(TRACE_EVENT_PAGE_SIZE).toBe(1000)`** — `packages/web/test/trace-events-loader.test.ts:215` · *nit*

Asserts a constant against its own literal, so it can only ever fail as a change detector. Left as
is — it pins the cross-package contract at `validate.ts:74-79` for a line of code, and the
alternative (importing the server's validator into a web test) is worse. ⬜ Left for the author

### Left for you

**Refresh cost.** The panel re-reads the selected file on every settled turn
(`chat-page.tsx:786` → `trace-panel.tsx:206`), and the walk restarts at offset 0 each time. Server
side that moves the per-turn cost from 2 whole-file reads (`analyze` plus one `readEvents`) to
`1 + ceil(n/1000)`: `readEvents` has no cache, every call goes through `readTraceTolerant`, which
reads and parses the entire file. On the wire the whole file now crosses per turn instead of its
first 1000 events. It is bounded — one Trace file is one model context (`core/src/trace/writer.ts:10-11`),
so 2-4 pages in practice, which is why I did not call it blocking. Trace files are append-only per
index, so a refresh could resume from `events.length` rather than restart; that changes what a
refresh re-validates (a replaced or truncated file would no longer be re-read from the start), so
it is your call, not a review fix.

**On the render ceiling**, since a bigger file now reaches the DOM: rows are collapsed by default
(`trace-event-row.tsx:417`), so each event is one `<li>` with a badge and a truncated summary — no
Markdown and no highlighting until a row is expanded, so this is not the #682 class of problem. The
grouping memo is O(events × rounds) through the linear scan at `trace-file-view.tsx:408-413` and
reruns once per page, which is ~100k iterations for a 2500-event / 40-round file. What is unbounded
is the node count, and there is no virtualization here before or after this PR. I did not measure it.

Also checked and clean: `loadingNote` is present in both catalogs with the same argument shape,
`truncatedNote` and `eventsOffset` have no remaining references, no hard-coded user-facing text was
added to the `.tsx`, and the changelog pair matches `changelog/README.md` field for field.

### Verification

```
# vitest on the build box (vps3), never locally
  → 135 files, 1859 tests passed

vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web exec vitest run test/trace-events-loader.test.ts
  → 9 passed  (at 50408416)
  → with trace-events-loader.ts:82 deleted: 1 failed | 8 passed  (the mutation above)

# locally
pnpm typecheck                                 → all packages Done
pnpm lint                                      → clean (oxlint)
pnpm format && pnpm format:check               → no files changed
git diff --check                               → clean
pnpm --filter @prismshadow/penguin-web build   → built in 10.09s
```

`packages/web/e2e` was not run. The branch is `MERGEABLE` against current `main`, so no sync is
needed.

PR #712, opened 2026-09-12
fix(web): a sidebar folder reveals ten rows at a time and folds back

hiyouga:
Bug: under time grouping, expanding the archived folder did not reveal ten conversations per click — from 28 remaining it showed 13 at once — and once everything was shown nothing folded it back. The active list has a per-group display cap (ten per "more", "show less" at the end), but a folder rendered every loaded row and its "more" only fetched the next server page, which under time grouping fans out to every contributing agent, so one click could load and show far more than a page.

Design half: [penguin-harness-design #156](https://github.com/Prism-Shadow/penguin-harness-design/pull/156).

## What ships

- `folderRevealPlan({ cap, loaded, total, fullyLoaded })` in `packages/web/src/lib/session-grouping.ts` → `{ shown, hidden, canCollapse }`, reusing `hiddenRowCount` so a folder's count and the active list's are one function (the `fullyLoaded` clause that discards a drifted server total included).
- `sidebar.tsx`: a per-folder display cap (`folderCaps`, reset with `groupCaps` on a Project switch and a mode switch); `renderFolder` shows the capped slice; "more" raises the cap by a page and fetches only when memory runs short and some agent still has more (mirroring `showMore`); "show less" drops the cap. Search still forces folders open with neither row; the first-expand fetch, the pending state and the active-session auto-expansion are unchanged. The cap survives closing and reopening a folder, as a group's does.
- `FolderSection` (`components/ui/group-list.tsx`) gained `less` / `onLess`, rendered after its "more" row as the active list renders its pair.
- Docs (`web-app`, both languages) and the changelog pair.

## Tests

`packages/web/test/group-paging.test.ts`: six `folderRevealPlan` cases — 28 total / 10 loaded / cap 10 → 10 shown, 18 hidden, no collapse; cap 20 / 23 loaded → 20, 8, collapse; cap 30 / 28 fully loaded → 28, 0, collapse; a cap above the loaded rows; a folder that fits in one page; a stale total once fully loaded.

## Verification

Remote test host (Node 24.20): `pnpm --filter @prismshadow/penguin-web typecheck` clean after building core, server, web and cli; `pnpm format:check` clean; web full suite 137 files, 1876 passed.


hiyouga:
## Review

Gives each collapsed folder its own display cap so it reveals ten rows per click instead of everything a fetch returned, with a "Show less" that folds it back. The reveal maths is right at every boundary I walked — a folder of exactly 10 offers no "More" (`hidden = loaded - shown = 0`); one of 11 shows the singular "Show 1 more chats", then lands on 11 and stops; "Show less" deletes the map entry so the cap falls back to `SIDEBAR_PAGE_SIZE`, i.e. to exactly the first ten and never to zero; and `folderCaps` is keyed by `folderKey(groupKey, category)`, a group identity, not an index, so reordering or renaming carries no expansion across. The two paging levels stay separate: `collapseFolder` touches only `folderCaps`, `collapseRows` only `groupCaps`, and only a mode switch or a Project change resets both. Ship after the fixes below, all of which I have pushed.

### Findings

**1. The same display rule ends up written twice** — `packages/web/src/lib/session-grouping.ts:163` vs `packages/web/src/components/layout/sidebar.tsx:1474-1485` (at c91818df5) · *should fix*

The PR's thesis is that a folder now obeys the active list's rule, but the active list kept computing that rule inline: its own `hiddenRowCount` call and a verbatim copy of `cap > SIDEBAR_PAGE_SIZE && parts.active.length > SIDEBAR_PAGE_SIZE`. Two copies of one rule is exactly the thing that drifts — a later change to `canCollapse` in the helper would silently leave the list above it behaving differently. `renderGroupBody` now reads the window from the helper, and since the helper is no longer folder-specific `folderRevealPlan` became `revealPlan`. The substitution is an identity, not a behaviour change: `orderedActive.slice(0, plan.shown)` equals `orderedActive.slice(0, cap)` because `plan.shown = min(cap, loaded)` and `orderedActive.length === parts.active.length` (`orderSessionRows` is a permutation); `plan.hidden` is the `hiddenRowCount` call it replaces with the same `shown`; `plan.canCollapse` is the same expression. ✅ Fixed in 8b6ef14b5

**2. The docs and the changelog lead describe the wrong ordering for "Show less"** — `packages/docs/content/web-app.en.md:55`, `web-app.zh.md:55`, `changelog/unreleased/2026-09-12-sidebar-folder-reveal.md:14` · *should fix*

Both said the reveal row disappears and a **Show less** row takes its place. Take a folder with a 28-row share: after one "More" that lands a page, `cap = 20` and `loaded = 20`, so `canCollapse` is true while `hidden = 28 - 20 = 8` is still positive — `FolderSection` renders both rows, which is what its own comment says ("the two can stand at once", `group-list.tsx:221`). A reader following the docs would expect "Show less" only at the end. Reworded in both languages, and the changelog lead and `revealPlan`'s doc comment along with them. ✅ Fixed in 8b6ef14b5

**3. The changelog entry's `Scope` omits `docs`** — `changelog/unreleased/2026-09-12-sidebar-folder-reveal.md:5` · *nit*

The PR edits `packages/docs/content/web-app.en.md` and `.zh.md`, and the convention is to name that: `changelog/unreleased/2026-09-10-dock-launcher-icons.md` is `web`, `docs` for #680, which touched the same two files. Now `` `web`, `docs` `` in both halves. ✅ Fixed in 8b6ef14b5

**4. Two boundaries the tests could not catch** — `packages/web/test/group-paging.test.ts:141` · *nit*

"offers no collapse on a folder that fits in its first page" passes `loaded: SIDEBAR_PAGE_SIZE`, so it cannot tell "folds back to ten" from "folds back to nothing" — the post-collapse state of a folder that actually has rows past the cap was unasserted, and so was the 11-row boundary. Added `revealPlan({ cap: 10, loaded: 10, total: 11, … })` → `hidden: 1` then `{ cap: 20, loaded: 11, … }` → `shown: 11, hidden: 0`, and `revealPlan({ cap: SIDEBAR_PAGE_SIZE, loaded: 28, total: 28, fullyLoaded: true })` → `{ shown: 10, hidden: 18, canCollapse: false }`. ✅ Fixed in 8b6ef14b5

**5. Pre-existing: the reveal row is live and enabled while the folder's first-expand fetch is still in flight** — `packages/web/src/components/layout/sidebar.tsx:885-889` · *nit, not introduced here*

`toggleFolder` fires `loadMoreFor` without a `pendingLoads` marker, so `pending` is false for that window while `hidden = total - 0 > 0` already renders an enabled "Show N more chats" over an empty folder body. Clicking it calls `trackedLoadMore` for the same (Agent, category, scope) pair while the first request is unresolved; `loadMoreFor` cannot filter it out (`pageState` has no entry yet), both requests read `offset: 0` and return the same rows, and `sessions.tsx:428` advances `fetched` once per landing — 20 for one page of unique rows, so the next "More" skips a page. This behaves identically at the merge base, so it is not this PR's regression and I have not touched it; the fix is to mark the first-expand fetch in `pendingLoads` like every other, which changes when the row reads as disabled. ⬜ Left for the author

**6. The wiring has no behavioural coverage** — *nit*

`revealPlan` is well tested, but nothing exercises `revealFolderMore` / `collapseFolder` / the `folderCaps` key / the reset on a mode switch — a folder cap keyed by index, or an `onLess` wired to the wrong key, would pass the whole suite. `packages/web/e2e/paging.spec.mjs` already walks exactly this dance for the active list ("a group reveals its own remainder one page at a time, and 收起 folds it back"); the folder equivalent belongs beside it. Playwright is not available in my environment, so I have not written it. ⬜ Left for the author

### Verification

```
Merged origin/main into the branch first (it was 11 commits behind; #705/#707/#710 among them
touch packages/web). Clean merge, no conflicts.

vitest on the build box (vps3), never locally:
      139 files, 1907 tests, 0 failures
  vps3-vitest.sh … -- pnpm --filter @prismshadow/penguin-web exec vitest run \
        test/group-paging.test.ts --reporter=verbose
      22 tests, 0 failures (12 under revealPlan)

Local:
  pnpm install --frozen-lockfile                      ok
  pnpm -r build                                       ok
  pnpm typecheck                                      ok
  pnpm lint                                           ok (oxlint)
  pnpm format && pnpm format:check                    ok
  pnpm --filter @prismshadow/penguin-web build        ok
  git diff --check                                    ok

Not run: packages/web/e2e (Playwright not installed on the build box).
```

PR #714, opened 2026-09-12
feat(server,web): a profile page sets the user's avatar and nickname

hiyouga:
Every user can now set an avatar and a nickname. The sidebar's user entry and the user menu's new header show them (the uploaded avatar, else the initial-letter tile; the nickname, else the user id), and a **Profile** page heads the personal group of system settings — visible in every session, the desktop shell included.

Design half: [penguin-harness-design #157](https://github.com/Prism-Shadow/penguin-harness-design/pull/157).

## What ships

- **Database**: migration 5 `user-profile` (swap-safe) adds nullable `display_name` and `avatar` to `users`, guarded by `PRAGMA table_info` so a database whose declarative track already created them adopts cleanly; `down` drops both (and loses every stored nickname and avatar, as its comment says). `schema.ts` declares both for fresh databases. Additive and reversible: an existing user simply has no profile until they set one.
- **API**: `UserInfo` gains `displayName?` / `avatar?` (a data URL); `PUT /api/me/profile` `{ displayName?: string | null, avatar?: string | null }` for any authenticated session (the desktop shell's token session included) — name trimmed, 1–32 characters counted by code point, no control characters; avatar a `data:image/(png|jpeg|webp);base64,…` URL of at most 128 KiB whose payload decodes; `null` clears; a body naming neither field is a 400. The admin users list carries `displayName` but strips `avatar`, so an unpaged list never carries 128 KiB per account.
- **Web**: `profile-section.tsx` (first in the personal group; its buttons on the `sm` rung and the module in the guard's `DIALOG_BODY_MODULES`): a 64 px preview, "change avatar" (browser-side centre-crop and downscale to 128×128, PNG, JPEG at 0.85 when the PNG exceeds 100 KB, an inline error past 128 KiB or for an undecodable file), "remove avatar", the nickname field, one save button; success reported like the other explicit-save page. A shared `UserAvatar` component draws the tile in the sidebar rail, the menu header, the settings page and the admin list. The auth state takes the response so the menu updates without a refetch.
- Docs (`web-app`, `server-api`, both languages) and the changelog pair.

## Tests

Server: `me-profile.test.ts` (set / clear / partial patch semantics, trimming and CJK counting, every rejection with nothing stored, the admin list's nickname-but-no-avatar, the desktop token session may call it); `db-migrations.test.ts` (up adds both columns and is a no-op when present; down removes them; the pre-profile fixtures are built without the two columns so the 0.2.9 round trip still compares equal). Web: `avatar-image.test.ts` (centre-crop geometry and the PNG → JPEG → reject budget through an injected encoder, cap parity with the server), `control-size.test.ts` and `settings-sections.test.ts` extended.

## Verification

Remote test host (Node 24.20): `pnpm typecheck` clean across every package after building core, server, web and cli; `pnpm format:check` clean; server full suite 126 files, 1830 passed; web full suite 138 files, 1878 passed; docs 53 passed.


hiyouga:
## Review

Adds a Profile page that gives an account an avatar and a nickname, a `users` column pair behind it, `PUT /api/me/profile`, and a shared `UserAvatar` the sidebar, the collapsed rail, the account menu and the admin list all draw through. The server side is sound: the route writes only `c.var.user.userId`, no admin path writes another account's profile (`/api/admin/users` is create / reset-password / delete only), the avatar is stored as a `data:` URL and only ever reaches the DOM as an ``, and `image/svg+xml` is refused at the pattern — so there is no route by which it becomes markup. I have reworked the page to the owner's model (finding 1) and merged `main`; the rest below is what reviewing it turned up, none of it blocking.

`main` was 11 commits behind and merged clean, no conflicts.

### Findings

**1. The page staged the avatar behind a page-level Save** — `packages/web/src/features/settings/profile-section.tsx` · *changed on request*

Reworked so that **every control on the page performs a write the moment it is used**, and the button that writes a value stands beside that value. The only thing ever staged is the text sitting in the input, which is inherent to a text field — which is also what keeps the nickname's two buttons from disagreeing: if Restore default merely emptied the box you would have one button next to the field that writes and one that does not.

- **Layout — four controls, two per row, never four in a row.** Avatar row: `64px preview · Change avatar · Restore default`. Nickname row: `field · Save · Restore default`. The alternative I rejected is the one the PR had — a single action row at the foot of the page — which puts the button away from the value it changes and would make one "Restore default" ambiguous between two rows. `SectionShell` is gone from this page for the same reason: its entire job is that trailing action row, and its own doc says a page that applies on the spot passes no actions.
- **"Remove avatar" → "Restore default", in both catalogs and in the key.** `profile.removeAvatar` is gone; `profile.restoreDefault` is shared by both buttons. It never deleted anything the app cannot draw again — it restores the letter tile — so the wording now says that. It is a real `Button variant="secondary"`, the same affordance as the picker beside it.
- **The nickname's default is the user id** — the username you sign in with. There is no email anywhere in `UserInfo`; every surface already spells the fallback as `user.displayName ?? user.userId` and `UserAvatar` takes the letter from the same pair. So Restore default is `PUT /api/me/profile {"displayName": null}`, a real write that persists, and the id comes back on the sidebar, the rail, the menu header and the admin list in the same commit — `setUserInfo(res.user)` feeds the server's row straight back into the auth state those four read.
- **Nothing is optimistic.** The preview and every other surface draw `user.avatar`, which only moves when the server's row comes back, so the displayed avatar always agrees with what is stored. While a write is in flight every control is disabled and a `Saving…` line shows under the rows; a failure replaces it with the message inline and leaves the stored value untouched. There is no state in which a picture appears and then silently disappears on the next load. The encode is inside that busy window too — decoding a phone photograph onto a 128×128 canvas is long enough to click through twice.
- **Disabled states follow `uploads-section.tsx`** rather than inventing a third convention: the `disabled` prop plus `Button`'s own `disabled:opacity-60`. Save is dead when the box would store what is already stored (whitespace-only edits included); each Restore default is dead when that field has nothing stored; all of them, and the picker, are dead while a write runs. The picker is a `<label>` and takes no disabled state, so it gets `pointer-events-none opacity-60` — a `cursor-not-allowed` beside `labelButtonClass`'s `cursor-pointer` would be decided by stylesheet order, not by the class string.
- **Control-size guard**: all three `Button`s pass `size="sm"`; the module was already in `DIALOG_BODY_MODULES` and `control-size.test.ts` passes.
- **i18n**: `restoreDefault` and `restoreDefaultOf(subject)` added to both catalogs in the same shape, plus `avatarInfo` disclosed behind the row's "?" (when a change takes effect is semantics, so it discloses rather than sitting in a `hint`). Two buttons reading "Restore default" on one page is not a name, so each carries `aria-label={restoreDefaultOf(…)}` — "Restore default: Avatar" / "Restore default: Nickname" — following `common.moreInfoAbout`'s existing pattern, and keeping the visible text a prefix of the accessible name.
- **Regression test**: `packages/web/test/profile-form.test.ts` pins the half a node-only vitest can reach — what each nickname button stores and when it is live — through a pure `profileControls` in `src/lib/profile-form.ts`, for exactly the reason `settings-sections.ts` holds the dialog's visibility rules outside the dialog. The avatar's apply-on-pick is a React effect of the pick handler and this package has no DOM (`environment: "node"`, no jsdom, no testing-library), so it has **no unit test** — saying that plainly beats a test that asserts the source text.
- Docs (`web-app.en.md` / `.zh.md`) and the changelog pair rewritten to the page as it now is. The Settings intro sentence, "Personal … apply the moment they are touched — no Save button", was made false by this page and now names the nickname as its one exception, in both languages.

✅ Fixed in b6fade0ab

**2. `displayName` rejects control characters but not the bidi and zero-width ones** — `packages/server/src/http/routes/me.ts:70` · *should fix*

`/\p{Cc}/u` is C0/C1 only. `PUT /api/me/profile {"displayName": "\u202Eexecutive"}` is 10 code points, survives `trim()`, matches nothing in the guard, and is stored — U+202E is category **Cf**, not Cc (verified on node 24). It is then rendered verbatim in the sidebar's user row, the account-menu header and, the one that matters, the admin **Users** table, where the nickname sits directly under another account's id and a right-to-left override reverses the visual order of what follows it on that line. U+200B likewise survives `trim()`, so two nicknames can render identically while comparing unequal.

The narrow fix is to reject the bidi controls specifically — U+202A–U+202E, U+2066–U+2069, U+200E, U+200F — and the zero-width space, rather than all of `\p{Cf}`, which would also reject the ZWJ that holds an emoji together. ⬜ Left for the author: narrowing what a published endpoint accepts, and rewording the "no control characters" sentence that now stands in four documents, is your call rather than mine.

**3. Nothing bounds the decoded image, only the base64 string** — `packages/server/src/http/routes/me.ts:77-99` · *nit*

Length, pattern and "the payload decodes" are all checks on the string. A ~98 KiB PNG of uniform data decodes to roughly 100 MB of pixels (deflate tops out near 1032:1), i.e. about 5000×5000 RGBA, and passes every check here. The server never decodes the pixels, so the server is fine; the browser does, on every surface the avatar is drawn. Reach is the account's own browser only — the Profile page always re-encodes to 128×128, `GET /api/me` returns your own row, and `listUsers` strips avatars — so it is self-inflicted, and a real bound needs an image decoder the server does not carry. ⬜ Left, stated so it is a known bound rather than an assumed one.

**4. `decodedLength === 0` cannot be reached** — `packages/server/src/http/routes/me.ts:98` · *nit*

The pattern requires at least one payload character, node's `atob` throws on a length ≡ 1 mod 4 and on a payload of only `=`, and the shortest decodable payload (`AA==`) yields one byte — so the length is never 0 by the time this line runs. Harmless, and defensive against a platform whose `atob` is laxer. ⬜ Left rather than trimmed, because removing a guard against a runtime's base64 behaviour is not obviously an improvement.

**5. One assertion in the new avatar test compares two constants that sit eight lines apart** — `packages/web/test/avatar-image.test.ts:91` · *nit*

`expect(AVATAR_PNG_BUDGET).toBeLessThan(AVATAR_MAX_CHARS)` can only fail if someone edits both in the module under test. Its sibling on line 90 does earn its place — pinning 131072 is the only thing tying the browser's budget to the server's, which is not exported and so cannot be imported. ⬜ Left.

### Left for you

**The migration is a data-on-disk question, so the strategy is yours.** What I established, rather than a recommendation:

- **What it does.** Migration 5 `user-profile` (`packages/server/src/db/migrations.ts:222`) runs on the first open by a build carrying this PR: one transaction adding `users.display_name TEXT` and `users.avatar TEXT`, both nullable, no default, no existing row rewritten, then `PRAGMA user_version = 5`. Every existing account reads NULL for both and keeps showing its user id and letter tile. Nothing for an operator to do.
- **Reversible, and lossy.** `down` is declared and drops both columns; every stored nickname and avatar goes with them, nothing else on disk holds a copy, and there is no export path for an avatar. Reachable only through `rollbackTo` — the swap path never calls it.
- **Rolling back to the previous build after it has run — I traced both shapes and neither breaks.** *Hot-update swap*: `swapSafe: true`, so it may apply while a pushed platform boots; if that boot fails the runtime reverts and the schema stays at 5, where the predecessor's `UsersRepo.insert` names its own five columns explicitly and the two nullable ones take NULL. *Plain downgrade to an older release*: its `LATEST_VERSION` is 4 while the file reads 5, and `migrate()` filters `m.version > from`, so nothing is pending and nothing errors; the columns sit unread and the values come back intact if the newer build is reinstalled.

So: additive, forward-safe, backward-safe, destructive only down the deliberate `rollbackTo` path. By the letter of the rule there is nothing that breaks if nothing is done, so a `changelog/unreleased/2026-09-12-backward-compatibility.md` pair may not be owed — but it touches data already on disk, which makes it your call. The options:

1. No compatibility entry; the changelog's migration bullet already says "nothing to do for an existing install".
2. Add the pair anyway, recording that `rollbackTo(db, 4)` destroys every stored avatar and nickname with no backup or export path, and naming who removes the note and when.
3. Ask for a non-destructive `down` before merging — parking the two columns rather than dropping them — which is a real change to the migration and not one I would make on your behalf.

**One storage consequence belongs with the same decision.** An avatar is up to 128 KiB of base64 in a `users` column, so `web.db` grows by that per account with a picture, the value rides on every `GET /api/me` — including the reconnect probe at `packages/web/src/features/chat/use-session-stream.ts:268` — and the only way to back one up is to back up the whole database. If files on disk beside the Traces would be the better home, this is the cheap moment to say so.

### Verification

```
vitest on vps3 (this box does not run it):
  vps3-vitest.sh …/rv-714 rv-714 -- pnpm --filter @prismshadow/penguin-web test       141 files / 1911 tests passed (exit 0)
  vps3-vitest.sh …/rv-714 rv-714 -- pnpm --filter @prismshadow/penguin-server test    127 files / 1839 tests passed (exit 0)
  vps3-vitest.sh …/rv-714 rv-714 -- pnpm --filter @prismshadow/penguin-docs test      7 files / 53 tests passed (exit 0)

locally, after merging origin/main (clean, no conflicts):
  pnpm install --frozen-lockfile   ok
  pnpm -r build                    ok
  pnpm typecheck                   ok (23 projects)
  pnpm lint                        ok (oxlint, exit 0)
  pnpm format && pnpm format:check ok
  git diff --check                 ok

packages/web/e2e not run (Playwright is not installed on the build box).
```

hiyouga:
## The profile rows keep their buttons on one line

Reported by the repo owner: in Chinese the nickname's **保存** and **恢复默认** each broke across two lines and stood twice as tall as the row beside them. Fixed in 4a45c87a7.

### Why only Chinese

Flex will not shrink an item below its min-content width. For a CJK label that width is **one character** — the browser may break between any two of them — so a squeezed button collapses to two characters wide and stacks its label. An English label is held open by its longest unbreakable word. That asymmetry is the whole bug, and it is why the row looked fine in one language and not the other.

The buttons now take `shrink-0` and the nickname field `min-w-0`, so the field gives up the width instead.

### Measured, not eyeballed

Rendered against the **built stylesheet** at settings-dialog widths, button height in pixels, one line being 26 and two being 42:

| container | 保存 | 恢复默认 | Restore default |
| --- | --- | --- | --- |
| 400px before | 26 | 26 | **42** |
| 380px before | 26 | 26 | **42** |
| 340px before | **42** | **42** | **42** |
| 280px before | **42** | **42** | **42** |
| any width after | 26 | 26 | 26 |

Two things that table settles:

- The Chinese threshold is **340px**, which matches the report.
- **English was wrapping too**, at every width tested including 400px, where the Chinese row still fit. Nobody had reported it. The same change fixes it.

The safety question with `shrink-0` is whether rigid buttons start pushing the row wider. They do not: the row's horizontal overflow is identical with and without the change at every width — 0px at 400, 3px at 340, 63px at 280. That overflow is the field's own minimum and predates this PR; holding the buttons rigid costs nothing it did not already cost, and `min-w-0` lets the field give back more than it did before (146px against 168px at 280).

### Also changed

The avatar row gets the same treatment. It escaped only because its two controls are narrower, not because it was built differently — same flex row, same missing `shrink-0`.

### Verification

```
web suite on vps3   141 files / 1911 tests passed
pnpm typecheck / pnpm lint / pnpm format:check / git diff --check   pass, clean
layout probe        real built CSS, 6 widths, both languages, before and after
```

vitest ran on the remote build host.

PR #715, opened 2026-09-13
feat(web): UI workbench panel — pick an element in the preview, send it to the conversation

bhh3349:
## What this changes

A right-dock panel for UI design work, plus the two places it has to touch to exist.

Point the panel at a dev server you are already running, and it embeds that page in the dock. Pick an element in it, and the panel hands the conversation a reference the agent can act on: the element's name, selector, text, size and styles, and — the part that makes it useful — **where that element is written in the project**, resolved back to `file:line:column` through the page's own source map, without the project installing anything or being modified.

This is the whole of what the branch does. It is one feature, so it lands as one PR: `feat(web): UI workbench panel`, `feat(web): state the four limits a picked location comes with`, `docs(web-app): describe the UI workbench panel`.

**Four honest tiers instead of one.** A location is `exact` (file, line, column), `ambiguous` (several candidates share that place in the source — reported with the count, never as exact), `file-only` (the framework names the file and no line, so none is invented), or absent — and when it is absent the panel says which of four situations produced it (no source map in the page; the element was created by framework or library runtime code; the framework reported nothing; a position was found that did not map back) rather than looking broken. The same honesty is on screen before the page is trusted: a **What it supports (L1)** fold states which frameworks were measured into which tier, and the four limits every location comes with (a location is where the JSX is written, not where a CSS rule lives; a style change may be landed in the stylesheet instead; an ambiguous pick may point at the wrapping element; picks return through the page's own console, which is not authenticated — which is why the payload is shown before it can be sent).

**Why this touches `packages/desktop`.** The preview has to embed the user's own dev server and inject the picker into that page's main world, and the guest must stay a guest: it keeps `contextIsolation: true`, `nodeIntegration: false`, `sandbox: true` and gets **no preload at all** — the only change to the shell is `webviewTag: true` on the existing window. The server's `/preview/:token/*` route cannot stand in for it: it serves static files out of the Session Workspace on an isolated origin and is deliberately not a reverse proxy, so it cannot front an arbitrary dev server. Everything else lives in `packages/web` and reads the page: the address probe and its failure classification, the injected picker, runtime framework sniffing, source-map resolution (inline or sibling `.map`), the payload, the send-time re-resolution, and the panel itself. Nothing is written to the user's project — a run that measured it left a real repository's `git status` and 112 file hashes unchanged — and the injected script is checked, as a unit test over the exact text the guest evaluates, for network primitives and host/storage capabilities.

Closes nothing; there is no issue for it.

## Verification

The chain CI runs, on this branch:

- `pnpm exec prettier --check .` — `All matched files use Prettier code style!`
- `pnpm --filter @prismshadow/penguin-web build` — ✅ 3.55s.
- `pnpm --filter @prismshadow/penguin-web test` — **2056 passed / 2056** (147 files), the ones that would have failed without this change included: the payload's four tiers and its schema, the picker's serialized script (the static half of "no network primitives"), source-map lookup, the framework sniffer, send-time re-resolution, the panel state's per-selection caching, and the copy invariants (`test/workbench-copy.test.ts` pins the matrix and the four limits, in both dictionaries, so a later edit cannot quietly promise a framework nobody measured).
- `npx oxlint` — 0 warnings, 0 errors.
- `packages/web/e2e/dock.spec.mjs` (the browser suite) — **9 passed, 1 failed**: `dock.spec.mjs:272` `getByText("根目录")`. That assertion is on `main` already (line 249 there) and the workspace tree's root segment has not been drawn as `根目录` since before this branch — `packages/docs/content/web-app.en.md` describes it as `.`. It is unrelated to this change and fails on `main` too; this branch's edit to that file only extends one loop with `"workbench"` and adds a test for the new panel.

Beyond CI, the panel was measured on a real desktop shell rather than a browser — 19 Electron runs, kept as raw evidence outside this repository (a private design repo; happy to attach any of them on request). The ones that say the most about the feature:

- A real project (`tokenrhythm-workbench`, 65 pickable elements): `exact` on 32 of 32 checked against an AST-derived truth table, **zero** elements reported exact with a wrong line; `ambiguous` on 33, all with a candidate count.
- An unmodified repository, put through the whole flow: `git status --porcelain` unchanged and 112 file hashes identical.
- A production build (real `vite build` + `vite preview`): degraded, not broken — the status line names the missing source map, the payload carries no invented file, and picking, adding and sending still work.
- `click` → the panel's card on screen, in a dev build: median **1 ms** (fixture) and **2 ms** (the real project), against the 100 ms target this feature set for itself.

## Checklist

- [x] Changelog entry pair under `changelog/unreleased/` — `2026-09-13-ui-workbench-panel.md` and `2026-09-13-ui-workbench-panel.zh.md`, in the format `changelog/README.md` describes.
- [x] Docs updated where this changes documented behavior — a **UI Workbench Panel** section in `packages/docs/content/web-app.{en,zh}.md`, the page every other docked panel is described on.
- [x] No API key, bot token or other credential in the diff, the logs, or the screenshots.


hiyouga:
Could you add me on WeChat (hiyouga)? I'd like to discuss how to merge this PR, since it's too big to merge directly.

PR #716, opened 2026-09-14
feat(models): add Penguin Go authorization and catalog sync

laodouuu:
## Summary
- Standardize the provider identity as **Penguin Go** across the UI, provider id (`penguin-go`), exported constants, server configuration, environment variables, tests, and bilingual documentation.
- Add a built-in **Penguin Go** model group with the same group-level “authorize a key” interaction as TokenDance, while keeping the platform-specific device start/poll protocol behind the PenguinHarness server.
- Preset the relay’s supported generation models and stable production routes, then use the catalog returned by authorization or manual sync to add newly available models and refresh platform-owned prices.
- Keep platform credentials out of the browser and fix Electron authorization so Windows opens the real HTTPS page without an `about:` protocol dialog.

## User experience
- The Models page now shows Penguin Go as a regular built-in provider group with a monochrome Penguin provider glyph.
- **Authorize key** starts only after an explicit click, opens the platform authorization page, polls for completion, and reports whether the key was applied.
- A configured group gets a separate **Sync** action immediately before **Add model**. Sync uses the stored key, adds missing platform models, refreshes changed prices, and reports either the same added/updated counts as preset sync or that the catalog is already current.
- If the stored platform key is rejected, Sync returns the user to authorization. The authorization action remains available so the owner can replace the group key with one from another account.
- Account profile, balance, and other platform account state are intentionally not stored or displayed.

## Model catalog
- Add the Penguin Go provider and eleven preset generation models under the production relay base URL.
- Keep Penguin Go authorization behavior outside `packages/core/src/agent.ts`; authorized keys continue through the existing per-model credential path.

## Server authorization and synchronization
- Add owner-only routes for start, status, retry, cancel, and catalog sync under `/api/projects/:projectId/platform-auth`.
- Keep flow ids, device secrets, delivered API keys, and retryable apply state in the server process. The browser receives only the local flow id, authorization URL, public status, and applied-model count.
- Validate platform response size, client id, key length, model ids, supported routes, endpoints, context/output limits, vision flags, and USD-per-million-token pricing before writing configuration.
- Preserve upstream 429 backoff as HTTP 429 plus `Retry-After`; polling also honors the platform backoff instead of repeatedly calling upstream.
- Keep connectivity, vision, and protocol probes provider-aware so they use the Penguin Go credential path rather than substituting Google or OpenAI environment keys.
- On authorization, write the delivered key across existing Penguin Go rows and add missing catalog rows. On Sync, keep existing routes and user configuration, update only platform-owned prices, add missing models with their metadata, and never delete local rows.
- Make the trusted platform origin server configuration (`PENGUIN_GO_ORIGIN`), requiring HTTPS except for explicit loopback integration environments.

## Web and desktop integration
- Add typed API clients, localized authorization states/errors, retry countdown handling, model refresh behavior, and the standalone Sync action.
- Preserve platform-fetched Penguin Go prices when the global **Sync presets** action runs, so the two synchronization paths do not overwrite one another.
- Prefill and require the relay base URL for manually added Penguin Go rows.
- Keep `about:blank` as a hidden Electron authorization bridge until `/start` returns, then open only the real HTTPS authorization URL in the system browser and close the bridge. This avoids the Windows protocol-selection popup without consuming a platform start request when the dialog merely opens.

## Data and compatibility
- New Projects receive the built-in Penguin Go rows from the model catalog.
- Existing Projects obtain those rows through the existing **Sync presets** action; no automatic migration or deletion is performed.
- Earlier draft-only local data is intentionally not migrated; reset the development Project before retesting the renamed provider.
- Model keys remain in `.project_config.toml` with the existing masked-response and file-permission behavior.
- Authorization flows are intentionally process-local and expire within ten minutes; restarting the server requires starting a new authorization flow but does not remove an already saved key.

## Verification

- [x] Merged current `origin/main` (`77c74695`) through `9c51dd0f`; Penguin Go now uses the current ModuleTree / `HttpModule.routes` assembly, with no legacy `AppDeps` business assembly restored.
- [x] `pnpm install --frozen-lockfile`
- [x] All workspace package builds: `pnpm -r build`
- [x] `pnpm format:check`, `pnpm lint`, and `git diff --check`
- [x] `pnpm typecheck`
- [x] Windows installer bundle, offline, rollback, and simulated-online tests
- [x] Core model catalog coverage: 42 tests
- [x] Penguin Go server authorization/model/protocol coverage: 69 tests
- [x] Web package: 175 files, 2,322 tests
- [x] Docs package: 7 files, 53 tests
- [x] Desktop authorization URL helper: 11 tests
- [x] Independent merge review confirmed the Penguin Go service and routes follow current `main` module boundaries and did not restore the retired server assembly.
- [ ] Complete an end-to-end authorization against the deployed production platform and invoke one Google-protocol and one OpenAI-protocol model
- [ ] Confirm a later platform catalog addition appears through **Sync** without replacing existing non-price model settings

## Local full-suite note

The complete workspace suite was run on Windows. Tests that do not require Windows symlink privilege passed after isolated reruns: Core reported 1,194 passing tests, Server 1,993, CLI 430, Web 2,322, and Desktop 200. The remaining local failures are pre-existing tests whose fixture setup calls `symlink` on a machine without that privilege (`EPERM`). One Core background-process timing assertion also missed its two-second bound under full parallel load and passed when its file was rerun alone. GitHub CI remains the authoritative Linux/macOS/Windows verification after this push.

## Simplification: promotions live in web.db, Project files keep list prices (2026-09-16)

Pushed as `9215e3b76` at the maintainer's request. Design: [penguin-harness-design#171](https://github.com/Prism-Shadow/penguin-harness-design/pull/171) (PRN-024 gains a deliberate exception; 05「模型价格」, 06「模型配置页」).

- **All groups, not just Penguin Go.** `.project_config.toml` `pricing` is always the list price. A flat promotion is a fraction in a new `model_promotions` table. It replaces the effective-price-in-TOML plus `provider_catalog_cache` fingerprint design, and migration 9 is edited in place because it was never released. Cost center, session cost, the chat's live estimate and company-mode budgets price usage at list × (1 − discount). Budgets are therefore no longer derived from files alone, which is the accepted exception.
- **Writers.** Three places write promotions:
  - Project creation seeds `presetPromotions()` once the `projects` row exists.
  - "Sync presets" declares each preset row's catalog promotion, or `null`. A Penguin Go row keeps the platform's promotion, restated when the merge rewrites its price.
  - Penguin Go authorization / Sync replaces that group's promotions with the platform's.
- **PUT /models** takes `discount` per entry: a number stores it, `null` clears it. An omitted one survives unless the row is renamed or its pricing changed, which clears it. Out-of-range values are a 400 before any write.
- **Platform catalog** is normalised at parse time to list price + `discount`; wire validation is unchanged. `ModelInfo.listPricing`, `PlatformCatalogPromotion`, the price fingerprint and the legacy TOML cleanup are gone.
- **Web.**
  - The badge multiplies the promotion with the catalog off-peak tier. The tier still applies only while the stored price is the catalog's peak.
  - The model dialog notes a running promotion and that editing the price cancels it.
  - The catalog delta counts promotion-only differences.
- **Penguin Go presets** carry list prices only (no static `discount`). Their DeepSeek rows follow DeepSeek's current lineup: `deepseek-flash` and `deepseek-v4-pro` shown as "DeepSeek V4 Pro 0813"; `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp` are removed.
- **Existing Projects are not migrated**, as decided. Rows holding discounted numbers keep billing at them until "Sync presets" rewrites list price + promotion.
- **Changelog.** Two bilingual entries: `2026-09-16-penguin-go` and `2026-09-16-model-promotions-in-database`.

### Verification (simplification)

On the remote test host (vps3), fresh sync of `9215e3b76`'s tree:

- `pnpm -r build`, `pnpm typecheck` and `pnpm format:check` are clean; `pnpm lint` reports 0 warnings and 0 errors.
- `pnpm -r --no-bail test`:

  | Package | Files | Tests |
  | --- | --- | --- |
  | core | 58 (+1 skipped) | 1224 passed |
  | server | 141 | 2054 passed |
  | web | 175 | 2332 passed |
  | cli | 31 | 440 passed |
  | desktop | 17 | 217 passed |
  | docs | 7 | 53 passed |
  | landing | 10 | 71 passed |

- The host-gated live tests in `plugins/sandbox-dsh` (2) and `plugins/sandbox-bwrap` (3) fail on that host identically on an untouched `origin/main` tree.


hiyouga:
Pushed `9215e3b76` on top of this branch: promotions now live in `web.db` (`model_promotions`) and every Project file keeps list prices, replacing the effective-price + `provider_catalog_cache` design. Details and verification are appended to the PR description; the matching spec change is Prism-Shadow/penguin-harness-design#171.

PR #717, opened 2026-09-14
fix(web): the session row's marks line up, and a Chinese date reads 8月30日

hiyouga:
In the sidebar's session list the marks right of a title — the run glyph, the background-task and schedule marks, the approval count — drifted from row to row, because the trailing time slot's width followed the time string (「刚刚」 vs 「31 分钟前」 vs 「7 月 1 日」) and the marks sit against that slot. The slot now takes a fixed width per interface language, so the marks form one vertical column; and the Chinese month-day is written without spaces, `8月30日`, with the English `Aug 30` unchanged.

Design half: [penguin-harness-design #158](https://github.com/Prism-Shadow/penguin-harness-design/pull/158).

## What ships

- `sidebar.tsx`: `SessionRow` takes the `locale`; the trailing slot is `w-[4.5rem]` in Chinese (room for 「12月31日」 and 「59 分钟前」) and `w-12` in English (room for "Dec 31"), never narrower than the archive / delete hover pair it swaps with; the time is right-aligned in tabular figures. The slot's comment explains the choice.
- `formatMonthDay` (`lib/format.ts`): zh `8月30日`. It also formats a one-off schedule's date on the scheduled-tasks rows and the version line's "last updated" date, which change the same way; the comments and the zh docs that quoted the spaced form were updated.
- Tests: `format.test.ts` and `schedule-describe.test.ts` pin the new form. Changelog pair.

## Verification

Remote test host (Node 24.20): `pnpm --filter @prismshadow/penguin-web typecheck` clean after building core, server, web and cli; `pnpm format:check` clean; web full suite 139 files, 1898 passed; docs 53 passed.


hiyouga:
## Review

Two changes ride here: a layout fix (the trailing time slot takes a fixed width per language so the marks stop drifting) and a format fix (the Chinese month-day loses its spaces). Both are right in substance — I measured the alignment claim in a headless browser and it holds: with the fixed slot the rightmost mark lands at the same x on every row, where `min-w-12` put it at 194.75px on one row and 188.75px on the next. Ship after the fixes below, two of which I have pushed.

### Findings

**1. The row's month-day named the UTC calendar day, not the reader's** — `packages/web/src/lib/format.ts:296,304` · *should fix*

`formatRelativeShort` handed the raw `...Z` timestamp to `formatMonthDay`, which reads the date fields straight out of the string. Concretely, with the process pinned to each zone:

```
TZ=Asia/Shanghai    2026-08-30T23:00:00Z  reader's clock 2026-08-31 07:00  ->  row read 8月30日
TZ=America/New_York 2026-09-01T02:00:00Z  reader's clock 2026-08-31 22:00  ->  row read Sep 1
```

So every row older than a week was off by a day for readers a few hours either side of UTC — wrong in both directions, and the more visible now that the PR is about that very string. Pre-existing, but this repo already has the answer one file over: `schedule-describe.ts` converted to the local calendar date first, with a comment saying the conversion "has to happen here, on the Date". Fixed by exporting that helper as `localYmd` from `format.ts`, routing both of `formatRelativeShort`'s fallbacks through it, and deleting `schedule-describe.ts`'s private copy — one representation of one fact. The version line keeps reading `BUILD_DATE` verbatim, which is a UTC calendar date by construction and must not be shifted. ✅ Fixed in 132fa52

Its test came with a trap worth naming: `formatRelativeShort("2026-07-01T00:00:00.000Z", …)` was zone-independent *because of* the bug, and would have started depending on the runner's zone the moment it was fixed. Those three assertions now build their instants from local parts, like every other assertion in that block and like the whole of `schedule-describe.test.ts`; a new case then pins two zones explicitly, so the regression is caught on a UTC runner too, where an ambient-zone test would quietly pass. Both halves earn their place — with the fix reverted, the suite fails twice:

```
× a week or older … fall back to the absolute month-day    expected '6月30日' to be '7月1日'
× that month-day is the reader's own calendar day …        expected '6月5日' to be '6月6日'
```

The first of those is the *existing* assertion, rebuilt from local parts: the test box happens to sit east of UTC, so `2026-07-01 00:00` local is the 30th in UTC and the old hard-coded `…Z` literal was hiding it in plain sight.

**2. The English slot was pinned at exactly the width it has to hold** — `packages/web/src/components/layout/sidebar.tsx:2793` · *should fix*

`w-12` is 3rem, which is the old `min-w-12` carried over rather than measured. Measured in headless Chromium with the app's own font stack, at the `sm` font tier (`state/theme.tsx:44`, root 16px — the tight one, because the slot is rem and the time inside it is a fixed `text-[11px]` that does not shrink with it):

```
en slot 48.00px   "Nov 30" 48.00px   0.00px spare
                  "Aug 30" 48.00px   0.00px spare
                  "Dec 31" 47.00px   1.00px spare
zh slot 72.00px   "12月31日" 58.00px  14.00px spare
```

Zero spare means the `px-1` inset is already gone, and the failure past it is not a clip: `whitespace-nowrap` in a `justify-end` flex box overflows to the *left*, over the marks the PR is lining up (I reproduced it — the time's left edge landed 25px inside the last mark). The two languages were also not held to the same standard. Fixed by `w-12` → `w-14`; that restores 8–9px of spare at the tightest tier and leaves the comment's "sized for the widest string" claim actually true. ✅ Fixed in 132fa52

**3. The changelog names the wrong hover pair** — `changelog/unreleased/2026-09-14-session-row-time-alignment.md` · *nit*

"the hover pair of archive and delete buttons" — `HOVER_ROW_ACTIONS` is `["archive"]` (`components/ui/session-row-menu.tsx:78`) and the second button is the ellipsis that opens the context menu; delete has lived in that menu only. Corrected in both files, along with the 3rem → 3.5rem figure. ✅ Fixed in 132fa52

**4. The same "archive and delete" claim is stale in two comments and in both docs** — `sidebar.tsx:20`, `sidebar.tsx:2797`, `packages/docs/content/web-app.en.md:48`, `web-app.zh.md:48` · *nit*

The module header says the slot "swaps to archive + delete icon buttons", the comment two lines under the slot explains the hover feedback as "red for delete" (no danger action is reachable from the hover pair any more, so nothing there turns red), and both docs tell the reader the hover reveals "archive and delete". All four predate this PR, so I left them alone — but the PR is already editing `web-app.zh.md` and rewriting that comment block, which makes this a cheap sweep to fold in. Doing it means touching `web-app.en.md` too, so the pair stays mirrored. ⬜ Left for the author

### Left for you

- **`2027 年 1月3日 10:00`.** `strings.ts:1248`'s `onDateWithYear` still spaces the 年 while the month-day no longer spaces 月/日, so the one string that carries all three renders half-converted. `${days} 天前` keeps its space too, so there is a defensible reading either way — it is your wording call, not a bug, and I did not touch it.
- **No year, ever, in the session row.** The fallback is month-day in both languages at the same 7-day cutoff, so a row last active in 2024 and one from this June both read `Jun 5` / `6月5日`. That is pre-existing and reasonable for a slot this narrow — but note the new fixed width has no room for a year if it is ever revisited.

### Verification

```
Merged origin/main into the branch first — it was 3 commits behind (#707, #713,
#712). #713 and #712 both touch sidebar.tsx; git merged them cleanly with no
conflicts, and `git diff origin/main HEAD` is exactly this PR's own hunks.

vitest on the remote build box, never on the review machine:
      139 files, 1908 tests passed
  pnpm --filter @prismshadow/penguin-web exec vitest run test/format.test.ts
      with the fix reverted: 2 failed | 43 passed  (the two quoted above)
  pnpm --filter @prismshadow/penguin-docs test
      7 files, 53 tests passed

Locally:
  pnpm install --frozen-lockfile                         ok
  pnpm -r build                                          ok
  pnpm typecheck                                         ok, every package
  pnpm lint                                              ok (oxlint, exit 0)
  pnpm format && pnpm format:check                       ok
  git diff --check                                       clean
  pnpm --filter @prismshadow/penguin-web build           ok

The layout numbers are measured, not estimated: headless Chromium, the app's own
font stack, all three font tiers (16 / 18 / 20px root), with the utility values
read out of the built stylesheet (--spacing = .25rem, so w-12 = 3rem,
w-14 = 3.5rem, w-[4.5rem] = 4.5rem, and the hover pair is two w-6 = 3rem).
Playwright e2e was not run.
```

PR #718, opened 2026-09-14
feat: allow agents to register its owned modules, a.k.a. workflow

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← #808 ← **#718** ← #544 ← #448 ← #450 ← #577 ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

It builds on **#808**, the plugin interface check and the travelling TypeScript compiler, which were split out of this PR at the owner's request — everything this PR says about comparing interfaces and about `typescript` is implemented there and only used here.

**Workflows as module trees** — an Agent keeps *workflows* in its own directory: `workflows/<id>/` is a plugin package the server boots as a module tree of its own, shows as tabs beside Chat, reloads on every edit and versions so any edit can be undone. Everything is checked before any of its code runs, and a failed load keeps the previous version serving with the problem named. Design: [penguin-harness-design#194](https://github.com/Prism-Shadow/penguin-harness-design/pull/194) (PRFC-0014).

## What a workflow is

```
workflows/<id>/
  package.json   "type": "module", penguin.modules: the manifests (hand-written; nothing to install)
  index.ts       default export { modules: { Workflow: { create } } } satisfies WorkflowPackage
  ui/            the pages its contributed tabs name
  state.json     its own document, kept by the server across reloads and rollbacks
  .harness/      written once by the server: the types it was written against
  .build/        the server's emitted code, per revision
```

- **Two interfaces cross the boundary.** The host publishes `WorkflowHost` — the SDK's verbs scoped to the workflow's Project: `listAgents()`, `createSession({ agentId? })`, `run(sessionId, [{ text }])` (a busy Session queues it; a Session of another Project is refused; the Agent hears the server, not a person), plus `sessionStatus`, `getState`/`setState`, `log`. The workflow provides `WorkflowMain`, a JSON handler mounted at `…/workflows/:id/api/*`.
- **TypeScript, checked by the server** (`workflows/compile.ts`). `index.ts` only — JavaScript is refused. One compiler program under host-fixed options (`strict`; the folder's tsconfig is not consulted), plus a virtual root assigning the default export to `WorkflowPackage`. Any diagnostic is the load error, with file, line and reason. Emitted into `.build/<revision>/`.
- **Types come from the harness that runs it, not from npm** (`workflows/harness-types.ts`). One machine can run several harnesses — a release, a checkout, a pushed platform with interfaces of its own — and none is a version on npm. The first load renders THIS platform's interface table into `.harness/plugin.d.ts` (what `@prismshadow/penguin-server/plugin` resolves to, through the compiler host) and `.harness/ifaces.json` (the slice it came from), then leaves them alone: the record of what the workflow was written against. Deleting the directory moves the workflow onto the current harness.
- **Interfaces compared across generations, by tsc** — **implemented in #808** (`plugin/iface-check.ts`, shared with plugins), used here for every workflow load. Looking both sides up in the platform's own table compared a declaration with itself — the platform could drop `setState` and every workflow still "passed". Now each side brings its own table, each is rendered as a self-contained `.d.ts`, and the compiler decides both ways: what the workflow requires, what it provides. Nothing is answered from the platform's entry; an expression that cannot be written as TypeScript is refused by name. Plugins get the same comparison at load against the table each package carries — which today holds no copy of the host interfaces a plugin requires (the generator writes only what a package declares: on the chain top, claude-code 1 of 2 and discord-bot 12 of 13 are uncompared), so for plugins those are logged, not fatal.
- **Tabs are contributions.** A workflow contributes to `WebModule.sessionTabs`, the slot a plugin writes, which the host publishes into its tree and scopes to the Agent: several tabs per workflow, pages under `ui/`, returned as `WorkflowInfo.tabs`. The slot gains `title`/`titleZh` and loses the unused `namespace`.
- **Loaded by content.** The folder's content hash is the revision: it names the emit directory (a new import URL per edit), and every successful load is recorded under `workflows-history/` (twenty kept) and restorable; `state.json`, `node_modules` and the dot-directories are no part of it. A watcher reloads on change; loads of one workflow run one at a time.
- **No global prompt.** Nothing is appended to any Agent's system prompt; the penguin-sdk skill documents workflows and is loaded on demand.

## Where it lives

- **Server** — `workflows/{store,service,routes,compile,harness-types}.ts` and the `Workflows` mechanism. The compiler and the interface check it leans on (`plugin/{typescript,iface-check}.ts`, `typescript` as a runtime dependency) are #808's: without them no workflow loads.
- **Web** — `features/workflows/workflow-tabs.tsx` (one tab per contribution, titled in the user's language; reload, history with restore, two-click remove), `lib/workflow-theme.ts` + `public/workflow-ui.css` (a workflow page follows the app's theme), and the full-page route `/app/<project>/<agent>/<workflow>[/<tab>]`, whose only exit is the command palette (Ctrl+P and Ctrl+Shift+P).
- **CLI** — `penguin web --app <project>/<agent>/<workflow>[/<tab>]`.
- **Docs and examples** — the penguin-sdk SKILL.md's Workflows section; *Mission control for your agents* joins the new-chat page's "Build and optimize agents" examples (in place of the docs-RAG one).

Changelog: `2026-08-30-workflows.md` (+ zh).

**For the reviewer:** the PR is unreleased, so workflows written against its earlier revisions (`index.mjs`, no `contributes`) are not migrated. Open questions are in the design PR: workflow code runs in the server process unsandboxed, the page is a same-origin iframe, `.harness/` is outside the recorded versions. Not in this PR: the module tree's own whole-tree check still compares within one table — PRFC-0001's revision.

## Verification

On this tip (vps3, Node 24): `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` clean; `pnpm -r test` green — server 2136, web 2338, cli 442. Tests cover: TypeScript checked and transpiled, JavaScript refused; the harness writing `.harness/` once and nothing being installed; both directions of the interface comparison (an addition on the platform's side passes, an unreadable table fails); several tabs and an unopened slot; the host's verbs scoped to the Project; reload, versions, rollback and removal; the web tab model and full-page route; `--app`.


packages/server/src/workflows/service.ts
```diff
@@ -0,0 +1,697 @@
...
+    try {
+      dir = fs.realpathSync.native(declared);
+    } catch {
+      dir = declared;
+    }
+    let watcher: fs.FSWatcher;
+    try {
+      watcher = fs.watch(dir, { recursive: true }, (_event, filename) => {
+        const id = typeof filename === "string" ? filename.split(/[\\/]/)[0] : undefined;
+        if (id === undefined || !isWorkflowId(id) || filename?.endsWith("state.json")) return;
```
Myriad-Dreamin: **(Blocker) Every `host.setState()` reboots the workflow.**

`writeState` writes `state.json.tmp` and renames it (`store.ts:136-140`). This filter ignores names *ending* in `state.json`, which `state.json.tmp` does not, and the dot-directory guard on the next line does not catch it either. So the create event for `<id>/state.json.tmp` reaches `schedule()`: ~300 ms after every state write the server runs a full `ts.createProgram`, boots a new tree, disposes the live one (all in-memory module state gone) and broadcasts `workflow_updated` to every Project user.

A workflow that persists state per request — the natural shape, and what `getState`/`setState` exist for — never stays up.

Second half: `readFolder` filters only the exact `STATE_FILE` (`store.ts:97`), so `state.json.tmp` is walked into `files`, enters the revision hash and is copied into history.

packages/server/src/workflows/service.ts
```diff
@@ -0,0 +1,697 @@
...
+          ifaces: { [HOST_MODULE]: { host: hostDecl() }, [WEB_MODULE]: { web: webSlotsDecl() } },
+          values: { [HOST_MODULE]: { host }, [WEB_MODULE]: { web: {} } },
+        },
+      });
+      const alias = Object.entries(root.manifest.provides).find(
+        ([, ref]) => ref === MAIN_IFACE,
+      )![0];
+      const main = tree.api<Loaded["main"]>(ROOT_MODULE, alias);
+      next = { folder, tree, main, tabs, loadedAt, error: null };
+      previous?.tree?.dispose();
```
Myriad-Dreamin, line 528: **(Blocker) A throw after this line leaves the workflow serving a disposed tree, and leaks the new one.**

`next` is assigned, the previous tree is disposed here, and then `pruneBuilds` + `await recordVersion` still run inside the same `try`. Any throw from those falls into the catch below, which rebuilds `next` with `previous.tree` / `previous.main` — the tree just disposed — and drops the freshly booted tree, which is then never disposed.

Concrete trigger: `recordVersion`'s `fs.copyFile` (`store.ts:174`) ENOENTs when a file listed at hash time has since been rewritten or removed — and the `state.json.tmp` from the finding above is exactly such a file. An author's own `effect(...)` cleanup rethrowing through `disposeAll` (`core/src/kernel/module.ts:550`) does it too.

Result: a green load is reported as an error, and the handler that keeps answering belongs to an instance whose cleanup has already run.

packages/server/src/workflows/store.ts
```diff
@@ -0,0 +1,220 @@
...
+    else if (e.isFile()) out.push(rel);
+  }
+  return out.sort();
+}
+
+async function hashFiles(dir: string, files: string[]): Promise<string> {
+  const h = createHash("sha256");
+  for (const rel of files) {
+    h.update(rel).update("\0");
+    h.update(await fs.readFile(path.join(dir, rel))).update("\0");
```
Myriad-Dreamin: **(Medium) `hashFiles` is not tolerant of a file vanishing mid-hash, and it runs on every request.**

Unguarded `fs.readFile` per walked path. `readFolder` is called from `current()` on every dispatch and from `list()`, so if a file disappears between `walk` and `readFile` — an editor's temp file, the Agent deleting a file, `state.json.tmp` — `readFolder` rejects and the route answers 500 (only `WorkflowNotFound` is caught) although the loaded instance is perfectly fine.

Everything else in this module is deliberately tolerant (a broken `package.json`, a broken `versions.json`); this path should be too. Related: hashing the whole folder on every `/api/*` call is O(folder) per request — a `ui/` with a bundle or images pays it every time, including each SSE connect.

packages/server/src/workflows/store.ts
```diff
@@ -0,0 +1,220 @@
...
+  try {
+    return JSON.parse(await fs.readFile(path.join(dir, STATE_FILE), "utf8")) as unknown;
+  } catch {
+    return null;
+  }
+}
+
+export async function writeState(dir: string, state: unknown): Promise<void> {
+  const file = path.join(dir, STATE_FILE);
+  await fs.writeFile(`${file}.tmp`, JSON.stringify(state ?? null, null, 2));
```
Myriad-Dreamin: **(Medium) Concurrent `setState` can corrupt `state.json`.**

A fixed `${file}.tmp`: two overlapping writes truncate and interleave into the same path, then both rename. A handler serving two requests that each call `setState` can leave invalid JSON, which `readState` silently turns into `null` on the next load — the document is gone with no error anywhere. A unique temp name per write (and/or serialising writes per workflow) fixes it.

packages/server/src/workflows/routes.ts
```diff
@@ -0,0 +1,263 @@
...
+    const request: WorkflowRequest = {
+      method: c.req.method,
+      path: sub === "" ? "/" : sub,
+      query: c.req.query(),
+      headers: forwardedHeaders(c.req.raw.headers),
+      body: null,
+    };
+    if (c.req.method !== "GET" && c.req.method !== "HEAD") {
+      const declared = Number(c.req.header("content-length") ?? 0);
+      if (declared > MAX_BODY_BYTES) throw tooLarge();
```
Myriad-Dreamin: **(High) The 50 MB cap is applied after the body has been buffered — twice.**

This pre-check reads only `content-length`: a `Transfer-Encoding: chunked` request has none, so `Number(undefined ?? 0) === 0` passes, and `content-length: abc` gives `NaN > MAX === false`, which also passes. What actually bounds it is the global Hono `bodyLimit` (`http/app.ts:71-89`), which with the default attachment limits is about 160 MB — base64(120 MB) plus headroom — and then `c.req.arrayBuffer()` on the next line makes a second full copy.

An authenticated Project member POSTing ~150 MB chunked here a few times concurrently costs ~300 MB peak each in a process that also hosts every Agent and every workflow; the 413 this code promises arrives only afterwards. Cap while reading, or give this route its own `bodyLimit` of `MAX_BODY_BYTES` ahead of the handler.

packages/server/src/workflows/routes.ts
```diff
@@ -0,0 +1,263 @@
...
+    const file = await deps.workflows.uiFile(projectId, agentId, id, decodeURIComponent(rel));
+    if (file === null) throw new HttpError(404, "not_found", "No such file in the workflow's ui/.");
+    const body = await fs.readFile(file);
+    return c.body(body, 200, {
+      "content-type": MIME[path.extname(file).toLowerCase()] ?? "application/octet-stream",
+      "cache-control": "no-cache",
+    });
+  });
+
+  app.all("/:id/api/*", async (c) => {
```
Myriad-Dreamin, line 268: **(Medium) A cross-site GET reaches a workflow handler with the user's cookie.**

`app.all` makes GET a workflow-defined verb, but the origin guard only fires for `WRITE_METHODS` (`auth/middleware.ts:104`, `:141`), and the session cookie is `SameSite=Lax` — which browsers do send on a cross-site **top-level navigation**. So `attacker.example` doing `window.open("https://<app>/api/projects/p/agents/a/workflows/w/api/purge")` runs that workflow's GET handler as the victim; and because `respond()` honours a handler-chosen `content-type`, the reply can render as an HTML document on the app's own origin.

Every other API route treats GET as a read, so this is new surface. Either run `fromThisOrigin` for all methods on this mount (a browser page always carries `Sec-Fetch-Site`/`Origin`), or state in the workflow contract that GET handlers must be side-effect-free.

packages/server/src/workflows/routes.ts
```diff
@@ -0,0 +1,263 @@
...
+ * download, a redirect — and `stream` sends it as it is produced. The same origin already serves the workflow's `ui/` files as
+ * written, so a handler that answers HTML is no wider than a file that is HTML.
+ */
+function respond(response: WorkflowResponse): Response {
+  const status = response.status ?? 200;
+  const headers = new Headers();
+  for (const [name, value] of Object.entries(response.headers ?? {})) {
+    if (!DROPPED_RESPONSE_HEADERS.has(name.toLowerCase())) headers.set(name, value);
+  }
+  const bodiless = status === 204 || status === 304 || (status >= 300 && status < 400);
```
Myriad-Dreamin: **(Low) 205 is a null-body status too.**

204, 304 and 3xx are covered; per fetch, 205 (and 101/103) are also null-body. `{ status: 205, body: {...} }` therefore reaches `new Response(JSON.stringify(...), { status: 205 })`, the constructor throws `TypeError`, and `handleError` turns it into a generic 500 with no usable diagnostic. Same for `{ status: 205, bytes }`.

Also in this function: a `stream` returned with a 3xx/204 status is discarded on the next line without `return()` ever being called on it — harmless for a lazy generator, a leak for an already-running iterable.

packages/server/src/mechanisms/workflows.ts
```diff
@@ -0,0 +1,177 @@
...
+   * Response headers. Naming a `content-type` is what makes the response something other than
+   * JSON; `location` with a 3xx status redirects. `set-cookie` is dropped — a workflow does not
+   * write the app's cookies.
+   */
+  headers?: Record<string, string>;
+  /** JSON by default (`null` when absent); with a `content-type` header, a string sent as written. */
+  body?: unknown;
+  /** A binary body (an image, a font, a download); wins over `body`. */
+  bytes?: Uint8Array;
+  /**
```
Myriad-Dreamin, line 47: **(Medium) The documented "client goes away → your `finally` runs" contract does not hold in the normal case.**

The wiring is right — @hono/node-server cancels the reader on close and `cancel()` calls `iterator.return()`. But an async generator suspended *inside an `await`* — exactly the state of "relaying an upstream response", which is the use this documents — queues `return()` behind the in-flight `next()`. If that upstream never settles, the `finally` never runs: the client is gone, yet the generator, the upstream socket and the workflow's loaded tree stay alive, and there is no idle or total response timeout to reclaim them.

Worth passing an `AbortSignal` into `WorkflowRequest` so the handler can abort at the source, and softening this comment meanwhile.

packages/web/src/features/workflows/workflow-app-page.tsx
```diff
@@ -0,0 +1,100 @@
...
+          agentId={agentId}
+          tab={tab}
+          bare
+          onChanged={() => undefined}
+          onRemoved={() => exit[0]!.run()}
+        />
+      ) : (
+        <div className="flex h-full items-center justify-center px-6 text-center text-sm text-gray-500 dark:text-gray-400">
+          {tab === undefined && failure === null
+            ? S.workflows.loadingHistory
```
Myriad-Dreamin: **(Medium) Two defects in this fallback block.**

1. The loading text is `S.workflows.loadingHistory` — "Loading history…" — used as the *page* loading message, so `penguin web --app …` greets the user with a message about history.
2. The `<br />` below sits directly inside a `flex items-center justify-center` container, where it is just another flex item: the message and `exitHint` render side by side on one line instead of stacked.

Also, this branch only renders `failure` when no tab has been shown yet: once a tab is up, a later list failure (403 after a permission change, a server restart) leaves the stale iframe with no indication at all.

packages/web/src/features/workflows/workflow-app-page.tsx
```diff
@@ -0,0 +1,100 @@
...
+      } catch (err) {
+        if (alive) setFailure(err instanceof Error ? err.message : String(err));
+      }
+    };
+    void load();
+    const onUpdated = (e: Event) => {
+      const d = (e as CustomEvent<{ projectId: string; agentId: string }>).detail;
+      if (d.projectId === projectId && d.agentId === agentId) void load();
+    };
+    window.addEventListener(WORKFLOW_UPDATED_EVENT, onUpdated);
```
Myriad-Dreamin, line 56: **(Medium) This listener is dead on this route.**

`WORKFLOW_UPDATED_EVENT` is dispatched from `applyUserEvent` in `state/sessions.tsx`, fed by the `/api/events` connection that `SessionsProvider` opens — and `SessionsProvider` is mounted only inside `RequireAuth` (`router.tsx:70`). The `/app/...` routes go through `RequireAuthBare` (`router.tsx:121`), outside that provider, so no event stream exists here.

Concretely: open `penguin web --app p/a/wf`, let the Agent edit its `ui/` — the page never reloads, because the iframe is keyed on `uiRev` and only this event refreshes it. If the workflow is deleted, the user keeps staring at a dead frame instead of the "no such page" message. Only a manual browser reload recovers.

packages/web/src/features/workflows/workflow-tabs.tsx
```diff
@@ -0,0 +1,373 @@
...
+  const [active, setActiveRaw] = useState<string | null>(null);
+
+  const refresh = useCallback(async () => {
+    if (projectId === null || agentId === null) {
+      setTabs([]);
+      return;
+    }
+    try {
+      const res = await api.getWorkflows(projectId, agentId);
+      setTabs(workflowTabsOf(res.workflows));
```
Myriad-Dreamin, line 56: **(Medium) Agent-switch race: the response is applied unconditionally.**

No generation counter and no abort, and `ChatPage` stays mounted across an Agent switch. Switch from a slow Agent A to Agent B and A's response can land last: the strip then shows A's tabs while the chat is B's. Clicking one mounts `WorkflowFrame` with B's `projectId`/`agentId` and A's `workflowId`, so Reload / History / Remove hit `/projects/p/agents/B/workflows/<A's id>` — a 404, or, if both Agents have a workflow of the same id, a Remove that deletes B's workflow and its history.

`workflow-app-page.tsx:38-49` does carry an `alive` guard, which makes this look like an oversight rather than a decision.

packages/web/src/api/endpoints.ts
```diff
@@ -1750,6 +1752,38 @@ export const getDesktopTray = () => apiFetch<DesktopTrayStatusResponse>("/api/de
...
+  projectId: string,
+  agentId: string,
+  workflowId: string,
+  revision: string,
+) =>
+  apiFetch<{ workflow: WorkflowInfo }>(
+    `${workflowsBase(projectId, agentId)}/${encodeURIComponent(workflowId)}/rollback`,
+    { method: "POST", body: { revision } },
+  );
+/** Delete the folder; its recorded versions stay on disk. */
```
Myriad-Dreamin: **(Low) This comment is wrong.** The server removes the history directory too (`workflows/service.ts:348-356`), and both the UI confirmation string and the SKILL.md guidance say so. Only this comment claims the versions survive.

Myriad-Dreamin (commented):
Reviewed the server core, the request path and its auth, and the web side. The shape holds up — load serialisation is genuinely correct (the chain sets its map entry synchronously and the `finally` only deletes its own), a failed load does not cause a reload storm, history is bounded, nothing in the background can take the process down, and shutdown disposes properly. On the request path I verified the things that would have been serious and found them sound: **the auth exemption does not widen auth** (`jsonOnlyWrites` runs before `authMiddleware`, the group declares `auth: "user"`, so an unauthenticated caller still gets 401 everywhere), the `WORKFLOW_API` regex is anchored and matches the same string Hono routes on (`%2F` cannot forge a segment), cookie and authorization are withheld from the handler, `ui/` cannot be traversed (double-decoded before `isSafeRelPath`), and compile diagnostics are relative so nothing leaks the server's layout.

Two findings I would fix before merge, both in the server core, both verified in the code — details inline:

1. **Every `host.setState()` reboots the workflow.** The watcher's own filter does not cover the temp file the state write creates, so a workflow that persists state per request never stays up.
2. **A throw after the previous tree is disposed leaves the workflow serving a disposed instance** and leaks the newly booted one — and finding 1 creates exactly the file that triggers it.

Then the request path (body cap applied after buffering; cross-site GET reaching a handler with the user's cookie; the documented disconnect→`finally` contract not holding while awaiting an upstream; 205 turning into a 500) and the web side (the full-page route's only exit stops working once the iframe has focus; that route never sees `workflow_updated`; an Agent-switch race in the tab list).

Test gaps worth closing along with the fixes: concurrent loads, concurrent dispatch, the watcher firing on the server's own `state.json.tmp`, a failure after a successful boot, a file changing mid-hash, `getState()` during `create()`, and `.build/` accumulation across failing loads.

Myriad-Dreamin:
Fixed on `b8d2b40d`…`acad2aaa`. Everything green on the test machine: `pnpm -r build`, `pnpm typecheck` (all packages), `pnpm lint` 0/0, `pnpm format:check` clean, full `pnpm -r test` — core 1223, server 2140 (151 files), web 2340, cli 442, desktop 216, docs 53, landing 71, sandbox plugins 68.

**The two blockers**

- **A state write is no longer an edit.** Staging now goes through a `.penguin-tmp<random>` suffix that both the watcher and the revision hash skip, so `host.setState` neither reloads the workflow nor enters its own document into the hash — and two concurrent writes cannot truncate each other's staging file. New test: after a `setState`, the revision and `loadedAt` are unchanged and nothing is left in the folder.
- **Recording a version cannot unseat a live tree.** `recordVersion` moved into its own `try` outside the one that decides the load: a copy that ENOENTs is logged and the new instance keeps serving. The catch also disposes a tree this attempt booted instead of leaking it, and never hands back the tree it disposed.

**Request path**

- The 50 MB cap is taken **while reading** (`readCapped`), so a chunked body with no `content-length` is refused at the byte that crosses it instead of after ~160 MB is already in memory; the reader is cancelled on refusal.
- The origin check runs on **every method** on the workflow mount, not only writes: a `SameSite=Lax` cookie rides a cross-site top-level navigation, and `app.all` makes GET a workflow-defined verb.
- 101/103/204/205/304/3xx are all treated as null-body, so `{ status: 205, body }` answers 205 instead of throwing into a 500; a stream handed a bodiless status is `return()`ed rather than dropped.
- A header value the platform refuses costs that header, not the whole response.
- The streaming contract now says what actually holds: a generator parked inside an `await` gets `return()` only once that await settles, so a relay should carry its own deadline. Same correction in the SKILL.md, plus a sentence on escaping what a handler echoes, since its responses are served from the app's own origin.

**Web**

- The full-page route's exit works again: the frame is same-origin, so its `keydown` is forwarded to the app's window (`forwardFrameKeys`) and the palette answers after the reader has clicked into the page.
- That route opens its own `/api/events` connection — it lives outside the provider that normally holds one, so it never saw `workflow_updated`/`workflow_removed`.
- The tab list takes a generation guard, so a slow answer for the Agent just left cannot land on the one now open.
- `WorkflowFrame` is keyed by tab, so a history fold, an error or an armed Remove cannot follow you to another workflow; an active tab that disappears is forgotten rather than filtered, so a re-created workflow does not snatch the view back.
- A workflow that contributes no tab but has an error or a hint now shows a notice row (`workflowNoticesOf`) — before this, a workflow whose first load failed was invisible everywhere in the app.
- The full-page fallback says "Opening the page…" instead of "Loading history…", stacks its two lines, and shows a list failure that arrives after the page is already up.
- `removeWorkflow`'s comment corrected: the server deletes the recorded versions too.

Also added the `PR:` line the changelog convention asks for, which this entry was missing.

**Not done, deliberately:** an `AbortSignal` on `WorkflowRequest` (the real fix for a hung relay — it changes the handler contract, so it wants its own change), and sharing one compiler program across loads.

hiyouga (approved):

Issue #621, opened 2026-09-05
Bug: Parallel Tool Call Causes 400 Error — Orphaned tool_result


paskaa:
## What happened

When a model emits **multiple parallel tool_calls in a single turn**, the second tool_result becomes orphaned, causing a `400 invalid_request_error: No function call found for function call output with call_id '...'`.

## Steps to reproduce

1. Ask the model to execute two commands in parallel:
   ```
   "请一次发出两个并行工具调用，分别执行 echo ALPHA123 和 echo BETA456，然后告诉我你看到的结果"
   ```
2. The model emits two `tool_call` messages in one turn (FC1 + FC2)
3. Both tool calls execute and return results
4. The **first** tool_result (OUT1) triggers `request_end` with `status=completed`
5. The **second** tool_result (OUT2) arrives after the request has closed
6. On the next request, OUT2 is sent as an orphan → **400 error**

## Version

v0.2.9

## How it is installed

CLI (penguin, from npm or an installer bundle)

## Operating system

Ubuntu 24.04 (Linux 6.8.0-136-generic)

## Does it reproduce on a fresh data root?

Not tested

## Relevant output

From trace `session-2026-09-05-12-31-07-f15228a8_001.jsonl`:

| Time | Event | call_id |
|------|-------|---------|
| 04:53:42.434 | `tool_call` FC1: `echo BETA456` | `call_01a06fea501d7323b33b735067c3b65f` |
| 04:53:42.436 | `approval_decision` allow | ← FC1 |
| 04:53:42.509 | `tool_call` FC2: `echo ALPHA123` | `call_01a06fea4f7c7601b73cdd32d73b65ae` |
| 04:53:42.510 | `tool_call_output` OUT1: `BETA456` | ← matches FC1 |
| 04:53:42.511 | `approval_decision` allow | ← FC2 |
| **04:53:42.525** | **`request_end` status=completed** ⚠️ | **Before OUT2!** |
| 04:53:42.581 | `tool_call_output` OUT2: `ALPHA123` | ← **orphan, FC2 already gone** |
| 04:53:42.582 | `request_begin` | |
| 04:53:43.560 | `request_end` status=fatal | `400 No function call found for call_id 'call_01a06fea4f7c7601b73cdd32d73b65ae'` |

## Root Cause

[...]

PR #720, opened 2026-09-14
fix(core,cli): AgentHub 0.4.15 so a parallel tool-call turn survives a gateway that interleaves the calls

hiyouga:
## What changed

`@prismshadow/agenthub` moves from 0.4.14 to 0.4.15 in `packages/core` and `packages/cli`, with the lockfile re-resolved and the `minimumReleaseAgeExclude` entry in `pnpm-workspace.yaml` renamed from 0.4.14 to 0.4.15 (the package is first-party and every bump lands inside pnpm's release-age gate).

0.4.15 carries agenthub [#218](https://github.com/Prism-Shadow/agenthub/pull/218): the four Responses clients (`openai_responses`, `gpt6`, `deepseek_v4`, `minimax_m3`) keep every function call of an interleaved parallel tool-call turn. A gateway that opens every call before closing any of them — Console Go streams that order — used to leave only the last call in the assistant message, so the next request replayed an orphaned `function_call_output`, the provider answered `400 … No function call found for function_call_output with call_id 'call_…'`, and the Task stopped. Every harness turn that ran two tools in parallel through such a gateway hit it.

No harness code changes: the fix lives entirely in the SDK.

## Verification

Run on the remote test host, not locally:

- `pnpm install --lockfile-only` re-resolved the lockfile with the new exclude entry; a second run after dropping the 0.4.14 entry left the lockfile byte-identical.
- `pnpm install --frozen-lockfile`, build of `@prismshadow/penguin-cli...` (core, server, cli), `pnpm typecheck` — clean.
- The full `@prismshadow/penguin-core` suite, the package that consumes the SDK.
- The installed `@prismshadow/agenthub` under `packages/core/node_modules` reads 0.4.15.

## Changelog

`changelog/unreleased/2026-09-14-agenthub-0.4.15.md` with its `.zh.md`.


hiyouga:
Reviewed at the PR head. The headline problem is not in the diff — it is what the diff does *not* do.

### 1. The harness never reads 0.4.15's `item_id`, so it is now the last single-slot attributor

`grep -rn item_id packages/ --include=*.ts` returns nothing, and `PartialToolCallPayload` (`omnimessage/types.ts:268`) has no field to carry it. Meanwhile `packages/core/src/llm/generative-model.ts:419` still does:

```ts
const providerKey = item.tool_call_id || this.activeToolCallId;
```

The four Responses clients still emit argument deltas with `tool_call_id: ""`; after this bump those deltas are distinguished *only* by the new `item_id`. So the release fixes the SDK and leaves its first-party consumer running exactly the heuristic the SDK just abandoned.

On a stream that opens both calls before streaming either — added(A), added(B), deltas(A), deltas(B), which is precisely the shape 0.4.15's map exists to support — every A fragment is appended to accumulator B. The Web card for B paints `{"path":"/etc/h{"cmd":"ls"}` while A's card stays empty, and the CLI prints the same splice permanently (it never re-renders from the complete message). Execution survives only because `emitCompleteTool` prefers `acc.completeArgs` over `argsBuffer` — so the tool runs correctly while the user watches a lie.

The agenthub fixture's Console Go order (added A, deltas A, added B, deltas B) happens to attribute correctly, which is why nothing caught it.

The right repair is upstream rather than here: have AgentHub stamp the resolved `tool_call_id` onto the delta fragment before yielding it. Routing on a Responses output-item id is provider knowledge the harness is not supposed to hold (`specs/05-ARCHITECTURE.md:54`), and it keeps the rule that a streamed fragment set is the call — nothing downstream has to reconcile deltas against a completed call.

### 2. Once the last-announced call closes, a still-open sibling's fragments are silently discarded

`generative-model.ts:569-570` — `emitCompleteTool` nulls `activeToolCallId` when it matches the closing accumulator, and `:420` then breaks on a falsy `providerKey`. Newly reachable *because* 0.4.15 closes each call at its own `function_call_arguments.done` while others are still streaming: added(A), added(B), deltas…, done(B) → `activeToolCallId` nulled → A's remaining id-less fragments hit the `break` and vanish, with no counter and no log. A's live preview freezes mid-argument; if the stream is cut before A's complete item, `finishInterrupted` backfills A from the truncated `argsBuffer` and writes *that* into the Trace. Partials are not persisted, so the discrepancy leaves nothing to diagnose from.

### 3. A call that closes with no accumulated argument text prints no CLI line at all

`packages/cli/src/render.ts:1060` — `renderPartialToolCall(..., { final: true })` returns null when `argsJson` is empty (`tool-render.ts:239`), and `renderWithheldCallLine` then writes nothing. Two ways to reach it after this bump: a call whose fragments were all mis-attributed to a sibling (finding 1), and any call a gateway announces with zero argument deltas — 0.4.15 now emits a valid complete `tool_call` for it (`parseToolCallArguments("")` returns `{}`) instead of dropping it.

Nothing rescues the line: `render.ts:739-745` deliberately never renders the complete `tool_call` (name map only), and in an auto-approved run `allow-all` returns in `approvals.ts:116` without publishing an `approval_request`, so the `ensuredCallLines` path never fires. The user sees `exec_command -> <output>` for a command that was never shown.

### 4. The published contract still promises what this bump makes false

`packages/docs/content/message-flow.en.md:75` (and `.zh.md:51-54`, and `design/specs/05-ARCHITECTURE.md:56`) guarantee that every segment goes strictly `start → delta* → stop` with the complete message right after, and that concatenated deltas equal the complete message; the annotated turn shows tool_call A's segment closing before B's opens, and the "not guaranteed" table lists interleaving only for tool *output* fragments. After 0.4.15 both premises can fail, and finding 1 makes the concatenation guarantee false on a real stream.

Per `CLAUDE.md` — 任何影响架构、行为或接口的代码改动都要同步更新对应的 specs;设计先落地,代码再跟上 — this needs the design change first. Until then an SDK consumer that keeps a single open-fragment slot (the CLI is one) is working as documented, and nobody is obliged to fix it.

### 5. The entry claims MiniMax is fixed; the MiniMax fix is not in this release

`minimax_m3` is listed among the four clients this release repairs, but 0.4.15's MiniMax client still builds tool calls from argument deltas — the shape that produced empty arguments for every MiniMax tool call. The actual repair is agenthub #221, which deletes the loop 0.4.15 adds to `minimax_m3` and reads the call off `output_item.done` instead; it landed on `dev` after the 0.4.15 fold and is deliberately unreleased.

So a user on the catalog's `minimax-m3` row (`model-catalog.ts:1613`) reads a release note naming MiniMax in a parallel-tool-call fix, points a session at it, and still gets tool calls with no arguments — and this PR's MiniMax delta has a lifetime of exactly one release. Either qualify the claim or wait for the version carrying #221.

### 6. The entry also overstates `item_id`

"AgentHub now carries an `item_id` on each `partial_tool_call` fragment" — it is optional, and only four of agenthub's ten clients set it. `openai_chat` (and its vLLM subclass), `kimi_k3`, `glm5_3` and the Gemini and Anthropic clients still emit fragments with `item_id === undefined` and still run their own single-slot accumulators. Anyone who reads that line and keys a fragment map on `item.item_id` gets `undefined` for every TokenDance / OpenRouter / Fireworks / SiliconFlow / Qwen row and collapses that turn's parallel calls into one bucket. AgentHub's own entry keeps both qualifiers ("gain an **optional** `item_id`: the Responses clients fill it").

### 7. Upstream, in agenthub — three things this bump inherits

Reporting these here because they decide whether the bump achieves its purpose; they belong in an agenthub PR.

- `openai_responses/client.ts:485` — `lastOpened` is assigned only in the start branch and never cleared when its call closes. For a gateway that sends no item ids, `done(A)` falls through to `lastOpened = "call_B"` and emits **B**; `done(B)` finds the key already deleted and `continue`s; `response.completed` then emits A. Three parallel calls reach the harness as C, A, B — against the contract that outputs are re-fed in call order. Identical code in `gpt6:505`, `deepseek_v4:431`, `minimax_m3:353` and all four Python twins.
- `:505` — the map is keyed from `item.id` at announce but probed with the fragment's `item_id`, and `keyOf` never falls back to the fragment's own `tool_call_id`. `ResponseFunctionToolCall.id` is *optional* in the OpenAI types while `item_id` is required on the delta and done events, so a gateway that omits it misses every lookup and reverts to pre-0.4.15 single-slot behaviour — ending worse than empty arguments: `done(A)` resolves to B, B closes with two concatenated JSON objects, `parseToolCallArguments` throws, A is never emitted, and the harness classifies the turn retryable and burns the reconnect ladder on a deterministic failure.
- `:541` — `parseToolCallArguments` is called inside the closing loop with no try/catch and *before* the terminal event is yielded. On a `response.completed` closing N leftover calls, the first malformed one aborts the rest and the terminal event never arrives, so the request produces no `token_usage` — a request the provider did bill leaves no cost record. The throw site predates 0.4.15, but its blast radius was one call and is now N.

### 8. Test coverage

Two of the four hand-copied client fixes ship with no test in either language: `RESPONSES_CASES` filters to `protocol === "responses"`, which selects only `OpenaiResponsesClient` and `DeepSeekV4Client`, and the near-byte-identical copies in `gpt6` and `minimax_m3` are executed by nothing — while both are live catalog rows. The new interleave test also asserts only the complete `toolCalls`, never the fragment stream the harness renders from, which is the assertion that would have caught finding 1.

On this side, `packages/core/test/llm.test.ts:460` ("attributes empty-id tool-call argument deltas to the active tool call") opens exactly one call, and the parallel test at `:2257` uses distinct explicit ids delivered as complete items. The intersection — two calls open with id-less argument deltas, the shape 0.4.15 exists to produce — is never constructed, so findings 1 and 2 are green in CI and a future reader will read that test name as an intentional guarantee. It needs no mock server: `translateEvents` with start(A), delta(id:""), start(B), delta(id:""), tool_call(A), tool_call(B), asserting each card's concatenated deltas equal its complete arguments.

### 9. Two changelog details

- `2026-09-14-agenthub-0.4.15.zh.md:12` heads its section `## 详情` where the convention is `## 细节` (211 uses against 8 strays; the sibling 0.4.14 entry two days earlier gets it right). Since the README sells `grep` as the deliberate substitute for an index, a release-prep sweep for `^## 细节` skips this entry's only body section.
- No `Issue:` line, although issue #621 ("Parallel Tool Call Causes 400 Error — Orphaned tool_result") is the report this bump answers: its Trace shows FC1 `echo BETA456` logged before FC2 `echo ALPHA123` with FC2's call_id orphaned into `400 No function call found` — the reversed-order-plus-orphan signature of the single-slot accumulator. It is currently closed as NOT_PLANNED on an incorrect harness-side root cause, so it is also worth reopening or re-closing with a pointer here.

### 10. One CLI nit

`packages/cli/src/render.ts:1336` — `finishLine()` unconditionally clears the single `partialToolCallLineId`, so one call's stop disowns a sibling's still-growing preview and the next fragment reprints the whole preview on a fresh line, leaving the truncated one above it. Needs a delta for B after A's stop, which was impossible before this bump and is possible now that calls close individually. Cosmetic, CLI only.

PR #721, opened 2026-09-14
fix(web): the send-to-background action waits ten seconds, and the mark shares its slot

hiyouga:
## What changed

- The **Send to background** action on a running `exec_command` / `run_subagent` row now appears only once the call has been executing for ten seconds (`BACKGROUND_ACTION_DELAY_MS`, measured from the execution segment's start — approval grant, else call completion). A command that returns in a few seconds used to flash the action and take it away again, so the row's right end jumped with every short call. A small `useElapsedPast(sinceMs, thresholdMs)` hook in `live-duration.tsx` schedules exactly one re-render at the crossing; nothing ticks.
- The `[Background]` mark moves out of the row button into the same right-end slot the action occupies (before the expand chevron), so it is right-aligned like the action and a click leaves the mark in its place. The two never show together.

`showsBackgroundAction` gains an `executedMs` parameter so the rule stays a pure, tested predicate.

## Round 2 — the note tells the model not to follow up

The note `exec_command` / `run_subagent` return when the user moves a running call to the background used to end with "Use input_command to interact", and models answered the move by opening `input_command` on the very process the user had just parked. It now says: leave it running unattended, do not poll it or send it input, its completion arrives on its own as a `[background_task_done]` user message, move on to other work or end the turn if nothing else is pending (`kill: true` / `abort: true` still stop it if the user asks). The render layer keeps matching on the shared `DETACHED_TOOL_NOTE_PREFIX`, so the card's mark is unaffected. Docs (tools en/zh) and design #160 say the same.

Design: penguin-harness-design #160 (stacked on #146, the send-to-background design that is still open).

## Verification

Run on the remote test host, not locally: build of core/server/web/cli, `pnpm typecheck` (every package), `pnpm format:check`, the full web suite (141 files, 1921 tests — the predicate's threshold cases added to `tool-call-preview.test.ts`, the detach fixture updated to the new note), and core's background-execution, exec-session and subagent suites (107 tests) for the note change.

## Changelog

`changelog/unreleased/2026-09-14-background-action-delay.md` with its `.zh.md`.


hiyouga:
Reviewed at `5470c3d72`. The timer lifecycle is sound — unmount clears the timeout, a call completing or aborting during the wait flips `executing` so `sinceMs` goes undefined and the effect clears both `reached` and the timer, and a card reused for a different call by a resync rebuild is harmless because the return expression reads the *new* `sinceMs`. The mark/action mutual exclusion holds, and the nested `[background_task_done]` inside the note's brackets is safe (`parseBackgroundTaskDoneMessage` requires `m.index === 0` on `user_text` items).

One real bug, and one simplification that removes a whole class of fragility.

### 1. The gate restarts when the approval timestamp lands, re-introducing exactly the flash this PR removes

`packages/web/src/features/chat/tool-call-card.tsx:326-329`

`sinceMs` is `executing ? (item.approvalAtMs ?? item.callStartedAtMs) : undefined`, and there is a real window where `executing` is already `true` while `approvalAtMs` is still `undefined` — so the gate first measures from `callStartedAtMs`, which predates the entire human approval wait.

Always-ask policy, `exec_command`:

1. `tool_call` completes → `callStartedAtMs = T`.
2. `approval_request` → `pending` → `executing === false`.
3. The user reads the command and clicks Allow at `T+15s`. `chat-page.tsx` awaits `api.postApproval(...)` then calls `stream.resolveApproval(key)`, which deletes the pending entry optimistically (`stream-controller.ts:644-646`) and fires `onPendingChange` → render.
4. At that render `pending` is gone but the `approval_decision` OmniMessage — the only thing that sets `approvalAtMs` (`stream-model.ts:1583`) — has not arrived; the POST response normally beats the broadcast. So `executing === true`, `sinceMs === callStartedAtMs`, elapsed ≈ 15 s ≥ 10 s → **the action renders immediately on click**.
5. The stream event lands, `approvalAtMs ≈ now`, `sinceMs` changes, the effect re-runs with `remaining ≈ 10_000` → `setReached(false)` → **the action disappears** and comes back ten seconds later.

So on every manually approved detachable call where the human took ≥10 s to decide, the row's right end flashes the action and takes it away — the precise symptom the changelog says this fixes. `main` did not do this (the action simply appeared on approval and stayed).

**On the fix:** the tempting one — stamp `card.approvalAtMs` optimistically in `resolveApproval` — has a trap. `approvalAtMs` also feeds `noteApprovalWait`, which does `approval - call` where `callStartedAtMs` is a *server* message timestamp; writing a local `Date.now()` into it mixes two clocks inside a stat, and a skewed client would corrupt `openApprovalWaitMs`. Better to leave `approvalAtMs` server-only and give the card the local click time under its own name — `resolveApproval` already has `findToolCard` imported, and the gate then reads `item.approvedLocallyAtMs ?? item.approvalAtMs ?? item.callStartedAtMs`. Local-first means `sinceMs` never moves again for that card, so there is nothing left to re-run the effect. That local value is also the *right* clock for this gate, which compares against `Date.now()`.

(The same `approvalAtMs ?? callStartedAtMs` gap makes `LiveDuration` at `:437` jump from ~15 s back to ~0 s in that window, but that line is unchanged from `main` — pre-existing, and the same fix would cover it.)

### 2. The hook returns a number its only caller immediately re-thresholds

`packages/web/src/features/chat/live-duration.tsx:19-37`

`useElapsedPast` computes the boolean it needs (`reached`), throws it away, and returns `Math.max(0, Date.now() - sinceMs)` so `showsBackgroundAction` can compare against the threshold again. There is exactly one caller and it uses the number for nothing else. The round-trip costs three things:

- a second, independent `Date.now()` read during render, so the render is impure;
- the `Math.max(0, …)` clamp, behaviourally dead for this caller — a negative elapsed and a clamped `0` both fail `>= 10_000` identically;
- a structural fragility: the crossing render re-derives the verdict from a fresh clock read, and if that read comes out even 1 ms short the action is withheld with **no further re-render scheduled** — the effect's deps `[sinceMs, thresholdMs]` have not changed. A silent command (`sleep 300`) produces no other re-render either. I could not construct this from timer behaviour alone (`setTimeout` never fires early), so it needs a backwards wall-clock step — but the whole class disappears with the fix.

Return the boolean: `useElapsedPast(sinceMs, thresholdMs): boolean` → `return reached && sinceMs !== undefined`, and `showsBackgroundAction(..., executedLongEnough: boolean)`. That deletes the clamp, the second clock read, and the four docstring lines explaining why a coarse number is acceptable. Keep the `sinceMs !== undefined` term — it is load-bearing on the render where `sinceMs` goes undefined and `reached` is still stale-`true`.

### 3. Changelog `Scope` omits `core` and `docs`

`changelog/unreleased/2026-09-14-background-action-delay.md:5` and `.zh.md:5` say `- **Scope:** \`web\``, but the diff changes `packages/core/src/environment/tools/{exec-command,run-subagent}.ts` and `packages/docs/content/tools.{en,zh}.md` — and the entry's own third bullet describes the core change. `grep -rl 'Scope:.*core' changelog/` will miss this model-facing prompt change.

### 4. The new note forbids follow-up unconditionally, but only excuses `kill`

`exec-command.ts:165-168`, `run-subagent.ts:234-238`:

> …leave it running unattended: do not poll it or send it input — …; `input_command` with `kill: true` still stops it if the user asks

"do not poll it or send it input" carries no escape, while "if the user asks" is attached only to `kill`. A user who then types "what's that build printing now?" puts the model between a direct request and an absolute instruction. Since the note exists to stop *unprompted* follow-up, scoping it — "do not poll it or send it input on your own initiative" — keeps the intent without the conflict.

(The sibling `run_in_background` launch notes at `exec-command.ts:132-133` and `run-subagent.ts:189-190` keep the old "no need to poll. Use input_command to interact" wording. I read that divergence as deliberate — the model chose that background, the user imposed this one — and am only noting it.)

### 5. Two smaller things

- `tool-call-card.tsx:466-470` — moving the `[background]` mark out of the row `<button>` makes it a bare `<span>` between two buttons: clicking it no longer expands the row, and it no longer contributes to the row button's accessible name. A behaviour change the entry does not mention.
- `live-duration.tsx:1-8` — the file docstring is entirely about `LiveDuration`'s per-second tick, and the new hook is explicitly *not* a ticking clock. Extend the file doc or move `useElapsedPast` below `LiveDuration`. Also `tool-call-card.tsx:195` writes `BACKGROUND_ACTION_DELAY_MS` bare in prose where the surrounding docstrings backtick identifiers.

---

Batch note across #721/#723/#724: all three changelog entries use `## What changed` / `## 变更内容` with no lead paragraph. `changelog/README.md` prescribes H1 → metadata → cross-link → lead paragraph → section and says to reuse an existing name first; `## Details` / `## 细节` appear 221/211 times in the corpus, `## What changed` zero times outside frozen `0.2.3/`. If the new shape is deliberate, the README should say so.

PR #722, opened 2026-09-14
test(core): the prompt cache — assembly invariants, an Anthropic-rule simulator, and the Session lifecycle

hiyouga:
## Why

PR #700 showed a request parameter (the thinking level) drifting between the turn requests and the compaction request of one Session, which breaks the provider's prompt cache exactly when the context is largest. The question behind it: is the harness's request assembly prefix-cache aware at all, and can a test tell? This PR adds that test, with a client-side diagnostic modelled on Anthropic's cache-diagnostics beta.

## Layout

`test/helpers/prompt-cache/` holds four small modules in reading order — `recording.ts` (capture what would go on the wire), `diagnostics.ts` (explain a divergence between two requests), `simulator.ts` (what a provider with Anthropic's rules would read), `fixtures.ts` (what the suites share) — and three suites: `prompt-cache-invariants.test.ts` (the harness's assembly), `prompt-cache-simulator.test.ts` (the simulator itself), `prompt-cache-lifecycle.test.ts` (the whole Session lifecycle through the simulator). Everything is called the prompt cache; "prefix" is used only for the bytes before the breakpoint.

## What the suite does

`packages/core/test/helpers/prompt-cache/recording.ts` runs a real `GenerativeModel` whose AgentHub provider client has its stream replaced by a scripted `UniEvent` stream shaped like `Claude5Client`'s (thinking with signatures, partial then complete tool calls, usage snapshots). AgentHub's own stateful history bookkeeping runs for real, and every request is recorded on the WIRE form the Claude client sends (`transformUniConfigToModelConfig` / `transformUniMessageToModelInput`). `diagnostics.ts`'s `diagnoseCacheMiss(prev, next)` reports the earliest divergence in Anthropic's order — model, system, tools, the other prompt-affecting parameters (thinking / effort / tool_choice / speed / betas / cache_control; `max_tokens` is ignored, it is not part of the cache key), then the first diverging message with an estimate of the input lost after it — the same vocabulary as the API's `cache_miss_reason`.

`packages/core/test/prompt-cache-invariants.test.ts` drives a `ContextEngine` through the cases that matter:

- every request of a Session extends the previous one on the wire (tool round, final reply, a second Task);
- steering delivered mid-task rides the same user turn as the tool result;
- moving the thinking level invalidates the message cache exactly once and then holds — the compaction request included, and the context the compaction opens receives the pin as its per-request override;
- a reconnect changes nothing before the retried turn's input, and that input only grows (see the finding);
- the compaction request extends the turn it follows (same object, same config, one appended user turn);
- a resumed Session replays the live history byte for byte, thinking signatures and tool blocks included.

## Findings

- **The harness is cache-aware where it counts.** Within a context the system prompt and the toolset are frozen (assembled once at context open, after MCP servers connected, and recorded in `session_meta`; a resume of an open context reuses the recorded bytes), the history is append-only through AgentHub's stateful client, the thinking level is the one soft parameter and is now read live for every request (#700), and `max_tokens` — the only per-request value that varies — is not part of the cache key. The compaction request runs on the same object with the same config. Nothing in the assembly re-orders tools, interpolates timestamps per request, or rewrites earlier messages.
- **One place the prefix legitimately changes: the reconnect ladder.** After a partially streamed attempt fails `retryable`, the retry re-sends the turn with a `[turn_retried]` user text carrying what the attempt already produced, so the retried turn's input grows instead of repeating. Everything before it is byte-identical, so tools, system and history still hit; what is re-read is the retried turn's own input — the same tail a normal turn re-reads anyway, because a cache entry is written at each request's end and the next request's new blocks are never cached yet. The suite pins the invariant in that form (the divergence never moves earlier; the original input still leads the retried one) rather than demanding an identical resend, which would cost the model the partial output.
- **The provider side, for the record.** AgentHub's Claude client enables caching by default with one automatic breakpoint (`cache_control: ephemeral` at the top level, 5-minute TTL); `PromptCaching.ENHANCE` (1-hour TTL) exists but the harness never asks for it. A thinking-level change maps to `output_config.effort`, which per Anthropic's rules invalidates the message cache — unavoidable when the user moves it; the harness's job is to move once and hold, which it now does.

## Round 2 — a fake provider with the real hit rules, and the session lifecycle under it

`packages/core/test/helpers/prompt-cache/simulator.ts` is a prompt cache that behaves like Anthropic's: the prefix is tools → system → a virtual parameters block (thinking / effort / tool_choice / speed / betas) → message blocks; a completed request writes **one entry per breakpoint** (the hash of the prefix ending there — content behind a breakpoint is not addressable on its own); a read hashes the prefix at the breakpoint and walks back at most 20 positions (the breakpoint counting as the first; runs of `tool_use` / `tool_result` blocks are one position each) for an entry a prior request wrote; entries expire five minutes after their last use, with an injectable clock; nothing below 1,024 tokens is cached; usage comes back as `cache_read_input_tokens` / `cache_creation_input_tokens` / `input_tokens`. Its breakpoint policy is an option: `"automatic"` (what AgentHub sends today — one top-level `cache_control`) or `"tools-system-automatic"` (explicit breakpoints on the last tool block and the system block as well; Anthropic allows four). Eleven unit tests in `prompt-cache-simulator.test.ts` pin the rules.

`packages/core/test/prompt-cache-lifecycle.test.ts` drives a real `Session` (real `Environment` with shell commands, a fake subagent runner, the Trace writer) with the simulator as the provider and asserts, request by request, that each read covers the previous request's whole prefix unless the step is a named, bounded exception. Measured under today's single breakpoint:

| scenario | hit |
| --- | --- |
| ordinary conversation, two Tasks with a tool round | 95–98% |
| interruption (abort mid-stream, then continue) | everything before the rewritten last user message; the interrupted turn's own input is re-read |
| `run_in_background` command + its `[background_task_done]` notice | 93% |
| a call the user moved to the background | 93% |
| three parallel tool calls in one turn | 91% (the round costs 2 lookback positions, not 6) |
| subagent (parent line) | 96%; the child's first request reads 0 — see below |
| scheduled task (`[scheduled_task]` block) | 95% |
| resume from Trace within the TTL | 98%, the whole replayed history from cache |
| resume past the TTL | read 0, `explainMiss` = no divergence (an expiry, not a bug) |
| the compaction request itself | 98% |
| thinking-level move | read 0 on the next request, then 98% |
| compaction reopen | read 0 on the new context's first request, then 98% |

**The sharpened finding.** With AgentHub's single automatic breakpoint, no entry ever ends at the tools or system boundary, so every prefix change that lands early — a thinking-level move (`output_config.effort`), a compaction reopen (same tools, same system prompt, new history), or a subagent child's first request (byte-identical tools and system prompt) — loses the whole prefix, tools and system included, not just the messages. The twelfth test reruns the thinking move and the compaction reopen with breakpoints after the tools and the system prompt and reads tools+system back in both (1,647 of ~1,700 tokens; 96–97% hit). That turns recommendation 4 below into a number.

## Recommendations (not in this PR)

0. **AgentHub: set explicit `cache_control` breakpoints on the last tool block and on the system block** next to the automatic one. It is what keeps tools and system cached across a thinking-level move, a compaction reopen and every subagent's first request; the suite measures the gain.

1. Offer the 1-hour cache TTL (`prompt_caching: enhance`) as a model or Agent option: agent Sessions routinely idle longer than five minutes between tool approvals and turns, and every such gap forfeits the whole prefix.
2. Record a request fingerprint (system hash, tools hash, thinking level, history length) on `request_begin`, so the Trace view can label a drop in the hit rate it already shows with a reason, the way `diagnoseCacheMiss` does here.
3. When the client is the Claude API directly, let AgentHub pass `diagnostics.previous_message_id` (beta `cache-diagnosis-2026-04-07`) and surface `cache_miss_reason` in `token_usage`.

## Verification

Run on the remote test host, not locally: `pnpm --filter @prismshadow/penguin-core typecheck`, the three suites (`prompt-cache-invariants` 6, `prompt-cache-simulator` 11, `prompt-cache-lifecycle` 12 — 29 tests), `pnpm format:check`.

## Changelog

`changelog/unreleased/2026-09-14-prompt-cache-awareness.md` with its `.zh.md`.


hiyouga:
Reviewed at `3cf5cb50c`. The suite's core claims are right and worth pinning: writes happen only at a breakpoint, reads land only on positions a previous request closed, and the three "designed miss" conclusions that follow — subagent child pays full price, compaction reopen pays full price, thinking move pays full price, all despite byte-identical tools and system — are genuinely correct. The harness-side assertions discriminate: I checked the interruption, reconnect, resume, steering, background-notice, detach, parallel-tool and scheduled-task cases, and each compares hash-derived reads or byte-wise wire slices, so a history rewrite upstream fails them. `expectHits`'s minimum guard is the right guard (the system prompt alone is ~1421 tokens against the 1024 floor). No float equality, no unawaited generators, no order dependence, `cleanups` drained in `afterEach`.

The problem is the simulator's rule set. Four rules are wrong, and two of them will actively mislead a later change.

### 1. The simulated cache is not model-scoped

`packages/core/test/helpers/prompt-cache-sim.ts:63`, `:159-179` — `prefixBlocks` hashes tools → system → `parameterView` → message blocks, and `model` is in none of them (`PARAMETER_KEYS = ["thinking","output_config","tool_choice","speed","betas"]`). Two requests differing only in `model` hash identically, so the simulator reports a **full hit across a model switch**. In reality caches are model-scoped, and a model switch is the one invalidator with no escape hatch — it kills tools, system and messages alike.

This is reachable: `openNextContext` re-stamps `model_id` per context (`session.ts:65`, `:96`), so a compaction rotation onto a different model is a supported path, and `driveCompaction` is exactly what the suite models. The sibling helper already knows this — `prefix-cache.ts:286` returns `model_changed` — so the two helpers silently disagree.

Fix: push a leading block in `prefixBlocks` — `makeBlock("model", \`model:${json(wireConfig.model)}\`)` — *ahead of* the tool blocks, so a model change invalidates the tools tier too.

### 2. Multi-breakpoint reads use one lookback window instead of one per breakpoint

`prompt-cache-sim.ts:282`, `:326-334`, `:340-355` — `breakpointOffsets` can write three entries under `breakpoints: "tools-system-automatic"`, but `lookup` is called once, anchored at the last position, walking back at most 20 from there. A conversation with ≥20 message positions can therefore never read the tools/system entries, while a real explicit breakpoint on the system block does its own lookback from its own position and hits at distance 0.

It passes today only because `driveThinkingMove` and `driveCompaction` produce ~7 positions. Extend either scenario to a realistic length and the headline test at `prompt-cache-hits.test.ts:752` flips to a 0-token read — which will read as a harness regression and will not be one.

Fix: run `lookup` once per entry in `breakpointOffsets`, each window anchored at that offset's position index, and take the longest live match.

### 3. The TTL test does not test the thing it names

`prompt-cache-sim.test.ts:227-242` — "…and a read refreshes it". Delete the read-refresh line (`prompt-cache-sim.ts:351`) and it still passes: every read goes through `sim.request()`, whose `complete()` immediately rewrites the same hash with `now + ttlMs`, so the write refresh masks the read refresh. After step 2 the expiry is `1e6 + 2·TTL − 1` either way, and step 3 finds it expired either way.

Fix: make the middle request an *extension* so `complete` writes a different hash, then re-issue the original short prefix just before `2·TTL` and assert it still reads.

### 4. `speed` is in the wrong prefix slot, and the thinking rule is stated as settled

`prompt-cache-sim.ts:10-18`, `:63` — the invalidation hierarchy puts a `speed` change **between tools and system**: the tools cache survives, the system cache does not. The sim lumps `speed`/`betas` with `tool_choice` in one virtual block *after* the system block, so a fast-mode toggle looks free for the system prompt when it is not. (`speed` + `betas` are exactly what AgentHub sets for `fast_mode`.)

Separately, the header asserts as fact that a thinking/effort change "invalidates the messages but neither tools nor system". The documentation calls that row **model-specific** — on models that render the thinking configuration ahead of tools and system it invalidates those tiers too. The AgentHub recommendation the changelog draws from this (the 96.3% / 96.8% figures) rests on the favourable branch of a rule the docs do not settle for `claude-sonnet-4-6`.

Fix: split the virtual block — `speed`/`betas` before the system block, `tool_choice` after — and say in the header that the thinking placement is the optimistic, model-dependent case.

### 5. The two helpers contradict each other about `cache_control`

`prefix-cache.ts:30`, `:76-77` keeps `cache_control` in the prefix and names it prompt-affecting in the `CacheMissReason` comment; `prompt-cache-sim.ts:57-63` excludes it, and `prompt-cache-sim.test.ts:182-191` pins the exclusion. The simulator is right — a moving marker is never an invalidator. Inert today because AgentHub emits a constant `{type:"ephemeral"}`, but the moment `prompt_caching: ENHANCE` is plumbed through, `diagnoseCacheMiss` reports a false `parameters_changed [cache_control]`. Add it to `NON_PREFIX_CONFIG_KEYS` and drop it from the comment. Same rule, third list: `prefix-cache.test.ts:126-130`'s `configFingerprint` strips only `max_tokens`, not `stream`.

### 6. Two of the eleven scenario tests assert only the simulator's own arithmetic

- `prompt-cache-hits.test.ts:690-708` re-drives the whole Trace-write/resume scenario — real `mkdtemp`, a `Writer`, two `Session`s — to restate TTL arithmetic already pinned at `prompt-cache-sim.test.ts:227`. Its one harness claim (`explainMiss(order[1], order[2]) === "no divergence"`) is asserted verbatim 20 lines above.
- `:752-780` re-drives `driveThinkingMove` **and** `driveCompaction` to measure a breakpoint policy this harness does not send. That is a proposal for AgentHub, not an invariant of this repo; when AgentHub changes it fails here for reasons unrelated to `packages/core`.

Drop the TTL duplicate; if the AgentHub measurement is worth keeping, express it on hand-built requests in `prompt-cache-sim.test.ts` rather than paying for two full scenario drives.

### 7. Scaffolding nobody uses

`PromptCacheSimOptions.lookbackPositions` and `.ttlMs` (`:109-111`) are never passed. `DEFAULT_LOOKBACK_POSITIONS`, `prefixBlocks`, `positionEnds`, `CacheBlock`, `CacheBlockKind`, `BreakpointPolicy`, `PromptCacheSimOptions` are exported but used only inside the helper. `ScriptedReply.cachedTokens` (`prefix-cache.ts:43`) is never set by any test yet threaded into both usage events. The `begin`/`complete` split — the thing that models "a write is invisible until the request completes", and the reason the subagent case is interesting — is exercised by exactly one unit test (`prompt-cache-sim.test.ts:272`); every scenario goes through `replay()`'s sequential `sim.request()`.

### 8. `expectHits`'s `allowed` map never reads its values

`prompt-cache-hits.test.ts:353-383` — `allowed.get` is never called (only `.keys()` and `.has()`), so the reason strings, including the two hoisted `THINKING_MOVE_REASON` / `CONTEXT_REOPENED_REASON` constants, are documentation the code never touches. The only validation is that the index is in range, so a stale exemption for a request that now *does* hit silently weakens the suite. Assert that an allowed index actually fell short of its predecessor — that makes the reason load-bearing.

### 9. Smaller things

- `prompt-cache-sim.ts:32-35` / `prompt-cache-sim.test.ts:244-260` count the breakpoint as the first of the twenty, pinning "19 appended positions hit, 20 miss". The documented wording is "up to 20 positions *before* the breakpoint", which would make 20 still hit. Nothing is near the boundary, so the cost is only a precise-looking test pinning one reading of an approximate rule — worth a comment saying so.
- `prompt-cache-hits.test.ts:411-413` — `usages[0].cache_read_input_tokens === 0` is true by construction for the first request into a fresh sim; `wrote === true` restates `expectHits`'s guard; `hitRatio > 0.9` is dominated by the system prompt and is strictly weaker than the `expectHits` two lines above.
- 45 byte-identical fixture lines (plus near-identical `SYSTEM_PROMPT`/`TOOLS`/`modelConfig`/fake Environment) are duplicated between `prefix-cache.test.ts` and `prompt-cache-hits.test.ts`; they belong beside `recordingModel` in `helpers/prefix-cache.ts`.
- The changelog quotes 96.3% / 96.8% but no test pins either figure — a fixture edit rots them silently.
- `waitFor`'s default `timeoutMs = 8000` (`prompt-cache-hits.test.ts:297`) can never be reached: `packages/core/vitest.config.ts` sets `testTimeout: 5_000`, so the test dies first.

### Fidelity verdict

Structurally faithful where it matters — prefix hashing over the exact rendered bytes, `tools → system → messages` order, write-only-at-a-breakpoint, read-only-on-closed-positions, the consecutive `tool_use`/`tool_result` run collapse, TTL refresh on read, and `minCacheableTokens = 1024` / `ttlMs = 5 min` are the right values for `claude-sonnet-4-6` with a default `ephemeral` marker. `breakpoints: "automatic"` correctly mirrors what AgentHub's non-Bedrock Claude client actually sends.

Wrong or over-simplified: not model-scoped (1), one lookback window for many breakpoints (2), `speed`/`betas` placed after the system block (4), thinking placement presented as settled (4). Two lesser omissions, both harmless here: the automatic breakpoint's walk-back to the nearest *eligible* block is not modelled, and the sim is stricter than reality about write visibility (an entry becomes readable once the first response begins streaming, not at completion) — being stricter only costs a pessimistic number.

PR #723, opened 2026-09-14
fix(web): the context panel's pending value reads larger, the room past the cutter is hatched, segments pin on click

hiyouga:
## What changed

Three adjustments to the context-usage ring's composition panel (`features/chat/context-gauge.tsx`):

- **The pending value is readable while dragging.** The chip that floats under the compaction cutter during a drag (or arrow-key adjustment) used ten-pixel type; it now uses the panel's own text size (regular weight — the bold of the first push was dropped on review) with a little more padding.
- **The room past the cutter is hatched.** The stretch of the bar to the right of the cutter, up to the model window, is drawn with a diagonal hatch (`.context-hatch`, plain CSS with a `.dark` rule): it is room the model has but compaction fires before the Session can use it, so it no longer looks like the plain free run left of the cutter. It sits under the coloured fills, follows the pending value live while the cutter moves, and carries a tooltip saying what it is.
- **Segments pin on click.** Bar segments and the six legend rows get a pointer cursor and a click pins the highlight (segment and row together) until the same item is clicked again or another is picked; hover still wins while it lasts, and the tool / file rankings stay hover-only. The pin gives the pointer cursor something to mean.

Design: penguin-harness-design #161 (stacked on #92, the panel's spec).

## Verification

Run on the remote test host, not locally: build of core/server/web/cli, `pnpm --filter @prismshadow/penguin-web typecheck`, `pnpm format:check`, and the full web suite (141 files, 1920 tests).

## Changelog

`changelog/unreleased/2026-09-14-context-gauge-polish.md` with its `.zh.md`.


hiyouga:
Reviewed at `c8d313ba8`. The hatch geometry and the pin state machine are both correct — I traced them: `beyondFraction` uses the same `pending ?? threshold` as the cutter's `shown`, so the hatch's left edge lands exactly on the cutter's 1px line; `beyondFraction < 1` correctly suppresses the hatch when the threshold is at or past the window; `pinned` only ever holds one of the six `CONTEXT_PART_KEYS` and `contextComposition` always returns all six, so a pin always has a row that can clear it; and the panel unmounts on close, resetting it. The pending chip's move from `text-[10px]` to `text-xs` is a real improvement — the old size was below the control-size floor.

The findings are an interaction gap, a CSS-token issue, and some prose.

### 1. The legend rows became clickable with no keyboard path

`packages/web/src/features/chat/context-gauge.tsx:876-883` — the `<li>` gains `onClick` and `cursor-pointer` but no `role`, `tabIndex`, `onKeyDown` or `aria-pressed`. The bar segments at `:487` sit inside the `aria-hidden` track and are deliberately mouse-only, which makes the legend *the* accessible surface for pinning — and it has none.

Tab to the ring, Enter to open the panel, Tab again: focus jumps straight past all six rows to the Tools/Files switch. Pinning is unreachable without a mouse, and a screen reader is never told the rows are interactive.

There is no `role="button"` anywhere in `packages/web/src` — the house answer is a real `<button>`. When `onSelect` is given, render the row's content in `<button type="button" aria-pressed={pinned === key}>` filling the `<li>`; the hover-only ranking rows keep the plain `<li>`.

### 2. A pin looks exactly like a hover, so the toggle's state is unobservable while you are on it

`:376` (`const lit = hovered ?? pinned`) and `:882` — one `bg-gray-100 dark:bg-gray-800` serves both states. Hover a row: it tints. Click it (pinned): nothing changes. Click again (unpinned): still nothing changes. You can only find out whether the pin is on by moving the pointer away.

The pinned resting state needs a second carrier hover does not use — a left rule or a ring — which also gives `aria-pressed` something to agree with.

### 3. The hatch hardcodes Tailwind v3 greys on a v4 project

`packages/web/src/styles.css:19-33` — `rgb(156 163 175 / 0.55)` and `rgb(75 85 99 / 0.75)` are the v3 hexes for gray-400/gray-600. This repo is on Tailwind 4.1, where `gray-*` resolves from the oklch scale, so the hatch is near but not equal to the greys beside it and will not follow the `.dark` overrides at `:69-72`. The rest of this file does use tokens (`var(--color-brand-50)` at `:357`).

The two themes also land at different strengths: light is `#9ca3af` @ 55% over `bg-gray-200` → ≈1.33:1; dark is `#4b5563` @ 75% over `--color-gray-800` → ≈1.77:1. The hatch is the only carrier of "this room is unusable" — its `title` is inside `aria-hidden` and reaches nobody — so it wants 3:1. Spell both from `var(--color-gray-*)` and raise the light alpha until the two read the same.

Also: `.context-hatch` sits between `@custom-variant` and `@theme`, while every other component class in the file lives from line 238 on. And the comment's closing sentence records an attempt ("a `dark:` variant inside `@apply` does not compile for this kind of rule") rather than a constraint.

### 4. The new English docs sentence is ungrammatical

`packages/docs/content/web-app.en.md:83` — "room the model has **that** compaction fires before the Session can use". The relative clause has no object. The code comments got it right (`context-gauge.tsx:421-422`, `styles.css:15-16`): "room the model has **but** compaction fires before the Session can use". The changelog at `:13` adds a resumptive "use **it**", also wrong. Use the code comment's wording in all three.

### 5. Smaller things

- `context-gauge.tsx:44-45` — the module header still says hover-only ("Hovering a segment or its legend row links the two") and never mentions the hatch. In a file where the header is the module's spec, both new behaviours need a clause.
- `:879` — `{...(onSelect !== undefined ? { onClick: onSelect } : {})}` is just `onClick={onSelect}`; React ignores a nullish listener and the two lines above already attach `onMouseEnter`/`onMouseLeave` unconditionally. Same for the paired `onSelect !== undefined ? "cursor-pointer " : ""` with its trailing space.
- `:419` — `hoveredPart` now carries a pinned key too. `litPart`.
- `strings.ts:2019-2023` — `contextBeyondThreshold` was inserted between `contextThresholdCutter` and the second half of its JSDoc ("and its tooltip naming the threshold it stands on"), which documents `contextThresholdHover`. Move it above the comment or below that key.
- `web-app.zh.md:82` ends in `。` where every sibling bullet ends in `；`; and the English inserted the new material *before* "Confirming writes it to the Agent…" while the Chinese appended it after, breaking the drag→release→confirm narrative. The Chinese also documents the chip's font size where the English does not.

---

Batch note across #721/#723/#724: all three changelog entries use `## What changed` / `## 变更内容` with no lead paragraph. `changelog/README.md` prescribes H1 → metadata → cross-link → lead paragraph → section and says to reuse an existing name first; `## Details` / `## 细节` appear 221/211 times in the corpus, `## What changed` zero times outside frozen `0.2.3/`. If the new shape is deliberate, the README should say so.

PR #724, opened 2026-09-14
fix(web): signing out asks first

hiyouga:
## What changed

**Sign out** in the Web App's user menu now opens the shared `ConfirmModal` before ending the session. The row sits in a menu of harmless entries (appearance, settings, update), and one slip used to land the user on the login page. The dialog names what happens — the session here ends and returns to the login page; running conversations carry on on the server — and Cancel keeps everything as it was. The dialog is mounted beside the settings dialog, outside the dropdown, so it outlives the menu that opened it. Hidden in desktop mode as before (the row itself is).

Design: penguin-harness-design #162.

## Verification

Run on the remote test host, not locally: build of core/server/web/cli, `pnpm --filter @prismshadow/penguin-web typecheck`, `pnpm format:check`, and the full web suite.

## Changelog

`changelog/unreleased/2026-09-14-logout-confirm.md` with its `.zh.md`.


hiyouga:
Reviewed at `ae14bfd5e`. The implementation itself is clean — it reuses `ConfirmModal` rather than inventing a second confirm, mounts it as a sibling of `Dropdown` so the menu closing cannot strand the dialog, opens focus on Cancel rather than the destructive button, and the failure path recovers because `auth.tsx` clears the user in a `finally` and `RequireAuth` redirects on null. Esc and overlay-click both cancel. Both dictionaries stay in lockstep, and `Strings = typeof zh` makes drift a type error anyway.

One thing actually breaks, plus a few notes.

### 1. The e2e sign-out step is now stale and will hang

`packages/web/e2e/draft-user-scope.spec.mjs:64`

```js
await page.getByRole("button", { name: UA }).last().click();
await page.getByRole("button", { name: "登出" }).click();
await page.waitForURL(/\/login/);
```

That click now only opens the confirmation, so nothing navigates and the spec burns its 90s timeout. It is the only sign-out click in the e2e suite.

It will not turn this PR red — root `pnpm test:e2e` is the core vitest LLM suite, and the web Playwright suite runs only under `pnpm --filter @prismshadow/penguin-web test:e2e` — so it breaks silently for whoever runs that suite next.

```js
await page.getByRole("button", { name: "登出" }).click();
await page.getByRole("dialog").getByRole("button", { name: "登出" }).click();
await page.waitForURL(/\/login/);
```

### 2. Nothing pins the confirmation

`packages/web/test/account-menu.test.ts` exists for this file and for exactly this kind of drift. Worth two assertions in its own idiom: that the sign-out row opens the confirm rather than calling `logout()` directly, and that `<ConfirmModal` sits outside `</Dropdown>` — the property the code comment claims, and the one that silently regresses the day someone moves it into the panel.

### 3. Cancelling drops focus to `<body>` — pre-existing, not yours

`Modal` captures the focus-restore target in the render where `open` flips true, which is the same commit that unmounts the dropdown panel; it therefore captures the sign-out `<button>`, which is then destroyed, and `restoreFocusRef.current?.focus()` is a no-op on a detached node. `Dropdown` only returns focus to its trigger on Escape, not on a row click.

The System settings row already does this on `main`, so the mechanism is not new — but this PR adds the first *cancel-and-stay* path where it is observable, since the old sign-out click always navigated away. Keyboard user: open menu, Enter on Sign out, Esc — focus is on `<body>` and the next Tab restarts at the top of the document. The fix belongs in `Dropdown` (return focus to the trigger on any close), as its own change.

### 4. Copy nit

`packages/web/src/lib/strings-en.ts:432` — "carry on on the server" doubles the preposition, and the neighbours prefer a second sentence to a semicolon. The zh side is idiomatic and matches `forkSessionConfirmBody`'s shape exactly.

> This ends your session here and returns to the login page. Running conversations keep going on the server.

### 5. `user-menu.tsx:56` and `:138-140` say the same thing

Both state that a menu row is one click from ending the session. The JSX comment carries the load-bearing part (mounted outside the dropdown so it outlives the menu); the `useState` JSDoc only restates the motive. Drop one.

---

Batch note across #721/#723/#724: all three changelog entries use `## What changed` / `## 变更内容` with no lead paragraph. `changelog/README.md` prescribes H1 → metadata → cross-link → lead paragraph → section, and says to reuse an existing name before inventing one; `## Details` / `## 细节` appear 221/211 times in the corpus and `## What changed` zero times outside frozen `0.2.3/`. If the new shape is deliberate, the README is the place to say so — otherwise it is worth aligning all three.

PR #725, opened 2026-09-14
fix(core,ci): a workspace checkout loads plugins from plugins/, and a plugin change must bump its version

hiyouga:
## What changed

**Root cause of the stale plugins under `pnpm dev`.** `pnpm-workspace.yaml` sets `injectWorkspacePackages: true`, so every `workspace:*` dependency — the twelve `@penguinharness/*` plugin packages included — is installed as an injected copy under `node_modules/.pnpm/…@file+plugins+…`: a snapshot taken at install time, refreshed only after the package's `build` script runs (`syncInjectedDepsAfterScripts`). Plugins have no build script, so a skill added or a `plugin.json` version bumped after the last install never reached the running server: core's loader resolved `@penguinharness/<name>/package.json` through Node and landed on the stale copy, the plugin library kept showing the old version, and installed agents never saw an update to take. The same copy is what the server's own test suite read (the known "passes locally, fails on CI" trap).

- **Loader:** `workspacePluginRoot(name, resolvedDir)` — when a `pnpm-workspace.yaml` sits above the loader's package and `<workspace>/plugins/<name>/package.json` names that package, the plugin is read from there; otherwise the resolved copy stands. An npm install and the packed desktop app have no workspace file above them and are unaffected. Four unit tests cover the redirect, the two fall-throughs, and that the live loader on this tree reads `plugins/`.
- **CI guard:** a new `plugin versions` job runs `scripts/check-plugin-versions.mjs <base sha>`: every plugin whose files changed in the pull request must carry a new `plugin.json` version, because that version is what tells an installed copy it is behind the library. A push has no base and passes; a brand-new or removed plugin passes. The job is in the `ci` gate's `needs`.
- `.github/CONTRIBUTING.md` explains both.

Design: penguin-harness-design #163.

## Verification

Run on the remote test host, not locally: `pnpm --filter @prismshadow/penguin-core typecheck`, core `test/plugins.test.ts` (21 tests), server `test/skills.test.ts` + `test/builtin-agents.test.ts` with core built (29 tests), `pnpm format:check`. The guard script was exercised against a throwaway commit that edited a skill without a bump (exit 1 with the plugin named) and against `origin/main` (no plugin change, exit 0).

## Changelog

`changelog/unreleased/2026-09-14-plugins-from-workspace.md` with its `.zh.md`.


hiyouga:
Reviewed at `cabc6f9a2`.

**The diagnosis is right, and confirmable from the tree.** `pnpm-workspace.yaml:22-24` sets `injectWorkspacePackages: true` with `syncInjectedDepsAfterScripts: [build]`; all twelve `@penguinharness/*` packages have no `scripts` key at all, so the injected copies can never re-sync. In a currently-installed checkout the injected copy is *hard-linked* from the source — `.pnpm/@penguinharness+goal@file+plugins+goal/…/plugin.json` and `plugins/goal/plugin.json` share an inode — which explains the symptom exactly: an in-place rewrite is shared, but a new file, a deleted file or any rename-based write is not.

One correction to the stated scope, which matters below: **the bug's environment is the consumer path, not core's own.** `packages/core/node_modules/@penguinharness/goal` is a *direct symlink* to `../../../../plugins/goal`, so core's own vitest run never had the bug. The server, CLI, web and desktop import `@prismshadow/penguin-core`, land on the injected core, and resolve plugins through *its* sibling `node_modules` — the injected copy. `packages/server/vitest.config.ts` defines no alias, so the server suite goes through that path. The changelog's "nor the server's own tests" is accurate; it is core's tests that were never affected.

### 1. `workspacePluginRoot` redirects the packed desktop app to the repo's `plugins/`

`packages/core/src/plugins/index.ts:228-236` — `workspaceRootAbove` walks to the filesystem root with no bound, and `packages/desktop/electron-builder.yml:29-30` sets `directories.output: out`. So a packed app lives at `packages/desktop/out/<name>/resources/app` — **inside the repo**, underneath its `pnpm-workspace.yaml`:

```
PKG_ROOT    = <repo>/packages/desktop/out/linux-unpacked/resources/app
resolvedDir = <repo>/packages/desktop/out/linux-unpacked/resources/app/node_modules/@penguinharness/goal
returns     = <repo>/plugins/goal          ← redirected
```

`.github/workflows/ci.yml:302-313` (the `runtime` job) builds exactly that tree on ubuntu/macos/windows and runs `<app>/bin/penguin` from it, and `pnpm --dir packages/desktop pack` does the same locally. So the packed app's own collected plugin copies are never the ones it reads: a packaging regression that ships stale or wrong plugin content is invisible to every in-tree smoke test, and a `git checkout` after packing silently changes the packed app's behaviour.

The claim that the packed app is unaffected is stated in four places — the docstring at `:225-226` and `:275-278`, `.github/CONTRIBUTING.md:62-69`, both changelog files, and design PR #163 — so all of them need the same correction.

Minimal fix: redirect only when the resolved copy is already inside the workspace's own store.

```ts
if (!resolvedDir.startsWith(path.join(workspace, "node_modules") + path.sep)) return resolvedDir;
```

That keeps both intended cases (injected core at `<root>/node_modules/.pnpm/…`, and core's own tests which already resolve to `<root>/plugins/<name>`), drops the packed app (`…/out/**/node_modules/…`), and incidentally bounds the "developer's home is itself a pnpm workspace" and "checkout nested inside another workspace" cases — along with the latent fragility in the second unit test, which fails if `TMPDIR` ever sits under a `pnpm-workspace.yaml`.

### 2. A version that moves *backwards* passes the guard

`scripts/check-plugin-versions.mjs:56` — `before !== null && before === after` tests inequality, not progress. `2026.09.01.1 → 2025.01.01.1` with a content change reports "changed and carry new versions", exit 0. That is worse than no bump: `agent-service.ts:239,245` gate the update flag on `comparePluginVersions(plugin.version, skill.version) > 0`, so a backwards version tells every installed copy it is *ahead* and the update is never offered — precisely the failure this job exists to prevent. A typo (`2026.09.14.1` → `2026.08.14.1`) produces it. Parse both sides as `YYYY.MM.DD.N` and require strictly greater.

Lower severity, same line: the current and legacy spellings of one version (`2026.09.01.1` / `2026-09-01.1`) compare *equal* in `comparePluginVersions` but *differ* as strings, so that swap also passes. `PLUGIN_VERSION_PATTERN` catches that one in core's suite; the backwards case is caught nowhere.

### 3. One non-ASCII filename under `plugins/` silently disables the guard for the whole PR

`scripts/check-plugin-versions.mjs:22-25` — git's default `core.quotePath=true` emits `"plugins/goal/skills/\344\270\255…/SKILL.md"` from `--name-only`. That line starts with `"`, so `.filter(line => line.startsWith("plugins/"))` drops it and the script reports "no plugin files changed", exit 0. No such filename exists under `plugins/` today, but `readDirFiles` reads every auxiliary file inside a skill with no name restriction, and this repo actively ships Chinese content. Use `-z` and split on `\0`; the `startsWith` filter then becomes redundant with the `-- plugins/` pathspec and can go.

### 4. The one test that claims to pin the live behaviour is tautological

`packages/core/test/plugins.test.ts:354-358` asserts `existsSync(<repo>/plugins/<name>/plugin.json)` for each loaded plugin. The names come from core's `dependencies` and those files exist by construction — it passes identically with the redirect deleted, and asserts nothing about which root the loader used.

Compounding it: per the correction above, core's own suite is the one environment that never had the bug, so all four new tests run where the fix is a no-op, and the suite that would actually regress — `packages/server` — has no coverage at all. Assert the root instead: `expect(workspacePluginRoot(plugin.name, "/nonexistent")).toBe(path.join(pluginsRoot, plugin.name))`. One line, and load-bearing. Also uncovered: the escape-the-repo case from finding 1, and the `catch` path (the third test exits via the name check, not the catch).

### 5. A non-plugin directory under `plugins/` prints a red `fatal:` and then a false success

`scripts/check-plugin-versions.mjs:44` (git's stderr is not suppressed) and `:65`. `plugins/README.md` and the four `plugins/sandbox-*` packages have no `plugin.json`, so editing any of them yields:

```
fatal: path 'plugins/README.md/plugin.json' does not exist in '<sha>'
plugin versions: README.md changed and carry new versions
```

A green job with a red `fatal:` in its log, claiming a directory carries a new version when it has no manifest. The sandbox backends are actively developed, so this will fire regularly. Pass `stdio: ["ignore","pipe","ignore"]` on the `git show`, and skip directories with no `plugin.json` on either side explicitly rather than by exception.

### 6. Smaller things

- `:34`'s `?? null` plus `:56` means deleting the `version` field entirely also passes (unequal, so it counts as changed). `readPluginDir` throws on it so core's suite catches it, but the guard's own null-tolerance is what opens the hole.
- `packages/core/src/plugins/index.ts:206-218` and `:228-236` are the same find-up loop with a different filename; one shared `findUp(from, file)` covers both. And `workspaceRootAbove(PKG_ROOT)` runs once per plugin inside `pluginRoots()` — roughly 7 `existsSync` × 13 plugins on every `loadLibraryPlugins()`, which the server calls per request — while `PKG_ROOT` itself is module-memoized at `:219`. Small next to the plugin file reads; it is the asymmetry that stands out.
- `packages/core/src/index.ts:37` is `export * from "./plugins/index.js"`, so `workspacePluginRoot` is now published SDK surface, exported only so a test can call it. Worth a deliberate decision rather than a side effect.
- The changelog pair has no lead paragraph and uses `## What changed` / `## 变更内容` where `changelog/README.md` says to reuse `## Details` / `## 细节` before inventing a name (34 of 34 other `unreleased/` entries do). The body also states current repository state — "an npm install and the packed desktop app keep resolving their own copy" — which the README forbids and which finding 1 shows is inaccurate.

### Checked and clean

CI wiring verified rather than trusted: `plugin-versions` is in the `ci` gate's `needs` (`:331-343`), the gate fails on any non-`success` result (`:348-357`), and the job sets `fetch-depth: 0` so the base-sha diff has its history. The three-dot `${base}...HEAD` gives merge-base semantics, so a stale base sha after `main` moves is handled. The deliberate pass paths all behave as claimed when executed — empty base, brand-new plugin, removed plugin — and a file *deleted* from an existing plugin correctly fails. The wrong-package-name guard sits inside the `try`, after the read and before the redirect is returned, and is covered. Windows separators are fine (`path.join`/`dirname`, terminating at `C:\`). The function is purely lexical and both inputs are realpath'd by Node, so symlinks stay consistent. Every packaging path other than desktop `out/` is safe: core's npm tarball ships `["dist","LICENSE"]`, the Docker runtime image copies into a clean `ubuntu:24.04`, and the release tarballs are assembled and packed, never run in-tree. Both `try/catch` blocks are load-bearing, not defensive. `format:check`, `oxlint` and `git diff --check` clean on all 7 files; no Chinese outside the `.zh.md`.

PR #740, opened 2026-09-15
fix: ignore non-ticket Markdown files in company boards

### LINKED ISSUE #735: Company mode: every .md in a ticket column is read as a ticket
Found by the automated reviewer on #587 and verified against the merged code.

The ticket store filters the month directory with `MONTH.test(...)` but does not filter the file basename, so every `.md` in a column is read as a ticket even though the method's own doc comment says "A file whose name is not a ticket id … is skipped".

Two outcomes, both reachable because the organization directory is hand-editable by employees by design:

- A stray file that fails `parseTicket` is reported as `org_ticket_invalid` noise on every 30-second reconcile pass instead of being skipped.
- One that *does* parse — a copied ticket renamed, say a `README.md` that kept its frontmatter — becomes a board row and an accounting entry whose id no route will accept. `requireTicketId` and the service both reject it, so it cannot be moved or closed through the API.

**Fix:** filter the derived basename before reading it — but the filter must accept `LEGACY_TICKET_ID_PATTERN` as well as `TICKET_ID_PATTERN`. Filtering on the current pattern alone would drop pre-existing tickets whose slug contains digits, which is exactly why the legacy pattern exists.

521Peter:
## What this changes

Company ticket discovery now filters Markdown files by the current and legacy ticket ID patterns, so files such as README.md are not parsed as tickets. A focused regression test covers valid current IDs, legacy IDs, and unrelated Markdown files.

Closes https://github.com/Prism-Shadow/penguin-harness/issues/735

## Verification

- pnpm --filter @prismshadow/penguin-server exec vitest run test/organization-store.test.ts (1 test passed)
- pnpm --filter @prismshadow/penguin-server typecheck (passed after building workspace dependencies)
- pnpm format and pnpm format:check (passed)
- git diff --check (passed)

The local runtime was Node 22.23.2, so pnpm emitted the repository Node >=24 engine warning.

## Checklist

- [x] Changelog entry pair under changelog/unreleased/ in the documented bilingual format.
- [x] Docs updated where this changes documented behavior (not applicable; internal file discovery now matches the documented ticket naming rules).
- [x] No API key, bot token or other credential in the diff, the logs, or the screenshots.


hiyouga:

PR #744, opened 2026-09-16
feat(core,server,web): a request's cost is fixed when it completes

hiyouga:
## Summary

A request's cost is now fixed the moment the request completes, at the price in force then, and nothing re-prices it afterwards. Editing a price, running **Sync presets**, or a promotion starting or ending changes only what later requests cost. Paired design PR: Prism-Shadow/penguin-harness-design#169.

### Why

The cost center priced every usage row at the Project's *current* price on every query. That had two consequences:

- A catalog promotion was written into the Project as a discounted price. After a release removed the promotion, the Project kept billing at the discount until the user clicked **Sync presets**.
- Clicking it re-priced the model's whole history, including usage that ran while the promotion was live.

The user decided that a cost is settled on the day it is incurred and only new usage follows later prices. To keep PRN-024 (SQLite holds only caches rebuildable from files), the rates are recorded in the Trace rather than only in `web.db`.

### What changed

**core**
- `token_usage` gains an optional `pricing` (`TokenUsagePricing | null`): the rates the request was billed at.
- `GenerativeModel` stamps it on a completed request through a new `resolvePricing` config hook. The hook is called with the event's own timestamp, before the event leaves the LLM layer, so the stream and the Trace carry the same rates. If the lookup fails, the field is left off and the usage is kept.
- `resolveBilledPricing` reads the Project's price on disk *now*, falling back to the Session's own entry when there is no file or the model is gone, and applies `billedPricing`.
- `billedPricing` is the single pricing rule. A stored price that still equals the catalog's list (or peak) price gets the discount live at that instant; any other stored price is billed as stored.
- Catalog:
  - `presetModelEntries` writes the list price for every row.
  - New `discountUntil` (exclusive end) and `discountRateAt`.
  - `offPeakScheduledRefs` removed.
  - OpenRouter's baked promotions are now declared: `openai/gpt-5.6-sol` stores the list price with `discount: 0.5`, and `z-ai/glm-5.3-flash` stores its list price with the promotion's documented end of 2026-09-09 16:00 UTC, so it bills list price now.
  - The Gemini 3.6/3.7/3.8 Flash launch discounts end at 2027-01-01T08:00Z.

**server**
- Migration 9 adds `usage_records.cost` and `usage_records.cost_settled`. It is swap-safe.
- `UsageRecorder` fixes each row's cost:
  - from `payload.pricing` when present;
  - `null` when the event says the model has no price;
  - otherwise priced at record time by the same rule. This covers title requests, failed requests (0 for a priced model), and a lookup failure (uncosted, row kept).
- The usage repo sums `cost` and counts uncosted rows. `peakExpr`, `PeakTier`, `TieredRates` and `ratesAt` are gone, and `getPricing` returns the stored price.
- `UsageService.settleUnsettledCosts` (compat) costs rows written before the upgrade, or by a rolled-back build, once at startup (`Startup.setup`). It uses the pre-change rule: stored price, off-peak tier only for rows still at catalog peak, no flat discount. The cost center therefore shows the same figures after the upgrade.
- The Trace page's per-turn cost is unchanged in spirit: the current stored price through `billedRates` at each request's timestamp.

**web**
- The discount badge compares against the list price and hides once a promotion has ended.
- The live cost chip and per-Task footer estimate use `billedBucketPricing` (the rate billed right now).

**docs / changelog**: models, configuration, omni-message and interfaces pages (en + zh), and bilingual changelog entries.

### Backward compatibility (decided by the user)
- **Usage records:** one-time settle at the pre-upgrade price. The compat code is removed in the release preparation of the second release after this ships.
- **Project configs holding baked discounted prices:** no migration. Those rows bill as stored, show no badge, and are counted by the Models sync badge until **Sync presets** runs once.
- **Traces:** additive optional field; nothing to migrate.

Details are in `changelog/unreleased/2026-09-16-backward-compatibility.md`.

### Overlap to note
#716 (Penguin Go) edits `model-catalog.ts`, `project-config-service.ts`, `catalog-sync.ts` and `discountedPrice`. It also stores platform-synced promotions in the Project as an effective `pricing` plus `list_pricing` / `discount`. Whichever PR merges second needs to reconcile `discountedPrice` and the preset/sync price convention.

## Verification

- [x] `pnpm install --frozen-lockfile`, `pnpm -r build`
- [x] `pnpm typecheck` (whole workspace)
- [x] `pnpm format:check`, `pnpm lint`, `git diff --check`
- [x] Full package suites on a 24-core Linux box (vps3):
  - core: 58 files / 1228 tests passed (5 skipped)
  - server: 140 files / 2045 tests passed
  - web: 175 files / 2321 tests passed
  - docs: 7 files / 53 tests passed
- New or changed coverage:
  - core `llm.test.ts`: the rates are stamped at the event's own timestamp; `null`, no resolver, and a failing resolver keep the usage.
  - core `state.test.ts`: `resolveBilledPricing` reads the price on disk at the moment, and falls back when the file is missing or no longer names the model.
  - core `model-catalog.test.ts`: `discountUntil` semantics, `billedPricing` rule, and list-price presets.
  - server `usage.test.ts`:
    - the recorder fixes costs from the event, prices title and failed requests itself, and keeps rows when the lookup fails;
    - queries never re-price;
    - a later price change leaves recorded rows alone;
    - the one-time settle uses the pre-change rule and is idempotent.
  - server `db-migrations.test.ts`: migration 9 up/down, swap-path, ADOPTION.
  - server `trace-service.test.ts`: turns priced at today's price per timestamp equal the cost center's figure while the price is unchanged (via the real recorder).
  - web `model-grouping.test.ts`: badge against list price, and hidden after `discountUntil`.
- [ ] Web e2e (Playwright) not run: no spec covers the badge or pricing rule, and the local e2e baseline has known pre-existing failures.


hiyouga:
Closing: we are keeping costs computed at query time from the current price, with no rates recorded in the Trace, so fixing a request's cost at record time is not needed. The stale OpenRouter `z-ai/glm-5.3-flash` price found here is fixed separately in a small catalog PR. The branch is kept in case any of this is revisited.

PR #746, opened 2026-09-16
fix(web): the terminal's "New shell" action reads "New terminal"

hiyouga:
## Summary

- The action that starts another terminal — the dock's add-tab menu item and the button on the standalone `/terminal` page, both reading `S.terminal.newShell` — said "新建 Shell" in Chinese and "New shell" in English, while every other terminal label (tab title, "关闭此终端", "Kill this terminal") says terminal.
- It now reads **新建终端** / **New terminal**. The dictionary key is unchanged, so both call sites pick it up.
- Bilingual changelog entry added.

## Verification

- `git diff --check` clean.
- `pnpm format:check` on the remote test host (vps3): all matched files use Prettier code style.
- No test pins the label (`grep -rn "新建 Shell\|New shell\|newShell" packages/web/test` is empty); the change is two dictionary values, so no typecheck-relevant surface moved.


packages/web/src/lib/strings-en.ts
```diff
@@ -76,7 +76,7 @@ export const en: Strings = {
   /** Server-side terminal (the in-app dock and the standalone /terminal page). */
   terminal: {
     title: "Terminal",
-    newShell: "New shell",
+    newShell: "New terminal",
```
hiyouga, line 79: **[copy]** The value is renamed, but the old label is still quoted in places the PR does not mention. `terminal-page.tsx:53`, `:120` and `:141` still say "New shell" (for example, "showing the exited shell with \"New shell\" one click away"), and so does `terminal-appearance.ts:22` ("detach, new shell, close"). The e2e spec does too: `e2e/terminal.spec.mjs:8` and the `:99` test title `New shell starts a fresh session`. None of these break anything, since the e2e test clicks `data-testid="terminal-new-shell"`. But the next reader will look for a button that no longer exists. Fix: say "New terminal" in those comments and in the test title. The Verification grep missed them because it searched only `packages/web/test`, not `packages/web/e2e`.

hiyouga (commented):
Review of the rename: both dictionaries, both call sites, tests, e2e, docs, specs, desktop, and changelog.

Findings (ranked):
1. **[copy]** Code comments still quote the button as "New shell" after the rename: `packages/web/src/features/terminal/terminal-page.tsx:53`, `:120`, `:141` (`showing the exited shell with "New shell" one click away`) and `packages/web/src/features/terminal/terminal-appearance.ts:22` (`detach, new shell, close`). These are outside the diff. The PR does not mention them. Fix: change the quoted label to "New terminal" in those four comments.
2. **[test]** `packages/web/e2e/terminal.spec.mjs:8` (header: `"New shell" is the one action that deliberately drops the session`) and `:99` (test title `New shell starts a fresh session`) still use the old name. Nothing breaks, because the click goes through `data-testid="terminal-new-shell"`. But the PR's Verification grep only searched `packages/web/test` and skipped `packages/web/e2e`. That is why it says no test mentions the label. Fix: rename the test title and header line to "New terminal".

Checked and correctly left alone:
- zh `新建终端` and en `New terminal` match each other and the labels around them (`title` 终端/Terminal, `killShell` 关闭此终端/Kill this terminal, `detach` "Open in new window", `dock.addTab` "Add panel"). The case style matches too.
- `dock-panel.tsx:607` and `terminal-page.tsx:184` are the only readers of `S.terminal.newShell`. Keeping the key name follows `killShell`, which already reads "Kill this terminal".
- The "New shell" / 新建 Shell mentions in `changelog/0.2.4/2026-08-21-dock-tab-sidebars.zh.md` and `changelog/0.2.10/2026-09-02-terminal-close-shortcut{,.zh}.md` are in released folders, which are frozen.
- `packages/docs/content/web-app.{en,zh}.md` (Ctrl+` "adopting or starting a shell" / 新建一个 Shell) describe starting the shell process, not the button label. `packages/cli/src/i18n.ts` "new shell" means the user's OS shell. Neither needs a change.
- There is no command palette entry or desktop tray/menu item for this action, and `specs/` does not name the label.
- `dock.killConfirmBody` still says Shell/shell, but it refers to the process being ended, so it is fine.
- The changelog pair follows `changelog/README.md`: metadata order, full PR URL, and the two files match section for section.
- There is no over-defensive code: the diff changes two string values.

Verdict: fix before merge: 1, 2 (stale label quotes only; no behaviour change).

packages/web/src/lib/strings-en.ts
hiyouga: The four comments in terminal-page.tsx (:53, :120, :141) and terminal-appearance.ts (:22), plus the e2e spec header (:8) and test title (:99), now say "New terminal" (f63ba4602).

PR #748, opened 2026-09-16
feat: switch the model inside a Session — compact on the current model, continue on the new one

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/170

## Summary

A Session can now switch models in place. The switch compacts on the **current** model first; only when that compaction completes does the next model context (and Trace file) open on the target model, carrying the summary. A Session can therefore run several models over its life. The thinking level is unchanged — the user still chooses whether to compact first — while a model switch always compacts. `/model` keeps its meaning (a handoff that opens a **new** conversation on another model) and its copy now says so.

**Protocol: no OmniMessage field or enum value added.** A switch's compaction is an ordinary `manual` pair, and the new model is recorded only where a model already lives — the new context's `session_meta`. The earlier revision of this PR added `reason: "model_switch"` and `next_provider` / `next_model_id` to `compaction_begin` / `compaction_end`; both are removed, along with `resumeTrace`'s `nextModel` / `danglingCompaction.next` and the `model_switch` resume branch.

### Core
- The model reference moved from `SessionSpec` (Session-fixed) to `AssembledContext` (per context); `openNextContext({ modelRef })` resolves the target from the Project config on disk.
- `Session.switchModel({ provider, modelId, signal })` → `{ status, previous, next }`: validates target + client construction before any event; same model is a no-op; always summarizes (ignores `compaction.mode: discard`).
- **Durability rule — the eager open.** A completed switch performs the Trace rotation at once instead of at the next message: the new file gets the context's `session_meta` (the target model), the opener's records, and in summarize mode the `[context_summary]` as the context's first input, which then heads the carry-over and is never written twice. This is the one exception to lazy file creation. The same `session_meta` is yielded as the switch's last record. `resumeSession` reads the model from the latest file alone (`ref` is that file's meta again), so a restart right after a switch lands on the new model with the summary pending.
- No-turn paths: just compacted → no request and **no second pair**, the pending rotation happens on the target; an open context without a completed turn → a `manual` / `discard` pair, and its text input rides to the target **in memory only** (carry-over may hold synthetic messages, never persisted) — except the summary that context opened with, which the engine remembers (`contextSummary`, until the context's first completed turn) and writes again at the head of the next file, so a resume never replays a history without it (switching twice before typing, a failed or stopped first request on the new model, a compaction followed by a failed request); never ran → silent re-assembly.
- Carry-over handed to a context on a model without vision is folded through `foldInputImages` first — the engine's held input and the Session's aborted-bootstrap input alike.
- Refusals are typed: `ModelSwitchRefusedError` with a `reason` (`model_not_configured`, `model_unavailable`, `compaction_not_configured`, `summary_too_large`) is thrown before any event; anything else a switch throws is a failure. A summary the switch's own compaction produced that does not fit the target ends the compaction `fatal`; a summary already held (a switch right after `/compact`) refuses the switch before any event (`summary_too_large`), compared against the same `summaryRoom`.
- A resumed file past the first with no completed turn answers `just_compacted` (`fromCompaction` derived from the file index, not from records).
- `Session.provider/modelId` are getters; spawn inheritance, image fold, `read_file` vision describer and bare LLM follow the current entry; explicit SDK credentials only apply to the creation model.
- The rule is written down: `omnimessage/types.ts` opens with the closed-vocabulary paragraph, the dev skill carries one bullet, and 05「消息格式」 states it in the spec.

### Server
- `POST /api/sessions/:id/switch-model` → 202 (streams, status `compacting`) or 200 `SessionResponse` for a never-ran Session; 409 `task_in_progress` / `compacting` / `same_model` / `model_not_configured` / `model_unavailable` / `compaction_not_configured` / `summary_too_large`. `switchRefusal` maps core's typed refusals only — no message regexes — and rethrows everything else, so a real failure before the first event answers 500 with its logging. A summary held from a compaction just made that does not fit the target answers a dedicated 409 `summary_too_large` (core's message names both sizes); the Web App and the CLI say so and point at a model with a larger context window. A row that vanished during the request answers 404 rather than the pre-switch row.
- The 202 stream is a plain manual pair when there was something to close, then the opener records and the new context's main-session `session_meta`. Row + runtime entry adopt the runtime's model as that meta passes through the drive (before its publish, so a client refetch cannot race it) and again after the drive; `completedSwitchTarget` is gone. `registerChildSession` returns early for an origin-less `session_meta`, and the drive's child-registration branch requires an origin. Usage: the switch's compaction bills to the old model; fork takes the cut shard's model; trace index reads the latest shard.

### Web
- The open conversation's model badge is a `ModelSelect`; picking another model confirms; disabled while running/compacting. The dialog is worded for what the switch will do (`switchContextShape`): **Compact and switch** when there is conversation since the last compaction; **Switch** with "no context yet" for an empty conversation; **Switch** with "continues from that summary, nothing is compacted again" right after a compaction — and the toast promises a compaction only in the first case.
- The stream shows a plain compaction row followed by a "模型已切换 · A → B" / "Model switched · A → B" marker (`ModelChangeItem`), derived from two consecutive main-session `session_meta` records; a just-compacted Session shows the marker alone, a failed compaction no marker. The compaction banner is back to main.
- The Session DTO stays the authority: `sessionRowStale` replaces the refetch watch — once the Session is idle, a row naming another model than the running context's `session_meta` is refetched. The pair is recorded as fetched only when the fetch succeeded; a failure is surfaced and retried at the next stream change, so a transient error cannot leave the old model showing.
- The marker counts as between-rounds housekeeping, like a compaction row, for the stats-row placement and the thinking-level cache-risk check.
- `/model` copy (slash description, picker title + footnote, chip, banner) states it opens a new conversation and points at the toolbar picker.

### CLI
- `penguin chat` `/switch-model <provider> <model_id>` (bare form prints the current model), 409 mapping, `model: A → B` from the Session re-read after the stream; the switch's compaction renders like any other (`render.ts` and the compaction lines are back to main). `--resume` still rejects model flags and points to `/switch-model`.

### Docs
Merged with main after #758's rewrite: this PR's intent is re-authored on main's text in `sessions-and-traces` (an "In-session model switch" section, then "New conversation on another model (/model)"; anchors from `chat` intact), `omni-message`, `message-flow`, `models`, `server-api` (the `/switch-model` row and bullet, the two-ways paragraph) and `cli`, in both languages. `chat.{en,zh}.md` (which merged cleanly but still said the model is locked) now describes both ways under the unchanged "Switch the model" heading, and `agent-loop.{en,zh}.md` says the model entry is per context and lists the in-session switch under the `manual` reason.

### Prompt cache
No expectation moved: `prompt-cache-invariants` (7), `prompt-cache-lifecycle` (11) and `prompt-cache-simulator` (14) pass unchanged; the review round edits none of them. The switch's compaction request is still an ordinary compaction request on the old LLM object (same config fingerprint, one appended user turn). The eager open and the review fixes change only the Trace, the carry-over's fold and where the held summary waits, so the new model's first request is still `[summary, prompt]`. The context on the new model is a new cache line, like any rotation.

### Backward compatibility
None needed: no OmniMessage field or value was added, no released build ever wrote the earlier `model_switch` shape, and the `sessions` / `trace_sessions` model columns already agree with every shard on disk.

### Decided
- The 409 code for a held summary that does not fit is a dedicated `summary_too_large` (the user's call), not a reuse of `model_unavailable`. The summarize path's equivalent stays a streamed `fatal` end.

### Known limitations
- The chat header's cost estimate prices the whole Session at the current model's rate; the cost center and Trace views are per record and correct.
- Crash windows: between the closing `compaction_end` and the new file's `session_meta` a crash resumes on the old model with the summary pending (the user switches again); a crash between a discard-path switch and the next prompt loses the in-memory text carry-over other than the summary (same as a discard `/compact`). The one remaining summary gap: a restart between two switches with no completed turn in between rebuilds the summary as ordinary pending input, so a second discard-path switch does not write it again (it stays in the closed file).
- On the live stream the new `session_meta` follows the opener's records, while the file holds it first; with MCP servers configured, the marker and the connect row swap places after a reload.
- A `tailLimit` window that starts after the previous context's meta derives no marker (the badge still follows the DTO).

## Verification
All builds, checks and vitest runs on the vps3 box (`pnpm -r build`, then the suites, `pnpm typecheck`, `pnpm lint`, `pnpm format:check` in one chain, exit 0). Locally only prettier on touched files and `git diff --check`.
- Red first: the amended `core/test/model-switch.test.ts` run against the pre-fix head `192e7c8c8` fails 9 of 22 — the seven behaviour cases (summary in the third file, failed-first-request switch, compact → failed request → switch, held-summary refusal, both image folds, the just-compacted-then-written discard) and the two typed-refusal cases — and passes 13.
- Core, 6 files / 121 tests: model-switch 22, resume 24, agent 43, prompt-cache-invariants 7, prompt-cache-lifecycle 11, prompt-cache-simulator 14.
- Server, 4 files / 77 tests: session-switch-model 14, session-manager 45, session-fork 12, trace-index 6. Round 2 (the dedicated code): session-switch-model + session-manager 59, web model-switch + api-error-text 22, CLI 33 files / 466, docs 62, typecheck / lint / format:check clean.
- Web, 4 files / 179 tests: stream-model 119, model-switch 15, api-error-text 7, thinking-level 38.
- Docs, 8 files / 62 tests (in-page and cross-page anchors included).
- CLI (`FORCE_COLOR=0`), 33 files / 464 tests.
- `pnpm typecheck`, `pnpm lint`, `pnpm format:check`: clean.
- Not run: Playwright e2e. `e2e/draft.spec.mjs`'s `/model` selectors were updated to this PR's renamed handoff copy but not executed; `e2e/layout.spec.mjs`'s 320px overflow check is worth a run for the toolbar model pill.


packages/core/src/engine/context-engine.ts
```diff
@@ -1062,6 +1113,100 @@ export class ContextEngine {
...
+      yield* this.startNewContext(next);
+      yield* this.openContextFile();
+      return "completed";
+    }
+    // The context's own records go into its file ahead of the pair that closes it.
+    await this.writeOwedBootstrapRecords();
+    this.pendingCarryOver = this.pendingCarryOver.filter(
+      (m) => (m.payload as { type?: string }).type !== "tool_call_output",
+    );
+    yield* this.discardContext("manual", next);
```
hiyouga, line 1242: **[bug]** The summary drops out of the Trace on this path. `openContextFile` (L1204) moves the summary into `pendingCarryOver`, where nothing marks it as the summary any more. This branch carries the whole carry-over into the next context in memory only. So if the context the summary opened has no completed turn when the user switches again, the summary never reaches the next file. Plausible flows:
- switch A→B, then B→C before typing;
- switch A→B, B's first request fails (bad key, a 400 on a parameter, quota) or is stopped, then switch to C or back to A;
- a compaction, the first request after it fails, then a switch.

The live engine still sends the summary, but C's file never holds it. Any resume of that context therefore replays a history without it: a desktop relaunch or a server restart at any point before C's next compaction, not only a crash before the next prompt. The model on C then silently loses everything before the first switch. Spec #170 (the durability clause of the model-switch section in 05-ARCHITECTURE) and design v2 §2.4 treat this as the crash-window gap of a discard `/compact`. There, though, the lost item is an interrupted prompt the user can re-send. Here it is the only record of the whole prior conversation. The "just-compacted context the user already wrote on" test pins the loss: `resumeTrace(opened).carryOver` is `[]` while the summary rides in memory.

Probe on vps3 against this head, both cases failing:
- A→B→C before typing, then one turn on C: C's file is `session_meta, tool_list_ready, text("task two"), request…`. `resumeTrace` of it gives the user texts `["task two"]`, with no summary.
- A→B, B's first request `fatal`, then B→C and one turn: the user texts are `["task three"]`. C's live first request was a `[turn_aborted]` flatten wrapping the summary and "task two", followed by "task three". After a failed request the summary sits inside that synthetic flatten.

Fix: the summary is a real record (`runToCompletion` writes it as a context's first input), so it may be written again. Remember the summary a context opened with: set it where `openContextFile` or `runToCompletion` writes it as the first input, and clear it at the context's first completed turn. In this branch, write it at the head of the next file right after the rotation, and leave the in-memory carry-over as it is. Moving it out of the carry-over would change request assembly after a failed request, while writing it only to the file does not. The replayed history then starts with the summary, as the live request does. Add a turn on the new model to "switching twice before typing" and to a failed-first-request variant, and assert `resumeTrace(<its file>).history` starts with the `[context_summary]`. (A restart between the two switches rebuilds `carryOver = [summary]` without that identity. Whether that narrower window is acceptable is a separate call.)

packages/core/src/engine/context-engine.ts
```diff
@@ -1062,6 +1113,100 @@ export class ContextEngine {
...
+      // records still owed (a Session resumed into this closed context and switched before
+      // running) describe the context being left, which never gets a file: dropped.
+      this.pendingBootstrapRecords = null;
+      yield* this.startNewContext(next);
+      yield* this.openContextFile();
+      return "completed";
+    }
+    // The context's own records go into its file ahead of the pair that closes it.
+    await this.writeOwedBootstrapRecords();
+    this.pendingCarryOver = this.pendingCarryOver.filter(
```
hiyouga, line 1236: **[bug]** Images carried across a switch reach a text-only target unfolded. This filter keeps everything except tool outputs, and nothing folds the rest for the model the next context opens on. The Session folds only new prompts: `runTask`, session.ts:577, which prepends `carryOverInput` after that fold at :582. The engine folds only steering.

Scenario: on a vision model, a prompt with a screenshot that never got an answer and was held as-is. That happens when it is stopped during the first MCP connect (the Session's `carryOverInput`), stopped during a retry wait, or out of retries (the engine's carry-over). The user then switches to a text-only model, which takes this branch. A `fatal` or mid-stream stop is not affected: its `[turn_aborted]` flatten already drops images. Probe on vps3 against this head: B's first request is `["text","image_url","text"]` both for a stop during the first connect and for a stop during the retry wait. The text-only model then either rejects the request, so the first message on it fails, or ignores the image. Either way the model never gets the `[attached image: …]` path line that lets `read_file`'s vision describer read the picture. Before this PR the carry-over was always built for the running model.

Fix: when the opened context's model has no vision, run the carried inputs through `foldInputImages`, the fold steering already uses. That means the engine's `pendingCarryOver` here and the Session's `carryOverInput`.

packages/core/src/engine/context-engine.ts
```diff
@@ -1062,6 +1113,100 @@ export class ContextEngine {
...
+      yield* this.openContextFile();
+      return "completed";
+    }
+    if (this.pendingTraceRotation) {
+      // Closed by a completed compaction, nothing written since: no pair — the closing pair is
+      // on the closed file. The pending rotation is performed now, on the target. First-run
+      // records still owed (a Session resumed into this closed context and switched before
+      // running) describe the context being left, which never gets a file: dropped.
+      this.pendingBootstrapRecords = null;
+      yield* this.startNewContext(next);
```
hiyouga, line 1230: **[bug]** The summary-size guard only runs in the summarize branch. Here the held summary goes to the target unchecked, although `target.summaryRoom` was computed for exactly this.

Probe on vps3 against this head: `/compact` on A produces a ~15k-token summary, then the user switches to an 8k-window target. The switch completes: no pair, and the Session moves to the target. The same summary in the first branch ends the switch `fatal`, which is your own test. On the target, the first request is rejected. The carry-over then becomes a `[turn_aborted]` flatten that still holds the whole summary, so every later message is rejected the same way until the user switches again.

Fix: before rotating, compare the held summary with `target.summaryRoom`, here and in the discard branch once it carries the summary (L1184). Refuse before any event, the way target validation does, so the Session stays on its model.

packages/server/src/runtime/session-manager.ts
```diff
@@ -128,6 +132,41 @@ function compactUnavailable(why: Exclude<CompactAvailability, "ok">): HttpError
...
+    return new HttpError(409, "compaction_not_configured", message);
+  }
+  if (isMissingCredential(err)) {
+    return new HttpError(
+      409,
+      "model_unavailable",
+      `Model ${ref.modelId} has no API key yet. Configure it on the Models page first.`,
+    );
+  }
+  return new HttpError(409, "model_unavailable", message);
```
hiyouga: **[over-defensive]** This catch-all labels every other pre-event failure `model_unavailable`. The Web App renders that code as "That model cannot be used yet (it may have no API key) — configure it on the Models page first", and the CLI as "is configured but cannot be used". Besides a target that cannot be constructed, it catches:
- an Agent State that no longer parses (`reassembleInitialContext` for a never-run Session, or the opener on the just-compacted path);
- a resumed Session's bootstrap failing in `ensureReady`.

Those also skip `handleError`'s logging of unexpected errors.

Fix: have core mark its own switch refusals (an Error subclass with a code, thrown by `validate`, the compaction check and the summary-size refusal), map exactly those, and rethrow the rest. That also removes the two message regexes at L147/L150, which tie the server to core's wording. The tests throw copies of those sentences from a fake (`session-switch-model.test.ts:283` and `:308`), so a rewording in core stays green while production falls through to this line.

packages/server/src/http/routes/sessions.ts
```diff
@@ -1289,6 +1291,29 @@ export function sessionsRoutes(deps: SessionsRouteDeps): Hono<AppEnv> {
...
+    const ref = {
+      provider: requireString(body, "provider", { minLen: 1 }),
+      modelId: requireString(body, "modelId", { minLen: 1 }),
+    };
+    const outcome = await deps.manager.startSwitch(row.sessionId, ref);
+    if (!outcome.switched) {
+      return c.json({ sessionId: outcome.sessionId } satisfies TaskCreateResponse, 202);
+    }
+    // Re-read after the switch: the manager moved the row to the new pair inside the call.
+    const fresh = deps.sessionsRepo.findById(outcome.sessionId) ?? row;
```
hiyouga: **[over-defensive]** `?? row` would answer 200 with the pre-switch row, which names the old model, as the switched Session. The row can only be missing here if it was deleted during the request. Answer 404 as `resolveSession` does, instead of reporting a model the runtime no longer runs on.

packages/web/src/features/chat/chat-page.tsx
```diff
@@ -1497,6 +1522,105 @@ export function ChatPage() {
...
+        provider: ask.to.provider,
+        modelId: ask.to.modelId,
+      });
+      const outcome = modelSwitchOutcome(res);
+      if (outcome.kind === "applied") {
+        applySessionRow(outcome.session);
+        toastSuccess(S.chat.modelSwitchInSessionApplied(to));
+        return;
+      }
+      toastInfo(S.chat.modelSwitchInSessionStarted(from, to));
```
hiyouga: **[copy]** For a Session that was just compacted and has no turn since, the switch runs no compaction and streams no pair, only the new context's opener records and meta. The copy still promises one:
- the dialog body says "The context is compacted on the current model first";
- the confirm reads "Compact and switch";
- this toast says "Compacting the context on …".

The page can already tell when the transcript ends in a completed compaction: `prefixCacheAtRisk` walks the same trailing run. Treat that case like the empty transcript, with its own body line (the conversation continues on B from the existing summary). The empty-transcript body, "has no context yet", would be wrong there too.

packages/web/src/features/chat/chat-page.tsx
```diff
@@ -1497,6 +1522,105 @@ export function ChatPage() {
...
+  // that still disagrees is not refetched in a loop. Runs per stream version because the model
+  // mutates in place.
+  const staleRowFetchRef = useRef<string | null>(null);
+  useEffect(() => {
+    const contextModel = stream.model.contextModel;
+    if (selectedSessionId === null || stream.loading || stream.taskState !== "idle") return;
+    if (contextModel === null || !sessionRowStale(contextModel, activeModelRef)) return;
+    const key = `${selectedSessionId}:${contextModel.provider}/${contextModel.modelId}`;
+    if (staleRowFetchRef.current === key) return;
+    staleRowFetchRef.current = key;
```
hiyouga: **[over-defensive]** The pair is marked as fetched before the fetch succeeds, and the `.catch(() => undefined)` below drops a failure. One transient `GET /sessions/:id` error leaves the badge, the context ring and the price on the old model until a reload, because this key then blocks every retry for that pair. Set the key in the `.then`, or clear it in the catch.

hiyouga (commented):
Independent review of `192e7c8c8`. I read the core engine, Session, Agent composition, server manager/route/trace index/fork, web stream model/picker/page and the CLI against design v2 and spec #170. I confirmed the core bugs with probe tests on vps3 (below). The findings are in the inline comments, except one docs finding below that has no diff line.

### Checked, no finding
- Refusals before any event: target validation from disk plus client construction, same model, compaction not configured.
- A failed or aborted switch compaction keeps the old model, context and file, and streams no `session_meta`.
- Eager open: `session_meta(B)`, records and the summary land in the new file. `resumeSession` of that file lands on B with `carryOver = [summary]`, and the `fromCompaction` derivation answers `just_compacted`.
- Usage: the switch's compaction request bills to A and later requests to B. Trace analysis prices each file by its own `session_meta`; fork takes the cut shard's model; the trace index follows the latest shard.
- Subagents spawned after a switch get B, B's thinking base and B's own credentials. An SDK credential override stays with the creation model.
- Vision answer, vision describer and tool selection follow each context.
- An origin-less `session_meta` no longer reaches child registration; the usage recorder and error watcher already ignore it.
- `pendingBootstrap` holds for the opener's records dedup against the eagerly written file.
- Web: the `model_change` derivation, stats-row placement, the thinking-level cache check and the `sessionRowStale` predicate. The CLI renderer ignores a top-level meta.
- zh/en dictionaries are in step, and the `/model` copy says "new conversation" on every surface.
- **Prompt cache:** no existing expectation moved in `prompt-cache-invariants`, `-lifecycle` or `-simulator` (the diff only adds one case to each of the first two and imports `sessionMeta`). The new lifecycle case declares one by-design miss, the target's first request, which is the inherent cost of switching models.

### Probes
I ran the three core `[bug]` findings as vitest cases on vps3 against `192e7c8c8`. They reuse the harness of `packages/core/test/model-switch.test.ts` and are not pushed anywhere. All of them fail on this head, and each inline comment quotes its result:
- summary lost after switching twice before typing, and after a failed first request on the new model then another switch;
- image carried unfolded after a stop during the first connect, and after a stop during a retry wait;
- an oversized held summary switched without the guard.

One scenario I expected to fail passes. An image prompt whose request ends `fatal` loses the image to the `[turn_aborted]` flatten before the switch, which is the existing multimodal TODO in `flattenCarryOver`, not this PR.

### Finding without a diff line
- **[copy]** `packages/docs/content/agent-loop.en.md:193` and `agent-loop.zh.md:186` (untouched by this PR) still say "What stays fixed for the Session's lifetime is the Session itself: its id, Workspace, model entry (credentials, window and per-model annotations included) and origin". That is exactly what this PR changes. The reason table's `manual` row (en:189, zh:182) also still says only "`/compact` or `session.compact()`". Say the model entry is per context and add the in-session switch to `manual`, as spec #170 does. On main these lines are en/zh 399 and 339.

### Merging main (for whoever merges)
`git merge-tree` against `origin/main` (`078a7aa83`) shows 14 conflicts, all docs: `cli`, `message-flow`, `models`, `omni-message`, `server-api`, `sessions-and-traces` and `web-app`, each in `.en.md` and `.zh.md`. #758 rewrote every page, so re-author this PR's intent onto main's text instead of resolving hunks. Where each edit lands on main (en; zh at the same line numbers):
- `sessions-and-traces`: l.17 (Session row, "model and Workspace are locked"), l.98-99 (resume step 2, "immutable"), `## Model switch (/model)` at l.134 → the in-session switch section plus the `/model` new-conversation section. Its link at l.142 to `/chat#switch-the-model` must survive any heading rename.
- `omni-message` l.50 ("The model and the Workspace are fixed for the Session's lifetime"). `message-flow` l.132 (the `session_meta` row).
- `models` l.547-556 ("Models and agents": "stays fixed for that Session", plus the `/model` handoff).
- `server-api`: the `POST /switch-model` row next to `/compact` (l.717), its 409s next to l.728, and the `/model` paragraph at l.758.
- `cli`: l.91 (`--resume`) and the REPL table at l.95-104 (the two `/switch-model` rows).
- `web-app`: #758 moved the composer content to the new `chat.{en,zh}.md`, so this PR's web-app edits belong there.

These files do not conflict but are wrong once this PR is merged, because this PR never touched them:
- `chat.{en,zh}.md`: l.23 ("its model and Workspace are locked") and l.93 (the `/model` row).
- `chat.{en,zh}.md`, `## Switch the model` (l.165-177): it quotes the removed "Type /model to switch models" hint and the old "Switched model (was …)" banner, which this PR renames. The "How it works" bullet at l.373 says the same. The in-page anchors to that section (`#switch-the-model` and its zh counterpart, l.10, 23, 93) must follow any heading change; `packages/docs/test/content.test.ts` checks in-page and cross-page anchors.
- `agent-loop` (above).
- Typography: English text writes lowercase "agent" mid-sentence (this PR's docs say "the Agent's compaction mode" in several places), zh writes "Agent", with CJK–Latin spacing and full-width punctuation.

Source files merge cleanly, and main's changes there are orthogonal: `api/types.ts`, `repos/sessions.ts`, `routes/sessions.ts`, `session-manager.ts` (+ its test), web `endpoints.ts`, `chat-input.tsx`, `chat-page.tsx`, both dictionaries, and core `agent.test.ts`. What main changed there:
- `session-manager.ts`: #785's `unattended` approvals for org sessions, and #751's `client = "org"` stamp on org sub-sessions.
- `chat-input.tsx`: #784's `ReferenceChip`.
- Dictionaries: #755's id-suggestion strings.

Core compaction, rotation, `session_meta` and resume are unchanged on main since the merge base (core changed only the model catalog), so the engine half needs no merge work. #749 retired `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp`. None of this PR's tests use them (they use `deepseek-flash`, `deepseek-v4-pro`, `claude-sonnet-4-6`, `claude-opus-4-7`), and the new-Project default is still `deepseek/deepseek-flash`. A retired row a Project still carries stays a valid switch target, because validation reads the config and not the presets.

Open PRs:
- #791 (org approval sync) edits `chat-input.tsx` and `chat-page.tsx` but no switch code path.
- #786 (OpenCode Go, not merged) edits `models.{en,zh}.md`.

Whichever of these lands second rebases.

Verdict: fix before merge: [bug] summary dropped from the Trace on the discard path (context-engine.ts:1184); [bug] carried images reach a text-only target unfolded (context-engine.ts:1181); [bug] no summary-size guard on the just-compacted path (context-engine.ts:1175); [over-defensive] `switchRefusal` catch-all and message regexes (session-manager.ts:160); [copy] agent-loop docs; plus the smaller inline items (routes/sessions.ts:1310, chat-page.tsx:1574 and :1606) and the docs merge with main above.

packages/core/src/engine/context-engine.ts
hiyouga: Fixed in 53f6de3eb. The engine remembers the summary a context opened with (`contextSummary`, set where `runToCompletion` or `openContextFile` writes it, cleared at the first completed turn and at every new context); the discard branch writes it at the head of the next file right after the rotation and leaves the in-memory carry-over as it is. Regression tests in `core/test/model-switch.test.ts`, each with a turn on the new model and `resumeTrace` of its file: switching twice before typing; a failed first request on the new model, then a switch; compact, a failed request, then a switch. The narrower window you named — a restart between two switches with no completed turn in between — is accepted and stated as the only remaining summary gap in spec #170 (design 434f2b7) and in `sessions-and-traces`.

packages/core/src/engine/context-engine.ts
hiyouga: Fixed in 53f6de3eb. `openContextFile` folds the engine carry-over through `foldInputImages` when the opened context has no vision (`foldCarryOverForContext`), and `Session.switchModel` folds `carryOverInput` the same way after a completed switch. Tests: both variants from your probe — the stop during the first connect (Session-held input) and the stop before the request (engine-held input) — assert that no `image_url` reaches the text-only model and that the `[attached image: …]` line does.

packages/core/src/engine/context-engine.ts
hiyouga: Fixed in 53f6de3eb. The just-compacted branch compares the held summary with `target.summaryRoom` before rotating (`assertSummaryFits`) and refuses before any event with `ModelSwitchRefusedError("summary_too_large")`, the message naming both numbers. The server maps it to 409 `model_unavailable` (the target is configured but cannot be switched to), with core message passed through; none of the six codes fits this exactly, so that mapping is flagged for the user in the report. Test: "a held summary the target's window cannot take refuses a just-compacted switch before any event, and a roomier target then takes it" — no event, no rotation, no open, the Session still `just_compacted` on A.

packages/server/src/runtime/session-manager.ts
hiyouga: Fixed in 53f6de3eb. Core throws `ModelSwitchRefusedError` with a `reason` (`model_not_configured` from `modelEntryFromDisk`, `model_unavailable` from the client construction in `validate` and from a Session without switch support, `compaction_not_configured`, `summary_too_large`); `switchRefusal` maps exactly those and rethrows everything else, so an unparseable Agent State or a broken bootstrap reaches `handleError` and its logging. Both regexes are gone. The tests throw the typed error instead of copied sentences, and a new case asserts that a plain Error answers 500 `internal`, is logged, and leaves the row and the entry untouched.

packages/server/src/http/routes/sessions.ts
hiyouga: Fixed in 53f6de3eb: a row missing after the switch answers 404 `session_not_found`, the same error `resolveSession` throws.

packages/web/src/features/chat/chat-page.tsx
hiyouga: Fixed in 53f6de3eb. `switchContextShape` walks the transcript's trailing run of compaction rows and model-change markers the way `prefixCacheAtRisk` does and yields `compact` / `empty` / `compacted`. Right after a compaction the dialog body says the conversation continues on the target from the summary already held, the confirm reads "Switch" / "切换", and the toast says "Switching to …" instead of "Compacting the context on …". Tests for the shape helper and the copy in `web/test/model-switch.test.ts`.

packages/web/src/features/chat/chat-page.tsx
hiyouga: Fixed in 53f6de3eb. The pair is recorded as fetched only in `.then`; a failure is surfaced with `toastError(apiErrorText(e))` and retried at the next stream change; a separate in-flight ref keeps one fetch per pair.

hiyouga:
Review round addressed in 53f6de3eb (fixes), 322782de4 (a test fixture) and 426982363 (the merge of `origin/main`, the docs re-authored on main's text, the changelog pair). The two findings without a diff line: `agent-loop.{en,zh}.md` now says the model entry is per context (an ordinary rotation keeps it, a switch resolves the target from the Project config on disk) and lists the in-session switch under the `manual` reason; `chat.{en,zh}.md` describes both ways under the unchanged "Switch the model" heading, so the `/chat#switch-the-model` links stand. Spec #170 is amended (design 434f2b7): the remaining summary gap is only a restart between two switches with no completed turn in between. One item is left for the user: the 409 code for a held summary that does not fit — mapped to `model_unavailable` for now, see "Needs a decision" in the description.

packages/core/src/engine/context-engine.ts
hiyouga: Follow-up in b684e6a7c, per the user: the held-summary refusal now answers a dedicated 409 `summary_too_large` (server, API types, server-api docs, spec #170 design c21151f); the Web App and the CLI have their own copy for it ("pick a model with a larger context window"), and the `model_unavailable` copy is back to its one meaning.

PR #751, opened 2026-09-16
fix(server,web): company-mode sessions stay out of the development list

hiyouga:

## Summary

Company-mode trial round, item 2: running company-mode sessions sometimes popped up in the development-mode Session list.

Root causes:
1. **Totals counted rows the web never loads.** `listSessions` counted every row of an Agent — desk and ticket sessions included — into `counts` / `workspaceCounts` / `workspaceLatest`; the web could only subtract the org rows it had loaded (10 per Agent). A busy employee's remaining rows kept a phantom workspace group alive, and the just-opened ticket session's `createdAt` sorted it to the top — exactly when a new org session opened.
2. **Subagent children of org sessions were unmarked.** `registerChildSession` inserted the child with `client` NULL, so it listed under Subagents with a live running mark.

Fix:
- `GET` session list gains opt-in `excludeOrg=1`, applied server-side before counting and paging (page, totals, breakdown, stamps and "has more" agree); the web's dev list uses it; the client-side totals correction is gone (a render-time guard stays for deep-linked rows).
- A child registered under a `client: "org"` parent inherits the stamp.
- Company surfaces take live run status from every `session_state` event, independent of the dev list; desk rows take their messaging-bot mark from the organization's own session data.
- The store carries the organization rows it holds (a desk or ticket session the chat page `add()`ed) across reloads, since no `excludeOrg` fetch can return them. The open desk's approval-mode and thinking-level changes, titles and statuses keep reaching its row.
- `resync_required` clears the remembered live statuses, and the company store re-reads its sessions route, organization list and open chart (a resync fan-out that bumps the `runs` and `orgs` versions). A held organization row no longer overrides the remembered statuses in `liveSessionStatuses`: no fetch refreshes it, so after a resync its stale status would keep the desk running the same way.
- The list mark and the desk mark read one guard, `isMessagingChannel`, built from `{ feishu, telegram, qq, wechat } satisfies Record<MessagingChannel, true>`, so a channel missing from it fails typecheck (checked by deleting `wechat`: TS1360). The messaging routes' `specOf` uses the guard instead of a cast, and `runtime/messaging/connector.ts` re-exports `MessagingChannel` from `api/types.ts` instead of redeclaring it.

Existing unstamped subagent rows created before this fix are left as they are (user decision).

## Merge with main

#759 (WeChat sidebar mark) conflicted in `packages/server/src/runtime/session-manager.ts`. Resolved toward this branch's `enabledMessagingChannel(...)` call, with #759's restored comment dropped; WeChat is covered by the guard above. The per-channel list-mark tests (Feishu, Telegram and #759's WeChat test) are folded into one `it.each` table over all four channels in `messaging.test.ts`, QQ included. The merge commit is 1b5d8446a, and the review fixes are in c8f150226.

## Verification
- vps3, on the final tree:
  - server: `session-index`, `session-manager`, `organizations-routes`, `organization-runtime`, `messaging`, `messaging-telegram`, `messaging-wechat`, `messaging-qq`: 8 files, 408 tests passed.
  - web: `sessions-own-rows`, `org-sessions`, `company-store`, `company-beta`, `company-click-targets`, `company-mode-write`, `company-nav`, `org-chart-tree`, `org-draft`, `org-examples`, `org-trigger`, `shell-org-status`, `session-status-events`, `install-scope`, `control-size`: 15 files, 178 tests passed.
  - `pnpm typecheck` (every package) and `pnpm lint` (0 warnings, 0 errors).
- The new tests fail without the fixes (vps3, with the pre-fix sources swapped in): the table's `wechat` row fails and the other three rows pass; the three new store tests fail (reload drops the desk, the resync leaves the remembered status, the held desk row still wins).
- Local: `prettier --check` on the touched files and `git diff --check`.
- Earlier rounds: an integration tree merging this branch with the other five branches of the batch and #748 (`integ/r14`, local only) was built, typechecked, linted and format-checked, and its full package suites ran on vps3: core 1250 passed / 5 skipped, server 2068, web 2417 of 2418 (the one failure was the temporary-session storage key missing from install-scope, fixed on #752), cli 466, docs 53, landing 71.


packages/web/src/state/sessions.tsx
```diff
@@ -290,6 +344,7 @@ export function createSessionsStore() {
                       offset: 0,
                       limit: SIDEBAR_PAGE_SIZE + 1,
                       category,
+                      excludeOrg: true,
```
hiyouga, line 345: **[bug]** With `excludeOrg`, `reload()` can never return an organization row. It also replaces `sessions` wholesale (`set({ sessions: nextSessions, ... })`), so the row the chat page's deep-link probe `add()`ed for an open desk or ticket session is gone after the next reload. Nothing adds it back: `resolveRoutedSession` falls back to the page's `fetchedSession`, `selected` stays non-null, and the probe never runs again.

In company mode that reload comes quickly: every `org_run` for the Project, every `schedule_fired`, and the chat page's own `reloadSessions()` when the open session spawns a subagent (`session_created`).

Scenario: open a desk from the 工位 group. The desk spawns a subagent, or any `org_run` arrives. Then change the approval mode or pin a thinking level in the chat page. The PATCH succeeds, but `replace(res.session)` finds no row and does nothing. The picker (`approvalMode={selected.approvalMode}`, `selected?.thinkingLevel`) snaps back to the probe-time value, and `setTitle` / `setStatus` for that session no longer reach `selected`. On main this only happened to a desk past its Agent's first page; with this PR it happens to every organization session.

Fix: carry the organization rows the store already holds through `reload()`, since no fetch can return or refresh them. For example, build the final list as `[...nextSessions, ...get().sessions.filter(isOrgSession)]` at the `set`. They are never counted (`adjustCount` skips them), and the sidebar filters them out at render. Add a case to `sessions-own-rows.test.ts`: `add(desk)`, `reload()`, and expect the desk to still be held.

packages/web/src/state/sessions.tsx
```diff
@@ -177,6 +188,15 @@ interface SessionsStoreState {
...
   workspaceLatestByAgent: ReadonlyMap<string, Readonly<Record<string, string>>>;
+  /**
+   * Run status by Session id as the user channel last reported it, for every `session_state`
+   * received — the rows this store holds AND the ones it never fetches (an organization's
+   * desks and ticket sessions, another category's unopened folder). The user channel is the
+   * only source that reports a run ENDING, so company mode's surfaces read this for the
+   * Sessions the development list deliberately leaves out. Not reset on a Project switch:
+   * ids are globally unique and a status is a fact about the Session, not about the list.
+   */
+  liveStatuses: ReadonlyMap<string, SessionStatus>;
```
hiyouga, line 201: **[bug]** Nothing ever invalidates this map, and company mode lets it win over a fresh snapshot (`deskRows`: `live?.get(d.sessionId) ?? d.status`; `employeeLiveState` does the same). If the user channel loses an event, the entry stays wrong.

Scenario: a desk is running when the server restarts, or the tab reconnects after the replay buffer has moved on. `sse.ts:72` sends `resync_required`, and `applyUserEvent` (line 716) only calls `reload()`. That refreshes rows but not `liveStatuses`, and no `idle` for the desk will ever come. The desk row keeps its hourglass, the org chart keeps the running dot, and the overview keeps counting the employee as running. `reloadOrgSessions()` cannot fix it either, because the stale live entry still beats the fresh `status` it returns. On main the desk's status lived on its list row, and the resync `reload()` refreshed it.

Fix: in the `resync_required` branch, also `set({ liveStatuses: new Map() })`, and have the company store re-read `orgSessions` and the open chart on that event, so the surfaces fall back to current snapshots. Add a store test: remember `running`, apply `resync_required`, and expect the entry to be gone.

packages/server/src/runtime/messaging/enabled-channel.ts
```diff
@@ -0,0 +1,20 @@
...
+import type { MessagingChannel } from "../../api/types.js";
+import type { MessagingBindings } from "../../mechanisms/messaging.js";
+
+export function enabledMessagingChannel(
+  bindings: Pick<MessagingBindings, "findEnabled">,
+  sessionId: string,
+): MessagingChannel | null {
+  const enabled = bindings.findEnabled(sessionId);
+  return enabled !== null &&
+    (enabled.channel === "feishu" || enabled.channel === "telegram" || enabled.channel === "qq")
```
hiyouga: **[bug]** This PR conflicts with main. #759 (merged) put `wechat` back into the `session-manager.ts` lambda this PR replaces, and `git merge-tree` stops on that file; it is the only conflicting file. If the conflict is resolved by taking this side, WeChat loses its mark again, on the development list and now on desk rows too, and #759's `messaging-wechat.test.ts:484` fails.

Required resolution (agreed on #759's review):
- `session-manager.ts`: keep `messagingChannel: (sessionId) => enabledMessagingChannel(this.messagingRepo, sessionId)` and drop the comment #759 restored, since this helper's doc covers it.
- Here: replace the `===` chain with one channel-id guard, so a missing channel becomes a type error. For example, beside `MessagingChannel`:
  ```ts
  const MESSAGING_CHANNEL_IDS = { feishu: true, telegram: true, qq: true, wechat: true } as const satisfies Record<MessagingChannel, true>;
  export const isMessagingChannel = (c: string): c is MessagingChannel => Object.hasOwn(MESSAGING_CHANNEL_IDS, c);
  ```
  The body then becomes `enabled !== null && isMessagingChannel(enabled.channel) ? enabled.channel : null`. The last sentence of the header ("a discriminator this reading does not name") then reads "a stored channel id that is not a `MessagingChannel`".
- Tests: fold the per-channel list-mark tests (`messaging.test.ts:2037` for Feishu, `messaging-telegram.test.ts:1914`, `messaging-wechat.test.ts:484`; QQ has none) into one `it.each(["feishu", "telegram", "qq", "wechat"])` table.

hiyouga (commented):
Checked: `excludeOrg` is applied to the row map before hydration, classification, `counts` / `workspaceCounts` / `workspaceLatest` and both paging paths (the no-classification slice included); both web fetch paths (`reload`, `loadMoreFor`, workspace-scoped pairs included) pass it, and no other web caller of the list route needs it. The child stamp is read from the driven entry's row, so nested and panel-driven sub-sessions inherit it too. `adjustCount` skips org rows on `add` / `remove` / `replace`. The desk `messagingChannel` wiring goes through the real bindings table and has a route test, and design PR #173 describes the same contract.

Findings, ranked:
1. **[bug]** `web/src/state/sessions.tsx:347`: `reload()` can no longer return an org row and replaces `sessions` wholesale, so the deep-linked desk or ticket row the chat page `add()`ed is dropped on the next reload (any `org_run`, or the open desk spawning a subagent). The chat page then keeps its probe-time `fetchedSession`, and approval-mode and thinking-level changes snap back after a successful PATCH. Fix: carry held org rows forward in `reload()`.
2. **[bug]** `web/src/state/sessions.tsx:199`: nothing invalidates `liveStatuses`, and company surfaces let it win over fresh snapshots. After a server restart, or a reconnect past the replay buffer (`resync_required`), a desk that was running keeps its hourglass, its running dot and its place in the running count indefinitely. Fix: clear the map on `resync_required` and re-read the company snapshots.
3. **[bug]** `server/src/runtime/messaging/enabled-channel.ts:17`: **the PR conflicts with main** in `packages/server/src/runtime/session-manager.ts` (the only conflicting file) because of #759. Taking this side drops WeChat from both the list mark and the desk mark, and fails #759's test. Required resolution: `enabledMessagingChannel` covers `wechat` through one `Record<MessagingChannel, true>`-backed guard, and the per-channel list-mark tests fold into one `it.each` table over all four channels.
4. **[over-defensive]** `server/src/runtime/organization/deps.ts:153`: `messagingChannel?` is optional only for two test literals. Make it required and delete the `?.() ?? null` at service.ts:2463.
5. **[copy]** `web/src/features/company/org-sessions.ts:19` (and lines 6, 31-34, 137): the doc comments still say live status comes from rows the session list holds.
6. **[simplify]** `server/src/services/session-service.ts:264`: needless `[...rows.values()]` copy, and `isOrganizationRow` is exported for a single call site. Inline both.
7. **[over-defensive]** `web/src/state/sessions.tsx:249`: the `if (oldest.done) break;` check can never fire.

Verdict: fix before merge: 1, 2, 3 (3 is required anyway to resolve the conflict)

packages/server/src/runtime/organization/deps.ts
```diff
@@ -139,6 +144,13 @@ export interface OrgDeps {
    */
   completeOnce?: (projectId: string, prompt: string) => Promise<UtilityCompletion>;
   usage: OrgUsageGateway;
+  /**
+   * The channel of a Session's ENABLED messaging binding, or null when none is enabled — the
+   * same reading behind `SessionInfo.messagingChannel`, so the company sidebar's desk row can
+   * carry the development row's mark without the development list holding the desk. Optional
+   * so a test binds nothing, and then no desk carries the field.
+   */
+  messagingChannel?: (sessionId: string) => MessagingChannel | null;
```
hiyouga: **[over-defensive]** Production always binds this (service.ts:2824). It is optional only so the two test literals (`test/org-harness.ts:131`, `test/organization-runtime.test.ts`) can leave it out, and service.ts:2463 pays for that with `?.() ?? null`. The risk is that a wiring which forgets it still type-checks and silently draws no desk marks, which is how #599 lost WeChat. Make it required, give the two test deps `messagingChannel: () => null`, and call `this.deps.messagingChannel(desk.sessionId)` directly.

packages/web/src/features/company/org-sessions.ts
```diff
@@ -19,6 +19,7 @@
  * snapshots stand in for the rows that list has not loaded.
```
hiyouga: **[copy]** This file's comments still describe the old source of live status, and after this PR they are wrong:
- this line: "the snapshots stand in for the rows that list has not loaded"
- line 6: the state is "the live one where the session list holds the row"
- `LiveSessionStatuses` (lines 31-34): "A Session the list has not loaded is simply absent"
- `employeeLiveState` (line 137): "by the live status where the list holds it"

The list never holds an organization row now. The live map is every `session_state` the user channel has reported (`liveStatuses`), and the snapshot is only used for a Session no event has named. Reword those four places.

packages/server/src/services/session-service.ts
```diff
@@ -220,21 +244,27 @@ export class SessionService {
...
+    const { paging, category, workspaceGroup, withCounts, excludeOrg } = opts;
     const rows = new Map(
       this.deps.sessions.listByAgent(projectId, agentId).map((r) => [r.sessionId, r]),
     );
     // One query for the whole Project's organization-owned sessions, looked up per row
     // below: the company caches are small, and a lookup per row would put a statement
     // behind every entry of a long sidebar list.
     const orgIds = this.deps.orgIdsOfProject?.(projectId) ?? EMPTY_ORG_IDS;
+    if (excludeOrg) {
+      for (const row of [...rows.values()]) {
```
hiyouga: **[simplify]** `[...rows.values()]` copies the whole map only so rows can be deleted from it, but deleting the current key while iterating a `Map` is safe. `isOrganizationRow` (line 112) is exported for this one call site, and its doc is longer than its body. Inline both: `for (const [id, row] of rows) if (row.client === "org" || orgIds.has(id)) rows.delete(id);`, and move the useful sentence of the doc onto the loop.

packages/web/src/state/sessions.tsx
```diff
@@ -206,6 +226,32 @@ interface SessionsStoreState {
...
+  sessionId: string,
+  status: SessionStatus,
+): ReadonlyMap<string, SessionStatus> {
+  if (live.get(sessionId) === status) return live;
+  const next = new Map(live);
+  next.delete(sessionId);
+  next.set(sessionId, status);
+  while (next.size > LIVE_STATUS_MAX) {
+    const oldest = next.keys().next();
+    if (oldest.done) break;
```
hiyouga: **[over-defensive]** This branch is dead. The loop only runs while `next.size > LIVE_STATUS_MAX` (1000), so `keys().next()` always returns a key. Delete the check, or write `next.delete(next.keys().next().value!)`. The same dead check exists in the `DELETED_IDS_MAX` loop at line 532, which this was copied from.

packages/server/src/runtime/organization/deps.ts
hiyouga: Fixed in c8f150226: `messagingChannel` is required, both test deps bind `() => null`, and `service.ts` calls `this.deps.messagingChannel(desk.sessionId)` directly.

packages/web/src/features/company/org-sessions.ts
hiyouga: Fixed in c8f150226: all four comments now say live status is every `session_state` the user channel has reported, with the snapshot standing in for a Session no event has named.

packages/server/src/services/session-service.ts
hiyouga: Fixed in c8f150226: `isOrganizationRow` is inlined as `for (const [id, row] of rows) if (row.client === "org" || orgIds.has(id)) rows.delete(id);`, with its useful sentence moved onto the loop.

packages/web/src/state/sessions.tsx
hiyouga: Fixed in c8f150226: the loop is now `while (next.size > LIVE_STATUS_MAX) next.delete(next.keys().next().value!);` (the older `DELETED_IDS_MAX` loop is outside this PR and left as it was).

packages/server/src/runtime/messaging/enabled-channel.ts
hiyouga: Resolved in merge 1b5d8446a: `session-manager.ts` keeps `enabledMessagingChannel` (#759's comment dropped), one `isMessagingChannel` guard built from `{ feishu, telegram, qq, wechat } satisfies Record<MessagingChannel, true>` backs it and `specOf`, `connector.ts` re-exports `MessagingChannel`, and the three list-mark tests are now one `it.each` over all four channels in `messaging.test.ts`.

packages/web/src/state/sessions.tsx
hiyouga: Fixed in c8f150226: `reload()` now keeps the organization rows the store holds (`[...nextSessions, ...held]`, deduped by id), and `sessions-own-rows.test.ts` adds a desk, reloads, and expects it still held.

packages/web/src/state/sessions.tsx
hiyouga: Fixed in c8f150226: `resync_required` clears `liveStatuses` and fires a company resync that bumps `runs`/`orgs` (re-reading `orgSessions`, the org list and the open chart); `liveSessionStatuses` also stops letting a held org row override the map, since its stale status would otherwise pin the desk after a resync. Tests are in `sessions-own-rows` and `company-store`.

PR #752, opened 2026-09-16
feat(web): ticket cards open on click and move by dragging; opened ticket sessions stay in a Temporary sidebar group

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/173 (merged; the 临时 clause is commit 64d0d11, the drag rule f501305). Follow-ups, open: [#198](https://github.com/Prism-Shadow/penguin-harness-design/pull/198) amends the 06 临时 clause for the user's 2026-09-19 decision (no cap, 全部关闭); [#189](https://github.com/Prism-Shadow/penguin-harness-design/pull/189) brings the 05 and 08 mentions of the group in line.

## Summary

Company-mode trial round, items 3, 4 and 5.

- **Ticket cards:** a click (or Enter/Space) anywhere on the card opens the ticket dialog. The card is the one declared exception to the "no whole-area click targets" rule. **Dragging a card moves it:** a mouse or pen press that moves more than 6 px lifts the card at once, with no hold and no timer. The ghost follows the pointer, and a drop on another column goes through the existing move confirmation. A press released within 6 px is a click. Only a gesture that lifted the card swallows the click that follows (a 600 ms window that the next press resets), so a drag never also opens the dialog, while Enter, Space and screen-reader activations are never eaten. Escape puts a lifted card back. HTML5 drag-and-drop was replaced by pointer events: `ticket-press.ts` is the pure state machine and `ticket-drag.ts` binds it to the DOM.
- **Touch keeps the long press** (350 ms, moving no more than 6 px). On a phone the board scrolls sideways and is almost all cards, so starting a drag on the first movement would leave nothing to scroll the board or the page with. A finger that moves before the hold completes gives the press up and scrolls; a tap still opens the card. A pen follows the mouse rule.
- **Ticket dialog:** the parent ticket, child tickets and ticket sessions open from their titles; the jump buttons are gone.
- **Ticket sessions open.** Root cause of "clicking does nothing": the dialog closed (a sync store update) and then navigated. React Router applies the location in a transition, so the board's `?ticket=` sync effect ran first and `replace`d the new `/chat/<id>` entry with the board. The effect now stands down once the pathname is no longer the board's.
- **A 「临时」 group under the desks.** The conversation opens as the full chat page, and the session joins a collapsible **Temporary** group in the company sidebar, **below** 工位, drawn with the same group header and collapse behaviour as the desks and shown only while it lists something. A row carries the employee's avatar, the session title, its run mark and a ✕, and the group's header carries 「全部关闭」 / "Close all"; the collapsed rail draws the entries as avatars with their running dots, after the desks. Ticket sessions are still never listed wholesale.

## Revision 1: move without a long press

Following user feedback, dragging a card now enters move mode right away on the desktop board. The machine's `void` phase ("a quick drag does nothing") and the hold timer for mouse and pen are gone; the timer remains only for touch. `company.tickets.dragHint` is back to 「拖到另一列即可移动」 / "Drag to another column to move it". It is a `title` tooltip, which touch screens never show, so the touch rule went into the board's "?" (`company.tickets.info`), the company mode guide (zh/en) and the changelog pair instead.

## Revision 2: a Temporary group, kept until dismissed

The first cut kept **one** row above the desks for the conversation on screen, with a 「返回工单」 link, and dropped it as soon as the reader went anywhere else. Following user feedback it became a list of its own:

- The group sits **below** 工位, and holds every ticket session opened from a ticket dialog, newest first. Opening one already listed moves it to the top.
- An entry stays until its ✕ removes it. Navigating away keeps it, a reload keeps it, and a ✕ pressed while the reader is on that very session removes only the row and stays on the page — there is no "back" any more. **`company.sessionList.backToTicket` and the return-path logic are gone**, in both dictionaries; `closeTemporary` now reads 「从临时中移除」 / "Remove from Temporary".
- The list lives in `localStorage` under `penguin.orgTempSessions.<userId>.<projectId>.<orgId>` — install-scoped in `install-scope.ts`, since it names Sessions and Agents of this root. (Revision 2 capped it at 20 entries; Revision 3 removes the cap.)
- A desk session opened from a ticket is never listed: its desk row already names it. It is left out when the entry is recorded and filtered again when the rows are drawn.
- The old `penguin.company.tempSession` record was `sessionStorage`-only and unreleased, so it is renamed with **no compatibility path** and its classification is gone.
- `temp-session.ts` is pure list logic (`withOpened` / `withDismissed` / `tempSessionRows`) plus a store that takes its storage as an argument; the app's instance watches `storage` events, so a tab still holding an entry another tab dismissed cannot write it back.

## Revision 3: review fixes and the user's decisions (2026-09-19)

Review: https://github.com/Prism-Shadow/penguin-harness/pull/752#pullrequestreview-5254962785 (fix before merge, findings 1–4). The user decided the two open questions it raised.

- **No cap** (user decision, 755a9f9cc). An entry stays until the user closes it: the 20-entry limit, which silently dropped the oldest entry when a 21st session was opened, is gone.
- **「全部关闭」 / "Close all"** (user decision, 755a9f9cc). A flat text action at the right end of the 临时 header empties the group in one click, with no confirmation — it only takes entries off the list, and every session is still one click away in its ticket. Nothing navigates. It exists only while the group does. `FolderSection` gained an optional `action`, drawn outside its toggle so a click on it never folds the folder.
- **Touch** keeps the 350 ms hold: that question is still open with the user.
- **Finding 1, [test]** (72d15d6ff). `e2e/company.spec.mjs` now clicks a card, drags one onto 进行中 and confirms the move, opens a ticket session from the dialog and checks the URL stays on it, and walks the 临时 group: listed below 工位, kept across a reload, a ✕ on the session on screen removing only its row, and 全部关闭 emptying it for good.
- **Finding 2, [bug]** (f1d98f86a). A session the ticket still names but whose record is gone comes back with an empty `agentId`; its title is now plain text, so no unopenable Temporary entry can be recorded.
- **Finding 3, [over-defensive]** (4bd80f2fc). The parser keeps the `JSON.parse` try/catch and reads the rest back as written — its only writer is the store. The two tests that pinned hand-edited lists are gone.
- **Finding 4, [copy]** (35a0996b4). zh `从「临时」中移除`. en stays "Remove from Temporary": the en dictionary cites a group capitalised, without quotes.

## Merge with main

`origin/main` is merged in twice (no rebase); the second merge brought only #766 (blog posts) and did not conflict. Two landings of the first met this branch:

- **#758** rewrote the company mode guide page by page, and its board paragraph went with it. The Temporary group and the board's click and drag are restated as a **Tickets** bullet in the new page-by-page list, in both languages (the only docs conflict).
- **#751** took the development session list out of the company sidebar. The Temporary rows therefore read their titles from the organization's own sessions route — new pure helper `ticketSessionTitles` in `org-sessions.ts`, tested in `org-sessions.test.ts` — instead of from the list store, which no longer holds a ticket session to ask. Run marks still come from `useLiveSessionStatuses`, which #751 made independent of that list.

## Verification

- vps3, on the final tree (1b0784229): `pnpm typecheck` (every package), `pnpm lint` (0 warnings, 0 errors), `pnpm format:check`, the web suites this diff touches — 22 files, 222 tests passed: `temp-session` 17, `org-sessions` 19, `group-list` 7, `install-scope` 27, `ticket-press` 19, every `company-*` / `ticket-*` / `org-*` suite, `sessions-own-rows`, `control-size`, `icon-scale`, `disclosure-anchor` — and the docs suite, 8 files, 62 tests passed.
- vps3, the company e2e (`e2e/company.spec.mjs`, run alone the way `e2e/run.sh` runs the specs: built web dist, the mock LLM, a scratch data root): 1 passed (13.5 s). **Negative control:** with the board's history guard (`tickets-page.tsx`, `if (window.location.pathname !== routedPath) return;`) removed, the spec fails at the session-open step — `toHaveURL(/\/chat\/session-…$/)` received `…/marketplace/tickets`.
- The `ticket-press` suite, run against the previous long-press machine, fails 9 of 19: the drag-on-move, drag-only click suppression and touch give-up cases. So it pins the new rules rather than passing on either machine.
- `temp-session.test.ts` covers add, move-to-top, dismiss, no cap (50 entries kept), Close all (every entry, the stored key removed, one notification, a no-op when empty, other organizations untouched), a desk session left out, per-user × Project × organization scoping, a simulated reload through a second store over the same storage, and rows recomputed from anywhere (only the on-screen mark moves).
- Not verified: how the touch long press and a pen on a touch screen feel on a real device, the drag ghost's look, and edge scrolling. The mouse drag, the session open and the 临时 group now run in the e2e above.


packages/web/src/features/company/tickets-page.tsx
```diff
@@ -173,7 +180,14 @@ export function TicketsPage() {
...
   const syncedTicket = useRef<string | null>(null);
   const { openTicket: openTicketInShell } = company;
+  const routedPath = location.pathname;
   useEffect(() => {
+    // The browser has already left this page: a click that closes the dialog and navigates
+    // (opening a ticket session) is rendered first with the dialog closed and this page still
+    // mounted, since the router commits its new location in a transition. Writing the query now
+    // would resolve against this page's stale location and replace the new history entry with
+    // the board — the navigation would silently land back here.
+    if (window.location.pathname !== routedPath) return;
```
hiyouga, line 190: **[test]** This one-line guard is the fix for the reported bug, "clicking a ticket session does nothing".
- **The cause:** this effect's `setParams(…, { replace: true })` resolved against the board's stale location and replaced the new `/chat/<id>` history entry.
- **What pins it:** nothing. No unit test reaches this effect, and the PR description says the drag, the session open and the Temporary group never ran in a browser.
- **Where to test it:** `e2e/company.spec.mjs` already files a ticket and opens the board (L166-179). Extend it:
  - a click on the card opens the dialog;
  - a mouse drag onto 进行中 (In progress) opens the move confirmation;
  - start a session for the ticket (`POST …/organizations/:orgId/tickets/:ticketId/start`) and click its title in the dialog. The URL should stay `/chat/<id>`; this step fails without this line.
  - 临时 then lists the session below 工位, a reload keeps it, and ✕ removes only the row.

packages/web/src/features/company/ticket-dialog.tsx
```diff
@@ -382,11 +390,25 @@ function TicketDialog({
...
   /**
-   * One field of the summary grid. Every row is exactly one line tall, so labels and values
-   * line up down the column: anything taller than the line — the parent's jump button is the
-   * only such thing — pulls its own box back in with a negative margin rather than growing
-   * the row and pushing its label off the others' baseline.
+   * Opens a ticket session: the one way out of the dialog. The session goes to the top of the
+   * company sidebar's Temporary group (a desk session keeps its desk row instead) and the
+   * conversation opens as its full page. The current Agent follows it, as a desk row's does.
    */
+  const openSession = (session: { sessionId: string; agentId: string; title?: string }) => {
```
hiyouga, line 397: **[bug]** (low) A ticket's frontmatter can still list a Session whose row is gone, for example after its Agent was deleted. `ticketDetail` then returns that session with `agentId: ""` and no title (`packages/server/src/runtime/organization/service.ts:1238-1246`).
- **What happens:** clicking that title records a Temporary entry with a `?` avatar and the title 「未命名会话」.
- **Why it matters:** the entry stays until ✕. Every click on it lands on `/chat/<id>`, which the chat page replaces with the latest development conversation (or a draft).
- **Fix:** when `agentId === ""`, render the dialog row as plain text instead of a `TitleButton`. There is nothing to open, so no entry should be recorded.

packages/web/src/features/company/temp-session.ts
```diff
@@ -0,0 +1,272 @@
...
+    .filter((e) => !desks.has(e.sessionId))
+    .map((e) => ({ ...e, active: e.sessionId === activeSessionId }));
+}
+
+/**
+ * A stored list, entry by entry. An entry that is not one (a truncated write, a hand edit) is
+ * skipped rather than failing the list; a repeated session keeps its newest place, and the cap
+ * holds even for a list written by hand.
+ */
+export function parseTempSessions(raw: string | null): TempSessionEntry[] {
```
hiyouga, line 100: **[over-defensive]** This parser re-validates data that only this module writes.
- **Why the checks are dead:** the only writer of this key is `write()` below. It always stores a deduplicated list of at most `MAX_TEMP_SESSIONS` well-formed entries. `localStorage.setItem` is atomic, so the "truncated write" this comment guards against cannot happen.
- **What to delete:** the per-entry shape checks, the `seen` dedupe and the cap.
- **What to keep:** the `JSON.parse` try/catch, which is the repo's localStorage convention (`page-hints.ts`).
- **Result:** after the parse, the rest of the function becomes `return Array.isArray(parsed) ? (parsed as TempSessionEntry[]) : [];`. Drop the two tests that pin the hand-edit handling (`temp-session.test.ts:115-136`).

packages/web/src/lib/strings.ts
```diff
@@ -3547,6 +3547,11 @@ Benchmark：
       noEmployees: "这个组织还没有员工",
       untitledSession: "未命名会话",
       loadFailed: "员工列表加载失败",
+      /** The group below the desks listing the ticket sessions opened from a ticket, each kept until removed. */
+      temporary: (n: number): string => `临时（${n}）`,
+      /** A Temporary entry's name on the collapsed rail, where no group header says where it belongs. */
+      temporaryEntry: (title: string): string => `临时 · ${title}`,
+      closeTemporary: "从临时中移除",
```
hiyouga: **[copy]** In `从临时中移除`, the unquoted 临时 reads as the adverb "temporarily". The tooltip then reads roughly "remove from … temporarily".
- **Convention:** this dictionary quotes a group name wherever it cites one. `channels.archiveConfirm`, for example, writes `…并折叠进「已归档」…`.
- **Fix:** `从「临时」中移除`.

hiyouga (commented):
## Review of #752 (head 865cfc053)

This PR does four things:
- **Ticket cards.** The whole card becomes one button: a click, Enter or Space opens the dialog. Dragging moves the card, and HTML5 drag-and-drop is replaced by pointer events (`ticket-press.ts` is the state machine, `ticket-drag.ts` the DOM binding).
- **The ticket dialog.** The parent, child and session titles become the click targets.
- **Opening a ticket session.** A click on a session no longer bounces back to the board.
- **A 「临时」 (Temporary) sidebar group.** Opened ticket sessions are listed there, kept in localStorage.

### What you asked for vs. what the PR does

| You asked | This PR |
|---|---|
| A click anywhere on a card opens the ticket dialog | **Done.** The card is one `<button>`, so a click, Enter or Space opens the dialog. |
| Dragging moves a card between states, starting immediately, with no long press | **Mouse and pen: done.** The card lifts once the pointer has moved 6 px, with no timer, and a drop goes through the existing move confirmation.<br>**Touch: not done.** A finger has to hold still for 350 ms before the card lifts (`ticket-press.ts:29`, `:146-152`). This keeps sideways scrolling of the board working on a phone. Design #173 (commit f501305) records this exception, but it contradicts "no long press".<br>**Your call:** keep it, or make touch drags start immediately. That would give up scrolling the board by swiping on a card. |
| Sub-tickets and ticket sessions open by clicking their title | **Done**, and the parent ticket works the same way. The three jump buttons are gone. |
| A ticket session opens as a temporary conversation in a collapsible 「临时」 section below 工位 | **Done.** The session opens as the full chat page. The group sits below 工位 (Desks), collapses, starts expanded and appears only when it has entries. The collapsed rail shows the entries' avatars after the desks. |
| No 「返回工单」 button | **Done.** The button, `backToTicket` and the return-path logic are gone from the code and from both dictionaries. |
| An entry stays until × closes it; nothing else removes it | **Done, with one exception.** Navigating away or reloading keeps an entry. ✕ removes only its row, even while that session is on screen.<br>**The exception:** the list holds at most 20 entries. Opening a 21st session silently drops the oldest (`temp-session.ts:44`, `:75-78`). Design #173 specifies this cap, but it breaks "nothing else removes it".<br>**Your call.** |

### Conflicts with main
- **Against main:** `git merge-tree --write-tree origin/main origin/feat/company-tickets-click` (main 078a7aa83) finds no conflicts.
- **Newer main commits:** the branch base is de6567cbf, so it already contains #751, #754, #755, #785 and #782–#784. Only #766 (blog posts) is newer.
  - #751 is handled: the Temporary rows read their titles from the organization's own sessions route (`ticketSessionTitles`), and their run marks from `useLiveSessionStatuses`.
  - #754, #755, #785 and #782–#784 touch nothing this PR uses.
- **Together with #753:** it merges cleanly in either order. `shared.tsx`, `ticket-dialog.tsx`, `strings.ts` and `strings-en.ts` auto-merge, and the merged dialog has no stale imports (`JumpButton`, `Md`).
- **vps3, on the three-way merge (main 078a7aa83 + #752 + #753):** all four checks pass. (1) `pnpm --filter @prismshadow/penguin-web typecheck`. (2) The 11 web suites the two PRs touch, 176 tests: ticket-press, temp-session, org-sessions, install-scope, company-click-targets, mention-draft, path-capsules, channel-mentions, control-size, icon-scale, disclosure-anchor. (3) `pnpm lint`: 0 warnings, 0 errors. (4) `pnpm format:check`: clean.

### Drag and click details
- **Mouse:**
  - A press released within 6 px is a click. Past 6 px the card lifts, and a column lights up under the pointer only when `canMove` allows the move.
  - Escape puts the card back, and losing window focus cancels the drag.
  - A drag never also opens the dialog. The guard that swallows the click after a drag expires after 600 ms and is cleared by the next press, so a later Enter, Space or screen-reader activation still opens the card.
- **Pen:** it follows the mouse rule.
  - On Android and iPad a stylus also fires touch events. Once the card has lifted, the board's non-passive `touchmove` handler therefore stops the page from scrolling.
  - Whether a pen on a touch screen drags the card or pans the page was not checked in a browser. The PR says so.
- **Touch:**
  - A tap opens the card.
  - A finger that moves before 350 ms scrolls the board or the page. Nothing sets `touch-action: none`, and `pointercancel` ends the press.
  - After the hold, the finger drags the card, and `onContextMenu` blocks the long-press menu.
- **Keyboard:** Enter or Space opens a card. There is no keyboard drag. The dialog's move control is the way to move a ticket without dragging, as the design says.

### Temporary list storage
- **Key:** `penguin.orgTempSessions.<userId>.<projectId>.<orgId>` in localStorage, classified as install-scope.
- **Reload:** the list survives it. The unit tests cover this with a second store over the same storage.
- **Other tabs:** a ✕ in another tab is re-read through the `storage` event.
- **A session deleted after it was listed** (its Agent deleted, or the session itself deleted): the row stays, with its stored title and avatar, because only ✕ removes a row. A click sends you to `/chat/<id>`. The chat page's lookup fails, and it replaces the route with the latest development conversation or a draft, with no message saying why. That is the chat page's existing behaviour for any dead link.
- **An archived session:** the row stays and opens normally.
- **A session that was already gone when opened from the ticket:** see finding 2.
- **Desk sessions:** a desk session opened from a ticket is left out when it is recorded and again when the rows are drawn.

### Other checks
- **i18n:** `temporary`, `temporaryEntry`, `closeTemporary` and the rewritten board "?" text exist in both dictionaries, in the same shape. For zh typography, see finding 4.
- **Changelog:** the pair mirrors section for section, in the past tense. The metadata block is in the fixed order and gives the full PR URL. The company mode guide gains a Tickets bullet in both languages.
- **Design #173:** these clauses match the code:
  - the board clause;
  - the dialog clauses: 点标题打开, no jump buttons;
  - the 临时 clause: placement, expanded by default, ✕ behaviour, per user × Project × organization, the cap of 20, no back button, desk sessions left out;
  - the rail clause.
- **Tests:**
  - `ticket-press` (19 tests) pins the state machine: click vs. drag, the touch give-up and the click-guard window.
  - `temp-session` (18 tests) pins the list rules and the reload.
  - The DOM binding and the navigation fix have no test (finding 1).

### Findings (inline)
1. **[test]** (medium) `tickets-page.tsx:190`: the navigation fix and the gesture have never run in a browser, and nothing pins them.
2. **[bug]** (low) `ticket-dialog.tsx:397`: a session the server no longer has gets recorded as a Temporary entry that cannot be opened.
3. **[over-defensive]** (low) `temp-session.ts:110`: the parser re-validates data written only by its own writer.
4. **[copy]** (low) `strings.ts:3554`: `从临时中移除` should quote the group name, as `从「临时」中移除`.

Verdict: fix before merge: 1 (at least one browser pass, plus an e2e for opening a session from the dialog), 2, 3, 4. Separately, two points need your decision rather than a code change: the touch long press and the 20-entry cap. Design #173 records both, but your wording rules them out.

packages/web/src/features/company/tickets-page.tsx
hiyouga: Fixed in 72d15d6ff. After filing its ticket, `e2e/company.spec.mjs` now:
- clicks the card and sees the ticket dialog;
- drags the card with the mouse onto 进行中, sees the move confirmation, confirms it, and finds the card in that column;
- starts a ticket session (`POST …/tickets/:ticketId/start`) and opens it from the dialog's session title. The URL must still be `/chat/<id>` once the conversation is on screen;
- checks that 临时 lists the session below 工位, with the on-screen row marked, and that a reload keeps it;
- opens a second session, which lands on top, and removes it with its ✕ while it is on screen. The page stays and the other entry remains;
- empties the group with 全部关闭 and checks it stays empty across a reload.

**vps3:** the spec passes (1 test, 13.5 s).

**Negative control:** with this guard line removed, the spec fails at exactly that step. `expect(page).toHaveURL(/\/chat\/session-…$/)` received `…/marketplace/tickets`.

packages/web/src/features/company/ticket-dialog.tsx
hiyouga: Fixed in f1d98f86a. When the server answers a session with `agentId: ""` (its row is gone), the dialog renders the title as plain muted text, with the Session id in its tooltip, instead of a `TitleButton` — so there is nothing to click and no Temporary entry can be recorded for it.

packages/web/src/features/company/temp-session.ts
hiyouga: Fixed in 4bd80f2fc. The parser keeps the `JSON.parse` try/catch and is otherwise `return Array.isArray(parsed) ? (parsed as TempSessionEntry[]) : [];`. The per-entry checks, the dedupe and the cap are gone, and so are the two tests that pinned hand-edited lists. (The cap itself is gone everywhere in 755a9f9cc, per the user's decision.)

packages/web/src/lib/strings.ts
hiyouga: Fixed in 35a0996b4: `从「临时」中移除`. The en string stays "Remove from Temporary": the en dictionary cites a group by its capitalised name without quotes (`channels.archiveConfirm`: "…folds away under Archived…"), and this already follows it.

PR #750, opened 2026-09-16
feat(company): one model for every hire, ask the board before heavy work, and a research protocol

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/173

## Summary

Company-mode trial round, items 1, 8 and 11 from the user's list.

- **One model for every hire.** Nothing in code made the CEO pick models; the prose did. The init body, `company-ceo`, `company-hr`, `company-finance`, `company-setup` and the handbook template no longer ask for a model per role: every employee runs on the organization's model, or the Project's default when the organization names none (resolved per session: employee → organization → Project default), unless the board named one. Runtime tests pin that a hire without a model opens desk and ticket sessions without a model pair, and that a hire in an organization created with a model gets no model on its chart entry and opens its desk on the organization's model.
- **Ask the board before it touches the machine, money or the outside.** `company-employee` gains "What you may not decide alone" — four tiers: ask the board first and wait (compute running for more than a few minutes, >1 GB downloads, long-lived processes, paid services, publishing/pushing/outside messages, anything outside the shared workspace, destructive operations, missing credentials); propose to your manager (organization changes); notify then proceed; just do it. The ask is an all-hands mention of the creator + `ticket block --by user:<id>` + end of run; only a clear yes to that proposal unblocks. Mirrored (condensed) in the CEO skill, the init body and a new handbook section for new organizations.
- **Research protocol** (after karpathy/autoresearch). New `company-research` skill: a resource envelope approved by the board before any experiment loop; fixed harness and metric, one editable surface, a fixed budget per run, `results.tsv` with keep/discard/crash; adversarial author/reviewer review on the ticket flow (dedicated reviewer, scored rounds, 3-round cap escalating to CEO then board). Each round, the author moves the ticket to `review` and starts the reviewer's session with `penguin org ticket start <id> --agent-id <reviewer>`; the reviewer reproduces from its own `git clone`. `results.tsv` and `logs/` are gitignored, a run commits only the editable surface, and a discard is `git checkout <last kept commit>` + `git branch -D run-<tag>`, so the log survives every discard. The research mission example is rewritten to name all three.
- `agent-company` → `2026.09.16.1`; reconcile reinstalls employees' plugins, so existing organizations get the skills. Existing handbooks on disk are left as they are (user decision).

No hard enforcement was added.

## Verification
[...]

PR #749, opened 2026-09-16
feat(model-catalog): refresh DeepSeek, TokenDance, OpenRouter, Fireworks, SiliconFlow and both Qwen groups

hiyouga:
## Summary

Catalog refresh requested on 2026-09-16. Prices are per million tokens, cache hit / input / output.

**DeepSeek** (api-docs.deepseek.com/zh-cn/quick_start/pricing lists exactly two names now)
- Keeps `deepseek-flash` (**DeepSeek V4.1 Flash**, default model, unchanged) and `deepseek-v4-pro`, renamed **DeepSeek V4 Pro 0813**. Its peak price stays CNY 0.30 / 9 / 27 on DeepSeek's off-peak schedule, and DeepSeek keeps V4 Pro after 2026-09-14 with unchanged billing.
- Retires `deepseek-v4-flash` and `deepseek-v4-flash-vision-exp` as presets (kept as retired rows, see below). DeepSeek still accepts both and serves them as V4.1 Flash.

**TokenDance**
- `deepseek-v4.1-flash` already matched DeepSeek's price and is unchanged. `deepseek-v4-flash-vision-exp` is retired as a preset (see below).
- Promotions are re-rated to the seller's 2026-09-16 quote, with the list price stored and the discount applied:

  | Model | List (CNY) | Discount | Billed (CNY) |
  | --- | --- | --- | --- |
  | `deepseek-v4-flash-0731` | 0.15 / 1.5 / 4.5 | 10% | 0.135 / 1.35 / 4.05 |
  | `deepseek-v4-pro-0813` | 0.45 / 4.5 / 13.5 | 10% | 0.405 / 4.05 / 12.15 |
  | `kimi-k3` | 1.6 / 20 / 100 | 40% | 0.96 / 12 / 60 |

**OpenRouter** (anonymous `/api/v1/models`)
- Adds `qwen/qwen3.8-27b`: USD 0.15 / 0.214 / 2.55, 1,000,000 context, vision.
- `deepseek/deepseek-v4.1-flash` already existed and matches. The GPT-5.4 rows are kept, as the maintainer asked.

**Fireworks AI** (each model's fireworks.ai page)
[...]

PR #753, opened 2026-09-16
feat(web): mentions show names, ticket paths become capsules, the organization status is a text pill

hiyouga:

## Summary

Company-mode trial round, items 6, 7 and 9.

- **Mentions show names in the channel composer, and each one is a single block.** Picking from the @ menu writes `@<display name>`, tinted like a sent chip. Picks are tracked as ranges and converted back to the wire form (`@id` / `@user:id` / `@all`) on send, so the server and the message format are untouched.
  - Any edit that reaches into a picked mention is widened to cover all of it: Backspace or Delete at an edge, word deletes, cuts, typed-over selections, drops and IME commits.
  - The caret and the selection snap out of a name. The arrow keys step over it, a click lands at the nearer edge, a pointer range grows to hold it, and Shift+arrow grows or shrinks a selection by the whole mention.
  - A mention that touches the text around it is sent with a separating space, so `x@Ada`, `@Ada.md` and a hand-typed `@id` right after a CJK name (`@张三@ceo`) all deliver.
  - Copy, cut and drag keep mentions. When the selection holds a mention, copy and cut write the clipboard themselves:
    - `text/plain` is the text as shown (`@<name>`), which is what other apps get.
    - `application/x-penguin-mentions` is JSON of the same text with each mention's offset, name and id. A drag adds the same type beside the browser's own `text/plain`.
    - On paste or drop, the browser inserts the text, so native undo still works, and the composer puts the mentions back over it.
    - A mention comes back only when the payload's text is the inserted `text/plain` and its name is still what this channel shows for its id. Text from elsewhere, or from another organization, stays plain.
  - The @ menu also searches names in any script, and Enter or Tab during IME composition no longer picks a candidate.
  - The logic lives in `mention-draft.ts` and is pure. The textarea is kept so native caret, IME and undo keep working.
- **Undo.** I measured this in Chromium. When script writes the value (a React controlled write), the native undo stack goes stale: Ctrl+Z does nothing, and Ctrl+Shift+Z re-inserts text from before the write. A text→mentions history alone would therefore never fire.
  - So a native `beforeinput` listener replaces Backspace at a mention's end, and Delete at its start, with `execCommand("delete")` over the whole mention. This is the repo's existing "execCommand keeps native undo" pattern, with a state fallback.
  - That delete is one native undo step: Ctrl+Z brings the name back, and a small text→mentions history turns it back into a mention.
  - A cut that holds a mention deletes its selection the same way, so Ctrl+Z restores it too.
  - Undo, redo and an IME composition in progress are never written back.
  - Other edits that reach into a mention (word deletes, keyboard-app edits) are widened after the fact and written back. For those, native undo is lost, as it always has been after a pick.
- **Path capsules in tickets.** In the ticket dialog's goal, acceptance criteria, progress, result and blocked reason, data paths render as a folder or file capsule labelled with the last segment. A data path is `<app_data_dir>/…`, or an absolute path through `/<projectId>/organizations|agents/`. Hovering shows the full path, and a click copies it as written.
  - Separators are `/` or `\`, mixed as they come, so the paths a Windows server hands employees (`C:\Users\…\.penguin\data\<project>\organizations\…`) become capsules too.
  - Markdown reads `\.` as an escape, so in the Markdown fields a path written with `\` is copied from the source as written, not from the rendered text.
  - In prose, a path is ASCII segments ending at whitespace, CJK or punctuation.
  - A path that runs on through a non-ASCII segment into a separator or a file extension stays text, rather than becoming a capsule that copies its parent folder. So does the tail of a path that runs in from a non-ASCII segment (`C:\Users\张三\…`).
  - `/api/…` routes are never paths.
  - An inline code span holding one path becomes a capsule; fenced blocks and links stay verbatim.
  - Stored text is unchanged.
- **Organization status as a text pill** to the right of the name in the switcher (运行中 / 已暂停 / 配置无效), replacing the dot. The name truncates first. The English label for an unpaused organization is now "Running".

## Verification

- vps3, on the merged tree 00f749430 (main 078a7aa83 plus this branch), after building core and cli, then hmr, then `pnpm gen:ifaces` and the server:
  - `pnpm typecheck` passed for every package.
  - `pnpm lint` reported 0 warnings and 0 errors. `pnpm format:check` was clean.
  - Web suites, 111 tests, all passed: `mention-draft` 48, `path-capsules` 31, `channel-mentions` 15, `control-size` 5, `icon-scale` 10, `disclosure-anchor` 2.
  - `e2e/company.spec.mjs` ran against the built server and web app with the mock LLM. It passed 3 of 3 runs, and once more before the merge. It now drives a picked mention through the real composer:
    - The pick shows the tinted name.
    - The arrow keys step over it, and Shift+ArrowLeft selects all of it.
    - Backspace at its end clears it, and Ctrl+Z restores it tinted.
    - Ctrl+X and Ctrl+V move it, and a mouse drag moves it back. After each move it is still a mention.
    - The stored message carries `agent:marketplace_ceo`.
  - Controls: with the copy/cut handler disabled the spec fails at the paste. With only the drag handler disabled it fails right after the drop.
- Local: prettier on the touched files, and `git diff --check`.
- Earlier revisions:
  - On 8267a38c1, 29 Playwright checks in Chromium were run but not committed. Besides what the spec now covers, they included Delete at a mention's start, Ctrl+Backspace into a name, an IME composition over part of a name, and a click and a double-click inside a name. They also checked that `@张三@lab_ceo 请看` is stored as `@lab_zhang @lab_ceo 请看`.
  - Before 8267a38c1 I fuzzed the old path module with 400k adversarial inputs, and no input reached the `<` guard or the `~`/drive label branches before they were deleted.
  - An integration tree of this branch with the other five branches of the batch and #748 (`integ/r14`, local only) was built, typechecked, linted and format-checked. Its full package suites passed on vps3, except one web failure that #752 fixed.
- Not checked visually: the capsules and the header pill. The composer tint was checked in the DOM only.


packages/web/src/features/company/mention-draft.ts
```diff
@@ -0,0 +1,175 @@
...
+ * the message that is sent still carries the token the server resolves (`@ceo`,
+ * `@user:alice`, `@all` — see channel-mentions.ts).
+ *
+ * Why ranges rather than matching names on send: names are not unique and may hold spaces or
+ * CJK, so `@Alice Chen` cannot be read back to an id from the text alone. A range remembers
+ * which candidate was picked; the text only has to still say what was inserted.
+ *
+ * The rules:
+ *
+ * - An edit that touches a mention's characters — typing inside it, deleting part of it,
```
hiyouga: **[bug]** (design) This rule is the opposite of the new requirement that a picked mention is one atomic block: Backspace or Delete at its edge removes the whole mention. Right now an edit inside the name leaves the rest as plain text (`@Ada Lovelac`), which is sent and delivered to nobody. The review body has the full analysis. In short: widen any edit that intersects a mention to the whole token inside `draftApplyEdit` (not in keydown, which misses word deletes, cut, drag and Android keyboards), and snap the caret and selection endpoints out of tokens in `onSelect`. Defer the rewrite while an IME is composing, and decide whether undo should restore the ranges. Doing this removes the "tracked but not live" state and most of `liveMentions`.

packages/web/src/features/company/mention-draft.ts
```diff
@@ -0,0 +1,175 @@
...
+    let offset = 0;
+    for (const run of mentionRuns(wire)) {
+      if (run.mention !== null) parsed.set(offset, run.mention);
+      offset += run.text.length;
+    }
+    let failing: DraftMention | null = null;
+    let shift = 0;
+    for (const m of draft.mentions) {
+      if (!live.has(m)) continue;
+      if (parsed.get(m.start + shift) !== m.wire) {
```
hiyouga: **[bug]** The parse-back check covers tracked mentions only, but converting one can break a plain `@id` typed right after it. Scenario (I ran it against this module): pick 张三 to get `@张三 `, delete the trailing space, and type `@ceo ` by hand without picking. The draft is `@张三@ceo `. Sent as typed, the server would read `ceo`, because the `@` follows `三`. `draftWireText` returns `@zhangsan@ceo `, which parses to `["zhangsan"]` only: `@ceo` is now glued to `n` and is silently not delivered, with no tint change to warn anyone. That contradicts this file's header ("a typed or pasted `@id` is left as it is") and the changelog ("works as before"). The picked-both case in the test at mention-draft.test.ts:169 (`@张三@李四`, where 李四 is lost) is the same failure.

Fix: treat a converted token directly followed by `@` as glued. Better still (see the atomic-block note in the review body), emit a space between the converted token and the next `@` so both are delivered. Add a test with a hand-typed `@id` after a CJK mention.

hiyouga (commented):
Reviewed the three areas named for this PR: the composer's mention draft, the ticket path capsules, and the organization status pill. I also judged the edit model against the new requirement that a picked mention behaves as one atomic block. I ran the pure modules on vps3 to confirm the mention and path scenarios below.

**1. [bug / design] A picked mention is not an atomic block (new user requirement).** Today `splice` (mention-draft.ts:50-59) drops the range on any overlapping edit and leaves the rest as plain text. After a pick, the second Backspace turns `@Ada Lovelace` into `@Ada Lovelac`, which is sent verbatim and delivered to nobody. The requirement is that Backspace/Delete at the edge removes the whole mention. On a textarea that needs:
- **Deletion widens to the whole token, in the reconciliation, not in `onKeyDown`.** A keydown handler misses word/line deletes (Ctrl/Alt/Cmd+Backspace), Ctrl+X over part of a name, drag-out, and Android virtual keyboards (they often report Backspace as keyCode 229 / `key: "Unidentified"`). In `draftApplyEdit`, when the recovered span `[from, to)` intersects mentions, widen it to their union and return the rewritten text plus caret `from`. The composer writes that value and restores the caret with `setSelectionRange` in a layout effect. This stays pure and unit-testable in mention-draft.ts.
- **The caret never rests inside a token.** In `onSelect`, a collapsed caret strictly inside `(start, end)` moves to the edge on the side it came from, comparing against the previous `caret` state. ArrowLeft/Right then skip the token and a click lands at an edge. A non-collapsed selection with an endpoint inside extends that endpoint outward, which covers double-click and drag selection. Typing can then never land inside a name, so the "typing inside a name makes it plain text" rule and its `@Anna`/`@Annaa` caret disambiguation go away.
- **Replacing a partial selection** (select `Love`, type `x`) uses the same widening: the whole mention is replaced.
- **IME:** don't rewrite the value while composing, because setting `value` mid-composition breaks the composition. Track `onCompositionStart`/`End` and apply any pending widening at `compositionend`. Once the caret is snapped out of tokens, a composition only starts inside one when a selection already cut into it.
- **Undo:** a value the composer writes is not a native undo step (a pick already has this property). If Ctrl+Z after a whole-token delete should bring the chip back, keep a small `text -> mentions` history and restore the ranges when the value returns to a recorded text. Without it, the name comes back as plain text. The alternative that keeps native undo is `preventDefault` on a native `beforeinput` listener, then `setSelectionRange` over the token and `document.execCommand("delete")`. It is not simpler: React's `onBeforeInput` does not fire for deletions, each `inputType` needs its own range math, and composition input is not cancelable, so the reconciliation fallback is still needed.
- **Prerequisite: finding 3.** With widening, an edit placed at the wrong offset no longer just un-tints a mention. It deletes one the user did not touch.
- **The simpler model this enables:** if a tracked mention can never be partially edited, every tracked mention reads `@<label>` by construction. The "tracked but not live" state, the `liveMentions` fixed point (findings 6 and 7) and the tint loss all go away. The remaining glue cases (`x@Ada`, `@Aday`, `@张三@ceo`) are better handled by emitting a separator space in the wire text next to a converted token than by reverting the token to its name. An atomic chip that silently stops being a mention is the failure this requirement removes. `contenteditable` with `contenteditable=false` chips gives native atomic deletion, but it brings back the IME, paste-sanitising and undo problems the textarea was kept to avoid, so I don't recommend it.
- **Tests:** widening and snapping as pure cases in mention-draft.test.ts, plus one Playwright case for Backspace at the edge (jsdom performs no native edits).

**Findings, ranked**
1. [bug/design] Mentions are not atomic: see above (mention-draft.ts:13, :50).
2. [bug] mention-draft.ts:142: converting a mention glues a following hand-typed `@id`, which silently stops being a mention (`@张三@ceo ` is sent as `@zhangsan@ceo `).
3. [bug] channel-composer.tsx:242: the edit is anchored at `selectionStart`. After an undo, redo or drop that leaves the inserted text selected, a picked mention next to the edit is dropped.
4. [bug] path-capsules.ts:61: this server's REST routes (`/api/projects/<projectId>/organizations|agents/...`) match the data-path marker and render as folder capsules.
5. [bug] path-capsules.ts:38: a path with a non-ASCII segment becomes a capsule of its ASCII prefix, so the capsule copies the parent folder, not the written path.
6. [simplify] mention-draft.ts:129: the fixed-point loop is a single left-to-right pass.
7. [over-defensive] mention-draft.ts:127: the `@<label>` text filter can never fail.
8. [over-defensive] mention-draft.ts:93: optional `caret`, `Math.max(0, ...)` and composer `?? value.length` are all dead.
9. [over-defensive] path-capsules.ts:60: the `startsWith("<")` guard and the `~`/drive branches of `pathLabel` are unreachable.
10. [copy] shell-org-status.ts:2: the new header says the pill takes `ORG_STATUS_TONE`, but the pill reads `ORG_STATUS_BADGE`.

**Checked, nothing to raise**
- **Wire form:** a pick still sends `mentionInsertId`'s token (`@id` / `@user:id` / `@all`), the same as before, and `MENTION_RE` matches the server's `extractMentionTokens`. The `mention_not_member` path is unchanged.
- **Mention edge cases:** duplicate names and CJK names convert correctly, and the IME Enter/Tab guard follows chat-input's convention.
- **Path capsules:** fenced blocks and link labels stay verbatim. The parser-split `__init__` case stays text. Stored text is untouched and the click copies the piece's own string.
- **Org status pill:** zh 运行中 / 已暂停 / 配置无效 and en Running / Paused / Invalid configuration are correct. Badge green/amber/red resolve to `toneSurface` success/attention/danger. The name is `min-w-0 truncate` and the pill `shrink-0`, so the name truncates first. No other English "Active" label remains for an organization's status. The trigger's `aria-label` carries the status.

Verdict: fix before merge: 1 (atomic mentions), 2, 3, 4, 5; 6-10 are small and can ride along.

packages/web/src/features/company/channel-composer.tsx
```diff
@@ -157,29 +202,54 @@ export function ChannelComposer({
...
+                aria-expanded={panelOpen}
+                aria-controls={panelOpen ? listId : undefined}
+                aria-activedescendant={
+                  panelOpen && active ? `${listId}-${active.principal}` : undefined
+                }
+                disabled={sending}
+                {...noAutofill}
+                onChange={(e) => {
+                  const value = e.target.value;
+                  const at = e.target.selectionStart ?? value.length;
```
hiyouga: **[bug]** The edit is anchored at `selectionStart`, but `draftApplyEdit` (mention-draft.ts:93) needs the end of the inserted text: "an insertion ends where the caret now stands". The two agree for a keystroke or a paste, where the caret is collapsed. They disagree whenever the browser leaves the inserted text selected: Ctrl+Z restoring a deleted selection, redo, and drag-and-drop.

Scenario (I ran it against the module): with the mention live, `@Ada Lovelace please` sends `@ceo please`. Select ` please`, press Delete, then Ctrl+Z. The value is back with 13–20 selected. Anchored at 13, the edit is recovered as `[12,12) + "e pleas"`, which overlaps the name, so the send is `@Ada Lovelace please` with no mention. Anchored at 20 it is `@ceo please`. Under the atomic-block requirement this mis-placed edit would delete a mention outright, so this fix is a prerequisite for that change.

Fix: `const at = e.target.selectionEnd;`. `?? value.length` is dead as well: `HTMLTextAreaElement.selectionStart` and `selectionEnd` are typed `number`, not nullable. The "undo" test at mention-draft.test.ts:127 passes the end offset (13), not what this handler passes, so it does not cover this path.

packages/web/src/features/company/mention-draft.ts
```diff
@@ -0,0 +1,175 @@
...
+ * the character the next one's `@` is judged against — `@张三@李四` becomes `@zhangsan@lisi`,
+ * where the second `@` is glued to a word — and that influence only ever runs left to right.
+ * So the first failing mention is reverted to its text and the rest are judged again: a later
+ * one may have failed only because of it.
+ */
+export function liveMentions(draft: MentionDraft): DraftMention[] {
+  const live = new Set(
+    draft.mentions.filter((m) => draft.text.slice(m.start, mentionEnd(m)) === `@${m.label}`),
+  );
+  for (;;) {
```
hiyouga: **[simplify]** This fixed point is a single left-to-right pass. Whether mention *i* parses back depends on two things only. One is the character before its `@` in the output, which is decided by the mentions to its left. The other is the draft characters after its label up to the next non-id character; a later mention's `@` counts as one whether or not that mention converts. So "revert the first failure and judge everything again" lands on the same set as a single loop. Walk the mentions in order, append the plain text between them, and convert each one iff the token `mentionRuns(lastOutputChar + "@" + wire + followingIdRun)` reads at that `@` is exactly `wire`. That removes re-converting the whole draft on every failure, the `parsed` offset map, the `shift` bookkeeping, and the paragraph arguing that the loop terminates. (This is moot if the atomic-block rework replaces reverting with a separator space.)

packages/web/src/features/company/mention-draft.ts
```diff
@@ -0,0 +1,175 @@
...
+ * The mentions that convert: each still reads `@<label>`, and converted, its token parses back
+ * as itself under the server's grammar (mentionRuns mirrors it). Converting one mention changes
+ * the character the next one's `@` is judged against — `@张三@李四` becomes `@zhangsan@lisi`,
+ * where the second `@` is glued to a word — and that influence only ever runs left to right.
+ * So the first failing mention is reverted to its text and the rest are judged again: a later
+ * one may have failed only because of it.
+ */
+export function liveMentions(draft: MentionDraft): DraftMention[] {
+  const live = new Set(
+    draft.mentions.filter((m) => draft.text.slice(m.start, mentionEnd(m)) === `@${m.label}`),
```
hiyouga: **[over-defensive]** This filter can never remove anything. `splice` keeps only mentions that lie wholly inside the common prefix or the common suffix of the old and new text, so their characters are unchanged. `draftInsertMention` writes `@${label}` at `start`, and nothing else produces a `MentionDraft`. So `text.slice(start, end) === "@" + label` always holds. Start from `new Set(draft.mentions)` and drop "each still reads `@<label>`" from the comment above.

packages/web/src/features/company/mention-draft.ts
```diff
@@ -0,0 +1,175 @@
...
+ */
+export function draftApplyEdit(draft: MentionDraft, next: string, caret?: number): MentionDraft {
+  const prev = draft.text;
+  if (prev === next) return draft;
+  const max = Math.min(prev.length, next.length);
+  let suffix = 0;
+  while (suffix < max && prev[prev.length - 1 - suffix] === next[next.length - 1 - suffix]) {
+    suffix += 1;
+  }
+  if (caret !== undefined) suffix = Math.max(0, Math.min(suffix, next.length - caret));
```
hiyouga: **[over-defensive]** `caret?` is optional, but every caller (the composer and all tests) passes it, so the `caret !== undefined` branch is always taken. The caret is a textarea selection offset, so `caret <= next.length` and `Math.max(0, ...)` never binds. Make the parameter required and write `suffix = Math.min(suffix, next.length - caret)`.

packages/web/src/features/company/path-capsules.ts
```diff
@@ -0,0 +1,114 @@
...
+/** The whole of an inline code span, as one path: the same roots, any non-space segment. */
+const CODE_PATH = /^(?:<app_data_dir>|~|[A-Za-z]:)?(?:\/[^\s/]+)+\/?$/;
+
+const escapeRe = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+/** Whether a well-formed candidate starts at a root this module capsules. */
+function rooted(path: string, scope: PathScope): boolean {
+  if (path.startsWith(`${PLACEHOLDER_ROOT}/`)) return true;
+  if (path.startsWith("<")) return false;
+  const marker = new RegExp(`/${escapeRe(scope.projectId)}/(?:organizations|agents)(?:/|$)`);
```
hiyouga, line 68: **[bug]** `/<projectId>/(organizations|agents)/` is also the shape of this server's own REST routes: `/api/projects/:projectId/organizations...` and `/api/projects/:projectId/agents/:agentId/schedules|hooks|plugins`. A blocked reason like `GET /api/projects/default_project/organizations/co_lab/tickets/T-12 returned 404` renders a folder capsule labelled `T-12`. I checked this: `splitPaths` returns the route as a path, and `codePath` accepts `/api/projects/default_project/agents/ceo/schedules` in a code span. The employees debugging the API are exactly the ones who write lines like this, and the module header says a route must not read as a folder. A full URL is safe (`:8787/api` has a digit before the `/`); a bare route is not.

Fix: reject a candidate that starts with `/api/`, in `rooted` so prose and code are both covered. Add a test next to the `/api/v1/organizations` case that uses the real project id.

packages/web/src/features/company/path-capsules.ts
```diff
@@ -0,0 +1,114 @@
...
+  /** The Project the ticket belongs to: the directory an absolute path must pass through. */
+  projectId: string;
+}
+
+export type PathPiece = { kind: "text"; text: string } | { kind: "path"; path: string };
+
+const PLACEHOLDER_ROOT = "<app_data_dir>";
+
+/** One segment of a prose path: plain characters, or a whole `<placeholder>`. */
+const PROSE_SEGMENT = String.raw`(?:<[a-z][a-z0-9_]*>|[A-Za-z0-9._~+=@%-])+`;
```
hiyouga, line 45: **[bug]** A path with a non-ASCII segment is capsuled as its ASCII prefix, so the capsule names and copies a different path from the one written. Example (checked): `结果在 <app_data_dir>/organizations/co_lab/workspace/调研报告.md` renders a `workspace` folder capsule that copies `<app_data_dir>/organizations/co_lab/workspace/`, followed by plain `调研报告.md`. The handbook asks employees for ASCII file names, but files people drop into the workspace, and zh users' own folders, are not bound by that. A capsule that copies the parent is worse than no capsule. The test at path-capsules.test.ts:90 (`workspace/实验/` becomes `workspace/`) codifies this.

Narrow fix: if the text right after a match continues as `[\p{L}\p{N}_-]+` followed by `/` or `.ext`, the path runs on through a non-ASCII segment, so leave it as text. Prose after a folder (`handbook/里有说明`, the case at :42) has no such continuation and keeps its capsule.

packages/web/src/features/company/path-capsules.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/** The whole of an inline code span, as one path: the same roots, any non-space segment. */
+const CODE_PATH = /^(?:<app_data_dir>|~|[A-Za-z]:)?(?:\/[^\s/]+)+\/?$/;
+
+const escapeRe = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+/** Whether a well-formed candidate starts at a root this module capsules. */
+function rooted(path: string, scope: PathScope): boolean {
+  if (path.startsWith(`${PLACEHOLDER_ROOT}/`)) return true;
+  if (path.startsWith("<")) return false;
```
hiyouga: **[over-defensive]** This line is unreachable. Both regexes accept a leading `<...>` only as `<app_data_dir>`, and they require a `/` after it, so that case already returned on the line above. The same goes for `name === "~"` and the drive-letter test in `pathLabel` (:107): they only matter when a path is just the root (`~/...` or `C:/...` trimmed to `~/` or `C:/`), and `rooted` rejects those before `hasSegment` runs. The one reachable job of `hasSegment` is `<app_data_dir>/...` trimmed to `<app_data_dir>/`. Delete this line and the two dead alternatives.

packages/web/src/features/company/shell-org-status.ts
```diff
@@ -1,6 +1,7 @@
 /**
- * The shell's organization status marks (pure, unit tested): the tone the switcher's dot and
- * the overview's status pill take, and what a session-list group should render given what
+ * The shell's organization status marks (pure, unit tested): the tone the status pill (the
```
hiyouga: **[copy]** The reworded header says `ORG_STATUS_TONE` is the tone "the status pill (the switcher's trigger, the overview, settings)" takes, but `OrgStatusPill` reads `ORG_STATUS_BADGE` in shared.tsx, a second invalid/paused/active → colour map. Only `OrgStatusDot` reads `ORG_STATUS_TONE`. The colours agree today (Badge red/amber/green resolve to `toneSurface` danger/attention/success), so nothing renders wrong; only the comment is wrong. Either say only the rows' dots take this tone, or have the pill derive its colour from `ORG_STATUS_TONE` and delete `ORG_STATUS_BADGE` so one map decides.

packages/web/src/features/company/mention-draft.ts
hiyouga: Fixed in 8267a38c1: an edit that reaches into a mention takes all of it, the caret and selection snap out of names, and Backspace/Delete at an edge is one native delete that Ctrl+Z restores as a mention; IME composition waits for compositionend.

packages/web/src/features/company/mention-draft.ts
hiyouga: Fixed in 8267a38c1: `draftWireText` sets a glued token apart with a space, so a hand-typed `@ceo` after `@张三` sends `@zhangsan @ceo` and both deliver (tests added).

packages/web/src/features/company/channel-composer.tsx
hiyouga: Fixed in 8267a38c1: the edit is anchored at `selectionEnd` and `?? value.length` is gone; a regression test selects ` please`, deletes it, undoes, and the mention survives.

packages/web/src/features/company/mention-draft.ts
hiyouga: Removed in 8267a38c1: `liveMentions` and its fixed point are gone, since every tracked mention now reads `@<label>` and the send is one left-to-right pass.

packages/web/src/features/company/mention-draft.ts
hiyouga: Removed in 8267a38c1 along with `liveMentions`; the `@<label>` invariant is now stated on `MentionDraft` instead of being filtered for.

packages/web/src/features/company/mention-draft.ts
hiyouga: Fixed in 8267a38c1: `caret` is required, `Math.max(0, …)` is dropped, and the composer's `?? value.length` is gone.

packages/web/src/features/company/path-capsules.ts
hiyouga: Fixed in 8267a38c1: `rooted` rejects `/api/`, with tests for the real-project route in prose and in a code span.

packages/web/src/features/company/path-capsules.ts
hiyouga: Fixed in 8267a38c1: a path that runs on through a non-ASCII segment into `/` or an extension stays text, while `handbook/里有说明` keeps its capsule (tests updated).

packages/web/src/features/company/path-capsules.ts
hiyouga: Removed in 8267a38c1, after fuzzing the old module with 400k adversarial inputs confirmed that neither the `<` guard nor the `~`/drive label branches could be reached.

packages/web/src/features/company/shell-org-status.ts
hiyouga: Fixed in 8267a38c1: the header now says only the switcher rows' dots take `ORG_STATUS_TONE` and the pill reads `ORG_STATUS_BADGE`, and `OrgStatusDot`'s comment no longer mentions the trigger.

packages/web/src/features/company/mention-draft.ts
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * A picked mention is one block, so every tracked mention reads `@<name>` by construction:
+ *
+ * - An edit that reaches into a mention takes all of it. The box reports only its new value, so
+ *   the edit is recovered from that and widened to the whole mention (draftApplyEdit): Backspace
+ *   at its end, Delete at its start, a word or line delete, a cut, a drop or a keyboard app's
+ *   own edit — whatever made the edit, the mention goes in one piece.
+ * - The caret never rests inside a mention, and a selection holds a mention whole or not at all
+ *   (draftSnapSelection), so typing lands before or after one.
+ * - Text typed or pasted anywhere else only moves the mentions after it. A pasted `@Name` is
```
hiyouga, line 19: **[bug]** Cutting and pasting inside the box drops the mention.

**Steps:**
1. Pick 张三 from the @ menu.
2. Select the mention; the selection snaps to all of it. Press Ctrl+X.
3. Paste it at the start of the line.

**Result:** the box shows an untinted `@张三`, and `draftWireText` sends it as it is. The server's grammar reads only ASCII ids (`channel-mentions.ts:140` mirrors it). The message posts, but 张三's desk is never triggered, and nothing warns the sender. Copying and pasting within the box does the same, and so does the drag this PR lists as a known limit.

**Why it is a regression:** before this PR the box held `@lab_zhang`, so the same cut and paste still delivered. This is new behaviour, not an old limit carried over.

**Fix, in `channel-composer.tsx`:**
- On `copy`/`cut`, and on `dragstart`, when the selection holds tracked mentions, also write a private clipboard type, e.g. `application/x-penguin-mentions`. It carries each mention's offset, label and wire.
- On `paste`/`drop`, re-create the mentions when that type is present and its text equals the `text/plain` being inserted.

This needs two pure helpers here: one slices a draft's mentions, one splices them back in. It also removes the drag limit. Text pasted from elsewhere stays plain, as this line and `mention-draft.test.ts:180` intend.

packages/web/src/features/company/path-capsules.ts
```diff
@@ -0,0 +1,129 @@
...
+const PROSE_SEGMENT = String.raw`(?:<[a-z][a-z0-9_]*>|[A-Za-z0-9._~+=@%-])+`;
+
+/**
+ * A candidate path in prose: a root, then `/segment` repeated, then an optional trailing `/`.
+ * The lookbehinds keep a path from starting in the middle of a word; a bare `/`, `~` or drive
+ * root additionally may not follow a path character or a `:`, which is what keeps a URL
+ * (`https://…`) and the inside of another path from reading as a root. The placeholder cannot
+ * be mistaken for either, so `Output:<app_data_dir>/…` still starts a path.
+ */
+const PROSE_PATH = new RegExp(
```
hiyouga, line 54: **[bug]** Windows paths never become capsules.

**Why:** `PROSE_PATH`, `CODE_PATH`, the marker in `rooted` and `pathLabel` accept only `/` as a separator. A ticket session's `Workspace:` line is `created.workspace` (`packages/server/src/runtime/organization/triggers.ts:258`), and the default workspace is `path.join(orgDir, "workspace")` (`packages/server/src/organization/paths.ts:135`).

**Result:** on a Windows install, employees name their deliverables as `C:\Users\…\<project>\organizations\<org>\workspace\…`. Those are exactly the long paths you wanted collapsed, and they stay as text.

**Fix:** accept `\` as a separator too.
- Run `rooted` on a copy with `\` normalized to `/`.
- Split the label on either separator.
- Keep copying the path exactly as written.

If Windows is out of scope, record that in design #173 instead. Its clause spells forward slashes only.

packages/web/src/features/company/channel-composer.tsx
```diff
@@ -82,13 +159,63 @@ export function ChannelComposer({
...
+    if (result.draft.text !== shown) pendingCaret.current = result.caret;
+  };
+
+  // Backspace at a mention's end, Delete at its start: the whole mention, deleted natively.
+  // React's onBeforeInput does not fire for deletions, hence the native listener; the refs keep
+  // it current without re-attaching.
+  useEffect(() => {
+    const el = inputRef.current;
+    if (el === null) return;
+    const onBeforeInput = (e: InputEvent) => {
```
hiyouga, line 189: **[test]** The pure rules are well pinned, but the wiring that makes them hold in a real box has no committed test. Untested:
- this native `beforeinput` → `execCommand("delete")` path;
- the caret after a write-back (`pendingCaret`);
- the `onSelect` snap;
- the `compositionend` widening.

The 29 Playwright checks in the PR description were not committed. `e2e/company.spec.mjs` only `fill()`s a typed `@id`, which takes the plain-text path.

**Suggested test:** move the core of that run into `company.spec.mjs`, next to its channel section (L192-205). There:
- type `@Plug` and press Enter. The box should show `@Plugin Marketplace CEO`, and once sent, the stored message's `mentions` should hold `agent:<org>_ceo`. The spec already checks `mentions` through the API at L236-240.
- delete the pick's trailing space. One Backspace at the name's end should then clear the whole mention, and Ctrl+Z should bring back the tinted name.

hiyouga (commented):
## Review of #753 (head 8267a38c1)

This PR changes three places in company mode:
- **Channel composer.** Picking from the @ menu now shows the display name. The draft tracks each pick as a range and converts it back to `@id` on send (`mention-draft.ts`, pure logic). A native `beforeinput` path keeps Backspace or Delete at a mention's edge on the browser's undo stack.
- **Ticket dialog text.** Data paths render as folder or file capsules through a remark pass (`path-capsules.ts`, `path-capsule.tsx`).
- **Organization switcher.** The trigger shows the organization's status as a text pill.

**Earlier review round.** The review of 2026-09-17 on ed38b3a6c raised ten findings, and 8267a38c1 addresses all of them. I checked each fix against the code. Every finding below is new.

### What you asked for vs. what the PR does

| You asked | This PR |
|---|---|
| @-mentioning an Agent in a channel shows its name in the input box, not its id | **Done.** An employee shows its display name. A Project member shows its user id, which is the name members carry everywhere in the app. `@all` shows 所有人 / Everyone.<br>**The Agent still receives the id:** the sent message carries `@lab_zhang` / `@user:alice` / `@all`, exactly as before. Names exist only in the box and in rendered chips. The server and the message format are unchanged. |
| A mention is one atomic block; deleting removes all of it | **Done.**<br>• Backspace at its end or Delete at its start removes the whole mention in one native step, and Ctrl+Z brings it back as a tinted mention.<br>• A word delete, cut, typed-over selection, drop or IME commit that reaches into it also removes all of it.<br>• The caret, the arrow keys, clicks and selections snap to its edges.<br>**Gap:** cut or copy followed by paste turns it into plain text that notifies nobody (finding 1). |
| Very long ticket paths render as a folder capsule | **Done** in the goal, acceptance criteria, progress, result and blocked reason. This covers `<app_data_dir>/…` paths and absolute paths through `/<projectId>/organizations|agents/`.<br>• A capsule shows a folder or file glyph and the last segment. Hover shows the full path, and a click copies it.<br>• Other absolute paths, `/api/…` routes, fenced blocks and link labels stay text, as design #173 says.<br>**Gap: not on Windows.** Backslash paths are never matched (finding 2). |
| The organization's status next to its name is a text pill (运行中 / 已暂停) | **Done.** 运行中 / 已暂停 / 配置无效 (Running / Paused / Invalid configuration) sits to the right of the name in the switcher trigger.<br>• The name truncates first, and the status is part of the trigger's accessible name.<br>• The menu rows keep their dots. Design #173's no-dot clause covers only the trigger.<br>• The English label "Active" becomes "Running" everywhere; its only use is `orgStatusLabel`. |

### Conflicts with main
- **Against main:** `git merge-tree --write-tree origin/main origin/feat/company-readable-names` (main 078a7aa83) finds no conflicts. `strings.ts` and `strings-en.ts` auto-merge.
- **Base is behind:** the branch is 15 commits behind main. Its base, 5b595e254 (#716), predates #751, #754, #755, #750, #785, #749, #758 and #782–#784. None of those touch a file this PR changes other than the two dictionaries.
- **Semantic checks:**
  - Nothing on main calls the removed `insertMention`.
  - After #751 and #754, the dialog's text fields are still plain strings.
  - #782–#784 change `message-stream.tsx`, which this PR does not touch.
  - The e2e's typed `@${ORG}_ceo …` still posts. `fill()` goes through the plain-text path, `draftWireText` leaves untracked text as it is, and the query regex rejects the space, so no panel opens.
- **Together with #752:** it merges cleanly in either order. `shared.tsx`, `ticket-dialog.tsx` and the dictionaries auto-merge.
- **vps3, on the three-way merge (main 078a7aa83 + #752 + #753):** all four checks pass. (1) `pnpm --filter @prismshadow/penguin-web typecheck`. (2) The 11 web suites the two PRs touch, 176 tests: ticket-press, temp-session, org-sessions, install-scope, company-click-targets, mention-draft, path-capsules, channel-mentions, control-size, icon-scale, disclosure-anchor. (3) `pnpm lint`: 0 warnings, 0 errors. (4) `pnpm format:check`: clean.

### The checks you asked for
- **Serialisation:** the Agent receives the id, never the name (see the table).
- **Paste:** pasted `@Name` text stays plain. That is intended for text from elsewhere, but it breaks cut-and-paste of a picked mention (finding 1).
- **Undo:**
  - An edge delete becomes `execCommand("delete")` over the whole mention, so it is one native undo step. This is the repo's existing `execCommand?.(…)` + state-fallback pattern (agent settings, memory tab). A text-to-mentions history then restores the tint.
  - Other widened edits, and a pick, are written from script and lose native undo. The PR states this.
- **IME:** nothing is written back while composing, and the widening waits for `compositionend`. Enter and Tab during composition neither pick nor send, through the repo's usual `isComposing` guard.
- **Chinese names:**
  - CJK labels work, and the menu matches `@张` (the query now allows `\p{L}`).
  - Tests cover text glued to a mention: `@张三你好`, and `@张三@ceo`, which is sent as `@zhangsan @ceo`.
  - Two employees with the same name stay apart.
- **i18n / zh:** `pathCapsule.hint` and `pathCapsule.copy` exist in both dictionaries and use their parameter. The zh copy puts a half-width space between the CJK text and a path.
- **Changelog:** the pair mirrors section for section in the past tense. The hard-wrapped zh lines match the other merged zh entries.
- **Design #173:** each clause matches the code:
  - the composer: display name, tint, `@id` on the wire, one block, a space added on send, matching in any script;
  - the capsules: roots, ASCII in prose, trailing full stop, inline code, the folder/file rule, copying as written;
  - the switcher trigger.
- **Tests:** `mention-draft` (38 tests) and `path-capsules` (22 tests) pin the rules well. The composer's DOM wiring has no committed test (finding 3).

### Findings (inline)
1. **[bug]** (medium) `mention-draft.ts:19`: cutting or copying a mention and pasting it inside the box turns it into a plain `@name` that triggers nobody. This is a regression.
2. **[bug]** (medium) `path-capsules.ts:54`: Windows backslash paths never become capsules.
3. **[test]** (medium) `channel-composer.tsx:189`: the browser checks in the PR description were not committed.
4. **[over-defensive]** (low) `path-capsules.ts:69`: the empty-`projectId` clause is dead code.

Verdict: fix before merge: 1, 2 (or record Windows as out of scope in the design), 3, 4.

packages/web/src/features/company/path-capsules.ts
```diff
@@ -0,0 +1,129 @@
...
+const CODE_PATH = /^(?:<app_data_dir>|~|[A-Za-z]:)?(?:\/[^\s/]+)+\/?$/;
+
+const escapeRe = (s: string) => s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+
+/** Whether a well-formed candidate starts at a root this module capsules. */
+function rooted(path: string, scope: PathScope): boolean {
+  if (path.startsWith(`${PLACEHOLDER_ROOT}/`)) return true;
+  if (path.startsWith("/api/")) return false;
+  const marker = new RegExp(`/${escapeRe(scope.projectId)}/(?:organizations|agents)(?:/|$)`);
+  return scope.projectId !== "" && marker.test(path);
```
hiyouga, line 69: **[over-defensive]** `scope.projectId !== ""` can never be false. The only caller is the ticket dialog, which passes the routed `projectId`, and that is never empty. The marker regex is also built before this check. Delete the clause.

packages/web/src/features/company/mention-draft.ts
hiyouga: Fixed in 1311195c7.

- When the selection holds a mention, copy and cut write the clipboard themselves. `text/plain` is the text as shown (`@<name>`), which is what every other app gets. `application/x-penguin-mentions` is JSON of the same text with each mention's `start`, `label` and `wire`. A drag adds the same type beside the browser's own `text/plain`.
- On paste or drop the composer parses that type and lets the browser insert the text, so the insertion and its undo step stay native. Then it puts the mentions back over the inserted text (`draftApplyClip`). A cut deletes its own selection with `execCommand`, so Ctrl+Z still restores it.
- A mention comes back only when the payload's text equals the `text/plain` being inserted and its label is still what this channel shows for its id. Text from elsewhere, and a name that would send a different id, stay plain.
- This also removes the drag limit the PR description listed.

Pure cases are in `mention-draft.test.ts`: slice, the serialize/parse round trip, malformed payloads, the label check, paste, a drag that moves the mention, and a drop into a mention. The browser path is in `company.spec.mjs` (63b77a23d). That spec fails at the paste when the copy/cut handler is disabled, and right after the drop when the drag handler is disabled.

packages/web/src/features/company/path-capsules.ts
hiyouga: Fixed in 94adf3b87.

- `\` is now a separator wherever `/` is, and the two can be mixed. This covers prose and code-span candidates, `rooted` (run on a copy with `\` read as `/`), `pathLabel` (the last segment by either separator) and `pathKind`. The capsule still copies the path as written.
- Markdown reads `\.` as an escape, so in the Markdown fields every default Windows data path arrived as `C:\Users\ada.penguin\…`. A path that holds a `\` is now taken from the Markdown source as written (`spellAsWritten`, using the text node's position). A path written with `/` keeps the text's spelling, because a source backslash there can only be an escape (`co\_lab`).
- A bare `/` or `\` root right after a non-ASCII segment (`C:\Users\张三\.penguin\…`) stays text, rather than becoming a capsule that copies only the tail. This mirrors the existing rule for a path that runs on into a non-ASCII segment.

Tests cover drive-letter paths with backslashes, mixed separators, code spans, labels, kinds, the run-in case, and the Markdown spelling in prose and in a code span.

packages/web/src/features/company/channel-composer.tsx
hiyouga: Fixed in 63b77a23d. In the all-hands channel, `company.spec.mjs` now types `@Plug`, picks the CEO and checks each step:

- The box shows `@Plugin Marketplace CEO`, tinted. The layer's mention spans carry `data-mention` so the spec can find them.
- ArrowLeft and ArrowRight step over the mention, and Shift+ArrowLeft selects all of it.
- After the trailing space is deleted, one Backspace clears the mention, and Ctrl+Z brings it back tinted.
- Ctrl+X and Ctrl+V move it behind a sentence, and it is still a mention.
- A mouse drag moves it back to the front, and it stays a mention.
- The stored message is `@marketplace_ceo 请看信息架构`, with `mentions: ["agent:marketplace_ceo"]`.

On vps3 the spec passed 3 of 3 runs on the merged tree. As controls, it fails at the paste when the copy/cut handler is disabled, and right after the drop when only the drag handler is disabled.

packages/web/src/features/company/path-capsules.ts
hiyouga: Removed in c508e8d2f.

hiyouga:
Design follow-up for this round (mention clipboard round-trip, Windows path separators, the non-ASCII-folder rule): https://github.com/Prism-Shadow/penguin-harness-design/pull/199

PR #754, opened 2026-09-16
refactor(server): tickets are read in the frontmatter format only

hiyouga:

## Summary

Company-mode trial round, item 12: tickets have one format — YAML frontmatter + prose sections. The user decided no compatibility with the old `# Ticket:` header format is needed; the shim was scheduled for removal one release after the frontmatter format shipped (0.2.13).

- Deleted `parseLegacyTicket`, its header/progress helpers, the old ticket-id pattern and `splitPrincipalList` (legacy-only). A file that does not start with `---` fails with `the file must start with \`---\` (YAML frontmatter)`.
- Such a file takes the existing invalid-ticket path: listed only under "Ticket files that failed to parse", `invalid` in `penguin org ticket ls`, `org_ticket_invalid` recorded, 404 on read, 409 `ticket_invalid` on write (400 `bad_request` on both when its slug is not letters-only, since every ticket route checks the id first); it is never rewritten and its id stays taken.
- Wording: "ticket fields" (`blocked` / `blocked_by`) instead of headers in the handbook template and `company-employee`, including the sweep step that still named `Blocked`.
- `agent-company` plugin version `2026.09.16.1` → `2026.09.17.1`, so installed copies pick up the skill wording.
- Changelog entry is marked Breaking. It gives the field mapping for converting a file by hand, and a separate recovery for a ticket whose slug is not letters-only: 0.2.13's write path cannot reach one, so the file is renamed to a letters-only id and every `parent` / `blocked_by` naming it is updated.

The escalation PR (#750) has merged into main. This branch merged main (`eb1a42ebb`), taking #750's post-review text wherever it conflicted, so the diff against main is this PR's own changes.

## Verification
- vps3, at this head (`a13c2b4ac`), in a fresh install: built core and cli, then hmr, then `pnpm gen:ifaces` and server. The seven organization server suites (`organization-{digest,files,notices,rota,runtime,scenario}`, `organizations-routes`) passed 134 of 134. That includes this PR's two tests: a `# Ticket:` file is rejected with the new message, and listing, creation and writes behave as above while the file stays byte-identical. Also core `plugins.test.ts` (21 passed) and docs `skills-sync.test.ts` (3 passed). `pnpm typecheck`, `pnpm lint` (0 warnings, 0 errors) and `pnpm format:check` are clean.
- Local (edits and git only): `prettier --check` on the touched files, `git diff --check`, and `node scripts/check-plugin-versions.mjs origin/main`, which exits 1 with the plugin at `2026.09.16.1` and 0 at `2026.09.17.1`.
- Earlier round, before main was merged: an integration tree merging this branch with the other five branches of the batch and #748 (`integ/r14`, local only) was built, typechecked, linted and format-checked locally, and its full package suites ran on the vps3 box: core 1250 passed / 5 skipped, server 2068, web 2417 of 2418 (the one failure, a storage key missing from install-scope, was fixed on the ticket-UX branch), cli 466, docs 53, landing 71.


plugins/agent-company/skills/company-employee/SKILL.md
```diff
@@ -211,7 +210,7 @@ penguin org ticket show <ticket_id> [--json]
 penguin org ticket create --title <s> (--goal <s> [--criteria <s>] | --body-file <path>) [--owner <principal>] [--slug <words>] [--parent <ticket_id>] [--notify <p,p>] [--priority P0|P1|P2] [--due <date>]
 penguin org ticket move <ticket_id> --to <col> [--reason <s>]   # moving into rejected requires a reason
 penguin org ticket assign <ticket_id> --owner <principal>
-penguin org ticket block <ticket_id> --reason <s> [--by <principal|ticket_id>]   # writes Blocked / Blocked-by; the ticket stays in its column
+penguin org ticket block <ticket_id> --reason <s> [--by <principal|ticket_id>]   # writes `blocked` / `blocked_by`; the ticket stays in its column
```
hiyouga, line 213: **[bug]** `agent-company` changes here, but `plugins/agent-company/plugin.json` stays at `2026.09.16.1`, the version #750 already set.

- **The green `plugin versions` check is stale.** Run 35144327104 checked against base `85fa4b401`, which is main at `2026.09.14.3`. That run was on 2026-09-16, before the base was switched to `feat/company-ask-before-heavy-work` on 09-17. Against the current base (`8638c543`, already at `2026.09.16.1`), `git diff --name-only 8638c543...HEAD -- plugins/` lists this `SKILL.md`, and the version is unchanged on both sides. So `scripts/check-plugin-versions.mjs` exits 1 with `agent-company (still 2026.09.16.1)`. The same happens once #750 is squash-merged and this PR is retargeted to main.
- **Beyond CI:** an instance that reconciles on main between the two merges installs `2026.09.16.1` with the old text (`# writes Blocked / Blocked-by`). It never receives this edit, because `reconcileEmployeePlugins` only updates when the library's version is greater than the installed one.

Fix: bump `plugins/agent-company/plugin.json` on this branch (e.g. `2026.09.17.1`).

changelog/unreleased/2026-09-16-tickets-frontmatter-only.md
```diff
@@ -0,0 +1,52 @@
...
+
+Affected: a ticket written in the header format by a build before 0.2.13 and never written to
+since — 0.2.13 rewrote every ticket in the frontmatter format on its first write. Such a file is
+left on disk untouched and is handled like any other ticket file that does not parse:
+
+- The board shows it only under "Ticket files that failed to parse", with its path and the error;
+  `penguin org ticket ls` prints it as an `invalid` row (`invalidFiles` under `--json`), and the
+  reconcile pass records an `org_ticket_invalid` error.
+- Opening it answers 404 `ticket_not_found`; every write to it is refused with 409
+  `ticket_invalid` (404 when its id still carries a digit), so nothing converts or overwrites it.
```
hiyouga: **[copy]** "(404 when its id still carries a digit)" is not what a client gets. Every ticket route runs `requireTicketId` (`packages/server/src/http/routes/organizations.ts:72-75`) before the service, and that uses the letters-only `TICKET_ID_PATTERN`. So for an id like `2026-09-02-launch-v2`, `GET …/tickets/:id` and every write (`PUT`, `move`, `block`, `unblock`, `progress`, `start`, `attach`) answer **400 `bad_request` "Invalid ticket id."**. The 404 only exists inside `requireTicket`, and neither the Web App nor `penguin org` reaches that directly. The runtime test only uses a letters-only id, so nothing covers this claim.

Fix: "(400 `bad_request` on read and write alike when its id still carries a digit)". Make the same change to the zh entry, lines 30-31 (「id 仍含数字时为 404」).

changelog/unreleased/2026-09-16-tickets-frontmatter-only.md
```diff
@@ -0,0 +1,52 @@
...
+
+To recover, write to the ticket once through 0.2.13 (a progress line or a move converts it), or
+replace the title line and headers by hand with a frontmatter block:
+
+- `# Ticket: <title>` → `title`; `Status` → `status` (it must match the column directory).
+- `Owner` → `owner`, or the `Initiator` value when `Owner` is empty.
+- `Initiator` → the first `history` entry: `{at: <time of the first progress line>, by: <initiator>, action: created}`.
+- `Notify: a, b` and `Sessions: a, b` → the lists `notify: [a, b]` and `sessions: [a, b]`.
+- `Parent`, `Priority`, `Due`, `Blocked` → `parent`, `priority`, `due`, `blocked`; `Blocked-by` →
+  `blocked_by`. A ticket id there needs a letters-only slug.
```
hiyouga: **[copy]** The recovery steps don't work for a legacy ticket whose **own** id has a digit. Those are the ids the deleted `LEGACY_TICKET_ID_PATTERN` existed for.

- **"Write to the ticket once through 0.2.13" fails.** 0.2.13's routes already ran `requireTicketId` with the letters-only pattern, so that write answers 400 and nothing is converted.
- **Following the mapping by hand while keeping the file name leaves a broken ticket.** The file parses and shows up on the board, because `store.listTickets` does not filter names. But opening it and every write answer 400. Line 49 says a `parent` / `blocked_by` value needs a letters-only slug, but not that the file it names has to be renamed to match. Keep the old digit id there and the child fails to parse ("parent must be a ticket id"). Change only the reference and the child is flagged `Parent … does not exist`.

Fix: add a step. Rename such a file to a letters-only slug (`<yyyy-mm-dd>-<letters>.md`, with a letter suffix if that id is taken), and update every `parent` / `blocked_by` that names it. Also say that the "through 0.2.13" path only works for letters-only ids. Make the same change to zh lines 34 and 40-41.

hiyouga (commented):
Reviewed only this PR's own diff against `feat/company-ask-before-heavy-work`.

Findings, most important first:
1. **[bug]** `plugins/agent-company/skills/company-employee/SKILL.md:213`: `plugin.json` is not bumped. The green `plugin versions` check ran against main before the base changed. Against the current base, and after #750 merges, it fails. Instances that took #750's version also never receive this text.
2. **[copy]** `changelog/unreleased/2026-09-16-tickets-frontmatter-only.md:36` (zh :30-31): a ticket id with a digit gets 400 `bad_request` from the routes on read and on write, not 404.
3. **[copy]** `changelog/unreleased/2026-09-16-tickets-frontmatter-only.md:41-49` (zh :34, :40-41): neither recovery path works for a legacy file whose own id has a digit. The file has to be renamed to a letters-only slug, and every reference to it updated.

Checked and found fine:
- The legacy reader is fully removed. `parseLegacyTicket`, `LEGACY_HEADERS`, `LEGACY_TICKET_ID_PATTERN` and the server's `splitPrincipalList` have no remaining references. The Web App's own `splitPrincipalList` is separate and still used. `isPersonPrincipal`, `historyNote`, `defaultTicketNotify` and `ticketSections` all still have callers, so no dead helpers are left.
- An invalid file behaves as described:
  - Listing puts it in `invalid` and records `org_ticket_invalid` (board, overview and CLI `ls`).
  - Reading returns 404 from `ticketDetail`.
  - Every write method goes through `requireTicket` and gets 409.
  - Creating a ticket with the same id takes a suffix, because `findTicket` is non-null.
  - `triggers.ts` only writes tickets it got from the listing, so the file is never rewritten.
- The `h.at === ""` branches in the web and `lastClosedAt` are still reachable: 0.2.13 already wrote converted files to disk with `at: ""`.
- The other field-mapping rows match the deleted parser: `Owner` falls back to `Initiator`, an empty `Notify` defaults the same way, unknown headers become fields of the same name, and progress lines become a sentence plus a history entry.
- The en and zh entries match section for section.
- No new defensive code was added.

Verdict: fix before merge: 1, 2, 3

plugins/agent-company/skills/company-employee/SKILL.md
hiyouga: Fixed in c8c267db4: `agent-company` is now `2026.09.17.1` and `node scripts/check-plugin-versions.mjs origin/main` passes (exit 1 at `2026.09.16.1`); the bump also ships deaef9abc, which renames the sweep step's leftover `Blocked` to the `blocked` field.

changelog/unreleased/2026-09-16-tickets-frontmatter-only.md
hiyouga: Fixed in fd49342c4: both entries now say 400 `bad_request` on read and write alike when the slug still carries a digit or is otherwise not letters-only.

changelog/unreleased/2026-09-16-tickets-frontmatter-only.md
hiyouga: Fixed in a13c2b4ac: both entries limit the 0.2.13 write path to letters-only slugs and give the hand recovery for any other ticket: rename the file to a letters-only id (letter suffix when taken), convert the headers with the mapping, and point every other ticket's `parent` / `blocked_by` at the new id.

PR #755, opened 2026-09-16
feat(server,web): AI-suggested ids for Projects, Agents and Benchmarks

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/173 (merged)

## Summary

The user asked for AI id generation everywhere ids are created, like company mode: type the name first, and the model suggests a valid id from it. The New Project, Create Agent and manual New Benchmark forms now have the same 「用 AI 生成」 id field as the organization and channel dialogs.

- **Route:** `POST /api/projects/:projectId/suggest-id` `{ name, kind, taken? }` → `{ id, source, reason? }`, kinds `project | agent | benchmark | org | channel`. One implementation (`services/semantic-id-suggest.ts`: model → ASCII slug → dated placeholder, one retry) serves every kind; `OrganizationService.suggestId` delegates, so org/channel prompts are byte-identical to main's.
- **Removed route (breaking):** `POST /api/projects/:projectId/organizations/suggest-id` is deleted, with its `ID_KINDS` and route tests. Nothing in the repo called it after this PR (the org and channel dialogs use the general route, and the CLI and plugins never called it), but it shipped in 0.2.13, so the changelog entry is marked `Breaking` with a Compatibility section: send the same body to `POST /api/projects/:projectId/suggest-id`. No compatibility code is kept, so there is no backward-compatibility file.
- **Kinds:** no new prefixes — Project and Agent ids bare snake_case, Benchmark ids kebab-case, a non-admin's Project id `<username>-<suffix>`; `co_` / `ch_` unchanged. An id that would start with a digit or be one character gets the kind's name in front.
- **Scope:** a Project suggestion asks the default model of the Project the dialog was opened from (membership required); Agent suggestions need membership, Benchmark suggestions ownership. The ids avoided are the names each create route refuses with 409, read as names only: `ProjectLifecycle.takenProjectIds` (every row plus every entry under the data root), `AgentLifecycle.takenAgentIds` (index rows plus every entry under `agents/`), `Benchmarks.takenIds` (every entry under `benchmarks/`). A leftover folder that no list shows (a Benchmark deleted mid-evaluation, an Agent folder without `system_config.yaml`, a Project directory without a row) is therefore avoided, and a click no longer opens every Agent and Benchmark. Avoided ids are never put in the prompt.
- **Web:** `SemanticIdField` + notice moved to `features/semantic-id/` and are shared by all five dialogs; the Benchmark title→id auto-slug was removed (typing a name never fills the id, as in the org dialog). Strings for the field moved to a top-level `semanticId` section; the Agent dialog's hint names its Name field (`S.agent.idGenerateHint`).
- Model failures for the new kinds are recorded under error source `id_suggest`; org/channel failures stay under `organization`.

Not included: the hire dialog's new-Agent option (employee ids carry an `<org_id>_` prefix by skill convention — a separate decision).

### Review follow-up

All six findings of the review are addressed in 90a720748, and main is merged in (65761ed8a; the only earlier conflict was `platform.ts`'s imports and `ApiModule` list, where both sides were kept).

- The Agent names include the index rows as well as the folders because `createAgent`'s 409 check reads both. An index row whose folder is gone still refuses the id.
- The Project names include every data-root entry, not only directories. A file there makes the create fail as well, because the Project's directory cannot be made.
- The semantic-id helpers take a `SemanticIdRule` only, `ruleOf` and the `uniqueSemanticId` defaults are gone, and `semanticIdPrompt` is module-private. The stranger test turns company mode on and asserts `project_not_found` for every kind. `proposalValue` loses the unreachable `""` check.

## Verification

At 65761ed8a, on vps3:
- server `semantic-id`, `suggest-id-routes`, `organization-runtime`, `organizations-routes`, `semantic-ids`, `utility-completion`, `benchmarks`, `errors`: 8 files / 207 tests passed (new: the config-less folder test in `suggest-id-routes`).
- web `semantic-id-field`, `id-suggest-notice`, `benchmark-prompts`, `control-size`: 4 files / 39 passed.
- docs: 8 files / 60 passed.
- `pnpm typecheck` (all 12 packages), `pnpm lint` (0 warnings, 0 errors), `pnpm format:check`: clean. Locally: `git diff --check`.

Before the review, at ef313d80b: server 8 files / 205, full web suite 176 files / 2330, docs 53. An integration tree merging this branch with the other five branches of the batch and #748 also ran its full package suites on vps3: core 1250 passed / 5 skipped, server 2068, web 2417 of 2418 (the one failure, a storage key missing from install-scope, was fixed on the ticket-UX branch), cli 466, docs 53, landing 71.


packages/server/src/http/routes/suggest-id.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    // Writing a Benchmark is Project management — owner only, as its create route demands.
+    if (kind === "benchmark") deps.access.requireProjectOwner(user.userId, projectId);
+    const reserved =
+      kind === "project"
+        ? deps.projects.listAll().map((p) => p.projectId)
+        : kind === "agent"
+          ? (await deps.agents.listAgents(projectId)).map((a) => a.agentId)
+          : (await deps.benchmarks.list(projectId)).benchmarks.map((b) => b.id);
```
hiyouga: **[bug]** The ids avoided here are not the ids the create routes refuse, so a proposal can still come back 409.

- **Benchmark:** `POST …/benchmarks` refuses any existing directory (`fs.mkdir(benchDir)` → EEXIST → 409 `benchmark_exists`). `benchmarks.list()` skips every directory that has no `benchmark_config.toml`, and its own comment names a way that happens: a Benchmark deleted while its evaluation is still running comes back as a directory with no config. Scenario: delete `report-writing` mid-evaluation → Generate from "Report Writing" → `report-writing` → Create → "A Benchmark with this id already exists; pick another". The list page shows no such Benchmark.
- **Agent:** `createAgent` returns 409 for any `agents/<id>` directory, but `listAgents` drops directories with no `system_config.yaml`.
- **Project:** `createProject` also checks `dirExists(projectDir(...))`, not only the DB row that `listAll()` returns.

These are also the costly way to get a list of names, on every click. `listAgents` loads the plugin library and reads every Agent's meta, vault, schedule count, installed plugins and memory count, and backfills the DB. `benchmarks.list` parses every config and scoreboard.

Fix: avoid the directory names instead, which is exactly what the 409 checks look at. Read `readdir(agentsDir)` / `readdir(benchmarksDir)` (ENOENT → `[]`), and for Projects use the rows plus the matching directories. That needs a names-only method on the two mechanisms instead of the list views.

packages/docs/content/server-api.en.md
```diff
@@ -259,7 +260,7 @@ All paths below are under `/api/projects/:projectId/organizations`. Every route
...
...` — plus `taken?`, the ids the proposal must avoid → `{id, source, reason?}`. The Project's default model translates the name into one snake_case English core (`source: model`), asked a second time with the format spelled out when the first answer yields no id; an ASCII slug of the name answers when no model is configured or neither answer is usable (`source: fallback`); and a name neither path can name gets `co_org_<yyyymmdd>` / `ch_channel_<yyyymmdd>` (`source: placeholder`) with a `reason` of `no_default_model`, `model_failed`, `unusable_answer` or `no_ascii`. The route never fails for a name it cannot translate — a dialog that asked for an id always gets one — and every model dead end is recorded as an `organization` / `id_suggest_failed` error. The server then prefixes the core by kind — `co_` for `org`, `ch_` for `channel`, never doubled when the core already carries it — before capping the length and avoiding `taken`, so a proposal is always prefixed and a hand-typed id never is. The completion runs with thinking off and the shared meta budget, belongs to no Session and is not metered. `POST /api/projects/:projectId/suggest-id` with `kind: org` / `channel` answers the same |
```
hiyouga: **[simplify]** After this PR nothing in the repo calls `POST /organizations/suggest-id`. `endpoints.ts` moves the org and channel dialogs to `/suggest-id`, and no CLI command or plugin ever called it. It shipped in 0.2.13 (`changelog/0.2.13/2026-09-02-company-mode.md`), so keeping it only serves API clients outside the repo. That makes it a compatibility shim. AGENTS.md makes that the user's call, and requires `changelog/unreleased/YYYY-MM-DD-backward-compatibility.md` to say how long it stays and who removes it. Design #173 says to keep it but gives no removal plan.

Pick one:
- Record it that way.
- Or delete it: the handler and `ID_KINDS` in `organizations.ts`, its cases in `organizations-routes.test.ts`, and this sentence in both docs. Then mark the changelog entry `Breaking`, with the `## Compatibility` section.

packages/server/test/suggest-id-routes.test.ts
```diff
@@ -0,0 +1,245 @@
...
+  });
+
+  it("validates the body, and asks nothing of a Project the caller cannot reach", async () => {
+    expect((await owner.post(url, { kind: "agent" })).status).toBe(400);
+    expect((await owner.post(url, { name: "", kind: "agent" })).status).toBe(400);
+    expect((await owner.post(url, { name: "Report" })).status).toBe(400);
+    expect((await owner.post(url, { name: "Report", kind: "team" })).status).toBe(400);
+    expect((await owner.post(url, { name: "Report", kind: "agent", taken: "x" })).status).toBe(400);
+    const stranger = apiClient(t.app, (await provisionUser(t.app, "stranger")).cookie);
+    for (const kind of ["project", "agent", "benchmark", "org", "channel"]) {
```
hiyouga, line 262: **[test]** For `org` and `channel`, this loop cannot catch a missing membership check. Company mode is off in this test, so a 404 comes back either way: `project_not_found` if access is checked, `company_mode_off` if it is not. `completion.prompts` stays empty in both cases too.

Scenario: move `requireProjectAccess` so it runs only for the three new kinds. This test still passes, yet a stranger could spend another Project's model on org/channel ids while company mode is on (`OrganizationService.suggestId` does no access check of its own).

Fix: assert `error.code === "project_not_found"` for every kind, or run the org/channel iterations after `t.deps.serverSettingsRepo.setCompanyMode(true)`.

packages/web/src/features/agents/agents-page.tsx
```diff
@@ -722,25 +723,29 @@ export function AgentsPage() {
...
             value={name}
             onChange={(e) => setName(e.target.value)}
             hint={S.agent.nameHint}
+            autoFocus
+          />
+          <SemanticIdField
+            projectId={projectId ?? null}
+            kind="agent"
+            label={S.agent.id}
+            hint={S.agent.idHint}
```
hiyouga, line 741: **[copy]** This dialog's first field is `S.common.name` (「名称」 / "Name"). No `generateHint` is passed, so the id hint ends with the default clause: 「；也可以从显示名生成」 / "; you can also generate one from the display name", naming a field this dialog does not have. The Benchmark dialog got `generateHint` for exactly this mismatch.

Fix: add `S.agent.idGenerateHint` to both dictionaries (「；也可以从名称生成」 / "; you can also generate one from the name") and pass it here.

packages/web/src/features/semantic-id/id-suggest-notice.ts
```diff
@@ -36,3 +37,15 @@ export function idSuggestNotice(
...
 }
+
+/**
+ * What goes in the box: the proposal, less the part the field draws in front of the box. The
+ * server answers a non-admin's Project id whole (`alice-research_lab`) while the box holds only
+ * what follows `alice-`; an id that does not start with the locked part is left whole, so the
+ * caller's validation names the problem instead of the field hiding it.
+ */
+export function proposalValue(id: string, lockedPrefix: string | undefined): string {
+  return lockedPrefix !== undefined && lockedPrefix !== "" && id.startsWith(lockedPrefix)
```
hiyouga: **[over-defensive]** Nothing ever passes `""` here. The only caller passes `lockedPrefix` only when `prefix` is non-empty (`project-dialogs.tsx`), and `SemanticIdField` itself treats `""` as a real prefix: it draws an empty locked span and `rounded-l-none`. So this guard and the field disagree about what `""` means.

Fix: drop the `lockedPrefix !== ""` clause, and the `proposalValue("co_research_lab", "")` row in `semantic-id-field.test.ts:96`.

hiyouga (commented):
Independent review of the kind table, the per-kind permissions, collision avoidance, the `SemanticIdField` move, the removed Benchmark auto-slug and the `id_suggest` error source.

Findings, ranked:
1. **[bug]** `suggest-id.ts:91-96`: the reserved ids come from the list views, which skip directories that the create routes still reject with 409. A proposal can therefore collide (a leftover Benchmark directory is the reachable case), and each click reads every Agent's and Benchmark's full card to get their names.
2. **[simplify]** `POST /organizations/suggest-id` no longer has a caller in the repo. It shipped in 0.2.13, so keeping it is a compatibility shim. That needs the backward-compatibility record AGENTS.md asks for (or delete it and mark the change as breaking).
3. **[test]** `suggest-id-routes.test.ts:206`: for `org` / `channel` the stranger loop passes whether or not the route checks membership.
4. **[copy]** `agents-page.tsx:740`: the Agent id hint says "display name" (显示名), but the field above it is labelled "Name" (名称).
5. **[over-defensive]** `id-suggest-notice.ts:48`: the `lockedPrefix !== ""` check can never be hit.
6. **[simplify]** `semantic-id.ts:123`: the `SemanticIdKind | SemanticIdRule` union, `ruleOf` and the `uniqueSemanticId` defaults are only used by tests.

Checked and fine: the org/channel prompts and the retry rule are byte-identical to main's. The org truncation (300) and the `organization` error source are unchanged. Every kind's `accepts` matches its create route: `SEMANTIC_ID_PATTERN` for admin Projects and Agents, `<userId>-` + `PROJECT_SUFFIX_PATTERN` ≤ 64 for non-admin Projects, `isValidId` ≤ 100 for Benchmarks. Permissions match the create routes (member / member / owner). The server-side reserved ids never reach the prompt; only a `taken` list the client sends does. A non-admin proposal can only collide inside that user's own namespace, so the `listAll()` avoidance leaks no one else's ids. `slugFromTitle`, `idTouched` and their test are gone, and nothing on main still references them. The zh/en `semanticId` sections match, and zh `idHint` / `idPrefixHint` end in "；创建后不可修改", so the appended "；也可以从…生成" reads correctly. No `<form>` wraps any of the three dialogs, and `Button` is `type="button"`, so Generate cannot submit.

**Conflicts with main:** only `packages/server/src/platform.ts`, in the import block and the `ApiModule` controllers list. Main added `PluginRegistryRoutes` / `InstalledPluginRoutes` next to `PluginRoutes`; keep both sides. `git merge-tree` merges every other file cleanly. The merged tree has no remaining reference to `features/company/semantic-id-field`, `organization/semantic-id` or `S.company.idSuggest`.

Verdict: fix before merge: 1 (and resolve the platform.ts conflict); 2–6 are non-blocking.

packages/server/src/services/semantic-id.ts
```diff
@@ -0,0 +1,234 @@
...
+    prefix,
+    maxLength: PROJECT_ID_MAX_LENGTH,
+    accepts: (id) =>
+      id.startsWith(prefix) &&
+      PROJECT_SUFFIX_PATTERN.test(id.slice(prefix.length)) &&
+      id.length <= PROJECT_ID_MAX_LENGTH,
+  };
+}
+
+function ruleOf(kind: SemanticIdKind | SemanticIdRule): SemanticIdRule {
```
hiyouga: **[simplify]** No production code passes a kind string: the route and `OrganizationService.suggestId` both pass a `SemanticIdRule`. These exist only so `semantic-id.test.ts` can keep writing `"org"`:
- the `SemanticIdKind | SemanticIdRule` union on `prefixSemanticId`, `fallbackSemanticId`, `sanitizeSuggestedId` and `placeholderSemanticId`;
- `ruleOf`;
- the `separator = "_"` / `maxLength = 64` defaults on `uniqueSemanticId`.

Fix: take `SemanticIdRule` only, delete `ruleOf` and the defaults, and have the tests pass `SEMANTIC_ID_RULES.org` etc. Also, `semanticIdPrompt` in `semantic-id-suggest.ts` is exported but never imported; make it module-private.

packages/server/src/http/routes/suggest-id.ts
hiyouga: Fixed in 90a720748: the route now avoids the names the create routes refuse, read as names only through `ProjectLifecycle.takenProjectIds` (rows + data-root entries), `AgentLifecycle.takenAgentIds` (index rows + `agents/` entries) and `Benchmarks.takenIds` (`benchmarks/` entries); a new test puts a config-less folder of each kind in place, checks it does not list but still 409s, and gets `_2` / `-2`.

packages/docs/content/server-api.en.md
hiyouga: Deleted in 90a720748 (maintainer decision): the handler, `ID_KINDS`, its `organizations-routes` cases and this row (folded into the general `/suggest-id` row, both languages); the changelog entry is `Breaking` with a Compatibility section, and no compat code is kept.

packages/server/test/suggest-id-routes.test.ts
hiyouga: Fixed in 90a720748: the stranger loop runs with company mode on and asserts `error.code === "project_not_found"` for every kind.

packages/web/src/features/agents/agents-page.tsx
hiyouga: Fixed in 90a720748: added `S.agent.idGenerateHint` (「；也可以从名称生成」 / "; you can also generate one from the name") to both dictionaries and passed it as `generateHint`.

packages/web/src/features/semantic-id/id-suggest-notice.ts
hiyouga: Fixed in 90a720748: dropped the `lockedPrefix !== ""` clause and the `proposalValue("co_research_lab", "")` test row.

packages/server/src/services/semantic-id.ts
hiyouga: Fixed in 90a720748: the helpers take a `SemanticIdRule` only, `ruleOf` and the `uniqueSemanticId` defaults are gone, the tests pass `SEMANTIC_ID_RULES.*`, and `semanticIdPrompt` is module-private.

PR #759, opened 2026-09-16
fix(server): the session list marks WeChat-enabled conversations again

hiyouga:
A conversation with an enabled WeChat binding loses its messaging mark (the paper plane beside the title) as soon as the session list reloads. The Web App draws the mark for whatever `SessionInfo.messagingChannel` names, and `S.messaging.enabledIndicator` has a `wechat` entry; the in-place update right after enabling the binding in the dialog sets it. The server's session list, though, only reported Feishu, Telegram and QQ, so the next reload took the mark away. Company mode's desk rows read the same field and lost it the same way.

## Cause

WeChat shipped in #533, and that PR did add `"wechat"` to the `messagingChannel` lambda in `app.ts`. The server's module-tree rearrangement (#599, released in 0.2.12) ported the lambda into `runtime/session-manager.ts` without the fourth disjunct, and dropped the comment above it. No test pinned WeChat on the session list, so nothing failed.

## What changed

- `packages/server/src/runtime/session-manager.ts`: the lambda narrows `enabled.channel` to all four channels again, and its comment is back word for word. `MessagingBindingRow.channel` is `string` (a stored discriminator is untrusted text, and an unknown one is skipped, as in the bridge and the routes), so the `===` chain still does the narrowing instead of a cast.
- `packages/server/test/messaging-wechat.test.ts`: one test mirroring the Telegram one. It enables a WeChat binding, lists the sessions and expects `messagingChannel: "wechat"`, and it fails without the fix.
- `changelog/unreleased/2026-09-16-wechat-sidebar-mark.{md,zh.md}`.

Design: 05-ARCHITECTURE (Messaging) says the row marks only the **enabled** channel (「会话行标识只标**启用中**的渠道」), and 06-PROTOTYPE says the same. Neither limits the mark to particular channels, so the specs need no change.

## Verification

Run on the remote test host, as the user asked (Node 24, `pnpm install --frozen-lockfile`, the server's workspace dependencies built). Nothing ran on the dev box.

```
pnpm --filter @prismshadow/penguin-server test messaging      # all 11 messaging test files
 Test Files  11 passed (11)
      Tests  421 passed (421)                                 # messaging-wechat.test.ts: 28, incl. the new test

pnpm --filter @prismshadow/penguin-server typecheck           # exit 0, no errors
pnpm format:check                                             # All matched files use Prettier code style!
```

The new test with the lambda reverted to origin/main and the test kept:

```
× wechat binding routes and the long poll > the session list marks a wechat-ENABLED row with messagingChannel wechat
  → expected undefined to be 'wechat' // Object.is equality
```

With the fix back in place, the same run gives `Tests 1 passed | 27 skipped (28)`.


packages/server/src/runtime/session-manager.ts
```diff
@@ -2397,12 +2397,16 @@ export class SessionsModule {
...
+      // the row); a point query per row keeps the repo out of the service. An unknown
+      // stored channel reads as none (same defensive skip as the bridge and the routes).
       messagingChannel: (sessionId) => {
         const enabled = this.messagingRepo.findEnabled(sessionId);
         return enabled !== null &&
           (enabled.channel === "feishu" ||
             enabled.channel === "telegram" ||
-            enabled.channel === "qq")
+            enabled.channel === "qq" ||
+            enabled.channel === "wechat")
```
hiyouga, line 2409: [simplify] This chain is the bug's root cause, and it can still happen again. `enabled.channel` is `string`, and any subset of the `===` tests narrows it to a subset of `MessagingChannel`, which is still assignable to the return type. When #599 dropped the `wechat` disjunct, typecheck passed, and it will pass again when a fifth channel is added to the union but not here.

A single source makes that miss a compile error. For example, next to `MessagingChannel`:

```ts
const MESSAGING_CHANNEL_IDS = { feishu: true, telegram: true, qq: true, wechat: true } as const satisfies Record<MessagingChannel, true>;
export const isMessagingChannel = (c: string): c is MessagingChannel => Object.hasOwn(MESSAGING_CHANNEL_IDS, c);
```

The lambda then becomes `enabled !== null && isMessagingChannel(enabled.channel) ? enabled.channel : null`. `routes/messaging.ts`'s `specOf` (`CHANNEL_SPECS[channel as MessagingChannel] ?? null`) could use the same guard instead of a cast. Separately, `runtime/messaging/connector.ts:16` redeclares `MessagingChannel` instead of re-exporting it from `api/types.ts`, so the union itself exists in two copies (three counting `bridge.ts:1891`).

#751 moves this lambda into `runtime/messaging/enabled-channel.ts` and keeps the three-channel list there, so the guard fits best in that file when #751 is rebased onto this PR. Not blocking here.

hiyouga (commented):
Checked: every `SessionInfo` the server builds goes through `SessionService.toInfo` (list, `GET`, `PATCH` and fork in `routes/sessions.ts`), so this one lambda feeds both the DTO and the list rows. On the web side, `sidebar.tsx` and `org-session-groups.tsx` both read `S.messaging.enabledIndicator`, which has a `wechat` entry in `strings.ts` and `strings-en.ts`. The binding editor's `enabledChannel` chain also includes `wechat`. The other hand-written channel unions (`api/types.ts:1849`, `runtime/messaging/connector.ts:16`, `bridge.ts:1891`, `CHANNEL_SPECS` in `routes/messaging.ts`) all list four channels. In the #599 diff (d446ae33a), this disjunct is the only `wechat` line that was removed and not added back elsewhere. So the fix is complete for the regression it describes.

Findings, ranked:
1. [simplify] `session-manager.ts:2405-2409`: the `===` chain is a hand-kept copy of `MessagingChannel`, and leaving a channel out of it still type-checks. That is how #599 dropped WeChat with no error. A guard built from a `Record<MessagingChannel, true>` would turn the next miss into a typecheck failure.
2. [test] `messaging-wechat.test.ts:484`: the new test pins WeChat, but QQ still has no session-list test. Deleting `enabled.channel === "qq"` leaves every server test green today.

Conflict with #751 (`fix/org-sessions-dev-list`): the two PRs do conflict. `git merge-tree` of the two heads stops on a content conflict in `packages/server/src/runtime/session-manager.ts`: this PR edits the lambda that #751 replaces with `enabledMessagingChannel(this.messagingRepo, sessionId)`. There is a semantic conflict too. #751's new `runtime/messaging/enabled-channel.ts` still spells `feishu || telegram || qq`, and #751 now uses the same helper for `OrgDeskItem.messagingChannel`. If the conflict is resolved by taking #751's side, WeChat is dropped again from both the development list and company-mode desk rows. Suggested order: merge this PR first. It is a small fix for a regression already shipped in 0.2.12 and 0.2.13, and its new test fails on a rebased #751 unless `wechat` goes into `enabledMessagingChannel`. Then rebase #751: add `wechat` there, which is also the natural place for the single-source guard from finding 1, and drop the comment this PR restores, since the helper's doc comment covers it. If #751 lands first instead, this PR shrinks to that one disjunct in `enabled-channel.ts` plus the test.

Verdict: merge as is (neither finding blocks this PR; merge it before #751 and carry finding 1 into #751's rebase)

packages/server/test/messaging-wechat.test.ts
```diff
@@ -481,6 +481,15 @@ describe("wechat binding routes and the long poll", () => {
     expect(body.bindings.map((b) => b.binding.channel).sort()).toEqual(["telegram", "wechat"]);
   });
 
+  it("the session list marks a wechat-ENABLED row with messagingChannel wechat", async () => {
```
hiyouga, line 484: [test] This pins WeChat, but the session-list mark is still tested one channel per file: Feishu in `messaging.test.ts:2037`, Telegram in `messaging-telegram.test.ts:1914`, and now WeChat here. QQ has no such test, so deleting `enabled.channel === "qq"` from the lambda leaves the whole server suite green. That is the same way the WeChat disjunct was lost.

If the lambda moves to a `Record<MessagingChannel, true>`-backed guard (see the comment on `session-manager.ts`), the type system covers every channel and this per-file pattern is enough. If the chain stays, replace the per-file copies with one `it.each(["feishu", "telegram", "qq", "wechat"])` in `messaging.test.ts` that writes an enabled row through `t.deps.messagingRepo.upsert` + `setEnabled` and reads the list, so every channel is pinned in one place.

packages/server/src/runtime/session-manager.ts
hiyouga: Non-blocking for this fix. Carried into #751, which owns the new `enabled-channel.ts` helper: it gains `wechat` through one channel-id guard, and the per-channel list-mark tests fold into one table over all four channels.

packages/server/test/messaging-wechat.test.ts
hiyouga: Non-blocking for this fix. Carried into #751, which owns the new `enabled-channel.ts` helper: it gains `wechat` through one channel-id guard, and the per-channel list-mark tests fold into one table over all four channels.

PR #761, opened 2026-09-16
feat(ui,web): W0b — fonts, the Frost and Console themes, and the style-hook list

hiyouga:

**Review round (2026-09-18, #765's review):** merged #765's fixes (fe7bd903b; `penguinUi()` now only emits the font licences, no aliases), Frost's and Console's dark rules declare only what dark changes (c7e593d0e), and `--ui-control-py-lg` is Primer's 0.625rem in both (c4ef75bfa).

**Second round (2026-09-18):** merged #765's second round (5c04bd6cb: `--ui-fg-on-emphasis` dropped from all three themes, 188 names), and Frost and Console take Primer's row and menu-row padding-block, with Frost's density comment corrected (911f41f7b).

**Third round (2026-09-19, #761's review):** fonts are never inlined into the stylesheet (f58ac8a3d: main CSS 156,168 → 122,323 B gzip -9, no data URIs); Frost's `--ui-fg-subtle` is the design's `#898683` / `#7f7a74` again (b9ecc81c3); the credit takes the menu's secondary ink and its test catches inline gates (b26b7d9e6); `build-misans.py` drops the dead private-use rows, the version fallback and the worker re-verification, and its comment says the three mirrored pairs sit in two slices each (b02d4f7bb, `--check` still byte-identical); the docs say what every session downloads, Noto is Console's, subsetting is settled, heights match from W1a, and `.prettierignore` lost its dead entry (414404630).

**Fourth round (2026-09-19):** current `main` is merged in through #765 (`a699c8f7f`); no change of this branch's own.

Row B of the W0 plan, revised in stage K (the de-slop round): the token-contract additions, the closed list of style hooks, the Frost and Console token values and hook recipes, the bundled fonts with Frost now set in **MiSans**, and the in-app MiSans credit that font's licence requires. Nothing in the Web App names a new theme yet, so the app looks as before, apart from the credit line at the foot of the account menu.

> MiSans slices are committed and shipped as part of the application; maintainer decision 2026-09-17.
>
> Still open for the maintainer before merge: which licence text is authoritative — see [Licence questions](#licence-questions). (Subsetting was settled on 2026-09-17: slicing is delivery, not adaptation.)

## What changed

**Token contract** (`tokens.ts`, 186 → 188 names after #765 dropped `--ui-fg-on-emphasis`; first commit, published to the other stage K agents)
- Added `--ui-radius-control` (bridged in `theme.css` as `--radius-control`, i.e. `rounded-control`): Primer `0.375rem` (the app's `rounded-md` buttons), Frost `9999px`, Console `0`. Added `--ui-stack-0` (`.25rem`) and `--ui-stack-4` (`2.5rem`). Removed `--ui-glass-highlight`.
- `themes/github.css` declares the three new names at current-app values and nothing else changes for Primer (its polish is W1a).

**Style hooks** (`hooks.ts`, new, exported from the package index)
- `HOOKS = ["ui-glass", "ui-eyebrow", "ui-display", "ui-live", "ui-frame", "ui-underline-nav"]`, with the markup each recipe relies on in the doc comment. Deleted everywhere: `.ui-wash`, `.ui-grid` / `.ui-grid-dots`, `.ui-ticks`, `.ui-pill-hover`, `.ui-button`.

**`theme.css`**
- Base rule `:where(button, a, [role="button"], [role="tab"]) { transition-property: color, background-color, border-color, opacity, box-shadow }` — property list only, so nothing animates until a component sets a duration.
- The layer comment now says what `ui-theme` holds besides custom properties (the hook recipes, which beat utilities on purpose).

**Frost** (`themes/modern.css`)
- No ambient wash. Radius xs/sm/md/lg/xl = 4/6/10/14/20 px (in rem), pill = control = 9999px.
- Untinted tight shadows: raised `0 1px 2px .06`; overlay `ring .06, 0 4px 12px -4px .12`; modal `ring .06, 0 16px 40px -16px .2`; drawer = modal; dark per the design table.
- Glass: `rgb(255 255 255/.78)` / 16px / saturate 1.2 / line `rgb(48 46 45/.12)` (dark `rgb(30 29 27/.72)`, line `.1`); the recipe sets fill, blur and line only.
- Shared type scale (h1 1.953 → h6 .75rem, ×1.25 heading steps, unitless line-heights, caption lh 1.35, code lh 1.6); h6 sentence case, label tracking 0; durations 120/200/320ms.
- Fonts: sans and CJK `"MiSans"`, mono `"JetBrains Mono Variable"`; weight synthesis off (MiSans ships 400 and 500 only).
- Hooks: glass (with `@supports not` / `prefers-reduced-transparency` → 96% opaque), eyebrow (the h6 rung, muted, sentence case), display (400, −.02em), live (linear; caret 1.2s), frame (radius-md, transparent head), underline-nav (accent bar), hover-only scrollbar thumbs.

**Console** (`themes/geek.css`)
- The modal casts no shadow (its card carries the `line-emphasis` border); h2 Plex Sans 600 sentence case; h5 Plex Sans, no tracking; h6 Plex Sans uppercase +.06em (the group label); shared type scale; `--ui-dur-fast` 120ms.
- Hooks: eyebrow (h6 rung, uppercase, no brackets, no mono); display on `h1` / `[aria-level="1"]` only (condensed uppercase 600, lh 1.05, zh lh 1.2 no tracking); live (stepped; spinner `steps(8)`); frame (hairline box, 2.25rem ruled head, ruled foot, pane rules, no font forcing); underline-nav (1px ink rule).

**Fonts** (`packages/ui/src/fonts/`)
- Frost: **MiSans** (Xiaomi), Latin and Chinese in one family. MiSans has no official npm package, so it is vendored: `scripts/build-misans.py` (`uv run`, fonttools 4.65.0 + brotli 1.2.0 pinned in its PEP 723 header) reads the official `MiSans.zip` from https://hyperos.mi.com/font/ (Regular and Medium TTFs pinned by sha256), and writes `fonts/misans/`: 99 WOFF2 `unicode-range` slices per weight (Regular = 400, Medium = 500) plus `misans.css`.
  - Slices follow Noto Sans SC's fontsource partition (each code point assigned to the last slice that lists it, which is how a browser resolves Google's overlapping ranges) plus one `extra` slice for MiSans's other non-ideograph characters; the ~15,500 rare ideographs outside the partition are not shipped (+2.46 MB per weight).
  - Fidelity: every glyph keeps its outline, instructions, metrics and name; all OpenType features (e.g. `tnum`) and all name records are kept (the subsetter's pruning of unreferenced name IDs ≥ 256 is undone); only glyph-indexing tables are rebuilt and `DSIG` dropped. The script reads each finished WOFF2 back and fails unless its cmap, every glyph's points/contours/components/instructions, hmtx/vmtx, the whole name table and fsType match the official TTF. `--check` rebuilds in a temp dir and compares byte for byte.
- JetBrains Mono for Frost is the face `fonts/github.css` already declares (no duplicate `@font-face`, 0 extra bytes). Console's IBM Plex Sans Condensed ships 600 only (500 dropped).
- Dependencies: `@fontsource-variable/geist` and `geist-mono` removed (lockfile −16 lines).
- Licences: `LICENSES/misans.txt` is the full bilingual MiSans Font Intellectual Property License Agreement transcribed from Xiaomi's PDF (the one the download dialog binds you to; sha256 recorded in `vendored-fonts.json` and `fonts/README.md`). `vendored-fonts.json` is the one list of fonts that are not dependencies. `sync-font-licenses.mjs` never rewrites or removes a vendored licence, and `--check` verifies it (present, names its licence, whitespace-normalized). The Vite plugin emits it with the rest as `fonts-licenses/misans.txt`. `fonts/misans/README.md` says in the directory that these files are not Apache-2.0.
- `fonts/README.md`: families and why, loading, bytes, what the MiSans licence asks and where the app meets each condition, and how to update.

**Web** — the credit (MiSans licence condition 1)
- The account menu ends with one muted line under a rule, for every account and backend: `S.settings.fontCredit` = "Uses the MiSans font by Xiaomi" / 「使用了小米 MiSans 字体」. `account-menu.test.ts` pins it as the menu body's last child, outside every gate.

## Deviations from K-redesign, and why
1. **Frost's font is MiSans, not Instrument Sans** — the user's decision of 2026-09-17. §2.3's "≈ −30 KB net" becomes about +4.15 MB in `dist/` (table below).
2. **Console `--ui-dur-fast` 100 → 120ms.** §2.0 says fast/base are clamped into 120–200ms in every theme; §2.4's table still lists E2B's 100ms. I followed the cross-theme rule.
3. **The glass recipe no longer paints `--ui-shadow-overlay`.** That shadow existed to carry the removed inset highlight; forcing it from a hook would override a modal card's modal shadow and put a drop shadow on a sticky header. Elevation stays with the host's shadow token.
4. **Console `--ui-h1-lh` stays at the shared 1.15**; §2.4's "lh 1.05" lives in the `.ui-display` h1 recipe, so switching themes still never reflows a plain h1.
5. **Eyebrows read the h6 tokens** (size, lh, weight, tracking, transform) rather than repeating values — §2.0 defines h6 as the group-label rung.
6. **Frost's frame no longer rounds its head's buttons** (buttons read `--ui-radius-control` now), and **Console's frame lost the centred three-part title grid** (the terminal title bar; §2.4's frame recipe lists box, ruled head/foot and panes only).
7. **Frost dark focus ring stays at 40% accent**: the design row gives one value, and like the scrollbar row next to it I read it as the light value.
8. **Frost turns weight synthesis off** — not in the design; a consequence of shipping MiSans at 400/500 only (a synthesized bold smears Han strokes, and Frost's own tokens stop at 500).
9. **`vendored-fonts.json`** is new: A-architecture's "npm dependencies, not vendored binaries" cannot hold for MiSans, so the exception is one reviewable list the sync script and the guard can share.

## Notes for the stack
- **#762 (guards):** `font-licenses.test.ts` fails two cases on this branch — "arrive as dependencies, never as binaries" (198 vendored WOFF2) and "not kept for fonts the package no longer depends on" (`misans.txt`). Suggested rule: read `vendored-fonts.json`; allow WOFF2 only under each entry's `files`; require its `license` to contain `licenseTitle`; exclude it from the orphan check. `packages/ui/package.json` conflicts trivially on rebase (`"test": "vitest run"` plus this branch's two scripts).
- **#764 (gallery):** `foundations.css` still reads `--ui-glass-highlight`; the shots' loaded-font check needs `MiSans` (and Condensed 600 only); `/fonts` will list ~200 MiSans faces.
- **#770 (THIRD-PARTY-NOTICES):** quotes the download dialog's short text (conditions 1–3). The PDF the dialog links adds condition 4 ("retain the copyright notice and this Agreement in any copies"), condition 5, a trade-secret clause, revocation and PRC governing law. It also describes MiSans Light for screenshots, while the app now bundles Regular and Medium.

## Licence questions
Facts: MiSans 4.009, fsType `0x0008` (editable embedding; the "no subsetting" bit is not set), copyright record "Beijing Xiaomi Mobile Software Co.,Ltd.", licensor Xiaomi Inc.; the official zip contains fonts only, and the licence is the PDF the download dialog binds you to.

**Decided — subsetting (condition 2).** Cutting the official TTFs into WOFF2 slices is delivery, not 「改编」; maintainer decision 2026-09-17. Glyph data, features and name records are verified identical; the glyph-indexing tables are rebuilt, `DSIG` is dropped, and the container changes from TTF to WOFF2.

**Decided — distribution (condition 3).** MiSans slices are committed and shipped as part of the application; maintainer decision 2026-09-17. That covers the public repository (each slice is downloadable from GitHub), the npm release (`@prismshadow/penguin-server` publishes `web-dist`) and the assets a running server serves, with the licence text beside the files (`fonts-licenses/misans.txt`, condition 4 together with the name table every slice keeps) and the in-app credit (condition 1: one muted line in the account menu, both languages).

**Open:**
1. **Which text is authoritative:** #770's short text (conditions 1–3) or the PDF the download dialog links (conditions 1–5, trade-secret clause, revocation, PRC law)? This PR ships the PDF's.
2. **FYI:** the unmodified font carries stylistic sets `ss05` / `ss06` that draw the Xiaomi and MIUI logos; they ship in the Latin slice (removing them would be modifying the font) and nothing enables them.

## Font inventory
`pnpm --filter @prismshadow/penguin-web build` (and a specimen app through the same `penguinUi()` + Tailwind pipeline, `assetsInlineLimit: 0`):

| family | files | bytes |
| --- | ---: | ---: |
| MiSans 400 (Frost) | 99 | 2,121,092 |
| MiSans 500 (Frost) | 99 | 2,135,204 |
| Noto Sans SC Variable (Console; Primer from W1a) | 101 | 4,516,508 |
| Mona Sans Variable (Primer, W1a) | 2 | 55,312 |
| JetBrains Mono Variable (Frost; Primer W1a) | 2 | 55,600 |
| IBM Plex Sans Variable (Console) | 2 | 76,676 |
| IBM Plex Sans Condensed 600 (Console) | 2 | 35,860 |
| Commit Mono 400 / 700 (Console) | 2 | 95,432 |
| **total** | **309** | **9,091,684** (B's build: 4,945,712) |

Per theme, the font files it can download: Frost 4,311,896 · Console 4,724,476 · Primer (W1a set) 4,627,420; Primer names no bundled family in W0, so it fetches no font file. Every session, Primer's included, does download all three themes' `@font-face` declarations in the main stylesheet: 93.6 KB of its 122.3 KB gzip -9 (user decision 2026-09-18: accepted, no per-theme lazy font sheet). Builds never inline a font (`assetsInlineLimit`).

Per-theme loading (Playwright network log on the specimen page, the same in light and dark):
```
modern only=en:  5 files,  92,604 bytes: misans-400 3/27,212, misans-500 1/24,988, jetbrains-mono 1/40,404
modern only=zh: 27 files, 627,088 bytes: misans-400 16/350,788, misans-500 10/235,896, jetbrains-mono 1/40,404
geek   only=en:  4 files, 121,668 bytes: ibm-plex-sans 45,712, condensed-600 19,816, commit-mono-400 48,128, noto 1/8,012
geek   only=zh: 17 files, 892,980 bytes: noto 14/779,324 + the three Latin faces
github only=en/zh: 0 files
```
MiSans is requested only under Frost; Noto is never requested under Frost.

## Verification
- **vps3**, this branch merged with #762's head (`165dd39b3`, the W0 guards) in a throwaway local worktree, `pnpm --filter @prismshadow/penguin-ui exec vitest run`: **8 files, 72 tests: 69 passed, 1 skipped, 2 failed.**
  - `token-contract.test.ts` 12/12 — `github`, `modern`, `geek`: canonical rules + every one of the 189 names in light and dark.
  - `contrast.test.ts` 18/18 — all six theme × mode cells clear every floor, no stale exception, all five accent presets.
  - `font-licenses.test.ts` 4 passed, **2 failed as expected**: "are not kept for fonts the package no longer depends on" (`['misans.txt']`) and "arrive as dependencies, never as binaries checked into the package" (the 198 MiSans WOFF2) — the vendored exception #762 needs (see Notes for the stack).
  - `boot-script`, `no-theme-reads`, `no-app-strings`, `testing-helpers`, `demo-coverage` pass (the one skip is demo-coverage's "no component modules yet").
- **vps3**, `pnpm --filter @prismshadow/penguin-web exec vitest run test/account-menu.test.ts`: 14/14, including the new credit case.
- Token contract + contrast, local node script over the theme files: `modern` and `geek` × light/dark 189/189 tokens, 0 missing / extra / duplicate, all contrast pairs pass; `github` 189/189.
- MiSans: `build-misans.py` built 198 slices in 31s with every slice passing the read-back fidelity check; `--check` rebuilt byte-identical output twice. Independent cross-check: every glyph of 10 slices (both weights, latin, extra and 3 random CJK) drawn through fontTools `RecordingPen` from the slice and from the official TTF — 0 mismatched drawings; copyright record, fsType `0x8` and `tnum` present.
- `pnpm --filter @prismshadow/penguin-ui sync:font-licenses --check`: current (6 packages, 1 vendored).
- `pnpm --filter @prismshadow/penguin-web build`: MiSans slices emitted (190 files + 8 inlined), `fonts-licenses/misans.txt` emitted, no Geist.
- Screenshots (local Playwright, specimen page, en and zh side by side): Frost and Console × light and dark. Checked by eye: MiSans renders Latin and zh as one family; 600/700 requests render as MiSans Medium (no synthetic bold); pill controls and boxes on the radius scale; glass only on the menu and modal; Console's condensed uppercase h1 alone, Plex Sans sentence-case h2, uppercase group label, sans button labels, mono only on data, single-border modal.
- `pnpm --filter @prismshadow/penguin-ui typecheck` ✅, `pnpm --filter @prismshadow/penguin-web typecheck` ✅, `pnpm lint` ✅, `pnpm format:check` ✅, `git diff --check` ✅.


packages/web/vite.config.ts
```diff
@@ -66,7 +70,7 @@ function katexWoff2Only(): Plugin {
 }
 
 export default defineConfig({
-  plugins: [react(), tailwindcss(), katexWoff2Only()],
+  plugins: [penguinUi(), react(), tailwindcss(), katexWoff2Only()],
```
hiyouga, line 73: **[bug]** Every session downloads MiSans, Primer included. That's the reverse of the requirement that only a Frost session downloads MiSans (K-COMMON's lazy-per-theme rule, `fonts/README.md` § Loading, this PR's "Primer downloads nothing in W0", and #175's 「会话只下载所用主题的、实际用到的分片」).

I measured `pnpm --filter @prismshadow/penguin-web build` on vps3 for `dffd7fb41` and `911f41f7b` (gzip -9):
- **The stylesheet grows 5.3×.** The app's one render-blocking stylesheet goes from 153,776 to 479,341 bytes (29.7 KB to 156.2 KB gzip). The entry script is 663 KB gzip, so a first load grows by about 18%.
- **Slices are inlined.** Vite inlines any asset under `assetsInlineLimit` (4,096 bytes). Eight MiSans slices (`misans-{400,500}-{87,88,91,97}`, 15,316 bytes) and three Noto slices (9,292 bytes) sit in that stylesheet as base64.
- **The rest is declarations.** 198 MiSans `@font-face` rules with their `unicode-range` lists come to 60 KB gzip without the inlined data, and Noto's 101 rules to 32 KB. In W0 Primer names neither family.

Fix:
1. **Required.** Never inline a theme font, e.g. `build: { assetsInlineLimit: (file) => (file.endsWith(".woff2") ? false : undefined) }` here, and the same in the gallery's config. Rebuilt that way on vps3, the stylesheet has no data URIs and is 124.1 KB gzip. KaTeX_Size3, which is inlined today, then ships as its own 3.6 KB file.
2. **Needs a decision.** The remaining ~92 KB gzip is what declaring every theme's faces unconditionally costs (`fonts/index.css`). That premise ("declaring a face downloads nothing") holds for the font files but not for the declarations. Two options:
   - Load each theme's font sheet on demand: the theme provider dynamic-imports `fonts/modern.css`, or `fonts/geek.css` plus `cjk.css`, which Vite splits into their own CSS chunks. Primer then loads no font sheet in W0.
   - Accept the cost and say so in the README and the changelog.

packages/ui/src/themes/modern.css
```diff
@@ -1,5 +1,482 @@
...
+    --ui-surface: #ffffff;
+    --ui-surface-muted: #eeeeee;
+    --ui-inset: #f6f5f3;
+    --ui-overlay: #ffffff;
+    --ui-overlay-backdrop: rgb(48 46 45 / 0.28);
+
+    /* Colour — text */
+    --ui-fg: #302e2d;
+    --ui-fg-muted: #696764;
+    --ui-fg-subtle: #726f6c;
```
hiyouga: **[bug]** `--ui-fg-subtle` has left the design, here and at `:301` (K-redesign §2.2: `#898683` light, `#7f7a74` dark). It isn't among the PR's listed deviations.
- **It collapses Frost's ink steps.** Light `--ui-fg-muted: #696764` and this `#726f6c` are 1.13:1 apart (L* 43.7 vs 47.0), so disabled and placeholder text reads the same as secondary text. The design's pair was 1.56:1. In dark the gap drops from 1.68:1 to 1.44:1.
- **Nothing required it.** `contrast.test.ts` exempts `fg-subtle` ("placeholder and disabled ink, which WCAG exempts"). The 4.5:1 floor that commit 6aaf6918d cites comes from the gallery's tokens drawer (`contrastFloor` in `packages/ui-gallery/src/lib/contrast.ts`, #764). That drawer also fails Console's subtle ink (3.68:1 light, 3.94:1 dark) and Primer's (2.60:1), and neither was touched.
- **It splits Frost's subtle ink in two.** The gray bridge still maps `gray-400` (light) and `gray-500` (dark) to the design values. So the app's existing `text-gray-400 dark:text-gray-500`, which is exactly Primer's `fg-subtle`, renders `#898683` under Frost, while anything reading the token renders `#726f6c`. In Primer the class and the token are the same colour.

Fix: restore `#898683` and `#7f7a74`, and make the gallery drawer exempt `fg-subtle` the way the suite does. If `fg-subtle` should have a 4.5:1 floor after all, that's a contract change for all three themes: add the pair to `contrast.test.ts`, move `fg-muted` so a visible step remains, keep the bridge in step, and record it in #175.

packages/web/src/components/layout/user-menu.tsx
```diff
@@ -131,6 +131,12 @@ export function UserMenu({
               {S.auth.logout}
             </button>
           )}
+          {/* The font credit. The app ships Xiaomi's MiSans (the Frost theme's typeface), and its
+              licence requires software that uses it to say so. A muted line under a rule, not a
+              row: nothing here is actionable, and it is present for every account and backend. */}
+          <p className="mt-1 border-t border-gray-100 px-3.5 pt-2 pb-1 text-xs text-gray-400 dark:border-gray-800 dark:text-gray-500">
```
hiyouga: **[bug]** The credit uses Primer's `fg-subtle` ink, `gray-400` light and `gray-500` dark, which this stack's contrast suite defines as placeholder and disabled ink. It measures 2.60:1 on the menu's white panel and 4.02:1 on its `#0d0d0d` dark panel. The licence asks the software to 「特别注明」 (specifically state) that it uses MiSans. This is informational text, so it needs 4.5:1. The account id two rows up already uses the menu's secondary ink, `text-gray-500 dark:text-gray-400` (4.84:1 and 7.47:1). Use that here: still quiet, and legible.

packages/web/test/account-menu.test.ts
```diff
@@ -163,4 +163,16 @@ describe("the account menu", () => {
...
   });
+
+  it("credits MiSans on every menu, behind no gate", () => {
+    // The app ships Xiaomi's MiSans font, whose licence requires software that uses it to say
+    // so, whatever the account or the backend. So the credit is not one of the gated rows: it
+    // is the menu body's last child, opened right after sign-out's gate closes.
+    const credit = source.indexOf("{S.settings.fontCredit}");
+    expect(credit).toBeGreaterThan(-1);
+    const lastGateEnd = source.lastIndexOf(")}\n", credit);
+    expect(source.slice(lastGateEnd, credit).match(/<\w/g)).toEqual(["<p"]);
```
hiyouga: **[test]** "behind no gate" only counts opening tags between the last `)}` + newline and the credit. So a gate written inline passes:
- With `{user && <p className=…>` or `{desktopMode ? null : <p …>`, this slice still yields exactly `["<p"]`.
- With `</p>}`, line 176 still yields `["</p", "</d"]`.

Assert what the test's comment claims:
- between that `)}` and `<p` there is only whitespace and the `{/* … */}` comment;
- `</p>` is followed by whitespace and `</div>`, with nothing in between.

packages/ui/scripts/build-misans.py
```diff
@@ -0,0 +1,372 @@
...
+        "b03e98374e971594b0b7a9706d0704241f76e1b88556cdda79c5039ef8a638d1",
+    ),
+}
+
+# Left out of the `extra` slice: ideographs the partition does not carry, and private-use code
+# points (a vendor's logos live there, and they are not text).
+EXCLUDED_FROM_EXTRA = (
+    (0x3400, 0x4DBF),  # CJK Unified Ideographs Extension A
+    (0x4E00, 0x9FFF),  # CJK Unified Ideographs
+    (0xE000, 0xF8FF),  # Private Use Area
```
hiyouga: **[over-defensive]** The two private-use rows here and at `:90` exclude nothing. I checked on vps3 against the official zip: the pinned MiSans-Regular and MiSans-Medium 4.009 map no code point in U+E000–F8FF or U+F0000–10FFFF. The logos that the comment at `:82-83` places there are the `ss05`/`ss06` alternates, reached through GSUB rather than the cmap. The faces are sha256-pinned, so a release that adds private-use characters arrives as a reviewed pin change anyway. Delete both rows and the private-use half of the comment.

The same kind of code, in this file:
- **`:263`**, `font_version`'s `or "unknown version"`. Name ID 5 exists in both pinned faces. If it ever went missing, this would silently write "MiSans unknown version" into the committed header. Drop the fallback and let it fail.
- **`:186-187`**, `_init_worker`. It re-reads the zip and re-hashes both TTFs in every worker, after `build()` has already verified them. Pass the verified bytes through `initargs`.

packages/ui/scripts/build-misans.py
```diff
@@ -0,0 +1,372 @@
...
+    unicode_json = PARTITION_DIR / "unicode.json"
+    if not unicode_json.exists():
+        sys.exit(f"build-misans: {unicode_json} is missing; run `pnpm install` first")
+    raw: dict[str, str] = json.loads(unicode_json.read_text(encoding="utf-8"))
+    version = json.loads((PARTITION_DIR / "package.json").read_text(encoding="utf-8"))["version"]
+    # Google's ranges overlap: a numbered CJK slice also lists the ASCII and punctuation that
+    # co-occur with its characters, and the script slices (latin, latin-ext, ...) come last. A
+    # browser tries overlapping faces last-declared first, so under the Noto CSS a code point is
+    # fetched from the last slice that lists it. Assigning each code point to that slice alone
+    # gives disjoint slices that load the same way, with no glyph shipped twice.
```
hiyouga: **[copy]** The slices aren't disjoint. Their declared ranges overlap on 6 code points per weight, the Bidi-mirrored pairs:
- ≤ / ≥ (U+2264 / U+2265), in slices 100 and 102;
- 〈 / 〉 (U+3008 / U+3009), in 100 and 101;
- ＜ / ＞ (U+FF1C / U+FF1E), in 5 and 101.

fontTools 4.65 adds a requested character's mirror to the cut; on vps3, subsetting U+2264 alone yields a cmap of {U+2264, U+2265}. `unicode-range` is then written from the cut's cmap (`held`), so both slices declare and carry each pair. The effect is harmless, since the browser takes the last-declared face. But "disjoint … with no glyph shipped twice" is false here and in the PR body. Either say so, or write each face's range from the points assigned to its slice (`points & held`), which keeps the declarations disjoint.

changelog/unreleased/2026-09-16-theme-fonts-frost-console.md
```diff
@@ -0,0 +1,67 @@
...
+`@prismshadow/penguin-ui` gained the fonts of all three themes, the token values of its two new
+themes, Frost (`modern`, after sierra.ai) and Console (`geek`, after e2b.dev), and the closed list of
+style hooks the themes give recipes to. Nothing in the Web App names a new theme yet, so the app
+looks as before, apart from a font credit at the foot of the account menu.
+
+## Fonts
+
+- Six fontsource packages, all SIL Open Font License 1.1: Mona Sans and JetBrains Mono (declared
+  for the later Primer polish; Frost already sets code in JetBrains Mono), IBM Plex Sans, IBM Plex
+  Sans Condensed (600 only) and Commit Mono for Console, and Noto Sans SC for Chinese under Primer
```
hiyouga: **[copy]** Three statements in this entry don't match the diff. The `.zh.md` mirrors them at `:18`, `:24` and `:39`.
- **"Noto Sans SC for Chinese under Primer and Console".** This PR leaves Primer on the system CJK stack (the README table says "Noto Sans SC Variable (W1a)"), so Primer's Chinese doesn't change. Write "for Console (and Primer from W1a)".
- **`:27`, "A page downloads only the faces its theme names and the slices its text touches".** That isn't true while the stylesheet inlines slices and carries every family's declarations (see `vite.config.ts:73`).
- **`:46`, "`theme.css` limits what a button, link or tab transitions".** The rule sets a default property list that any `transition-*` utility replaces, as the PR body says, so it limits nothing.

packages/ui/src/fonts/README.md
```diff
@@ -0,0 +1,108 @@
...
+line breaks, and the Kangxi-radical code points the PDF's text layer uses (`⽤` for `用`) written as
+the ideographs they stand for. The sync script never rewrites it; its `--check` confirms that it is
+present and names its licence (`vendored-fonts.json` lists it).
+
+What the licence asks, and where the app does it:
+
+| condition | how it is met |
+| --- | --- |
+| 1. The software states that it uses MiSans. | The Web App's account menu ends with a credit line in both dictionaries (`S.settings.fontCredit`). |
+| 2. No adaptation or redevelopment of the font or its components. | The slices are subsets delivered as WOFF2: every glyph they carry keeps its outline, hinting, metrics and name, every feature and every name record is kept, and `build-misans.py` reads each finished file back and fails unless it matches the official TTF. Whether subsetting itself counts as adaptation is for the licensor to say; it has not been confirmed. |
```
hiyouga: **[copy]** Condition 2 is settled. The maintainer decided on 2026-09-17 that cutting the official TTFs into WOFF2 slices is delivery, not 「改编」. The PR body records the matching decision for condition 3. As written, "it has not been confirmed" reads as an open blocker, and so does item 1 of the PR body's "Still open for the maintainer before merge". Record the decision instead.

Also `:104-105`: `pnpm --filter @prismshadow/penguin-ui fonts:misans <MiSans.zip>` runs in `packages/ui`. So a relative path typed at the repository root, as the sentence suggests, fails with a raw `FileNotFoundError` from `zipfile`. Say the path must be absolute, or drop the alternative.

packages/ui/src/fonts/cjk.css
```diff
@@ -0,0 +1,10 @@
+/**
+ * The CJK face every theme shares: Noto Sans SC, variable `wght` 100–900, OFL-1.1 — fontsource's
```
hiyouga: **[copy]** "The CJK face every theme shares" stopped being true with MiSans. Frost's CJK is MiSans, and Primer stays on the system CJK stack until W1a, so in this PR only Console names Noto. `index.css:6-7` makes the same claim ("a session downloads the active theme's Latin faces plus the Noto Sans SC slices its CJK text touches"). Name the themes that use it.

hiyouga (commented):
Independent review of `911f41f7b` against its base `feat/theme-w0-package` (`dffd7fb41`). I read it against K-redesign §2 and Appendices A/B, the 2026-09-17 user decisions, design Prism-Shadow/penguin-harness-design#175 (`ffb6582`), and the #765 review rounds. I haven't re-raised anything settled there: MiSans at 400/500, bundling with an in-app credit, slicing, dark blocks with only mode-dependent tokens, shared row paddings, 188 names. The open `--ui-panel-p` question is also left out.

**What I checked on vps3.** Before running anything, I matched both trees blob for blob against the two commits.
- **Web builds.** Both build. `web/dist/fonts-licenses/` holds all 7 texts, `misans.txt` included. `release.yml`, the `Dockerfile` and `electron-builder.yml` each copy the whole of `web/dist`, so the licence reaches the npm `web-dist`, the tarball and the desktop app.
- **Where MiSans comes from.**
  - I downloaded the official `MiSans.zip` again from `hyperos.mi.com` (228 MB, Last-Modified 2025-07-14). Both pinned TTF members match their sha256.
  - `build-misans.py --check` rebuilt all 199 files byte-identically, using Python 3.13 with the pinned fonttools 4.65.0 and brotli 1.2.0.
  - The licence PDF matches `licenseSourceSha256`. `LICENSES/misans.txt` equals the PDF's text layer after whitespace and NFKC normalization. The only differences are three radical-supplement code points, written as the ideographs they stand for (⺠→民, ⻛→风, ⻄→西).
- **Slices.**
  - 99 per weight, with the same partition for both weights.
  - Every slice's cmap equals its declared `unicode-range`, 198 of 198.
  - Each weight ships 14,062 code points. The 15,509 it leaves out are all ideographs: 6,555 Ext A, 8,734 URO, 18 compatibility, 202 SIP.
  - Each slice has the source's tables minus `DSIG`. `GDEF` is pruned from the 174 slices that need no glyph classes.
  - The latin slice keeps every feature its glyphs reach: `tnum`, `ss05`/`ss06`, `kern`.
  - `hhea` and `OS/2` vertical metrics and `fsType` equal the source's. `head.flags` gains bit 11, the flag WOFF2 requires.
- **Theme values.**
  - Every Frost and Console value matches K-redesign §2.2/§2.4 or one of the PR's listed deviations. The exception is Frost's `--ui-fg-subtle` (inline, `modern.css:55`).
  - All three base rules declare exactly the 188 contract names.
  - `c7e593d0e` changed no effective value in either mode, and no dark declaration repeats its base value.
- **Hook recipes.** They follow §2.2/§2.4 and Appendix A. Console's glass is a documented no-op. Primer has none yet (W1a).
- **What `contrast.test.ts` measures.** At 4.5:1:
  - `fg` on the four page surfaces and the overlay;
  - `fg-muted` on canvas, surface and surface-muted;
  - the link on canvas and surface;
  - each tone's ink on its own tint;
  - each solid badge's label on its fill;
  - the accent label at rest and on hover, for each theme and all five presets.

  At 3:1, each non-neutral tone's ink as a mark on the four page surfaces. It doesn't measure `fg-subtle`, which it exempts, or the gray bridge that the app's existing classes render through.

**Primer: nothing visible changes except the credit.**
- The base rule `:where(button, a, [role="button"], [role="tab"])` does nothing in today's app. No web element sets a `duration-*` without a `transition-*` utility, none puts that utility behind a variant, and none sets an inline transition. Animations don't read `transition-property`.
- Primer's font stacks name no bundled family, so it requests no font file. It does still download a much larger stylesheet (`vite.config.ts:73`).

**Findings with no line in the diff**
- **[copy]** #175, 06 「主题」 › 字体:
  - 「会话只下载所用主题的、实际用到的分片」 is false while the stylesheet inlines slices (`vite.config.ts:73`).
  - 「分片清单记在共享 UI 包的 `vendored-fonts.json`」: that file records the source package, the slice directory and the licence. The slice list is `misans/misans.css`.
- **[copy]** #175 (06 「主题」 › 令牌契约) states 「控件、列表行与菜单行的高度（纵向内边距与行高）跨主题一致」 with no caveat, and so does `tokens.ts:15-17`. Until W1a, Primer's rem line-heights differ from Frost's and Console's unitless ones. At the 18 px root, a `sm` control is 29 px in Primer and 31.25 px in Frost. Add "from W1a", as the theme file headers already do.
- **[copy]** Two statements in the PR body are out of date:
  - "Primer downloads nothing in W0" is false; see `vite.config.ts:73`.
  - Item 1 under "Still open for the maintainer before merge", whether subsetting counts as adaptation, is settled; see `README.md:93`.

Verdict: fix before merge: every session downloads MiSans and a 5× larger stylesheet (`vite.config.ts:73`), plus the decision on declaring every theme's `@font-face` rules eagerly; Frost's undocumented `fg-subtle` change (`modern.css:55`); the credit's contrast (`user-menu.tsx:137`); the test that misses inline gates (`account-menu.test.ts:174`); the dead exclusions and fallbacks in `build-misans.py`; and the copy fixes (changelog, README, `cjk.css`, `.prettierignore`, #175).

.prettierignore
```diff
@@ -11,6 +11,14 @@ packages/desktop/stage/
 packages/desktop/out/
 packages/desktop/skills/
 
+# font licence texts mirrored verbatim from their packages (scripts/sync-font-licenses.mjs
+# rewrites them; a formatter pass would read as drift)
+packages/ui/src/fonts/LICENSES/
```
hiyouga: **[simplify]** This entry is dead. Prettier has no parser for `.txt`: `prettier --file-info packages/ui/src/fonts/LICENSES/misans.txt` reports `"inferredParser": null`. `prettier --write .` only expands to files it can parse. On vps3, with this entry removed, `prettier --check packages/ui/src/fonts` reports nothing under `LICENSES/`. Delete lines 14–16. The `misans/` entry below stays, since it covers `misans.css`.

packages/web/vite.config.ts
hiyouga: Fixed in f58ac8a3d (and c1db975b0 for the gallery): `build.assetsInlineLimit` refuses `.woff2`, taking the main CSS from 156,168 to 122,323 B gzip -9 with no data URIs; the eager `@font-face` rules stay, per the user's decision, and 414404630 plus design 3874d6b say so (about 94 KB of the 122 KB).

packages/ui/src/themes/modern.css
hiyouga: Fixed in b9ecc81c3: back to `#898683` / `#7f7a74`; the gallery drawer now shows `fg-subtle`'s ratios without judging them, as the contrast suite exempts it (c1db975b0).

packages/web/src/components/layout/user-menu.tsx
hiyouga: Fixed in b26b7d9e6: the credit uses `text-gray-500 dark:text-gray-400`, the account id's ink.

packages/web/test/account-menu.test.ts
hiyouga: Fixed in b26b7d9e6: the test asserts only whitespace and the `{/* … */}` comment before the `<p`, and only whitespace and `</div>` after `</p>`; `{user && <p …}` and an extra node after `</p>` both fail it (checked on vps3).

packages/ui/scripts/build-misans.py
hiyouga: Fixed in b02d4f7bb: both private-use rows, the `"unknown version"` fallback and the worker re-verification are gone (workers get the verified bytes through `initargs`); `--check` still rebuilds all 199 files byte-identically on vps3.

packages/ui/scripts/build-misans.py
hiyouga: Fixed in b02d4f7bb: the comment now says the subsetter adds a requested character's Bidi mirror, so ≤ ≥, 〈 〉 and ＜ ＞ are held and declared by two slices each (harmless, the last-declared face wins).

changelog/unreleased/2026-09-16-theme-fonts-frost-console.md
hiyouga: Fixed in 414404630: Noto is Console's CJK face (Primer's from W1a), the download bullet separates font files from the declarations every user downloads, and `theme.css` is said to set a default transition list, not a limit (both languages).

packages/ui/src/fonts/README.md
hiyouga: Fixed in 414404630: condition 2 records the maintainer's 2026-09-17 decision, and the pnpm form of `fonts:misans` asks for an absolute path; the PR body's open item is marked settled.

packages/ui/src/fonts/cjk.css
hiyouga: Fixed in 414404630: `cjk.css` and `index.css` name Noto as Console's CJK face (Primer's from W1a) and say what each theme downloads.

.prettierignore
hiyouga: Fixed in 414404630: the `LICENSES/` entry is removed; the `misans/` entry stays.

PR #762, opened 2026-09-16
test(ui,web): W0d — token contract, contrast, de-slop and hook guards, roots-aware style guards

hiyouga:

**Review round (2026-09-18, #765's review):** merged the fixed #761 and #765 (c9b434bc2); `token-contract` now checks each theme × mode as the base rule plus that mode's own against the 189 names and fails on a dark declaration that repeats a base value, and `contrast` measures `--ui-fg-on-emphasis` on the neutral emphasis fill (5fa7e85be, changelog df94e03f8).

**Second round (2026-09-18):** merged the stack (f6e99a099); `token-contract` counts 188 names, confirms `--ui-fg-on-emphasis` is gone and holds the theme import order in the app's and the gallery's entry stylesheets (a swapped order fails it, checked on vps3), and `contrast` drops the fg-on-emphasis pair (1f6c4be4f).

**Third round (2026-09-19):** merged #761's review fixes (1c1d09634); nothing in the guards changed, and the ui and web suites pass on vps3.

**Fourth round (2026-09-19):** merged current `main` into the stack (`db316f05e`) and fixed this PR's review (`edb415c1a`). The class-token rules now run over the package's `.ts` files — only 4, 5, 7, 10, 15, 16, 17, 22 and rule 14's eyebrow half wait for the first `.tsx`; the stray-token check exempts a named `FAMILY_PREFIXES` list instead of any spelled prefix; `.ui-display` is judged on `Heading`'s `level` and a hook is attributed to its element's own component; rule 4 reads `xs`/`2xl`/`3xl` and rule 13 reads `text-[11px]/4`; the web policy's pulse home is `skeleton.tsx` alone, with a new test that fails a home naming no file; rule 10's literal-children scope is documented where the check is. The allowlist is re-seeded against `main` by running the suite: **390 hits in 99 files, 169 entries** (`chat-input.tsx` 12 → 6, new `reference-chip.tsx` 12 → 1). On vps3: ui 12 files / 179 tests, web 187 files / 2488 passed / 2 skipped, `pnpm typecheck`, `pnpm lint` 0/0, `pnpm format:check` clean.

Theme system W0, PR 0d, revised in stage K (the de-slop round). It adds the shared UI package's test harness, the **de-slop guards** that keep the tells of the W0 drafts from coming back (gradient atmosphere, radius that eases on hover, a terminal costume, one-hue status boxes, 10–11 px uppercase micro-titles, a tinted stop button), the **closed list of six style hooks** with their hosts, and the **188-name token contract**. The web app's 21 style guards scan `packages/ui/src` as well as `packages/web/src`.

> **Merge note.** To code and test against the revised themes, this branch merged `feat/theme-w0-fonts` (#761) up to `cc7355207` (token names, `hooks.ts`, the Frost and Console revision, MiSans). The diff against #765 therefore includes #761's commits; merging this PR lands them too.

## De-slop guards

`packages/ui/src/testing/deslop.ts` reads Tailwind classes out of the TypeScript AST (string literals and template text, never comments) and out of `@apply` lists, together with each JSX element's classes, direct children and enclosing component, plus CSS declarations. A string counts as classes when it sits in a `className`-like attribute or every word in it is class-shaped, so dictionary prose is never read as classes.

- `packages/ui/test/deslop.test.ts`: every check over `packages/ui/src`, **empty allowlist**. Named homes only: the one Spinner file, the chevron / sheet / launcher / drawer motion files, and the dot / skeleton / streaming-caret pulse. It also checks the theme heading tokens: no h2–h5 is uppercased, and no heading rung is set in mono. A rule with nothing to read yet is a named `PENDING` skip. On this branch that is every markup rule, until the screens (#763), the modules (#764) or the W1 components land.
- `packages/web/test/deslop.test.ts`: the same checks over `packages/web/src` against an allowlist seeded from the app as it is: **390 hits in 99 files, 169 entries** (re-seeded against `main` at `db316f05e`). Each entry is `{ rule: [count, wave] }`. Counts are exact both ways: a new hit fails as new slop, and a removed one fails until its entry shrinks. The list only tightens, and a file that moves into the package leaves stale entries behind and is then judged by the package's empty list. By wave: W1 4, W1b 35, W1b+W7 5, W2 11, W3 9, W4 202, W5 1, W6 59, W7 36, W8 28.

| § 3 rule | guarded | what the check reads |
| --- | --- | --- |
| 1 transitions | yes | `transition-all`, bare `transition`, `transition-transform` off the motion files, `transition-[…]` lists off colour / opacity / shadow; CSS `transition` / `transition-property` with `all` or other properties |
| 2 hover transforms | yes | `scale-*` / `translate-*` under hover, active or focus variants; CSS transforms under `:hover` / `:active` |
| 3 durations | yes | `duration-*` other than 150 / 200 or the fast / base token (300+ only in entrance files); CSS transition durations outside 120–200 ms |
| 4 nested radius | partly | a `rounded-(sm\|md\|lg\|xl)` direct child of a rounded `p-*` element, unless inner / full / control radius |
| 5 clipping box owns the border | yes | a full border on a direct child of a rounded `overflow-hidden\|clip` element |
| 6 status marks | yes | `animate-ping`; `animate-pulse` off `.ui-live` outside the pulse homes; a tinted ring or shadow on a dot |
| 7 icon on its own tint | partly | tone (or hue) tint + ink on an element whose only content is icons |
| 8 left-border callouts | yes (class) | `border-l-2\|4` beside a tone, accent or chromatic border colour; one-per-view is review |
| 9 one-hue box | yes | a tone's tint beside the same tone's line (hue tint + border in the web app) |
| 10 badges | partly | `<Badge>` text that is a mood word or an emoji; two-per-row is review |
| 11 one spinner | yes | `animate-spin` outside `components/icons/spinner/spinner.tsx`; `border-t-transparent` |
| 12 rhythm steps | partly | `gap` / `space-x\|y` / `p` off 1, 1.5, 2, 3, 4, 6, 10; two-steps-per-kind is review |
| 13 text rungs | yes | `text-[Npx]`, `text-[N rem]` |
| 14 uppercase | yes | `uppercase`, `tracking-wide*`; an eyebrow directly above an h1–h4; CSS `text-transform: uppercase` outside eyebrow / display; theme h2–h5 transforms |
| 15 tabular figures | partly | `StatChip` / `Count` / `KeyValue` without `tabular-nums`; call sites are review |
| 16 mono for data | partly | `font-mono` in `Button` / `NavRow` / `Heading` / `MenuItem`; hook recipes forcing mono; theme heading fonts |
| 17 one surface | partly | a bordered `bg-surface*` block nested in another in one file; across components is review |
| 18 decoration | yes | gradient backgrounds; backdrop filters without `.ui-glass`; gradients in hook recipes |
| 19 shadows | yes | anything but `shadow-sm\|lg\|xl\|none` and `shadow-[var(--ui-shadow-*)]`; `drop-shadow` |
| 20 colour | yes (package) | palette classes, `dark:`, hex outside `fixtures/` and `avatar.ts`; not applied to the web app |
| 21 copy | partly | emoji in `strings*.ts` and `fixtures/`; voice is review |
| 22 page header | yes | a `PageHeader` taking or rendering an eyebrow |

## Style hooks

`packages/ui/test/hooks.test.ts` checks K1's `src/hooks.ts`:
- The list holds exactly the six hooks, and each hook has hosts.
- No `ui-*` class and no `.ui-*` selector outside the list, in the package, the web app and the gallery (when it exists).
- Each hook sits only inside its Appendix B hosts, matched by the nearest enclosing PascalCase function, so a stand-in carries its host's name.
- `.ui-live` carries `data-live`, `.ui-display` sits on an h1, and `.ui-frame` slots are head, body, foot or pane.

## Token contract and font licences

- `token-contract`:
  - The contract has 188 names: `--ui-radius-control`, `--ui-stack-0` and `--ui-stack-4` are in, and `--ui-glass-highlight` and `--ui-fg-on-emphasis` are out.
  - `theme.css` bridges `--radius-control`.
  - Every `--ui-*` name that stylesheets or string literals spell in the package, the web app or the gallery is a contract name. This catches a recipe still reading a removed token.
- `font-licenses` follows K1's `vendored-fonts.json` rule, the one `sync-font-licenses.mjs --check` applies:
  - woff2 is allowed only under an entry's `files` directory, and TTF / OTF never.
  - The entry's licence file must contain its `licenseTitle`.
  - Vendored entries are left out of the orphan check.
  - `KNOWN_LICENSE` still checks dependencies only.
- `contrast` is unchanged.

## W0d harness (unchanged from the first draft)

- `packages/ui/test`:
  - `token-contract`, `contrast`, `no-app-strings`, `no-theme-reads`, `demo-coverage`, `font-licenses` and `boot-script`.
  - A case that can't run yet is a named skip, never a pass.
- `packages/ui/src/testing`:
  - `source-roots`, `css`, `theme-tokens`, `color`, `render` and now `deslop`.
  - `test/testing-helpers.test.ts` exercises the helpers.
- `packages/web/test/helpers/roots.ts`: the 21 guards scan both roots, assert that each root yielded files, and assert that each module they read lives in one place.
- `GlyphIcon` strokes at `var(--ui-icon-stroke, 1.7)`.

## Verification

- **Before-run on the unrevised stack.** `integ/theme-w0` at `d963c65f1`, taken with `git archive`, with the new guard files and K1's `hooks.ts` overlaid; vitest on vps3.
  - `packages/ui` **fails**: 18 failed, 51 passed, 2 skipped, **121 hits**.
  - Every hit §5.5 expects is there:
    - all six deleted hooks: 11 selectors and 22 classes
    - wash / grid in the four screens
    - the 8 `ui-pill-hover` sites
    - 6 uppercase: 4 utilities and 2 CSS declarations
    - 19 off-rung text sizes: 13 of them `text-[0.6875rem]`
    - the one-hue approval block (rule 9) and the tinted stop button (rule 7)
  - Also found: 10 hooks outside their hosts, 1 `.ui-display` on an h2, 4 mono hook recipes, 5 hook gradients, 5 hex swatches, 9 off-step spacings, the StatChip without `tabular-nums`, and 8 geek heading tokens.
  - `packages/web` passes: 22 passed, 2 skipped.
- **This branch on vps3** (K1 merged):
  - `packages/ui`: 10 files, **123 passed, 25 skipped**. Every skip is named: 21 de-slop markup rules and 3 hook checks wait for the first `.tsx`, and 1 is demo-coverage.
  - `packages/web`: 176 files, **2,372 passed, 2 skipped**.
  - After the final refactor of the check table: `packages/ui` again 123 passed and 25 skipped, and web `deslop` + `account-menu` 36 passed and 2 skipped.
- **Known shapes.** Each rule has positive and negative samples in `deslop.test.ts`, and the allowlist arithmetic is covered (new, excess, shrink, stale).
- **Local:** `pnpm --filter @prismshadow/penguin-ui typecheck`, `pnpm --filter @prismshadow/penguin-web typecheck`, `pnpm lint`, `pnpm format:check` and `git diff --check` all pass.
- **Post-reboot verification** (2026-09-17, vps3, after `main` `a17323a53` came in through #765): this branch at `e6af829c9` passes ui `deslop` 41 passed / 21 skipped, `hooks` 6 passed / 3 skipped, `token-contract` 17 passed, `contrast` 18 passed, `font-licenses` 7 passed, web `deslop` 22 passed / 2 skipped, and ui and web `typecheck`; on `integ/theme-w0` (#765 + this + #761 + #763, no gallery) the same suites pass except ui `deslop` 9 failed / 51 passed / 2 skipped (47 hits) and `hooks` 3 failed / 5 passed / 1 skipped (33 hits), every hit in #763's unrevised `screens/`; the web allowlist now follows main's web app (`e6af829c9`: 390 hits in 98 files, 168 entries; W1b 35, W4 202).

## Notes

- **Readings of §3 where the text leaves room**, for review:
  - Rule 6: `animate-pulse` is allowed on an element that also carries `.ui-live`. The hook test then confines `.ui-live` to Dot, Spinner, StreamingCaret and ActivityIcon.
  - Rule 17 reads block containers only, so a bordered, filled `<span>` input stand-in is a control, not a card.
  - Rule 18's CSS half reads only hook recipes. The web app's `.context-hatch` stripe carries meaning.
  - Rule 20's hex half also reads `.ts`, and skips `#001`-style labels.
  - Rule 1 refuses size tweens: `transition-[width]` and `transition-[grid-template-rows]`. They are allowlisted in the web app for W4 (CollapsibleSection) and W7 (dock, sidebar).
- **Open questions in the design**, not decided here:
  - §2.0 clamps `--ui-dur-fast` to 120–200 ms, while §2.4 gives Console `fast` 100 ms.
  - §2.1 gives Primer buttons an 80 ms colour transition, which rule 3 would refuse.
  - `theme.css` still bridges `--shadow-2xl`, which rule 19 forbids. There is one web call site, the sheet, allowlisted for W3.
- The frontend skill's "De-slop rules" section (§3) is not in this PR.


packages/web/test/deslop.test.ts
```diff
@@ -0,0 +1,221 @@
...
+  "features/agents/schedules-tab.tsx": { 12: [1, "W4"] },
+  "features/agents/skills-tab.tsx": { 12: [1, "W4"], 13: [2, "W4"] },
+  "features/ai-create/ai-create-panel.tsx": { 13: [1, "W4"] },
+  "features/benchmark/benchmark-case-browser.tsx": { 13: [2, "W4"] },
+  "features/benchmark/benchmark-detail.tsx": { 13: [2, "W4"] },
+  "features/benchmark/benchmark-page.tsx": { 13: [2, "W4"], 18: [1, "W4"] },
+  "features/benchmark/create-benchmark-modal.tsx": { 13: [1, "W3"] },
+  "features/benchmark/evaluation-detail-modal.tsx": { 13: [2, "W3"] },
+  "features/chat/agent-topology-view.tsx": { 12: [1, "W6"], 13: [3, "W6"] },
+  "features/chat/chat-input.tsx": { 7: [1, "W6"], 12: [7, "W6"], 13: [5, "W6"] },
```
hiyouga: **[test]** The allowlist is seeded against the stack's merge-base with `main` (`a17323a53`) and `main` has moved: #782–#784 landed, and two entries no longer match, so `packages/web/test/deslop.test.ts` fails the moment this stack rebases or merges.

- **This entry.** Rule 12 is `[7, "W6"]`, which is right here (`chat-input.tsx` lines 477, 2299, 2525, 2554, 2576, 2600, 2636). #784 moved one of those `p-0.5` chips out, so `main` has six and the suite reports `features/chat/chat-input.tsx: 6 left of 7 — shrink the entry to [6, "W6"]`.
- **The file it moved to.** `packages/web/src/features/chat/reference-chip.tsx` is new on `main` and has no entry; its `:82` `className="shrink-0 rounded p-0.5 …"` is a rule-12 hit, so the suite reports `features/chat/reference-chip.tsx: 1 new`.

That is the guard working exactly as designed — a hit that moves fails on both sides — so it is a merge-time chore, not a defect in the rule. Nothing else in the four merged PRs adds a hit I could reach by grep: `stream-selection-menu.tsx` (245 lines, new) styles through shared class constants and spells no utility at all, and `reference-chip.tsx`'s other classes (`gap-1`, `duration-150`, `py-0.5`/`pl-2`/`pr-1`, bare `rounded`) are all on the allowed lists or outside what the rules read. Re-seed by running the suite against `main` rather than patching by hand, and re-seed again if `main` moves before this merges.

packages/ui/test/deslop.test.ts
```diff
@@ -0,0 +1,550 @@
...
+function nothingToRead(rule: DeslopRule, surface: "markup" | "stylesheet"): string | null {
+  const active = ACTIVE.filter((f) => (surface === "markup" ? MARKUP : /\.css$/).test(f.name));
+  if (surface === "stylesheet") return active.length > 0 ? null : "no stylesheet to read";
+  if (rule === 21) {
+    return active.some((f) => f.rel.startsWith("fixtures/") || /^strings.*\.ts$/.test(f.name))
+      ? null
+      : "no fixtures or strings dictionary yet";
+  }
+  if (!hasTsx) {
+    return "no .tsx under packages/ui/src yet — the screens (#763), the modules (#764) and the W1 components are read when they land";
```
hiyouga: **[test]** `hasTsx` switches off the class-token rules over the `.ts` files the engine can already read.

`nothingToRead` returns this reason for every markup rule but 21 whenever the package holds no `.tsx`. But the markup surface is `/\.tsx?$/`, and most of these rules read class *strings*, not JSX — this suite's own samples prove it in `.ts`: `found("parts.ts", 'export const LABEL = "text-xs uppercase";')` → `14:uppercase` (:216), and `found("toast.ts", …)` → `9:red: tint and line on one box` (:359).

`ACTIVE` holds five readable `.ts` files today — `boot.ts`, `fixtures/index.ts`, `index.ts`, `tokens.ts`, `vite-plugin.ts` — and rules 1, 2, 3, 11, 12, 13, 14, 18, 19 and 20 are all reported PENDING over them. That includes rule 20's hex half, whose documented home is `hexHomes: ["fixtures/", "avatar.ts"]` (:64) — one of the five files that exist. A W1 `parts.ts` of class recipes, or #764's modules, can land class strings before the first `.tsx` and go unguarded while the suite says PENDING.

Gate on what each check needs: skip on `!hasTsx` only the rules that read elements or components — 4, 5, 7, 10, 15, 16, 17, 22 and rule 14's `eyebrowSiblingCheck` half — and let the token checks run over the `.ts` files that exist, the way rule 21 already does with its own reason at :84-88.

packages/ui/test/token-contract.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  });
+
+  /**
+   * Names the file spells that are not tokens. A family prefix (`startsWith("--ui-chart")`) names
+   * tokens that exist; a removed or misspelt token prefixes none.
+   */
+  const strays = (file: SourceFile) =>
+    spelled(file).filter(
+      (found) =>
+        !contract.has(found.name) && !TOKEN_NAMES.some((t) => t.startsWith(`${found.name}-`)),
```
hiyouga: **[test]** The family-prefix skip exempts 56 names that are not tokens.

The doc above says "a removed or misspelt token prefixes none". That does not hold: the skip fires for any spelled name that is a proper prefix of a contract name, and the 188 names produce 56 such prefixes —

`--ui-border, --ui-card, --ui-chart, --ui-code, --ui-control, --ui-diff, --ui-dur, --ui-ease, --ui-focus, --ui-font, --ui-glass, --ui-h1 … --ui-h6, --ui-icon, --ui-live, --ui-md, --ui-menu, --ui-panel, --ui-radius, --ui-row, --ui-scrollbar, --ui-selection, --ui-shadow, --ui-stack, --ui-text, --ui-tone, --ui-tracking, --ui-weight`, and the two-level ones (`--ui-tone-danger`, `--ui-text-code`, `--ui-control-px`, …).

So a recipe reading `var(--ui-shadow)`, `var(--ui-radius)`, `var(--ui-dur)` or `var(--ui-glass)` resolves to nothing and passes this check — which is the exact failure the block was written for. `--ui-glass` is the family the removed `--ui-glass-highlight` belonged to.

Make the exemption a named list instead of a derived one: `const FAMILY_PREFIXES = ["--ui-chart"];` and skip only those. The probe at :235-245 still passes (`--ui-chart-*` and `startsWith("--ui-chart")` are the only two cases it covers), and every other prefix-shaped read becomes a stray.

packages/ui/test/hooks.test.ts
```diff
@@ -0,0 +1,233 @@
...
+      analyzeFile(file).elements.flatMap((element) => {
+        const names = new Set(element.classes.map((t) => t.utility));
+        const at = `${file.id}:${element.line} <${element.tag}>`;
+        const problems: string[] = [];
+        const live = element.attributes.get("data-live");
+        if (names.has("ui-live") && live !== null && !LIVE_SIGNALS.has(String(live))) {
+          problems.push(`${at} .ui-live needs data-live="dot|caret|spinner", has ${String(live)}`);
+        }
+        const level = element.attributes.get("aria-level");
+        if (names.has("ui-display") && element.intrinsic && element.tag !== "h1" && level !== "1") {
```
hiyouga: **[test]** `.ui-display belongs on an h1` cannot fire on the host most likely to get it wrong.

`element.intrinsic` limits the check to `<div>`, `<h2>`, … so `<Heading level={2} className="ui-display">` is never looked at — and `Heading` is `ui-display`'s own Appendix B host (`HOSTS` :45) and takes the level as a prop, which makes it the likeliest way to put the display face on something that is not a page title. `deslop.ts:893-898` already has `headingLevel()`, which reads `Heading`'s `level` attribute as well as `h1`–`h6`; using it here judges a component host too, and `intrinsic` can stay for the tags it cannot resolve.

Second, smaller: `componentOf` (:87-91) approximates "the nearest enclosing PascalCase function" by sorting candidates on `classes.length`. When an outer component's only class tokens are the ones written inside a nested one, the two lists are equal and the stable sort returns the outer name:

```tsx
export function Tooltip() {
  const Bubble = () => <div className="ui-glass rounded-md p-2" />;
  return <Bubble />;
}
```

`Tooltip.classes` and `Bubble.classes` both hold the same three tokens, so the hook is attributed to `Tooltip`. `analyzeFile` already computes the true nearest host on every element (`JsxElementInfo.component`, `deslop.ts:419-426`); for a hook written in a class attribute that is the value to read, with the length sort left only for a loose module-scope token.

packages/ui/src/testing/deslop.ts
```diff
@@ -0,0 +1,1276 @@
...
+const isHexColour = (hex: string) =>
+  hex.length > 5 || /[a-f]/i.test(hex) || new Set(hex.slice(1)).size === 1;
+
+/** The components §3 names: tabular figures (15), mono-free labels (16), no eyebrow (22). */
+export const TABULAR_COMPONENTS: readonly string[] = ["StatChip", "Count", "KeyValue"];
+export const LABEL_COMPONENTS: readonly string[] = ["Button", "NavRow", "Heading", "MenuItem"];
+export const PAGE_HEADER_COMPONENTS: readonly string[] = ["PageHeader"];
+
+const BORDER_WIDTH = /^border(?:-(?:2|4|8))?$/;
+const ROUNDED_STEP = /^rounded-(?:sm|md|lg|xl)$/;
```
hiyouga: **[test]** Rule 4 does not recognise the radius steps `theme.css` deliberately bridges.

`ROUNDED_STEP` is `sm|md|lg|xl`, so neither the parent test (`nestedRadiusCheck` :832) nor the child test (:835-836) sees `rounded-2xl`, `rounded-3xl` or `rounded-xs`. `theme.css:128` bridges `--radius-xs`, and `:136-137` bridges `--radius-2xl` and `--radius-3xl` onto the theme's own scale — with a comment saying that is so "a 2xl/3xl keeps following the theme's radius, including all the way down to 0". The web app spells `rounded-2xl` three times and `rounded-3xl` once. A `rounded-2xl p-4` card holding a `rounded-md` child is precisely the violation the rule names, and it produces no hit.

None of today's four sites is a padded parent with a rounded child, so nothing is currently hidden. Rule 5 is unaffected — `clippedBorderCheck` (:852) uses the wider `^rounded(?:-(?!none$|0$).+)?$`.

Same shape one rung down, if it is worth closing in the same pass: `textSizeToken` (:707) ends the pattern at `]`, so Tailwind's line-height modifier `text-[11px]/4` escapes rule 13. Zero occurrences in `packages/web/src` today.

packages/web/test/deslop.test.ts
```diff
@@ -0,0 +1,221 @@
...
+import { expectEveryRootScanned, scanSources } from "./helpers/roots";
+
+const SCAN = scanSources();
+const WEB = SCAN.files.filter((file) => file.root === "web");
+
+/** The homes §3 names, as the web app spells them today. It has no Spinner of its own: W1 adds it. */
+const POLICY: DeslopPolicy = {
+  transformMotion: ["chevron.tsx", "sheet.tsx", "drawer.tsx", "dock-launcher.tsx"],
+  entranceMotion: [],
+  pulseHomes: ["dot.tsx", "skeleton.tsx", "streaming-caret.tsx"],
```
hiyouga: **[over-defensive]** Two of these three name a file the web app does not have.

`packages/web/src` holds `components/ui/skeleton.tsx`, but no `dot.tsx` and no `streaming-caret.tsx` — and the comment above this policy says it is "the homes §3 names, **as the web app spells them today**". `matchesPolicyPath` matches a bare name anywhere under the root, so each of the two is a standing exemption waiting for a file: the first `features/**/dot.tsx` to carry `animate-pulse` on something that is not live passes rule 6 silently, with no allowlist entry and no wave attached.

W1 builds `Dot` and `StreamingCaret` in the package, where `packages/ui/test/deslop.test.ts:59` already names both. Delete them here; `skeleton.tsx` is the only one the app currently needs.

(`:181` asserts that every allowlisted file still exists, but nothing asserts it of the policy homes. The package's own POLICY is legitimately forward-looking — this is about the web list only.)

packages/ui/src/testing/deslop.ts
```diff
@@ -0,0 +1,1276 @@
...
+/** Rule 10: a `<Badge>` whose literal text is a mood word or an emoji. */
+const moodBadgeCheck: Check = (analysis) =>
+  analysis.elements.flatMap((element) =>
+    element.tag !== "Badge"
+      ? []
+      : element.children.flatMap((child) => {
+          if (child.kind !== "text") return [];
+          const word = child.text.trim().toLowerCase();
+          const mood =
+            MOOD_WORDS.has(word) && !(word === "beta" && element.component === "BetaBadge");
```
hiyouga, line 893: **[test]** Rule 10 guards nothing in this app, and this carve-out is dead.

`moodBadgeCheck` reads only a `<Badge>`'s literal text children. Every `<Badge>` in `packages/web/src` takes an expression — `{S.models.freeBadge}`, `{S.company.tickets.blocked}`, `{S.company.orgPaused}`, `{stopReason}`, `{priority}`, … — the single literal in the app is `origin`. The `BetaBadge` exception on this line is dead for the same reason: `features/company/beta-badge.tsx:60-66` renders a `<span>` holding `{S.company.beta}`, not a `<Badge>`, so no `<Badge>beta</Badge>` exists to excuse.

K-redesign §3 rule 10 scopes the guard exactly this way ("`Badge` children that are **literal strings** from a `MOOD_WORDS` list (`new`, `hot`, `popular`, `beta` outside `BetaBadge`)") and marks it "partly", so this is the design's shape rather than a coding slip. But in an app whose badge copy lives in `lib/strings.ts` / `lib/strings-en.ts`, a mood word arrives as a dictionary value — the files rule 21 already reads. Worth settling with the user before the PR body's table counts rule 10 among the guarded ones.

changelog/unreleased/2026-09-16-theme-w0-guards.md
```diff
@@ -0,0 +1,88 @@
...
+  file and no `border-t-transparent`; gaps, stacks and padding on the rhythm steps; no literal text
+  sizes; `uppercase` and wide tracking only through the eyebrow and display hooks, and no eyebrow
+  directly above a heading; tabular figures in stat chips, counts and key-value values; no mono in
+  button, nav, heading or menu labels; no bordered surface inside another; no gradient or backdrop
+  blur without a job; shadows from the shadow tokens; colours from tokens; no emoji in dictionaries
+  or fixtures; no eyebrow on a page header.
+- `packages/ui/test/deslop.test.ts` runs every check over `packages/ui/src` with an empty allowlist,
+  and checks the theme files' heading tokens: no h2–h5 is uppercased and no heading is set in mono.
+- `packages/web/test/deslop.test.ts` runs the same checks over `packages/web/src` against an
+  allowlist of the 374 hits the app held in 96 files, counted exactly per file and rule, each entry
```
hiyouga: **[copy]** The allowlist figures are two rounds out of date. The list at head holds **390 hits in 98 files, 168 entries** — the PR body's own post-reboot note gives the same 390 / 98 / 168, and I counted it from `packages/web/test/deslop.test.ts`. The `.zh.md` repeats 374 / 96 at `:51` (「现有的 374 处命中，分布在 96 个文件中」), so both files need 390 and 98.

The by-wave table in the PR description is stale the same way: W1b 33 → 35 and W4 188 → 202, which is where the 374 → 390 difference sits.

hiyouga (commented):
Independent review of `1c1d09634`, read as its own incremental diff against `origin/feat/theme-w0-fonts` (`414404630`) — #761's and #765's content is excluded, and nothing settled in their reviews is re-raised. Read against K-redesign §3 and Appendices A/B, `K-COMMON.md`, A-architecture §7/§7.2, and design PR Prism-Shadow/penguin-harness-design#175.

**Do the guards bite?** I reproduced five rules' hit counts against the branch tree with independent greps, and every one matched the allowlist exactly:

- rule 13: 175 `text-[Npx|rem]` = 175 allowlisted.
- rule 11: 28 `animate-spin` / `border-t-transparent` = 28.
- rule 14: 26 — and the 10 occurrences it correctly drops are `//` comments, a `{/* … */}` JSX comment, a prop named `uppercase` and its type member. Prose is not read as classes.
- rule 1: 17 — `transition-[border-color,box-shadow]` (`field.tsx`, `chat-input.tsx`) and `dock-launcher.tsx`'s four-property list are correctly allowed, `chevron.tsx`'s `transition-transform` correctly homed, `styles.css`'s `transition: none !important` correctly ignored.
- rule 19: 5 — `shadow-lg/sm/xl` allowed, `shadow-2xl` and the four `shadow-[…]` caught, and the two bare `shadow` words in comments not counted.

The counting arithmetic (new / excess / shrink / stale) is covered by a unit case, every parser has positive *and* negative probes, and `unscannedRoots` plus the per-root "spells at least one `--ui-*`" assertion close the "guard scans nothing" hole the design names as this PR's risk. All 23 web test files that import `helpers/roots` call `expectEveryRootScanned`.

**Contrast** is computed correctly: WCAG linearisation and coefficients, the ratio formula, per-layer compositing down `layerBelow`, a translucent `fg` composited over its backdrop, and a throw rather than a guess when the backdrop is not opaque. The skipped pairs match A-architecture §7.2 (`fg-subtle` exempt as placeholder/disabled ink; `neutral` exempt from the 3:1 tone-ink floor, but still held to 4.5:1 on its own tint). The oklch path clamps per channel rather than gamut-mapping, which can differ from the browser by a fraction of a ratio on an out-of-gamut value — not worth changing.

**Token contract.** The parser handles what it needs to: comments stripped with newlines preserved, `;` inside `url()`/`rgb()` not split, a missing final semicolon, `@layer`/`@supports`/`@media` nesting (a token outside the two canonical rules is a `structure` failure, and one inside the right selectors but the wrong at-rule is too), attribute-quote normalisation, `@keyframes` steps parsed as rules. Names built at runtime are correctly dropped on both sides of an interpolation. `TOKEN_NAMES` really is 188 unique names and the "each name once" test would catch a duplicate. The one weakness is the family-prefix skip — finding 3.

**Findings** (all inline): 1 stale web allowlist vs `main`, 2 the `.tsx` gate over `.ts` files, 3 the prefix skip, 4 `.ui-display` and host attribution, 5 rule 4's radius steps, 6 dead pulse homes, 7 rule 10, 8 the changelog counts.

No `try`/`catch` anywhere in the new modules and nothing else over-defensive worth deleting.

Verdict: fix before merge: the web allowlist against today's `main` (`web/test/deslop.test.ts:97`), the `hasTsx` gate that turns off the class-token rules (`ui/test/deslop.test.ts:90`), the prefix skip that exempts 56 non-tokens (`token-contract.test.ts:220`), and the changelog's 374/96 (`2026-09-16-theme-w0-guards.md:62` and `.zh.md:51`). Findings 4–7 are judgement calls on how far each rule should reach.

packages/web/test/deslop.test.ts
hiyouga: Re-seeded by running the suite against `main` (merged into the stack first, `db316f05e`): `features/chat/chat-input.tsx` rule 12 is `[6, "W6"]` and the new `features/chat/reference-chip.tsx` carries `{ 12: [1, "W6"] }`. Nothing else moved. The list is now **390 hits in 99 files, 169 entries**, which is what both changelog files and the PR body's wave table say. edb415c1a

packages/ui/test/deslop.test.ts
hiyouga: Gated on what each rule reads: only 4, 5, 7, 10, 15, 16, 17 and 22 skip on `!hasTsx`, and rule 14's eyebrow-sibling half gets a named skip of its own while its class-token half runs. Over this branch's five `.ts` files that turns fourteen PENDING rules into live ones (1, 2, 3, 6, 8, 9, 11, 12, 13, 14, 18, 19, 20, 21), rule 20's hex half included. edb415c1a

packages/ui/test/token-contract.test.ts
hiyouga: Now a named list: `const FAMILY_PREFIXES = ["--ui-chart"]`. The probe gained `var(--ui-shadow)` as a case — a name that prefixes five tokens and resolves to nothing, which the derived skip passed. edb415c1a

packages/ui/test/hooks.test.ts
hiyouga: Both taken. `headingLevel()` is exported and the recipe check judges `<Heading level={2} className="ui-display">` as well as the intrinsic tags; `componentOf` reads the element's own `component` (the analyzer's nearest host) and keeps the length sort only for a token in a loose string. The recipe check moved into `recipeProblems(file)` so both cases are pinned on known shapes, your `Tooltip`/`Bubble` one included. edb415c1a

packages/ui/src/testing/deslop.ts
hiyouga: `ROUNDED_STEP` is `xs|sm|md|lg|xl|2xl|3xl` — every step `theme.css` bridges — and `textSizeToken` accepts Tailwind's line-height modifier (`text-[11px]/4`). Neither adds a hit in the app or the package today. edb415c1a

packages/web/test/deslop.test.ts
hiyouga: `pulseHomes` is `["skeleton.tsx"]` — `Dot` and `StreamingCaret` are W1's, in the package, where `packages/ui/test/deslop.test.ts` already names them. A new test in the same suite fails any policy home that names no file under `packages/web/src`, so the next standing exemption cannot sit there unnoticed. edb415c1a

packages/ui/src/testing/deslop.ts
hiyouga: Left as §3 scopes it, and said out loud where the check is: only literal `<Badge>` children are read, a badge whose copy comes from a dictionary is out of reach here, and the dictionaries themselves are rule 21's (for emoji). So in this app the rule guards the hard-coded badge only. edb415c1a

changelog/unreleased/2026-09-16-theme-w0-guards.md
hiyouga: Both changelog files now read 390 hits in 99 files, and the PR body's by-wave table is recomputed from the list at head (W1 4, W1b 35, W1b+W7 5, W2 11, W3 9, W4 202, W5 1, W6 59, W7 36, W8 28). edb415c1a

PR #763, opened 2026-09-16
feat(ui): W0e — fixtures and de-slopped full-screen compositions for the gallery

hiyouga:

**Review round (2026-09-18, #765's review):** merged the fixed stack through #762 (4e32088ca, 17cf287cf); the `/screens` alias entry went with the alias layer, and the `./screens` export stays.

**Second round (2026-09-18):** merged the stack (87df719c3); the settings screen's switch knob reads `--ui-tone-neutral-emphasis-fg` now that `--ui-fg-on-emphasis` is gone.

**Third round (2026-09-19):** merged #761's review fixes through #762 (a84c0a3d2); no change of its own.

**Fourth round (2026-09-19):** merged current `main` and #762's review fixes (`3330081ec`), and fixed this PR's review (`8618fc527`). The catalog rows follow `main`'s refresh — `DeepSeek V4 Pro 0813`, and the retired Flash row gives way to the live `deepseek-flash` / `DeepSeek V4.1 Flash` — and `fixtures-screens.test.ts` now also fails a row the catalog has retired. The screens print what the product prints: the stats line leads with the reply's time and carries the app's five numbers, the Trace's Overall shows Avg tools / turn, the composer's ring fills against the compaction threshold (68%, not a 1 px dash), a file size reads `47.1KB`, the login's language switch has the app's three options, and the dock, sidebar and accent labels are the app's. Dead code went with them (the approval block's try/catch, the `?? "?"` fallbacks, the login form's `onSubmit`, `Turn`'s silent clamp), `Dot` takes a tone, `Badge` takes §5.1's size, and a preference row's "?" carries "More info: <label>" with the explanation in its popover. On vps3: ui 11 files / 172 passed / 3 skipped, ui typecheck.

Row E of the theme-system W0 breakdown, revised in stage K: the screens are now held to the
de-slop rules the theme work adopted (K-redesign §1.3 rows 14–23), and the fixtures carry the sets
the module gallery composes (§4.6).

## What

- **`packages/ui/src/fixtures/`** is one typed mock dataset in `en` and `zh`.
  - `dataset.ts` holds the structure (ids, states, numbers, order) once. The locale files supply prose only, so the two cannot drift in shape.
  - Session: the landing page's "Build Claude Code docs expert" story, extended to two turns:
    - thinking, and assistant text that is still streaming
    - `exec_command` with output, `read_file`, and `edit_file`/`write_file` with unified diffs
    - `run_subagent` carrying the child's own transcript
    - a command waiting for approval, since the composer is in read-only mode
  - Trace: the file list, overall figures, and per-turn timeline lanes with event rows.
  - Models: ten rows copied from `model-catalog.ts`. A test checks each row still exists there.
  - Also: the Workspace tree and a file preview, sidebar session groups, and company mode (employees, tickets in all five columns, a week of calendar events).
  - **New in stage K (§4.6):** `notices` (one per tone plus a toast pair), `forms` (field values with one invalid field, one held by a server policy, and the accent swatches), `menus` (a message context menu with a shortcut and one destructive item), `vault` (six rows), `plugins` (four installed), `commandPalette` (two groups) and `usage` (7 days × 3 buckets). Colours, versions, icons, shortcuts and series are locale-independent and live in `fixtures/shared.ts`; only words are per locale.
  - Chrome copy is taken from the web app's two dictionaries, plus `chat.runStates` (the five run-state words), the icon-button labels the screens now require, and `traces.fromSubagent`.
  - Type specimens (display, heading, paragraph, UI line, caption, code, numerals) mix CJK and Latin text with code and numbers, for `/fonts` and the Typography foundation.
- **`packages/ui/src/screens/`** holds the four compositions the architecture names (§3.14): `chat`, `traces`, `settings` and `login`.
  - They are static JSX on token utilities and the six remaining style hooks only, with no palette classes and no `dark:`. A test enforces this.
  - Exported as `@prismshadow/penguin-ui/screens` (`SCREENS`, `SCREEN_IDS`, `screenById`). Each entry is `{ id, title, description, Component: ({ lang }) }`, so the gallery's `/screens/:name` route renders `screenById(name).Component`.
  - Stand-in pieces (`glyph.tsx`, `markdown.tsx`, `parts.tsx`, `transcript.tsx`) each name the component and wave that replaces them.
- `package.json` exports gain the `/screens` subpath.

## De-slop (stage K)

- The five deleted hooks leave the markup: no `.ui-wash` / `.ui-grid` behind the chat, the settings
  page or the Trace panel, no `.ui-grid-dots` behind the login (the circuit traces stay — the
  page's one decoration, in every theme), no `.ui-ticks` on the approval block, no `.ui-pill-hover`
  on buttons and cards, no `.ui-button` mono uppercase labels. Controls read `--ui-radius-control`
  through `rounded-control` instead, and `data-shape="card"` is gone with the hook that read it.
- Each remaining hook sits on the host Appendix B names, so the `hooks` suite's host check passes:
  `Modal` and `FloatingPanel` own the settings glass, `ComposerCard` the composer's, `PagedDialog`
  the rail's eyebrows, `GroupHeader` the sidebar's and the call graph's, `DiffViewer` and
  `CodeBlock` their frames, `Spinner` its live signal. The settings page title is an ordinary `h2`
  without the display face (which belongs to a page's one display title) or the sun glyph the rail
  already shows.
- Status is a word again: sentence case at the caption rung in its tone's ink, from the new
  `chat.runStates` in both dictionaries. No `uppercase` or `tracking-wide` utility is left on the
  screens, and nothing is set below `.75rem` — the code block reads its size from
  `--ui-text-code-size` rather than `text-[0.8125rem]`.
- Nesting and one-hue boxes: the pending-approval block keeps its amber fill but takes the neutral
  rule above it and drops the alias chip for mono text; the subagent shortcut becomes a row under
  its call instead of a bordered card inside the work group; the Trace's Overall is a ruled section
  and its timeline sits in the turn's body, neither in a box of its own; the composer's stop is a
  solid `tone-danger-emphasis` control instead of a danger glyph on a danger tint.
- Spacing sits on the rhythm steps (`gap-0.5`, `p-0.5`, `gap-2.5`, `gap-x-8` are gone).
- `Pill` → `Badge` and `IconSquare` → `IconButton`, in the shapes W1's components take, so the
  later swap is a rename: `Badge { tone, variant }` with `soft` taking the neutral line;
  `IconButton { icon, label, size, pressed }`, flat at rest and filled only when pressed, with the
  label required.

### One W1 component landed early

Rule 11 gives `animate-spin` exactly one home, `src/components/icons/spinner/spinner.tsx`, and the
screens need a spinner now — so the real `Spinner` is in that file: an SVG arc at `xs` / `sm` / `md`
(10 / 12 / 14), `tone` or `inherit`, a required `label`, and `.ui-live[data-live="spinner"]` so a
theme re-times it (Console steps it eight times a turn) without redrawing it. `screens/parts.tsx`
re-exports it, so W1 deletes one line rather than rewriting call sites.

Its `spinner.demo.tsx` is deferred to W1 through `demo-coverage`'s exemption list, because the
gallery reaches a demo through a module's `parts` and those modules land with the component set.

## Notes for the other W0/W1 PRs

- **#764 (gallery):** `modules/parts.tsx:185` calls `<Spinner size={12} />`; the W1 signature takes
  `size="sm"`. One token to change when the branches meet.
- **#764 (gallery):** the module-local fixtures this PR replaces — `status.module.tsx`'s
  `LOCAL.notices`, `overlays.module.tsx`'s `LOCAL.menus.message`, `LOCAL.toasts` and `LOCAL.palette`,
  `tables.module.tsx`'s `LOCAL.vault` and `LOCAL.plugins.rows`, `pages.module.tsx`'s
  `LOCAL.plugins.rows`, `stats.module.tsx`'s `LOCAL.usage`, and `forms.module.tsx`'s `FORMS` field
  values and swatch labels — can now read `f.notices`, `f.menus`, `f.commandPalette`, `f.vault`,
  `f.plugins`, `f.usage` and `f.forms`.
- **B:** Console's `--ui-tone-neutral-bg` is `transparent`. The screens therefore draw neutral wells
  (bubbles, chips, segmented tracks, timeline lanes) with an ink-derived fill (`NEUTRAL_FILL`). The
  contract has no opaque neutral-fill token; worth deciding in W1/W2.

## Verification

vps3 (`vps3-vitest.sh`), on this branch with `feat/theme-w0-guards` and `feat/theme-w0-fonts`
merged in:

- `pnpm --filter @prismshadow/penguin-ui test`: 11 files, **164 passed | 3 skipped**. That includes
  `deslop` (62 | 2 skipped) and `hooks` (9 | 1 skipped) with **zero hits in
  `packages/ui/src/screens/`** — 47 de-slop hits and 33 hook hits before this change — plus
  `token-contract` (17), `contrast` (18), `font-licenses` (7), `demo-coverage` (3) and
  `fixtures-screens` (18).
- `pnpm lint` (oxlint): 0 warnings, 0 errors. `pnpm format:check`: clean. `pnpm typecheck`: clean.
- `git diff --check`: clean.


hiyouga (commented):
Reviewed the incremental diff only (`origin/feat/theme-w0-guards` `1c1d09634` … `a84c0a3d2`): 22 files, the fixtures, the four screens, the early `Spinner`, `fixtures-screens.test.ts` and the changelog pair. Settled points from #765 and #761 are not re-raised.

**What holds.** All twelve style-hook sites land on an Appendix B host and carry the markup their recipes select on — `ui-glass` in `Modal` / `FloatingPanel` / `ComposerCard`, `ui-eyebrow` in `PagedDialog` / `GroupHeader`, `ui-display` on `LoginScreen`'s `h1`, `ui-frame` in `CodeBlock` / `DockFrame` / `DiffViewer` / `WorkGroup` with `head` / `body` slots, `ui-live` in `Spinner` and `StreamingCaret` with `data-live`. The de-slop fixes are real style changes: `deslop.test.ts` is untouched, `ALLOWLIST` is still `{}`, `POLICY.hexHomes` already carried `fixtures/` on the guards branch, and the only new exemption is `demo-coverage`'s `icons/spinner` entry, which carries its reason and is discharged by a file, not by a rule. Nothing leaks into the app: `packages/ui/src/index.ts` exports only `hooks` and `tokens`, `./screens` and `./fixtures` are separate subpaths, no app file imports either, and `packages/web/src/styles.css:30-31` already excludes both from Tailwind's sources. Every `--ui-*` name the screens read is in the contract; `bg-surface-inset` resolves through the `--ui-inset` alias. The fixtures hold no personal data, the `ericbuess` corpus URL and `127.0.0.1:7890` are both already established elsewhere in the repo, the sizes are computed or plausible, the zh prose follows the CJK/Latin spacing and full-width punctuation rules, and the internal arithmetic checks out (turn totals sum to `overall`, `1m12s` matches `elapsedMs`, the diff hunk's line numbers match `RAG_TS_BEFORE`). The changelog pair is complete and correctly ordered.

**What does not.** One finding is a merge blocker; the rest are honesty and dead-code items, all inline.

The blocker: this branch's merge base is `a17323a53`, and `main` is 9 commits ahead, including #749 (`11c8cd05c`), which renamed the catalog's `deepseek-v4-pro` to `"DeepSeek V4 Pro 0813"` and retired `deepseek-v4-flash`. `fixtures-screens.test.ts`'s catalog check passes against this branch's copy and fails against `main`'s — I ran the test's own regex against both.

The honesty items all point the same way: five of them are places where a mock-up shows something the product does not. The per-turn stats line and the context ring are the two a theme reviewer would actually be looking at.

Two claims to settle before merge, both stated in `fixtures/types.ts` and the changelog: that the chrome copy is "taken from the Web App's two dictionaries" and that `ACCENT_SWATCHES` are "the accent presets `theme.css` declares". Neither is guarded and both have already drifted (inline). If the intent is that the gallery keeps its own wording, say so in the comment instead — the claim is what makes the drift a defect.

Verdict: fix before merge: the catalog rename (`fixtures/shared.ts:335`) will break CI on `main`; the stats line (`transcript.tsx:89`), the context ring (`dataset.ts:387`), the Overall row (`traces.tsx:59`), the login language switch (`login.tsx:73`) and the dock tab (`chat.tsx:113`) show what the product does not; the KB figure contradicts itself (`format.ts:38`); the dead fallbacks and the unreachable `try`/`catch` (`transcript.tsx:227`, `:409`, `:490`, `parts.tsx:278`, `login.tsx:101`) should go.

packages/ui/src/fixtures/shared.ts
```diff
@@ -0,0 +1,575 @@
...
+  output,
+});
+
+/** Catalog rows without their localized notes; each locale adds `note` where it has one. */
+export const MODEL_ROWS: readonly Omit<ModelFixture, "note">[] = [
+  {
+    provider: "deepseek",
+    providerLabel: "DeepSeek",
+    modelId: "deepseek-v4-pro",
+    displayName: "DeepSeek V4 Pro",
```
hiyouga: **[bug]** This row will fail `fixtures-screens.test.ts` the moment the branch meets `main`.

The merge base is `a17323a53`; `main` is 9 commits ahead and one of them is #749 (`11c8cd05c`), which renamed the catalog row:

```
modelId: "deepseek-v4-pro",
displayName: "DeepSeek V4 Pro 0813",
```

`fixtures-screens.test.ts:87-95` builds `modelId … \s* displayName … \s* provider … \s* contextWindow` and requires all four adjacent, so `deepseek/deepseek-v4-pro` no longer matches. I ran the test's own regex over both catalogs: 10/10 pass against this branch's copy, 9/10 against `main`'s, and the one failure is this row.

The same refresh also gives `deepseek-v4-flash` `retired: true`, so the fixture's second row is a model the product no longer offers — and `en.ts:314` / `zh.ts:307` advertise it ("Model added — DeepSeek V4 Flash is ready to use").

Fix after rebasing on `main`: `displayName: "DeepSeek V4 Pro 0813"`, and swap the Flash row for a live one (`deepseek-flash` / "DeepSeek V4.1 Flash" is the successor in the same group), then re-run that test against `main`'s catalog.

packages/ui/src/screens/transcript.tsx
```diff
@@ -0,0 +1,570 @@
...
+  );
+}
+
+/** The per-turn footer under a settled reply (the app's TaskStatsLine). */
+export function StatsLine({ stats, f }: { stats: TurnStats; f: Fixtures }) {
+  const t = f.copy.traces;
+  const input = stats.inputTokens;
+  return (
+    <div className="-mt-1 mb-3 flex h-5 items-center gap-3 whitespace-nowrap text-fg-subtle">
+      <StatChip glyph="wrench" value={String(stats.toolCalls)} title={t.toolCalls} />
```
hiyouga: **[copy]** The stats line shows two numbers the app deliberately does not put here, and drops the one it leads with.

`packages/web/src/features/chat/task-stats-line.tsx:140-190` renders, in order: the reply timestamp, input tokens (bare `humanizeTokens(input)`), output tokens, TPS, cost, elapsed — then copy and fork. There is no tool-call chip anywhere on that line (`STAT_ICONS.toolCalls` appears only in the Traces view and the Agents page), and no cache percentage. Its header comment says so outright (`:21-24`): the line "doesn't break down the cache composition — … belongs to the Trace page's turn card", and the timestamp "leads: it's this reply's identity (when it was said), the stat numbers are an annotation".

This mock adds `wrench 2` and turns input into `15.6k (79%)`, and omits the timestamp. That is the most-photographed row of the chat screen, so a reviewer judging its density and figure rhythm would be judging a line the product does not have.

Fix: drop the wrench chip and the `(NN%)` parenthetical, and lead with the reply time — every text item already carries `atIso`, and `fixtures/shared.ts:28` already has `clock()`.

packages/ui/src/screens/traces.tsx
```diff
@@ -0,0 +1,286 @@
...
+  const o = f.trace.overall;
+  const t = f.copy.traces;
+  return (
+    <section className="border-t border-line pt-3">
+      <p className="mb-2 text-xs font-(--ui-weight-strong) text-fg-muted">{t.overall}</p>
+      <div className="grid grid-cols-3 gap-x-6 gap-y-1">
+        <div>
+          <SummaryRow label={t.turns} value={String(o.turns)} />
+          <SummaryRow label={t.toolCalls} value={String(o.toolCalls)} />
+          <SummaryRow label={t.compactions} value={String(o.compactions)} />
```
hiyouga: **[copy]** The app's Overall has no Compactions row, and the row it does have is missing.

`packages/web/src/features/traces/trace-file-view.tsx:521-526` — the counts column is Turns / Tool calls / **Avg tools / turn**, with a comment explaining that the average is exactly the two rows above it divided, so a reader can check the division by eye. Compaction is surfaced as a per-turn badge, not as a global count.

The fixture has both numbers (`overall.toolCalls` 7, `overall.turns` 2), so the honest row is `formatAverage(7, 2)` = `3.5`. Either replace this row or drop `compactions` from `TraceFixture.overall` and `copy.traces.compactions` with it — nothing else reads either.

packages/ui/src/fixtures/dataset.ts
```diff
@@ -0,0 +1,907 @@
...
+    session: {
+      id: SESSION_ID,
+      title: s.title,
+      agentId: AGENT_ID,
+      model: { provider: "deepseek", modelId: "deepseek-v4-pro" },
+      workspace: WORKSPACE,
+      createdAtIso: at(0),
+      running: true,
+      totals: { tokens: 43_830, costUsd: 0.0231, elapsedMs: 71_840 },
+      context: { tokens: 27_310, window: 1_000_000 },
```
hiyouga: **[copy]** The composer's context ring renders as an empty circle in all 12 shots, and the basis is the one the app rejects.

`transcript.tsx:454-472`: `share = 27310 / 1000000 = 0.0273`, `c = 2π × 5.5 = 34.56`, so the arc is `Math.max(0.94, 1)` = a **1 px dash on a 34.6 px circumference** — clamped to the minimum, i.e. indistinguishable from zero in every theme × mode × language capture. Nothing about the ring is reviewable.

`packages/web/src/features/chat/context-gauge.tsx:9-16` names this exact trap: the app fills against the **effective compaction threshold, not the model window**, "because against a window that dwarfs the threshold every answer looks like 'plenty' (64k used at a 128k threshold on a 1M window would draw 6% instead of 50%)". `fixtures/types.ts:197-198` claims the opposite ("tokens in the current context against the model's window").

Fix: make `context.window` the compaction basis the app fills against (and rename the field to say so), with a fill worth looking at — the app turns amber past 80% and red past 95%, so something in the 0.5–0.8 band shows the ring at its resting colour.

packages/ui/src/screens/format.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/** Money in USD with the precision small agent spend needs: `$0.0231`, `$2.31`, `$41.37`. */
+export function usd(n: number): string {
+  return n < 1 ? `$${n.toFixed(4)}` : `$${n.toFixed(2)}`;
+}
+
+/** A byte size: `942 B`, `48.2 KB`. */
+export function bytes(n: number): string {
+  return n < 1024 ? `${n} B` : `${trimZero(n / 1024)} KB`;
```
hiyouga: **[copy]** The same Trace file is two different sizes, and the unit is spaced where the product's is not.

`bytes(48_213)` = `47.1 KB`, which is what the Traces screen prints (`traces.tsx:248`). But the copy for the same file says `48.2 KB` (decimal KB) in three places: `en.ts:291` / `zh.ts:284` ("48.2 KB written to trace-001.jsonl.") and the caption specimen `en.ts:417` / `zh.ts:410`. The notices module and the Traces screen sit in the same gallery, so both numbers are on screen for one 48 213-byte file.

Separately, the app writes `47.1KB` with no space (`packages/web/src/lib/format.ts:154-163`), and this file's own header claims it follows those rules. In a typography gallery the figure/unit gap is exactly the kind of thing being judged.

Same class, one line up in the specimens: `en.ts:413` / `zh.ts:406` claim "Prompt cache hit 81%" for a run whose own Traces screen prints `33.7k · 83%` (`traces.tsx:65`), and whose turns are 79% and 86%.

Fix: `${n}B` / `${trimZero(kb)}KB` to match the product, and set the three prose figures to `47.1 KB` and `83%`.

packages/ui/src/screens/login.tsx
```diff
@@ -0,0 +1,129 @@
...
+  );
+}
+
+function Corner({ f }: { f: Fixtures }) {
+  const a = f.copy.auth;
+  const s = f.copy.settings;
+  return (
+    <div className="absolute right-4 top-4 flex items-center gap-2">
+      <div aria-label={a.language}>
+        <Segmented options={[a.langZh, a.langEn]} value={f.lang === "zh" ? 0 : 1} />
```
hiyouga: **[copy]** The product's login language switch has three options, not two.

`packages/web/src/pages/login.tsx:99-107` puts two segmented controls in the corner: language as `langEn` / `langZh` / `followSystem` ("English" / "中文" / "System"), and theme as Light / Dark / System. This mock gives language two options in the reverse order, so it sits next to a three-wide theme control at two-thirds its width — a layout difference a theme reviewer reads as a deliberate composition.

`copy.settings.system` is already in both dictionaries, and `Segmented` already branches on `options.length === 3` (`parts.tsx:209`), so this is `options={[a.langEn, a.langZh, s.system]}` with the matching `value`.

packages/ui/src/fixtures/shared.ts
```diff
@@ -0,0 +1,575 @@
...
+  },
+];
+
+// ---------------------------------------------------------------------------
+// Notices, forms, menus, secrets, plugins, the palette and usage (K-redesign §4.6)
+// ---------------------------------------------------------------------------
+
+/** The accent presets `theme.css` declares, in the order the swatch picker draws them. */
+export const ACCENT_SWATCHES: readonly { id: string; color: string }[] = [
+  { id: "theme", color: "var(--ui-accent)" },
```
hiyouga: **[copy]** The comment does not describe this list, and the first swatch is not the product's.

`theme.css` declares five presets — `blue`, `green`, `violet`, `rose`, `amber` — and says of the sixth (`:199`) that "`data-accent` (the stored `neutral`) leaves the theme's own accent in place". The app draws that sixth option first, named "Neutral" / "灰白" and painted `#6b7280` (`packages/web/src/state/theme.tsx:212-219`, `strings-en.ts:248`, `strings.ts:264`). This entry replaces it with `id: "theme"` — a value the app never stores — painted `var(--ui-accent)` and labelled "Theme default" / "跟随主题".

The other five labels diverge from the app's zh dictionary too: `蓝色 / 绿色 / 紫色 / 玫红 / 琥珀` against the app's `蓝 / 绿 / 紫 / 红 / 橙` (`strings.ts:263-270`).

Fix either way, but make the two agree: use `neutral` / `#6b7280` and the app's `accentNames`, or keep the theme-accent swatch and change the comment to say this list is the gallery's own substitution rather than what `theme.css` declares.

packages/ui/src/screens/chat.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <Sidebar f={f} activeSessionId={f.session.id} />
+      <div className="flex min-w-0 flex-1 flex-col">
+        <ChatHeader f={f} dock="right" />
+        <div className="flex min-h-0 flex-1">
+          <main className="flex min-w-0 flex-1 flex-col">
+            <ChatTranscript f={f} />
+            <Composer f={f} />
+          </main>
+          <div className="w-[24rem] shrink-0">
+            <DockFrame f={f} tab={f.copy.dock.subagents(1)} glyph="bot" edge="right">
```
hiyouga: **[copy]** A dock tab is rendered text, and this one is not the product's.

`packages/web/src/features/dock/panel-meta.tsx:20-35` labels that panel `S.chat.openAgents` = **"Agents panel"** (zh "智能体面板"), with no count. The traces tab on the other screen is right (`S.nav.traces` = "Trajectories").

While here, the dock's tooltip copy has drifted as well — invisible in a capture, but it is the same dictionary claim: `dock.bottomDock` "Bottom panels" vs the app's "Bottom panel", `dock.rightDock` "Right panels" vs "Right sidebar", `dock.close` "Close" vs `hideDock` "Hide sidebar", `dock.movePanel` vs the app's `moveToBottom` / `moveToRight`. Same in the sidebar: `nav.filterSessions` "Filter and sort" vs `listSettings` "List options", `nav.newFolder` "New folder" vs `newWorkspaceEntity` "New workspace" (that button browses for an existing directory; it creates no folder). `dock.newPanel` "Add panel" and `chat.approve` "Allow" are both correct.

packages/ui/src/screens/transcript.tsx
```diff
@@ -0,0 +1,570 @@
...
+
+/**
+ * The pending-approval block: the one genuine ask in a view, so it keeps its amber fill — but the
+ * rule above it is the neutral line, because a tint and a line of the same hue is the one-hue box
+ * (§1.3 row 19). The alias is mono text; it needs no chip of its own.
+ */
+function ApprovalBlock({ item, f }: { item: ToolCallItem; f: Fixtures }) {
+  let preview = item.argumentsJson;
+  try {
+    const args = JSON.parse(item.argumentsJson) as { cmd?: string };
```
hiyouga: **[over-defensive]** This `try`/`catch` cannot fire. `argumentsJson` is produced by `json()` (`fixtures/dataset.ts:169` — `JSON.stringify(value, null, 2)`), and `ApprovalBlock` only renders for `item.state === "waiting"`, which is `tc7` alone, whose arguments always carry `cmd`. So the `preview = item.argumentsJson` seed, the `cmd?:` optionality and the swallowed catch are all unreachable.

Delete all three:

```tsx
const { cmd } = JSON.parse(item.argumentsJson) as { cmd: string };
const preview = `$ ${cmd.replace(/\s+/g, " ")}`;
```

packages/ui/src/screens/transcript.tsx
```diff
@@ -0,0 +1,570 @@
...
+  f: Fixtures;
+  expansion?: Expansion;
+  from?: string;
+  /** The dock's narrower column: prose one step smaller. */
+  dense?: boolean;
+}) {
+  const start = from
+    ? Math.max(
+        0,
+        turn.items.findIndex((i) => i.id === from),
```
hiyouga: **[over-defensive]** `findIndex` returning `-1` is clamped to `0`, so an unknown `from` silently renders the whole turn instead of its tail.

`ChatTranscript` passes `from="tx3"` (`chat.tsx:33`). Rename `tx3` in `fixtures/dataset.ts:433` and the chat screen quietly grows turn 1's full answer — a fenced tree block and a three-row table — into every capture, and nothing fails: `fixtures-screens.test.ts:112` only asserts `html.length > 1000` and that the session title appears.

Make it loud instead — the neighbouring screens already do this (`chat.tsx:42,44`, `traces.tsx:225`):

```tsx
const start = from === undefined ? 0 : turn.items.findIndex((i) => i.id === from);
// with a non-null assertion or an explicit throw on -1
```

packages/ui/src/screens/transcript.tsx
```diff
@@ -0,0 +1,570 @@
...
+    <div className="ui-glass @container rounded-lg border border-line-emphasis bg-surface px-2.5 pb-2 pt-2">
+      {children}
+    </div>
+  );
+}
+
+export function Composer({ f, compact = false }: { f: Fixtures; compact?: boolean }) {
+  const s = f.session;
+  const c = f.copy.chat;
+  const model = f.models.find((m) => m.modelId === s.model.modelId);
```
hiyouga: **[over-defensive]** Two dead fallbacks that hide a broken fixture rather than failing on it.

Here: `f.session.model.modelId` is `deepseek-v4-pro`, which is `MODEL_ROWS[0]`, so `model` is never `undefined` — yet `:552` falls back to `"?"` for the provider label and `:553` renders nothing for the name. Same shape at `parts.tsx:278`: every `SessionListItem.agentId` is `AGENT_ID` or `REVIEWER_AGENT_ID` and both are in `f.agents`, so `agent?.name ?? item.agentId` is dead.

Both files already assert elsewhere on exactly this pattern (`chat.tsx:40-44`, `traces.tsx:225`). Use `!` in both places so a fixture that loses a row fails the render test instead of drawing a `?` into a screenshot.

packages/ui/src/screens/login.tsx
```diff
@@ -0,0 +1,129 @@
...
+    <div className="relative flex h-screen w-full items-center justify-center overflow-hidden bg-canvas p-4 text-fg">
+      <Circuit />
+      <Corner f={f} />
+      <div className="relative w-full max-w-sm">
+        <img src={logoSrc} alt="" className="mx-auto mb-3 h-16 w-16 rounded-2xl" />
+        <h1 className="ui-display mb-6 text-center text-3xl font-(--ui-weight-strong) tracking-tight text-fg">
+          {f.copy.appName}
+        </h1>
+        <div className="rounded-xl border border-line bg-surface p-6 shadow-sm">
+          <form className="space-y-4" onSubmit={(e) => e.preventDefault()}>
```
hiyouga: **[over-defensive]** Dead handler. This `<form>` holds no `<input>` and no submit control — the two fields are `<span>`s (`:103-115`) and "Sign in" is a `role="button"` `<span>` (`:116-121`) — so nothing can ever submit it. Delete `onSubmit`.

packages/ui/src/screens/settings.tsx
```diff
@@ -0,0 +1,209 @@
...
+                      i === 0 ? "ring-2 ring-line-emphasis ring-offset-2 ring-offset-overlay" : ""
+                    }`}
+                  />
+                ))}
+              </div>
+            </PrefRow>
+            <PrefRow label={c.launcher} info={c.launcherInfo}>
+              <Switch on />
+            </PrefRow>
+            <PrefRow label={c.toolAliases} info={c.toolAliasesInfo} popover={c.toolAliasesInfo}>
```
hiyouga: **[simplify]** `popover` is passed the same string as `info`, and it only means "this row's popover is open". Make it `open?: boolean` and have `PrefRow` render `<FloatingPanel>{info}</FloatingPanel>`; the string then lives in one place.

While here: `:89` puts the whole explanation in a native `title`. In the app the "?" carries only "More info: &lt;label&gt;" as its accessible name and the explanation lives in the click-opened popover (`packages/web/src/components/ui/info-popover.tsx:58,65`) — worth mirroring, since the "?" appears on all six rows in the capture.

packages/ui/src/screens/parts.tsx
```diff
@@ -0,0 +1,461 @@
...
+// ---------------------------------------------------------------------------
+
+/**
+ * The one Spinner already lives in the component set (rule 11 gives it the only `animate-spin`);
+ * the screens re-export it so W1 removes this line rather than rewriting call sites.
+ */
+export { Spinner };
+
+/** A 6 px state dot. Dots stay round in every theme; that is what `rounded-full` is kept for. */
+export function Dot({ className }: { className: string }) {
```
hiyouga: **[simplify]** This stand-in will not swap mechanically, unlike its neighbours.

W1's `Dot` is `{ tone: Tone; size?: "xs" | "sm" | "md"; live?: boolean; label?: string; knockout?: boolean }` (K-redesign §5.1), and `Dot` is one of `ui-live`'s four hosts. This one takes a raw `className`, so the two call sites are a rewrite, not a rename — and one of them (`:288`, `bg-accent`) has no `Tone` at all, while the other (`transcript.tsx:213`, `bg-tone-attention-emphasis`) uses an emphasis fill where W1's `tone` would pick the `-fg` ink. Two lines today; a question to answer in W1 otherwise.

Smaller, same class: `Badge` omits the `size` prop that both §1.3 row 20 ("becomes `Badge size=\"sm\"`") and §5.1 name, so today's call sites will inherit whatever W1 picks as the default.

packages/ui/src/fixtures/shared.ts
hiyouga: `main` is merged into the stack (`196817030`) and the rows follow its catalog: `displayName: "DeepSeek V4 Pro 0813"`, and the retired Flash row is replaced by the live `deepseek-flash` / `DeepSeek V4.1 Flash` (vision, same price), with the notice that advertises it. `fixtures-screens.test.ts` passes against `main`'s catalog, and it now also fails a row the catalog has marked `retired: true` — a rename fails on the regex, a retirement used to pass. 8618fc527

packages/ui/src/screens/transcript.tsx
hiyouga: The line is the app's now: the reply's time leads (`messageTime(atIso, lang)`, fixed to UTC so a capture reads the same anywhere), then input, output, TPS, cost, elapsed, then copy and fork. The wrench chip and the `(NN%)` parenthetical are gone — both belong to the Trace page's turn card. 8618fc527

packages/ui/src/screens/traces.tsx
hiyouga: Replaced with Avg tools / turn — `average(7, 2)` = `3.5`, the two rows above it divided, as `trace-file-view.tsx` prints it — and `compactions` left `TraceFixture.overall` and both dictionaries with it. 8618fc527

packages/ui/src/fixtures/dataset.ts
hiyouga: The field is `context: { tokens, threshold }` and the ring fills against the threshold the app fills against, with the comment naming `contextFillBasis`. The mock sits at 27.3k of a 40k threshold = 68%, below the app's amber at 80%, so the arc is worth looking at in every cell. 8618fc527

packages/ui/src/screens/format.ts
hiyouga: `bytes()` writes `${n}B` / `${trimZero(kb)}KB`, as `lib/format.ts` does, so the Traces screen prints `47.1KB`. The three prose figures now say `47.1 KB`, and the specimen's cache line says 83%. 8618fc527

packages/ui/src/screens/login.tsx
hiyouga: Three options in the app's order — `langEn` / `langZh` / `system` — so the two corner controls are the same width. 8618fc527

packages/ui/src/fixtures/shared.ts
hiyouga: Took the first option: the list opens with the app's stored `neutral` at `#6b7280`, then the five presets `theme.css` declares, and both dictionaries use the app's `accentNames` (`Neutral / 蓝 / 绿 / 紫 / 红 / 橙`). The comment says what the list is. 8618fc527

packages/ui/src/screens/chat.tsx
hiyouga: The tab is `dock.agentsPanel` — "Agents panel" / "智能体面板", no count — and the drifted tooltips are the app's too: `bottomDock` "Bottom panel", `rightDock` "Right sidebar", `hideDock` "Hide sidebar", `moveToBottom` / `moveToRight` (the button flips by edge), `nav.listSettings` "List options" and `nav.newWorkspace` "New workspace". `AppCopy`'s doc now says the two dictionaries are not guarded against drifting, and that this review found some. 8618fc527

packages/ui/src/screens/transcript.tsx
hiyouga: Deleted, exactly as written: `const { cmd } = JSON.parse(item.argumentsJson) as { cmd: string }` and a template for the preview. 8618fc527

packages/ui/src/screens/transcript.tsx
hiyouga: Loud now: `const start = from === undefined ? 0 : turn.items.findIndex(…)` and a throw on `-1`, with the comment naming the neighbours' `find(…)!` idiom. 8618fc527

packages/ui/src/screens/transcript.tsx
hiyouga: Both are `!` now — the composer's model lookup and `SessionRow`'s agent — so a fixture that loses a row fails the render test instead of drawing `?` into a screenshot. 8618fc527

packages/ui/src/screens/login.tsx
hiyouga: Deleted. 8618fc527

packages/ui/src/screens/settings.tsx
hiyouga: `PrefRow` takes `open?: boolean` and renders `<FloatingPanel>{info}</FloatingPanel>`, so the explanation lives in one place. The "?" carries `moreInfoAbout(label)` — "More info: Tool short names" / "说明：工具短名" — as its `title` and `aria-label`, mirroring `info-popover.tsx`; the old bare `moreInfo` key is gone. 8618fc527

packages/ui/src/screens/parts.tsx
hiyouga: `Dot` takes `tone` (§5.1's shape at its `xs` size) and paints the tone's **ink**, never an emphasis fill; `accent` is the one addition, for the unread mark, and it is documented as such. The two call sites are `<Dot tone="attention" />` and `<Dot tone="accent" />`, so W1's swap is a rename. `Badge` gained §5.1's `size` (`sm` 20px / `md` 24px, default `md`); the Trace's dense event rows take `sm`. 8618fc527

PR #764, opened 2026-09-16
feat(ui-gallery): the component gallery — fifteen modules, one composition each

hiyouga:

**Review round (2026-09-18, #765's review):** merged the fixed stack (957a83742, 2bbff4238); the gallery resolves the package through the workspace link (no vitest alias, no tsconfig `paths`), the tokens drawer judges `--ui-fg-on-emphasis` on the neutral emphasis fill (af4f7c45d), and the switch mock-ups' knob keeps a hairline on Primer's near-white dark accent (b481a4984).

**Second round (2026-09-18):** merged the stack (fc69eef05; the switch knobs and the tokens drawer read `--ui-tone-neutral-emphasis-fg` now that `--ui-fg-on-emphasis` is gone), and the modules pass the three web guards that failed: dialog-footer Buttons say `size="sm"`, the overlays InfoPopover sits after its title in the Appearance row it explains, and the stats ring's stroke follows its size (8c50c55b2). The web suite is green; 30 module shots and the page-top view changed and were refreshed in the K3 set.

**Third round (2026-09-19):** merged #761's review fixes (66a045fae); the gallery build never inlines a font either, and the tokens drawer shows `--ui-fg-subtle`'s ratios without judging them, as the contrast suite exempts it (c1db975b0). 48 Frost module shots, 30 of them in the K3 set, and 6 page views changed with Frost's subtle ink and were refreshed.

**Fourth round (2026-09-19):** merged current `main` and #762's and #763's review fixes (`3246cd411`), and fixed this PR's review (`abd7acf35`, `6ae67f0d2`). `.g-app code` and `.g-app ::selection` no longer reach inside a preview and the Parts drawer carries `g-chrome` on its own elements, so the page, the compare frames and the PNGs agree; the preview frame reads the contract tokens with no fallback; every ticket state gets a column; `shots.mjs` validates all four axes and fails on a module that does not render; a variant key the map lacks throws (`viewFor`) and a new test renders every variant of every package-owned module; `MATRIX_KEY` is rejected as an axis value; the Icons board counts files and says how many paths it could not attribute, badging only spelled-out names; the mock week, the Log variant, the usage chart and the accent presets are read from the data and fail loudly; `tok/s`, `/M`, a file's `A`/`M` mark and the breadcrumb label come from the dictionaries; the tokens drawer reports what it measured, and `tokensFromRules` is pure and tested; `packages/ui/test/app-source-scope.test.ts` holds the app's `@source not` list to the package's directories. On vps3: ui 12 files / 179 tests, gallery 8 files / 55 passed / 1 skipped, web 187 files / 2488 passed / 2 skipped, `pnpm typecheck`, `pnpm lint` 0/0, `pnpm format:check`, both builds. The tokens probe still resolves 188 tokens in all six cells with 15 drawers, 0 unset and 0 mismatches. 84 module shots moved (54 of them in the K3 set) and `page-top.png` with them; all were refreshed.

Row K3 of the theme revision: the gallery is rebuilt around **modules**. One long page, fifteen modules, each one realistic composition of the product's own UI built from the shared fixtures — not a wall of 124 numbered sections. Start it with `pnpm dev:gallery` → http://localhost:7372.

## The shape

- **Fifteen modules** in one order: Foundations, Conversation, Composer, Sidebar & navigation, Buttons & actions, Status & feedback, Forms, Overlays, Tables & lists, Stats & charts, Markdown & code, Files & trees, Pages & sections, Company board, Screens. Each has three to five variants (`Streaming · Settled · Approval · Failed`), a header of `Title — one sentence`, and a card on the theme's own canvas with the variant pills bottom-centre and a quotable breadcrumb bottom-left: `Frost › Conversation › Approval · dark · zh`.
- **The module contract** is `packages/ui/src/module.ts`: `{ id, title, description, width, variants, parts, render(variant, ctx) }`, collected with `import.meta.glob`. Thirteen modules live in `packages/ui/src/modules/*.module.tsx`; Foundations and Screens live in the gallery because they are made of gallery machinery (the boards, the `/screens` frames). A module renders static parts today and swaps them for real components wave by wave — the module file is the seam.
- **Compare per width**: a narrow module's three themes stand side by side, a wide module's stack at full width (three chat transcripts at a third of the column each are unreadable), each frame with its theme name and breadcrumb. `compare=<module>` pins one module in compare while the rest of the page stays single, and widens the frame.
- **Three drawers per module**: *parts* (each component demo as a small card with its own axis pills and `all` matrix; a planned component as one quiet line — `Button · W1 · replaces button.tsx, ≈40 icon buttons`), *tokens* (the tokens the rendered composition actually reads, resolved in the active theme and mode, with WCAG ratios and the declared `tokensUsed` of any demo), *code* (the module's source).
- **URL v2**: `?theme=&mode=&tier=&lang=&compare=1|<module>&v.<module>=<key>#<module>`, plus `/embed?module=<id>&variant=<key>` (new) and `/embed?demo=<part-id>&variant=` (kept, so part breadcrumbs quoted in earlier feedback still resolve).
- **Foundations** is a module with eight variants: colour, type, shape & depth, spacing, icons, motion, focus, hooks. The boards replaced the per-token rows; the token list is now the drawer.
- **No data in the modules.** Every word and number a composition prints comes from `packages/ui/src/fixtures/`: #763's sets (notices, forms, menus, vault, plugins, command palette, usage) where they hold the same data, and new ones for the rest — the citation test failing, the Task's to-do list (built from the session's own calls), the slash menu, a docs answer, the plugin library's totals, the group chat, the dock's panel menu, the dialog's search and choice groups, and `AppCopy` sections for the pages the modules imitate. The swatch picker paints the presets' own colours from `forms.swatches`.
- **`/fonts`** groups the declared faces per family and weight with a slice count and load status (MiSans ships 99 slices per weight, Noto Sans SC 101 — one row each, not 200 rows).
- **Deleted** (K-redesign §4.5): the coverage strip, section ordinals and `g-num`, group runs and group headers, `PlannedCard`, the `uncatalogued` trailing group (a demo outside every module is now a test failure), `FoundationSection`, the hatch and dashed chrome rules.

## De-slop pass (K-redesign §3)

Every composition follows the component rules: no `transition-all`, no hover scale, no tinted icon chips, no coloured left-border callouts, status marks without haloes, one spinner, spacing by relationship, numbers that align in tabular figures, colour through tokens only. The parts in `packages/ui/src/modules/parts.tsx` are named after the components they stand in for (`Button`, `Badge`, `Dot`, `FloatingPanel`, …) so W1 swaps an import, and the style hooks sit only in their allowed hosts.

## Deviations (and why)

- **Foundations and Screens are gallery-owned modules.** Both are built out of the gallery's own pages and the `/screens` iframes; moving them into `packages/ui` would invert the dependency. K4: the hook specimens live in `packages/ui-gallery/src/foundations/hooks.tsx`, outside the package's host check.
- **Foundations has eight variant pills, not seven** — §4.5 keeps the Style hooks page, and it is the eighth.
- **The tokens drawer measures.** No W0 demo declares `tokensUsed`, so the drawer reads the CSSOM rules that match the rendered composition and resolves those names; declared ones are merged in when they appear.
- **Two glyphs join the screens' registry** (`history`, `terminal`): fixture data names icons by `GlyphName`, and the slash menu and the panel menu need these two.
- **Fixture shapes #763 defined grew a little**: a form field can carry `required`, `placeholder` and `prefix`; `forms` gains `search` and `groups` (the checkbox and radio groups §4.1 lists); a menu entry can carry a `description`; `menus` gains `panels`; the session gains `runCommand`. Additive only.

## Verification

All on vps3 (this box only edits files):

- Stacked on #761 at `6aaf6918d` and #763 at `b96c39aba` (both merged in).
- `pnpm --filter @prismshadow/penguin-ui typecheck`, `pnpm --filter @prismshadow/penguin-ui-gallery typecheck`, `pnpm lint`, `pnpm format:check`: clean.
- Package tests: 11 files, 167 tests passed, the guards among them — `deslop` 62, `hooks` 9, `token-contract` 17, `contrast` 18, `font-licenses` 7 — and `fixtures-screens` 18 (both locales keep one shape with the new sets).
- Gallery tests: 8 files, 45 tests passed — module collection and coverage (every `MODULE_IDS` entry has a file, every catalog section's module exists, every demo is in some module's parts, zh titles for every module and variant), URL v2 round trip with `compare=<module>` and `v.<module>=`, both breadcrumb forms, demo collection, the measured-token reader, the chrome palette guard.
- Screenshots: 90 module shots (15 modules × 3 themes × 2 modes, en) + 24 zh + the page views (rail and header, a section with both drawers open, narrow compare, wide compare, `/fonts`), all read by eye against §3, plus every variant of every module in Frost light, en and zh. After the fixture move, `shots.mjs diff` against the previous set moved 90 of 122; those were re-read and replaced. Fonts are the real ones: the run fails if a family the theme names has no loaded face. They are in `_coord/theme/K3-shots/`.

## Findings for other rows

- Frost light `--ui-fg-subtle` now clears the drawer's 4.5:1 floor (`#726f6c`: 4.6 on canvas, 5.0 on surface) since #761's `6aaf6918d`.
- Primer has no underline recipe for `ui-underline-nav` in W0, so the tab marker is flat there.


hiyouga (commented):
Reviewed **only this PR's own incremental diff** — `git diff origin/feat/theme-w0-fixtures..c1db975b0` (merge-base is exactly `a84c0a3d2`, so the 95 files / +15,299 below are K3's work and nothing from #765, #761, #762 or #763). Points settled in #765's and #761's reviews are not re-raised. Screenshots read: the `page/` set plus a representative slice of `K3-shots/{github,modern,geek}/{light,dark}/{en,zh}`.

## It does its job

The compositions are the product, not a swatch wall: `conversation` is a real transcript with an opened diff and a pending approval, `company` is the board with its priority and status badges, `stats` is the Trace panel. Compare is genuinely comparable — three real roots through `/embed`, side by side for `narrow` and stacked for `wide`, each frame reporting its height so none scrolls. The tokens drawer measures the live CSSOM rather than a hand-kept list, and it handles the three hard cases correctly: a name built at runtime (`--ui-tone-${tone}-fg`) is already substituted by the time the rule is read, an unset name renders as `unset` with an empty swatch, and nothing in the tree re-declares a `--ui-*` below `:root`, so the probe's root values really are the effective ones. The probe's trick — copy the page's CSS into a blank same-origin frame and flip its root through all six cells — is right, and `paintColor`'s two-sentinel `fillStyle` test is the correct way to tell "not a colour" from "a colour that happens to equal the sentinel".

Also checked clean: no app package imports `@prismshadow/penguin-ui-gallery`, and the package is `private`; `shots.mjs`'s `.ts` imports work under Node 24 type stripping because `catalog.ts` and `module.ts` carry type-only imports; determinism is covered (fixed 758px viewport, `deviceScaleFactor: 1`, `motion=reduced` applied in a layout effect before first paint, plus Playwright's `animations: "disabled"`); `THEME_FONTS.github` being empty is correct, since `themes/github.css:155` is system stacks only in W0; en/zh parity is type-enforced (`GalleryStrings = typeof zh`) and the three `Partial` maps are either covered by `modules.test.ts` or complete; the changelog pair follows the repo contract; and design PR #175's 06-PROTOTYPE 组件画廊 section matches the code (fifteen modules, four header buttons, eight Foundations variants, compare per width, the routes, both breadcrumb forms).

Guards: nothing was exempted. The diff touches no test file; `packages/ui/test/deslop.test.ts` keeps `ALLOWLIST = {}` over the whole of `packages/ui/src`, so the new `modules/` passes all 22 rules with no seat at the table, and `hooks.test.ts` / `token-contract.test.ts` scan `packages/ui-gallery/src` for stray `ui-*` and `--ui-*` names. One thing worth saying out loud in `hooks.test.ts` rather than leaving implicit: its Appendix-B host check runs over `filesOf("ui")` only (`hooks.test.ts:70`), so `ui-gallery/src/foundations/hooks.tsx` — which applies all six hooks outside any host on purpose — is outside it by construction. That is the right call; the test should say so, because a *future* gallery component that misuses a hook slips through the same hole.

## Findings without an inline anchor

**[test] Nothing guards the app stylesheet's `@source` exclusions — the only thing keeping this PR out of the app's CSS bundle.**
`packages/web/src/styles.css:27-32` is `@source "../../ui/src"` plus four `@source not` lines (`**/*.demo.tsx`, `modules`, `screens`, `fixtures`, `testing`). This PR is what fills `modules/` with ~5,600 lines of app-lookalike Tailwind markup. No test reads that list (`git grep '@source' packages/**/test` is empty), so renaming a directory, or adding a sixth gallery-only one, silently ships its utility classes in the shipped app's sheet and nothing fails. The package's own guards already hold exactly this doctrine — `testing/source-roots.ts:5-10`, "a root that yields nothing is a failure, never an empty pass". Fix: a test in `packages/web/test/` that parses the `@source not` globs out of `styles.css` and asserts they cover every directory under `packages/ui/src` the app does not import.

**[test] `tokensReadBy` — the function the tokens drawer's truthfulness rests on — has no test.**
`packages/ui-gallery/test/tokens-read.test.ts` covers `tokenNamesIn`, `matchableSelector` and `sortTokens`. The part that decides what the drawer claims — walking `document.styleSheets`, descending a nested rule with its parent selector, declarations nested in an at-rule *inside* a style rule, and inline `[style]` attributes (`lib/tokens-read.ts:50-92`) — is untested, because the package's vitest runs in `node`. Fix: one `// @vitest-environment jsdom` file that builds a small document with a Tailwind-shaped nested rule, an `@media` block inside a style rule and `style="color:var(--ui-fg)"`, and asserts the names.

**[simplify] The stack's `existsSync(gallery)` conditionals are dead the moment this merges.**
`packages/ui/test/hooks.test.ts:60-66` and `:119-121`, and `token-contract.test.ts:147-151` and `:172-178`, branch on whether `packages/ui-gallery` exists; `hooks.test.ts:120` is an `it.skip` that names "#764". This is the PR that lands the package and it already carries #762 merged in, so the false branch can never be taken again. Drop the guards, make the gallery an unconditional root, delete the skip. In the same breath, `packages/web/test/helpers/roots.ts:34-37` still reads "The gallery … joins when that package lands; its chrome is deliberately unthemed, so each guard decides then whether it applies there" — it has landed and nothing was decided. Record the decision and its reason in that comment: the gallery stays out, because `expectSingleHome` would fail on names it reuses (`icons.tsx`, `strings.ts`).

**[over-defensive] Small deletions, each unreachable on today's types and each silent if it ever fires.**
`status.module.tsx:37` types `STATE_MARK` as `Record<string, …>` and then needs `!` at `:74`, while the same file already types `TICKET_TONE` (`:45`) and `PRIORITY_TONE` (`:56`) by their unions — use `Record<RunState, …>` and drop the `!`. `status.module.tsx:121` `workspace?.items.length ?? 0` prints a confident `0` for a missing group. `status.module.tsx:232-251` re-checks `budgetUsd` that the `find` predicate already guaranteed, with a silent `[]` arm — give `find` a type predicate. `company.module.tsx:132` `at += step || Infinity` is dead: the loop already breaks on `!step` at `:146`. `navigation.module.tsx:309` `glyphs[i] ?? "eye"` over two literal 3-element arrays declared four lines apart. `composer.module.tsx:145-146` `model?.displayName ?? …` / `?? "?"` — the second would render a literal question-mark avatar; throw on a miss like `conversation.module.tsx:19-23`'s `item()`. `content.module.tsx:222` `call.diff!` after a filter that already tested it. `overlays.module.tsx:46-49` swallows a missing turn into an empty stage. `chrome/drawers.tsx:147` `values["--ui-canvas"] || "#ffffff"` falls back onto a contract token that cannot be unset, and `:151-153` composites twice (`paintColor(fg, toHex(bg))` already returns an opaque composite, so `composite(fg, bg)` is identity). `foundations/specimens.ts:10-18` adds `short` as a copy of `specimens.ui` "named for where the old pages used it" — §4.5 deleted those pages in this PR.

**[bug] `navigation.module.tsx:131` hard-codes a count the fixture owns.** `row.key === "plugins" ? 4 : undefined` — `f.plugins.length` is exactly 4 (`fixtures/shared.ts:524`) and is read from the fixture at `pages.module.tsx:67` and `tables.module.tsx:305`.

**[bug] A family of fixture lookups that render an empty or wrong composition instead of failing.**
`forms.module.tsx:33` + `:64` and `:140` + `:167` guard on `find(...)` and drop the whole row when the invariant in `fixtures/dataset.ts:160`'s comment ("exactly one carries `error` and exactly one is `disabled`") breaks. `company.module.tsx:166` borrows the *usage bar chart's* day labels for a 7-column calendar grid with nothing tying `f.usage.days.length` to 7 or its first entry to `weekStartIso`'s weekday. `company.module.tsx:207` silently drops any event outside 09:00–18:00, and `weekly-retro` at 17:00 + 60 min fits that window exactly. `company.module.tsx:247`'s legend lists four outcomes while `CalendarOutcome` and `copy.company.outcomes` have five. `conversation.module.tsx:45,:82` and `overlays.module.tsx:50` type fixture ids (`"th3"`, `"tc4"`, `"tc7"`, `"u1"`, `"tx1"`) straight into the modules — `item()` throws on a renamed id, but the *expansion* sets and `overlays`' `filter` just leave a group collapsed or a stage empty. Fix, uniformly: route these through the throwing `item()` helper `conversation.module.tsx:19-23` already has.

**[copy] Untranslated fragments beyond the two flagged inline.** `files.module.tsx:53` prints `"A"` / `"M"` as the visible change mark while the `title` on the same element correctly uses `copy.added` / `copy.modified`. `parts.tsx:703` `aria-label="Breadcrumbs"` is the one hard-coded user-visible string in `parts.tsx` — every other label there arrives as a prop. `stats.module.tsx:314` appends a `k` magnitude suffix the y-axis three lines above omits. `stats.module.tsx:367-368` writes the two budget thresholds as `left-[80%]` / `left-[95%]` while the caption right below is `copy.thresholds` ("Warn at 80%, stop at 95%" / "80% 时提醒，95% 时停止") — editing the copy moves the words and not the lines. `actions.module.tsx:83` writes `bash` inline.

**[copy] `company.module.tsx:164` has a Chinese comment.** ``// The week's day names with their dates: `Mon 14`, `周一 14`.`` — AGENTS.md keeps comments English-only; CJK belongs in the fixtures and the dictionaries. Rewrite the example.

**[simplify] Two byte-identical duplications small enough to miss and large enough to drift.** `company.module.tsx:29-36` (`PRIORITY`) and `status.module.tsx:56-63` (`PRIORITY_TONE`) are the same eight lines, same type, same three entries — a tone change in one will not reach the other. `overlays.module.tsx:297-305` and `status.module.tsx:214-222` repeat the same toast block. Both belong in `parts.tsx` beside `TONE_INK`/`TONE_GLYPH`.

**[bug] Two more boards print a number that is not the number they are about.**
`foundations/foundations.css:638-645` — `.gf-control` declares no border, so `Measured` reports `2 × --ui-control-py-sm + --ui-text-small-lh` = 27px at the 18px root, while `packages/ui/src/tokens.ts` states the same control is "29px at the 18px root, 31.25px in Frost and Console" — the contract's figure includes `--ui-border-w` on both edges. The group is titled `controlHeights` and `density.tsx:5-6` calls it "the number that must match across themes". Fix: `border: var(--ui-border-w) solid transparent` on `.gf-control`. — `pages/fonts.tsx:160` prints `S.fonts.noFaces` ("no faces declared") on the first paint, before the effect has read `document.fonts`, so "not measured yet" and "none declared" are the same sentence; initialise the state to `null` and show `S.intro.resolving` until the first read.

**[bug] `foundations/shared.tsx:69-88`** — `Glyph` hardcodes `fill="none"`, so `ELLIPSIS_ICON`, whose own doc says it "renders through GlyphIcon's `filled` mode" because hairline dots vanish at row size (`packages/web/src/components/ui/session-row-menu.tsx:21-28`), is drawn on the Icons board as three hollow rings rather than the mark the app draws.

**[over-defensive] Gallery-side deletions.** `foundations/icons.tsx:15` `find(…"GEAR_ICON")?.d ?? WEB_ICONS[0]?.d ?? ""` — rename `GEAR_ICON` and the size-rung row silently swaps to whatever icon sorts first; an empty registry renders `<path d="">`, five blank boxes labelled `rowMark 12 … caretDense 10`, with no message. `icons.tsx:13` `icon.sources[0] ?? ""` and `:50` `icon.names[0]?` are both unreachable — `extractIconPaths` seeds each array with one element (`lib/icon-registry.ts:163`) — so type them `[string, ...string[]]` instead. `lib/icon-registry.ts:179` `m[1] ?? ""` (group 1 always participates; `""` would render a nameless rung) and `:122` `(m.index ?? 0)` (a `matchAll` match always has an index; `0` would parse the object body from offset 0) are `noUncheckedIndexedAccess` artefacts — use `!`, not a value that renders. `lib/icon-registry.ts:176` + `:82` + `sources.ts:36` are three silent empties on one path: move `packages/web/src/lib/icon-scale.ts` and `import.meta.glob` matches nothing (a Vite warning, not an error), so the size-rung row becomes an empty flex container. `foundations/shared.tsx:32-35` exports `Resolving()`, which nothing in `src/` or `test/` imports — dead code that makes the "not resolved yet" contract look honoured.

**[bug] `lib/demos.ts:41`** — the matrix branch of `parseVariantKey` lacks the `axes && Object.keys(axes).length > 0` guard `defaultPick` applies at `:23`. `defineDemo({ matrix: true })` with no axes is type-legal and `collectDemos` does not report it, so `?variant=all` returns `{ kind: "matrix" }` and `DemoView` (`preview.tsx:31-44`) renders one cell with an empty label. Mirror the guard, and report a declared matrix with no axes as a problem rather than ignoring the flag.

**[test] `test/icon-registry.test.ts:46-68` cannot fail on a dropped icon.** It asserts `icons.length > 30` and five names. Five declaration forms the regexes silently drop are uncovered: a quoted object key (`ENTRY`'s key class is `[A-Za-z_$][\w$]*`, its separator class excludes `"`), a concatenated value (`CONST_DECL` requires `[;\n]` right after the value), a lowercase or camelCase constant name (`CONST_DECL` and `REF` require `[A-Z][A-Z0-9_]*`), a template literal containing a backslash, and the JSX `d=` form that loses six real icons today. Fix: assert an exact expected count — overall and per file — so a silently dropped icon turns it red.

Also checked clean on the gallery side: `typography.tsx`, `shape.tsx`, `motion.tsx`, `focus.tsx` and `hooks.tsx` read only live tokens and print nothing hand-copied; every `--ui-*` name the gallery spells, including the template-interpolated ones (`--ui-radius-${step}`, `--ui-h${n}-*`, `--ui-control-p${axis}-${size}`, `--ui-tone-${name}-${part}`, `--ui-chart-${n}`), is in the contract, and no `--g-*` is read without a definition; `/fonts` groups MiSans's ~200 slices into one row per family and weight, measures load status from `face.status` rather than assuming it, and never touches `document.styleSheets`, so no cross-origin sheet can throw there; `Measured` prints `…` until it has measured; and `demos.ts`'s `danger.sm` round trip, the stale-key path and `allSelections` ordering are correctly pinned by `test/demos.test.ts`.

None of this is structural. The module contract, the URL scheme, the compare layout, the token probe and the guards' coverage are right, and the shots show the gallery doing the job it was rebuilt for. The list below is what stands between it and being trustworthy as the artifact the three themes get judged on.

Verdict: fix before merge: `.g-app code` and `.g-app ::selection` reaching inside every preview, so the Content module's inline code renders one way on the page and another in every compare frame and every PNG (`chrome.css:62,78`); the `rejected` ticket that renders nowhere (`company.module.tsx:76`); `shots.mjs` validating only `--modules`, so a mistyped `--themes`/`--modes`/`--langs`/`--tier` writes a convincing set under the wrong name and silently disarms the font gate (`shots.mjs:124`), and a module that fails to collect leaving a short set that exits 0 (`shots.mjs:216`); the fifteen `?? <first variant>` fallbacks that turn a drifted variant key into a silently duplicated screenshot (`conversation.module.tsx:124` and thirteen siblings, `foundations.module.tsx:46`); the `var(--ui-*, var(--g-*))` fallbacks that let a dropped token look fine in the very card the themes are judged in (`chrome.css:433,438-442`); `MATRIX_KEY = "all"` colliding with a legal axis value, cheap to fix before W1 names one (`demos.ts:113`); the Icons board's undercount and its six false `×N` badges (`foundations/icons.tsx:20,52`); `tok/s` and `/M` printed untranslated in the zh gallery (`stats.module.tsx:147`, `composer.module.tsx:198`); the four uncoordinated copies of the mock week (`company.module.tsx:114,116,178,233`); the Log variant's unchecked `{ cmd }` cast and hard-coded exit 0 (`content.module.tsx:257,265`); the Compare switch that misreports a pinned module and cannot clear it (`rail.tsx:100`); and the two test gaps — nothing guards `packages/web/src/styles.css`'s `@source not` list, which is the only thing keeping this PR's markup out of the shipped app's CSS, and `tokensReadBy` is untested.

packages/ui/src/modules/company.module.tsx
```diff
@@ -0,0 +1,487 @@
...
+          {ticket.costUsd > 0 ? usd(ticket.costUsd) : "—"}
+        </span>
+      </div>
+    </li>
+  );
+}
+
+function Board({ f }: { f: Fixtures }) {
+  const status = f.copy.company.ticketStatus;
+  const columns = ["proposed", "in_progress", "review", "done"] as const;
```
hiyouga: **[bug]** The dataset's one `rejected` ticket renders nowhere.

`TicketStatus` has five members (`fixtures/types.ts:364`), the fixture ships a `rejected` ticket (`fixtures/dataset.ts:307`, index 6 of 8), and both locales translate `copy.company.ticketStatus.rejected` (`en.ts:287`, `zh.ts:282`). But this line hard-codes four columns, so the Board drops it — and `status.module.tsx:151`'s `.slice(0, 6)` cuts it too, because it sits at index 6. `status.module.tsx:53` even defines a tone for it. So one of the eight tickets, one whole ticket state, and a translated label in both dictionaries are invisible in every theme, and nothing fails.

Fix: derive the columns from the ticket-status record (`Object.keys(f.copy.company.ticketStatus)`) so a state cannot go missing, or drop `rejected` from the fixture if the board is deliberately four-wide.

packages/ui/src/modules/conversation.module.tsx
```diff
@@ -0,0 +1,127 @@
...
+    "layout-disclosure-row",
+    "icons-status-icon",
+    "icons-avatars",
+    "content-prose",
+    "content-code-block",
+    "content-diff-viewer",
+    "data-stat-chip",
+  ],
+  render: (variant, { lang }) => {
+    const View = VARIANTS[variant as keyof typeof VARIANTS] ?? Streaming;
```
hiyouga: **[over-defensive]** Every module's `render` ends in a fallback that can only fire on a bug, and hides it.

`module.ts:76` says "Unknown variant keys never reach it" and `lib/modules.ts:91` (`pickVariant`) enforces it, so the `?? Streaming` arm is unreachable for any key the gallery, `/embed` or `shots.mjs` can produce. What it *is* reachable for is drift between the `variants` array and the `VARIANTS` map — and then the pill for that key silently renders variant #1. Nothing catches it: `collectModules` only checks key shape and uniqueness, `modules.test.ts` never renders a module, and `shots.mjs --variants all` writes two byte-identical PNGs under different names (`shots.mjs:229-234` names the file from the *requested* key), which `diff` compares across trees and not within one.

The same line is in all fifteen modules:
`actions:228 ?? Toolbar` · `company:484 ?? Board` · `composer:294 ?? "idle"` · `content:297 ?? Prose` · `conversation:124 ?? Streaming` · `files:198 ?? Tree` · `forms:245 ?? VARIANTS.settings` · `navigation:455 ?? Sidebar` · `overlays:415 ?? Menus` · `pages:219 ?? Settings` · `stats:402 ?? Overview` · `status:336 ?? Live` · `tables:341 ?? Band` · `ui-gallery/src/modules/foundations.module.tsx:46 ?? ColourBoard` · `screens.module.tsx:30` (the `!screen` arm).

Fix: delete the `?? X` and make the map the single source of the `variants` array — one literal per variant carrying `title` and the component — so the cast is safe by construction and the two lists cannot drift. (`composer.module.tsx:261` needs `STATES` retyped `as const` or `Record<VariantKey, ComposerState>` for that to compile.) If the two literals stay split, the cheap alternative is a test: for each package-owned module, `renderToStaticMarkup(module.render(key, ctx))` must differ between variants — `renderStatic`/`markupStructure` already exist in `packages/ui/src/testing/render.ts`.

packages/ui-gallery/scripts/shots.mjs
```diff
@@ -0,0 +1,374 @@
...
+} else {
+  process.exitCode = await shoot();
+}
+
+async function shoot() {
+  const base = (args.base ?? "http://localhost:7372").replace(/\/$/, "");
+  const out = path.resolve(args.out ?? path.join(HERE, "..", "shots"));
+  const modules =
+    args.modules === undefined || args.modules === "all" ? [...MODULE_IDS] : list(args.modules);
+  const themes = list(args.themes, ["github", "modern", "geek"]);
```
hiyouga: **[bug]** Only `--modules` is validated; the other four axes are taken verbatim and silently mislabel the output tree.

`--modules` gets the check at :131-135, but `themes`, `modes`, `langs` (:124-126) and `tier` (:127) do not. `--themes primer` writes `<out>/primer/light/en/*.png` while `/embed?theme=primer` falls back to `github` (`url-state.ts:82`), so the run produces a complete, convincing set of **Primer** shots filed under another theme's name — and exits 0. `--modes drak` and `--tier xl` do the same. `--langs fr` is worse: `SPECIMEN["fr"]` and `THEME_FONTS[theme]["fr"]` are both `undefined`, so `checkFonts` runs with `wanted = []` and the run's font gate — the thing that makes these shots evidence — is silently switched off.

A `diff` of two such trees then reports "0 moved", because both are wrong the same way.

Fix: validate all four against the same lists the defaults come from and `return 2` naming the known values, exactly as :131-135 does.

packages/ui-gallery/scripts/shots.mjs
```diff
@@ -0,0 +1,374 @@
...
+            }
+            const root = page.locator("#embed-root");
+            const info = await root.evaluate((el) => ({
+              renderable: el.dataset.renderable === "true",
+              variant: el.dataset.variant ?? "",
+              variants: JSON.parse(el.dataset.variants ?? "[]"),
+              axes: JSON.parse(el.dataset.axes ?? "{}"),
+              matrix: el.dataset.matrix === "true",
+            }));
+            if (!info.renderable) return info;
```
hiyouga, line 237: **[bug]** A module that fails to collect produces a short shot set that looks complete.

`capture` returns here when `#embed-root` reports `data-renderable="false"`, and the module loop at :223-226 ignores the result. If a `*.module.tsx` is missing, exports no `module`, or lands in the wrong file name, `collectModules` reports it (`lib/modules.ts:47-81`) on the gallery page — but `/embed?module=<id>` just renders "There is no module named …", so the run writes no PNG for it, prints nothing about it, and exits 0. The reviewer reads fourteen modules believing they have fifteen.

Only the `--parts` loop records this (`skipped`, :243-245), and even there it is an informational line, not a failure.

Fix: collect non-renderable *module* ids the way `fontFailures` is collected and `return 1` naming them.

packages/ui-gallery/src/chrome.css
```diff
@@ -0,0 +1,931 @@
...
+  margin-top: 12px;
+  border: 1px solid var(--g-line);
+  border-radius: 12px;
+  overflow: hidden;
+  background: var(--ui-canvas, var(--g-surface));
+}
+.g-preview {
+  min-height: 24rem;
+  padding: 24px;
+  background: var(--ui-canvas, var(--g-surface));
```
hiyouga: **[over-defensive]** The preview frame falls back off the token contract, in the one surface whose job is to expose a missing token.

`.g-card` (:433) and `.g-preview` (:438-442) read `var(--ui-canvas, var(--g-surface))`, `var(--ui-fg, var(--g-ink))`, `var(--ui-font-sans, var(--g-font))`, `var(--ui-text-body-size, 1rem)`, `var(--ui-text-body-lh, 1.5)`. `token-contract.test.ts` pins every contract name in all six theme × mode cells, and `boot.ts:53` notes the default theme's selectors match a bare `<html>`, so the fallback arm is reachable only before the stylesheet parses — when nothing is painted. Meanwhile it contradicts the gallery's own rule (`foundations/shared.tsx:2-5`: "an undefined token must look broken"): drop `--ui-canvas` from a theme and the card renders in the gallery chrome's `#ffffff` / `#19191c` and looks fine.

These five names are also what the tokens drawer attributes to the composition (`tokensReadBy` matches `.g-preview` itself), so the drawer inherits the same fiction.

Fix: `background: var(--ui-canvas)` etc., no second argument.

packages/ui/src/modules/stats.module.tsx
```diff
@@ -0,0 +1,405 @@
...
+  return (
+    <div className="grid grid-cols-[minmax(0,1fr)] gap-6">
+      <div className="flex divide-x divide-line">
+        <StatTile label={t.cost} value={usd(o.costUsd)} detail={t.overTurns(o.turns)} />
+        <StatTile
+          label={t.tokens}
+          value={tokens(o.inputTokens + o.outputTokens)}
+          detail={`${tokens(o.outputTokens)} ${t.outputTokens}`}
+        />
+        <StatTile label={t.elapsed} value={duration(o.elapsedMs)} detail={`${o.outputTps} tok/s`} />
```
hiyouga: **[copy]** `tok/s` is an English unit typed into the module; the zh gallery prints it untranslated.

Every other label on this tile row comes from `f.copy.traces`. This one does not, so the Chinese view reads `71 tok/s`. Separately, `copy.traces.outputTps` exists in both dictionaries (`en.ts:124` "Output TPS", `zh.ts:121` "输出 TPS") and is read by nothing — dead copy in two files.

Fix: add the unit to `copy.traces` (the label itself reads oddly as a suffix in either language) and either use `outputTps` or delete it.

packages/ui/src/modules/composer.module.tsx
```diff
@@ -0,0 +1,296 @@
...
+    <span
+      className={`flex items-center gap-2 rounded-[var(--radius-inner)] px-2 py-1.5 ${selected ? "bg-surface-muted" : ""}`}
+    >
+      <AgentTile id={model.provider} name={model.providerLabel} size={18} />
+      <span className="min-w-0 flex-1">
+        <span className="block truncate text-sm text-fg">{model.displayName}</span>
+        {model.note && <span className="block truncate text-xs text-fg-muted">{model.note}</span>}
+      </span>
+      <span className="shrink-0 text-right font-mono text-xs tabular-nums text-fg-muted">
+        {tokens(model.contextWindow)} · {usd(model.pricing.output)}/M
```
hiyouga: **[copy]** `/M` is typed inline while the fixture already owns that unit.

`copy.models.perMTok` exists in both dictionaries (`en.ts:212` `` `${bucket} / MTok` ``, `zh.ts:207` `` `${bucket} / 百万 tokens` ``) and `navigation.module.tsx:282-283` uses it correctly. Here the model picker prints `/M` in both languages.

Fix: `m.perMTok(...)`, as navigation does.

packages/ui/src/modules/company.module.tsx
```diff
@@ -0,0 +1,487 @@
...
+
+/** The working day the week view draws: 09:00 to 18:00, one row an hour. */
+const FIRST_HOUR = 9;
+const END_HOUR = 18;
+const HOURS = Array.from({ length: END_HOUR - FIRST_HOUR }, (_, i) => FIRST_HOUR + i);
+const ROW_REM = 2.75;
+/** The org's zone is Asia/Shanghai, UTC+8. */
+const OFFSET_H = 8;
+/** "Now" in the dataset: Wednesday 16, 15:10 local — after the morning runs, before the eval run. */
+const NOW = Date.parse("2026-09-16T07:10:00.000Z");
```
hiyouga: **[bug]** Four copies of one fact about the mock week, three of which are invisible constants.

`OFFSET_H = 8` (:114) duplicates `f.company.org.timezone` (`"Asia/Shanghai"`, `fixtures/dataset.ts:933`); `NOW` (:116) is a fourth hard-coded instant that must track `f.company.calendar.weekStartIso` (`"2026-09-14"`); and the "today" column is written as a bare `2` twice more — `i === 2` at :178 (bolds the header) and `dayIndex === 2` at :233 (draws the now-line).

They agree today. Move the fixture week by a day and every event silently flips past/upcoming while the now-line stays parked on Wednesday — a wrong calendar that renders cleanly in all six theme × mode cells.

Fix: put the "now" instant and the zone offset in the calendar fixture beside `weekStartIso`, and compute the today column as `floor((now − weekStart) / 86_400_000)` instead of writing `2`.

packages/ui-gallery/src/chrome/rail.tsx
```diff
@@ -0,0 +1,134 @@
...
+          </span>
+        </a>
+        <ModeSwitch />
+      </div>
+
+      <ViewControls />
+      <div className="g-switches">
+        <SwitchRow
+          label={S.rail.compare}
+          checked={state.compare === true}
```
hiyouga: **[bug]** The Compare switch misreports a pinned module and cannot clear it.

`checked={state.compare === true}` reads `false` when `compare=<module>` — which is exactly what the section header's link button writes (`chrome/section.tsx:86`), so it is the state a quoted link puts a reader in. The switch then shows "off" while one module is in compare; clicking it turns compare on for **all fifteen** (45 iframes), and a second click is needed to get back to none. There is no other control for compare, so the URL is the only way out.

Fix: `checked={state.compare !== false}` and `onChange={(on) => update({ compare: on })}`.

packages/ui-gallery/src/chrome/drawers.tsx
```diff
@@ -0,0 +1,245 @@
...
+/** Measures the tokens a rendered composition reads, retrying while its root is not mounted yet. */
+function useMeasuredTokens(root: () => Element | null, key: string): string[] | null {
+  const [names, setNames] = useState<string[] | null>(null);
+  useEffect(() => {
+    let tries = 0;
+    let timer = 0;
+    const measure = () => {
+      const el = root();
+      if (el) setNames(tokensReadBy(el));
+      else if (tries++ < 20) timer = window.setTimeout(measure, 250);
```
hiyouga, line 140: **[bug]** After the retries run out the drawer says "Resolving tokens…" forever.

When `root()` never returns an element, `measure` stops after 20 tries (~5 s) without ever calling `setNames`, so `measured` stays `null` and `TokensDrawer` renders `S.intro.resolving` (:190-191) until the drawer is toggled or a preference changes — indistinguishable from "still working".

It is reachable: in compare mode the root is `frames.current.get(state.theme)?.contentDocument?.getElementById("embed-root")` (`chrome/section.tsx:97-100`) and those frames carry `loading="lazy"` (`chrome/compare.tsx:87`), so a frame the browser has not begun loading yields `null` for the whole window.

Fix: on exhaustion set a state that says so — `[]` and let `S.section.noTokens` speak, or a distinct "the frame never mounted" line.

packages/ui-gallery/src/lib/tokens-read.ts
```diff
@@ -0,0 +1,93 @@
...
+
+/** The contract tokens the subtree at `root` reads, in contract order. */
+export function tokensReadBy(root: Element): string[] {
+  const doc = root.ownerDocument;
+  const found = new Set<string>();
+  const matches = (selector: string) => {
+    try {
+      return root.matches(selector) || root.querySelector(selector) !== null;
+    } catch {
+      return false;
```
hiyouga, line 79: **[over-defensive]** This `catch` makes an under-reporting drawer look like an honest one.

A selector `querySelector` cannot parse is swallowed and read as "does not match", so that rule's `--ui-*` names quietly drop out of the tokens drawer. The catch guards a reachable case (the sheets here are Tailwind's output plus three theme files, and `matchableSelector` rewrites nesting by string surgery), but silently: a drawer that lost half its rules is indistinguishable from a composition that reads few tokens — and the drawer is this PR's evidence that a module reads what it claims.

Fix: count the rejected selectors and surface the count beside the drawer title (or push it into `MODULES.problems`), so a measurement that lost rules says so.

packages/ui-gallery/src/modules/screens.module.tsx
```diff
@@ -0,0 +1,68 @@
...
+  );
+}
+
+export const module = defineModule({
+  id: "screens",
+  title: "Screens",
+  description:
+    "The four full-viewport composites the modules add up to, each opening at full size on its own page.",
+  width: "wide",
+  variants: [
```
hiyouga: **[simplify]** The Screens module keeps a third copy of the screen registry.

`SCREENS` already names the four (`packages/ui/src/screens/index.ts:19,31`), this list repeats them, and `pages/screen.tsx:38` derives the module's default variant a third way (`Object.keys(SCREENS)[0]`). They agree today and nothing pins that: a fifth screen is simply absent from the gallery, and a renamed one renders `S.screens.notFound` at :30 instead of failing a test.

Fix: build `variants` from `SCREENS` (`SCREEN_LIST.map((s) => ({ key: s.id, title: s.title }))`). That makes the `!screen` arm at :29-30 unreachable so it can go too, and lets `screen.tsx` ask the module for its default key rather than re-deriving it.

packages/ui/src/modules/content.module.tsx
```diff
@@ -0,0 +1,300 @@
...
+      >
+        {foot}
+      </div>
+    </section>
+  );
+}
+
+function Log({ f }: { f: Fixtures }) {
+  const c = f.copy.chat;
+  const clone = f.session.turns[0]!.items.find((i): i is ToolCallItem => i.kind === "tool_call")!;
```
hiyouga: **[bug]** The Log variant takes *any* first tool call and casts its arguments unchecked.

`find((i): i is ToolCallItem => i.kind === "tool_call")` does not look at `clone.name`, though `ToolCallItem.name` is a typed `FixtureToolName` (`fixtures/types.ts:138`). Put a `read_file` call ahead of `tc1` in turn 1 — its arguments are `{ file_path, offset, limit }` (`fixtures/dataset.ts:490`) — and the log head renders `$ undefined` with no error. `clone.output ?? ""` at :260 then swallows a call that has no output.

:265-266 compound it: the foot hard-codes a green `circleCheck` and `exitStatus(0, …)` while ignoring `clone.state`, so a failed call would still be drawn as a success in the module whose job is to show what a log looks like.

Fix: `find((i) => i.kind === "tool_call" && i.name === "exec_command")`, throw on a miss the way `conversation.module.tsx:19-23`'s `item()` does, and drive the glyph and exit code from `clone.state`.

packages/ui/src/modules/stats.module.tsx
```diff
@@ -0,0 +1,405 @@
...
+        </ul>
+      </section>
+    </div>
+  );
+}
+
+function Usage({ f }: { f: Fixtures }) {
+  const copy = f.copy.usage;
+  const u = f.usage;
+  const totals = u.days.map((_, i) => u.cacheRead[i]! + u.cacheWrite[i]! + u.output[i]!);
```
hiyouga: **[bug]** A usage series one element short renders a blank chart with no error.

`UsageFixture` (`fixtures/types.ts:570-575`) declares `days`, `cacheRead`, `cacheWrite` and `output` as four independent `readonly` arrays with nothing linking their lengths. All are 7 today (`fixtures/shared.ts:571-575`). Drop one entry from any series and every `!` here yields `undefined`: `totals` becomes `NaN`, `max` becomes `NaN`, and each bar's `height: NaN%` is an invalid declaration the browser ignores — an empty chart that looks like a rendering choice.

Fix: make the fixture one array of `{ day, cacheRead, cacheWrite, output }` rows so the shape cannot skew, or assert the four lengths once before the map.

packages/ui-gallery/test/modules.test.ts
```diff
@@ -0,0 +1,158 @@
...
+    for (const { module } of MODULES.list) {
+      for (const id of module.parts)
+        expect(catalogSection(id), `${module.id}: ${id}`).toBeDefined();
+    }
+    for (const section of CATALOG) {
+      expect(MODULES.byId.get(section.module)?.module.parts, section.id).toContain(section.id);
+    }
+  });
+
+  it("reach every demo through some module's parts", () => {
```
hiyouga, line 50: **[test]** These two assertions pass over nothing, and will keep passing if the demo glob breaks.

No `*.demo.tsx` exists yet, so `DEMOS.byId` is empty: `expect(DEMOS.problems).toEqual([])` compares `[]` with `[]` and the loop body never runs. The same holds for `catalogSection` coverage of `module.parts` in the test above — that one does check something — but the demo half is vacuous today and, once W1 lands demos, a glob that silently stops matching (`registry.ts:27`, the path is relative to the gallery root) would restore it to vacuous without failing.

The package's own suites already solved this: `source-roots.ts:5-10` makes "a root that yielded nothing" a failure, and `deslop.test.ts:15-17` turns a rule with nothing to read into a **named skip, never a pass**. This suite should follow the same rule.

Fix: assert `DEMOS.byId.size` against a count (0 today, with a named `it.skip` explaining that W1 brings the demos), so the day the glob breaks is the day this fails.

packages/ui/src/modules/status.module.tsx
```diff
@@ -0,0 +1,339 @@
...
+  return (
+    <span className={TONE_INK[tone]}>
+      <GlyphIcon name={icon} size={14} />
+    </span>
+  );
+}
+
+function Settled({ f }: { f: Fixtures }) {
+  const company = f.copy.company;
+  const reasons = ["completed", "tool_use", "max_tokens", "error"] as const;
```
hiyouga: **[bug]** Four stop-reason strings invented in the module and printed as badge text at :181.

Only the *heading* comes from the fixture (`f.copy.traces.stopReasons`); the values do not. The Trace fixture carries real ones on `TraceEvent.stopReason` (`fixtures/types.ts:260`) and `stats.module.tsx:280` renders them from there — so the Status module shows a set of reasons the mock dataset does not contain, against the PR's own "no data in the modules" rule.

Fix: collect the distinct `stopReason` values out of `f.trace.turns[].events`, or add the list to the fixture beside them.

packages/ui/src/modules/actions.module.tsx
```diff
@@ -0,0 +1,231 @@
...
+          </li>
+        ))}
+      </ul>
+    </div>
+  );
+}
+
+function Footer({ f }: { f: Fixtures }) {
+  const a = f.copy.agents;
+  const agent = f.agents[1] ?? f.agents[0]!;
```
hiyouga: **[over-defensive]** `?? f.agents[0]!` is dead, and the `!` on it is a second untruth.

`f.agents` is built from `Record<"default" | "reviewer", …>` (`fixtures/dataset.ts:88`), so it always has exactly two entries and `f.agents[1]` is always present.

Fix: `const agent = f.agents[1]!;`.

packages/ui/src/modules/files.module.tsx
```diff
@@ -0,0 +1,201 @@
...
+ */
+import type { ReactNode } from "react";
+import { fixturesFor } from "../fixtures";
+import type { FileNode, Fixtures } from "../fixtures";
+import { defineModule } from "../module";
+import { bytes } from "../screens/format";
+import { Breadcrumbs, EmptyState, GlyphIcon, IconButton, SearchInput } from "./parts";
+
+/** Folders shown open: the ones leading to this session's changes. */
+const OPEN = new Set(["claude-code-expert", "claude-code-expert/src", "claude-code-expert/test"]);
```
hiyouga: **[bug]** Four workspace paths retyped, so a fixture rename leaves the tree collapsed.

The fixture owns these: `fixtures/shared.ts:19` (`WORKSPACE`), `:130` (`RAG_PATH`), `:131` (`TEST_PATH`), and `f.fileTree.path` / `f.filePreview.path` are already in hand here. `:166` retypes one of them again as `<DropOverlay folder="claude-code-expert/src" …>`. Rename the project in the fixture and the Tree variant renders fully collapsed while the Drop variant names a folder that no longer exists — in both cases a plausible-looking screenshot.

Fix: derive `OPEN` from the ancestor segments of `f.filePreview.path` plus `f.fileTree.path`, and pass the preview file's parent folder to `DropOverlay`.

packages/ui/src/modules/content.module.tsx
```diff
@@ -0,0 +1,300 @@
...
+      </blockquote>
+      <h3 className={`pt-2 text-(length:--ui-md-h3-size) leading-snug ${h}`}>{d.rankingTitle}</h3>
+      <p>{d.ranking}</p>
+      <p className="overflow-x-auto text-center font-mono text-sm text-fg">{d.formula}</p>
+    </article>
+  );
+}
+
+/** A code block: the language and path in its head, a copy button, numbered lines. */
+function CodeBlock({
```
hiyouga: **[simplify]** Three code-block renderers, two of them near character-identical.

This one and `actions.module.tsx:76-95` repeat the same head bar (`ui-frame … rounded-md border border-line` → `data-slot="head" … border-b border-line bg-surface-muted px-3 py-1` → mono language span → `min-w-0 flex-1` spacer → `IconButton icon="copy" size="sm"`) and the same body (`data-slot="body" … bg-[var(--ui-code-bg)] … font-mono text-xs leading-relaxed text-fg`); this one and `files.module.tsx:105-114` repeat the same numbered gutter verbatim (`w-10 shrink-0 select-none pr-3 text-right text-[var(--ui-code-gutter)]` + `{i + 1}`). About 35 lines of mechanical repetition across three files — and the drift already happened: the actions copy is where `"bash"` got baked in as a literal (`actions.module.tsx:83`).

`parts.tsx` is where these stand-ins live, and W5's `CodeBlock` will replace exactly one of them.

Fix: one `CodeBlock({ lang, path?, code, copy, numbered? })` in `parts.tsx`; all three call it.

packages/ui-gallery/src/chrome.css
```diff
@@ -0,0 +1,931 @@
...
+  color: inherit;
+}
+.g-chrome button:focus-visible,
+.g-chrome a:focus-visible,
+.g-chrome input:focus-visible,
+.g-chrome summary:focus-visible {
+  outline: 2px solid var(--g-accent);
+  outline-offset: 1px;
+}
+.g-chrome code,
```
hiyouga: **[bug]** `.g-app code` reaches inside every preview, so the page and every screenshot disagree about inline code.

This file's own header (lines 7-9) states the contract: unlayered on purpose, and "match nothing inside a preview except the pills and labels that carry a `.g-*` class themselves". `.g-app` is the outermost wrapper of the whole gallery page (`pages/gallery.tsx:51`), so `.g-app code` matches every `<code>` in every composition — and being unlayered it beats Tailwind's `@layer utilities` outright, whatever the specificity.

`content.module.tsx:28` renders inline Markdown code as `<code className="… font-mono text-[0.85em] …">`. On `/` that renders in the chrome's `--g-mono` at a fixed **12px**, identically in all three themes. In `/embed` there is no `.g-app` (`pages/embed.tsx:61-71`), so the compare frames and **every PNG `shots.mjs` writes** render the same markup in the theme's own mono at `0.85em`. The single view, the compare view and the screenshots show three different things for the one element the "Markdown & code" module exists to compare.

It also hits the Foundations boards: `color.tsx:47`, `color.tsx:109` and `typography.tsx:88` use `<code className="gf-caption gf-mono">`, and `.g-app code` (0,1,1) out-specifies `.gf-mono` (0,1,0) — both unlayered — so those labels sit at a fixed 12px and ignore the 16/18/20 root tier under test, while the `<span className="gf-caption gf-mono">` labels in the same rows follow it.

`.g-app ::selection` (:62-63) is the same mistake: it overrides theme.css's layered `::selection { background: var(--ui-selection-bg) }` inside every preview, including on the Focus board, whose subject is that token.

Fix: drop `.g-app` from both selectors. Every chrome subtree already carries `.g-chrome` (`rail.tsx:84`, `section.tsx:104,143,150`, `gallery.tsx:55`, `fonts.tsx:99`), so `.g-chrome` alone covers the chrome — once the drawer nesting is fixed (see the comment on `section.tsx:150`).

packages/ui-gallery/src/lib/demos.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (!demo || typeof demo.render !== "function" || typeof demo.id !== "string") {
+      problems.push(`${path}: no \`demo\` export (export const demo = defineDemo({ … }))`);
+      continue;
+    }
+    const earlier = byId.get(demo.id);
+    if (earlier !== undefined) {
+      problems.push(`${path}: demo id "${demo.id}" is already used by ${earlier.path}`);
+      continue;
+    }
+    const badValues = Object.entries(demo.axes ?? {}).flatMap(([axis, values]) =>
```
hiyouga, line 116: **[bug]** `all` passes this check, and `MATRIX_KEY` is `"all"` — so that axis value is unaddressable.

`AXIS_VALUE` accepts `"all"`, so `defineDemo({ axes: { scope: ["all", "mine"] }, matrix: true })` collects without a problem. Then `formatVariantKey` (:55-58) formats the single selection `{ scope: "all" }` as `"all"`, and `parseVariantKey` (:41) reads that straight back as `{ kind: "matrix" }`. The round trip is broken: `?v.<id>=all` and `/embed?demo=<id>&variant=all` cannot address that pick, and via `chrome/drawers.tsx:58-63` clicking the pill computes `fallback === "all"`, stores `null`, and re-opens the matrix.

`test/demos.test.ts` uses `variant`/`size` axes with no colliding value, so nothing catches it. No demo exists yet, which is exactly why this is cheap to fix now and expensive once W1 names an axis value `all`.

Fix: reject `MATRIX_KEY` here alongside the shape check — `values.filter((v) => v === MATRIX_KEY)` into `badValues` — and pin it in `demos.test.ts`.

packages/ui-gallery/src/foundations/icons.tsx
```diff
@@ -0,0 +1,65 @@
...
+export function IconsBoard() {
+  const { S } = useGallery();
+  const sources = [...new Set(WEB_ICONS.map((icon) => icon.sources[0] ?? ""))];
+  const sample =
+    WEB_ICONS.find((icon) => icon.names.includes("GEAR_ICON"))?.d ?? WEB_ICONS[0]?.d ?? "";
+  return (
+    <div className="gf-board">
+      <BoardGroup
+        title={S.foundations.iconSizes}
+        aside={S.foundations.iconRegistry(WEB_ICONS.length, sources.length)}
```
hiyouga: **[bug]** The Icons board's count and its `×N` badges both say something that is not true.

**The count.** `sources` (:13) is derived from the icons that were *extracted*, not from the files that were *read*, and `iconRegistry(WEB_ICONS.length, sources.length)` presents both as the registry total. Six line icons in a file `sources.ts` does read carry their path in a JSX `d=` attribute rather than a constant, and `extractIconPaths` does not parse those: `ChevronDown`, `CheckIcon`, `PlusIcon`, `DownloadIcon`, `UploadIcon`, `CloseIcon` (`packages/web/src/components/ui/icons.tsx:22,39,65,87,109,133`). The board shows 64 of 70 and says nothing. If a globbed file ever stops yielding icons, it drops out of the *file* count too, rather than reporting an empty group.

**The badges.** `:52-56` badges `icon.names.length > 1` in `--ui-tone-attention-fg` under the title "One path, several names", and the docstring calls the list W1's deduplication list. But every merged entry here comes from an alias `ref`, not a re-typed path — `GROUP_MODE_ICONS.workspace → FOLDER_ICON`, `SORT_MODE_ICONS.recent → CLOCK_ICON`, `NAV_ICONS.agents → …`. Those are lookup tables already pointing at the one constant: the opposite of a dedup candidate. The real duplicates are in files `sources.ts` does not read (`PLUS_ICON` is byte-identical in `features/agents/memory-tab.tsx:80`, `features/company/calendar-page.tsx:108` and `components/ui/icons.tsx:65`; `TRASH_ICON` is declared in three files).

Fix: count files with `Object.keys(iconSources).length`, have the extractor report paths it saw but could not attribute so the aside can say "N unread", and badge only names whose declaration is a literal (keep alias names in the tooltip).

packages/ui-gallery/src/foundations/color.tsx
```diff
@@ -0,0 +1,245 @@
...
+import { BoardGroup } from "./shared";
+
+const SURFACES = ["--ui-canvas", "--ui-surface", "--ui-surface-muted", "--ui-inset"] as const;
+const INKS = ["--ui-fg", "--ui-fg-muted", "--ui-fg-subtle"] as const;
+const LINES = ["--ui-line", "--ui-line-muted", "--ui-line-emphasis"] as const;
+
+/** `:root[data-accent="blue"] { --ui-accent: #2563eb; … }` blocks from theme.css, as inline custom properties. */
+function presetStyle(preset: string): CSSProperties {
+  const block =
+    new RegExp(String.raw`\[data-accent="${preset}"\]\s*\{([^}]*)\}`).exec(themeCss)?.[1] ?? "";
```
hiyouga: **[over-defensive]** A failed preset parse paints the active theme's accent under another preset's label.

On a miss `block` is `""`, `style` is `{}`, and the `.gf-accent` column inherits `--ui-accent*` from `:root` — so the column labelled `blue` / `green` / `violet` / `rose` / `amber` shows **the theme's own accent**. A partial block is the same shape: a missing `--ui-accent-fg` pairs the theme's foreground with the preset's background. That is precisely the failure this board exists to catch, rendered as a plausible-looking colour.

All five blocks parse today, so this is latent — but the parser is a single-`{}` regex over raw CSS, so a nested rule, a `}` inside a comment, or a renamed token turns the column into a lie.

Fix: `themeCss` is a build-time `?raw` import, so throw at module scope when a preset block is missing or does not yield all six `ACCENT_PRESETS` names. The failure is then immediate and visible.

packages/ui-gallery/src/chrome/section.tsx
```diff
@@ -0,0 +1,164 @@
...
+          </div>
+        )}
+        <div className="g-card-foot g-chrome" data-compare={compare || undefined}>
+          {!compare && <Breadcrumb text={crumb} />}
+          <VariantPills module={module} current={variant} onPick={onPick} />
+        </div>
+      </div>
+
+      {open.size > 0 && (
+        <div className="g-drawer g-chrome">
```
hiyouga: **[bug]** `g-chrome` on the drawer wraps the Parts previews, so the same demo renders differently in the card and in the drawer.

The chain is `.g-drawer.g-chrome` → `.g-drawer-panel` → `.g-parts` → `.g-part` → `.g-preview.g-part-preview` (`drawers.tsx:71`) → the demo. Because `chrome.css` is unlayered, these beat every Tailwind utility regardless of specificity:

- `.g-chrome button, .g-chrome a { font: inherit; color: inherit }` (`chrome.css:66-70`) — every button and link in a part demo renders at the chrome's 13px `--g-font` / `--g-ink`;
- `.g-chrome button:focus-visible, a:…, input:…, summary:…` (`:71-77`) — part demos get the chrome's `--g-accent` ring instead of `--ui-focus-ring`;
- `.g-chrome code` and `.g-chrome ::selection` likewise.

The same demo rendered on the module card at `:139` is not under `.g-chrome` and renders themed, so card and drawer will disagree about the same component — in the drawer whose job is to show that component on its own. No `*.demo.tsx` exists at this head, so nothing shows it yet; it fires on the first one W1 lands.

Fix: put `g-chrome` on the drawer's own chrome elements (`g-drawer-title`, `g-part-head`, `g-card-foot`) rather than on the `g-drawer` wrapper, or add `:not(.g-preview *)` to those four rule groups.

packages/ui/src/modules/company.module.tsx
hiyouga: Derived. `TICKET_STATUSES` is now the tuple `TicketStatus` is built from, the Board maps over it (five columns), and the Status module shows one ticket per state instead of `.slice(0, 6)` — so every state, its label and its tone are on screen and none can be cut off the end. abd7acf35

packages/ui/src/modules/conversation.module.tsx
hiyouga: The `?? <first>` arms are gone. `viewFor(map, key)` (in `module.ts`) throws with the map's keys when a key is missing, so drift between `variants` and the map fails where it happens rather than drawing one composition under another's name. Both halves of your fix: `test/modules.test.ts` also renders every variant of every package-owned module and fails if two render the same markup. abd7acf35

packages/ui-gallery/scripts/shots.mjs
hiyouga: All four axes are validated against the lists the defaults come from (`THEME_IDS` and `THEME_MODES` imported from `packages/ui/src/tokens.ts`, the languages from `SPECIMEN`'s own keys, the tiers from `url-state.ts`'s list), and an unknown value returns 2 naming the known ones. abd7acf35

packages/ui-gallery/scripts/shots.mjs
hiyouga: Collected like `fontFailures`: a module whose `/embed` reports `data-renderable="false"` is recorded, printed at the end and the run returns 1. abd7acf35

packages/ui-gallery/src/chrome.css
hiyouga: No second arguments: `.g-card` and `.g-preview` read `--ui-canvas`, `--ui-fg`, `--ui-font-sans`, `--ui-text-body-size` and `--ui-text-body-lh` bare, with a comment pointing at `foundations/shared.tsx`'s rule. A dropped token now looks broken in the card the themes are judged in, and the drawer's attribution is honest about it. abd7acf35

packages/ui/src/modules/stats.module.tsx
hiyouga: The unit joined the dictionary: `copy.traces.tps(rate)`, used by the Overview tile and by the three screen sites that spelled `tok/s` inline. Both locales spell `tok/s`, which is what the app does in zh as well (`strings.ts` says so where the model card's figures are defined). `outputTps` is read — it labels the Overall row and titles the stats chip — so it stays. abd7acf35

packages/ui/src/modules/composer.module.tsx
hiyouga: `f.copy.models.perMTok(usd(model.pricing.output))`, as navigation does. abd7acf35

packages/ui/src/modules/company.module.tsx
hiyouga: The week is the fixture's: `calendar` gained `nowIso` and `utcOffsetH` beside `weekStartIso`, and one `week(f)` helper derives the Monday, the offset, "now" and `today = floor((now − start) / 86_400_000)`. The header's bold column and the now-line both read `today`, so moving the fixture week moves them with it. abd7acf35

packages/ui-gallery/src/chrome/rail.tsx
hiyouga: `checked={state.compare !== false}` with the comment explaining the three states; `onChange` already wrote the boolean, so the switch now clears a pinned module in one click. abd7acf35

packages/ui-gallery/src/chrome/drawers.tsx
hiyouga: Exhaustion is its own state: the hook returns `"unmounted"` after the 20 tries and the drawer says so (`S.section.notMeasured`, both locales) instead of sitting on "Resolving tokens…". abd7acf35

packages/ui-gallery/src/lib/tokens-read.ts
hiyouga: Counted and surfaced. The walk moved into a pure `tokensFromRules(sheets, probe)` returning `{ names, unreadable }`; `tokensReadBy` wraps it, and the drawer title carries `S.section.unreadRules(n)` when anything was rejected. The new function is tested — matching, nesting through the parent, an at-rule inside a style rule, a selector the probe cannot parse — which also closes the "`tokensReadBy` is untested" finding in the body, without a DOM. abd7acf35

packages/ui-gallery/src/modules/screens.module.tsx
hiyouga: The module builds `variants` from `SCREENS` (`SCREEN_LIST.map(…)`), the `!screen` arm is gone (the id is a variant key, resolved with `viewFor`), and `screen.tsx` asks the module for its default key instead of re-deriving it. abd7acf35

packages/ui/src/modules/content.module.tsx
hiyouga: The Log variant takes the `exec_command` it means (`find(… && i.name === "exec_command")`, throwing on a miss), destructures `cmd` with no cast fallback, renders `clone.output!`, and drives the glyph and the exit code from `clone.state`. abd7acf35

packages/ui/src/modules/stats.module.tsx
hiyouga: The fixture is one row per day now — `UsageDay { day, cacheRead, cacheWrite, output }` — so the series cannot skew and no `!` is left in the chart. abd7acf35

packages/ui-gallery/test/modules.test.ts
hiyouga: The count is asserted: `expect(DEMOS.byId.size).toBe(0)` with a message telling W1 to raise it, plus a named `it.skip` for the demos that are still to land. A glob that stops matching now fails here rather than going quiet. abd7acf35

packages/ui/src/modules/status.module.tsx
hiyouga: Collected from the data: the reasons are the distinct `stopReason` values of `f.trace.turns[].events`, and `TraceEvent.stopReason` is now the `StopReason` union, so `REASON_TONE` is total and typechecked. The module shows what this run reported (`completed`, `tool_use`). abd7acf35

packages/ui/src/modules/actions.module.tsx
hiyouga: `const agent = f.agents[1]!;` abd7acf35

packages/ui/src/modules/files.module.tsx
hiyouga: Derived from the fixture: `openFolders(f)` is the root plus the ancestors of `f.filePreview.path` and of every path a tool call changed (which keeps `test/` open for the new test file), and the Drop overlay names `folderOf(f.filePreview.path)`. Renaming the project in the fixture now moves both. abd7acf35

packages/ui/src/modules/content.module.tsx
hiyouga: One `CodeBlock({ lang, path?, code, copy, numbered? })` and one `CodeLines` in `parts.tsx`; `content`, `actions` and the file preview call them, and the baked-in `"bash"` is a call-site argument. abd7acf35

packages/ui-gallery/src/chrome.css
hiyouga: `.g-app` is off both selectors, so `code` and `::selection` inside a preview are the theme's — on the page, in the compare frames and in every PNG alike. The Foundations boards' `<code class="gf-caption gf-mono">` labels follow the root tier under test now; `page-top.png` is the refreshed shot that shows it. abd7acf35

packages/ui-gallery/src/lib/demos.ts
hiyouga: `MATRIX_KEY` is rejected alongside the shape check, with the reason in the message, and `demos.test.ts` pins it. While there: `parseVariantKey` no longer reads `all` as a matrix for a demo with no axes, mirroring `defaultPick`. abd7acf35

packages/ui-gallery/src/foundations/icons.tsx
hiyouga: Both halves. The aside counts the files read (`WEB_ICON_FILES`) and adds `iconsUnread(n)` from a new `unreadPaths()`, which reports path-data strings no constant carries — the six JSX `d=` icons — so the board says 64 icons from 4 files · 6 drawn inline, not read. The `×N` badge now counts `declaredNames`, the names whose declaration spells the path out, so an alias is a name in the tooltip and not a duplicate. Both are pinned in `test/icon-registry.test.ts`. abd7acf35

packages/ui-gallery/src/foundations/color.tsx
hiyouga: `parseAccentPresets(css)` throws on a block it cannot read and on one that declares fewer than the six names, and the page parses at module load. Under Vitest CSS imports are not processed, so `themeCss` is empty there and the parse is the suite's own: `color.test.ts` reads `theme.css` from disk, checks all five presets and their six values, and pins both failures. abd7acf35

packages/ui-gallery/src/chrome/section.tsx
hiyouga: `g-chrome` moved off the drawer wrapper onto the drawer's own elements — the titles, the Parts drawer's head and foot rows, and the Tokens and Code panels, which hold no preview. A part demo in the drawer now renders themed, exactly as the same demo on the card does. abd7acf35

PR #765, opened 2026-09-16
feat(web,ui): W0a — the penguin-ui package, the token contract, and the default theme wired with zero visual change

hiyouga:
W0a of the theme system: the private, source-only `@prismshadow/penguin-ui` package, its token contract, and the current look wired through it as the default theme (`github`, display name Primer) with no visual change to the app.

Design: Prism-Shadow/penguin-harness-design#175. Stacked on this PR (base `feat/theme-w0-package`): #761 (W0b fonts + Frost/Console), #762 (W0d guards), #763 (W0e fixtures + screens), #764 (W0c gallery).

**Review round (2026-09-18):** the alias layer is gone (436931e3f), `--ui-fg-on-emphasis` is the ink on the neutral emphasis fill and the eight labels are `text-white` again (d436f81dc), the dark rule declares only what dark changes (61d590556), theme-shots fails on a missed Benchmark click (cf1310310), and the `.font-sans` move is recorded (f5b8f97c8, deviation 14).

**Second round (2026-09-18):** `--ui-fg-on-emphasis` is dropped as a duplicate of `--ui-tone-neutral-emphasis-fg`, leaving 186 names here (dffd7fb41), and the app's Tailwind scan skips `packages/ui/src/modules` (a94ff3cfb).

**Fourth round (2026-09-19):** current `main` is merged into the branch and propagated up the stack (`855ccad6b`), so the guards above are recorded against what ships. Nothing else changed here. On vps3: `pnpm typecheck`, `pnpm lint` 0/0, `pnpm format:check`, the web suite (186 files, 2438 tests) and the web build all pass; the package has no suite of its own until #762.

## What changed

- **`packages/ui`** (new, private, no build): `tokens.ts` holds the contract, 186 `--ui-*` names in 15 groups, plus `THEME_IDS`, `TONES` (the new `done` and `info` included) and `ACCENT_PRESETS`. `theme.css` has the `dark` variant, the `ui-theme` / `ui-accent` layers after Tailwind's, the `@theme inline` bridge, the base rules that read tokens, and the five accent presets. `themes/github.css` defines every token: the base rule all of them, the dark rule only the values dark changes. `boot.ts` holds `BOOT_SCRIPT` and `applyThemeAttributes()`. `modern.css`, `geek.css`, `fonts/`, `fixtures/` and `testing/` are stubs that W0b/d/e fill.
- **Gray bridge.** In light, `--color-gray-*` is re-pointed to Tailwind's stock values as exact `oklch()`. In dark, 950/900/800/700 get the app's pure-black overrides. Because of that, the 3,555 existing gray classes follow the theme without any TSX edits.
- **Pre-paint.** `BOOT_SCRIPT` is inlined in `index.html`. It sets `html.dark`, `data-theme`, `data-accent` and the root font size before the first frame. That fixes the dark flash and the 16px→18px first-frame reflow. The dark `theme-color` is now `#000000`.
- **`state/theme.tsx`** gains `themeId` (key `penguin.themeId`, default `github`), applied through `applyThemeAttributes`. No UI offers it yet. The key is also classified in `install-scope.ts`, because the existing guard requires that.
- **Web wiring.** The `workspace:*` dependency, which pnpm links to `packages/ui`; Vite, vitest and tsc read the source through the package's `exports`. `styles.css` imports the package CSS, adds `@source "../../ui/src"`, and excludes demos, `modules/`, `screens/`, `fixtures/` and `testing/` from the scan. It also drops the rules the package now owns. `--font-app-sans` became an alias of `--ui-font-sans`, and `--accent-bg` / `--accent-fg` stay as aliases for one wave.
- **Hand edits.**
  - 27 `var(--accent-bg/fg)` → `bg-accent` / `border-accent` / `ring-accent` / `text-accent-fg`.
  - 3 dialog backdrops `bg-black/45` → `bg-[var(--ui-overlay-backdrop)]`.
- **`packages/web/scripts/theme-shots.mjs`** is the pixel-diff harness described under Verification.
- `ports.ts` reserves **7372** for the gallery. The CI `web-cli` shard now runs `@prismshadow/penguin-ui` tests. A changelog pair is included, and the stale comments that pointed at the moved `styles.css` rules were updated, including in the frontend skill.

## Deviations from the architecture (A-architecture.md §1–2)

1. **`--ui-inset` utility alias is `surface-inset`, not `inset`.** A `--color-inset` would make Tailwind also emit the existing static `ring-inset` as a ring colour.
2. **`--radius-2xl` = `calc(var(--ui-radius-xl) * 4 / 3)` and `--radius-3xl` = `xl × 2`,** not `var(--ui-radius-xl)`. The app uses `rounded-2xl` (the mobile sheet) and `rounded-3xl`, and mapping 2xl to xl would shrink them from 16px to 12px.
3. **The light gray bridge uses Tailwind v4's exact `oklch()` values,** not the hex values in the doc. Rounded hex values move pixels.
4. **`--color-white` stays literal `#ffffff` in github dark.** `dark:border-white/10`, `dark:bg-white/10`, `dark:hover:bg-white` and `ring-white` read it as light ink. Light keeps `var(--ui-surface)`, which is `#fff`.
5. **None of the 12 `text-white` became a token class,** and 3 of the 6 `bg-black/N` became the backdrop token. The contract names the label of the neutral solid fill `--ui-tone-neutral-emphasis-fg` (`--ui-fg-on-emphasis`, which duplicated it, is gone); the app's white labels sit on palette fills (`bg-gray-900 dark:bg-gray-100/200`, `gray-700`, a photo scrim) that no token pairs with yet. The lightbox's `bg-black/50` and `/70` scrims also stay. All of these move with their components in later waves.
6. **The code/pre/kbd base rule now reads `--ui-font-mono`,** which is Tailwind's stock mono stack. That stack adds Monaco and "Courier New" to the old rule's stack, but it resolves to the same face on Linux, macOS and Windows.
7. **`BOOT_SCRIPT` differs from the doc's sketch in four ways.** It is an IIFE, so no globals leak. It validates stored values against the allowed sets, so a key like `constructor` cannot pick a function. The script is generated from constants. The HTML copy carries `<!-- prettier-ignore -->`.
8. **No Vite plugin, vitest alias or tsconfig `paths`.** pnpm links the `workspace:*` dependency to `packages/ui` (`link:../ui` in the lockfile), so every tool reads the live source through `exports`; the web `dist` is byte-identical without the layer. #761 adds a `penguinUi()` that only emits the font licences.
9. **Accent presets set all six accent tokens,** not four.
10. **`ACCENT_SWATCHES`' neutral swatch is not yet rendered from the live `--ui-accent`.** That is a visible change, so it moves to W9 with the "Theme default" relabel.
11. **`github.css` values the app never had** (tones on tints, fills, `done`/`info`, code/diff, glass, type roles) are chosen to clear the contrast suite in #762, and nothing reads them yet:
    - Tone ink on tints and solid fills use the 700 step. Neutral ink is `gray-600` in light and `gray-300` in dark.
    - Dark tints are 950/40, matching the app's dark notices.
    - Dark `--ui-line-muted` is `gray-800`, and dark `--ui-surface-muted` is `#161616`, so neither vanishes against the surface.
12. **The control rungs describe the app's real controls,** not Primer's 28/32/40px. Content heights are 1.5, 2 and 2.5rem (Button sm, Button md, Input base), which is 27/36/45px at the 18px root, or 29/38/47px with borders. Reaching 28/32/40 would change every control's height. The derivation is written above the density group in `github.css`.
13. **The gallery port is 7372, not 7370** (coordinator decision).
14. **`.font-sans` changes stack.** The bridge re-points Tailwind's `--font-sans`, so the class now reads the app's `--ui-font-sans` instead of Tailwind's stock `ui-sans-serif, system-ui, sans-serif, …`. One element uses it, the full-prompt preview folded into the Create-with-AI dialogs (`ai-create-panel.tsx:63`): Latin text keeps its face (`system-ui` is second in both stacks), CJK resolves through the named PingFang SC / Microsoft YaHei instead of the OS fallback, the same face on most systems but not guaranteed on Windows. The pixel run cannot see it (the fold is collapsed, the run is on Linux); the changelog says so.

## Verification

- **Second round, on vps3 at dffd7fb41:** `pnpm typecheck`, the web suite (181 files, 2391 tests), `pnpm lint`, `pnpm format:check` and the web build pass. The CSS loses only `--ui-fg-on-emphasis:#fff;` (−25 B) and the JS the name (−22 B). The modules exclusion moves nothing here, where no modules exist yet; on #764 it takes 12,431 B of module-only utilities out of the web CSS.
- **Review round, on vps3 at f5b8f97c8:** `pnpm typecheck` (every package), the web suite (181 files, 2391 tests), `pnpm lint`, `pnpm format:check` and the web build pass. The web `dist` at 436931e3f (the alias layer removed, nothing else) is byte-identical to f2ab6fcbf's, 82 files; at f5b8f97c8 only the index CSS (−4,207 B: the trimmed dark rule and the now unused `.text-fg-on-emphasis`) and the index JS (−72 B: the eight labels) differ, and every other chunk only by the hash it imports. The pixel run below was not repeated: the labels are back to main's classes, and the trimmed dark declarations resolve to the same values.
- `pnpm lint`, `pnpm format:check`, `git diff --check`: clean. Typecheck for `@prismshadow/penguin-ui` and `@prismshadow/penguin-web`: clean.
- On vps3, `pnpm --filter @prismshadow/penguin-web --filter @prismshadow/penguin-ui test` → web 175 files, 2323 tests passed; ui has no tests yet (#762 adds them).
- The contrast and contract logic from #762, run against this `github.css` (a tsx script, not vitest), gives 44/44 pairs in light and 44/44 in dark with no contract problems.
- **Built CSS diff, `main` vs this branch (prettified):**
  - Token declarations were added.
  - Theme variables became `var(--ui-*)` reads.
  - Unused aliases for the new utility classes were added (`bg-surface`, `text-fg-muted`, `border-line`, `caret-line`).
  - The literal accent, backdrop and white classes were swapped for token-backed ones.
  - No other rule changed.
- **Pre-paint.** Checked with the app bundle blocked, so only the inline script runs:
  - Stored `dark` / `geek` / `blue` / `lg` gives `class="dark"`, `data-theme=geek`, `data-accent=blue` and a 20px root.
  - With nothing stored and a dark system preference, the page gets `dark` and an 18px root.
  - Garbage values give no attributes and an 18px root.
  - No globals leak.
  - `main` sets none of these.
- `pnpm build` (the full `pnpm -r build` chain) and the root `pnpm typecheck` (`gen:ifaces` + every package) pass.
- **Pixel diff, threshold 0: 48/48 shots identical.**
  - **Setup.** `node packages/web/scripts/theme-shots.mjs capture --out <dir> before=<main dist> again=<main dist> after=<branch dist>`, then `diff before after`. It covers 12 pages (login, new chat, chat with a tool call + Markdown table/code/links, Trace dock, user menu, Settings › Appearance, Agents, Agent settings, Plugins, Models, Cost Center, a Benchmark) × light/dark × en/zh.
  - **Control.** `before` vs `again` (the same `main` build shot twice) is also 48/48 identical, so the harness's noise floor is zero.
  - **Shared conditions.** All three builds are served by one server and one data set, with a frozen `Date.now`.
  - **Rendering.** Chromium runs without the GPU. Each shot is taken only after 3 identical frames, in a fresh context.
  - **Re-shoots.** A page is re-shot when variants disagree, up to 3 rounds. One page (en light settings) needed a second round, and every other page settled in one.
  - **Excluded from the comparison.**
    - The session cost figure is masked, because it is a load-order data race.
    - The login page is shot under reduced motion, because its 14s circuit animation otherwise shifts anti-aliasing behind the form.
  - The before/after flash and reflow fix only changes first-frame behaviour, which settled shots do not capture; the pre-paint check above covers it.


packages/ui/src/vite-plugin.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * `penguinUi()` — the Vite plugin every consumer of this package installs.
+ *
+ * The package is source-only. The workspace injects workspace packages
+ * (`injectWorkspacePackages: true`), so `node_modules/@prismshadow/penguin-ui` is a snapshot that
+ * pnpm only re-syncs after an install or a `build` script — and this package has no build. The
+ * alias therefore points every `@prismshadow/penguin-ui` specifier, JS and CSS alike
+ * (`@tailwindcss/vite` resolves `@import` through Vite's resolver), at the live `src/` beside
+ * this file, following the `exports` map in package.json.
```
hiyouga: **[simplify]** The premise doesn't hold: web's `node_modules/@prismshadow/penguin-ui` is a symlink, not a snapshot.
- This branch's lockfile resolves web's dependency as `version: link:../ui`, and so does every stacked branch (#761–#764), including after the fontsource dependencies land.
- Every installed tree has `packages/web/node_modules/@prismshadow/penguin-ui -> ../../../ui`: the local theme-w0-* worktrees, and all 14 checkouts under `~/penguin-wt` on vps3 that carry the package.

Through the symlink, Vite, `@tailwindcss/vite` (`style` condition), Vitest and tsc (`moduleResolution: Bundler`) already reach the live `src/` via the `exports` map. That makes these dead weight:
- `penguinUiAliases`, `EXACT_SUBPATHS` (a hand copy of `exports` with a keep-in-step note) and `escapeRegExp`;
- the `resolve.alias` in `web/vitest.config.ts`;
- the three `paths` entries in `web/tsconfig.json`.

Two of them do harm:
- The `@prismshadow/penguin-ui/*` → `../ui/src/*` wildcard bypasses `exports`, so tsc accepts `@prismshadow/penguin-ui/tokens` and Vite then rejects it.
- `./vite` is exported, aliased and path-mapped, while lines 11–14 forbid importing it by name.

Verified on vps3 with a scratch copy of `f2ab6fcbf`, after a fresh `pnpm install --frozen-lockfile` (which made the symlink again). With the `penguinUi()` line, the vitest alias and the three `paths` entries removed:
- `packages/web/dist` is byte-identical to the build with them (`diff -r`, 80 assets);
- `penguin-web` typecheck passes;
- the web suite passes, 181 files and 2391 tests.

Fix: delete them. In this PR `penguinUi()` is then empty, so drop it together with the `./vite` export; #761 re-adds a plugin for the licence emission only. The stacked web and gallery configs drop the same lines. Two texts name the aliases and need the same edit: design #175's 05-ARCHITECTURE sentence (consumers read the source through the workspace dependency, a Vite alias and tsconfig `paths`), and the changelog bullet.

packages/ui/src/themes/github.css
```diff
@@ -0,0 +1,511 @@
...
+    --ui-focus-ring-input: 0 0 0 2px color-mix(in oklab, var(--color-gray-400) 30%, transparent);
+    --ui-selection-bg: rgb(0 0 0 / 0.1);
+    --ui-scrollbar-w: thin;
+    --ui-scrollbar-thumb: rgb(60 64 67 / 0.28);
+    --ui-scrollbar-thumb-hover: rgb(60 64 67 / 0.45);
+    --ui-scrollbar-track: transparent;
+    --ui-scrollbar-visibility: auto;
+  }
+
+  :root.dark,
```
hiyouga, line 281: **[simplify]** Contract shape: 130 of this dark block's 187 `--ui-*` declarations repeat the light block verbatim. That includes six whole groups that can't vary by mode: shape, families, type scale, density, motion and icons (96 names). `:root.dark` also matches `:root`, so a dark block only has to declare what the mode changes.

The copies aren't free:
- `tokens.ts:15-17` promises equal heights everywhere, but nothing checks equality. #762's contract test checks presence per mode.
- W1a's radius and type retune has to edit both blocks, and an edit to one block alone splits light from dark silently.
- The cross-theme half already differs on #761: `--ui-control-py-lg` is `0.5rem` in modern and geek, `0.625rem` here.

Either:
- let dark blocks declare only mode-dependent tokens, and have the contract test check the union per mode (the cascade still gives every mode every token); or
- keep this shape, and have #762 assert that these groups are equal across modes, and across themes where they must not differ.

Settle it before #761 copies the pattern into two more files.

packages/ui/src/themes/github.css
```diff
@@ -0,0 +1,511 @@
...
+    --ui-overlay: var(--color-gray-900);
+    --ui-overlay-backdrop: oklab(0% 0 0 / 0.45); /* = bg-black/45, alpha exactly .45 */
+
+    /* ---- Colour — text ---- */
+    --ui-fg: var(--color-gray-100);
+    --ui-fg-muted: var(--color-gray-400);
+    --ui-fg-subtle: var(--color-gray-500);
+    /* The neutral emphasis fills invert in dark (`bg-gray-900 dark:bg-gray-100`), and so does the
+       label on them: `text-white dark:text-gray-900`. */
+    --ui-fg-on-emphasis: var(--color-gray-900);
```
hiyouga: **[bug]** `--ui-fg-on-emphasis` means two different things in the stack:
- Here it's the label on the neutral fill that inverts in dark: gray-900 in dark, to mirror `text-white dark:text-gray-900`.
- #761's Frost and Console set it to `#ffffff` in dark.

No fill token pairs with it, so none of the eight migrated sites is actually tokenised. Their fill is still palette, and it isn't one fill:
- `dark:bg-gray-100` at `chat-input.tsx:2504` and `:2890`, `context-gauge.tsx:766` and `overview-page.tsx:298`;
- `dark:bg-gray-200` at `user-avatar.tsx:71`, `channel-header.tsx:79`, `channel-view.tsx:577` and `shared.tsx:459`.

Every site also keeps `dark:text-gray-900`, so this dark value is read nowhere, and nothing is visible in this PR. The first component that trusts the token on one of these fills renders gray-900 in Primer dark and white on `#e0dcd6` in Frost dark.

Fix: revert the eight labels to `text-white`, which is still zero change and a smaller diff. Then write the token's meaning next to its name in `tokens.ts` before W1 reads it. Alternatively, keep the edits and make the dark value agree across the three theme files.

packages/ui/src/theme.css
```diff
@@ -0,0 +1,228 @@
...
+  --color-chart-5: var(--ui-chart-5);
+  --color-chart-6: var(--ui-chart-6);
+  --color-chart-cache-read: var(--ui-chart-cache-read);
+  --color-chart-cache-write: var(--ui-chart-cache-write);
+  --color-chart-output: var(--ui-chart-output);
+  --color-chart-grid: var(--ui-chart-grid);
+  --color-chart-axis: var(--ui-chart-axis);
+
+  /* Tailwind's own namespaces follow the theme, so existing classes need no edits. */
+  --font-sans: var(--ui-font-sans);
```
hiyouga, line 121: **[copy]** This line changes one rendered font stack, which the PR's "No other rule changed" and the changelog's "pixel-identically" don't account for. `.font-sans` used to be Tailwind's stock `ui-sans-serif, system-ui, sans-serif, 'Apple Color Emoji', …`; it now reads the app stack (the built sheet has `.font-sans{font-family:var(--ui-font-sans)}`).
- One element uses the class: the prompt `<pre>` in `ai-create-panel.tsx:63`.
- Latin text is unchanged, because `system-ui` is second in both stacks.
- CJK there now resolves through the named `"PingFang SC"` / `"Microsoft YaHei"` instead of the OS fallback. That's the same face on most systems but not guaranteed on Windows.
- The pixel run can't catch it: the fold sits in the Create-with-AI dialogs, collapsed by default; none of the 12 pages opens them; and the run is on Linux.

The new stack is what the class was reaching for, so no code change is needed. List it next to deviation 6, and qualify the changelog lead.

packages/web/scripts/theme-shots.mjs
```diff
@@ -0,0 +1,707 @@
...
+  {
+    id: "benchmark",
+    run: async (page) => {
+      await page.goto(`${BASE}/benchmark`);
+      await page.waitForTimeout(1500);
+      await page
+        .getByText(/Example Benchmark|示例/)
+        .first()
+        .click({ timeout: 5000 })
+        .catch(() => {});
```
hiyouga: **[over-defensive]** With `.catch(() => {})`, the Benchmark shot is whatever `/benchmark` happens to render. If the seeded Example Benchmark is renamed or moved, or the click misses, every variant shoots the list and the run still reports all shots identical. Drop the catch so the capture fails instead. The `try {} catch {}` wrappers around `srv.kill()` and `mock.close()` (492–497) cover calls that don't throw, and can go too.

hiyouga (commented):
Independent review of `f2ab6fcbf`, read against A-architecture §1–2 and §7 (W0a), the K-redesign notes on #765, and design PR Prism-Shadow/penguin-harness-design#175.

**Does today's default look change?** Only where the design says it should, plus one small unlisted case. I checked every rule this PR moves or re-points by reading, including the edits in pages the 12-page pixel run doesn't open: company, dock launcher, update row and modal, context gauge.
- Each hand edit gives the same computed colour in light and dark, and under all five accent presets. `--ui-accent*` carries the old `--accent-bg/fg` values, and a preset still beats `.dark`: by layer order here, by specificity on main.
- Nothing in `packages/web` puts the `dark` class on any element but `<html>`, so narrowing `.dark {…}` to `:root.dark` drops nothing.
- No call site combines a bridged `shadow-sm/lg/xl/2xl` with a `shadow-<colour>`, so losing the `--tw-shadow-color` hook changes nothing.
- The swapped utilities keep Tailwind's sort order against every same-property class on the same element.
- The base rules that moved into `theme.css` share no property with the `styles.css` base rules they now precede.
- The radius, shadow, transition and gray values equal Tailwind 4.3.2's stock `theme.css`.

What a user can still notice:
1. The first frame: no dark flash, and no 16→18px reflow. Intended (A §2.4).
2. In dark system mode, mobile browser chrome turns `#000000` instead of `#030712`. Intended (A §2.4) and listed.
3. `code`/`pre`/`kbd` gain Monaco and "Courier New" in their stack. Listed as deviation 6, and on stock installs of all three OSes it resolves to the same face.
4. `.font-sans` changes stack. Not listed; see `theme.css:122`.

**Package boundary and packaging are clean.**
- `packages/ui` is private and has no build. Web's `workspace:*` dependency on it is fine: `check-publishable` only checks published packages, and web is private.
- Docker's `COPY . .`, `install --frozen-lockfile` and `--filter penguin-web build` all work with a source-only package.
- The desktop copies `web/dist`.
- The release runs `pnpm -r build` and publishes only core, server, cli and the plugins.
- No font files are added. The branch merges into today's main without conflicts (`git merge-tree`).
- The `--accent-bg`/`--accent-fg` aliases are still needed: open #752 and #718 add new `var(--accent-bg)` uses.

**i18n:** no user-facing strings added. **Changelog:** the pair mirrors section for section.

**Finding with no line in the diff**
- **[copy]** Design #175, 06-PROTOTYPE's token-contract paragraph, says the full token list and values live in the shared UI package's README, and A §8 asks for that README too. No PR in the stack (#765, #761–#764) adds `packages/ui/README.md`. Add it here, since this PR creates the package: the groups are in `tokens.ts` and the values in `themes/*.css`; also cover the utility names the bridge exposes, the layer order, and how a consumer wires the package in. Or point the spec at `tokens.ts` instead.

Decide before #761 lands: the per-mode restatement of mode-independent tokens (`github.css:273`).

Verdict: fix before merge: the alias layer (`vite-plugin.ts:4`), the two meanings of `--ui-fg-on-emphasis` (`github.css:301`), the package README the spec points to, the unlisted `.font-sans` change (`theme.css:122`), and the swallowed benchmark click (`theme-shots.mjs:451`)

packages/ui/src/vite-plugin.ts
hiyouga: Fixed in 436931e3f: `vite-plugin.ts`, the `./vite` export, the vitest aliases and the tsconfig `paths` are gone (web `dist` byte-identical on vps3); #761 keeps a licence-only `penguinUi()` (fe7bd903b) and the gallery dropped its alias and `paths` too (957a83742).

packages/ui/src/themes/github.css
hiyouga: Fixed in 61d590556 (user decision: a dark rule declares only what dark changes; Frost and Console in c7e593d0e), with `--ui-control-py-lg` at the specified 0.625rem in all three themes (c4ef75bfa) and #762's contract test checking base plus each mode and rejecting dark repeats (5fa7e85be).

packages/ui/src/themes/github.css
hiyouga: Fixed in d436f81dc: `tokens.ts` now defines it as the ink on the neutral emphasis fill `--ui-tone-neutral-emphasis`, white in both modes in all three themes, and the eight labels are back to `text-white` with `dark:text-gray-900`; #762's contrast suite measures that pair (5fa7e85be).

packages/ui/src/theme.css
hiyouga: Fixed in f5b8f97c8: the changelog lead says what "pixel-identical" covers and a bullet describes the `.font-sans` stack move; the PR description lists it as deviation 14.

packages/web/scripts/theme-shots.mjs
hiyouga: Fixed in cf1310310: the Benchmark click now throws when it misses, and the try/catch around `srv.kill()` and `mock.close()` is gone.

PR #674, opened 2026-09-11
feat(server,web,plugins): settings groups on the Settings dialog's Plugins page — plugin options, the sandbox, and its backends' own settings

Myriad-Dreamin:
> **Stack:** `main` ← #599 ← #546 ← #656 ← #383 ← **this** ← #672 ← #551 ← #718 ← #719 ← #544 ← #448 ← #450 ← #577 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #637 ← #628 ← #642 ← #639 ← #645 ← #641 ← #526 ← #726 ← #727 — one linear chain, in this order since 2026-09-14. Review the diff against `feat/plugin-list`.

Split out of #383 on 2026-09-10, so the plugin platform lands without a configuration surface and this one is reviewed on its own. It is the plugin configuration mechanism that #641 had introduced for the Discord bot, generalized — and, since 2026-09-14, the place the sandbox is configured: #673 is merged into this PR, and its bespoke page became data on this mechanism.

## What

- **A module declares a settings group as a contribution.** A contribution to `PluginConfigProvider.groups` — data only, so a plugin's is in its generated `ifaces.json`, not `package.json`. Its id is the group's name; its data a titled group of typed fields — `string`, `secret`, `boolean`, `number`, `enum` (with `options`) or `list` (lines, optional `maxItems`) — each with a title, description, placeholder, default and `required` in English and optionally Chinese, plus `parent` (drawn inside another group's card) and `order`. A malformed declaration drops that group with a warning. There is no Project-picker type: a Project is named by its id, a string.
- **The declaring module reads its own group.** It `requires` `PluginConfig` and pulls it: `get(name)` (stored values over the declared defaults), `watch(name, cb)`, `saved(name)`. Nothing carries values on a module's behalf; each turns the document into its own typed settings.
- **The kernel: a data-only contribution does not order the boot.** A slot's owner is created after its contributors only where the slot takes a code half (`packages/core/src/kernel/module.ts`, with a kernel test), so a module may declare a group and require `PluginConfig` in the same class. PRFC-0001's sentence is updated on Prism-Shadow/penguin-harness-design#117.
- **Two nodes.** `PluginConfigProvider` holds values and takes the declarations; `PluginConfigPage` is what the admin API reads and takes live notices as code (`PluginConfigPage.status`). One node would close a cycle: the sandbox's status reads the sandbox service, and a sandbox backend reading its own group comes after the store.
[...]

PR #767, opened 2026-09-17
Each Session keeps its own permissions, and the composer shows the level

Myriad-Dreamin:
> **Stack:** `main` ← #768 (replaces #551) ← #771 ← #769 (replaces #674) ← #672 ← #718 ← #719 ← #544 ← #448 ← #450 ← #577 ← #614 ← #613 ← #619 ← #623 ← #625 ← #626 ← #637 ← #628 ← #642 ← #639 ← #645 ← #641 ← #526 ← #726 ← #727 ← **this** — one linear chain, in this order since 2026-09-17. Review the diff against `fix/bwrap-vendored`.

## Summary

Changing the Sandbox settings used to change the confinement of every running agent, because the policy was one server-wide value read at each spawn. Now each Session keeps a snapshot of the policy, taken when the Session is created.

- **Per-Session snapshot.** The policy is stored on the Session row (migration 12 adds `sessions.sandbox`). The settings only seed new Sessions. A subagent confines under its root Session's policy. Forks, model switches and handoffs carry the policy over.
- **Admin ceiling kept.** A non-admin may tighten a Session but cannot pick a filesystem mode or network looser than the server settings. The API answers `403 sandbox_forbidden`.
- **Permission button.** The composer's approval control is now one shield glyph whose colour gives the level: red for all, amber for partial, green for read-only, grey for off. The menu has Filesystem, Network, and More…, which unfolds the four approval modes inside the panel.
- **API.** `SessionInfo.sandbox` is `{ mode, network }`. Create and `PATCH /api/sessions/:id` accept `sandbox`. `GET chat-defaults` also serves the starting policy. In core, `CreateAgentOptions.confineSpawn(ctx)` is evaluated per Session, like `controlEnv`.

Design: Prism-Shadow/penguin-harness-design#176.

## Backward compatibility

Existing rows have an empty `sandbox` column. Each one takes the settings in force at its next command and keeps them from then on. The fallback stays until a later migration backfills the column; see `changelog/unreleased/2026-09-16-backward-compatibility.md`.

## Known gap

Filesystem and Network confine commands only. `write_file` and `edit_file` are still governed by the approval mode alone, as they were before this change.

## Verification

- `pnpm -r typecheck`, `pnpm lint`, `pnpm format:check` are clean.
- core 1223 passed, server 2101 passed, web 2200 passed. One server test (`POST /api/version/restart`) fails only in this shell, which inherits `PENGUIN_SUPERVISED=1`. It passes with that variable unset.
- New `session-sandbox.test.ts` covers the admin ceiling, the row round-trip, migration on the swap path, a confiner that ignores a settings change, and the HTTP flow: snapshot at creation, a settings change that leaves an existing Session alone, a non-admin tightening (200) and loosening (403).
- New `permission-level.test.ts` covers the level rule and tones. The draft-cache test covers the stored picks.


Myriad-Dreamin:
Closed: #767 was split on 2026-09-17, at the maintainer's request, into the two PRs its work belongs to.

- **#771** (sandbox backends) now carries the Windows account sandbox — the unreviewed `feat/sandbox-winuser` and `fix/bwrap-vendored` branches below this PR, and this PR's Windows fixes (the real home, full access that can cut the network, the setup script's grants).
- **#769** (plugin settings) now carries this PR's session permissions (per-Session sandbox snapshot, admin ceiling, permission button; the migration is now number 9) and the settings side of the Windows backend (its options and setup button).

The stack's top tree is unchanged apart from the migration renumbering (sessions-sandbox 9, machines-columns 10, sessions-surface 11, user-profile-adoption 12). The branches `feat/session-permissions`, `feat/sandbox-winuser` and `fix/bwrap-vendored` are deleted.

PR #769, opened 2026-09-17
feat(server,web,plugins): settings groups on the Settings dialog's Plugins page

Myriad-Dreamin:
Stack: main ← #771 ← **#769** ← #808 ← #718 ← #544 ← #448 ← #450 ← #577 ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

The plugin configuration mechanism that #641 had introduced for the Discord bot, generalized, and the place the sandbox is configured.

## What

- **A module declares a settings group as a contribution.** A contribution to `PluginConfigProvider.groups` — data only, so a plugin's is in its generated `ifaces.json`, not `package.json`. Its id is the group's name; its data a titled group of typed fields — `string`, `secret`, `boolean`, `number`, `enum` (with `options`) or `list` (lines, optional `maxItems`) — each with a title, description, placeholder, default and `required` in English and optionally Chinese, plus `parent` (drawn inside another group's card) and `order`. A malformed declaration drops that group with a warning. There is no Project-picker type: a Project is named by its id, a string.
- **The declaring module reads its own group.** It `requires` `PluginConfig` and pulls it: `get(name)` (stored values over the declared defaults), `watch(name, cb)`, `saved(name)`. Nothing carries values on a module's behalf; each turns the document into its own typed settings.
- **The kernel: a data-only contribution does not order the boot.** A slot's owner is created after its contributors only where the slot takes a code half (`packages/core/src/kernel/module.ts`, with a kernel test), so a module may declare a group and require `PluginConfig` in the same class. PRFC-0001's sentence is updated on Prism-Shadow/penguin-harness-design#117.
- **Two nodes.** `PluginConfigProvider` holds values and takes the declarations; `PluginConfigPage` is what the admin API reads and takes live notices as code (`PluginConfigPage.status`). One node would close a cycle: the sandbox's status reads the sandbox service, and a sandbox backend reading its own group comes after the store.
- **`GET|PUT /api/admin/plugin-config`** (admin only). GET lists every declared group in order: schema, stored values merged onto the defaults (secrets masked `first4…last4`), `parent?`, `notices?`. PUT `{ name, values }` validates one group's update: omitted fields keep their value, `null` or `""` clears one, a secret sent back as its mask keeps the stored one, an enum must be one of its options, a list is trimmed and de-duplicated under its cap, a required field may not end up empty. 400 `plugin_config_invalid` names the field; 404 `plugin_config_unknown` for a name no group answers to. Storage: `server_settings`, `plugin-config:<group>`.
- **The sandbox is a group** (moved here from #673): mode, cut off the network, masked paths (at most 64), with a notice listing the mounted backends or saying none is. `SandboxSettings` declares it and applies the stored document on boot and after every save — only once something was saved, so a hot swap never un-confines; the service's parked context is unchanged. `/api/admin/sandbox` and the hand-written section are gone.
- **Backends' own settings, inside the Sandbox card.** Each backend declares a group with `parent: "sandbox"` and reads it at each confine through an interface of its own (`BwrapConfigReader { get }`), parsed into typed settings (`bwrapSettingsOf` → program and probe timeout; Seatbelt and MXC: program). `SandboxPolicy` carries no settings bag. Probes are cached per program.
- **The Settings dialog** (renamed from "System settings" / "系统设置" to "Settings" / "设置", `S.settings.title`; comments, tests, docs and the unreleased changelogs follow, including those further up the chain) has a Plugins page in the server group: one card per group with no parent, its children inside, one Save per card. The plugin list's header carries a gear icon button (admin) that opens the dialog there. Design side: Prism-Shadow/penguin-harness-design#159.

## Session permissions (from #767)

Changing the Sandbox settings used to change the confinement of every running agent, because the policy was one server-wide value read at each spawn. Now each Session keeps a snapshot of the policy, taken when the Session is created.

- **Per-Session snapshot.** The policy is stored on the Session row (migration 9 adds `sessions.sandbox`). The settings only seed new Sessions. A subagent confines under its root Session's policy. Forks, model switches and handoffs carry the policy over.
- **Admin ceiling kept.** A non-admin may tighten a Session but cannot pick a filesystem mode or network looser than the server settings. The API answers `403 sandbox_forbidden`.
- **Permission button.** The composer's approval control is one shield glyph whose colour gives the level: red for all, amber for partial, green for read-only, grey for off. The menu has Filesystem, Network, and More…, which unfolds the four approval modes inside the panel.
- **API.** `SessionInfo.sandbox` is `{ mode, network }`. Create and `PATCH /api/sessions/:id` accept `sandbox`. `GET chat-defaults` also serves the starting policy. In core, `CreateAgentOptions.confineSpawn(ctx)` is evaluated per Session, like `controlEnv`.
- **Backward compatibility.** Existing rows have an empty `sandbox` column; each takes the settings in force at its next command and keeps them from then on (`changelog/unreleased/2026-09-16-backward-compatibility.md`).
- **Known gap.** Filesystem and Network confine commands only; `write_file` and `edit_file` are still governed by the approval mode alone.

Design: Prism-Shadow/penguin-harness-design#176.

## Settings groups can offer an action (from below #767)

- A group's live status may list `actions()` and `run(id)`, drawn as buttons under its notices; `POST /api/admin/plugin-config/action` (`{ name, action }`) runs one.
- `sandbox-winuser` declares its group (**Open the shell's directory**, on by default) and a status that, until the accounts exist, offers **Set up sandbox accounts**: it raises Windows' own consent prompt for the setup script and reports what happened.
- bwrap's and Seatbelt's program settings fall back to what the backends run by default: the bubblewrap bwrap ships, and `/usr/bin/sandbox-exec`.

## The WSL backend's settings (2026-09-18)

#771 now carries the WSL backend in place of the account backend (#779 merged into it), and is merged into this branch. The merge drops the account backend's settings group and setup action. The commit after it adds the WSL side:

- **`sandbox-wsl` declares its group** under the Sandbox card: show the Windows drives read-only (off by default), the base Linux (Ubuntu or Alpine), the package list, a package mirror, and the distro's name. Its provider is a loader that reads them at each spawn.
- **Its card says what the machine has** (WSL and its version, the distro, the last check line by line) and offers the next step as a button: Install WSL, Initialize sandbox distro, Check confinement, Remove distro.
- **A notice may carry the `progress` tone.** The Plugins page draws it with a spinner and reads the card's notices and actions again every two seconds while one is shown.
- **An action may hand back `settled`**, the work it started. When that ends, the card that draws the group settles as after a save, so a sandbox backend that could not load before loads without one. `settled` stays on the server.

Verified at this branch's tip (Node 24, shared box): build, `pnpm typecheck`, `pnpm lint` and the Prettier check are clean, and the WSL plugin (23), the server's plugin-registry, plugin-config, sandbox and sandbox-settings suites (47) and core's exec-session suite (44) pass. The `sandbox-winuser` bullet above is history.

## Changelog (2026-09-18)

One entry, `changelog/unreleased/2026-09-07-plugin-configuration.md` (and `.zh.md`): plugin settings, the sandbox card with each backend's settings and the WSL card, and a section for permissions per Session, which used to be its own entry. The backward-compatibility note for the `sessions.sandbox` column stays a separate file, as the changelog rules require. #771 was merged in again after it took #727 and merged its own entries into one; the two of its entries this PR had edited are gone, and what those edits said about settings is in this PR's entry. Verified at this tip: build, `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, the plugin packer (four backends), and the WSL, plugin-config, sandbox, session and migration suites.

## Network levels (2026-09-18)

Design: Prism-Shadow/penguin-harness-design#176. The Sandbox card's network field and the composer's Network section offer **no network**, **local network** (localhost only) and **full access**. Local needs a backend declaring `network-local` (#771): where none does, it is greyed out and marked unsupported in both places, and the API refuses it — a settings save with 400 on the field, a Session pick with 400 `sandbox_unsupported`. `SessionSandbox` carries `localNetworkSupported` (response only), and a non-admin's ceiling ranks none < local < open. A settings status may now list `unavailable()` enum options, which the page greys out with the reason. The separate backward-compatibility changelog file is gone: nothing here has shipped, and its one sentence about Sessions from before snapshots is in the entry. Verified at this tip: typecheck, lint, format, server 2120 and web 2331 tests.

## Not in scope

The values stay on this server: the fleet sync (#448) carries a Project's plugin table to its machines, not these values.

## Verification

2026-09-17, at this branch's tip after #767's session permissions and settings work moved in (Node 24, shared box): `pnpm install --frozen-lockfile`, build, `pnpm typecheck`, `pnpm lint` and `pnpm format:check` are clean. Passing: web 2331, core 1223, server 2112 across six shards, and the plugin suites (winuser 31, bwrap 22, Seatbelt 15, DSH 5). The stack tip's tree is byte-identical to the pre-split tip plus the migration renumbering; there every suite passes except the desktop `cli-link` macOS-symlink case, which fails the same way on `main` on this box. Typecheck and the sandbox, plugin-settings, session and migration suites also pass at #719, #448 and #641. #626 fails typecheck on an `authMiddleware` reference that fails identically on its previous head and is fixed further up the stack.


plugins/sandbox-bwrap/src/index.ts
```diff
@@ -191,13 +232,43 @@ export function createPenguinBwrapProvider(internals: PenguinBwrapInternals = {}
...
   },
 })
 export class SandboxBwrap {
+  @Use() private readonly config!: BwrapConfigReader;
   @Bind("sandbox-bwrap.provider") provider!: SandboxProviderSource;
 
   setup() {
-    this.provider = loadPenguinBwrapProvider();
+    const config = this.config;
+    this.provider = loadPenguinBwrapProvider({
```
Myriad-Dreamin: **Medium: a bad saved runner can only be fixed by restarting.** The loader probes the configured runner at boot and throws if it fails. `SandboxService` records that error once, in its constructor, and never loads the backend again. So after an admin saves `/opt/bwarp` and restarts, correcting the path on the Settings page doesn't mount bwrap. Every saved mode that needs bwrap keeps refusing commands until the next restart. The same happens when `bwrap` is missing at boot and the admin later points the setting at a real binary. This contradicts the README ("a changed program is probed afresh") and the comment saying the confine-time probe stays "for a runner changed later": that probe only runs if the backend loaded in the first place. `plugins/sandbox-seatbelt/src/index.ts:256` has the same pattern.

packages/server/src/sandbox/settings-store.ts
```diff
@@ -0,0 +1,182 @@
...
+          const declined = sandbox.declined();
+          const elsewhere =
+            declined.length === 0
+              ? ""
+              : ` ${declined.join(", ")} ${declined.length === 1 ? "is" : "are"} installed, but for another platform.`;
+          const elsewhereZh =
+            declined.length === 0 ? "" : `已安装 ${declined.join("、")}，但它们适用于其他平台。`;
+          notices.push({
+            tone: "attention",
+            text: `This deployment has no usable sandbox backend: a mode confines nothing until one for this platform is installed from the Plugins page.${elsewhere}`,
```
Myriad-Dreamin: **Low: this notice says the opposite of what happens.** "a mode confines nothing" (选择任何模式都不会产生约束) suggests turning on a mode is harmless. In fact `SandboxService.pick` fails closed, so any mode other than `danger-full-access` makes every agent command fail. An admin who believes this notice will break every command.

packages/web/src/features/settings/plugins-section.tsx
```diff
@@ -0,0 +1,418 @@
...
+      const v = valueOf(field, draft[name]);
+      if (typeof v === "number" && !Number.isFinite(v)) {
+        setFieldErrors((prev) => ({
+          ...prev,
+          [`${entry.name}\0${name}`]: S.settings.pluginFieldNotNumber,
+        }));
+        return false;
+      }
+      if (!sameValue(v, entry.values[name])) changed = true;
+      values[name] = v;
```
Myriad-Dreamin, line 212: **Low: saving a card sends every non-secret field, not just the changed ones.** `entry.values` has the defaults merged in, so declared defaults (`cutNetwork: false`, `writableTemp: true`, `probeTimeoutSeconds: 5`) get stored as real values. If a later version changes a default, any deployment that ever saved the card keeps the old value. This also defeats `applyUpdate`'s rule that omitted fields keep their value, so the page can save one field at a time.

plugins/sandbox-bwrap/src/index.ts
```diff
@@ -143,19 +180,23 @@ export async function loadPenguinBwrapProvider(
...
+    internals.settings ??
+    (() => ({ runner: internals.runner ?? "bwrap", probeTimeoutMs: PROBE_TIMEOUT_MS }));
+  const usable = new Map<string, boolean>();
   return {
     dimensions: ["fs-write", "network", "mask-paths"],
     confine(argv, policy): ConfinedArgv {
-      usable ??= probe(PROBE_TIMEOUT_MS, runner);
-      if (!usable) {
+      const { runner, probeTimeoutMs } = settings();
+      if (!usable.has(runner)) usable.set(runner, probe(probeTimeoutMs, runner));
```
Myriad-Dreamin, line 248: **Low: an uncapped probe timeout can freeze the server.** The confine-time probe is a blocking `spawnSync` that uses the admin-set `probeTimeoutSeconds`, and nothing caps that value. After a runner change, the first spawn blocks the whole server event loop for up to that timeout. With a runner that hangs and a timeout of 600, the server is unresponsive for 10 minutes.

packages/server/src/sandbox/settings-store.ts
```diff
@@ -0,0 +1,182 @@
...
+            descriptionZh:
+              "被封禁的命令在两种模式下都可以写系统临时目录。Shell 与大多数工具需要它才能启动；关闭后临时目录同样只读。",
+            default: true,
+          },
+          maskPaths: {
+            type: "list",
+            title: "Masked paths",
+            titleZh: "屏蔽路径",
+            description:
+              "One absolute path per line, hidden from confined commands (reads included).",
```
Myriad-Dreamin, line 108: **Low: the field says absolute paths, but nothing checks it.** `applyUpdate` accepts any string, and `bwrapProfileArgs` resolves a relative path against the server's working directory. Saving `.ssh` hides `<server cwd>/.ssh` (or nothing, if that doesn't exist), and the admin gets no error.

Myriad-Dreamin (commented):
Review: one medium and four low findings, inline below. Found by reading the diff and the code around it; not verified by running it. Fixes follow on this branch.

plugins/sandbox-bwrap/src/index.ts
Myriad-Dreamin: Fixed in dfab9122d. `SandboxProviderSource` now also accepts a loader function, and bwrap and Seatbelt bind one. `SandboxService.retryFailed()` re-runs the loaders that failed, keeping routing order; when retries overlap, the latest one's outcome wins. The sandbox card's status gained `saved()`, which the page calls after a save of the card or any group inside it. It retries failed backends, and the PUT waits for that before answering, so the returned notices are current. There's a unit test (`sandbox.test.ts`) and an API test that fixes the runner through the PUT (`sandbox-settings.test.ts`).

packages/server/src/sandbox/settings-store.ts
Myriad-Dreamin: Fixed in 5e4aee814. The notice now says that until a backend for this platform is installed, every mode but Off refuses every agent command (zh matches). The docs and changelog had the same claim and are corrected too.

packages/web/src/features/settings/plugins-section.tsx
Myriad-Dreamin: Fixed in 5e4aee814. `updateOf` now sends only the fields whose value differs from the entry's current one. Clearing a box still sends `null`, so the field falls back to its default.

plugins/sandbox-bwrap/src/index.ts
Myriad-Dreamin: Fixed in 5e4aee814. Number fields can now declare `minimum` / `maximum`, checked on save and on the declared default. `probeTimeoutSeconds` is limited to 1–30, and the README says the probe blocks the server while it runs.

packages/server/src/sandbox/settings-store.ts
Myriad-Dreamin: Fixed in 5e4aee814. String and list fields can now declare `pattern` (with `patternErrorMessage`), checked on every value or line. `maskPaths` requires a POSIX, drive-letter or UNC absolute path, so saving `.ssh` returns 400 `plugin_config_invalid` naming the field.

hiyouga (approved):

PR #766, opened 2026-09-16
docs(blog): three new posts on remote control, the Evaluation Center and company mode

hiyouga:
## Summary

Adds three blog posts in English and Chinese:

- **`connect-agents-to-feishu-qq-telegram-wechat`** (practice): create a conversation, open **Remote control** from its right-click menu, bind WeChat by scanning a QR code, then talk to the agent from a phone. The post warns about the wait before the first WeChat message, covers Feishu, Telegram and QQ, and ends with limits, safety and troubleshooting.
- **`create-evaluate-evolve-agents`** (practice): a tutorial for the Evaluation Center. Have AI create a Benchmark and take a baseline, ask AI what a case tests, replace a Benchmark and delete the old one, then evaluate and optimize an agent. Diagrams show the Optimizer/Evaluator separation (the Optimizer changes the agent but never reads rubrics; the Evaluator scores strictly but changes nothing) and one optimization round.
- **`agent-company-mode`** (perspectives): an essay on company mode. It covers the move from a single workbench to shared spaces (Multica, Raft, Grok Bot, described from their own pages), what an organization adds, the one-person company, the mirror company, and the beta's limits.

`blog.test.ts` now expects 24 posts, with the new ones at the top of the practice and perspectives lists.

## How the text was produced

The same pipeline as #757:

1. An English draft from research in the code.
2. A rewrite by GLM-5.3 Flash (TokenDance) from one bare prompt, with no system prompt and no tools.
3. An editor pass against the code.
4. A Chinese translation in a separate call carrying the zh UI labels.
5. A Chinese edit pass.

Three blind reviews read the posts against the code, and this PR carries the fixes from all of them. For example:

- **Remote control**
  - Readers select the channel's tab (the dialog has no visible "Channel" label).
  - The paper-plane mark appears once the connection is enabled.
  - Wait about 15 seconds after **Connected** before the first WeChat message.
  - The QQ steps include creating the bot, the WebSocket event subscription and **Save**.
  - New conversations start in **Approve everything**, and the post says so before readers bind a bot, next to the safer **Approve read-only**.
- **Evaluation Center**
  - The baseline runs the agent on the model of the conversation that builds the Benchmark; evaluations from the **Evaluate** tab use the agent's configured model, and optimization keeps the baseline's model.
  - Calibration aims for a baseline below 50 over at most four rounds and publishes anything below 85; a Benchmark built with AI runs each case once.
  - Evaluator subagents are distinguished from the **Evaluator agent** that dispatches them.
[...]

PR #757, opened 2026-09-16
docs(landing): rewrite every blog post for readability

hiyouga:
## Summary

Rewrites all 21 blog posts, in English and Chinese, to read like standard, well-organized product writing:

- **News and release posts** follow the plain, factual style of the Obsidian blog: a short summary first, then one section per change (what users can do, how to use it, limits), with internal-only implementation detail removed.
- **Practice posts** are step-by-step tutorials (prerequisites, numbered steps, expected results, troubleshooting). `natural-language-training-loop` was turned from an argument into steps.
- **Perspectives essays** state the question and thesis up front and give each step of the argument its own section.

## How the text was produced

Per the request, each post went through a fixed pipeline:

1. The English post was rewritten by a separate model (GLM-5.3 Flash on TokenDance) from one explicit, type-specific prompt, with no system prompt and no tools.
2. An editor pass adopted what was accurate, restored anything the rewrite lost (every code span, bold label, number and URL was diffed against the old post), deleted anything invented, and checked UI labels against the Web App's own `strings-en.ts` / `strings.ts`.
3. The final English was translated into Chinese by the same model with a separate prompt carrying the product's zh UI labels and term list, then edited for translationese, UI wording (e.g. 插话 for steering) and CommonMark bold rendering next to CJK punctuation. Chinese posts keep their China-specific material and `-zh` screenshots.
4. A blind review (reviewers who had not seen the pipeline) read every post against the code, and a second pass fixed what it found (second commit).

## What the second pass changed

- **Posts describe the product as of their date.** UI labels follow the app at each post's release tag (for example **Set API key for group**, **Costs**, **Trajectory**, 模型仓库), and claims that are no longer true today (built-in tool lists, unsigned builds, offline bundles, the download page's source picking) are tied to their version instead of reading as current.
- **Factual fixes**, among them: `natural-language-training-loop` sets its approval mode explicitly and makes "your data stays local" conditional on a local model driving the agent; `penguin-harness-self-improvement-with-amd-gpu` now creates `meeting_summary_agent`, because the id rule rejects the hyphenated id the old post used (its screenshot, from the original run, is captioned as such); the 0.2.6 post's `input_subagent`, `penguin-orchestration` and environment-variable details; the Gemini post's catalog count (57 → 70); the 0.2.13 post no longer implies the server enforces "the CEO proposes, the board decides".
[...]

PR #772, opened 2026-09-17
fix(landing): blog links to a Chinese heading anchor land on the heading

hiyouga:
## Root cause

`Layout` in `packages/landing/src/router.tsx` resolved the URL hash with `document.getElementById(hash.slice(1))`. `useLocation().hash` is read from `window.location`, and browsers report a CJK fragment percent-encoded (`#升级须知` → `#%E5%8D%87%E7%BA%A7%E9%A1%BB%E7%9F%A5`), while heading ids (`slugifyHeading`) hold the decoded text. The lookup missed and the effect fell through to `window.scrollTo(0, 0)`. Three ways in, all through that one effect:

- **Deep link / shared URL** — initial load, lands at the top.
- **Hand-written `[…](#…)` link in a post** — `MdLink` opens every link in a new tab, so this is an initial load too; react-markdown emits the href percent-encoded.
- **TOC click** — native fragment navigation fires `popstate`; react-router hands the effect a new location with key `"default"`. When the post was opened from the blog list (a router `Link`, random key), that key differs from the handled one, so the effect runs and pulls the page back to the top. Later clicks keep the key `"default"` and happen to work.

`nav-state.ts` (`getActiveNavItem`) and `sections/quickstart.tsx` also read the hash, but only map ASCII home-page section ids, which percent-encoding never changes. Left as is.

**Docs site.** `packages/docs` already decoded (router effect, the TOC pin on load, the `hashchange` handler), but with a bare `decodeURIComponent`. A malformed escape such as `/quickstart-cli#100%` or `#%E5%AE` threw `URIError: URI malformed` inside the effects, React unmounted the tree, and the page went blank. Fixed in this PR as well.

## Change

- `hashTargetId(hash)` in `packages/landing/src/lib/hash.ts` and `packages/docs/src/lib/hash.ts` (the two sites share no package, like `toc.ts`): strip the optional `#`, `decodeURIComponent`, fall back to the raw fragment on a malformed escape.
- Landing router effect uses it. Docs router effect, TOC pin and `hashchange` handler use it; the docs content test's in-page anchor check resolves anchors through it too.
- `MD_COMPONENTS` in `blog-post.tsx` is exported so a test can render with the page's own pipeline.
- Slugger check: every CJK heading slug (`[\p{L}\p{N}-]` only) survives the URL round trip — `new URL('#' + id).hash` then `hashTargetId` gives back `id`. Across all 42 post files on main, `docs/blog-rewrite` (#757) and `docs/blog-new-posts` (#766), TOC ids equal the rendered heading ids, with no duplicate ids.

## Not changed: smooth scrolling on image-heavy posts

Both sites set `scroll-behavior: smooth` on `html`. For a deep link into a post whose images above the target load *during* the smooth scroll (`penguinharness-0-2-4`), the animation ends at the heading's pre-image position, above the heading. Measured in Chromium, on the ASCII `#upgrading` anchor as well, whose code path this PR does not touch, so the problem predates this PR. With reduced motion (instant scroll), CSS scroll anchoring keeps the heading at the top while the images load. A possible follow-up is `scrollIntoView({ behavior: "instant" })` in the route effect, or intrinsic image sizes in posts. It changes scroll feel, so it is left for a separate decision.

## Removed anchors (#757)

`git diff origin/main...origin/docs/blog-rewrite` removes four in-page links. The target headings still exist on the rewrite branch:

| Post | Removed link | Now reads | Before this PR | After |
| `penguinharness-0-2-4.zh.md` | `[升级须知](#升级须知)` | 处理方法见文末的「升级须知」。 | broken | works |
| `penguinharness-0-2-6.zh.md` | `[升级须知](#升级须知)` | 详见下文的「升级须知」。 | broken | works |
| `penguinharness-0-2-4.en.md` | `[Upgrading](#upgrading)` | See Upgrading at the end of this post for what to do. | worked (ASCII) | works |
| `penguinharness-0-2-6.en.md` | `[Upgrading](#upgrading)` | See Upgrading below. | worked (ASCII) | works |

Nothing is restored here. If they come back, the new landing content test fails on any anchor that names no heading.

## Verification

- vps3, `pnpm --filter @prismshadow/penguin-landing test`: 12 files, **80 passed** (new: `hash.test.ts` 7, `blog-post-anchors.test.ts` 2).
- vps3, `pnpm --filter @prismshadow/penguin-docs test`: 8 files, **60 passed** (new: `hash.test.ts` 7; `content.test.ts` 7 still pass).
- Mutation check on vps3: helper without decoding → 5 of the 9 new landing tests fail, including the content test on main's two zh posts; helper without the fallback → the malformed-escape test fails.
- Chromium (Playwright, local Vite dev), zh post `penguinharness-0-2-4`, anchor `#升级须知`:
  - before: deep link `scrollY 0`; hand-written link in a new tab `scrollY 0`; TOC click after arriving from `/blog` pulled back toward the top.
  - after: TOC click after arriving from `/blog` lands (heading top 108px); deep link with reduced motion lands (heading top 80px = `scroll-mt-20`) and stays there while images load; with smooth scrolling, see the section above.
- Chromium, docs `/quickstart-cli`: `#100%` and `#%E5%AE` on main's code → `#root` empty and 3× `URI malformed`; with the fix → page renders, no errors. `#安装参考` lands (top 80px) before and after.
- `pnpm --filter @prismshadow/penguin-landing typecheck`, `pnpm --filter @prismshadow/penguin-docs typecheck`, `pnpm lint`, `pnpm format:check`, `git diff --check`: clean.

Spec impact: none. The design repo does not describe hash resolution on the landing or docs sites.


packages/landing/test/blog-post-anchors.test.ts
```diff
@@ -0,0 +1,88 @@
...
+const render = (body: string) =>
+  renderToStaticMarkup(
+    createElement(
+      Markdown,
+      { remarkPlugins: REMARK_PLUGINS, rehypePlugins: REHYPE_PLUGINS, components: MD_COMPONENTS },
+      body,
+    ),
+  );
+
+const unescapeAttr = (value: string) =>
```
hiyouga: **[over-defensive]** `unescapeAttr` cannot change the result of any assertion in this file.

- **Heading ids (line 42):** every id is `slugifyHeading` output, so it contains only `\p{L}`, `\p{N}` and `-`. None of the five characters React escapes in an attribute (`&`, `'`, `"`, `<`, `>`) can appear in one, so this call never changes anything.
- **Hrefs (line 46):** an href that React escaped contains one of those characters, so it can never equal a heading id, unescaped or not. For example, `[x](#a&b)` renders `href="#a&amp;b"`. Neither `a&b` nor `a&amp;b` is in `ids`, so both are reported as dangling. Only the wording of the failure message differs.

Fix: delete lines 32-38 and use `match[1]!` directly in `headingIds` and `inPageHrefs`.

hiyouga (commented):
Reviewed the two `hashTargetId` helpers, every hash reader in `packages/landing/src` and `packages/docs/src`, the three new test files, the `content.test.ts` change and the changelog pair.

**Findings (ranked)**
1. [over-defensive] `packages/landing/test/blog-post-anchors.test.ts:32`: `unescapeAttr` cannot change the result of any assertion. Delete it. This is optional cleanup and does not block the merge.

**Checked, no finding**
- **Duplicated helper:** landing and docs share no workspace package, and they already carry duplicate `toc.ts`, `frontmatter.ts` and `nodeText`. A shared package for a 7-line function would cost more than the copy. The two copies are identical except for the example heading in the doc comment, and each package's own `hash.test.ts` covers its copy.
- **try/catch:** proportionate. `decodeURIComponent` on a string can only throw `URIError`. The input is whatever someone types or pastes after the `#` (for example `/quickstart-cli#100%`), and neither site has an error boundary, so a throw in the router effect unmounts the root. On landing the catch also stops this PR from adding that crash, since the old `hash.slice(1)` could never throw. The raw-fragment fallback cannot pick the wrong element: `slugifyHeading` ids never contain `%`.
- **Hash readers:**
  - Landing: an initial load, a native TOC click and an in-post anchor opened in a new tab all go through `router.tsx:52` (native clicks arrive via popstate). Landing has no `hashchange` listener, and its TOC highlight reads `entry.id`, not the hash.
  - Docs: the router effect (`router.tsx:66`), the initial TOC pin (`doc-page.tsx:193`) and `hashchange` (`doc-page.tsx:231`) all use the helper. A TOC click pins `entry.id` directly, and the search dialog's `navigate(page#anchor)` goes through the router effect.
  - Still reading the raw hash: `nav-state.ts:40` and `quickstart.tsx:66`, which only match ASCII section ids.
- **`MD_COMPONENTS` export:** acceptable. It follows `MdLink`, `MdImage`, `REMARK_PLUGINS` and `REHYPE_PLUGINS`, which the same module already exports for tests. Rendering with the page's own component map is what lets the new test catch a drift between TOC ids and rendered heading ids.
- **Fixture case in `blog-post-anchors.test.ts`:** not redundant. [#757](https://github.com/Prism-Shadow/penguin-harness/pull/757) removes all four in-page links from the posts, and after that the fixture is the only test of a hand-written CJK anchor.
- **Changelog pair:** the metadata order, the `[中文版]`/`[English]` links and the `## Details`/`## 细节` headings follow `changelog/README.md`.

Verdict: merge as is

packages/landing/test/blog-post-anchors.test.ts
hiyouga: Deleted `unescapeAttr`; `headingIds` and `inPageHrefs` now use `match[1]!` directly (984238ef9).

PR #758, opened 2026-09-16
docs: reorganize the docs site around tasks and rewrite every page

hiyouga:
## Summary

Reorganizes the documentation site in the spirit of a help center (modelled on Obsidian Help) and rewrites every page in both languages.

**New sidebar** (`packages/docs/src/lib/nav.ts`; section labels in both `strings` dictionaries):

| Section | Pages |
| --- | --- |
| Get Started | introduction, quickstart (desktop / CLI / Docker / SDK), **concepts** (new), **updates** (new) |
| Guides | web-app (now an overview), **chat**, **files**, **schedules**, **remote-control** (new), **agents**, skills, models, **usage**, **settings** |
| Advanced | **evaluation-center**, self-improvement, goal-mode, company-mode |
| How It Works | architecture, server-boot, omni-message, agent-loop, message-flow, interfaces, tools, sessions-and-traces |
| Reference | cli, server-api, configuration, security |

- The 63k-character Web App guide was split verbatim into the bold pages above, then each was rewritten; its Plugin Library and Model Configuration sections were merged into `skills` and `models` (which moved from the design section into Guides), and its source-layout block into `architecture`.
- **All existing slugs are unchanged**, so every inbound link (README, installers, compose file, blog posts, code comments) keeps working.
- New pages: `concepts` (a topic-grouped glossary checked against `specs/02-TAXONOMY.md`), `remote-control` (Feishu / Telegram / QQ / WeChat bindings, from a code survey), `updates`.

## How the text was produced

For every page: the English page was rewritten by a separate model (GLM-5.3 Flash on TokenDance) from one explicit prompt per page type (guide / quickstart / concept / reference; no system prompt, no tools); an editor pass adopted what was accurate, diffed every code span, bold label, number and URL against the old page, removed anything invented and verified UI labels against `strings-en.ts` / `strings.ts`; the final English was then translated with a separate prompt carrying the page's zh UI labels and term list, and edited for translationese, the product's Chinese wording (e.g. 插话 for steering) and anchors that point at the Chinese headings.

Much of the old content had drifted from the code, and was corrected rather than just reworded. A selection:
- four stop reasons with `error_code`/`error_message` (the pages still described six), `compaction_end` outcomes, resume behavior;
- server boot: early port bind + 503 while starting, `packages/hmr`, resource-contract families, supervised child process, versioned migrations;
- the CLI is a client of the server; `LLMOutcome` statuses; missing interface members; tool failure outcomes and MCP connection reuse;
[...]

PR #773, opened 2026-09-17
fix(web): New chat honours the Project's default Agent; the login, goal and draft slash-menu copy says what actually happens

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/179

Four Web App fixes: "New chat" now starts on the Project's new-chat defaults, and three pieces of copy stop describing behaviour the product no longer has.

## 1. New chat honours the Project's default Agent

**Root cause.** The draft page seeds each field by precedence: route state first, then the draft cache, then the Project's `[default_chat]`, then the built-in fallback. `[default_chat]` shipped in #191, but the entry points written before it kept naming an Agent in route state. The pinned "New chat" and the collapsed rail named `default_agent`, and a Workspace group's "+" named the current Agent. Route state outranks the project default, so the default Agent never showed. The chosen Agent was also persisted to the draft cache, so it went on overriding the default on later visits.

Two more paths let leftovers win over the defaults:
- The draft persists every selection on mount. A text-less draft left behind by a group "+", or by a pill pick that was never sent, kept its Agent, Workspace and approval mode in the new-chat slot. That cache rung outranks the project defaults on the next "New chat".
- With nothing cached, the Agent fell back to the last-used `currentAgent` rather than `default_agent`.

**Change.**
- New `features/chat/new-chat.ts` with two helpers:
  - `newChatAgentId(agents, defaults)`: `[default_chat].agent_id` while it names a listed Agent, else `default_agent`, else the first Agent. It reuses `pickDefaultAgent`.
  - `prepareNewChatDraft(userId, projectId)`: `parkActiveDraft`, then rewrites the slot to `{ modelRef, skills }`. Typed text is still parked, and the model carry-over and staged skills stay. Everything else a text-less draft left is released, including the `source: "benchmark"` mark of an Evaluation Center draft whose prompt was deleted, which used to turn the next ordinary chat into an evaluation run.
- The entry points name only what they are about, and all of them call `prepareNewChatDraft`:

  | Entry point | Names in route state |
  | --- | --- |
  | Sidebar pinned "New chat", header create button (chat grouping), collapsed rail, chat page empty state | nothing |
  | Agent group "+", Agent card "New chat", plugin quick start | its Agent |
  | Workspace group "+" | its path (`""` = temporary). No Agent any more |

- `draft-view.tsx`: the Agent state starts at `cached.agentId ?? null`. The resolution effect is route state > valid current value > `newChatAgentId`, and it waits for the defaults to load instead of flashing a stand-in Agent. The one-shot project-default block and the `agent` touched flag went away, since a valid selection is never replaced. The live reseed (`onDefaultsChanged`) uses the same helper.
- Thinking level: unchanged. New chat never set it. The draft shows the selected Agent's explicit level, then the project default, then medium (#191's chain).
- The AI bridge ("Create with AI") is untouched. It names `default_agent` by design, because the canned prompts depend on its preinstalled plugins, and it keeps its documented "absent Workspace keeps the cached choice" contract.

## 2. Login hint for `penguin server reset-admin-password`

**Root cause.** The footer copy predates the claim-link flow. `resetAdminPassword` sets a random password that is discarded unread and marks it as initial. It also revokes admin's sessions and prints two lines: "returned to its unclaimed state" and "Start the service: it will print a sign-in link…". The next server start then prints the framed first-login link (`printFirstLoginNotice`).

**Change.** Both dictionaries now say to stop the server and run the command. The next start prints a new first-login link, and opening it sets a new password. This mirrors the first footer line's wording.

## 3. Goal mode's image description in the "+" menu

**Root cause.** #71 folded a goal's images into the objective as path lines, so the menu showed `imagesAsPathHint` in goal mode on every model. #542 moved goal mode into the goal plugin: round 1 is the user's own message with its images, passed through as ordinary input (`startGoal`). `Session.runTask` folds those images to paths only when the model has no vision, and later rounds restate the objective text. The composer kept `vision && !goalOn`.

**Change.** The image entry's description is now `vision ? uploadImageDesc : imagesAsPathHint`, the same as outside goal mode. The stale fold comments in `chat-input.tsx` and `recall-draft.ts` were corrected. No dictionary string needed rewording: `imagesAsPathHint` is accurate wherever it is still shown.

## 4. `/compact` in the draft slash menu

**Root cause.** `DraftView` passed `onCompact={async () => undefined}`, and the menu offered `/compact` whenever a handler existed.

**Change.** The draft no longer passes the no-op handler. The menu stays gated on the handlers its host passes: only the session composer passes `onCompact`, `onSwitchModel` and `onHandoff`, so `/compact`, `/model` and `/agent` are offered there alone. `/goal` is hidden only for the subagent variant. The other draft-state commands checked out fine: `/goal` works from a draft (the first send creates the Session and starts the goal), `/model` and `/agent` were already session-only, and skill commands work. Only `/compact` needed fixing.

## Spec

Design PR https://github.com/Prism-Shadow/penguin-harness-design/pull/179 records all four:
- 06「Session 列表」: a new conversation defaults to the Agent in the Project's new-conversation defaults, then `default_agent`, then the first Agent; an in-app "New chat" entry rewrites the draft slot down to the model selection and staged skills and reseeds Agent / Workspace / approval mode from the defaults (the desktop tray's "New Session" is a full page load and does not take this path); only what the entry itself names overrides the defaults.
- 06「登录页」: footer line 1 points at the first-login link; line 2 says `reset-admin-password` returns admin to unclaimed and the next start prints a new link (matches 05「账号与鉴权」).
- 06 input area: `/compact` is not offered in draft state.
- Issue 3 already matches 05「Goal 模式」.

## Not in this PR

- Docs on main: `packages/docs/content/web-app.en.md` (goal mode bullet) says a goal's images "are always sent as scratchpad paths". Docs are being rewritten in #758. No doc repeats the reset-admin-password claim.
- The desktop tray's "New Session" loads `/chat/new` as a plain URL, so it resumes the slot like a refresh, without parking or releasing.

## Verification

- Local: `pnpm --filter @prismshadow/penguin-web typecheck` passed. `pnpm lint` (oxlint), `pnpm format` + `pnpm format:check` and `git diff --check` are clean.
- vps3: `vitest run` on 25 web files, 242 tests passed. New: `new-chat.test.ts` (9). Its evaluation-draft case fails against the previous `prepareNewChatDraft` (`expected 'benchmark' to be undefined`). Also run: `draft-cache`, `draft-sessions`, `default-agent`, `ai-bridge`, `composer-send`, `recall-draft`, `skill-slash`, `skill-use`, `slash-token`, `goal-use`, `input-history`, `example-fill`, `thinking-level`, `nav-group-collapse`, `pinned-sessions`, `session-order`, `session-row-menu`, `group-list`, `placeholders-parity`, `pages`, `api-error-text`, `control-size`, `disclosure-anchor`, `icon-scale`. String parity is enforced by the `Strings` type.
- Local Playwright (run on 29703d9, not repeated after bc3ef04), against a web-only Vite dev server with `/api` mocked, `[default_chat]` = Writer / `/srv/defaults-ws` / read-only:
  - The login footer renders the new hint in en and zh.
  - Pinned "New chat" from /agents. Before: Agent pill "General Agent", and the slash menu lists `/compact`, `/goal` and the skills. After: pill "Writer", and the menu lists `/goal` and the skills only.
  - With a stale text-less slot (`coder`, `/srv/leftover`, allow-all), "New chat" lands on Writer / `defaults-ws` / read-only.
- Browser e2e was not run here. `e2e/draft.spec.mjs` covers the group "+" flows this touches.


hiyouga (commented):
Reviewed against design PR Prism-Shadow/penguin-harness-design#179: the 06 对话页 default-Agent chain, the text-less-draft release clause, and `/compact` not being offered in draft state. The code matches all three clauses.

**Findings, ranked**
1. **[bug]** `new-chat.ts:46`: `prepareNewChatDraft` leaves `source: "benchmark"` in a text-less slot, so the next plain New chat is created as an evaluation run.
2. **[over-defensive]** `draft-view.tsx:379`: the `agents.length > 0` guard in the reseed can't be reached while a draft is mounted, and if it were reached it would keep the stale pick.
3. **[simplify]** `slash-commands.ts:31`: the draft/session state split repeats what handler presence already decides. The actual fix is deleting the no-op `onCompact`.
4. **[simplify]** `new-chat.ts:26`: `ChatDefaultsDto | null`, but no caller passes null.
5. **[copy]** PR description, which becomes the squash commit message: "Not in this PR" still says the CLI's `resetPassword.desc` promises "a fresh initial password". f4f15d3 fixed that in both locales, so drop the bullet.
6. **[copy]** The release clause in design #179 reads as covering every 「新建对话」 entry. The desktop tray's "New Session" (`packages/desktop/src/tray-menu.ts:36`) loads `/chat/new` directly, though, so a text-less leftover (for example an abandoned Agent-group "+" for `coder`) still beats `[default_chat]` there. The PR body mentions this, but the spec sentence doesn't. Name the exception or limit the clause to in-app entries.

**Checked and found sound**
- Every in-app way into `/chat/new` calls `prepareNewChatDraft` and names only its own object in route state: pinned New chat, the chat-grouping header button, the Agent and Workspace group "+", the collapsed rail, the chat-page empty state, the Agent card and the plugin quick start. The AI bridge is left alone on purpose. The parked-draft delete fallback and the `/chat` redirect aren't New-chat entries. Clicking New chat while already on `/chat/new` remounts the view (its key includes `location.key`), and the flush event has already cleared the debounce timer, so the old instance can't write the released selections back.
- The draft Agent picker's wait can't hang. A `getChatDefaults` error resolves to `{}`, members can read the route, and the defaults-changed event sets `chatDefaults` itself. Agents named by route state, the cache or a parked draft resolve without waiting. There's no wrong-Agent flash either: the pill shows "Loading…" rather than a stand-in until the fetch settles. Dropping the `agent` touched flag is safe because a valid pick is never replaced. (Pressing Send inside that window does nothing and shows no message, because `onSend` returns on `!agentId`. That was already true while the Agent list loads.)
- The goal-mode image copy matches `startGoal` (round 1 passes images through) and `Session.runTask` (it folds images into paths only when the model has no vision). The login and CLI reset copy matches `resetAdminPassword` plus `printFirstLoginNotice`. zh and en are in parity.

Verdict: fix before merge: 1, 2

packages/web/src/features/chat/new-chat.ts
```diff
@@ -0,0 +1,48 @@
...
+ * afresh. What stays is the model carry-over, as after any park or send, and staged skills.
+ * Returns the parked entry's id, or null when nothing was typed.
+ */
+export function prepareNewChatDraft(
+  userId: string,
+  projectId: string,
+  storage?: DraftStorage,
+): string | null {
+  const parked = parkActiveDraft(userId, projectId, storage);
+  clearDraftChatDefaults(userId, projectId, storage);
```
hiyouga: **[bug]** A text-less slot keeps `source: "benchmark"`, so the next ordinary New chat is created as an evaluation run.

Path:
1. Evaluation Center → Use → Evaluate. `use-benchmark-modal.tsx:178` calls `openAiChat` with `source: "benchmark"`, and the slot gets `aiPrefill` + `source`.
2. Edit the prompt. That clears `aiPrefill`.
3. Select all and delete it.
4. Click the pinned "New chat".

`parkActiveDraft` returns early because the text is empty, and `clearDraftChatDefaults` drops only agentId, workspace and approvalMode. The remounted DraftView reads `cached.source`, carries it through every persist, and `onSend` sends `body.source = "benchmark"`. The user's next ordinary conversation lands in the sidebar's Evaluations folder, with the benchmark skill still staged.

The underlying behaviour predates this PR (it's in `parkActiveDraft`), but this helper is now the one place that promises a clean slot. Fix: allow-list what survives instead of deny-listing three fields. After the park, rewrite the slot as `{ modelRef, skills }`, the two things this doc comment says stay. That drops `source` too.

packages/web/src/features/chat/draft-view.tsx
```diff
@@ -383,30 +370,20 @@ export function DraftView({
...
-          setAgentId(
-            currentAgent?.agentId ??
-              (agents.find((a) => a.agentId === "default_agent") ?? agents[0])?.agentId ??
-              null,
-          );
-        }
+        // The Agent a fresh mount would now start on (the new block's default while it names
+        // an Agent, then default_agent, then the first). Skipped while the list is empty:
+        // nothing to validate against, so the pick stays.
+        if (agents.length > 0) setAgentId(newChatAgentId(agents, d));
```
hiyouga: **[over-defensive]** `if (agents.length > 0)` guards a state that doesn't happen while a draft is mounted. `agents` is `[]` only while a Project loads or switches (`state/project.tsx` clears it in `setCurrentProjectId` and in the provider's reload effect), and a Project switch remounts this view because it is keyed by projectId.

If the state did happen, the guard would do the wrong thing. The old pick stays, and once the list arrives the resolution effect sees a valid `agentId` and never adopts the new default.

Delete the guard and call `setAgentId(newChatAgentId(agents, d))` unconditionally. It sets `null` for an empty list, and the resolution effect, which waits until the list loads, then picks `newChatAgentId(agents, d)` from the `chatDefaults` set at line 372.

packages/web/src/features/chat/slash-commands.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * `available` says whether the host supplied what a command runs on — the compaction handler,
+ * the switch handler with models to pick from, the handoff handler with Agents to pick from.
+ */
+export function builtinSlashCommands(
+  state: ComposerState,
+  available: { compact: boolean; switchModel: boolean; handoff: boolean },
+): BuiltinSlashCommand[] {
+  const commands: BuiltinSlashCommand[] = [];
+  if (state === "session" && available.compact) commands.push("/compact");
```
hiyouga: **[simplify]** The `state === "session"` gates never change the result. Only chat-page's session composer passes `onCompact` / `onSwitchModel` / `onHandoff`. draft-view passes none of them (once this PR deletes the no-op `onCompact`), and subagents-view passes none either. So `available.*` is already false outside a session.

The only state difference that matters is that the subagent variant gets no `/goal`, which is the old `variant === "session"` check. The draft fix is therefore the one-line deletion in draft-view.tsx. On top of it, this module adds:
- `ComposerState`, inferred from `onChangeModel`.
- A `Record` of all four built-ins, which forces `void onCompact?.()` at chat-input.tsx:1544.
- A test (slash-commands.test.ts:40) that pins `("draft", everything)`, a combination no caller produces.

Go back to the inline, handler-gated list with `/goal` gated on `variant !== "subagent"`. Or keep the helper, but replace `state` with a single `subagent: boolean`.

packages/web/src/features/chat/new-chat.ts
```diff
@@ -0,0 +1,48 @@
...
+import { parkActiveDraft } from "./draft-sessions";
+
+/**
+ * The Agent a new conversation starts on when nothing names one: the Project's new-chat default
+ * while it still names an Agent of this Project, else the built-in `default_agent`, else the
+ * first Agent. Null only for an empty list.
+ */
+export function newChatAgentId(
+  agents: readonly AgentSummary[],
+  defaults: ChatDefaultsDto | null,
```
hiyouga: **[simplify]** No caller passes `null`. The resolution effect returns on `chatDefaults === null` first (draft-view.tsx:292), and the reseed passes `detail.defaults` inside its `if (detail.defaults)`. The null arm exists only for new-chat.test.ts:43.

Type the parameter as `ChatDefaultsDto`, read `defaults.agentId`, and drop that assertion.

packages/web/src/features/chat/new-chat.ts
hiyouga: Fixed in bc3ef04: after parking, `prepareNewChatDraft` rewrites the slot to `{ modelRef, skills }` (removed when both are absent), so `source`, the Agent / Workspace / approval mode, leftover text and chip fields are all released; `new-chat.test.ts` now reproduces the emptied Evaluation Center draft.

packages/web/src/features/chat/draft-view.tsx
hiyouga: Done in bc3ef04: the guard is gone and the reseed calls `setAgentId(newChatAgentId(agents, d))` unconditionally; an empty list sets null and the resolution effect picks the default once the list arrives.

packages/web/src/features/chat/slash-commands.ts
hiyouga: Done in bc3ef04: back to the inline handler-gated list in `chat-input.tsx` with `/goal` gated on `variant !== "subagent"`; `slash-commands.ts`, its test and the `void onCompact?.()` workaround are deleted, and the draft simply passes no `onCompact`.

packages/web/src/features/chat/new-chat.ts
hiyouga: Done in bc3ef04: `newChatAgentId` takes `ChatDefaultsDto` and reads `defaults.agentId`; the null assertion is dropped.

PR #774, opened 2026-09-17
fix(server): a login for an unknown username costs the same scrypt check as a wrong password

hiyouga:
## Root cause

`AuthService.login` (`packages/server/src/auth/service.ts`) did:

```ts
const row = this.users.findById(userId);
const ok = row !== null && (await verifyPassword(password, row.passwordHash));
```

With no row, `verifyPassword` never ran, so the 401 for an unknown username came back without any scrypt work, while a wrong password on a real account paid for a full derivation. Measured locally at production cost (N=16384, median of 15): **0.00 ms** for an unknown username against **30 ms** for a wrong password. Response time was an account-existence oracle, even though the throttle beside it deliberately counts unknown usernames so that it is not one.

`verifyPassword` also returned early, without deriving, for a stored string it could not use (empty, wrong format, parameters scrypt refuses), which was a second, smaller oracle for such accounts.

## Change

- `password.ts`: `checkPassword` returns `match` / `mismatch` / `unverifiable` and never throws; the dead try/catch around `Buffer.from(..., "base64")` and the always-true length check it inherited are gone. `verifyPassword` keeps its contract on top of it. New `verifyAccountPassword(password, stored | null, dummyHash, check = checkPassword)`: when the account is missing, its hash is empty, or `check` finds the hash unverifiable, it checks the password against the dummy hash and returns false. Every call runs exactly one derivation.
- `AuthService.login` passes `row?.passwordHash ?? null`. The dummy hash comes from `loginDummyHash()`: made **lazily, then kept, per `AuthService`** by the injected `PasswordHasher` from a random password nobody holds, so it carries the cost parameters of the hashes this server writes (in tests, the suite's token cost). The cache holds the finished string, so a hasher failure propagates, as it does for every other caller of the hasher.
- No DI interface changes. `PasswordHasher` and `Auth` are untouched, and the generated `ifaces.json` is byte-identical before and after.

## Every credential check, audited

| Check | Account may not exist? | Result |
| --- | --- | --- |
| `POST /api/auth/login` → `AuthService.login` | yes | **fixed** |
| `penguin auth login` (CLI) | yes, but it posts to `/api/auth/login` | covered by the fix |
| `PUT /api/me/password` → `changePassword` | no: the userId comes from the authenticated session | unchanged |
| `PUT /api/me/password` (setup/desktop) → `setInitialPassword` then `login` | no | uses the fixed `login` |
| `adminPasswordIs` | startup notice gate only, not reachable over HTTP | unchanged |
| Bearer API token, first-login link, desktop claim token, preview token | no username involved; constant-time compares (`tokensEqual` / `timingSafeEqual`) | unchanged |

Other paths checked:

- **Disabled users.** No such state exists: the `users` table has no disabled or locked column, and deleting a user removes the row, which makes it an unknown username.
- **Wrong-format hashes.** `password_hash` is `NOT NULL`, but an empty or corrupted value now costs the dummy derivation too. That includes `scrypt$3$…`, which Node rejects synchronously with `ERR_CRYPTO_INVALID_SCRYPT_PARAMS`.
- **Rate limit.** The throttle is keyed by the submitted userId and applied before the lookup. Its 429 does no scrypt work for either kind of username, and the failure is recorded the same way for both. Existing test "throttles unknown usernames identically" still passes.
- **Message.** Both failures come from the same `throw new HttpError(401, "invalid_credentials", …)`. The new route test pins status, headers and body as equal.

Residuals, not changed here:

- **The first unknown-username sign-in after a boot** also computes the dummy hash, so that request takes about twice as long; a concurrent first burst may compute it more than once, which is harmless. That is the design's lazy computation, and later sign-ins reuse the cached dummy.
- **Hashes written at a different cost.** If `SCRYPT_COST` is ever raised, accounts whose hashes predate the change still verify at the old cost, which would differ from the dummy's.
- **Row size.** `findById` reads the avatar column, capped at 131072 characters. That costs well under a millisecond, far below scrypt and network jitter.

## Verification

- vps3, `packages/server`: `password`, `auth`, `first-login`, `auth-sessions`, `auth-token`, `api-token`, `me-profile`, `admin-users`, `reset-admin-password` and `authz` suites: 10 files, **84 passed**. New tests: 4 in `password.test.ts` (8 total) and 2 in `auth.test.ts` (17 total).
- The unit tests count derivations through an **injected `check`** that wraps the real `checkPassword`, never wall-clock time. An unknown username gives one derivation, against the dummy. A real account gives one, against its own hash, with no dummy. An empty, malformed or refused-parameter hash gives one, against the dummy.
- Route test: `nobody` / wrong password, `carol` / wrong password and `grace` (stored hash emptied) / her real password return identical `{status, headers, body}`: 401 `invalid_credentials`, no `set-cookie`.
- Service test with a counting `ScryptHasher` replacement: a wrong password on `admin` makes no dummy. The first unknown username makes it, and later unknown usernames reuse it.
- Mutations on the vps3 copy at `36246f8cd`, each caught:
  1. `login` back to `row !== null && verifyPassword(...)` → service test fails.
  2. `verifyAccountPassword` returning early for `null` → unit and service tests fail.
  3. An unverifiable stored hash failing without the dummy → unit test fails.

  With the code restored, the two files pass 25/25.
- Local timing probe with tsx (not vitest), production cost, median of 15, run before the review fixes (they keep one derivation per sign-in):

  | Case | Median |
  | --- | --- |
  | Before, unknown username | 0.00 ms |
  | Before, wrong password | 29.98 ms |
  | After, unknown username | 30.54 ms |
  | After, empty hash | 30.15 ms |
  | After, wrong password | 30.21 ms |
- `pnpm --filter @prismshadow/penguin-server typecheck`, `pnpm lint`, `pnpm format` + `format:check`, `git diff --check`: clean.

Spec impact: `specs/05-ARCHITECTURE.md` (多用户与数据布局) gains one clause beside the throttle's 「以免成为账号探测信号」: an unknown username or an uncheckable hash also costs one scrypt check against a dummy hash, so a failed login's timing and response do not reveal whether the account exists. Paired design PR: penguin-harness-design [#184](https://github.com/Prism-Shadow/penguin-harness-design/pull/184).


packages/server/src/auth/service.ts
```diff
@@ -177,6 +178,25 @@ export class AuthService implements Auth {
...
+  /**
+   * The hash a sign-in with no account to check is verified against (verifyAccountPassword), so
+   * it costs what a wrong password costs. Made by this server's own hasher, hence at the cost its
+   * real hashes carry, from a random password nobody holds; computed on the first such sign-in
+   * and kept, except that a failed computation is dropped so the next sign-in retries it.
+   */
+  private loginDummyHash(): Promise<string> {
+    this.dummyHash ??= this.hasher
+      .hash(randomBytes(18).toString("base64url"))
+      .catch((err: unknown) => {
```
hiyouga: **[over-defensive]** This `.catch` reset handles a failure that can't happen in practice. It works together with the swallowing try/catch in `verifyAccountPassword` (`password.ts:120-124`).

The only production hasher is `ScryptHasher` → `hashPassword(pw, SCRYPT_COST)`. Its parameters are constants (N=16384, r=8, p=1, 16 MiB, under the 128 MiB `maxmem`), so scrypt can't refuse them, and it can only reject on a native allocation failure. Every other caller of the same `this.hasher.hash` lets that propagate with no guard: `seedAdmin`, `setPassword`, `AdminService.createUser`/`resetPassword`, and `reset-admin-password`'s `hashPassword`. Only the login path has retry logic for it.

That logic currently costs:
- a promise cache that clears itself
- a swallowed catch
- the `failNextHash` fake and its two ghost-retry steps in `auth.test.ts:303-309, 330-336`
- the unit test "fails without throwing when the dummy hash cannot be made" (`password.test.ts:121-126`)
- a sentence in each changelog file

Fix: keep the resolved value, not the promise, so a failure is never stored:

```ts
private dummyHash: string | null = null;

private async loginDummyHash(): Promise<string> {
  this.dummyHash ??= await this.hasher.hash(randomBytes(18).toString("base64url"));
  return this.dummyHash;
}
```

A concurrent first burst may compute it more than once, once per process, which is harmless. Then remove:
- the try/catch in `verifyAccountPassword`, plus its "Never throws: a dummy hash that cannot be produced…" sentence
- the tests listed above
- "If computing it fails, that sign-in still fails with the same 401, and the next one tries again." / 「计算失败时，这次登录照样以同样的 401 失败，下一次登录会重新计算。」

packages/server/src/auth/password.ts
```diff
@@ -57,30 +59,70 @@ export async function hashPassword(password: string, cost: number = SCRYPT_COST)
...
+ */
+export async function verifyAccountPassword(
+  password: string,
+  stored: string | null,
+  dummyHash: () => Promise<string>,
+  check: (password: string, stored: string) => Promise<PasswordCheck> = checkPassword,
+): Promise<boolean> {
+  const outcome = stored === null ? "unverifiable" : await check(password, stored);
+  if (outcome !== "unverifiable") return outcome === "match";
+  try {
```
hiyouga: **[over-defensive]** Delete this try/catch (see the comment on `service.ts:193`). `check` defaults to `checkPassword`, which never throws, so the only thing it catches is `dummyHash()` rejecting. `ScryptHasher` can't do that with its constant parameters. The body becomes `await check(password, await dummyHash()); return false;`.

packages/server/src/auth/password.ts
```diff
@@ -57,30 +59,70 @@ export async function hashPassword(password: string, cost: number = SCRYPT_COST)
...
-  if (parts.length !== 6 || parts[0] !== "scrypt") return false;
+  if (parts.length !== 6 || parts[0] !== "scrypt") return "unverifiable";
   const n = Number(parts[1]);
   const r = Number(parts[2]);
   const p = Number(parts[3]);
-  if (!Number.isInteger(n) || !Number.isInteger(r) || !Number.isInteger(p)) return false;
+  if (!Number.isInteger(n) || !Number.isInteger(r) || !Number.isInteger(p)) return "unverifiable";
   let salt: Buffer;
   let expected: Buffer;
   try {
```
hiyouga: **[over-defensive]** This try/catch is dead code: `Buffer.from(string, "base64")` never throws. Invalid input such as `"!!"` or `"===="` decodes to a zero-length or short buffer, and line 85 already rejects the empty case. `parts[4]`/`parts[5]` are always strings because line 72 checked `parts.length === 6`. Likewise, `actual.length === expected.length` on line 92 is always true, because `scryptAsync` is asked for exactly `expected.length` bytes. These lines are older than this PR, but the PR rewrites the function and adds a "Never throws" contract, so now is the time to remove them:

```ts
const salt = Buffer.from(parts[4]!, "base64");
const expected = Buffer.from(parts[5]!, "base64");
...
return timingSafeEqual(actual, expected) ? "match" : "mismatch";
```

packages/server/test/auth.test.ts
```diff
@@ -269,6 +270,84 @@ describe("auth", () => {
...
+      expect(hashed).toHaveLength(1);
+      // …so the next unknown username makes it, and every later one reuses it.
+      await expect(deps.authService.login("ghost", "whatever-123")).rejects.toMatchObject(invalid);
+      expect(hashed).toHaveLength(2);
+      await expect(deps.authService.login("phantom", "whatever-123")).rejects.toMatchObject(
+        invalid,
+      );
+      expect(hashed).toHaveLength(2);
+      // Same algorithm and cost parameters as the hash of the real account.
+      expect(hashed[1]!.split("$").slice(0, 4)).toEqual(hashed[0]!.split("$").slice(0, 4));
```
hiyouga: **[test]** This assertion can't fail. `hashed` only records this fake's own `hashPassword(password, 2)` output, so `hashed[0]` and `hashed[1]` share the prefix `scrypt$2$8$1` by construction. The property it's meant to show, that the dummy came from the injected hasher, is already pinned by `hashed` going from 1 to 2. Delete the line and its comment.

hiyouga (commented):
Checked every credential check on main: `AuthService.login` (route and CLI `penguin auth login`, plus the `me.ts` set-initial-password sign-in), `changePassword`, `adminPasswordIs`, and the unguarded hash writers (`seedAdmin`, `setPassword`, `AdminService.createUser`/`resetPassword`, `reset-admin-password`). The fix is correct. Cost parity holds: in production the dummy comes from `ScryptHasher` at N=16384, r=8, p=1 with a 64-byte key, and `reset-admin-password` writes at the same `SCRYPT_COST`. The derivation always runs over the submitted password, and the throttle and the 401 are the same for both kinds of username. No bug found.

Findings, ranked:
1. [over-defensive] `service.ts:190-197` + `password.ts:120-124`: the dummy-hash failure/retry path handles a failure that can't happen in practice (constant scrypt parameters). Every other `hasher.hash` caller lets a failure propagate. Cache the resolved string, and delete the `.catch` reset, the try/catch, and their tests and changelog sentences.
2. [over-defensive] `password.ts:79-84`, `:92`: `Buffer.from(string, "base64")` never throws, and scrypt always returns `expected.length` bytes, so this try/catch and the length check are dead code.
3. [test] `auth.test.ts:341-342`: the prefix assertion compares two outputs of the same fake hasher, so it can't fail.
4. [spec] The login security paragraph in `specs/05-ARCHITECTURE.md` (line 1366 on the design repo's main) names the throttle as the defense against account probing (「对不存在的用户名同样计数以免成为账号探测信号」: unknown usernames are counted too, so the count isn't a probe signal). It says nothing about this second defense, and there is no paired design PR. AGENTS.md requires behavior changes to be reflected in specs. One clause is enough, e.g. 「不存在的用户名或无法校验的 hash 同样对占位 hash 做一次 scrypt 派生，失败耗时不暴露账号是否存在」 (an unknown username or an uncheckable hash also costs one scrypt derivation against a dummy hash, so how long a failure takes doesn't reveal whether the account exists).

Verdict: fix before merge: 1 and 4 (2 and 3 are small enough to go in the same push).

packages/server/src/auth/service.ts
hiyouga: Done in 36246f8cd: `loginDummyHash` caches the finished string (`this.dummyHash ??= await this.hasher.hash(...)`) with no `.catch` reset, and the `failNextHash` fake, its failure step, the "fails without throwing" unit test and the retry sentence in both changelog files are gone.

packages/server/src/auth/password.ts
hiyouga: Done in 36246f8cd: the try/catch and its "Never throws: a dummy hash that cannot be produced…" sentence are gone; the body is now `await check(password, await dummyHash()); return false;`.

packages/server/src/auth/password.ts
hiyouga: Done in 36246f8cd: `salt` and `expected` are plain `const Buffer.from(..., "base64")` with no try/catch, the empty-buffer rejection stays, and the return is `timingSafeEqual(actual, expected) ? "match" : "mismatch"`.

packages/server/test/auth.test.ts
hiyouga: Done in 36246f8cd: the prefix assertion and its comment are gone; the injected hasher stays pinned by `hashed` going from 1 to 2 on the first unknown username and staying at 2 on the next.

PR #775, opened 2026-09-17
fix(web): the Evaluation Center's score lines connect across other agents, owner-only creation is hidden from members, and English labels no longer truncate

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/179

Three Web App fixes in the Evaluation Center and around it.

## 1. Score lines break across other agents' evaluations

**Root cause.** The chart lays every evaluation on one shared x axis (one slot per scoreboard entry). `seriesValues` turned a series into a full-length array with `null` in every slot another series held, and `lineSegments` split that array at every `null` — so a slot owned by *another* agent read as a gap in this one. With two agents taking turns, every segment was a single point and no line was drawn at all (seeded scoreboard: 6 dots, 0 strokes).

**Fix.** `seriesPoints(evaluations, series)` (pure, in `benchmark-metrics.ts`) maps a series' indices to their Scores, each point at its global slot index, and the chart strokes them with `segmentPath` as one line. Point positions on the x axis are unchanged. `seriesValues` had no other caller and `lineSegments` had no caller left, so both are removed (with the `lineSegments` tests; `segmentPath` keeps its own).

There is no rule for an evaluation without a Score: the server drops any scoreboard entry whose `score` is not a number in 0..100 (`toEvaluation` / `scoreOr` in `benchmark-service.ts`), so the chart never receives one.

## 2. Owner-only actions offered to members

**Root cause.** The server guards both writes with `requireProjectOwner` — `POST /api/projects/:projectId/benchmarks` and `POST /api/projects/:projectId/agents/:agentId/traces/import` — but the Web App rendered them without reading the role: the Evaluation Center passed `onManual` to `CreateButtons` unconditionally, and the Import Trace row (System settings › General; it is not on the Evaluation Center) listed every Project the viewer can reach, members' ones included, and even started on the open Project when the viewer was only a member of it.

**Fix**, reusing the role read every other owner-only surface uses (`currentProject?.role === "owner"`, as the scheduled-tasks tab does for the same button pair):
- **Create manually** is offered to the owner only, in the page header and in the empty state (`BenchmarkCreateButtons`). A member keeps **Create with AI** (it opens a conversation, which members may start), and the first step card's text drops the manual path for members (`guideCreateMember`, both dictionaries).
- **Import Trace** lists only owned Projects and starts on the open one only when the viewer owns it (`traceImportTargets`); a viewer who owns no Project gets no row.
- **A failed Benchmark's copy** told every viewer to delete it and create it again, but only the owner has the delete button (`canDelete={isOwner}`). A member now gets the same text without the delete step, on the card (`creationFailedHintMember`) and on the Benchmark's own page (`creationFailedDetailMember`, drawn by `UnpublishedNotice`), in both dictionaries.
- Other owner-only actions on the same pages: **Delete Benchmark** was already gated, and the Benchmark's own page has no owner-only action. Nothing else was exposed.

Checked against a live server as a member (`bob`, member of `default_project`, owner of `bob-default_project`): before, the header showed Create manually and the import picker offered `default_project` (preselected) and `bob`; after, Create with AI only, and the picker offers `bob` only. Direct calls as that member return `403 owner_required` for both routes.

## 3. English labels cut off

**Root causes and fixes** (no font size changed):
- **"first evaluation" on the Benchmark card.** The score column was a fixed `w-16` (72px at the 18px root) with `truncate`, sized for "首次评估" (72px); the English label is 83px. The same column also broke "Not evaluated yet" over three lines. The column is now `min-w-16` with `whitespace-nowrap`, so it grows to its label; Chinese is unchanged (still 72px). Trade-off: on a card whose label outgrows 4rem, the sparkline sits a few px further left than on its neighbours, and the title gives up that width.
- **"Scoring rubric" in the case dialog.** The tree column was a fixed 240px, and the rubric folder row carries the "Hidden from Target Agent" badge; the English row needs 318.44px with the aside's border, which left the name 32px ("S…"). The case browser now passes one fixed `treeWidth` per language: **330px in English, 240px in Chinese**. The width does not depend on which folders are open, so the preview never moves; a long file name is cut and its row's tooltip holds the path. `FileBrowser` is unchanged, and the plugin detail keeps its width.
- **Long CEO names on the org chart.** The card is a fixed 240px box by design (the tree layout is text-free), so names truncate intentionally and the card face already carries the full name as its tooltip. What made CEO names worse than everyone else's was the CEO chip sharing the name line: 95px for the name versus 145px on any other card, and the CEO's Agent is created as `<organization name> CEO`, so it is the likeliest long name. The chip is dropped from the chart card: the CEO is always the tree's root and the server titles it "CEO", so the chip only repeated the title line. The name now gets the same 145px. (The finance page's CEO badge is untouched.)

## Changes after review

In 519324c9f291ed8866652fb17158f0a5ed5f6d01:
- The tree column's `fit-content(360px)` (a `treeMaxWidth` prop on `FileBrowser`) sized it to the widest *visible* row, so opening a folder moved the preview and a long file name widened the Chinese tree too. It is replaced by the fixed per-language width above, and `treeMaxWidth` is removed.
- The CEO chip had moved to the title line, where every CEO card read `CEO [CEO]`. It is dropped instead.
- The rule for an evaluation with no finite Score (skip the point, join its neighbours) covered input the server never sends. It is removed with its test, its doc sentences and its changelog sentence.
- The failed-Benchmark member copy in section 2 is new.

## Spec impact

- 05's API table already marks both routes owner-only, and 06 says "owner 可向任一 Agent 导入 Trace 文件". 06's scheduled-tasks tab already specifies the pattern this follows ("「手动创建」即上述弹窗，仅 owner 可见，成员只看到 AI 那一枚"). 06's Evaluation Center section says nothing about members; it could gain one clause — 「手动创建」仅 owner 可见，成员只看到「用 AI 创建」，步骤块 ① 也只提 AI 一条 — and its Trace import sentence could say the picker lists owned Projects only. Not changed here.
- 06 gives a failed Benchmark's text as 「题目难度未能校准完成，请删除后重新创建」 on the card and 「创建失败」（删除后重新创建） on its page, with no member case. Members now see it without the delete step; the spec could say so. Not changed here.
- 06 says a tested Agent's versions on one model and thinking level "连成一条线"; this makes the chart do that when other series interleave.
- 06's org chart node is 头像 + 名称 + 头衔 + 运行态圆点 + 支出 / 预算 + 工位工作区尾段, with no CEO chip; dropping the chip matches it.
- No spec impact for the layout fixes.

## Noticed, not changed

- At the default 18px root the org chart card's content is ~97–100px tall inside its fixed 92px box, so the spend ratio bar collapses to 0px height on every card (measured on both seeded cards, before and after this change). The box was sized for a 16px root while its paddings and line heights are rem-based. It needs its own decision (scale the card box with the font tier, or fix its internals in px).
- The docs' Web App guide describes the Import Trace row without the owner rule; #758 rewrites that page, so the docs follow-up belongs after it.
- `treeWidth` is in px, so it does not follow the font tier: at the 20px tier the English rubric row needs 337.17px and its name is cut by about 7px at 330px. The Chinese row needs 239.97px there and still fits.

## Verification

- vps3 (`vps3-vitest.sh`, name `z2-eval`): the benchmark, company, usage, settings, file-tree and markdown suites plus `control-size`, `icon-scale`, `disclosure-anchor`, `create-with-ai-surfaces` and the new `owner-only-actions` — 46 files, 546 tests passed. New or changed: `benchmark-metrics.test.ts` (25: `seriesPoints` for interleaved series and a single-point series), `usage-charts.test.ts` (`segmentPath` over non-adjacent slots), `owner-only-actions.test.ts` (8: rendered create buttons for owner and member, a failed Benchmark's card and page notices for owner and member, `traceImportTargets` offering, defaulting, fallback and the no-owned-Project case).
- Local: `pnpm --filter @prismshadow/penguin-web typecheck`, `pnpm lint`, `pnpm format` + `pnpm format:check`, `git diff --check`.
- Local Playwright against the dev app (own data root and ports, 18px root, light mode, en and zh), with a seeded interleaved scoreboard and an organization named "Northwind Research Laboratory". Measured with `scrollWidth` / `clientWidth` and bounding boxes (Linux falls back to DejaVu Sans, wider than macOS and Windows UI fonts). Before is `main`:

| Spot | Before (en) | After (en) | zh |
| --- | --- | --- | --- |
| Card "first evaluation" | 72 / 83px, cut | 83 / 83px | 72 / 72px, unchanged |
| Card "Not evaluated yet" | 3 lines | 1 line | 1 line, unchanged |
| Tree "Scoring rubric" | 32 / 110px, cut | 122 / 122px (tree 330px) | 75 / 75px (tree 240px) |
| CEO name | 95px visible | 145px visible, tooltip has full name | same data, same widths |
| CEO chip | beside the name, repeating the title | none; the title line reads "CEO", the only "CEO" on the card | same |
| Chart, 3 interleaved series | 0 strokes, 6 dots | 2 strokes, 6 dots, same positions | same |

  The tree and CEO rows were measured again after review; the card and chart rows are from the first revision's run, and the review commit does not touch that code (`seriesPoints` returns the same points for scored evaluations, which is all the server sends).

  After review, with a case whose `statement/sources/` holds `quarterly_revenue_reconciliation_notes_fy2026.md`, the tree column while that folder is opened and closed again:

| Locale | First revision (`fit-content(360px)`) | Now (fixed width) |
| --- | --- | --- |
| en | 318.44 → 360 → 318.44px; preview edge moves 41.56px | 330 → 330 → 330px; preview edge moves 0px |
| zh | 240 → 360 → 240px; preview edge moves 120px | 240 → 240 → 240px; preview edge moves 0px |

  The rubric row alone, laid out at max-content: en 298.70 / 317.44 / 336.17px and zh 213.97 / 228.47 / 238.97px at the 16 / 18 / 20px roots (the column adds the aside's 1px border).


packages/web/src/features/benchmark/benchmark-case-browser.tsx
```diff
@@ -350,6 +353,7 @@ export function BenchmarkCaseBrowser({ projectId, benchmarkId, caseSummary }: Pr
       emptyPreview={S.benchmark.caseFileUnavailable}
       resolveRef={resolveRef}
       treeWidth={240}
+      treeMaxWidth={360}
```
hiyouga: **[bug]** `fit-content(360px)` sizes the column to the widest row **currently in the tree**, not to the rubric row. A file row's max-content is its depth indent (6 + 14px per level), the chevron column, the icon, the full `text-sm` name (it only truncates when squeezed, `file-tree.tsx:257`) and the size tag. At depth 1 that is about 117px before the name, so any name longer than about 15 characters (`evaluation_notes.md`) pushes the column past 240px.

What this does:
- **Chinese widens too.** A case whose statement lists such a file opens with a tree wider than 240px, so the PR's "Chinese stays 240px" only holds for the seeded case.
- **The preview jumps.** Expanding a subfolder with a longer name (or collapsing it, once the close animation ends) resizes the column anywhere from 240 to 360px. The preview pane's left edge moves and the README being read reflows on each toggle. Before this change the column was fixed and only the names were cut.

Fix: keep the width independent of what is expanded. One way is a width chosen for the rubric row in this caller only (e.g. `treeWidth={locale === "en" ? 330 : 240}`), which also removes the need for `treeMaxWidth` in `FileBrowser`. If the resizing is acceptable, say so in the PR and fix the zh claim in the changelog body.

packages/web/src/features/company/chart-card.tsx
```diff
@@ -170,15 +170,19 @@ export function ChartCard({
...
+                other card: a CEO's Agent is created as "<organization name> CEO", which makes its
+                name the likeliest on the chart to be long. A name the card still cuts is in the
+                face's tooltip in full. */}
             <span className="flex items-center gap-1.5">
-              <span className="min-w-0 truncate text-[13px] leading-4 font-semibold text-gray-900 dark:text-gray-100">
-                {employee.name}
+              <span className="min-w-0 truncate text-[11px] leading-4 text-gray-500 dark:text-gray-400">
+                {employee.title}
               </span>
               {isCeo && <Badge tone="brand">{S.company.ceo}</Badge>}
```
hiyouga: **[copy]** Every organization's CEO is written with `title: "CEO"` (`packages/server/src/runtime/organization/service.ts:620`, in both languages), and `S.company.ceo` is `"CEO"` in both dictionaries. With the chip on the title line, every CEO card's second line reads `CEO [CEO]`: the same word twice, side by side. Before, the chip followed the name and the title stood alone.

Fix: draw the chip only when the title doesn't already say it (`isCeo && employee.title !== S.company.ceo`). Or drop the chip from the chart card, since the CEO is always the tree's root card and its position already marks it.

packages/web/src/features/benchmark/benchmark-metrics.ts
```diff
@@ -130,13 +131,25 @@ export function labelSeries(evaluations: readonly EvaluationLabelLike[]): Evalua
...
+  evaluations: readonly MetricSourceLike[],
   series: EvaluationSeries,
-): (number | null)[] {
-  const own = new Set(series.indices);
-  return scoreValues(evaluations).map((v, i) => (own.has(i) ? v : null));
+): LinePoint[] {
+  const values = scoreValues(evaluations);
+  return series.indices.flatMap((index) => {
+    const value = values[index];
+    return value === null || value === undefined ? [] : [{ index, value }];
```
hiyouga: **[over-defensive]** An evaluation with no finite Score can't reach this function. `toEvaluation` drops any scoreboard entry whose `score` fails `scoreOr` (a number in 0..100, so NaN and ±Infinity are gone, `benchmark-service.ts:72,191`), and JSON can't carry NaN or Infinity anyway. The PR description calls this "a defensive path" itself. Yet this PR adds a rule for it: four doc sentences (lines 138-142), a 29-line test (`benchmark-metrics.test.ts:238-266`), a sentence in both changelog files, and the clause 「没有有效分数的评估不画点，线跨过它连接前后两点」 in design #179.

Delete them. The function becomes:
```ts
export function seriesPoints(evaluations: readonly MetricSourceLike[], series: EvaluationSeries): LinePoint[] {
  return series.indices.map((index) => ({ index, value: evaluations[index]!.score }));
}
```
This also stops re-running `scoreValues` over every evaluation once per series. `scoreValues`' own null branch predates this PR and can stay for now.

hiyouga (commented):
Review of the three fixes: the chart lines, the owner-only gates and the English truncation.

**Findings, ranked**
1. **[bug]** `benchmark-case-browser.tsx:356`: `fit-content(360px)` sizes the tree to its widest *visible* row. Expanding or collapsing a folder can resize the column and move the preview pane, and the Chinese tree grows past 240px too once any file name runs past about 15 characters.
2. **[copy]** `chart-card.tsx:184`: every CEO card's title line now reads `CEO [CEO]`. The server writes the CEO's title as `"CEO"`, and the chip says `CEO` in both dictionaries.
3. **[over-defensive]** `benchmark-metrics.ts:151`: the rule for an evaluation with no score covers something the server never sends. It comes with a test (`benchmark-metrics.test.ts:238-266`), doc sentences, a changelog sentence and a spec clause in design #179.
4. **[copy]** (unchanged code, not blocking) A member looking at a failed Benchmark is told to "delete this Benchmark and create it again", but a member gets no delete button (`canDelete={isOwner}`, `benchmark-page.tsx:254`). The same text shows as `creationFailedHint` at `benchmark-page.tsx:270,281`, and `creationFailedDetail` at `benchmark-detail-page.tsx:109` says the same thing. It is the same kind of member copy this PR already fixed for Create manually (`guideCreateMember`), so the member version could leave out the delete step.

**Checked and found correct**
- `seriesPoints` + `segmentPath` handle interleaved series (the x slot is the scoreboard position, and one stroke crosses the other series' slots), a single point (bare `M`, dot only, guarded by `points.length > 1`) and an empty series. `lineSegments` and `seriesValues` had no other callers on main.
- Owner gates: both routes call `requireProjectOwner`, and `project-access.ts` has no admin bypass. Delete was already gated. Nothing else on the Evaluation Center or a Benchmark's page calls an owner-only route. `Use`, `Ask AI` and `Create with AI` only prefill conversations, the case browser only reads, and `GET models` only needs `requireProjectAccess`. So no member-visible action still returns 403.
- `TraceImportRow`: the `return null` comes after every hook. `traceImportTargets` handles a pick, the open Project, the first owned Project and "none owned" correctly.
- `FileBrowser`'s other caller (plugin detail) is unchanged: its track is still `${treeWidth}px`, and the new `md:min-w` on the aside equals that track.
- `guideCreateMember` exists in both dictionaries with the same shape, and it matches the button labels and the new clause in design #179.

Verdict: fix before merge: 1, 2, 3 (4 is optional follow-up)

packages/web/src/features/benchmark/benchmark-case-browser.tsx
hiyouga: Fixed in 519324c9f: the case browser now passes a fixed `treeWidth` per language (330px en, 240px zh) and `treeMaxWidth` is gone from `FileBrowser`, so opening a folder no longer resizes the column (measured 330→330px en, 240→240px zh; before, 318→360px en and 240→360px zh).

packages/web/src/features/company/chart-card.tsx
hiyouga: Fixed in 519324c9f: dropped the chip from the chart card, since the CEO is always the root and its title already reads "CEO"; the title line now shows "CEO" once.

packages/web/src/features/benchmark/benchmark-metrics.ts
hiyouga: Fixed in 519324c9f: `seriesPoints` is now `series.indices.map((index) => ({ index, value: evaluations[index]!.score }))`, and the unscored doc sentences, test and changelog sentence are removed; the matching clause in https://github.com/Prism-Shadow/penguin-harness-design/pull/179 is being removed there.

PR #776, opened 2026-09-17
fix(server): outbound files a channel cannot take are recorded in the cost center, not posted into the chat

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/181

## Summary

QQ was posting `"xx.md" could not be sent to the chat. 文件"xx.md"未能发送到会话。(QQ cannot receive "xx.md": sending a file to QQ requires a publicly reachable URL for it, which this server has no way to provide)` into the chat after a reply. The request: keep that out of the chat, put it in the Cost Center's error records, and do the same on every channel.

All four outbound file notices stop being posted into the chat, on Feishu, Telegram, QQ and WeChat:

1. a failed upload (`messagingFileFailedNotice`)
2. an upload the bot's app lacks a permission for (`messagingFilePermissionNotice`)
3. a file over the outbound ceiling (`messagingFileTooLargeNotice`)
4. the files past the per-reply count (`messagingFilesSkippedNotice`)

Instead, each reply files **one error record per cause** through the bridge's existing `recordError`. Each record names every file it covers, the channel and the reason. It carries the Session's Project/Agent attribution and appears in `GET /api/projects/:id/usage/errors`. The reply text is still delivered as before. Unchanged: the inbound download notices, the unsupported-message notice and the approval notice.

## What gets recorded

One reply files at most one record per row below. Rows 1-3 share a code and are filed in that order.

| # | Cause | Code | Kind |
| --- | --- | --- | --- |
| 1 | Missing upload permission | `messaging_file_send_failed` | unexpected |
| 2 | Any other refused upload (or failed read) | `messaging_file_send_failed` | unexpected |
| 3 | A refusal the channel repeats for every file (QQ) | `messaging_file_send_failed` | expected |
| 4 | Over the 10MB picture / 30MB file ceiling | `messaging_file_too_large` (new) | expected |
| 5 | Past the 5 files one reply may carry | `messaging_files_skipped` (new) | expected |

Message formats (`<channel>` is the binding id: `feishu` / `telegram` / `qq` / `wechat`):

- **A lone file:** `"<file>" was not sent to the <channel> chat: <reason>`
  - Permission: `"<file>" was not sent to the <channel> chat: this bot's app is missing a permission.`⏎`<scopes>`⏎`<grant URL>`
  - Over the ceiling: `"<file>" was not sent to the <channel> chat: it is larger than the 10MB limit for a picture` (`30MB limit for a file`)
- **Several files:** `<n> files were not sent to the <channel> chat: "<a>", "<b>", "<c>"`, followed by:
  - for a permission, `this bot's app is missing a permission.`⏎`<scopes, deduplicated>`⏎`<each distinct grant URL>`, and nothing else
  - for any other cause, the reason they share, or one `"<file>": <reason>` line per file when reasons differ
  - over the ceiling, the shared reason reads `each is larger than the 10MB limit for a picture`, and per-file lines read `"<file>": larger than the 30MB limit for a file`
- **Skipped:** `2 more files the reply mentioned were not sent to the <channel> chat: at most 5 ride along with one reply ("f.md", "g.md")`

The permission record quotes nothing the channel threw. Feishu's refusal already carries the scopes and the console link, and ends in a `log_id` that differs on every request. Quoted per file, it gave every file its own line repeating the fix, each clipped before the `log_id`. A lone file's permission record is now about 240 characters.

Examples:

```
3 files were not sent to the qq chat: "a.md", "b.md", "c.md"
QQ cannot receive files: sending a file to QQ requires a publicly reachable URL for it, which this server has no way to provide
```

```
2 files were not sent to the feishu chat: "huge.png", "vast.png"
each is larger than the 10MB limit for a picture
```

```
"report.pdf" was not sent to the feishu chat: this bot's app is missing a permission.
im:resource:upload, im:resource
https://open.feishu.cn/app/<app id>/auth?q=im:resource:upload,im:resource&op_from=openapi&token_type=tenant
```

**Length.** Every message, the skipped record included, is kept within the recorder's 500-character `MESSAGE_MAX`, instead of letting the recorder cut the reasons off the end:

1. The file list gives way first, down to `…`.
2. If per-file reasons still don't fit, they are cut to equal lengths. Each keeps its file name and the start of its reason.
3. A lone file's message is clipped with `…`.

## Classification (`error-kind.ts`)

The old rule filed `messaging_file_send_failed` as expected for the typed refusals because the chat was told. Nobody in the chat is told any more, so what can still count as routine at the outbound points is what nobody can act on:

- A structural refusal (`MessagingUnsupportedError`) and the bridge's own caps (`MessagingOutboundCapError`) → `expected`.
- A missing permission on an upload → `unexpected`. The same denial on an inbound download stays `expected`, because the chat is still given the scopes and the link there.
- Any other upload failure → `unexpected`.

Records are split by the type their kind is read from, so kinds never mix within a record. The structural-refusal and cap records carry that type. The permission and other-refusal records are plain `Error`s, which are `unexpected` at `messaging_file_send_failed` exactly as the thrown types are. The inbound fetch codes are classified exactly as before. Records already in the table keep their kind.

## Implementation notes

- **Grouping lives in one pure function.** `messagingFilesNotSentRecords(channel, oversize, refused)` in `bridge.ts` groups a reply's failures and builds the messages; it has unit tests. The skipped record's name list is fitted by the same `quotedNames`. `deliverFiles` collects failures during the loop and files the records once the batch is done. `deliverOneFile` returns the ceiling that held a file back instead of throwing.
- **New `MessagingOutboundCapError` in `media.ts`.** It is the typed error for the bridge's own caps. `MessagingMediaTooLargeError` couldn't be reused: its message is fixed to `<what> is larger than the <N>MB limit`.
- **QQ's refusal no longer names the file.** The text is now `QQ cannot receive files: sending a file to QQ requires …`. Three QQ files would otherwise be three different reasons, which means three near-identical lines, about 770 characters for five files. With the change they share one reason line (204 characters for five files). The record's first line already names every file.

## Residual behavior worth knowing

The recorder deduplicates on (source, code, Project) over 2 seconds, and rows 1-3 share a code. So in the rare reply where uploads fail for two of those causes at once (e.g. a permission denial plus a network error), only the first record survives. The order is permission, other, structural, so the kept record is the one someone has to act on. Avoiding it entirely needs a separate code for the structural refusal, or the kind in the recorder's dedup key. That is a design decision this PR does not make.

## Docs

`packages/docs/content/server-api.{en,zh}.md` said every mentioned file that didn't arrive was "said in the chat". That sentence now says each cause is one record per reply, with its code and kind, and that a name matching no file is only logged. #758 rewrites the docs, and the same claim appears there in `remote-control.{en,zh}.md` (around line 181) and `server-api.{en,zh}.md` (around line 950). It needs carrying over.

## Spec

The 消息软件接入 / 出站 and 异常落库 sections of `specs/05-ARCHITECTURE.md` still describe the in-chat notices and the old classification. They need a paired design PR.

## Verification

- `pnpm --filter @prismshadow/penguin-server typecheck`: exit 0 (after rebuilding core for the merge of main)
- On vps3, six messaging suites plus `errors.test.ts` and `usage.test.ts`: 8 files, 386 tests passed
  - `messaging.test.ts` 124
  - `messaging-telegram.test.ts` 71
  - `messaging-qq.test.ts` 35
  - `messaging-wechat.test.ts` 28
  - `messaging-wire.test.ts` 23
  - `messaging-error-kind.test.ts` 5
  - `errors.test.ts` 68
  - `usage.test.ts` 32
- On vps3, `@prismshadow/penguin-docs` tests: 8 files, 60 passed
- Negative controls on vps3, running the new or changed tests against the source they fix:
  - first commit: exactly the 8 new or changed cases failed
  - review round (9c3eada25): exactly the 5 new or changed cases failed (two permission-record tests, two skipped-record tests, and the per-cause record list)
- `pnpm format` + `pnpm format:check`, `pnpm lint`, `git diff --check`: clean
- Web: not touched. The Cost Center errors panel shows raw codes and has no per-code label table.


packages/server/src/runtime/messaging/bridge.ts
```diff
@@ -314,33 +323,240 @@ export const MESSAGING_OUTBOUND_FILE_MAX_COUNT = 5;
...
+    ];
+    const fix = ["this bot's app is missing a permission.", scopes.join(", "), ...grantUrls].join(
+      "\n",
+    );
+    records.push({
+      code: "messaging_file_send_failed",
+      err: new MessagingPermissionError(
+        scopes,
+        grantUrls[0] ?? null,
+        filesNotSentMessage(channel, permission.map(thrownReason), AS_THROWN, fix),
```
hiyouga: **[bug]** (low) With Feishu's real scope denial, this record does not list the fix "once", and the per-file lines add nothing.

`larkFailure` (feishu-sdk.ts) builds the `MessagingPermissionError` message from Feishu's own `msg`. That text already contains the scopes and the console URL (see the fixture in `messaging-wire.test.ts`, "reports the scopes to grant and the console link…"), followed by `(code 99991672, log_id …)`. Every request gets its own `log_id`, so the reasons never match and each file gets its own line. I ran `messagingFilesNotSentRecords` via tsx on that `msg` shape, with a real-length app id and log_id:

- 1 file: 500 chars. The grant URL appears twice (fix block, then inside the reason), and the clip removes `(code 99991672, log_id …)`, the only part of the reason that differs.
- 3 files: 500 chars. The head list shrinks to `…`, and every line reads `"report-N.pdf": Access denied. One of the following scopes is required: [im:resource:u…`. That is three truncated copies of what the fix block above already says.

The unit fixture `uploadDenied` (messaging.test.ts:650) uses a synthetic `Access denied (code 99991672, log_id N)`, which is why the "lists a permission's scopes and console links once" test passes. The docs (server-api.en/zh) and the changelog repeat the "listed once" claim.

Fix: for the permission group, leave out the thrown reasons. The head naming the files plus the fix block is the whole actionable record, and nobody needs a `log_id` to grant a scope. That gives about 240 chars for one file, with nothing clipped. Also change `uploadDenied` to Feishu's real `msg` shape, so the test covers what production produces.

packages/server/src/runtime/messaging/bridge.ts
```diff
@@ -314,33 +323,240 @@ export const MESSAGING_OUTBOUND_FILE_MAX_COUNT = 5;
...
+      ),
+    });
+  }
+  if (other.length > 0) {
+    records.push({
+      code: "messaging_file_send_failed",
+      err: new Error(filesNotSentMessage(channel, other.map(thrownReason), AS_THROWN), {
+        cause: other.map((file) => file.err),
```
hiyouga: **[over-defensive]** This payload is never read.

- `cause: other.map(...)`: the recorder stores only `err.message` (`messageOf` in error-recorder.ts), and `recordError` never logs `err`.
- The `scopes` / `grantUrls[0] ?? null` passed into the rebuilt `MessagingPermissionError` at 402-406: nothing reads them. Its type doesn't matter either: `messagingErrorKind` returns `unexpected` for a `MessagingPermissionError` at `messaging_file_send_failed`, the same as for a plain `Error`. Only the `unsupported` record actually needs its type.

Delete `cause`, build the permission record as `new Error(...)`, and narrow the doc comment's "each keeping that type because error-kind.ts reads the kind off it" to the unsupported record.

Same kind of thing: `Math.max(0, room)` in `fitLines` (line 541) is redundant. `clip` already returns `""` for `max < 1`, so a negative `room` produces identical output without the clamp.

packages/server/src/runtime/messaging/bridge.ts
```diff
@@ -354,22 +570,37 @@ const MESSAGING_NOTICE_NAMES_MAX = 5;
...
+ */
+function filesSkippedError(
+  channel: string,
+  fileNames: readonly string[],
+): MessagingOutboundCapError {
+  const n = fileNames.length;
+  const files =
+    n === 1 ? "1 more file the reply mentioned was" : `${n} more files the reply mentioned were`;
+  return new MessagingOutboundCapError(
+    `${files} not sent to the ${channel} chat: at most ${MESSAGING_OUTBOUND_FILE_MAX_COUNT} ride along with one reply (${nameList(fileNames)})`,
```
hiyouga: **[copy]** (low) This is the one record that is not fitted under `MESSAGE_MAX`. The PR body says "Every message is kept within the recorder's 500-character `MESSAGE_MAX`", and the changelog says "Each message is kept under the recorder's 500-character cap by shortening the file list first".

`nameList` limits how many names are listed (5), not their length. The fixed text around the list is about 106 chars. If five skipped names average more than ~77 chars (e.g. `quarterly-revenue-breakdown-by-region-and-product-line-2026-q3-final-v2-draft.xlsx`), the message goes past 500. The recorder then `slice`s it mid-name, with no `…` and without the closing `)`. The names here are also unquoted, while every other record quotes them.

Fix: build the list with `quotedNames(names, MESSAGE_MAX - <rest of the message>.length)`, as `filesNotSentMessage` does, or narrow the claim in the changelog.

hiyouga (commented):
Reviewed the diff against origin/main. What I checked:

- **Grouping:** one record per cause per reply, filed after the batch, with the skipped tail as its own record.
- **Message formats and 500-char fitting:** ran the pure functions via tsx on realistic inputs.
- **Kinds:** checked against the `error-kind.ts` rules and every typed-error construction site (only QQ throws `MessagingUnsupportedError`, and only Feishu throws `MessagingPermissionError`).
- **Nothing left in chat:** no outbound-file `sendText` remains in `deliverFiles` / `deliverOneFile`.
- **Inbound notices untouched:** they and their helpers `noticeReason` / `permissionDetail` are still in use, and no references to the removed exports remain.
- **Docs and spec:** docs en/zh agree with each other, and design PR #181 agrees with the code.

The dedup collapse of same-code records is already disclosed in the PR body. Nothing more to add there.

Findings, ranked:
1. **[bug]** (low) bridge.ts:405: a Feishu permission record repeats the scopes and URL inside every per-file reason and clips each reason before its `log_id`. The "listed once" claim in the docs and changelog doesn't hold for Feishu's real refusal text, and the synthetic test fixture hides this.
2. **[over-defensive]** bridge.ts:413 (also 402-406, 541): the `cause` array and the rebuilt `MessagingPermissionError`'s scopes and type are never read, and the `Math.max(0, room)` clamp duplicates `clip`'s own guard.
3. **[copy]** (low) bridge.ts:602: the `messaging_files_skipped` message isn't fitted under `MESSAGE_MAX`, despite the "every message" claim. Names averaging over ~77 chars get sliced by the recorder with no `…`.

Verdict: fix before merge: 1, 2 (both small); 3 optional.

packages/server/src/runtime/messaging/bridge.ts
hiyouga: Fixed in 9c3eada25: the permission record is now the file names plus the scopes and grant links listed once, with no thrown reasons (239 chars for one file); `uploadDenied` uses Feishu's real `msg` shape, and the test asserts under 500 chars, scopes and each link once, every file named.

packages/server/src/runtime/messaging/bridge.ts
hiyouga: Fixed in 9c3eada25: removed the `cause` array, built the permission record as a plain `Error` (still `unexpected`, kinds unchanged), narrowed the doc comment to the unsupported record, and dropped the `Math.max(0, room)` clamp.

packages/server/src/runtime/messaging/bridge.ts
hiyouga: Fixed in 9c3eada25: the `messaging_files_skipped` list is built with `quotedNames(names, MESSAGE_MAX - rest)`, so names are quoted and cut with `…` under 500 chars; a new test covers five 96-char names.

PR #780, opened 2026-09-17
fix(web,desktop): a reply's link to a Workspace file opens it in the Files panel, and the desktop never opens a second app window

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/182

A user on the Windows desktop app clicked `[pelican-bike.html](pelican-bike.html)` in an assistant reply, and a new PenguinHarness main window opened. Each further click opened another one.

## Root cause

1. `packages/web/src/features/chat/md.tsx` `MdLink` gave every chat link `target="_blank"`, including relative hrefs. `pelican-bike.html` resolves against the SPA route (`http://localhost:<port>/chat/pelican-bike.html`), which is on the app origin.
2. In `packages/desktop/src/main.ts`, the main window's `setWindowOpenHandler` allowed every URL for which `isLocalSurfaceUrl` was true. That exception was added for the Workspace preview's "Open in a new tab" link. Any other app-origin URL matched it too, so each click opened a BrowserWindow that booted the whole SPA. In a plain browser, the same click opened a tab on a route that does not exist.

## Web

`lib/reply-link.ts` (new) has two functions. `resolveReplyLink(href, workspace)` sorts an href into one of four kinds, and `replyLinkBehavior` turns each kind into the attributes the link gets:

| href | kind | on click |
| --- | --- | --- |
| has a scheme (`https:`, `mailto:` …) or is protocol-relative (`//host`) | external | new tab, unchanged |
| `#id` (percent-decoded) | anchor | default prevented; scrolls to the nearest element with that id. GFM footnote ids repeat in every reply, so the search starts at the link and moves outward. |
| anything else that `toWorkspaceRelative` accepts once percent-decoded and stripped of its query and hash | file | default prevented (any modifier keys); opens the file in the Files panel through the same `onOpenFile` the file card's rows use |
| the rest: a `..` that leaves the Workspace, an absolute path outside it, `~`, an href react-markdown emptied (it replaces `file:` URLs and Windows drive-letter paths with `""`) | none | default prevented; nothing opens |

- Relative hrefs are resolved the same way the file card resolves paths in replies (`toWorkspaceRelative`). A leading `./` and `.` segments drop, an inner `..` pops, and an absolute path counts only inside the Session's Workspace. Windows Workspaces compare with either separator.
- Existence is not checked. A missing file opens the panel on its path, and the panel reports it (see Known limits).
- `md.tsx`: a new `WorkspaceLinksProvider` puts the Workspace and file opener into a context, and `MdLink` reads it. It follows the same pattern as `channel-markdown.tsx`'s reader context, so the components map stays a module constant. Without a provider, links behave as before: every link opens a new tab. This covers Memory, the handbook, tickets and channels.
- `message-stream.tsx` wraps `MessageItems` in the provider, using `ctx.workspace` and `ctx.onOpenFile`. It therefore covers the main stream and the subagent conversations in the agents panel, including reasoning and compaction summaries.
- `chat-page.tsx`: `onOpenFile` is now a stable `useCallback`. A new context value on every render would re-render every link in the transcript on every streamed frame.

## Desktop

- `util.ts`: `classifyWindowOpen(url, origin)` returns `window`, `external` or `deny`. There are also two smaller helpers: `isExternalScheme` and `urlForLog`, which logs origin and path only because a query can carry a token.
  - **window** (an app window): `/api/sessions/<id>/files/preview-redirect` on the app origin; `/preview/…` on the preview host, `127.0.0.1` on the same port (the desktop always loads the App on `localhost`, and the server's `loopbackHostRoles` always assigns previews to `127.0.0.1`); `/terminal` on the app origin.
  - **external** (the system browser): an `http(s)` or `mailto` URL that is not this instance, including other local ports such as a dev server a conversation started (the process list's service link).
  - **deny** (refused, logged as `[shell] refused to open a window for <origin><path>`): every other path on this instance. This includes non-preview paths on the preview host, which 302 back to the App, and every non-web scheme, `about:blank` included. With no app origin every request is denied, although no window exists before boot sets one.
- `main.ts`: the main window and every window opened from it share one handler (`openWindowFor`). `guardOpenedWindow` re-applies the handlers on `did-create-window`, recursively. The navigation rules themselves are unchanged: the main window stays on the app origin, and a child window stays on the loopback surface.
- **Penguin Go authorization bridge (#716, kept through the merge).** The Authorize button opens `about:blank` synchronously and then points it at the platform. Only the main window's handler allows it, as a hidden, sandboxed window, and it checks for the bridge before calling `openWindowFor`. The check is not in `openWindowFor`, which every opened window shares, so a preview window asking for `about:blank` is refused. `did-create-window` passes `isAuthorizationBridgeUrl(details.url)` into `guardOpenedWindow`. Its `will-navigate` hands the authorization URL to `openInSystem` and then closes the bridge, so no invisible window is left behind. Windows opened further down pass `false`.

### Decisions for review

1. **`/terminal` is allowlisted.** The brief did not name it. `detachTerminal` opens `/terminal?id=…` with `window.open` and keeps the handle to put the tab back when the window closes. Denying it would turn "detach" into "remove the tab" in the desktop app (`dock.spec` covers the browser side).
2. **Only `http:`, `https:` and `mailto:` reach `shell.openExternal`**, whether from a window-open request or from `will-navigate`. Before, any scheme was passed through, including `file:` and custom protocol handlers, and Agent-written preview HTML could reach that call.
3. **Windows opened by a child window get the handlers too.** Before, a window opened from a preview page had no window-open handler and no navigation guard, so it could open the App again.
4. **A window-open request for an SPA route is refused, not turned into a navigation of the main window.** The shell cannot tell a real route from a relative file link that resolved onto one, which is exactly the reported bug.
5. **In a conversation, an href with nowhere to go does not navigate.** It no longer opens a new tab on a route that does not exist, and the desktop app would refuse that window anyway.

### Known limits (unchanged here)

- react-markdown's `defaultUrlTransform` empties Windows drive-letter paths and `file:` URLs before the link adapter sees them. They cannot be recognized as Workspace files, so in a conversation they are inert. Relative links, as in the report, work on every platform.
- Child windows can still *navigate* (not open windows) anywhere on the loopback surface. A preview page that links to `/` still loads the App inside that preview window, as it did before.
- A middle-click on a file link is left to the browser.
- For a missing HTML file, the panel's own not-found state is a "File does not exist." toast plus the server's JSON 404 inside the preview iframe.
- The authorization bridge can still be opened from an HTML preview in the main window's Files panel. The isolated preview iframe has `allow-popups` and runs inside the main window, and Electron's window-open handler does not say which frame asked. Agent-written HTML there can therefore open a hidden `about:blank` window and keep it alive, because the bridge closes only when it navigates off the loopback surface. Main has behaved this way since #716. Closing the gap is a design decision, so this PR leaves it as it is and notes it at the handler.

## Specs

Paired design PR: [penguin-harness-design #182](https://github.com/Prism-Shadow/penguin-harness-design/pull/182). The clauses that change in `specs/06-PROTOTYPE.md`:

- 对话页 › 消息流: "thinking 为暗色可折叠块；text 为流式 Markdown 正文。" (needs the link rules above)
- 文件浏览面板: "消息文件卡片的「点击预览」入口展开树定位到该文件并预览。" (reply links take the same path)
- 桌面端 › 进程模型: "页面内的外部链接一律转系统浏览器打开；但 **app 源的新窗口请求留在应用内**（…）：这类请求打开一个同样禁用 Node 的子窗口，子窗口可在本实例的回环面（app 源 + 同端口的预览对偶主机）内自由导航，其余一律外送。" (only the three URL shapes get a window, other app-origin requests are refused, only web and mail schemes go to the OS, and the rules apply to every descendant window. The main window alone also opens Penguin Go's authorization bridge: a hidden `about:blank` window that hands the URL to the system browser and closes itself.)

The old rule that "every chat link opens in a new tab" was never in the specs; it lived only in `md.tsx`'s comment and `md.test.ts`. The impl docs (`packages/docs/content`) do not describe chat link behaviour, so they are not touched.

## Verification

This round merged main (including #716) and addressed the review. Everything below ran on `80c647331`. The head, `ef3d9d5d6`, only changes a comment in `main.ts` on top of it, and desktop typecheck, lint and format:check ran again there.

- On vps3, where the runner builds core, hmr and server first: `pnpm --filter @prismshadow/penguin-web typecheck`, `pnpm --filter @prismshadow/penguin-desktop typecheck`, `pnpm lint` (0 warnings, 0 errors) and `pnpm format:check` are all clean. `git diff --check` is clean too.
- vitest on vps3:
  - `pnpm --filter @prismshadow/penguin-desktop exec vitest run`: 17 files, 228 tests passed. `util.test.ts` has 22 tests: 11 from this PR and #716's `isAuthorizationBridgeUrl` test. The 11 are 9 for `classifyWindowOpen`, one each for `isExternalScheme` and `urlForLog`. One of the 9 is new: it pins `about:blank → deny` as the rule that only the main window opens the bridge.
  - `pnpm --filter @prismshadow/penguin-web exec vitest run md.test reply-link message-stream chat`: 6 files, 54 tests passed. That includes `reply-link.test.ts` (15), `md.test.ts` (13) and the new `message-stream.test.ts` (1).
  - `message-stream.test.ts` replaces the source-text match in `md.test.ts`. It renders `MessageStream` with a Workspace, a file opener and one reply that links a relative and an absolute Workspace file. Both anchors must render without `target`, and each click must open its file through `ctx.onOpenFile`. It fails as expected under three mutations of `message-stream.tsx` (no opener, a different opener, no Workspace), each run on vps3 and then reverted. The chat page's stable `onOpenFile` is no longer pinned by a test, because rendering `ChatPage` is out of reach of the node-only suite.
  - The first round ran the whole web suite (176 files, 2346 tests). This round ran only the files above: its own web changes are tests, and the web code merged from main was checked in its own PRs.
- Browser check (first round, before the merge) with local Playwright against the built server and a mock LLM that replies with the reported message. The spec was throwaway and is not committed. 9 of 9 checks passed:
  - the file link renders without `target` and keeps `href="pelican-bike.html"`;
  - clicking it opens the file's preview in the Files panel, no page opens and the URL does not change;
  - a link to `./nope/missing.html` opens the panel on that path;
  - a `../outside.html` link and a footnote reference open nothing and leave the URL unchanged (no hash);
  - the external link keeps `target="_blank"`.
- The desktop window behaviour itself, including the bridge's hidden window closing after the hand-off, was not exercised, because there is no display on the Linux dev box. The unit tests cover `classifyWindowOpen` and `isAuthorizationBridgeUrl`.


packages/desktop/src/main.ts
```diff
@@ -142,54 +146,74 @@ function createWindow(url: string): void {
...
-  // in a window of this app; handing it to the system browser would land on a 401.
-  // Denying it outright (as this did at first) made the entry silently do nothing.
-  // Genuinely external links still go to the system browser.
-  win.webContents.setWindowOpenHandler(({ url: target }) => {
-    if (isLocalSurfaceUrl(target, appOrigin)) {
+  // New windows, from this window and from every window it opens, go through one rule
+  // (classifyWindowOpen): only the Workspace preview hand-off, a preview page and a detached
+  // terminal get a window of this app; other sites go to the system browser; anything else
+  // on this instance is refused, since every other app path boots a second copy of the App.
+  win.webContents.setWindowOpenHandler(({ url: target }) => openWindowFor(target, iconPath));
```
hiyouga: **[bug]** This line and `guardOpenedWindow` (lines 202-210) conflict with #716 on main. The conflict is in behaviour, not only in text, so a textual resolution can silently break Penguin Go authorization in the desktop app.

On main, `platform-key-auth-dialog.tsx:138` calls `window.open("about:blank", "_blank")` synchronously from the click. The main window's handler special-cases that URL (`isAuthorizationBridgeUrl`) and opens a hidden, sandboxed window. The child's `will-navigate` then hands the real HTTPS URL to the OS and calls `child.close()`.

In this PR, `classifyWindowOpen("about:blank")` returns `deny`, and `util.test.ts:136` asserts it. Two ways the merge goes wrong:
- **The PR's side wins:** `window.open` returns `null`, and the dialog falls through to `setPhase("ready")`. The first Authorize click opens nothing, and the user has to click again.
- **The bridge is kept, but the close is lost:** `guardOpenedWindow` has no `details`, so nothing closes the hidden window. Every authorization leaves an invisible BrowserWindow alive.

Resolution:
- Check `isAuthorizationBridgeUrl(target)` in the **main window's** handler only, before `openWindowFor`, and keep the hidden `overrideBrowserWindowOptions`. Do not move the check into `openWindowFor`: every child window shares that function, so Agent-written preview HTML could then open hidden windows.
- Pass `isAuthorizationBridgeUrl(details.url)` into `guardOpenedWindow`, and call `child.close()` after `openInSystem(target)` in its `will-navigate`.
- Keep main's `isAuthorizationBridgeUrl` test next to the `about:blank → deny` assertion. `classifyWindowOpen` should keep denying it, because the bridge is decided before classification.

packages/web/test/md.test.ts
```diff
@@ -65,6 +103,92 @@ describe("Md links", () => {
...
+    }
+  });
+
+  it("with no file opener wired, a conversation's links keep the new tab", () => {
+    const tags = anchors(renderWithOpener("[x](pelican-bike.html)", undefined));
+    expect(tags).toHaveLength(1);
+    expectNewTab(tags[0]);
+  });
+
+  it("the message stream provides the Workspace, and the chat page's opener is stable", () => {
```
hiyouga: **[test]** This test checks the source text, not the behaviour, and it fails in both directions:
- **False failure:** the `message-stream.tsx` line is 98 columns under `printWidth: 100`. One more wrapper level makes Prettier split the attributes onto separate lines, and the regex fails with nothing broken.
- **False pass:** `useCallback((path) => {…}, [fileOpenRequest])` still matches `/const openWorkspaceFile = useCallback\(/` while giving a new function on every change. That is the per-frame re-render this test says it guards against.

The provider → link contract is already tested behaviourally through `renderInConversation`. The only thing this test adds is the `MessageStream` wiring. Guard that by rendering it: render `MessageStream` with a ctx carrying `workspace` and `onOpenFile` and one assistant item containing `[x](a.html)`, then assert the anchor has no `target`. Otherwise delete the test rather than match the source.

packages/desktop/src/util.ts
```diff
@@ -55,8 +57,96 @@ export function isLocalSurfaceUrl(url: string, origin: string | null): boolean {
...
-  return loopback.has(target.hostname) && loopback.has(app.hostname);
+  return LOOPBACK_HOSTS.has(target.hostname) && LOOPBACK_HOSTS.has(app.hostname);
+}
+
+/**
+ * The host Workspace previews are served from while the App is on `appHost`: the server
+ * canonicalizes the App onto one loopback name and serves previews only from the other
+ * (loopbackHostRoles in the server's preview-token service).
+ */
+function previewHostFor(appHost: string): string | null {
```
hiyouga: **[simplify]** Both branches past the first are unreachable. The desktop's App host is always `localhost`: `appOriginFor` hardcodes it, the attach path at `main.ts:432` builds `http://localhost:<port>`, and the server's `loopbackHostRoles` returns `app: "localhost"` unconditionally. Spec 06 (desktop section) says the window must use localhost and that `127.0.0.1` is reserved as the preview host.

The comment ("canonicalizes the App onto one loopback name and serves previews only from the other") reads as if the roles could swap. They cannot. Replace the function with `const PREVIEW_HOST = "127.0.0.1"` beside `LOOPBACK_HOSTS`. On line 131, compare `target.hostname === PREVIEW_HOST`.

packages/desktop/src/util.ts
```diff
@@ -55,8 +57,96 @@ export function isLocalSurfaceUrl(url: string, origin: string | null): boolean {
...
+ * preview host redirects every non-preview path back to it, so a relative link in a chat reply
+ * (`/chat/x.html`) would boot a second copy of the App in a new window on a page that does not
+ * exist — one more window per click. Other schemes (`file:`, custom protocol handlers) are
+ * refused outright rather than handed to the OS (isExternalScheme).
+ */
+export function classifyWindowOpen(url: string, origin: string | null): WindowOpenAction {
+  if (!isExternalScheme(url)) return "deny";
+  const target = new URL(url);
+  if (target.protocol === "mailto:") return "external";
+  if (origin === null || !isLocalSurfaceUrl(url, origin)) return "external";
```
hiyouga: **[over-defensive]** Two lines here decide nothing:

- **`origin === null`** is unreachable. No window exists before `appOrigin` is assigned (`main.ts:380` / `:432`), and nothing resets it to null. The branch also fails in the wrong direction: if it ever ran, it would send this instance's own `http://localhost:<port>/…` URLs to the system browser, which gets a 401. The test at `util.test.ts:144-147` pins that as intended behaviour.
- **The `mailto:` return on line 122** is redundant. `isLocalSurfaceUrl` already returns false when the protocol differs from the App's, so a mailto URL reaches `external` on line 123 anyway.

Fix: delete line 122. Narrow the type at the caller (`openWindowFor` can return `deny` when `appOrigin === null`), or have this function return `deny` for a null origin. Drop the first assertion of the "with no origin yet" test.

hiyouga (commented):
Reviewed the reply-link classifier and resolver (`reply-link.ts` against `toWorkspaceRelative`, micromark's percent-encoding and react-markdown's URL emptying), the `WorkspaceLinksProvider` wiring (every `Md` render site: only `MessageStream`, meaning the main stream and the subagent panel, provides it; Memory, the handbook, tickets, channels and the Files panel's own Markdown adapters are unaffected), `classifyWindowOpen` against the server's `loopbackHostRoles` and every `target="_blank"` or `window.open` in `packages/web/src` on main, the recursive child-window guard, the changelog pair, and paired design PR #182.

The web side holds up. No other app surface loses a window it needs: downloads use `download`, and the only app-origin window openers are the preview hand-off and `/terminal`, both allowlisted. Relative links in Memory, the handbook and tickets are now refused in the desktop app, where they used to open a second copy of the App. That is intended (decision 4).

**Conflicts with main:** `packages/desktop/src/main.ts` and `packages/desktop/test/util.test.ts`, both against #716 (5b595e254, the Penguin Go authorization bridge). `util.ts` and the web files merge cleanly.

Findings, ranked:
1. **[bug]** `main.ts:153`: the conflict with #716 is behavioural, not textual. `classifyWindowOpen("about:blank")` is `deny` (and `util.test.ts:136` pins it). If the PR's handler wins the merge, the desktop app's Penguin Go Authorize click opens nothing, and the user has to click a second time. If the merge keeps the bridge but loses #716's `child.close()`, every authorization leaves a hidden window behind. Details inline.
2. **[test]** `md.test.ts:178-189`: the source-text regex test pins formatting, not behaviour. A Prettier reflow breaks it, while a real regression (an unstable dep array) passes.
3. **[simplify]** `util.ts:63-72`: `previewHostFor` has two unreachable branches. The App host is always `localhost`.
4. **[over-defensive]** `util.ts:122-123` and `util.test.ts:144-147`: the `origin === null` branch is unreachable and would fail open, and the `mailto:` early return repeats what `isLocalSurfaceUrl` already decides.

Verdict: fix before merge: 1 (the conflict resolution must keep the authorization bridge, restricted to the main window). Fold 2-4 into the same rebase.

packages/desktop/src/main.ts
hiyouga: Resolved in the merge c34b0b985: `isAuthorizationBridgeUrl` is checked in the main window's handler only, before `openWindowFor` (hidden, sandboxed options kept), and `guardOpenedWindow(child, iconPath, isAuthorizationBridgeUrl(details.url))` closes the bridge after `openInSystem` (descendants pass `false`); both sides' tests are kept, with `about:blank → deny` restated as the main-window-only rule. One caveat, noted at the handler in ef3d9d5d6 and in the PR description: the Files panel's HTML preview iframe (`allow-popups`) runs inside the main window, so it can still reach the bridge, as on main since #716.

packages/web/test/md.test.ts
hiyouga: Replaced in b84ec9649: the source-text test is gone, and the new `test/message-stream.test.ts` renders `MessageStream` with a Workspace and `onOpenFile`, then asserts both reply file links render without `target` and open the file through `ctx.onOpenFile` on click (it fails under no-opener, other-opener and no-Workspace mutations).

packages/desktop/src/util.ts
hiyouga: Done in 4b07fd553: `previewHostFor` is replaced by `const PREVIEW_HOST = "127.0.0.1"` beside `LOOPBACK_HOSTS` (confirmed: `appOriginFor` and attach mode both build `http://localhost:<port>`, and `loopbackHostRoles` returns `app: "localhost"` unconditionally).

packages/desktop/src/util.ts
hiyouga: Done in 80c647331: the `mailto:` early return is deleted, a null origin now returns `deny` inside `classifyWindowOpen`, and the no-origin test keeps only its `file:` assertion (renamed to match).

PR #781, opened 2026-09-17
fix(web): turning company mode on offers the mode, it does not switch to it

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/183

## Summary

Turning company mode on, whether through the admin master switch (**System settings → Server → Company mode**) or the user's own switch (**System settings → General**), could put the shell straight into company mode. The sidebar became the company sidebar while the main area stayed on the page it was showing, usually the new-chat page instead of the new-organization landing. The user report asked for the opposite: enabling should only make **公司** appear in the top-left mode switch, and the user switches over from development mode by hand.

After this PR, turning either switch on only brings the mode switch back. The shell stays in development mode on the current page. Clicking **公司** opens the company landing as before (new organization when there are none).

## Root cause

The mode the shell stood in was a mask over the stored choice: `available ? workMode : "dev"`. Turning a switch off never touched the stored choice (the store, the `penguin.workMode` localStorage mirror, `ui_prefs.workMode`). So a `company` chosen before the mode became unavailable survived the off spell and took effect again the moment either switch came back on. No navigation happened, so the page underneath stayed where it was. A reload or another tab read the same stored `company`. A user who had never chosen company mode was not affected, which is why it only showed up for some people.

The org layout's route-entry assertion was not involved: it only runs on `/org` routes, and those redirect to `/chat` while the mode is unavailable.

Reproduced on `main` in a browser (Playwright against a local server + Vite, scratch data root, admin on the new-chat page):

| Step | Mode switch | Sidebar | Path | Stored choice (mirror / prefs) |
| Server switch on, never chose company | dev | development | `/chat/new` | dev / dev |
| Click 公司 | company | company | `/org` | company / company |
| Server switch off | hidden | development | `/chat/new` | company / company |
| **Server switch on again** | **company** | **company** | `/chat/new` | company / company |
| **Reload; other tab reloaded** | **company** | **company** | `/chat/new` | company / company |
| **Personal switch off → on** | **company** | **company** | `/chat/new` | company / company |
| **Loaded while server off, then on** | **company** | **company** | `/chat/new` | company / company |

## The rule

**While company mode is unavailable, the chosen work mode is development: in the store, the localStorage mirror and the preference.** The store applies it whenever it learns a switch's value:

- the server switch reaching the store (seeded at mount, then on every `/api/me` read; the admin's own flip refreshes it)
- the personal switch turned off
- preferences arriving while a switch is off

In each case a stored `company` is written back as `dev` through `setWorkMode`. `setWorkMode("company")` is refused while the mode is unavailable. `lastOrgKey` is kept, so switching to company mode later still lands on the last organization. Entering company mode is unchanged: the sidebar switch, the collapsed rail's toggle, or an `/org` route (the org layout still asserts company mode on route entry only).

Why clear on the off side instead of resetting at the off → on transition: a transition is visible only to a tab that is running when the switch flips. Three cases would still find the stale `company` with nothing to reset it: a reload after enabling, a browser that was closed during the flip, and a switch turned on elsewhere (another admin, the API). The transition render would also show the company shell for a frame before an effect could reset it. The level rule needs no memory of the previous availability, is idempotent, and runs on every load while the mode is unavailable. The effective-mode mask is kept for the first render after such a load, before the store has settled.

One limit remains, and it predates this PR: nothing tells other open tabs that the master switch changed. A tab that has not re-read `/api/me` since the switch went off still believes company mode is available. If the user deliberately switches into company mode in that tab, it records `company` until its next load settles it again.

## `login()` adopts the user together with `/api/me`

`AuthProvider.login` set the user before re-reading `/api/me`. The shell mounts as soon as there is a user, so it briefly mounted on the pre-login flags (`companyMode: false` on a fresh page). Under the new rule the store would read that stale "off" as unavailable and write a stored `company` back to `dev` on every sign-in through the login page. Checked in the browser: with the store change but the old `login()`, signing in lost the company choice. With both changes it is kept.

`login()` now sets the user in the same update as the `/api/me` answer. If that read fails, the user is adopted alone, as before — except on a 401. That is the one answer meaning the session cookie never took (blocked cookies, a cross-site context, a proxy stripping `Set-Cookie`); `api/client.ts` has already cleared the user over it, and adopting it back in the same continuation would navigate into a shell with no session behind it that bounces to `/login` on its next request. The login page stays put instead. On the remaining failure paths company mode reads as off for the session, as it already did, and a stored company choice is settled to development.

## Stored data

There is no format change and no compatibility code. A user whose stored `workMode` is `company` while company mode is unavailable to them gets it rewritten to `dev` on their next load, which is the fix itself. A user whose app never runs while the mode is off keeps their `company` choice. From their point of view nothing was turned on.

## Review round (2883ced19)

Three low findings, all taken.

- **`state/auth.tsx`** — the post-login catch adopted the user on a 401 as well, undoing the client's 401 handler in the same continuation. It is guarded now, and the decision is the exported `loginSessionSurvives`: this package's vitest is `environment: "node"` with no jsdom, so the Provider cannot be mounted and the predicate is the seam the `wechatScanStep` / `updateCheckOutcome` idiom already uses. New test `packages/web/test/login-session.test.ts`. The changelog pair's login bullet was reworded to match.
- **`state/company.tsx`, `applyPrefs`** — it wrote the `penguin.workMode` mirror with the stored mode and only then let `settleWorkMode` rewrite it, so a tab closed in that window left the next load a choice this one had already refused. The mirror is written once now, after the settle, from `get().workMode`, and the doc comment says so. `company-work-mode.test.ts` records every write to `WORK_MODE_KEY` through the stubbed `localStorage` and asserts the unsettled value is never among them.
- **`state/company.tsx`, `prefsLoaded`** — write-only across the repo; the field, its initializer and both writers are gone.

`origin/main` (`078a7aa83`) is merged in, no conflicts.

## Verification

- `pnpm --filter @prismshadow/penguin-web typecheck`: passes.
- `pnpm format` + `pnpm format:check`, `pnpm lint`, `git diff --check`: clean.
- vitest on vps3:
  - Company and state files (`vitest run company org- shell-org install-scope`): 13 files, 138 tests passed. This includes the new `test/company-work-mode.test.ts` (9 tests).
  - Whole web package (`pnpm --filter @prismshadow/penguin-web test`): 177 files, 2341 tests passed.
  - The new test file run against `main`'s `state/company.tsx`: all 9 fail. Most fail because the new actions are missing. "Entering company mode" fails on the old behaviour itself: `setWorkMode("company")` was accepted with the server switch off.
- Browser check (Playwright, local server + Vite on a scratch data root, same script as the table above), on this branch: after every step (server switch on again, reload, other tab reloaded, personal switch off → on and reload, loaded while off then enabled) the mode switch shows **dev**, the development sidebar is shown, the page stays on `/chat/new`, and the stored choice is `dev`. Clicking 公司 still opens `/org` with the landing, navigating to `/org` still enters company mode, and 开发 goes back. Signing in through the login page keeps a stored `company` choice. Reverting only `state/auth.tsx` loses it.
- Review round, on the tree with `origin/main` merged in (vps3, `2883ced19`):
  - `pnpm --filter @prismshadow/penguin-web exec vitest run test/company-work-mode.test.ts test/company-store.test.ts test/company-nav.test.ts test/company-beta.test.ts test/company-click-targets.test.ts test/company-mode-write.test.ts test/login-session.test.ts`: 7 files, 63 tests passed.
  - `pnpm typecheck` (25 projects): passes. `pnpm lint` (oxlint, 1458 files): 0 warnings, 0 errors. `pnpm format:check`: clean. `git diff --check`: clean.
  - Not re-run in the review round: the browser check and the whole web package, neither of which the three fixes reach beyond the suites above.
- Not run: the full `pnpm test`, and the Playwright `e2e/` suite. `e2e/company.spec.mjs` enables the switch through the API before signing in, which this change leaves on the same path.


packages/web/src/state/company.tsx
```diff
@@ -245,13 +298,45 @@ export function createCompanyStore() {
...
+     * The stored preferences have arrived, and they win over the localStorage mirrors: the
+     * user's switch, the mode and the organization last opened. A company choice among them
+     * while company mode is unavailable is settled at once rather than adopted.
+     */
+    applyPrefs: (prefs) => {
+      const patch: Partial<CompanyStoreState> = { prefsLoaded: true };
+      if (prefs.companyMode === false) patch.personalEnabled = false;
+      if (prefs.workMode === "company" || prefs.workMode === "dev") {
+        patch.workMode = prefs.workMode;
+        storeWorkMode(prefs.workMode);
```
hiyouga: **[simplify]** `applyPrefs` adopts the choice it is about to undo. The doc comment three lines up says a company choice is "settled at once rather than adopted", but the code adopts it — `patch.workMode = "company"` plus this `storeWorkMode("company")` — and only then lets `settleWorkMode` rewrite the store, the mirror and the server copy.

Scenario: a load with the master switch off and `ui_prefs.workMode === "company"`. The mount effect has already fired `putPrefs({ workMode: "dev" })`; the `getPrefs` from the next effect is in flight at the same time, so whenever the GET is answered before that PUT lands, the response still carries `"company"`. This line then writes `penguin.workMode` back to `"company"` and `settleWorkMode` immediately writes it to `"dev"` again with a second `putPrefs`. Close the tab in that window and the mirror is left on `"company"` for the next load to settle all over again.

Fix: drop the eager mirror write here and write the mirror once after `settleWorkMode(get())`, from `get().workMode` — the settled value. `setWorkMode` already owns the mirror and the preference for the settled case, so nothing is undone and the comment becomes true.

packages/web/src/state/auth.tsx
```diff
@@ -139,7 +143,8 @@ export function AuthProvider({ children }: { children: ReactNode }) {
       setUploadLimits(me.uploadLimits);
       setCompanyMode(me.companyMode);
     } catch {
-      // Login itself succeeded; keep the optimistic default.
+      // Login itself succeeded; adopt the user and keep the optimistic defaults.
+      setUser(res.user);
```
hiyouga: **[bug]** This catch adopts the user even when `/api/me` failed with 401, which is the one failure that means there is no usable session. `api/client.ts:106` fires `onUnauthorized()` for any 401 outside `/api/auth/`, and the handler registered at line 92 sets the user to `null`; this line then puts it straight back in the same continuation, so `setUser(null)` loses.

Scenario: `POST /api/auth/login` succeeds but the session cookie is not stored or not sent back — blocked cookies, a cross-site context, a proxy stripping `Set-Cookie`. Before this PR the ordering was the other way round and the user stayed on the login page. Now `login()` resolves, `login.tsx` navigates to `/chat`, the shell mounts on the pre-login flags, `settleWorkMode` writes `penguin.workMode` back to `"dev"` locally (the matching `putPrefs` 401s, so the server copy survives), and the first other request bounces back to `/login`. Same destination, via a flash of a dead shell.

Fix, one line, `ApiError` is already imported at line 11:

```ts
} catch (e) {
  // Login itself succeeded; adopt the user and keep the optimistic defaults. A 401 here
  // says the session cookie never took, so there is no user to adopt.
  if (!(e instanceof ApiError && e.status === 401)) setUser(res.user);
}
```

packages/web/src/state/company.tsx
```diff
@@ -245,13 +298,45 @@ export function createCompanyStore() {
...
       void api.putPrefs({ companyMode: enabled }).catch(() => undefined);
     },
 
+    /**
+     * The stored preferences have arrived, and they win over the localStorage mirrors: the
+     * user's switch, the mode and the organization last opened. A company choice among them
+     * while company mode is unavailable is settled at once rather than adopted.
+     */
+    applyPrefs: (prefs) => {
+      const patch: Partial<CompanyStoreState> = { prefsLoaded: true };
```
hiyouga: **[simplify]** `prefsLoaded` has no reader. `git grep -n prefsLoaded` over the whole repo returns four hits and all four are in this file: the interface field, the initializer, this write and the one in the provider's `.catch`. Nothing renders on it, and the field's own comment ("before that the mirrors below stand in") describes a gate that does not exist.

Pre-existing — I raise it because this PR moves one of the two writers into `applyPrefs`. If nothing is going to read it, the field, the interface entry and both writers go together; otherwise it is worth a line saying who is expected to read it.

hiyouga (commented):
Independent review. I read the diff against today's `origin/main` (`078a7aa83`), the full head files, the callers on main, and design PR penguin-harness-design#183. No commands were run.

**The rule holds.** I traced every path that can reach `setWorkMode`, and every one of them is gated on availability, so a switch coming on cannot enter the mode:

- Enabling changes no route. Nothing navigates on `company.available` going true — the two `if (available) setWorkMode("company")` effects live in `OrgIndexRedirect` and `OrgLayout`, which only mount on `/org*`. The sidebar's `Segmented` (`sidebar.tsx:1744`) and the collapsed rail's button (`app-layout.tsx:202`) are render-gated on `company.available`, so the new refusal in `setWorkMode` can never leave a rendered control dead; the shell stays where it was and only gains the switch.
- The seed closes the reload case. `createCompanyStore({ serverEnabled: companyMode })` is safe because `RequireAuth` renders `null` until `user !== undefined`, and `AuthProvider`'s mount fetch sets `user` and `companyMode` in one batched continuation — so `CompanyProvider`'s first render already has the real master switch and the store never settles a choice against a not-yet-known `true`. The one path where it could (the shell mounting on pre-login flags) is exactly what the `login()` change removes.
- Disabling from company mode is not a dead end. `OrgLayout` returns `<Navigate to="/chat" replace/>` once `available` goes false, and `lastOrgKey` survives, so re-entering lands on the same organization. On `/chat/:deskSessionId` the conversation stays open without a sidebar row, but that is main's behaviour too (#751) and not introduced here.
- Concurrent `putPrefs` writes are safe: `PUT /api/me/prefs` (`packages/server/src/http/routes/me.ts:230-249`) does its read-merge-write with no `await` between the read and the write, so the two writes `setPersonalEnabled(false)` now fires back to back cannot lose each other's field.
- Against today's main: `git merge-tree --write-tree origin/main origin/fix/company-enable-stays-in-dev` succeeds, and the merged `state/company.tsx` carries both sides — main's `setDeskMessagingChannel`/`resync` (#751) and this PR's `setServerEnabled`/`applyPrefs`/`effectiveWorkMode`. #755, #782–#784, #785 touch no file here; #791 is still open and touches `features/chat` and the server only. No conflicts.
- `e2e/company.spec.mjs` is unaffected: it enables the switch over the API before signing in and then clicks 公司 explicitly. `company-store.test.ts`'s 18 `createCompanyStore()` calls never touch `setWorkMode`, so the new `serverEnabled: false` default does not silently refuse anything there. Changelog pair is section-for-section and the metadata block is in order; design #183 states the same rule the code implements.

Three findings, all low, none of them affects the behaviour the PR is for. Detail inline.

Verdict: merge as is

packages/web/src/state/auth.tsx
hiyouga: Taken, in 2883ced19.

The catch now guards the adoption, and the predicate is exported so the decision can be asserted
without a DOM (`loginSessionSurvives` in `state/auth.tsx`):

```ts
} catch (e) {
  // Login itself succeeded; adopt the user and keep the optimistic defaults — unless the
  // read came back 401, which says the session cookie never took. ...
  if (loginSessionSurvives(e)) setUser(res.user);
}
```

Test: `packages/web/test/login-session.test.ts` — a 401 leaves no user to adopt; a network error,
a 500, a 403 and a non-`ApiError` throw all still adopt. This package's vitest is `environment:
"node"` with no jsdom and no testing-library, so the Provider cannot be mounted; the predicate is
the seam the `wechatScanStep` / `updateCheckOutcome` idiom already uses for a decision that lives
inside an effect.

The changelog pair's login bullet was reworded to match (a 401 now leaves the login page standing).

packages/web/src/state/company.tsx
hiyouga: Taken, in 2883ced19.

`applyPrefs` no longer writes the mirror eagerly; it writes it once after the settle, from the
settled mode:

```ts
if (prefs.workMode === "company" || prefs.workMode === "dev") patch.workMode = prefs.workMode;
...
set(patch);
settleWorkMode(get());
storeWorkMode(get().workMode);
```

The doc comment now says what happens — the choice is settled back to development before the
mirror is written, so a tab closed in between cannot leave the next load a choice this one has
already refused.

Test: `company-work-mode.test.ts` now records every write to `WORK_MODE_KEY` through the stubbed
`localStorage`, since the end state alone does not catch a value the mirror only passes through.
Both settling cases assert `modeWrites` never contains `"company"`, and the available case asserts
a single `["company"]` write.

packages/web/src/state/company.tsx
hiyouga: Taken, in 2883ced19. Confirmed the same four hits and no reader, so the field is gone rather than
documented: the interface entry, the initializer, the `applyPrefs` write and the provider's
`.catch` write. That `.catch` is now `.catch(() => undefined)`, the idiom the rest of the file
uses for a best-effort write, with the comment kept.

PR #782, opened 2026-09-17
feat(web): the process list clears every exited process in one click, and a long command reads whole on hover

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/185

## Summary

User request (verbatim): "会话右上角的后台进程列表，加一个一键清除「已退出」进程的按钮。补充一下，a1 后台进程hover要显示全部的，现在缩略了看不到"

The list in question is **会话进程** in the conversation's details card (the stats chips at the toolbar's far right). It is the only place that renders these rows: nothing in the dock or the subagents panel reuses them.

### Clear exited

- While at least one listed process has exited, the list heading shows a text action, 「清除已退出」 / "Clear exited". It is styled like #783's 「打开记忆列表」: `text-xs`, gray, darker on hover, no background, no border, and the same text size as the heading beside it, so the row keeps its height when the action appears. It is hidden when nothing has exited, like the section itself when there are no processes. Like a row's Remove button, its native `title` notes what leaves with the rows: 「清除所有已退出的进程——它们已捕获的输出也会一并丢弃」 / "Clear every exited process — the output captured from them is discarded too".
- **Server-side removal, through the existing path.** The server already keeps exited entries, and a single exited row is removed today with `DELETE /api/sessions/:sessionId/processes/:processId` (issue [#312](https://github.com/Prism-Shadow/penguin-harness/issues/312)). The new action sends that same request for every exited row (`Promise.allSettled`). No new route, no server change.
  - Running processes are never touched: only `running: false` rows are sent, and the server refuses to remove a live entry (409) anyway.
  - 404/409 still count as "the list was stale" and stay quiet. Any other failure produces at most one toast for the batch (`reportableProcessFailure`). The list refreshes afterwards, as it does for Stop / Remove.
  - No confirmation, same as the per-row Remove. The hint says what leaves with the rows.
- `runProcessAction` now takes a list of ids (a row's Stop / Remove is a batch of one), and `procBusy` holds the ids in flight. While the batch runs, every exited row's Remove shows the loading label and all process buttons are disabled.

### Full command on hover

- The command line was `truncate` with a native `title`. It now uses `Truncated` with a new `codeTooltip` option: the app's styled `Tooltip`, below the line, opened only while the command is actually truncated (the same measured rule that decides `Truncated`'s native `title`). The native `title` is gone, so there are never two tooltips.
- `Tooltip` gained a `code` content kind: monospace, `whitespace-pre-wrap` + `wrap-anywhere`, up to 24rem wide.
- `Tooltip` also caps **every** panel's width at the room left on screen on the side it grows toward (`min(cap, room)`). Before this change a left-anchored panel on a phone could run off the right edge; a 24rem code panel would have. For the existing labels nothing changes unless that room is below 16rem. The geometry functions take the viewport as a parameter now, so they are unit-tested.

## Decisions worth a look

- **Text, not an icon, for the clear action (user's decision).** The first version used a flat icon button with a tooltip, as the brief asked. After review the user chose text 「清除已退出」 in #783's style. The icon (`CLEAR_LIST_ICON`) and its Tooltip wrapper are gone; the Tooltip width clamp stays.
- The e2e specs waited for a chip titled `1 个运行中的服务`, a string that no longer exists anywhere in `packages/web/src`. The chip now reads `1 个后台任务`, so those tests failed on main regardless of this change. Both `processes.spec.mjs` and (after review) `background-notify.spec.mjs` are updated here.

## Verification

All on vps3 via `vps3-vitest.sh` (core/hmr/server built, web rebuilt for Playwright).

Head `5d99440` (review follow-ups: the `background-notify` chip title, and the unreachable `processIds.length === 0` guard):

- `pnpm --filter @prismshadow/penguin-web test`: 180 files / 2375 tests passed
- `pnpm format:check`: clean
- Playwright `e2e/processes.spec.mjs` + `e2e/background-notify.spec.mjs`: **4/4 passed**

Head `9b525cc` (text action): the same chain passed (180 files / 2375 tests, lint and format clean, `processes.spec` 3/3).

First version (icon button): the same chain passed (180 files / 2375 tests, lint and format clean, processes.spec 3/3; before the chip-title fix, test 1 failed on its stale title). Screenshots at 1280×800 and 390×800 showed the command tooltip wrapping and staying on screen on a phone.


packages/web/src/features/chat/chat-page.tsx
```diff
@@ -965,29 +967,31 @@ export function ChatPage() {
...
    */
   const runProcessAction = useCallback(
     async (
-      processId: string,
+      processIds: readonly string[],
       request: (sessionId: string, processId: string) => Promise<void>,
       staleStatuses: readonly number[],
     ) => {
-      if (!selected || procBusy !== null) return;
+      if (!selected || procBusy !== null || processIds.length === 0) return;
```
hiyouga: **[over-defensive]** `processIds.length === 0` is unreachable. The clear action renders only inside `{exitedIds.length > 0 && (...)}` in the same commit whose closure the click runs, and `onKillProcess` / `onRemoveProcess` always pass exactly one id. Even if an empty array did arrive, `Promise.allSettled([])` resolves at once, `reportableProcessFailure([])` returns null and the `finally` refresh runs — the guard buys nothing it does not already have. Drop `|| processIds.length === 0`.

hiyouga (commented):
Independent review at head `9b525cc`, read against `origin/main` and design PR `penguin-harness-design#185`. Nothing built or run here.

## Checked, and clean

- **The batch stays consistent.** `exitedProcessIds` judges "exited" as `!p.running`, the same test the row render uses to choose Remove over Stop, so the action's set and the rows' labels cannot disagree. A partial failure leaves the successful removals in place, reports one toast, and the `finally` refresh — still gated on `selectedSessionIdRef` — is the source of truth. A process that exits between the poll and the click is simply not in the batch; it comes back as an exited row after the refresh, with the action visible again.
- **It cannot fire twice.** The button is `disabled={procBusy !== null}` and `runProcessAction` re-reads `procBusy` (it is in the `useCallback` deps), so the second click of a double-click is dropped twice over. A click landing after the batch resolves but before the refresh re-sends DELETEs that answer 404 — swallowed, which is the intended reading.
- **The new viewport clamp does not narrow any existing tooltip.** Every `placement="right"` caller is the left icon rail (`packages/web/src/components/layout/app-layout.tsx`), where `room = viewportWidth - (rect.right + 8) - 8` is nearly the whole viewport. Every `placement="bottom"` caller is the Files-panel toolbar (`packages/web/src/features/chat/workspace-browser.tsx`); the left-anchored branch is only reachable when `rect.left <= viewportWidth / 2` (so `room >= viewportWidth/2 - 8`), and the right-anchored branch gives `room = rect.right - 8`. Both stay far above 16rem even at 390px wide. `max-w-64` and `contentMaxWidth.label` are the same 16rem, so a label's wrapping is byte-identical to before.
- **The changed Playwright expectation is right.** `packages/web/src/features/chat/chat-page.tsx:1958` renders `title={S.chat.backgroundTasks(backgroundCount)}`, and zh is `(n) => \`${n} 个后台任务\``. `1 个运行中的服务` exists nowhere in `packages/web/src`; the string has read this way since #585.
- Truncation survives the `Truncated` swap: the `Tooltip` wrapper is `display:flex` and therefore block-level, and `truncate`'s `overflow:hidden` zeroes the flex item's automatic minimum size, so the line still shrinks to the column and `scrollWidth > clientWidth` still measures what it did.
- Design #185's clause matches the code clause for clause (text action shown only while something has exited, same per-entry route, no confirmation, hint naming the discarded output, hover panel raised only while the command is truncated). i18n parity and the bilingual changelog pair are in order.

## Findings

**[test] `packages/web/e2e/background-notify.spec.mjs:62` still carries the dead chip title this PR retires.**

```js
await expect(detailsBtn.locator('span[title="1 个运行中的服务"]')).toBeVisible({
```

That is the same selector, failing for the same reason, in the other spec that waits on the same chip. The PR's own justification for touching `processes.spec.mjs` ("a string that no longer exists anywhere in `packages/web/src`") applies verbatim here, so leaving it means the web Playwright suite still has one guaranteed failure that this PR already knows how to fix. Fix: the same substitution to `1 个后台任务`.

Verdict: merge as is

packages/web/src/features/chat/chat-page.tsx
hiyouga: Dropped in 5d99440 — `runProcessAction` now guards on `!selected || procBusy !== null` alone.

hiyouga:
Review body finding (the one without an inline thread) is fixed too: `packages/web/e2e/background-notify.spec.mjs:62` now waits on `span[title="1 个后台任务"]`, the same substitution as in `processes.spec.mjs` (5d99440). Verified on vps3: `background-notify.spec.mjs` + `processes.spec.mjs` = 4/4 passed.

PR #783, opened 2026-09-17
fix(web): the memory-change card opens the memory list with a text action, not a second brain icon

hiyouga:
Design: none. `06-PROTOTYPE` describes the memory-change card only as memories listed under the file summary, where a click locates the entry in the Memory panel. No clause describes the card's header button or names its icon, so there is no spec text to change.

## Summary

User request (verbatim): "现在会话框记忆更新以后会出现两个记忆图标，用户也不清楚是什么意思，右边的打开记忆列表就不要用图标了吧，文字更清晰。"

`packages/web/src/features/chat/memory-changes-card.tsx`: the header showed the card's own brain glyph on the left (`MEMORY_ICON`), and on the right an icon-only button with the same glyph that opens the Memory panel on its list.

- The right-hand button is now text: 「打开记忆列表」 / "Open memory list". These are new dictionary keys `chat.memoryOpenList` in `strings.ts` and `strings-en.ts`. `memoryViewTitle`, which names the panel, is left alone.
- Styled like the rows' right-aligned 「查看内容」 / "View content" hint: `text-xs`, gray-400 (gray-500 in dark), darker on hover, no background. The visible words are its accessible name, so the old `title` and `sr-only` label are gone and no tooltip repeats the text.
- The left glyph stays, so the card carries exactly one memory icon.
- The header's height now matches the file-summary card's header. The old 24px square button made it slightly taller.
- The file's header comment and the prop doc no longer mention "the header's brain button".
- No existing unit or e2e test found the button by its title or icon. The new `test/memory-changes-card.test.ts` (static markup) pins: the action is a `<button>` whose text is the label, with no `title` and no `<svg>`; the brain path is drawn exactly once, with or without the action wired; and the label follows the UI language. After review, the action is matched whole (the card's first `<button>…</button>`) and the match is asserted non-null, so a glyph creeping back inside it fails the test instead of emptying the string the assertions read.

## Verification

All on vps3 via `vps3-vitest.sh` (fresh install, core/hmr/server built):

- `pnpm --filter @prismshadow/penguin-web test`: 182 files / 2395 tests passed, including the 4 new card tests
- `pnpm format:check`: clean
- Negative check of the repaired test (vps3, remote copy only): putting a `GlyphIcon` back inside the action fails 3 of the 4 tests, including `not to contain '<svg'` against the real button markup.

Not checked in a browser: the e2e mock LLM has no branch that writes a memory file, so no spec can raise this card. The change is a class and text swap, covered by the markup test above.


packages/web/test/memory-changes-card.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const html = render(() => {});
+    expect(html).toMatch(/<button type="button"[^>]*>打开记忆列表<\/button>/);
+    expect(memoryMarks(html)).toBe(1);
+  });
+
+  it("names the action in words the button shows, with no tooltip repeating them", () => {
+    const html = render(() => {});
+    const action = html.match(/<button type="button"[^>]*>打开记忆列表<\/button>/)?.[0] ?? "";
+    expect(action).not.toContain("title=");
+    expect(action).not.toContain("<svg");
```
hiyouga, line 55: **[test]** This assertion can never fail. `action` comes from a regex whose `[^>]*` cannot cross the opening tag's `>` and whose body is pinned to the label, so a button that regressed to carrying an `<svg>` would not match at all — `action` would be the `?? ""` fallback, and `not.toContain("<svg")` would pass on the empty string. Same for line 49 in that case. Assert the match exists first (`expect(match).not.toBeNull()`), then check the attributes on `match[0]`.

hiyouga (commented):
Independent review at head `0e6bbac`, read against `origin/main`. Nothing built or run here.

## Checked, and clean

- **Both dictionaries.** `memoryOpenList` is added at the same position in `strings.ts` and `strings-en.ts`, with the same doc comment and the same shape, so the `Strings` type covers it.
- **The style matches the rows' hint.** The row's "View content" span is `shrink-0 text-xs text-gray-400 transition-colors duration-150 group-hover:text-gray-600 dark:text-gray-500 dark:group-hover:text-gray-300`; the new button is the same with `hover:` in place of `group-hover:` (correct — it is its own button, not a child of the row button) plus `cursor-pointer`. Note that this also changes the dark resting ink from gray-400 to gray-500, which is what makes it match the rows rather than stand a shade brighter than them.
- **Nothing else refers to the old icon button.** `memoryViewTitle` is still live (`packages/web/src/features/dock/panel-meta.tsx:28` names the panel with it), so dropping it here leaves no dead key. `packages/web/test/icon-scale.test.ts:74` ("draws the memory brain in exactly one place") counts occurrences of the path literal in source, not renders of `MEMORY_ICON`, so removing one render does not touch it. The file header comment and the `onOpenPanel` prop doc were both updated.
- The visible words are now the accessible name, so dropping `title` and the `sr-only` span is right rather than a loss.
- Bilingual changelog pair present, metadata block in the fixed order, no forbidden sections.

## Findings

**[test] `packages/web/test/memory-changes-card.test.ts:48-50` — the two assertions in "names the action in words the button shows" cannot fail in the case they exist for.**

```js
const action = html.match(/<button type="button"[^>]*>打开记忆列表<\/button>/)?.[0] ?? "";
expect(action).not.toContain("title=");
expect(action).not.toContain("<svg");
```

`[^>]*` cannot cross the opening tag's `>`, and the pattern then demands the button's content be exactly the label. So the moment the button regresses to an icon button — a `<svg>` child, or the label replaced by a glyph — the match fails, `action` falls back to `""`, and both `not.toContain` assertions pass on the empty string. The one thing they could still catch (a `title=` attribute alongside the visible text) is real, but the `<svg>` assertion is dead: if an `<svg>` were inside the button the regex would not have matched it in the first place.

The regression is still caught by the first test (`expect(html).toMatch(...)` plus `memoryMarks(html) === 1`), so this is a hardening fix, not a hole in the suite. Fix: assert the match first, e.g. `const match = html.match(...); expect(match).not.toBeNull();` and then run the attribute checks against `match[0]`.

Verdict: merge as is

packages/web/test/memory-changes-card.test.ts
hiyouga: Fixed in 5f8c102. The header action is now matched whole (`/<button[\s\S]*?<\/button>/`, the card's first button), the match is asserted non-null, and the label is checked on it, so a glyph inside the button fails the test instead of emptying the string it is checked against. Verified on vps3 by putting a `GlyphIcon` back in the remote copy: 3 of the 4 tests fail, including `not to contain '<svg'` against the real button markup.

PR #784, opened 2026-09-17
feat(web): selected conversation text gets a right-click menu — Copy, and Add to conversation

hiyouga:
Design: https://github.com/Prism-Shadow/penguin-harness-design/pull/186

## Summary

User request (verbatim): "然后对话框里面的文字也支持选中后右键菜单，一个是复制，一个是直接添加到对话（输入框）。"

### Behaviour

- **When the app's menu opens.** A non-blank selection must lie wholly inside the message stream (user messages, replies and tool output all count). While it does, a secondary click **anywhere in the stream** (not only on the selected text), Shift+F10 (when focus is inside the stream) or a keyboard-synthesized `contextmenu` opens the app's own menu instead of the browser's. A pointer-opened menu hangs at the pointer; a keyboard-opened one hangs at the end of the selection's last line, and Escape hands focus back to the element inside the stream that held it.
- **Items, in order:**
  1. 「复制」 / "Copy" writes the selection with `writeClipboard` and confirms with `toastSuccess(S.common.copied)`. That is the menu-row convention the Files panel's copy-path row already uses, because a row cannot carry the copy button's at-the-control feedback.
  2. 「添加到对话」 / "Add to conversation" (reuses `S.files.addToChat`) stages the excerpt through `ComposerControl.addReference`, the same way the Files panel's 「将选中内容添加到对话」 stages a quoted range. Nothing already typed is touched and nothing is sent. The chip label is the excerpt's start (whitespace folded, cut at 32 code points with "…", then CSS-ellipsized to the chip), and its tooltip is the whole excerpt. The message carries the excerpt as a Markdown blockquote ahead of the typed text.
- The selection stays highlighted after either action. It is captured (text + cloned range) at the gesture and restored with `restoreSelection`, the Files panel's helper, now shared.
- **Declined, so the native menu stays:** no selection, a blank one, a selection that runs outside the stream (into the composer, say), a selection built from several ranges (Firefox's Ctrl+drag — its text is every range's, so one range in the composer would ride into the clipboard and the excerpt), and a right-click on an editable field. A touch or pen press-and-hold is also declined, since mobile OSes have their own selection menu: pointer type comes from the `contextmenu` PointerEvent where the browser provides it, otherwise from the stream's last `pointerdown`. `preventDefault` runs only once the rules have accepted the gesture, inside the stream's own handler. No global listeners are added (the context-menu suppression-scope test still passes).
- **Where it applies.** `MessageStream` takes a required `onAddExcerpt`. The chat page passes its composer. The Agents panel's child stream passes that child's composer (new `controlRef` on `SubagentComposer`), so "Add to conversation" stages into the conversation the text was selected in. There is no composer-less stream in this app, so the menu's rows are the constant `SELECTION_MENU_ITEMS`.
- **Auto-follow is held while the menu is open.** During a live reply, streaming updates no longer snap the view to the bottom, so the menu stays anchored and usable. Closing the menu resumes following, and a view that was pinned to the bottom catches up at once. The user's own scrolling still closes the menu through the Dropdown's scroll dismissal, which is unchanged.
- The desktop app gets a copy menu for conversation text this way too. Electron shows no native context menu, as the shell registers none.

### Structure

- `lib/selection-menu.ts` holds the pure rules: `opensSelectionMenu` (which reads `rangeCount` and `firstRangeInStream`, so the multi-range case is unit-tested), `SELECTION_MENU_ITEMS`, `selectionEndAnchor`, `normalizeExcerpt` / `excerptLabel` / `excerptBlockquote` / `excerptReference`.
- `features/chat/stream-selection-menu.tsx` is the DOM half. `useStreamSelectionMenu` reuses `useRowContextMenu` for anchor state and dismissal, without spreading its press-and-hold handlers, plus the `Dropdown` in `anchorRect` mode as `workspace-file-menu.tsx` does. `SelectionMenuRows` uses the overflow-menu row styling.
- `ComposerReference` became `WorkspaceReference | ExcerptReference`, a new path-less `excerpt` kind. The composer chip moved out of `chat-input.tsx` into `reference-chip.tsx` (unchanged for existing kinds, quote glyph for an excerpt) so it can be rendered in a node test.
- `restoreSelection` moved from `workspace-browser.tsx` to `components/ui/text-selection.ts`, unchanged.
- New keys `common.copy`: 「复制」 / "Copy".
- The follow hold: `stream-follow.ts`'s model gained `hold(held)` and `snaps` (`stick && !held`). A hold defers the snap and never changes the intent (`stick`), so a wheel-up during the hold still exits follow and a scroll back to the bottom still resumes it. `MessageStream` sets the hold from `useStreamSelectionMenu`'s new `open` in a layout effect declared before the commit snap, snaps on release if `snaps`, and gates both snap sites (commit and ResizeObserver) on `snaps`. Prepend anchoring now applies whenever the snap does not own the position (`!snaps`), so a held view is anchored like any other.

## Decisions worth a look

- **Agents panel:** its child stream gets the same menu, and "Add to conversation" stages into the child's composer below it, not the main one. The brief's rule was "omit where there is no composer", and that panel has one.
- **Selection anywhere in the stream counts (user confirmed: kept as is).** While a selection lies inside the stream, a right-click anywhere in the stream shows the menu, including on a link elsewhere in the stream. Design #186's clause now states this exactly.
- **The keyboard path is best-effort without a document-level listener.** Shift+F10 is heard only while focus is inside the stream. The Menu key works where the browser dispatches its synthesized `contextmenu` into the stream.
- **Live replies (user's decision: hold auto-follow while the menu is open).** Previously the snap that follows a live reply scrolled the stream, and the primitive's scroll dismissal closed the menu on the next chunk. Now the menu holds the follow instead. One edge remains: a snap made in the same frame as the right-click can still deliver its scroll event after the menu has opened and close it. It is rare, and the retry in the e2e test exists for exactly that. Special-casing the Dropdown's scroll dismissal was ruled out.

## Verification

All on vps3 via `vps3-vitest.sh` (core/hmr/server built, web rebuilt for Playwright).

Head `c1d625c` (review follow-ups: the multi-range decline, the required `onAddExcerpt`, the un-exports):

- `pnpm --filter @prismshadow/penguin-web test`: 183 files / 2417 tests passed, including `selection-menu.test.ts` (17, with the new multi-range case), `stream-selection-menu.test.ts` (4), `stream-follow.test.ts` (17)
- `pnpm format:check`: clean
- Playwright `e2e/selection-menu.spec.mjs`: **2/2 passed**

Head `09a3269` (follow hold):

First version (before the hold): the same chain passed (183 files / 2413 tests, lint and format clean, selection-menu.spec 1/1), and screenshots of the menu and the excerpt chip were checked at 1280 and 390 wide.


packages/web/src/features/chat/stream-selection-menu.tsx
```diff
@@ -0,0 +1,246 @@
...
+    const host = menu.anchorOwner();
+    // Events raised in the portaled panel reach this handler through React's tree; only a
+    // gesture that really happened inside the stream counts.
+    if (host === null || !(target instanceof Node) || !host.contains(target)) return false;
+    const selection = window.getSelection();
+    const range =
+      selection !== null && selection.rangeCount > 0 && !selection.isCollapsed
+        ? selection.getRangeAt(0)
+        : null;
+    const selectedText = range !== null && selection !== null ? selection.toString() : "";
```
hiyouga, line 177: **[bug]** The text and the containment test come from different places when a selection has more than one range. `selectedText` is `selection.toString()`, which concatenates **every** range; `selectionInStream` on the next line is judged on `getRangeAt(0)` alone. Firefox is the reachable case (Ctrl+drag builds a multi-range selection): select a sentence in a reply, then Ctrl+drag a second range inside the composer, and range 0 is inside the stream, so the menu opens — Copy writes the composer's text too, and "Add to conversation" stages it into the excerpt. That is precisely what the rule (and design #186's 「选区非空且两端都落在消息流内」) says must be declined. `restoreSelection` then compounds it: it `removeAllRanges()` and puts back only the cloned range 0, so the other ranges vanish. Fix: decline outright unless `selection.rangeCount === 1`, or require every range to be inside `host` and build the text from those ranges.

packages/web/src/features/chat/reference-chip.tsx
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * A staged reference, split for display: its name, and the quoted lines as a `:from-to` suffix.
+ * The `file:line` form rather than a worded one — it is the shape every editor and stack trace
+ * already uses, it needs no translating, and a chip has no room for a sentence.
+ *
+ * Returned in two pieces because the chip draws them differently: a long name ellipsizes, and the
+ * line numbers must not go with it. They are the smaller half and the half the name does not
+ * already say. An excerpt has no path and no lines, so its name is the excerpt's own start.
+ */
+export function referenceParts(reference: ComposerReference): { name: string; lines: string } {
```
hiyouga: **[simplify]** `referenceParts` and `referenceTitle` are exported but consumed only by `ReferenceChip` in this same file — `chat-input.tsx` no longer references either (it imports `ReferenceChip` alone), and `test/stream-selection-menu.test.ts` imports only `ReferenceChip` and `excerptLabel`. They were module-private in `chat-input.tsx` before the move; drop both `export` keywords so the module's surface stays one component.

packages/web/src/features/chat/message-stream.tsx
```diff
@@ -216,8 +220,15 @@ export function MessageStream({
   outline?: ReactNode;
   /** Scroll-up backfill of older history windows; omitted = the whole transcript is loaded (no top affordance). */
   older?: OlderHistoryControls;
+  /**
+   * Stages text selected in the stream in this conversation's composer, as a chip (the
+   * selection menu's "Add to conversation"). Omitted where the view has no composer: the menu
+   * then offers Copy alone.
+   */
+  onAddExcerpt?: (reference: ComposerReference) => void;
```
hiyouga: See the "read-only stream" finding in the review body: this prop is optional, and its doc ("Omitted where the view has no composer") describes a caller that does not exist — both `MessageStream` sites pass it.

hiyouga (commented):
Independent review at head `09a3269`, read against `origin/main` and design PR `penguin-harness-design#186`. Nothing built or run here.

## Checked, and clean

- **The auto-follow hold cannot strand the view.** `follow` is `followRef.current ??= createStreamFollow()` inside `MessageStream`, and the hold is driven from `selectionMenu.open`, which is state in the *same* component — so an unmount (route change, the Agents panel's `key={active.sessionId}` remount) destroys the holder and the held object together, and there is nothing left to be stuck. The only way `menuOpen` stays true is the panel still being rendered, i.e. visibly open. I looked for a session switch that keeps `MessageStream` mounted and leaves the menu open: the chat route's stream has no `key`, so a switch does keep it mounted, but every switch I can reach goes through a pointer press (a sidebar row, a link, the composer), and the `Dropdown`'s `window` `mousedown` listener closes the menu before that navigation runs. The one programmatic navigation on that page, `syncHealedSessionId`, only fires inside a send, which already required a click in the composer. No path found.
- **The hold's shape is right.** `hold()` touches only `snaps`, never `stick`, so a wheel-up during the hold still exits follow and a scroll back to the bottom still resumes it — and because `scrolled()` is only reached from a real scroll event, content growing under a held view never rewrites the intent. Flipping the prepend anchoring from `!follow.stick` to `!follow.snaps` is behaviour-identical whenever nothing is held, and correct when something is (nothing snaps, so the anchor has to own the position). Declaring the hold effect above the commit-snap effect does order them as claimed. `jumpToLatest`'s two `stickToBottom` calls stay ungated, which is right — that button's job is to move the view — and the `stream-follow.test.ts` source pin does not match them, so it is precise even though its comment says "every snap".
- **Selection rules.** `selectionInStream` is `host.contains(range.commonAncestorContainer)`, which is exactly "both ends inside the stream": a common ancestor inside the host implies both endpoints are. A selection running from a reply into the composer therefore declines, and `preventDefault` is only reached after the rules take the gesture, so the native menu survives everywhere else. The `host.contains(target)` guard is load-bearing rather than defensive: `ZoomableImage` (`components/ui/image-zoom.tsx`) portals its overlay to `document.body` from inside a message item, so a right-click on a zoomed image really does reach this handler through React's tree, and that check is what keeps "Save image as…" alive there.
- **Shift+F10 and focus.** `isContextMenuKey` is heard on the stream container only, the keyboard path passes `pointerType: ""` explicitly (so it is never mistaken for a press-and-hold), and `contextMenuAnchor` routes it to `selectionEndAnchor` because `button: 0`. `focusBefore.current` is only filled when the active element is inside the host, which is exactly the keyboard case; for a mouse-opened menu it is null and `Dropdown`'s fallback also resolves to null (`button={null}`, panel portaled), so Escape leaves focus on `<body>` — no worse than not returning it, and the design only asks for the keyboard path.
- **What the chip carries.** `excerptReference` stages `{kind:"excerpt", excerpt, text: blockquote}`; `withReferences` puts `r.text` ahead of the typed draft joined by `\n\n`, and the subagent variant uses the same composition (`chat-input.tsx` builds it for both send and steer, and `draftHasContent` counts references, so an excerpt alone can be sent). The blockquote is correct Markdown: a blank line becomes a bare `>` so the quote does not end there, and an unterminated fence inside the quote cannot swallow the typed text, because the typed text is outside the blockquote's own lines.
- **The subagent path.** `childComposerRef` is read at call time through a stable `useCallback`, and both the child stream and `SubagentComposer` are keyed by `active.sessionId`, so the excerpt always lands in the composer belonging to the conversation the text was selected in.
- **Clipboard and toast.** `writeClipboard` is documented as best-effort and never throws; the unconditional `toastSuccess(S.common.copied)` after it is the established menu-row convention (`workspace-browser.tsx:1525` copy-path and `sidebar.tsx:2943` copy-id do exactly this, because a row closes out from under the at-the-control feedback). Nothing to change.
- Design #186's clause matches the implementation clause for clause. i18n parity for `common.copy` is in order, and the bilingual changelog pair has the metadata block in the fixed order with no forbidden sections.

## Findings

Two of the three are inline. The third:

**[simplify] The "read-only stream offers Copy alone" path has no caller, and the spec states it as shipped behaviour.** Both `MessageStream` call sites pass `onAddExcerpt` (`chat-page.tsx:2200`, `subagents-view.tsx:274`), so `selectionMenuItems(false)` is only ever reached from `test/selection-menu.test.ts`. Design #186 nonetheless says 「没有输入框的只读消息流只有「复制」」, which describes a surface that does not exist. Either drop the optionality (make `onAddExcerpt` required and `selectionMenuItems` unconditional) or name the stream that will need it, and match the clause to whichever it is.

Verdict: merge as is

packages/web/src/features/chat/stream-selection-menu.tsx
hiyouga: Fixed in c1d625c, taking the first of your two options. The rule now reads the range count: `SelectionMenuRequest` carries `rangeCount` and `firstRangeInStream`, and `opensSelectionMenu` declines unless `rangeCount === 1` and that range lies inside the host — so a Ctrl+drag selection reaching into the composer keeps the browser menu, and nothing it holds can reach the clipboard, the excerpt, or `restoreSelection`. `selection-menu.test.ts` gains the multi-range case (2 ranges and 0 ranges, first range inside). Design #186's clause now reads 「选区非空、为单一区段且两端都落在消息流内」 and names the Ctrl+drag case among what it declines.

packages/web/src/features/chat/reference-chip.tsx
hiyouga: Dropped both `export`s in c1d625c; the module exports `ReferenceChip` alone.

packages/web/src/features/chat/message-stream.tsx
hiyouga: Took the first option in c1d625c: `MessageStream.onAddExcerpt` is required, `useStreamSelectionMenu` and `SelectionMenuRows` take it non-optional, and `selectionMenuItems(hasComposer)` is now the constant `SELECTION_MENU_ITEMS`. Its test case, the changelog line and design #186's 「没有输入框的只读消息流只有「复制」」 sentence are gone with it. `test/message-stream.test.ts` passes the prop now.

PR #786, opened 2026-09-18
feat(model-catalog): the OpenCode Go group, after TokenDance and Penguin Go

hiyouga:
## Summary

Adds **OpenCode Go** (`opencode-go`) as a built-in model group with the 27 models [opencode.ai/docs/go](https://opencode.ai/docs/go/) lists (re-read 2026-09-18). The Models page's default group order becomes **TokenDance → Penguin Go → OpenCode Go → DeepSeek → the rest, unchanged**.

Design: Prism-Shadow/penguin-harness-design#191. Group order and the per-row protocol pins are in 06「模型配置页」 and 01 PRN-021.

### Where the order lives

- `MODEL_PROVIDERS` in `packages/core/src/state/model-catalog.ts` is the default display order. The web's `groupModelRows`, `allGroupKeys` and `orderModelsLikeLibrary` map over it.
- `opencode-go` sits between `penguin-go` and `deepseek`. The core `model-catalog` and web `model-grouping` tests pin the whole sequence.
- `MODEL_CATALOG`'s cluster order is only the order a new Project's `[[models]]` rows are written in. It is not the display order, so the new rows follow Penguin Go's there.
- A Project whose groups were dragged keeps its stored order, and the new group trails the groups that order names (`model-group-order.ts`).

### The group

- **Provider entry.** Label "OpenCode Go". Key page `https://opencode.ai/auth`, models page `https://opencode.ai/docs/go/`. `gatewayBaseUrl` is `https://opencode.ai/zen/go/v1`, so a hand-added model is prefilled with it and gets `openai-chat`. No group-level `clientType`, because the models use three protocols. Env pair `OPENAI_*`.
- **Per-row protocol**, from the docs page's endpoint table. Every row pins its client, since AgentHub routes by id substring and never looks at `base_url`:

  | Client | Rows | Base URL |
  | --- | --- | --- |
  | `openai-chat` | 16 | `https://opencode.ai/zen/go/v1` |
  | `openai-responses` | 4 | `https://opencode.ai/zen/go/v1` |
  | `ant-messages` | 7 | `https://opencode.ai/zen/go`, no `/v1`: the Anthropic SDK appends `/v1/messages` itself |

  The paths are not interchangeable. `/chat/completions` answers 401 "not supported for format oa-compat" for `grok-4.6` and 500 for `gpt-5.6-luna`. For `minimax-m3` it returns the `<think>` block inside the reply text. The Messages rows read `ANTHROPIC_*` when the key is blank.
- **Pricing.** Go is $10/month. Its usage limits are dollar amounts per model: a monthly allowance, at most 20% per 5 hours and 50% per week. The rows store the per-token rates a request draws from that allowance, so the cost center shows allowance used. This follows how Qwen Token Plan stores per-token list prices rather than a plan fee.
  - Where the page lists no cached-write price, `cache_write` carries the input price.
  - Four rows are tiered and store their base tier: `gpt-5.6-luna` (≤272K), `grok-4.6` (≤200K), `qwen3.7-plus` and `qwen3.6-plus` (≤256K).
  - The four DeepSeek rows publish Peak and Off-Peak prices on DeepSeek's own windows: peak is 01:00–04:00 and 06:00–10:00 UTC on weekdays, which is `DEEPSEEK_OFF_PEAK`, and Off-Peak is exactly half. They store the peak price and declare that schedule.
  - DeepSeek V4.1 Flash's "4x, ends Sep 20" offer raises its allowance, not its price, so no row has a `discount`.
- **Metadata.** Context windows and input modalities come from models.dev's `opencode-go` provider, the registry OpenCode maintains. The docs page publishes neither.
- **Glyph.** The pixel "G" OpenCode publishes as the group's 24×24 logo: one `currentColor` path, rendered and checked in light and dark.
- **Attribution.** Both base URLs are on `opencode.ai`, so every row already gets `x-opencode-session` (#612). A new catalog-invariant test pins this.
- **Docs.** `models.{en,zh}.md` gains a table row and a "The OpenCode Go group" section, and the DeepSeek schedule and GLM-5.3 Flash lists are updated. `configuration.{en,zh}.md` lists the group's env vars.

### Not added

- **`/models` extras.** The gateway's `/models` listing also answers ten older ids the docs page does not list and models.dev marks `deprecated`: `minimax-m2.5`, `kimi-k2.5`, `glm-5`, `deepseek-flash`, `qwen3.5-plus`, `mimo-v2-pro`, `mimo-v2-omni`, `hy3-preview`, `grok-4.5`, `omen-alpha`. `minimax-m2.5` still has a price row, but it is not in the lineup.
- **`union-alpha`** ("Union Alpha Free", limited time). It was listed on 2026-09-17. By 2026-09-18 it is gone from the docs page, `/models` and models.dev, and the gateway answers `401 Model union-alpha is not supported`.

## Models

Prices are USD per million tokens, in the Go page's order: input / output / cache read / cache write.

| id | protocol | context | max out | vision | price |
| --- | --- | ---: | ---: | :---: | --- |
| `deepseek-v4.1-flash` | chat | 1,000,000 | 384,000 | yes* | 0.30 / 1.20 / 0.006 / – (peak; off-peak ½) |
| `deepseek-v4-flash` | chat | 1,000,000 | 384,000 | no* | 0.30 / 1.20 / 0.006 / – (peak; off-peak ½) |
| `deepseek-v4-flash-vision-exp` | chat | 1,000,000 | 384,000 | yes | 0.30 / 1.20 / 0.006 / – (peak; off-peak ½) |
| `deepseek-v4-pro` | chat | 1,000,000 | 384,000 | no* | 1.32 / 3.96 / 0.044 / – (peak; off-peak ½) |
| `glm-5.3` | chat | 1,000,000 | 131,072 | no | 1.40 / 4.40 / 0.26 / – |
| `glm-5.3-flash` | chat | 1,000,000 | 131,072 | yes | 0.15 / 0.50 / 0.03 / – |
| `glm-5.2` | chat | 1,000,000 | 131,072 | no | 1.40 / 4.40 / 0.26 / – |
| `glm-5.1` | chat | 202,752 | 32,768 | no | 1.40 / 4.40 / 0.26 / – |
| `gpt-5.6-luna` | responses | 1,050,000 | 128,000 | yes | 0.20 / 1.20 / 0.02 / 0.25 (≤272K) |
| `grok-4.6` | responses | 500,000 | 500,000 | yes | 2.00 / 6.00 / 0.50 / – (≤200K) |
| `hy4-preview` | chat | 1,024,000 | 64,000 | no | 0.834 / 2.501 / 0.042 / – |
| `hy3` | chat | 192,000† | 128,000 | no | 0.14 / 0.58 / 0.035 / – |
| `kimi-k3` | chat | 1,048,576 | 131,072 | yes | 3.00 / 15.00 / 0.30 / – |
| `kimi-k2.7-code` | chat | 262,144 | 262,144 | yes | 0.95 / 4.00 / 0.19 / – |
| `kimi-k2.6` | chat | 262,144 | 65,536 | yes | 0.95 / 4.00 / 0.16 / – |
| `longcat-2.0` | chat | 1,000,000 | 131,072 | no | 0.30 / 1.20 / 0.006 / – |
| `mimo-v2.5` | chat | 1,000,000 | 128,000 | yes | 0.14 / 0.28 / 0.0028 / – |
| `mimo-v2.5-pro` | chat | 1,048,576 | 128,000 | no | 0.435 / 0.87 / 0.003625 / – |
| `minimax-m3` | messages | 1,000,000 | 131,072 | yes | 0.30 / 1.20 / 0.06 / – |
| `minimax-m2.7` | messages | 204,800 | 131,072 | no | 0.30 / 1.20 / 0.06 / 0.375 |
| `muse-spark-1.3-contributor` | responses | 1,048,576 | 131,072 | yes* | 0.10 / 0.20 / 0.002 / – |
| `muse-spark-1.2-contributor` | responses | 1,048,576 | 131,072 | yes* | 0.10 / 0.20 / 0.002 / – |
| `qwen3.8-flash` | messages | 1,000,000 | 131,072 | yes | 0.15 / 0.47 / 0.016 / 0.20 |
| `qwen3.8-max` | messages | 1,000,000 | 131,072 | yes | 2.00 / 6.00 / 0.25 / 2.50 |
| `qwen3.7-max` | messages | 1,000,000 | 65,536 | no | 2.50 / 7.50 / 0.50 / 3.125 |
| `qwen3.7-plus` | messages | 1,000,000 | 65,536 | yes | 0.40 / 1.60 / 0.04 / 0.50 (≤256K) |
| `qwen3.6-plus` | messages | 1,000,000 | 65,536 | yes | 0.50 / 3.00 / 0.05 / 0.625 (≤256K) |

\* Could not be reached with the test key (see below), so the flag rests on models.dev alone. The catalog has no max-output field; that column is for review only.

† models.dev lists a 256,000 context but caps Hy3's input at 192,000. The row stores the cap, because compaction derives from `contextWindow` (review finding 1).

## Live test

**Method.** Each model ran through core's `GenerativeModel`, the object a Session uses: AgentHub `AutoLLMClient` with the row's `client_type` and base URL, attribution headers and a Session id. Three checks:
- **Text:** a streaming one-word reply.
- **Tool:** a `get_weather` call, the tool result sent back, and a final answer.
- **Image:** a generated PNG of the digits "73". A model that cannot see the image cannot guess them.

**Runs.**
- 2026-09-17 from the dev box.
- 2026-09-18 from vps3, which is in mainland China, after the coordinator moved every live call there.

On both runs every request carried `x-opencode-session`, and the SDKs sent their default user agents.

| Result | Models |
| --- | --- |
| Text ✓, tool ✓, image read "73" ✓ | `deepseek-v4-flash-vision-exp`, `glm-5.3-flash`, `grok-4.6`, `kimi-k3`, `kimi-k2.7-code`, `kimi-k2.6`, `mimo-v2.5`, `minimax-m3`, `qwen3.8-flash`, `qwen3.8-max`, `qwen3.6-plus` |
| Text ✓, tool ✓, image correctly not read | Refused: `glm-5.3`, `hy4-preview`, `hy3`. Not seen: `glm-5.2`, `glm-5.1`, `longcat-2.0`, `minimax-m2.7`. Upstream 500 on the image, both days: `mimo-v2.5-pro`, `qwen3.7-max`. All are marked text-only. |
| Text ✓, image ✓, tool intermittent | `qwen3.7-plus`: the call was made in 3 of 5 runs; in the other two the model answered the text "call" without calling. |
| Works except from mainland China | `gpt-5.6-luna`: text, tool and image ✓ from the dev box on 09-17. From vps3: `403 unsupported_country_region_territory`. |
| 403 until the workspace opts in | `muse-spark-1.3-contributor`, `muse-spark-1.2-contributor`: Meta training consent. From vps3 they also answer `403 This model is not available in your country`. `deepseek-v4.1-flash`, `deepseek-v4-flash`, `deepseek-v4-pro`: China-hosting consent. The test key's workspace has not opted in. |

**Retries on 09-17.** Two transient failures passed on rerun: a `kimi-k2.6` malformed stream on the image and a `mimo-v2.5` tool-call timeout. Both passed first time on 09-18.

## Notes and follow-ups

- **User agent.** OpenCode asks clients to identify with their own user agent rather than a generic SDK name. The harness sends `OpenAI/JS 6.42.0` and `Anthropic/JS 0.91.1`. Sending `PenguinHarness/<version>` for `opencode.ai` hosts would be a small follow-up next to `attributionHeaders`. Not done here.
- **Hand-added models.** A model added to this group by hand gets `openai-chat`, the gateway rule. A future Messages-only model would need `penguin config model add --client-type ant-messages --base-url https://opencode.ai/zen/go`, or the custom group.
- **Spec wording.** 06's sentence about the groups after DeepSeek predates this PR and does not match `MODEL_PROVIDERS` exactly. For example, SiliconFlow sits after Anthropic, and MiniMax is not named. The design PR leaves it alone.

## Verification

Merged `origin/main` into this branch. It brings the #749 catalog refresh and the #758 docs restructure; the merge commit lists how each conflict was resolved.

On vps3, using `vps3-vitest.sh` on this branch's tree with a fresh install and a core/hmr/server build:
- [x] core `test/model-catalog.test.ts`: 1 file, 48 tests passed.
- [x] web `model-grouping`, `catalog-sync`, `protocol-path`, `model-group-expansion` and `model-group-order`: 5 files, 98 tests passed.
- [x] server `test/models.test.ts`: 1 file, 29 tests passed.
- [x] docs package: 8 files, 62 tests passed.
- [x] `pnpm typecheck` (12 packages), `pnpm lint` (0 warnings, 0 errors) and `pnpm format:check`: all clean.

Other checks:
- [x] Locally: `git diff --check` clean, and prettier run on the touched files.
- [x] opencode.ai/docs/go, `/zen/go/v1/models` and models.dev re-read 2026-09-18. Ids, endpoints and prices are unchanged, except that `union-alpha` has been removed.
- [x] Live test as described above. The key was copied to vps3 over stdin into a mode-600 file and deleted after the run. The probe output contains no key material or workspace id.
- [x] The glyph was rendered at 16 and 32 px in light and dark themes, next to the other marks.

After the review fixes (`ec5f90622`), on vps3 from a fresh sync and build:
- [x] core `test/model-catalog.test.ts`: 1 file, 48 tests passed.
- [x] web `protocol-path` and `model-grouping`: 2 files, 43 tests passed.
- [x] `pnpm typecheck` (12 packages), `pnpm lint` (0 warnings, 0 errors) and `pnpm format:check`: all clean.


hiyouga (commented):
Independent review at `d39f95d56`, against `origin/main` (with #749 and #758) and design Prism-Shadow/penguin-harness-design#191. I checked the rows against opencode.ai/docs/go and models.dev's `opencode-go` provider, both fetched 2026-09-18. I made no live calls.

**Checked, no finding**
- **The rows against the docs page.** All 27 ids match. The endpoint table maps to 16 `openai-chat`, 4 `openai-responses` and 7 `ant-messages` rows. Both base URLs are right: `/zen/go/v1`, and `/zen/go` for Messages, where the SDK adds `/v1/messages`. Every input, output, cache-read and cache-write rate matches (`cache_write` is the input rate where the page shows `-`), and so do the four base tiers.
- **Context windows and vision flags.** For all 27 rows they equal models.dev's `limit.context` and `modalities.input`. The exception is hy3; see the inline comment.
- **DeepSeek schedule.** OpenCode's peak windows (01:00–04:00 and 06:00–10:00 UTC, Mon–Fri) are exactly `DEEPSEEK_OFF_PEAK`. The UTC windows never cross midnight, so the weekday agrees. Each Off-Peak figure is exactly half the peak one. `offPeakScheduledRefs` carries the four pairs, so `tieredRates`, `ratesAt`, `peakExpr` and the web's `discountedPrice` split them by tier while the stored price is the catalog peak. No `discount` is correct: the 4x offer raises an allowance, not a rate.
- **Group order.** `groupModelRows`, `allGroupKeys`, `orderModelsLikeLibrary`/`visibleChatModels`, the dialog's group select and both docs tables follow `MODEL_PROVIDERS`. A stored per-Project order is read as is, and `opencode-go` trails it (`orderModelGroups`), as the changelog says. A new Project's `[[models]]` rows and the Sync-presets ref list follow `MODEL_CATALOG` cluster order, which was never the display order.
- **`x-opencode-session`.** Both bases are on `opencode.ai`. `attributionHeaders` is keyed on the host, and all three AgentHub 0.4.15 generic clients pass `defaultHeaders`. The rows therefore get the header exactly as #612's custom entries do, on Session requests only.
- **Docs and changelog.** The en and zh files mirror each other section for section. CJK–Latin spacing, full-width punctuation and the quoted UI labels (`Set key`/`手动设置密钥`, `Sync presets`/`同步预置`) are right, and so are the metadata order and the PR link.
- **Over-defensive code.** None. The runtime diff is data, two constants and a glyph, and the glyph's path is identical to models.dev/logos/opencode-go.svg.

**Findings** (inline unless noted)
1. **[bug]** hy3's `contextWindow` is above its input cap (`model-catalog.ts:1949`).
2. **[bug], design call, not blocking:** what the #191 env-fallback clause means for this group (`model-catalog.ts:279`).
3. **[test]** `protocol-path.test.ts:71` asserts nothing about the group.
4. **[simplify]** `model-catalog.test.ts:276` and `:297` restate other assertions.
5. **[copy]** (design #191, not in this diff): the 06「模型配置页」 sentence that #191 edits still lists the groups after DeepSeek in the wrong order. It gives OpenRouter, Fireworks AI, SiliconFlow, Qwen Token Plan, Qwen Pay-As-You-Go, then Google Gemini, Anthropic, OpenAI, Z.AI, Moonshot, vLLM, custom. `MODEL_PROVIDERS` and this PR's docs table run OpenRouter, Fireworks AI, Google Gemini, OpenAI, Anthropic, SiliconFlow, Z.AI, Moonshot, MiniMax, Qwen Pay-As-You-Go, Qwen Token Plan, vLLM, custom. #191 already rewrites that sentence, so correct its tail there. No code change is needed.

Verdict: fix before merge: 1 (hy3 `contextWindow` 256000 → 192000, plus its pinned test row). 3 and 4 are deletions for the same push; 2 is a design decision for the maintainer; 5 goes into #191.

packages/core/src/state/model-catalog.ts
```diff
@@ -1762,6 +1784,324 @@ export const MODEL_CATALOG: ModelCatalogEntry[] = [
...
+    pricing: usd(0.042, 0.834, 2.501),
+    supportsVision: false,
+    clientType: "openai-chat",
+    baseUrl: OPENCODE_GO_BASE_URL,
+  },
+  {
+    modelId: "hy3",
+    displayName: "Hy3",
+    provider: "opencode-go",
+    contextWindow: 256000,
```
hiyouga: **[bug]** `hy3` stores models.dev's `limit.context`, but the same entry also caps the prompt: `opencode-go/hy3` is `"limit":{"context":256000,"input":192000,"output":128000}`. 9 of the 13 Hy3 listings on models.dev carry the same `input: 192000`, OpenRouter's included.

Compaction is derived from this field alone. An Agent on the seeded `max_context_length` compacts at `effectiveMaxContextLength(256000, 256000)` = 253,952.

Scenario: a Session on this row grows past 192K tokens. From there to ~254K, every request is over Hy3's input cap and the upstream refuses it. Nothing compacts in reaction to that error, so every turn fails until the user compacts by hand.

Fix: set `contextWindow: 192000` here and in the pinned row at `test/model-catalog.test.ts:219`. Compaction then fires at 189,952, and the per-request output clamp keeps input plus output under the cap. OpenRouter's `tencent/hy3` (262144) has the same exposure, but it is outside this PR.

packages/core/src/state/model-catalog.ts
```diff
@@ -263,6 +272,19 @@ export const MODEL_PROVIDERS: ModelProviderInfo[] = [
     apiKeyUrl: "https://token.penguin.ooo/",
     modelsUrl: "https://token.penguin.ooo/",
   },
+  {
+    // One key serves the whole group, but its models sit on three protocols (see the group's
+    // block comment in MODEL_CATALOG), so no group-level pin. The env pair is the one its
+    // Chat Completions majority reads and a model added here by hand gets: the preset base
+    // URL below with openai-chat. Its Messages rows fall back to ANTHROPIC_* instead.
```
hiyouga, line 279: **[bug], design call, not blocking.** The PRN-021 clause that #191 adds is accurate:
- AgentHub 0.4.15 `ant_messages/client.ts:64-65` reads `ANTHROPIC_API_KEY`/`ANTHROPIC_BASE_URL`.
- `openai_chat` and `openai_responses` read `OPENAI_*`.
- Core passes only `entry.api_key`.

What that means here: a keyless OpenCode Go row authenticates to `opencode.ai` with the user's own Anthropic or OpenAI key. Three paths reach it:
- **Speed test and Test connection.** `probeApiKey` → `modelEnvironmentApiKey` resolves `process.env.ANTHROPIC_API_KEY` for the 7 Messages rows and `OPENAI_API_KEY` for the other 20. One confirmed click on the group's Speed test sends 27 requests to OpenCode: 7 carry the Anthropic key and 20 carry the OpenAI key.
- **The edit dialog suggests it.** `envHintKey` checks `detectedEnvKeys`, and any first-party Anthropic row fills that set. A Messages row's key field then reads "Leave empty to use the ANTHROPIC_API_KEY env var". The docstring of `envFallbackFirstParty` calls this steer "exactly the misconfiguration the preview must not encourage".
- **Any Session on the row.**

The `OPENAI_*` half is already the rule for every gateway. The `ANTHROPIC_*` half is new for a gateway group; the only precedent is the custom Atria preset. If this should not ship, the smallest change is to stop the probe paths and the dialog hint from resolving environment keys for groups that have a `gatewayBaseUrl`. `detectProtocol` already passes `env: {}` for the relay.

packages/web/test/protocol-path.test.ts
```diff
@@ -48,6 +49,28 @@ describe("protocolPathForModel", () => {
...
+    expect(rows).toHaveLength(27);
+    const hit = new Set<string>();
+    for (const m of rows) {
+      const url = `${m.baseUrl}${protocolPathForModel(m.provider, m.clientType ?? "")}`;
+      expect(endpoints.has(url), `${m.modelId} -> ${url}`).toBe(true);
+      hit.add(url);
+    }
+    expect(hit).toEqual(endpoints);
+    // A model added to the group by hand takes the Chat Completions default.
+    expect(protocolPathForModel("opencode-go", "openai-chat")).toBe("/chat/completions");
```
hiyouga: **[test]** This line asserts nothing about OpenCode Go. `protocolPathForModel` returns `/chat/completions` for `openai-chat` whatever the provider, because an explicit client type returns before `switch (provider)`.

The default that the comment claims to cover lives in two places this test never exercises:
- the new-row form in `models-page.tsx`: `pinnedClientType ?? (vendorAdd || isCustomLikeGroup(addProvider) ? "" : "openai-chat")`;
- the CLI's `defaultClientType`.

Delete lines 70–71.

packages/core/test/model-catalog.test.ts
```diff
@@ -182,6 +183,137 @@ describe("model-catalog", () => {
...
+    for (const m of scheduled) {
+      expect(m.offPeakDiscount, m.modelId).toBe(DEEPSEEK_OFF_PEAK);
+      const off = effectivePricing(m, saturday)!;
+      expect([off.cache_read, off.cache_write, off.output], m.modelId).toEqual(
+        m.modelId === "deepseek-v4-pro" ? [0.022, 0.66, 1.98] : [0.003, 0.15, 0.6],
+      );
+    }
+    // No promotion is on a rate: the running "4x" offer raises an allowance, not a price.
+    expect(go.filter((m) => m.discount !== undefined)).toEqual([]);
+    expect(presetPromotions().filter((p) => p.provider === "opencode-go")).toEqual([]);
```
hiyouga: **[simplify]** Two assertions in this test restate others:
- Line 297 cannot fail once line 296 passes, because `presetPromotions` only emits rows that carry a `discount`.
- Line 276 checks `unit === "usd_per_mtok"`, which the literal type of `ModelPricing.unit` and `usd()` already guarantee.

Delete both.

packages/core/src/state/model-catalog.ts
hiyouga: Fixed in ec5f90622: hy3 now stores 192000 (models.dev's input cap), with a row comment saying why, and the pinned test row matches. OpenRouter's `tencent/hy3` is left for a separate change.

packages/web/test/protocol-path.test.ts
hiyouga: Fixed in ec5f90622: lines 70–71 removed.

packages/core/test/model-catalog.test.ts
hiyouga: Fixed in ec5f90622: both assertions removed (the `usd_per_mtok` check and the `presetPromotions` one).

packages/core/src/state/model-catalog.ts
hiyouga: Waiting on a product decision

hiyouga: Decided by the user: gateway groups never fall back to vendor env keys. Implemented separately as a Breaking change in #794 (design #196), which covers OpenCode Go's rows by their opencode.ai base URLs; nothing to change in this PR.

PR #787, opened 2026-09-18
feat(web): keyboard shortcuts follow the platform, from one registry

hiyouga:
## Summary

Design: penguin-harness-design [#190](https://github.com/Prism-Shadow/penguin-harness-design/pull/190). This is PR A of three (A: registry and platform-correct modifiers; B: settings page and per-account storage; C: six more commands).

The Web App's four hand-written shortcut checks become commands in `packages/web/src/lib/shortcuts/`:

- **Registry** (`registry.ts`): `palette.toggle` (Mod+P, reserved for #768), `terminal.toggle` (Ctrl+` on every platform — ⌘` is macOS window cycling), `terminal.close` (Mod+W), `editor.save` (Mod+S). `Mod` is ⌘ on macOS and Ctrl elsewhere; a literal `Ctrl+` token is allowed only where macOS must not follow ⌘, and the registry test holds that line.
- **Matcher** (`chord.ts`, `match.ts`): chords match on `KeyboardEvent.code`, so a CJK IME (`key: "Process"`), Shift, Caps Lock and a macOS Option chord (`key: "∑"`) do not change what a key means; auto-repeat and IME composition never match; Meta on Windows/Linux belongs to the OS. A focus scope (terminal, editor) beats global on the same chord, then registry order.
- **Formatter** (`format.ts`, `components/ui/kbd.tsx`): `⌘W`, `⌃⇧\`` on a Mac; `Ctrl+Shift+W` elsewhere; layout-map letters when the browser exposes one.
- **Store** (`store.ts`): defaults overlaid by the `penguin.keybindings` mirror (`{ v: 1, mac|windows|linux: { id: chord | null } }`, overrides only, unknown ids preserved, a row set back to its default removed); the `storage` event keeps every tab live; on Chromium/Electron the defaults relocate to the key that types their letter (AZERTY `KeyW` → `KeyZ`). No settings UI and no server copy yet — PR B.
- **Dispatcher** (`dispatcher.ts`): one window keydown listener installed at module evaluation; `onCommand` / `runCommand` / `hasCommandHandler` so #768 can register `palette.toggle` in an effect.
- **Terminal** (`terminal-view.tsx`): the clipboard keys split by platform (`terminal-clipboard.ts`: macOS keeps ⌘C/⌘V native, Ctrl+C is always SIGINT, Ctrl+V reaches the shell); then the keymap — `terminal.close` is consumed with `stopPropagation`, a global command that has a handler is skipped from the shell and left un-prevented for the window dispatcher (VS Code's commands-to-skip-shell), and anything else goes to the shell as typed.
- **Display sites** format from the registry and follow a binding change live: the terminal tab's × tooltip, the panel picker's kbd, the Files panel's save title (`S.common.save` + chord; `S.files.saveTitle` removed), the handbook's editor hint (now a function of the chord).
- **Guard** (`test/shortcut-guard.test.ts`): a `.ctrlKey` / `.metaKey` read anywhere outside `lib/shortcuts/` fails, naming file and line.
- `KEY_RULES` gains the `browser`-scoped `penguin.keybindings` row.

Behaviour changes for a Mac user: the terminal tab closes on ⌘W and Ctrl+W now reaches the shell (readline delete-word); the editors save on ⌘S only; Ctrl+C over a selection no longer copies (⌘C does) and Ctrl+V is no longer swallowed. Windows/Linux behaviour is unchanged: xterm never sent Ctrl+` to the shell, and Ctrl+P still reaches it until a surface answers `palette.toggle`.

Deviations from the design document, each deliberate: the dispatcher does not skip events targeted inside `.xterm` (the design's two bullets on this contradicted each other; the terminal handler's `stopPropagation` and the shared matcher make the skip unnecessary and the global-command bubble path requires its absence); the xterm handler only skips the shell for a global command that has a registered handler, so `palette.toggle`'s Ctrl+P keeps reaching readline until #768 lands; the save title reuses `S.common.save` rather than adding `S.files.save`; Ctrl+Q and Ctrl+M are added to the Windows/Linux menu-accelerator table (Electron's `quit` and `minimize` roles carry them).

## Verification

All on vps3 via `vps3-vitest.sh` (the local box only edits and runs git):

- `pnpm lint` — 0 warnings, 0 errors
- `pnpm format:check` — clean
- `pnpm --filter @prismshadow/penguin-web test` — 196 files, 2526 tests passed after the review fixes (the ten new suites: chord, match, format, registry, conflicts, store, dispatcher, terminal-clipboard, terminal-keys, shortcut-guard; and the touched `install-scope`, `dock-state`, `control-size`)
- `git diff --check` — clean

Not run: a manual browser check (Playwright is not set up on vps3 and dev servers do not run locally); `e2e/dock.spec.mjs` keeps `Control+Backquote`, which is unchanged on Linux.

## Review fixes

The second commit answers [the review](https://github.com/Prism-Shadow/penguin-harness/pull/787#pullrequestreview-5248315253) items 1–6 and the cleanups; each thread names it. A browser tab now shows no chord the browser reserves (the × tooltip on a Mac browser no longer advertises ⌘W); a held chord's repeats are kept from the browser without re-running the command; relocation covers every code in the layout map (US-Dvorak); the desktop menu table matches Electron 43.2's roles and the shell table is empty on main; the terminal's key decision is a pure function with a table test.


hiyouga (commented):
Independent review of 5117be61d against origin/main, the design (S-design), xterm 5.5.0's key handling and Electron 43.2.0's menu roles. Details are inline; summary here.

**Fix before merge**
1. **[bug]** `test/shortcut-guard.test.ts` fails on Windows: `test-windows (rest)` is red on this PR on both attempts, and on #789 and #790, because `path.relative` yields `lib\shortcuts\…` there. One-line fix inline.
2. **[bug]** Auto-repeat is dropped from matching, so a held chord runs once and its repeats reach the browser's own action (Save Page today; Print once the palette follow-up lands; Downloads and address-bar search with #790's Ctrl+J / Ctrl+K). Inline at chord.ts:109.
3. **[bug]** In a Mac browser the terminal tab's × tooltip now advertises ⌘W, which closes the browser tab (the whole app) before the page sees it; main's "Ctrl+W" worked there. Inline at dock-panel.tsx:329.
4. **[bug]** Layout relocation searches only `Key*` codes, so with US-Dvorak (s and w sit on `Semicolon` and `Comma`) the desktop app moves save to Ctrl+O and terminal close to the comma key, while Ctrl+S and Ctrl+W, which main matched by `key`, stop working. Inline at match.ts:52.
5. **[bug]** `MENU_BY_PLATFORM` disagrees with Electron 43.2's roles (Windows has no Ctrl+Q, Linux redo is Ctrl+Shift+Z, zoom-in carries Shift, macOS lacks ⌥⇧⌘V), and `SHELL_BY_PLATFORM` describes #625/#628, which main does not have. Inline at reserved.ts:104.
6. **[copy]** The changelog pair renders with broken code spans in three places each, "no longer on either" reads as "on neither", and "closes on ⌘W on a Mac" needs "in the desktop app". Inline.

**Cleanups (can ride along):** **[simplify]** the dispatcher's install/uninstall machinery exists only for its test; **[simplify]** `relocateDefaults`, `isCommandId` and `ShortcutCommand.rebindable` have no reader anywhere in #787–#790; **[test]** the terminal's key decision, which carries the Mac behaviour change, is untested; **[copy]** two comments contradict the code, and the dictionaries' terminal doc comment now sits on `shortcuts`.

**Command palette.** origin/main has no palette: #551's base was `feat/plugin-npm-install`, its merge commit ea3024a59 is not an ancestor of origin/main, and `git grep` on origin/main finds no `command-palette` or `isCommandPaletteShortcut` under `packages/web`; #768 is still open. So on main ⌘P / Ctrl+P is the browser's Print, and Ctrl+P in a focused terminal is readline's ^P. Both are unchanged on this branch (no handler, so the dispatcher prevents nothing and the xterm handler returns true): the description's claim holds and nothing regresses. `palette-follow-up.patch` edits `features/palette/command-palette.tsx`, `lib/command-palette.ts` and `test/command-palette.test.ts`, none of which exist here, so it cannot go in this PR; it belongs on #768 when that rebases onto this one, and the guard test will fail #768 until it does (`isCommandPaletteShortcut` reads `.metaKey` / `.ctrlKey`). Two effects to expect when it lands: without fix 2 a held ⌘P ends in the print dialog, and once `palette.toggle` has a handler, Ctrl+P in a focused terminal opens the palette instead of reaching readline (VS Code's default, but a change for shell users).

**PR description.** "a global chord pressed inside a focused terminal is no longer also sent to the pty": xterm 5.5 maps nothing for `` Ctrl+` `` (keyCode 192 is not in `evaluateKeyboardEvent`'s Ctrl table), so main never sent it, and Ctrl+P still reaches the pty here. On Windows/Linux the terminal behaves exactly as before in this PR.

**Deviations.** (1) No `.xterm` skip in the dispatcher: agreed. The design's two bullets contradict each other, the skip-shell path needs the bubble, `terminal.close` is stopped, and whatever xterm leaves uncancelled re-matches on the window against a subset of the scopes, so nothing fires twice. (2) Skip the shell only for a command with a handler: agreed, and better than the design, which would have swallowed Ctrl+P in the terminal with nothing to run it. (3) `S.common.save`: fine. (4) Ctrl+Q / Ctrl+M: Ctrl+M is right on both, Ctrl+Q only on Linux (fix 5).

**Checked, no finding.** All four `ctrlKey`/`metaKey` sites on main moved (terminal clipboard and close, the toggle, both editor saves), and no other keydown handler on main reads Ctrl or ⌘; on a US layout every Windows/Linux chord is the same as before. Global commands fire inside inputs exactly as `` Ctrl+` `` did on main; there is no Monaco, CodeMirror or contenteditable editor, and both editors are textareas whose React handler prevents `editor.save` before the window listener runs. No ⌘+letter reaches the pty (xterm 5.5 maps only ⌘A on a Mac). `isComposing` is honoured, and keyCode 229 needs no check because every binding carries a modifier. Dropdown, Modal and the esc-layer act only on Escape and arrows. A malformed, foreign or wrong-`v` `penguin.keybindings` degrades to the defaults, and the `storage` listener re-reads on another tab's write or a clear. Both dictionaries carry the same keys and signatures.

Verdict: fix before merge: 1 (Windows guard-test path), 2 (auto-repeat reaches the browser default), 3 (⌘W in the Mac browser tooltip), 4 (Dvorak relocation), 5 (Electron menu/shell tables), 6 (changelog).

packages/web/test/shortcut-guard.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  const elsewhere: string[] = [];
+  const walk = (dir: string): void => {
+    for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+      const full = path.join(dir, entry.name);
+      if (entry.isDirectory()) {
+        walk(full);
+        continue;
+      }
+      if (!/\.tsx?$/.test(entry.name)) continue;
+      const rel = path.relative(WEB_SRC, full);
```
hiyouga: **[bug]** This fails on Windows: `test-windows (rest)` is red on this PR on both attempts, and on #789 and #790. `path.relative` returns `lib\shortcuts\chord.ts` there, so `rel.startsWith(ALLOWED_DIR)` on line 39 is false, all ten allowed reads land in `elsewhere`, and the "not vacuous" case sees 0. The PR's verification ran on Linux only. Normalize the way `control-size`, `disclosure-anchor` and `icon-scale` do:

```ts
const rel = path.relative(WEB_SRC, full).replaceAll(path.sep, "/");
```

packages/web/src/lib/shortcuts/chord.ts
```diff
@@ -0,0 +1,128 @@
...
+  return /^[A-Z][A-Za-z0-9]*$/.test(key) ? key : "";
+}
+
+/**
+ * The chord an event IS, or null when it is not one: a modifier-only press, an auto-repeat (a
+ * held ⌘W must not close two tabs), a key inside an IME composition (the IME owns it), a Meta
+ * chord on Windows/Linux (the Super key belongs to the OS), or an event with no usable code.
+ */
+export function chordOf(e: KeyLike, platform: Platform): Chord | null {
+  if (e.repeat === true || e.isComposing === true) return null;
```
hiyouga: **[bug]** Dropping auto-repeat here makes a repeat "not a chord", so nothing prevents it and the browser's own action for a bound chord fires from the first repeat on; main matched, and prevented, every repeat. Today: hold Ctrl+S / ⌘S a little past the repeat delay in the Files editor with nothing to save (`requestSave` only toasts, so focus stays in the textarea) and the browser's Save Page dialog opens. It gets louder up the stack: with the palette follow-up, a held ⌘P / Ctrl+P opens the palette and then the print dialog (#768's own listener prevents every match for exactly that reason), and #790's Ctrl+J / Ctrl+K hand their repeats to Chrome's Downloads page and address-bar search.

Fix: leave `repeat` out of `chordOf` and have each caller consume a repeat without re-running it: the editors `e.preventDefault(); if (!e.repeat) onSave();`; `terminal.close` prevents and stops as now but calls `onCloseRequest` only when `!event.repeat`; the dispatcher does `if (event.repeat) { if (hasCommandHandler(id)) event.preventDefault(); return; }` before `runCommand`. A held ⌘W still closes one tab.

packages/web/src/features/dock/dock-panel.tsx
```diff
@@ -323,6 +326,7 @@ export function DockPanel({
 }: DockPanelProps) {
   useSyncExternalStore(subscribeDock, dockVersion);
   const terminals = useSyncExternalStore(subscribeTerminals, liveTerminals);
+  const closeShortcut = useShortcutLabel("terminal.close");
```
hiyouga, line 329: **[bug]** In a browser tab on a Mac the × tooltip now reads "… (⌘W)", and ⌘W there closes the browser tab, the whole app, before the page sees it. On main it read "Ctrl+W", which on a Mac browser did close the terminal tab. The page already has both facts (`currentHost()`, `browserReserved`): leave the chord out of the tooltip when the host is a browser and the chord is browser-reserved. That also retires the Windows/Linux browser tooltip's "Ctrl+W", which has always closed the browser tab instead.

packages/web/src/lib/shortcuts/match.ts
```diff
@@ -0,0 +1,65 @@
...
+ * "the W key" means the key labelled W on an AZERTY board (`KeyZ`) too. Only letter keys move;
+ * user-recorded chords are physical and are never relocated.
+ */
+export function relocateChord(chord: Chord, layout: ReadonlyMap<string, string>): Chord {
+  const letter = /^Key([A-Z])$/.exec(chord.code)?.[1];
+  if (letter === undefined) return chord;
+  const produced = layout.get(chord.code)?.toUpperCase();
+  if (produced === undefined || produced === letter) return chord;
+  for (const [code, char] of layout) {
+    if (/^Key[A-Z]$/.test(code) && char.toUpperCase() === letter) return { ...chord, code };
```
hiyouga: **[bug]** This looks for the letter only on `Key*` codes, but the layout map (Chromium, Electron) also covers `Semicolon`, `Comma`, `Period`, `Slash` and `Quote`, and US-Dvorak puts s, w, v and z there. Windows or Linux desktop app with US-Dvorak: `KeyS` types "o" and no `Key*` code types "s", so `editor.save` stays on `KeyS`: the editors now save on Ctrl+O (tooltip "Save (Ctrl+O)") and Ctrl+S does nothing (Chrome: Save Page). `terminal.close` stays on `KeyW`, which types a comma, so Ctrl+W goes to readline and Ctrl+comma closes the tab. Main matched `e.key`, so Ctrl+S and Ctrl+W worked there.

Fix: drop the `/^Key[A-Z]$/` test on this line so any code whose character is the letter qualifies, and have `keyLabel` (format.ts:66-70) read `layout.get(code)` for any code rather than letters only, so a relocated `Semicolon` is labelled S, not ";". One caveat if it matters: macOS's "Dvorak – QWERTY ⌘" input source switches to QWERTY while ⌘ is held, which no layout map shows; relocation already moves ⌘P to `KeyR` there, and this would do the same to ⌘S and ⌘W, so that source wants relocation off for Mod chords on a Mac.

packages/web/src/lib/shortcuts/reserved.ts
```diff
@@ -0,0 +1,158 @@
...
+    "Ctrl+KeyR", // reload
+    "Ctrl+Shift+KeyR", // force reload
+    "Ctrl+Shift+KeyI", // toggle DevTools
+    "Ctrl+Digit0", // actual size
+    "Ctrl+Equal", // zoom in
+    "Ctrl+Minus", // zoom out
+    "F11", // toggle full screen
+    "Ctrl+KeyM", // minimize
+    "Ctrl+KeyW", // close (hides the window to the tray)
+    "Ctrl+KeyQ", // quit
```
hiyouga: **[bug]** Checked against Electron's role table at v43.2.0 (`lib/browser/api/menu-item-roles.ts`; the desktop package pins ^43.2.0):
- `quit`: `accelerator: isWindows ? undefined : 'CommandOrControl+Q'`. Windows' Exit item has no accelerator, so this line goes; Ctrl+Q is Linux-only.
- `redo`: `isWindows ? 'Control+Y' : 'Shift+CommandOrControl+Z'`. Linux redo is Ctrl+Shift+Z, so line 108's `Ctrl+KeyY` should be `Ctrl+Shift+KeyZ`.
- `zoomin`: `CommandOrControl+Plus`, and `shell/common/keyboard_util.cc` maps `plus` to `VKEY_OEM_PLUS` with a shifted character, which `accelerator_util.cc` turns into Shift. The accelerator is Ctrl+Shift+= (⌘⇧= on a Mac), so lines 80, 99 and 117 want `…+Shift+Equal`.
- The macOS `editMenu` also carries `pasteAndMatchStyle` (`Cmd+Option+Shift+V`), which the mac list lacks.

Ctrl+M (`minimize`, `CommandOrControl+M`) is right on both. Separately, `SHELL_BY_PLATFORM` (lines 61-65) is the `before-input-event` handling of #625/#628, both open on other branches; main's shell has none, so on main F10 and F12 reach the page and Ctrl+Shift+I is the View menu's toggleDevTools accelerator (`"menu"`, not `"shell"`). As written, #789 would tell a Windows user that Ctrl+Q shadows a menu shortcut that does not exist, and say nothing to a Linux user who binds Ctrl+Shift+Z.

changelog/unreleased/2026-09-18-keyboard-shortcuts.md
```diff
@@ -0,0 +1,33 @@
...
+- **Scope:** `web`
+- **PR:** [#787](https://github.com/Prism-Shadow/penguin-harness/pull/787)
+
+[中文版](2026-09-18-keyboard-shortcuts.zh.md)
+
+The Web App's shortcuts now come from one registry (`lib/shortcuts/`) instead of four hand-written
+modifier checks, and the modifier follows the platform: ⌘ on macOS, Ctrl elsewhere. The terminal
+tab closes on ⌘W on a Mac, where Ctrl+W now passes to the shell as readline's delete-word; Windows
+and Linux keep Ctrl+W. The editors in the Files panel and the handbook save on ⌘S / Ctrl+S, no
+longer on either. The terminal toggle stays Ctrl+` on every platform (⌘` is macOS's own window
```
hiyouga: **[copy]** Three code spans in this file break when rendered (CommonMark/GFM; I rendered both files with micromark): a lone backtick after `Ctrl+` opens a span that the next lone backtick closes, and a backslash does not escape inside a span. Line 14 renders " on every platform (⌘" as code; line 22 turns "), " and " (⌘W / Ctrl+W), " into code and leaves `terminal.close` and `editor.save` as plain text; line 29 swallows the ⌃-backquote chord and the text after it into one span and leaves a stray backtick after Ctrl+W. The .zh.md has the same three at lines 10, 14 and 16. Double-backtick spans fix all six:

```md
longer on either. The terminal toggle stays `` Ctrl+` `` on every platform (`` ⌘` `` is macOS's own window
  (`` Ctrl+` ``), `terminal.close` (⌘W / Ctrl+W), `editor.save` (⌘S / Ctrl+S). Each has a scope
  writes it (`⌘W`, `` ⌃` ``; `Ctrl+W`), and updates when the binding changes.
```

Two wording points: "save on ⌘S / Ctrl+S, no longer on either" (lines 13-14) reads as "on neither"; say "save on ⌘S on a Mac and Ctrl+S elsewhere, instead of accepting both everywhere". And "The terminal tab closes on ⌘W on a Mac" wants "in the desktop app": a browser keeps ⌘W for closing its own tab, which the spec in design #190 states and this entry does not.

packages/web/src/lib/strings.ts
```diff
@@ -80,6 +80,22 @@ export const zh = {
   },
 
   /** Server-side terminal (the in-app dock and the standalone /terminal page). */
+  /** Keyboard shortcuts: the registry's command and group labels (lib/shortcuts/registry.ts). */
```
hiyouga, line 82: **[copy]** The `shortcuts` block went in between the terminal section's doc comment (line 82, "Server-side terminal (the in-app dock and the standalone /terminal page)") and `terminal:` (line 99), so `shortcuts` now carries two doc comments and `terminal` none. Same in strings-en.ts at lines 77-78. Move the block above line 82 in both dictionaries.

packages/web/src/features/terminal/terminal-view.tsx
```diff
@@ -350,46 +355,43 @@ export function TerminalView({
...
-       * - copy: Ctrl+Shift+C, Ctrl+Insert, or plain Ctrl+C while a selection exists
-       *   (SIGINT still goes through when nothing is selected);
-       * - paste: Ctrl+V, Ctrl+Shift+V and Shift+Insert all ride the browser's NATIVE paste
-       *   event into xterm's textarea (no clipboard permission involved) — the browser
-       *   fires `paste` for every one of these, so returning false (skip xterm's own key
-       *   handling, keep the browser default) is the whole implementation; calling the
-       *   async clipboard API here as well double-pastes.
+       * Keys the terminal decides before xterm does. xterm hands this handler its own
+       * textarea's events and nothing else, so everything here is seen only by the terminal
+       * that has focus, with no window-level listener involved.
```
hiyouga: **[copy]** "with no window-level listener involved" no longer holds: point 2 below relies on the window dispatcher for global commands. A similar slip in store.ts:227: the `storage` event never fires in the tab that wrote, so "the settings dialog in this one, through the mirror" is not something that listener covers; `writeStored` notifies the writing tab itself (store.ts:100).

packages/web/src/features/terminal/terminal-view.tsx
```diff
@@ -350,46 +355,43 @@ export function TerminalView({
...
-        // Ctrl+W closes this terminal when the host offers a close (the dock tab's ×, the
-        // standalone page). Consumed here so it never reaches the shell, where it is
-        // readline's delete-word, and seen only by the terminal that has focus — xterm hands
-        // this handler its own textarea's events and nothing else, so no window-level
-        // listener is involved. Browsers keep Ctrl+W for closing the browser tab and may act
-        // first; the desktop shell delivers it here.
-        const closeCombo =
-          event.ctrlKey && !event.shiftKey && !event.altKey && !event.metaKey && key === "w";
-        if (closeCombo && callbacks.current.onCloseRequest) {
+        const command = matchShortcut(event, keymap(), ["terminal", "global"], platform);
```
hiyouga: **[test]** This is the decision behind the Mac behaviour change the changelog leads with (⌘W closes, Ctrl+W reaches readline, the terminal toggle pressed inside a focused xterm runs without reaching the pty, Ctrl+P reaches the pty while nothing answers it), and nothing pins it. `shortcuts-match.test.ts` covers `matchShortcut`, not the scopes passed here, the missing-`onCloseRequest` fallback, the `hasCommandHandler` gate, or which branch prevents and stops; `e2e/dock.spec.mjs` presses Control+Backquote only with the composer or the page focused, never inside the xterm. Lifting lines 386-395 into a pure function next to `terminal-clipboard.ts` (say `terminalKeyAction(e, keymap, platform, { canClose, hasHandler })` returning `"close" | "skip-shell" | "shell"`) turns this into a table test.

packages/web/src/lib/shortcuts/dispatcher.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+  return false;
+}
+
+export interface KeydownTarget {
+  addEventListener(type: "keydown", listener: (event: KeyboardEvent) => void): void;
+  removeEventListener(type: "keydown", listener: (event: KeyboardEvent) => void): void;
+}
+
+const installed = new WeakSet<KeydownTarget>();
```
hiyouga: **[simplify]** `KeydownTarget`, this WeakSet, the idempotency and the returned uninstall exist only for the tests: production installs once, on line 78, at module evaluation (and a Vite HMR re-evaluation gets a fresh WeakSet anyway, so the guard does not cover the one re-run that happens). Export the listener itself, register it here with `window.addEventListener("keydown", …)`, and let the tests call it with their fake event; `installShortcutDispatcher`, `KeydownTarget`, `installed` and the "installs once per target" case go.

packages/web/src/lib/shortcuts/match.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return chord;
+}
+
+export function relocateDefaults(
```
hiyouga: **[simplify]** No production caller anywhere in the stack (#787 through #790): the store relocates one chord at a time with `relocateChord` (store.ts:112), so this function and its test case only test each other. The same holds for `isCommandId` (registry.ts:62, called only by the registry test) and `ShortcutCommand.rebindable` (types.ts:64-65: `true` on every row and read by nothing, #789's settings page included). Delete all three.

packages/web/test/shortcut-guard.test.ts
hiyouga: Fixed in 5ec02126b: `path.relative(...).replaceAll(path.sep, "/")`, as the other guard tests do.

packages/web/src/lib/shortcuts/chord.ts
hiyouga: Fixed in 5ec02126b: `chordOf` no longer drops a repeat. The editors `preventDefault` every match and save only when `!e.repeat`; the terminal's decision (now the pure `terminalKeyAction`) returns `consume` for a repeat of the close chord — prevent and stop, run nothing; the dispatcher prevents a repeat of a command that has a handler and returns before `runCommand`. Tests: the chord test asserts a repeat is the chord, the dispatcher test holds Ctrl+` and sees one toggle with the repeat prevented, and the terminal-keys table has the held ⌘W case.

packages/web/src/features/dock/dock-panel.tsx
hiyouga: Fixed in 5ec02126b: `shownOnHost(chord, platform, host)` (reserved.ts) is false for a browser-reserved chord in a browser, and the display hooks (`useShortcutLabel`, new `useDisplayedBinding`) return null for it. That covers the × tooltip, the picker's kbd and the two save labels; the settings page (#789) reads the raw keymap and says why instead. Tests in shortcuts-conflicts.

packages/web/src/lib/shortcuts/match.ts
hiyouga: Fixed in 5ec02126b: `relocateChord` looks for the default's US character (`usCharacter`, letters, digits and the punctuation codes) across every code in the layout map, and `keyLabel` reads the layout for any code first. Test with a US-Dvorak-shaped map: Mod+KeyS → Semicolon, Mod+KeyW → Comma, Mod+KeyP → KeyR, and `keyLabel("Semicolon", …)` reads S. The "Dvorak – QWERTY ⌘" caveat is in the function's comment as unhandled.

packages/web/src/lib/shortcuts/reserved.ts
hiyouga: Fixed in 5ec02126b against the 43.2 role table: Windows loses Ctrl+Q, Linux redo is Ctrl+Shift+Z, zoom in is Mod+Shift+= on all three, macOS gains ⌥⇧⌘V. `SHELL_BY_PLATFORM` is empty on every platform with a comment that main's shell binds nothing on `before-input-event`; Ctrl+Shift+I stays a menu entry. The conflicts test pins each of these.

changelog/unreleased/2026-09-18-keyboard-shortcuts.md
hiyouga: Fixed in 5ec02126b: double-backtick spans for the three chords in each file, "in the desktop app" for the ⌘W sentence, and "save on ⌘S on a Mac and Ctrl+S elsewhere, instead of accepting both everywhere".

packages/web/src/lib/strings.ts
hiyouga: Fixed in 5ec02126b: the `shortcuts` block sits above the terminal section's doc comment in both dictionaries.

packages/web/src/features/terminal/terminal-view.tsx
hiyouga: Fixed in 5ec02126b: the terminal comment no longer claims no window listener is involved, and the store's `storage` comment says the event never fires in the writing tab.

packages/web/src/features/terminal/terminal-view.tsx
hiyouga: Done in 5ec02126b: `lib/shortcuts/terminal-keys.ts` exports `terminalKeyAction(e, keymap, platform, { canClose, hasHandler })` → `close | consume | skip-shell | shell`, and `test/terminal-keys.test.ts` is the table: ⌘W closes on a Mac, Ctrl+W is the shell's there, Ctrl+W closes elsewhere, the toggle skips the shell, Ctrl+P and an editor chord reach the shell, no close on offer falls through, a held close is consumed.

packages/web/src/lib/shortcuts/dispatcher.ts
hiyouga: Done in 5ec02126b: `handleShortcutKeydown` is exported and registered once with `window.addEventListener`; `installShortcutDispatcher`, `KeydownTarget`, the WeakSet and the install-once case are gone, and the test calls the listener with its fake event.

packages/web/src/lib/shortcuts/match.ts
hiyouga: Done in 5ec02126b: `relocateDefaults`, `isCommandId` and `rebindable` are deleted (with their test cases); #789 and #790 did not use them, and the merge into those branches drops the field from #790's new registry rows.

PR #788, opened 2026-09-18
fix(desktop,web): the shell opens no blank window, and Penguin Go authorization opens its page directly there

hiyouga:
Design PR: Prism-Shadow/penguin-harness-design#192 (amends #182's window-open clause).

## The gap

Since #716 the desktop main window allowed exactly one window outside the #780 rule (`classifyWindowOpen`): the hidden, sandboxed `about:blank` window the Web App opens synchronously in the Penguin Go **Authorize** click, to be navigated to the authorization URL once `/start` returns. Electron's `setWindowOpenHandler` does not say which frame asked. The Files panel previews Agent-written HTML in an iframe that allows popups, so that HTML could call `window.open('about:blank')` and receive the same hidden window — and, because `about:blank` inherits its opener's origin, script it, invisibly, for as long as it never navigated off this instance.

Review found a second route to the same window, present since #780: Electron builds a new window's options as `{ show: true, width, height, ...featuresFromThePage, ...override }` (`lib/browser/guest-window-manager.ts`), and the page's feature string may set `show`, `skipTaskbar`, `opacity`, `transparent`, `focusable`, a position and size limits (`allowedWindowOptions` in `lib/browser/parse-features-string.ts`). A preview page opening `location.href` with `show=no,skipTaskbar=yes` got the window the rule allows for `/preview/`, hidden, and same-origin with the preview.

## The fix

**No bridge.** Two candidates were considered: arming the bridge through a one-shot call from the App (with a TTL), or dropping the bridge on desktop. The second is smaller and holds up: the shell has no preload and no IPC into the page (by design, see `main.ts`), Electron has no popup blocker (`-add-new-contents` ignores `_userGesture`), so a `window.open(url, "_blank", "noopener,noreferrer")` after the `await` reaches the handler like any request and is routed `external`. The dialog's own "Open again" button already does exactly this.

**Desktop.** `isAuthorizationBridgeUrl` and the main window's allow branch are gone; every window, the main one included, answers with `openWindowFor` → `classifyWindowOpen`, which refuses `about:blank` with every other non-web scheme. New `APP_WINDOW_OPTIONS` in `util.ts` pins every feature-string key that can hide, dim or shrink a window (`show`, `skipTaskbar`, `opacity`, `transparent`, `focusable`, `hiddenInMissionControl`, `enableLargerThanScreen`, size and its limits — `maxWidth: 0` is Electron's "no maximum"); the one `allow` spreads it last. `guardOpenedWindow` centers the child (a position is not an option to pin: `left`/`top` land as `x`/`y`) and refuses the page's `moveTo`/`resizeTo` via `content-bounds-updated`. No page of the App moves or resizes its window. Every other rule from #780 is untouched.

**Web.** New `lib/desktop-renderer.ts`: `isElectronRenderer(userAgent)` — `Electron/` in the user agent, which every window of the shell carries (it sets no user agent of its own) and no browser does. `PlatformKeyAuthDialog` uses it: in the shell's window it opens no blank tab and, once `/start` returns, opens `authorizeUrl` directly and enters `waiting`. The renderer decides, not the session: in attach mode (the app attaching to a server `penguin web` already runs) the shell's window signs in through the login page, so `isDesktopShellWindow` is false there and a session check would have left the first Authorize click doing nothing. Browsers — including one signed into a desktop-mode server, which has a popup blocker like any other — keep the blank-tab path unchanged.

## Tests

- `packages/desktop/test/main-window-open.test.ts` (source-level, in the style of `esm-cjs-banner.test.ts`, since `main.ts` imports Electron): exactly one `action: "allow"`, inside `openWindowFor`; both `setWindowOpenHandler` calls delegate to `openWindowFor`; the override spreads `APP_WINDOW_OPTIONS` last and spells out none of the pinned keys itself; `guardOpenedWindow` calls `child.center()` and blocks `content-bounds-updated`. Fails on main on every count.
- `packages/desktop/test/util.test.ts`: `about:blank`, `about:blank#other`, `about:srcdoc` → `deny` for every window; an https authorization URL → `external`; `APP_WINDOW_OPTIONS` pins `show: true`, `skipTaskbar: false`, `opacity: 1`, `transparent: false`, `focusable: true`, the Mission Control and larger-than-screen flags, a positive minimum size and the unbounded maximum.
- `packages/web/test/desktop-renderer.test.ts`: `isElectronRenderer` on an Electron UA, Chrome/Firefox/Safari UAs, the empty string, Node's, and two near-misses (`ElectronNotes/`, lower-case `electron/`).

## Verification

Run on vps3 (the local box only edits and runs git). Results in the comments below.

## Not verified

No Electron app can be launched on vps3, so the shell flow (Authorize click → system browser in both embedded and attach mode; a preview popup with `show=no` appearing as a normal centered window; `about:blank` refused and logged) is not manually checked here. The refusal log line reads `[shell] refused to open a window for a about: URL`.


hiyouga:
## Verification (vps3, head 9cec12fb8)

Run through `vps3-vitest.sh` on a fresh remote checkout: `pnpm install --frozen-lockfile`, then

| Step | Result |
| --- | --- |
| `pnpm -r build` | ok |
| `pnpm --filter @prismshadow/penguin-desktop test` | 17 files, 227 tests passed |
| `pnpm --filter @prismshadow/penguin-web test` | 187 files, 2440 tests passed |
| `pnpm typecheck` | ok |
| `pnpm lint` | 0 warnings, 0 errors |
| `pnpm format:check` | ok |

Locally only `prettier --write` on the touched files and `git diff --check` (clean).

Not verified: no Electron app can be launched on vps3, so the Authorize click in the shell window and a preview popup being refused were not exercised by hand.

packages/desktop/src/main.ts
```diff
@@ -151,28 +150,11 @@ function createWindow(url: string): void {
...
-    }
-    return openWindowFor(target, iconPath);
-  });
-  win.webContents.on("did-create-window", (child, details) =>
-    guardOpenedWindow(child, iconPath, isAuthorizationBridgeUrl(details.url)),
-  );
+  // This window gets no exception of its own: the handler is not told which frame asked, and
+  // the Files panel previews Agent-written HTML in an iframe that allows popups, so any window
+  // allowed here is one that HTML can open too — a hidden one, it could own outright.
+  win.webContents.setWindowOpenHandler(({ url: target }) => openWindowFor(target, iconPath));
```
hiyouga, line 157: **[bug]** Every request now goes through `openWindowFor`, and its `"window"` branch still gives Agent-written HTML a hidden window it can script. That is the capability this PR exists to remove.

Electron 43 builds the child from `{ show: true, width: 800, height: 600, ...parseFeatures(features), ...overrideBrowserWindowOptions }` (`lib/browser/guest-window-manager.ts` at v43.2.0). `parseFeatures` takes `show`, `skipTaskbar`, `opacity`, `left`/`top` and `focusable` straight from the page's feature string (`allowedWindowOptions` in `lib/browser/parse-features-string.ts`). The override in `openWindowFor` (lines 179-187) sets none of them.

Scenario: an HTML file previewed in the Files panel runs the line below. (The isolated iframe is `allow-scripts allow-same-origin allow-popups` on `http://127.0.0.1:<port>/preview/…`.)

```js
const w = window.open(location.href, "_blank", "show=no,skipTaskbar=yes");
```

The URL is a preview page, so `classifyWindowOpen` answers `"window"`. The child is created with `show: false, skipTaskbar: true`. It is same-origin with the preview, because the popup inherits `allow-same-origin`. It lives as long as the main window's page, and close-to-tray only hides that page. This is the old bridge window exactly: that one also inherited the preview's origin. `opacity=0`, or `left`/`top` far off-screen, does the same to a window that is nominally shown. The `/preview/` rule predates this PR (#780), but the comment above now states the invariant, and the changelog and design #192 rest on it.

Fix: pin what hides a window in the `"window"` override: `show: true`, `skipTaskbar: false`, `opacity: 1`. In `guardOpenedWindow`, call `child.center()` and `preventDefault()` on `content-bounds-updated` (Electron applies `moveTo`/`resizeTo` by default). Then neither a `left`/`top` feature nor a later `moveTo` can park the window off-screen. If a pure helper in `util.ts` builds the override, a unit test can pin those keys.

hiyouga (commented):
Independent review of head 9cec12fb8, with design #192.

Checked and fine: Electron 43 has no popup blocker. `-add-new-contents` receives `_userGesture` and ignores it (`lib/browser/api/web-contents.ts` at v43.2.0), so the post-`await` `window.open` always reaches `setWindowOpenHandler`. `noopener` suits the dialog, which polls by `flowId` and never uses the handle. `isDesktopShellWindow` is false for a browser signed into a desktop-mode server (`sessionVia` is `password` there). `about:srcdoc`, `data:`, `blob:` and `javascript:` all fall to `deny` through `isExternalScheme`. Every `window.open` / `target="_blank"` the App itself issues (dock terminal, TokenDance, terminal links, the anchors) carries no feature string and routes as intended. Design #192 matches the code apart from finding 2. No i18n strings changed; the changelog pair mirrors section for section.

Findings (details inline):

1. **[bug]** High. A Files-panel preview can still get a hidden window it scripts, through `/preview/` with `show=no` instead of `about:blank` (`main.ts:156`).
2. **[bug]** Medium. Attach mode lost one-click authorization. There the shell's window is a password session, so it still asks for `about:blank`, which the shell now refuses (`platform-key-auth-dialog.tsx:47`).
3. **[test]** Low. The new desktop case passes on main unchanged, and nothing pins the `main.ts` change (`util.test.ts:147`).
4. **[simplify]** Low. `lib/authorization-window.ts` and its test re-wrap and re-test `isDesktopShellWindow` (`authorization-window.ts:24`).

Verdict: fix before merge: 1, 2

packages/web/src/features/models/platform-key-auth-dialog.tsx
```diff
@@ -37,6 +39,12 @@ export function PlatformKeyAuthDialog({
   onClose: () => void;
   onApplied: (applied: number) => void;
 }) {
+  const { desktopMode, sessionVia } = useAuth();
+  // A browser needs a tab opened inside the click, before the server has the URL, or its
+  // popup blocker eats the navigation. The desktop shell has no popup blocker and refuses
+  // every blank window (see lib/authorization-window), so there the URL is opened once it is
+  // known and the shell hands it to the system browser.
+  const bridge = usesAuthorizationBridge({ desktopMode, sessionVia });
```
hiyouga: **[bug]** This checks the session, but whether a blank tab can open depends on the renderer. In attach mode the shell's own window is not a desktop session.

Scenario: `penguin web` (or `pnpm dev:server`) already holds the data root. The release app shares `~/.penguin/data` with the CLI, so this is an ordinary setup. The app starts in attach mode (`main.ts:432-439`), and its window signs in through the login page, so `sessionVia` is `"password"` and `desktopMode` is false. `bridge` is true, and `window.open("about:blank", "_blank")` reaches the shell, which now refuses it (`[shell] refused to open a window for a about: URL`). `authorizationTab` is null, so once `/start` returns the dialog goes back to `ready` (line 164). The first Authorize click opens nothing and shows nothing; only a second click reaches the browser.

On main, the main window's allow branch served the bridge here too, so this is a regression. It also makes three claims false: the changelog's "the Web App does not ask for one inside the shell", the new `util.ts` comment ("Nothing of this app opens one") and design #192's rule.

Fix: decide by the renderer, not the session. Every shell window's `navigator.userAgent` carries `Electron/` (the shell sets no user agent of its own), and no browser's does. Otherwise, keep this check, accept two clicks in attach mode, and correct those three places.

packages/desktop/test/util.test.ts
```diff
@@ -145,20 +144,20 @@ describe("classifyWindowOpen", () => {
...
-  });
-});
-
-describe("isAuthorizationBridgeUrl", () => {
-  it("accepts only the inert blank window used while authorization starts", () => {
-    expect(isAuthorizationBridgeUrl("about:blank")).toBe(true);
-    expect(isAuthorizationBridgeUrl("about:blank#other")).toBe(false);
-    expect(isAuthorizationBridgeUrl("about:srcdoc")).toBe(false);
-    expect(isAuthorizationBridgeUrl("https://example.com")).toBe(false);
+  it("refuses a blank window from every window, the main window included", () => {
```
hiyouga, line 148: **[test]** This case passes on main unchanged:

- Main already asserted that `classifyWindowOpen("about:blank", origin)` is `deny`, in the case this one replaces.
- `about:blank#other` and `about:srcdoc` fail `isExternalScheme` like any other scheme.
- https → `external` is the "sends other sites … to the system browser" case above.

The change the comment describes is removing the main window's allow branch from `main.ts`. No test pins that, so a new exception added there would still pass the suite.

Fix: assert it from the source; `esm-cjs-banner.test.ts` already reads sources in this package. For example, assert that `src/main.ts` holds exactly one `action: "allow"` (inside `openWindowFor`), and that both `setWindowOpenHandler` calls delegate to `openWindowFor`. That fails on main. Otherwise, drop the case and leave the claim to review, as the PR body already does.

packages/web/src/lib/authorization-window.ts
```diff
@@ -0,0 +1,26 @@
...
+ * for one there.
+ *
+ * Both halves of isDesktopShellWindow are required: a browser signed into a desktop-mode
+ * server over loopback has a popup blocker like any other browser and keeps the blank tab.
+ */
+import type { AccountMenuSession } from "./account-menu";
+import { isDesktopShellWindow } from "./account-menu";
+
+/** Whether authorization opens a blank tab first and navigates it once the URL is known. */
+export function usesAuthorizationBridge(session: AccountMenuSession): boolean {
```
hiyouga: **[simplify]** This is a named negation of `isDesktopShellWindow` under a 19-line header. Its test re-runs the `desktopMode` × `sessionVia` combinations that `account-menu.test.ts:32-35` already pins. `workspace-browser.tsx:499`, `appearance-section.tsx:47` and `use-tray-locale.ts:35` call `isDesktopShellWindow` directly.

Fix: call it in the dialog, and delete this file and `test/authorization-window.test.ts`. If the fix on `platform-key-auth-dialog.tsx:47` switches to a renderer check, that check is the one worth a helper and a test.

hiyouga:
## Verification, round 2 (vps3, head 93f778074)

Same remote checkout as round 1 (full `pnpm -r build` there in round 1; core/hmr/server rebuilt by the runner each time). An intermediate commit (2d077c321) failed `pnpm typecheck` — `content-bounds-updated` was attached to the `BrowserWindow`, which has no such event; it is a `WebContents` event — fixed in 93f778074, then:

| Step | Result |
| --- | --- |
| `pnpm --filter @prismshadow/penguin-desktop test` | 18 files, 233 tests passed (new: `main-window-open.test.ts`, two `APP_WINDOW_OPTIONS` cases) |
| `pnpm --filter @prismshadow/penguin-web test` | 187 files, 2441 tests passed (`authorization-window.test.ts` removed, `desktop-renderer.test.ts` added) |
| `pnpm typecheck` | ok |
| `pnpm lint` | 0 warnings, 0 errors |
| `pnpm format:check` | ok |

Locally only `prettier --write` on the touched files and `git diff --check` (clean).

Still not verified by hand: no Electron on vps3 (see "Not verified" in the description).

packages/desktop/src/main.ts
hiyouga: Fixed in 2d077c321 (+ 93f778074 for the event's owner). `APP_WINDOW_OPTIONS` (util.ts) pins `show: true`, `skipTaskbar: false`, `opacity: 1`, `transparent: false`, `focusable: true`, `hiddenInMissionControl: false`, `enableLargerThanScreen: false`, the size and its limits (`minWidth/minHeight` 320×240; `maxWidth/maxHeight: 0` = Electron's no-maximum, so a `maxWidth=1` feature cannot clamp the window either); the one `allow` spreads it last. `guardOpenedWindow` calls `child.center()` and `preventDefault()`s `content-bounds-updated` on `child.webContents` (it is a WebContents event — the first attempt put it on the window and typecheck caught it); no page of the App uses `moveTo`/`resizeTo`. There is no other override path — the bridge branch was the only one. Pinned by `util.test.ts` (values) and `main-window-open.test.ts` (the override spreads the constant and spells out none of the keys itself; the guard centers and blocks). Changelog and design #192 (556280c) state the visibility rule.

packages/web/src/features/models/platform-key-auth-dialog.tsx
hiyouga: Fixed in 2d077c321. The dialog decides by the renderer: `isElectronRenderer(navigator.userAgent)` in new `lib/desktop-renderer.ts` (`/\bElectron\//`). Verified the premise: nothing in packages/desktop calls `setUserAgent` or sets `userAgentFallback`, so Electron's default UA with `Electron/43.x` stands in every shell window, attach mode included. The three texts are corrected: the changelog pair, the `util.ts` comment (now 'when it is drawn by this shell's renderer … whatever session it holds'), and design #192 (556280c: 凭渲染器判定 … 与登录方式无关，接入模式下壳窗口经登录页登录). Test: `web/test/desktop-renderer.test.ts`.

packages/desktop/test/util.test.ts
hiyouga: Fixed in 2d077c321 (+ 93f778074): `packages/desktop/test/main-window-open.test.ts` reads `src/main.ts` and asserts exactly one `action: "allow"` inside `openWindowFor`, exactly two `setWindowOpenHandler` calls both of the form `(({ url: target }) => openWindowFor(target, iconPath))`, the override spreading `APP_WINDOW_OPTIONS` last with none of `show`/`skipTaskbar`/`opacity`/`width`/`height` spelled out, and the guard's `center()` + `webContents.on("content-bounds-updated", …preventDefault)`. Fails on main on each count. The `util.test.ts` case's comment now points at it instead of claiming to pin the main.ts change.

packages/web/src/lib/authorization-window.ts
hiyouga: Done in 2d077c321: `lib/authorization-window.ts` and its test are deleted. The renderer check from the thread above (`lib/desktop-renderer.ts`, `isElectronRenderer`) is the helper that stays, with its own test.

PR #789, opened 2026-09-18
feat(web,server): a Shortcuts settings page, with bindings stored per account

hiyouga:
## Summary

Design: penguin-harness-design [#190](https://github.com/Prism-Shadow/penguin-harness-design/pull/190). **Stacks on #787** (PR A): this branch is cut from A's tip, so its diff against `main` includes A until A merges; after that, `origin/main` is merged in (no rebase). Review the commits after `5117be61d`.

- **Settings page** (`features/settings/shortcuts-section.tsx`): a new `shortcuts` page in the personal group after Appearance, listing every registry command by group as `PrefRow`s. Each row's control is a `ShortcutRecorder`; an overridden row also shows a flat restore-default icon button; the trailing action row holds "Reset all" (disabled when nothing is overridden, confirmed through `ConfirmModal` when more than one row is). The pane's "?" (`S.settings.shortcutsInfo`) states the platform rule, what live apply covers (tabs of the same browser now; other browsers and the desktop app on their next load), per-account and per-platform storage, and the browser-reserved caveat.
- **Row hints**, one line in priority order: a same-scope conflict (both rows), a shadowed conflict (the global row only — the focus-scoped command wins), a browser-reserved chord (browser host only: "只在桌面应用里生效"), a chord the desktop menu also carries (desktop host only; the binding takes it over). Conflicts use the new `PrefRow.hintTone="attention"` (`toneInk.attention`); reserved notices stay muted.
- **Recorder** (`lib/shortcuts/recorder.ts` pure + `features/settings/shortcut-recorder.tsx`): `idle → recording → idle`. While recording a window **capture** keydown listener with `stopPropagation` feeds the state machine, so Esc cancels without also closing the dialog; a held modifier previews as `⌘…` / `Ctrl+…`; Backspace/Delete commit an unbind; a key without a real modifier (Mod, literal Ctrl, Alt off a Mac — Shift alone and ⌥ alone on macOS are typing) is refused with `S.shortcuts.needsModifier` unless it is an F key; blur or a pointer-down outside cancels.
- **Per-account storage**: `UiPrefs.keybindings?: StoredKeybindings`; `services/keybindings.ts` validates on `PUT /api/me/prefs` (`v === 1`; only `mac|windows|linux`; ≤ 64 entries per section; id `^[a-z][a-z0-9]*(\.[a-z][a-z0-9]*)+$` ≤ 64 chars; chord `null` or grammar-conforming ≤ 48 chars; whole document ≤ 8 KiB; otherwise `400 invalid_keybindings` and nothing stored). The store gains `setKeybindingsPersister` (installed by `ShortcutRuntime` once signed in, so nothing is sent before auth; installing or removing it resets the "edited this session" mark, so an edit under the previous account is never read as this account's) and `hydrateFromServer`: the server document replaces the mirror; an absent server copy keeps a mirror this account edited here (the persister already carried it) and otherwise clears it. A failed PUT toasts `S.shortcuts.saveFailed` and leaves the mirror.
- `Kbd` gains a `size` rung (`picker` = the dock picker's `text-[10px]`, `control` = `text-xs` for the settings control).
- Wiring: `lib/settings-sections.ts` (`shortcuts` after `appearance`), `settings-dialog.tsx` (keyboard glyph, label, info, render), `AppLayout` mounts `<ShortcutRuntime/>`, `control-size.test.ts` gains the two new dialog-body modules, `settings-sections.test.ts` expectations updated.

## Verification

All on vps3 via `vps3-vitest.sh` (the local box only edits and runs git):

- `pnpm --filter @prismshadow/penguin-web typecheck` and `pnpm --filter @prismshadow/penguin-server typecheck` — clean
- `pnpm lint` — 0 warnings, 0 errors
- `pnpm format:check` — clean
- `pnpm --filter @prismshadow/penguin-server exec vitest run test/keybindings.test.ts test/draft-shortcuts.test.ts` — 2 files, 17 tests passed
- `pnpm --filter @prismshadow/penguin-web test` — 197 files, 2527 tests passed (new: `shortcuts-recorder`, `shortcuts-section` (static render of the page against an injected keymap, both dictionaries, both hosts), and the store's hydrate/persister cases)
- `git diff --check` — clean

No browser check: Playwright is not available on vps3 through the helper, and dev servers do not run locally. The recorder's DOM wiring (capture listener, blur/pointer-down cancel) is therefore exercised only through the pure state machine's tests.

Follow-up commit `f885b218d`: the id grammar accepts a camelCase segment (`^[a-z][a-zA-Z0-9]*(\.[a-z][a-zA-Z0-9]*)+$`), since the registry ids `dock.toggleRight` / `dock.toggleBottom` that #790 adds are camelCase; the design document spelled a lowercase-only grammar beside those ids.

## Review fixes

Commit `da8603501` answers [the review](https://github.com/Prism-Shadow/penguin-harness/pull/789#pullrequestreview-5254314238): the account-boundary reset and the no-repush hydrate (two-account test), the real-modifier rule in the recorder, the live-apply wording, and the cleanups (shell hint and strings removed, `HydrateOutcome` and a try/catch gone, `StoredKeybindings` declared once, zh menu hint). Design #190 `3c026c2` carries the two spec sentences.


packages/web/src/lib/shortcuts/store.ts
```diff
@@ -0,0 +1,295 @@
...
+ * or nothing was ever stored for this account, in which case the mirror is cleared, so a browser
+ * that signs into another account cannot resurrect the previous account's bindings.
+ */
+export function hydrateFromServer(stored: unknown): HydrateOutcome {
+  if (stored !== undefined && stored !== null) {
+    storeMirror(compactDoc(sanitizeStored(stored)));
+    invalidate();
+    return "applied";
+  }
+  if (writtenThisSession) {
```
hiyouga: **[bug]** Any edit since the page loaded sets `writtenThisSession`, and nothing ever resets it, but sign-out is client-side: `user-menu.tsx:147` runs `logout().then(() => navigate("/login"))` with no reload, and a 401 goes through the same `setUser(null)`. The sequence:

1. Account A changes a binding, then signs out.
2. Account B signs in on the same tab.
3. `ShortcutRuntime` installs B's persister and fetches B's prefs.
4. B has no `keybindings`, so this branch PUTs A's mirror into B's account.

From then on B works with A's bindings. This is exactly the shared-browser case that the design and the changelog (lines 27-29) say cannot happen.

Once the flag is scoped to one account, this branch has nothing left to send. The persister is installed before `getPrefs()` goes out, and the settings page only exists inside `AppLayout`, so the persister has already PUT every edit made under the current account. The store test "pushes the mirror when the server has no copy…" records the same document twice for this reason.

Fix: `setKeybindingsPersister` resets `writtenThisSession = false` (installing or removing a persister is an account boundary either way), and this branch keeps the mirror without calling the persister again. A test to pin it: edit under one persister, `setKeybindingsPersister(null)`, install a second persister, then `hydrateFromServer(undefined)`. Expect the mirror cleared, nothing sent to the second persister, and `editor.save` back on its default.

hiyouga (commented):
Independent review of #789's own commits (c97dcc2fe, a9bd64297, f885b218d and the merge d4033a8c9), diffed against `origin/feat/keyboard-shortcuts` at 5ec02126b, so nothing from #787 or its review is raised again. I checked it against S-design §3 and §7 B, design #190, the frontend skill, and the sign-in and sign-out flow on this branch. Details are inline.

**Fix before merge**
1. **[bug]** If a tab edits a binding, the next account that signs in on that tab gets those bindings pushed into its prefs: `writtenThisSession` survives the client-side sign-out, so the new account's hydrate takes the "pushed" branch. Inline at store.ts:144.
2. **[bug]** The recorder treats Shift (and ⌥ on macOS) as enough of a modifier. It accepts Shift+letter, Shift+Enter and Shift+Tab, and any of those, once bound, swallows typing in every input. Inline at recorder.ts:60.
3. **[copy]** "applies at once in every open tab of this account" (the page's "?" text in both dictionaries, and the changelog) promises more than the `storage` event delivers. The changelog also promises that the leak in 1 cannot happen, and it describes a "shell" hint that no chord can produce. Inline.

**Cleanups (can ride along)**
- **[simplify]** The `"shell"` branch of `rowHint` and `desktopShellReserved` in both dictionaries are unreachable, because `SHELL_BY_PLATFORM` is empty on every platform after #787's fix. Inline at shortcuts-section.tsx:65.
- **[over-defensive]** store.ts:148-152: the try/catch around `removeItem` guards nothing. `removeItem` does not throw, and the `localStorage` getter's SecurityError is already caught in `storage()` (store.ts:51-58).
- **[simplify]** `HydrateOutcome` has no production reader (`ShortcutRuntime` ignores the return value). The store tests already assert the mirror and the persister calls, and those are the behaviour.
- **[simplify]** `StoredKeybindings` is now declared twice, in `server/src/api/types.ts` and in `web/src/lib/shortcuts/types.ts`. The web already imports `UiPrefs` and its other prefs types from `@prismshadow/penguin-server/api`, so it can import this one too; the design says the server type is exported for the web.
- **[copy]** zh `desktopMenuReserved` 「会覆盖桌面应用菜单里的同名快捷键」 (strings.ts:115): the menu item shares the keys, not the name. Suggest 「会覆盖桌面应用菜单里使用同一组合的快捷键」. The en string already says "of the same keys".

**Checked, no finding.** *Server:* `PUT /api/me/prefs` sits under the module's `auth: "user"` and reads and writes only `c.var.user.userId`; in the route test, the second user sees `{}`. Validation runs before the read-merge-write, and there is no `await` between `prefsRepo.get` and `set`, so two concurrent PUTs cannot interleave inside the server. Each of these is a 400 `invalid_keybindings` that stores nothing: `null`, a non-object, `v !== 1`, a non-object section, more than 64 entries, an id outside the camelCase grammar or longer than 64 chars, a chord outside the grammar or longer than 48 chars, and a document over 8 KiB. Unknown sections are dropped. Unknown ids that fit the grammar are kept, and the client carries them through on write. The byte cap measured with `.length` is exact, because both regexes admit ASCII only. Every registry id in #787 to #790 passes the grammar, and every chord `chordOf` can produce fits the chord grammar and the 48-char cap, so the app cannot provoke a 400 by itself. *Store:* a server copy replaces the mirror and notifies; other tabs follow through the `storage` event; a failed PUT toasts, and the next load's hydrate restores the account's copy (the edit is lost there, as §3.4 decides). *Recorder:* the capture listener with `stopPropagation` runs before `Modal`'s bubble listener, so Esc cancels without closing the dialog. It also runs before the window dispatcher, so a recorded chord never runs its command as well. A repeat that lands after the listener is removed is prevented by the dispatcher (#787's fix). Blur and a pointer-down outside both cancel. A modifier-only press shows a preview. `isComposing`, and a Meta chord on Windows or Linux, commit nothing. Backspace and Delete unbind. A same-scope clash shows on both rows and a shadowed one on the global row only, which matches the matcher's resolution. The reserved hints read #787's corrected `reserved.ts` (Linux desktop: Ctrl+W is flagged as a menu chord). Reset-row and reset-all go through the same persister, and reset-all asks for confirmation only when more than one row is overridden. *UI:* the page sits in the personal group after Appearance. Its explanation lives in the pane's "?" (a title exists, so `InfoPopover`). The row hints stay visible as facts about the current state, and the conflict hint uses `toneInk.attention`. Both new modules are in `DIALOG_BODY_MODULES`, and the Reset all button is `sm`. Both dictionaries have the same keys and signatures.

Verdict: fix before merge: 1 (cross-account push in `hydrateFromServer`), 2 (recorder accepts Shift-only and ⌥-only chords), 3 (copy: "every open tab of this account" and the changelog's claims).

packages/web/src/lib/shortcuts/recorder.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (isModifierCode(code)) {
+    return { state: { phase: "recording", preview: modifiersOf(e, platform), notice: null } };
+  }
+  const bare = !e.ctrlKey && !e.metaKey && !e.altKey && !e.shiftKey;
+  if (bare && code === "Escape") return { state: RECORDER_IDLE };
+  if (bare && (code === "Backspace" || code === "Delete"))
+    return { state: RECORDER_IDLE, commit: null };
+  const chord = chordOf(e, platform);
+  if (chord === null) return { state }; // Meta on Windows/Linux belongs to the OS
+  if (!hasModifierOrFKey(chord)) {
```
hiyouga: **[bug]** `hasModifierOrFKey` counts Shift, so the recorder commits Shift+B, Shift+Enter, Shift+Tab or Shift+Backspace. On every layout, each of those is typing or editing. The dispatcher matches global chords inside inputs and prevents what it handles, so:

- with `terminal.toggle` bound to Shift+T, every capital T typed in the composer toggles the terminal and never appears;
- with Shift+Enter bound, the composer stops inserting newlines;
- `editor.save` bound to Shift+S eats every capital S in both editors.

The docstring above says a modifier-less chord is refused because it "would steal typing from every input", and S-design §2.7 says "a global command cannot steal typing from an input". A Shift-only chord defeats both. ⌥ alone on macOS is the same case, because ⌥+letter types characters: ⌥E is the accent key, and `@` is ⌥L on a German Mac.

Fix: accept a chord only when it has Mod or literal Ctrl, or an F key, or Alt on Windows/Linux (this keeps the design's suggested Alt+W rebinding possible). Shift counts only alongside one of those. `needsModifier` changes with it in both dictionaries, for example "Hold Ctrl or Alt (⌘ or ⌃ on macOS); Shift only together with them. Or use an F key". The registry test "no default is a bare key" should use the same predicate, and the recorder test gains "Shift+B on Linux → needsModifier". S-design §3.3's copy lists Shift as sufficient, so this needs your OK. In design #190, 「无修饰键的组合（F 键除外）不接受」 gains half a sentence: Shift, and ⌥ on macOS, does not count on its own.

packages/web/src/lib/strings-en.ts
```diff
@@ -140,6 +178,10 @@ export const en: Strings = {
     profile: "Profile",
     generalTitle: "General",
     appearanceTitle: "Appearance",
+    shortcutsTitle: "Keyboard shortcuts",
+    /** The Shortcuts page's "?": what follows the platform, where bindings live, and the browser's own claims. */
+    shortcutsInfo:
+      "Shortcuts follow the platform: ⌘ on macOS, Ctrl elsewhere. A change applies at once and in every open tab of this account; bindings are stored per account and per platform, so a Mac and a Windows machine each keep their own. A browser tab cannot receive the chords the browser itself reserves (Ctrl+W / ⌘W, for example); those work in the desktop app and are marked in the list.",
```
hiyouga: **[copy]** "A change applies at once and in every open tab of this account" claims more than the code does. Live apply rides the mirror's `storage` event, which only reaches tabs in the same browser profile. The desktop app (it has its own Electron storage), another browser and another machine keep the old binding until they next load and hydrate. Suggest: "A change applies at once in every tab of this browser; the account's other browsers and the desktop app pick it up the next time they load." The same claim appears in strings.ts:204 (「并同步到本账号已打开的全部标签页」), in the changelog (line 16, and zh line 10) and in design #190's 06 paragraph.

packages/web/src/features/settings/shortcuts-section.tsx
```diff
@@ -0,0 +1,160 @@
...
+      tone: "attention",
+    };
+  }
+  if (chord === null) return null;
+  const platform = currentPlatform();
+  if (currentHost() === "browser") {
+    return browserReserved(chord, platform) ? { text: S.shortcuts.browserReserved } : null;
+  }
+  const desktop = desktopReserved(chord, platform);
+  if (desktop === "shell") return { text: S.shortcuts.desktopShellReserved };
```
hiyouga: **[simplify]** This line is unreachable. `SHELL_BY_PLATFORM` has been empty on every platform since #787's fix, so `desktopReserved` never returns `"shell"`. Delete the line and `desktopShellReserved` in both dictionaries; the `"shell"` kind can come back with shell code that actually takes keys. The changelog promises this hint as well: line 22 says "a chord the shell binds" (zh line 14: 「壳自身绑定」). The same changelog sentence says a row explains when a binding "will not fire as bound", and the menu case contradicts that: a menu-reserved chord does fire, because the page takes it from the menu. Suggest "…or, in the desktop app, a chord its menu also carries (the binding takes it over)", and in zh 「或在桌面应用中，菜单同样带有的组合（绑定会接管它）」.

packages/web/src/lib/shortcuts/store.ts
hiyouga: Fixed in da8603501: `setKeybindingsPersister` resets `writtenThisSession` (installing or removing the persister is the account boundary), and the absent-server branch now keeps the mirror without calling the persister again. Two-account test added: edit under persister A, remove it, install persister B, hydrate with nothing — the mirror is cleared, B receives nothing, `editor.save` is back on its default. The changelog no longer claims the push, and says the previous account's edit is never read as this account's.

packages/web/src/lib/shortcuts/recorder.ts
hiyouga: Fixed in da8603501: `isBindableChord(chord, platform)` replaces `hasModifierOrFKey` — Mod, literal Ctrl, Alt off a Mac, or an F key; Shift alone and ⌥ alone on macOS are refused. `needsModifier` reworded in both dictionaries, the registry test uses the same predicate, and the recorder test gains Shift+B, Shift+Enter and ⌥E (mac) → needsModifier while Alt+W on Linux still commits. Design #190 says Shift and ⌥ alone do not count (3c026c2).

packages/web/src/lib/strings-en.ts
hiyouga: Fixed in da8603501: the "?" text in both dictionaries now says a change applies at once in every tab of this browser and that the account's other browsers and the desktop app pick it up on their next load; the changelog pair and design #190's 06 paragraph (3c026c2) say the same.

packages/web/src/features/settings/shortcuts-section.tsx
hiyouga: Done in da8603501: the `"shell"` branch and `desktopShellReserved` are gone from both dictionaries; the changelog line now says what stands in a binding's way and that a menu-reserved chord is taken over by the binding. The other cleanups ride along: the `removeItem` try/catch is gone, `hydrateFromServer` returns nothing (`HydrateOutcome` deleted), `StoredKeybindings` is declared once in the server's api types and re-exported by `lib/shortcuts/types.ts`, and the zh menu hint reads 「会覆盖桌面应用菜单里使用同一组合的快捷键」.

PR #791, opened 2026-09-18
fix(server,web): an organization's approval mode reaches its open sessions, and their composer offers no always-ask

hiyouga:
Two follow-ups to #785 (design Prism-Shadow/penguin-harness-design#187).

Since #785 an organization's session never waits for a person, so its approval mode is the only
thing that decides what an employee may do. Two gaps remained:

1. The organization runtime copies the mode onto a desk or ticket session when it opens it
   (`triggers.ts`), and a session's approvals read its own row. Changing the mode in the
   organization settings therefore changed nothing for the desks and ticket sessions already
   open; only `penguin org desk renew` or a new ticket session picked it up.
2. The chat composer still offered `always-ask` on an organization's session, where since #785
   it means "deny every call that needs approval". The organization settings already leave it
   out.

The review added a third point: the company skills let the CEO write `org_config.toml`, mode
included, which the sync cannot see. Since approval mode is security policy, it is now the
board's alone (see "Review follow-up" below).

Design PR: https://github.com/Prism-Shadow/penguin-harness-design/pull/193

### Server: the change is applied where the config is written

- `OrganizationService.patch` is the one place an organization's `org_config.toml` is rewritten
  after creation. `PATCH /api/projects/:projectId/organizations/:orgId` (the settings dialog)
  goes through it, and the CLI has no command that writes the mode. It now notes whether
  `approval_mode` changed and, after the write and still under the organization's lock, calls
  `syncApprovalMode` (`runtime/organization/reconcile.ts`).
- `syncApprovalMode` writes the new mode onto every session the organization's files name: the
  current and previous desks in `desks.toml`, and every session a ticket's `sessions` lists.
  It skips archived rows and rows already on that mode. `SessionManager.entryApprove`'s
  `getMode` re-reads the row on every decision, so a session mid-run applies the new mode from
  its next approval decision. Nothing else is needed there.
- "Which sessions are the organization's" is the walk the reconcile pass already makes to stamp
  `client = "org"`. It is factored out of `syncCaches` into `ownedSessions` unchanged,
  including the Project guard against a hand-edited ledger that names another Project's
  session, so the stamp and the mode cannot disagree about it.
- Sessions are touched only when the mode actually changes. A PATCH that leaves it as it was
  (a rename, say) does not overwrite a mode someone set on one desk from its composer.

### Web: the composer's approval picker

- `features/chat/approval-mode.ts` (new) holds `APPROVAL_MODES` and
  `approvalModeChoices(client, current)`. For `client === "org"` (the stamp the server's denial
  reads) it lists `read-only`, `allow-all` and `deny-all`, as the organization settings do.
- `ChatInput` takes the list as a required `approvalModes` prop, so no surface can forget it.
  The chat page's main composer and the Agents panel's sub-session composer (which edits the
  parent session's mode) pass `approvalModeChoices(selected.client, selected.approvalMode)`.
  The draft view passes all four.

**The stored `always-ask` edge case.** An organization's session may already store
`always-ask`, set through the API or before this change. The picker then keeps listing it
while it is the current value, in its usual first position and ticked, so the open menu shows
the mode in effect rather than a list with nothing selected. The trigger button names it as it
always did. Once any other mode is picked, it drops out of the list and is not offered again.

### Review follow-up: the approval mode is the board's alone

The sync only sees writes that go through the settings PATCH, and `org_config.toml` had a
second documented writer. `company-employee` listed the file, mode included, as written by
"humans, the CEO", and `company-ceo` had the CEO change `org_config.toml` once the board
agreed. A mode written that way never reached an open session, and re-saving the same mode in
the settings could not repair it. Approval mode is security policy, which by the
settings-placement principle no Agent changes. So this takes the minimum fix (the settings are
the only writer) rather than a reconcile pass that remembers the last-applied mode:

- `company-employee`: the file table marks `approval_mode` as changed only by the board, in the
  organization's settings. "What you may not decide alone" says the mode belongs to the board,
  that an employee never edits it in `org_config.toml` (not even after a yes, since a hand edit
  reaches only sessions opened after it), and that an employee who thinks the mode should change
  asks the board in the all-hands channel ("Asking the board").
- `company-ceo`: the list of `org_config.toml` changes the CEO proposes and then makes excepts
  the approval mode, which the CEO only proposes.
- `handbook.ts`: the handbook index a new organization is created with says the same in its
  directory table, in both the English and the Chinese template.
- `agent-company` `2026.09.17.2` → `2026.09.18.1`. The reconcile pass reinstalls an employee's
  company plugins when they fall behind the library, so existing employees get the rule on
  their next pass.

## Existing data

- Sessions already on disk whose mode differs from their organization's stay as they are until
  the organization's mode is next changed. Nothing is migrated.
- An existing organization's `handbook/README.md` is not rewritten, so its directory table
  still lists `org_config.toml` as written by "people, CEO" without the exception. The updated
  skills, which every employee carries, state the rule explicitly. Whether to touch handbooks
  already on disk is left to the user; this PR does nothing to them.
- This changes no stored format and adds no compatibility code, so there is no
  backward-compatibility changelog file.

## Worth a look, and decisions for the user

- **A person hand-editing `org_config.toml` is still not synced.** After this PR no skill
  tells an employee to edit the mode, and the settings dialog is the documented way to change
  it. Covering hand edits would need the last-applied mode remembered per organization and
  compared on every reconcile pass. That was deliberately not taken.
- **Previous (renewed) desks are included.** They are the organization's by the ledger and not
  archived, which is the rule given.
- **Re-sending the same mode is a no-op for sessions**, so there is no "apply to every session
  again" action. To realign sessions that drifted, change the mode and change it back.
- **An already-open desk chat page shows the mode it last fetched** until it refetches the
  session. No server event carries a session's approval mode, and that is already true when
  another tab changes a mode. The server-side mode is right either way.

## Verification

On vps3 (Node 24) through `vps3-vitest.sh`: `pnpm install --frozen-lockfile`, then the core and
cli builds, the hmr build, `pnpm gen:ifaces` and the server build.

First round (head `b17b54965`):

- `pnpm --filter @prismshadow/penguin-server exec vitest run test/session-manager.test.ts
  test/organization-runtime.test.ts test/organizations-routes.test.ts
  test/organization-scenario.test.ts`: 4 files, 138 passed (session-manager 44,
  organization-runtime 79, organizations-routes 14, organization-scenario 1). The new case:
  after the organization's mode changes, the current and previous CEO desks, the HR desk and an
  open ticket session all report the new mode, and an archived ticket session keeps the old
  one. A PATCH that leaves the mode unchanged keeps a desk's own override, and the next real
  change replaces it.
- Red check: with the `syncApprovalMode` call disabled on the remote copy only, that case fails
  at the first assertion after the change (`['allow-all', …]` vs `['read-only', …]`). With the
  code restored, both approval-mode cases in the file pass.
- `pnpm --filter @prismshadow/penguin-web exec vitest run test/approval-mode.test.ts
  test/control-size.test.ts test/disclosure-anchor.test.ts test/icon-scale.test.ts
  test/stream-selection-menu.test.ts`: 5 files, 24 passed. `approval-mode.test.ts` is new
  (3 cases); the other four parse the JSX of the components touched here.
- `pnpm typecheck` (every package; `ifaces.json` unchanged), `pnpm lint` (0 warnings,
  0 errors), `pnpm format:check` clean.

Review follow-up (head `f066a0fbe`; `origin/main` was still `078a7aa83`, so merging it into
the branch was a no-op):

- Server: the same 4 files, 138 passed. No test pins the handbook's table text, so the template
  edit is covered by typecheck and review, not by an assertion.
- `pnpm --filter @prismshadow/penguin-core exec vitest run test/plugins.test.ts`: 21 passed.
- `pnpm --filter @prismshadow/penguin-docs exec vitest run test/skills-sync.test.ts`: 3 passed.
- `pnpm typecheck`, `pnpm lint` (0 warnings, 0 errors) and `pnpm format:check` clean.
- `node scripts/check-plugin-versions.mjs origin/main`, run in a temporary worktree of the vps3
  clone at `f066a0fbe`: "agent-company changed and carry new versions".
- Locally: `git diff --check` clean.


packages/server/src/runtime/organization/service.ts
```diff
@@ -669,8 +669,16 @@ export class OrganizationService {
         await this.validateModel(projectId, req.model);
         next.model = req.model;
       }
+      const modeChanged = next.approvalMode !== org.config.approvalMode;
```
hiyouga, line 672: **[bug]** Only a PATCH can detect the change, but `org_config.toml` has a second documented writer. `company-employee/SKILL.md:22` lists the file's approval mode as written by "humans, the CEO", and `company-ceo/SKILL.md:128` has the CEO change `org_config.toml` once the board confirms.

Scenario: the org is on the default `allow-all`, and the board asks for `read-only`. The CEO edits `approval_mode` with its file tools, which `allow-all` permits. The settings dialog now shows Read only, but every desk keeps `allow-all`, and so does every open ticket session. Desks are long-lived: they are renewed only on a workspace move or an explicit renew. Re-saving Read only in the settings cannot repair this, because `modeChanged` is false. The only way out is to switch to another mode and back. The CEO can only tighten the mode this way (under `read-only` or `deny-all` its file writes are denied), and tightening is the direction that matters.

Fix, at minimum: make the settings dialog the one place the approval mode is changed, and say so in those two skill lines (the CEO proposes; the board applies the change in the settings). Complete fix: have the reconcile pass apply any mode it has not applied yet, keeping the last-applied mode per organization in the org cache. That covers every writer, hand edits included.

hiyouga (commented):
Independent review of head b17b54965, with design #193.

Checked and fine:

- **Lock order.** Every path that opens a desk or ticket session loads the org inside the same `KeyedLocks` key as `patch`: `create`, `hire`, `desk`, `startTicket`, and the reconcile pass's provisioning and dispatch. None of them can open a session on the old mode after the sync. `patch` takes no `SessionManager` lock, awaits no run and does not re-enter the org lock, so it has no cycle with the pass or with runs.
- **Mid-run.** `entryApprove`'s `getMode` re-reads the row on every decision. Sub-sessions are judged through the parent entry's approve (the subagent fallback sink; panel messages go through the parent runtime), so the desk and ticket rows are the right ones to write.
- **Skips.** Archived rows are skipped, and a PATCH that leaves the mode unchanged touches no session. `ownedSessions` is a faithful extraction of the old `syncCaches` walk, Project guard included.
- **Test and design.** The server test runs on the real `SessionsRepo` and fails without the call. Design #193's two sentences match the code.
- **Composer.** A stored `always-ask` stays listed and ticked, the draft passes all four modes, and the sub-session composer gets the parent's list. No other approval picker exists. No i18n strings changed; the changelog pair mirrors section for section.

Finding (details inline):

1. **[bug]** Medium. A mode written to `org_config.toml` by anyone other than the settings dialog never reaches open sessions, and the skills tell the CEO to write it that way (`service.ts:672`).

Verdict: fix before merge: 1

packages/server/src/runtime/organization/service.ts
hiyouga: Fixed in f066a0fbe with the minimum fix, as decided: the organization's settings are now the only writer of the approval mode.

- `company-employee`: the file table marks `approval_mode` as changed only by the board, in the organization's settings. "What you may not decide alone" says the mode belongs to the board, that an employee never edits it in `org_config.toml` (not even after a yes, since a hand edit reaches only sessions opened after it), and that one who thinks it should change asks the board in the all-hands channel ("Asking the board").
- `company-ceo` line 128: the list of `org_config.toml` changes the CEO proposes and then makes excepts the approval mode, which the CEO only proposes.
- The handbook index template (`handbook.ts`, English and Chinese) says the same in its directory table. That was the only other text naming the mode as agent-editable; an existing organization's handbook is not rewritten, which is noted in the PR description for the user.
- `agent-company` 2026.09.17.2 → 2026.09.18.1; `node scripts/check-plugin-versions.mjs origin/main` passes. The reconcile pass brings employees up to the new version.
- Design Prism-Shadow/penguin-harness-design#193 gains the board-only sentence (f349cf8).

The reconcile "last-applied mode" machinery was not added. On vps3: core plugins 21, docs skills-sync 3, and session-manager, organization-runtime, organizations-routes and organization-scenario 138 all pass; typecheck, lint and format:check are clean.

PR #794, opened 2026-09-19
fix: a keyless model entry borrows an environment variable only for the vendor's own endpoint

hiyouga:
## Summary

A model-catalog row with no API key of its own used to fall back to the **server's environment variables** through AgentHub: its clients read `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `GEMINI_API_KEY` (and their `*_BASE_URL` siblings) whenever they are handed no key, whatever base URL they were pointed at. So a keyless row in a **gateway group** sent the user's own vendor key to a third party — found in the #786 review: one group speed test on OpenCode Go made 27 requests, 7 with the Anthropic key and 20 with the OpenAI key; the model dialog told a Messages-protocol gateway row to "leave empty to use the ANTHROPIC_API_KEY env var" once any Anthropic row had shown the variable set; and any Session on such a row did the same.

User decision (2026-09-18): gateway groups never read a vendor env key. A keyless gateway row asks for its own key: a clear error at use, no env hint in the dialog, and the test paths refuse too. **Breaking**, shipped on its own.

## The rule, as implemented

Defined from data, by the **destination** of the request rather than the group's label (`packages/core/src/state/model-catalog.ts`, `modelEnvFallback` / `resolveModelCredential`). A keyless entry may lean on the variable AgentHub's routed client reads when

- it has **no base URL** — the client's default endpoint is the vendor's own, or the `*_BASE_URL` the user set beside the key (that pairing is AgentHub's own and is left to it entirely);
- its base URL is one of that **vendor's own official endpoints** (`VENDOR_ENDPOINTS`: the SDK defaults plus the second official host where a vendor runs two; the catalog pins the DeepSeek and MiniMax rows this way).

Per the user's decision (2026-09-18, second round): **environment keys are for official endpoints only.** A row whose own base URL equals the `*_BASE_URL` variable's value gets no exception — an earlier revision of this PR had one; it is removed, with its tests.

Everything else is refused before any client exists: every gateway group's preset endpoint (TokenDance, OpenRouter, Fireworks AI, SiliconFlow, both Qwen gateways, OpenCode Go when #786 lands — its rows carry `opencode.ai` base URLs, so it needs no special handling), `custom`, `vllm` and user-created rows with their own endpoints, and a vendor row re-pointed at a proxy. The `custom` group's Atria Dawn Preview preset, which used to read `ANTHROPIC_API_KEY`, now needs its own key.

**Custom / vLLM decision:** judged by the endpoint like everyone else. A custom row pointed at `api.openai.com/v1` keeps `OPENAI_API_KEY`; a private server does not, and `OPENAI_API_KEY` + `OPENAI_BASE_URL=<that server>` in the environment does not change that while the row carries its own base URL — the key goes on the row. A custom row with no base URL is left to AgentHub's own env pairing, as before.

**Penguin Go:** its provider-scoped `PENGUIN_GO_API_KEY`, which no AgentHub client reads, is read by the harness and passed explicitly for every row of the group; unset, the row is refused (the probes already did this via `assertRelayConnection`; Sessions did not, and would have let the client read `GEMINI_API_KEY`). A relay row with no base URL is refused whatever its key's source, so the relay key cannot travel to a vendor.

Where the fallback **is** allowed, the client is still handed no key and reads the variable itself — byte-for-byte the previous behaviour for vendor rows, including a Bedrock `ANTHROPIC_BASE_URL` with no `ANTHROPIC_API_KEY`.

## Where it is enforced

Harness side: every construction site applies core's `modelEnvFallback` — the entry paths through `resolveModelCredential`, the listing and protocol detection through `endpointEnvApiKey`:

- core: `Agent.createSession` / `resumeSession` (→ `model_credential_missing` on the server), the vision describer's `createLLM`, `listEndpointModels`;
- server: `testModel` (connection test **and** the group speed test), `detectVision`, `utilityCompletionConfig`, `detectProtocol` (relay key), `listEndpointModels`; `protocol-detect.ts`'s per-probe env fallback is gated on the probed URL being the vendor's own; `GET /models` reports `envKey` / `envKeyMasked` only where the fallback is allowed (`envFallbackFirstParty` replaced by the core rule);
- web: the dialog hint resolves through `envHintKeyFor` (group, id, protocol, base URL as drafted); the group-level "Set key" dialog names a variable only for vendor groups and Penguin Go (`providerEnvFallbackKey`).

**Preview vs. fallback (review finding 2).** `envKey` reports the rule; the masked `envKeyMasked` preview and the dialog hint read the narrower `modelEnvPreviewKey`: a row whose own base URL is a vendor endpoint, or a keyless row in a vendor group / Penguin Go. The eight vLLM presets and a custom row saved without a base URL do fall back under rule 1 but are not shown as "key configured" — that reading would list a self-hosted id in the chat picker and send the OpenAI key to api.openai.com. Server and web read the same function, and a web test pins their agreement.

**AgentHub follow-up (review findings 1 and 4) — this PR does not fix these, and an earlier revision of this description was wrong to say AgentHub needs no change.** Two leaks sit one layer down, in what the vendor SDKs read on their own, and the user decided they are fixed in AgentHub, in a separate PR being written now; they reach the harness through the next AgentHub release and bump:
1. `DeepSeekV4Client`, `GLM5_3Client` and `KimiK3Client` compute `options.apiKey || process.env.<VENDOR>_API_KEY || undefined` and hand the OpenAI SDK an undefined key, which the SDK fills from `OPENAI_API_KEY` — so a keyless `deepseek-flash` (the default model) with only `OPENAI_API_KEY` exported sends the OpenAI key to api.deepseek.com. No harness guard is added for this.
2. `Claude5Client` never passes `authToken`, so the Anthropic SDK attaches the environment's `ANTHROPIC_AUTH_TOKEN` beside a row's own key; and it passes neither `apiKey` nor `authToken` to `AnthropicBedrock`, which then sends `ANTHROPIC_API_KEY` to bedrock-runtime. Until that lands, a keyless `bedrock://` row on the entry is refused here with a message naming the way out (no base URL on the row plus `ANTHROPIC_BASE_URL=bedrock://<region>`), and the changelog lists it under what stops working.

## Design

Design PR: penguin-harness-design [#196](https://github.com/Prism-Shadow/penguin-harness-design/pull/196) (PRN-021 gets a destination-rule bullet; PRN-005's `api_key` comment, the 05 SDK comments, 06's credential bullet and the Atria preset note follow; one dated 08 line). It rewrites PRN-021's env-fallback clause; #191 (OpenCode Go) touches the same clause and is open — whichever lands second rebases onto the other: #191 adds OpenCode Go to the list of gateway groups and the per-row protocols, this PR adds the destination rule as a separate bullet after it, and the two compose without conflict in meaning.

## Compatibility

Anyone relying on `OPENAI_API_KEY` (or another vendor variable) for a gateway group, a self-hosted or custom server with an explicit base URL on the row (even with `OPENAI_BASE_URL` set to it), or a re-pointed vendor row now gets "Model `<provider>/<id>` has no API key … set the API key on the model entry" — put the key on the row (Model settings → API key, the group header's Set key, or `penguin config model add … --api-key`). Nothing on disk changes shape, no migration runs, no compatibility code, hence no backward-compatibility changelog file. Full text in `changelog/unreleased/2026-09-18-gateway-no-env-keys{,.zh}.md`.

## Verification

All on vps3 (`vps3-vitest.sh`, worktree synced from this branch at 83cbd393c, after the review fixes; `origin/main` was already contained); the local machine only edited files and ran git, prettier on touched files and `git diff --check`. No live API calls anywhere — every new test either refuses before a client exists or drives a local fake.

- builds: core, cli, hmr, `gen:ifaces`, server — ok
- core: `vitest run test/model-catalog.test.ts test/agent.test.ts test/plugins.test.ts` — 3 files, **117 passed**
- server: `vitest run test/models.test.ts test/protocol-detect.test.ts test/utility-completion.test.ts test/model-probe.test.ts test/session-index.test.ts test/models-list.test.ts test/models-vision.test.ts` — 7 files, **131 passed**
- web: `vitest run test/protocol-detect.test.ts test/models-input.test.ts test/model-grouping.test.ts` — 3 files, **103 passed**
- docs: `vitest run test/skills-sync.test.ts` — **3 passed** (the penguin-sdk skill changed; plugin `2026.09.18.1`)
- `pnpm typecheck` — ok; `pnpm lint` — 0 warnings, 0 errors; `pnpm format:check` — all files formatted

Coverage: every catalog row of every gateway group is refused a fallback and `resolveModelCredential` throws for it with vendor variables set; vendor presets still hand the client nothing; Penguin Go rows read `PENGUIN_GO_API_KEY`, are refused without it, and refused without a base URL even with a key; custom / vLLM / user-defined rows judged by endpoint, and `OPENAI_BASE_URL` naming the same server earns no exception; a keyless `bedrock://` row gets the Bedrock message and the no-base-URL form still works; `sameEndpoint` refuses the lookalike host, the userinfo trick and `http:`; `modelEnvPreviewKey` hides the vLLM presets and keyless custom rows without a base URL while keeping vendor / relay / vendor-endpoint rows; `Agent.createSession` on a keyless OpenRouter preset rejects with `OPENAI_API_KEY` stubbed and succeeds once the row has its own key; `POST /models/test` (plain and `speed: true`) on keyless TokenDance / OpenRouter / Atria presets and `/detect-vision` return `ok:false` with the harness's sentence (no vendor variable in it) and a fetch tripwire proving no request left the process; protocol detection probes a gateway URL anonymously even with `OPENAI_BASE_URL` set to it and authenticates each probe of a vendor's own URL only for that URL's protocol; the endpoint listing refuses a keyless gateway URL before constructing a client; the web dialog hint agrees with the server's preview rule over the vLLM presets and the other cases, and follows a base URL edit.


packages/core/src/state/model-catalog.ts
```diff
@@ -2504,6 +2510,206 @@ export function resolveProviderModelEnv(
...
+      `Model ${ref} has no base URL. Its API key belongs to the ${providerInfo(entry.provider)?.label ?? entry.provider} endpoint and cannot be sent to the vendor's default endpoint: set the base URL on the model entry.`,
+    );
+  }
+  if (apiKey !== undefined) return { apiKey, ...(baseUrl !== undefined ? { baseUrl } : {}) };
+  if (fallback === undefined) {
+    throw new ModelCredentialError(
+      `Model ${ref} has no API key. Its endpoint is not the vendor's own, so no environment variable is used for it: set the API key on the model entry.`,
+    );
+  }
+  if (fallback.readByClient) return baseUrl !== undefined ? { baseUrl } : {};
```
hiyouga, line 2696: **1. [bug] On the allowed path, `OPENAI_API_KEY` still reaches DeepSeek, Z.AI and Moonshot.** `readByClient` assumes the routed client reads only its own pair; lines 2658-2660 say the client "reads the pair itself … including the SDK's own error when the variable is unset". That is not what AgentHub's `DeepSeekV4Client`, `GLM5_3Client` and `KimiK3Client` do. They compute `key = options.apiKey || process.env.<VENDOR>_API_KEY || undefined`, then call `new OpenAI({ apiKey: key, baseURL })`. openai 6.42 fills an `undefined` key from `OPENAI_API_KEY`.

Probe with only `OPENAI_API_KEY` set:
- `deepseek/deepseek-flash`, the shipped default model: the harness credential is `{"baseUrl":"https://api.deepseek.com"}`, and the SDK holds `apiKey: "sk-openai-PROBE"`. That key is sent as `Authorization: Bearer` to api.deepseek.com.
- `zhipu/glm-5.3` and `moonshot/kimi-k3` send the same key to api.z.ai and api.moonshot.cn.

So on every new Project's default model, a user who exported `OPENAI_API_KEY` but not `DEEPSEEK_API_KEY` sends the OpenAI key to DeepSeek. The card reads "No key", and the chat guide nags once but never blocks, so the first sign is DeepSeek's 401. This predates the PR. It is still exactly what the user's rule forbids, on the path this PR vouches for.

Fix: in `resolveModelCredential`, refuse a `readByClient` fallback on `DEEPSEEK_API_KEY` / `ZAI_API_KEY` / `MOONSHOT_API_KEY` when the variable is unset. The third message already fits. The other clients do not need it: `MiniMaxM3Client` already throws, and the Anthropic and Gemini SDKs substitute that same vendor's own credentials. The alternative is to make those three AgentHub clients throw on a missing key, as MiniMax does. Either way, add a test: `deepseek-flash` with `OPENAI_API_KEY` set and `DEEPSEEK_API_KEY` unset throws `ModelCredentialError`.

packages/server/src/services/project-config-service.ts
```diff
@@ -1161,18 +1098,13 @@ export class ProjectConfigService implements ProjectConfigStore {
...
         // presence is implied by the field, the plaintext never leaves the server, and an
         // empty variable counts as absent — it would not authenticate either. Read from this
         // process's env, which on the desktop already includes the imported login-shell
         // variables.
-        const envValue =
-          envKey !== undefined &&
-          envFallbackFirstParty({ provider, modelId, clientType, baseUrl: credBaseUrl })
-            ? (process.env[envKey] ?? "")
-            : "";
+        const envValue = envKey !== undefined ? (process.env[envKey] ?? "") : "";
```
hiyouga: **2. [bug] The env preview now covers the eight vLLM presets.** On main, `envKeyMasked` was gated by `envFallbackFirstParty`, which excluded every group that pins a protocol, plus custom and the gateways. Its comment gave the reason: "steering it to … the user's own server is exactly the misconfiguration the preview must not encourage". `modelEnvFallback` allows any row with no base URL.

The vLLM presets are seeded into every Project with `client_type = "openai-chat-vllm-adapter"` and no `base_url` (`presetModelEntries`; `baseUrlRequired` is false for presets). The probe gives `vllm/Qwen/Qwen3.8-27B` the fallback `OPENAI_API_KEY` with `readByClient: true`, while this PR's own `providerEnvFallbackKey("vllm")` returns nothing.

With `OPENAI_API_KEY` exported (the desktop imports it), all eight rows now return `envKeyMasked`:
- the cards show the OpenAI key's mask as "read from environment variable";
- `hasConfiguredKey` turns true, so `visibleChatModels` lists them in the chat picker;
- a Session sends the OpenAI key and `Qwen/Qwen3.8-27B` to api.openai.com and gets model-not-found.

A custom row saved without a base URL does the same, for example one added with `penguin config model add --provider custom` and no `--base-url`.

Fix, without touching the refusal rule: report the preview only when `fallback !== undefined && (credBaseUrl !== undefined || providerEnvFallbackKey(provider) !== undefined)`.
- A row with no base URL is then previewed only in groups whose defaults point at the vendor's endpoint (vendor groups, Penguin Go), as on main.
- A custom row at `https://api.openai.com/v1` keeps its preview.

Apply the same condition in `envHintKeyFor`, add a vLLM preset to the preview test, and revert the `custom/my-model` expectation at `models.test.ts:248-249`.

packages/core/src/state/model-catalog.ts
```diff
@@ -2504,6 +2510,206 @@ export function resolveProviderModelEnv(
...
+ * The endpoints AgentHub's vendor clients talk to when handed no base URL, per credential
+ * variable: the OpenAI and Anthropic SDK defaults, Google's Generative Language host, the
+ * defaults the vendor-specific clients (deepseek_v4, glm5_3, kimi_k3, minimax_m3) carry, plus
+ * the second official host where a vendor runs two (Z.AI's mainland bigmodel.cn, Moonshot's
+ * international .ai, MiniMax's mainland minimaxi.com). A key taken from the environment is
+ * sent only to one of these — see modelEnvFallback.
+ */
+export const VENDOR_ENDPOINTS: Readonly<Record<string, readonly string[]>> = {
+  OPENAI_API_KEY: ["https://api.openai.com/v1"],
+  ANTHROPIC_API_KEY: ["https://api.anthropic.com"],
```
hiyouga, line 2523: **3. [bug] Keyless Bedrock rows are now refused, and the changelog does not say so.** The harness supports an entry-level `base_url = "bedrock://<region>"`: `fastModeProtocol` tests for it, and the docs list "Claude on Bedrock". When such a row has no `access,secret` key, `Claude5Client`'s `AnthropicBedrock` signs with the AWS default credential chain (SSO, instance roles, `AWS_*`), which is the usual way to run Bedrock.

The rule compares `bedrock://us-east-1` against `https://api.anthropic.com`. The probe gets "Model anthropic/claude-sonnet-5 has no API key. Its endpoint is not the vendor's own…". Main ran these rows. The only way to keep one working is to inline long-lived `access,secret` on the row, and that format cannot express SSO or role credentials.

Allowing `bedrock://` here is not enough on its own. `AnthropicBedrock` inherits the base client's `ANTHROPIC_API_KEY` default, so with that variable exported it sends `x-api-key: <Anthropic key>` to bedrock-runtime.<region>.amazonaws.com. The probe shows `[["x-api-key","sk-ant-PROBE"]]`; see 4 in the review body.

For this PR:
- List keyless `bedrock://` rows under "What stops working", with the way out: a row with no base URL plus `ANTHROPIC_BASE_URL=bedrock://<region>` still runs under rule 1.
- Give these rows a message that does not call Bedrock "not the vendor's own".

Allow `bedrock://` once AgentHub stops lending env keys to `AnthropicBedrock`.

changelog/unreleased/2026-09-18-gateway-no-env-keys.md
```diff
@@ -0,0 +1,67 @@
+# A keyless model entry borrows an environment variable only for the vendor's own endpoint
+
+- **Date:** 2026-09-18
+- **Type:** fix
+- **Scope:** `core`, `server`, `web`, `docs`
+- **PR:** [#794](https://github.com/Prism-Shadow/penguin-harness/pull/794)
+- **Breaking:** yes — a model entry with no `api_key` in a gateway group (TokenDance, OpenRouter, Fireworks AI, SiliconFlow, Qwen Pay-As-You-Go, Qwen Token Plan), in `custom`, `vllm` or a user-created group, or a vendor entry with a `base_url` of its own, no longer reads `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY` (or any vendor variable) from the server environment; Sessions, the connection test, the group speed test, the vision probe, protocol detection and the endpoint listing refuse it instead
```
hiyouga: **6. [copy] The Breaking line describes groups, but the rule is per destination.** Two kinds of row the line names still fall back:
- a custom, vLLM or user-created row with no `base_url`. That is rule 1, and the Compatibility section says so.
- a vendor row whose `base_url` is the vendor's own, such as `deepseek-flash`, or a Moonshot row at `https://api.moonshot.ai/v1` (the dialog tells platform.kimi.com users to set that URL).

Suggest: "a model entry with no `api_key` whose `base_url` is not the vendor's own official endpoint (every gateway preset; custom / vLLM / user-created rows with their own endpoint; a vendor row re-pointed at a proxy) no longer reads …". Same in the .zh.md.

Two more stale phrases:
- Line 18, and `model-catalog.ts:2578`: "an environment key goes only where the environment put it" is left over from the dropped `*_BASE_URL` clause. The rule is now "environment keys are for official endpoints only".
- Line 22, and `model-catalog.ts:2651-2655`: "the one place … every client … the endpoint listing" is not accurate. The listing and protocol detection go through `endpointEnvApiKey`, not `resolveModelCredential`; the one shared function is `modelEnvFallback`.

packages/server/test/models.test.ts
```diff
@@ -991,6 +1009,69 @@ describe("model-reference rekeying and the connectivity test", () => {
...
+        provider: "tokendance",
+        modelId: "glm-5.3",
+      });
+      expect((await vision.json()) as ModelVisionDetectResponse).toMatchObject({
+        outcome: "failed",
+        message: expect.stringMatching(/has no API key/) as string,
+      });
+      expect(requests).toBe(0);
+      // The refusal is not a 400 with the SDK's own text: the message never names a vendor variable.
+      expect(JSON.stringify(await (await api.get(url())).json())).not.toContain("must-not-cross");
```
hiyouga: **8. [test] This assertion cannot fail.** `GET /models` only ever carries `maskApiKey(value)`, which is at most `vend…ross`. So `must-not-cross` never appears in the body, whatever the code does. The comment above it also describes a different check, the refusal message. Either assert on the refusal bodies inside the loop (`expect(body.message).not.toMatch(/OPENAI_API_KEY|ANTHROPIC_API_KEY/)`) or drop the line.

Line 1028 says the Atria preset is covered, but the loop runs only tokendance and openrouter. Atria is the Messages-protocol path, the one that carried the Anthropic key in the #786 finding. Add `["custom", "Atria-Dawn-Preview"]` to the loop.

packages/core/test/model-catalog.test.ts
```diff
@@ -1753,3 +1761,208 @@ describe("off-peak schedules", () => {
...
+      "qwen-token-plan",
+      "vllm",
+      "custom",
+      "my-group",
+    ]) {
+      expect(providerEnvFallbackKey(id), id).toBeUndefined();
+    }
+  });
+
+  it("sameEndpoint ignores case in the host and trailing slashes, and nothing else", () => {
```
hiyouga, line 1941: **9. [test]** The title says "and nothing else", but the test pins none of the cases that make this a security check. Add these three, each expected false:
- `https://api.openai.com.evil.example/v1`
- `https://api.openai.com@evil.example/v1`
- `http://api.openai.com/v1`

The implementation already handles all three. The test is what keeps a later string-prefix "simplification" out.

packages/web/src/features/models/models-page.tsx
```diff
@@ -2841,20 +2843,26 @@ function ModelDialog({
...
+  const routedEnvKey = resolveProviderModelEnv(
     form.provider,
     form.modelId.trim(),
     envHintClientType(form.provider, form.clientType),
   )?.envKey;
+  // The variable a blank key may actually FALL BACK to for the entry as drafted (core's
+  // modelEnvFallback, the rule the server enforces): undefined for every row whose endpoint
+  // is not the vendor's own — a gateway's preset base URL, a custom or vLLM server, a vendor
+  // row re-pointed at a proxy. The hint follows this, never routedEnvKey.
+  const liveEnvKey = envHintKeyFor(form.provider, form.modelId, form.clientType, form.baseUrl);
```
hiyouga, line 2866: **10. [bug]** The placeholder now follows the drafted base URL, but the "Read from environment variable" block at :3245 does not, because it renders the stored `form.envKeyMasked`. To reproduce, edit a keyless Anthropic row and type a proxy base URL. The placeholder disappears, the env mask stays, and Test connection answers "has no API key". Gate that block on `liveEnvKey === form.envKey` as well.

packages/core/src/state/model-catalog.ts
```diff
@@ -2504,6 +2510,206 @@ export function resolveProviderModelEnv(
...
+ * entry yet. Same rule as modelEnvFallback: only a vendor's own endpoint gets the vendor's
+ * key; a gateway or a private server gets none.
+ */
+export function endpointEnvApiKey(
+  clientType: string,
+  baseUrl: string | undefined,
+  env: Readonly<Record<string, string | undefined>> = process.env,
+): string | undefined {
+  const fallback = modelEnvFallback({ provider: "custom", modelId: "", clientType, baseUrl });
+  if (fallback === undefined || !fallback.readByClient) return undefined;
```
hiyouga: **11. [over-defensive]** `|| !fallback.readByClient` can never be true. The provider is hard-coded to `custom`, and only the Penguin Go branch returns `readByClient: false`. Delete it.

hiyouga (commented):
Independent review of `8537acd82`, read against design Prism-Shadow/penguin-harness-design#196.

**Coverage.** Every client construction takes its credential from the new rule: all `new GenerativeModel` sites (agent.ts ×3, project-config-service.ts ×3) and both `new AutoLLMClient` sites (generative-model.ts, list-models.ts). Subagents, resume, company-mode, scheduled and evaluation Sessions all enter through `Agent.createSession` / `resumeSession`. The CLI reaches models only through the server, and no plugin reads a vendor key. `sameEndpoint` compares the WHATWG-parsed scheme, host and path, the same parse `fetch` uses, so lookalike hosts, userinfo tricks and `http:` do not pass. `VENDOR_ENDPOINTS` matches AgentHub 0.4.15's client defaults plus the two alternate hosts the dialog tells users to type.

The holes are one layer down, in credentials the vendor SDKs under AgentHub read on their own, and there are two regressions. I checked the SDK claims on vps3 with a probe that makes no network calls. It builds the clients the harness builds (core from this head, AgentHub 0.4.15, @anthropic-ai/sdk 0.91.1, openai 6.42.0) and reads the credential each SDK instance holds.

Findings 1, 2, 3, 6, 8, 9, 10 and 11 are inline comments. 4, 5 and 7 are below because they point at code outside the diff.

**4. [bug] AgentHub follow-up: `Claude5Client` lets the Anthropic SDK attach env credentials, keyed rows included.** Every Claude 4.6+/5 id, which means the whole Anthropic group, routes to `Claude5Client`. It never passes `authToken`, and it passes neither `apiKey` nor `authToken` to `AnthropicBedrock`. The SDK fills them from the environment:
- A keyed `claude-sonnet-5` row at `https://proxy.example/anthropic`, with `ANTHROPIC_AUTH_TOKEN` exported, sends `[["authorization","Bearer tok-env-PROBE"],["x-api-key","sk-row-KEY"]]`. The same row on `ant-messages`, which passes `authToken: key`, sends only its own key.
- A `bedrock://us-east-1` row with `ANTHROPIC_API_KEY` exported builds an `AnthropicBedrock` that sends `[["x-api-key","sk-ant-PROBE"]]` to bedrock-runtime.us-east-1.amazonaws.com.

`ANTHROPIC_AUTH_TOKEN` plus `ANTHROPIC_BASE_URL` is what Claude Code users export for third-party Anthropic-compatible endpoints, and the desktop app imports the login shell. Keyed rows never reach the harness's rule, so the fix belongs in AgentHub: pass `authToken` explicitly in `Claude5Client`, and `apiKey: null, authToken: null` to `AnthropicBedrock`. This does not block this PR, but the description's "AgentHub needs no change" is not accurate, and the gap should be tracked.

**5. [copy] The Import dialog still says to leave the key empty.** `groupImportKeyHint` (`packages/web/src/lib/strings-en.ts:978` and its zh counterpart at `strings.ts:964`) is the key placeholder at `models-page.tsx:1949`. It reads "Leave empty to read the protocol's OPENAI_* / ANTHROPIC_* env vars". After this PR, a keyless import of any URL that is not a vendor's own fails with "No API key for …". The docs step was updated but the UI was not. Drop the placeholder, or say the endpoint's key is required.

**7. [copy] Docs and comments that still promise the old fallback.**
- `packages/docs/content/models.{en,zh}.md:370-376` ("Add a custom entry") and the vLLM recipe at `:355` list only `client_type`, `base_url` and `model_id`. A keyless entry at a local server is now refused, and `OPENAI_API_KEY=dummy` no longer covers it. Add an `api_key` step: any placeholder works if the server does not check keys.
- `quickstart-cli.{en,zh}.md:51` says AgentHub reads the vendor variables whenever the entry has no `api_key`. Add "for the vendor's own endpoint".
- `quickstart-desktop.{en,zh}.md:53` says "Gateway and custom groups are not matched against these variables", and the `envKeyMasked` contract at `packages/server/src/api/types.ts:526-531` says gateway, custom and user-defined groups never carry it. After this PR both are false for custom and vLLM rows with no base URL. The fix in 2 makes them true again.
- These need the same vendor-endpoint qualifier:
  - the public SDK JSDoc at `packages/core/src/agent.ts:218,230` ("then to AgentHub reading environment variables");
  - `project-config-service.ts:1003-1004` ("the wrapped SDK reads the protocol's own variable");
  - `plugins/agent-development/skills/penguin-sdk/SKILL.md:55` ("Environment variables cover the credential only").
- `packages/core/src/llm/list-models.ts:46` says "The environment lends a key only to a vendor's own endpoint". When the URL is the vendor's own and the variable is simply unset, this blames the URL. Say the variable is unset.

**Design Prism-Shadow/penguin-harness-design#196.** PRN-021 states the rule the code implements, limited to custom / vLLM / user-group rows "with their own endpoint". The 06 credential bullet drops that qualifier: it says every gateway / custom / vLLM / user-group row without a key is unusable and gets no env hint. The eight vLLM presets ship with no base URL and do fall back under rule (1). Whichever way 2 is fixed, 06 should say "with their own endpoint", as PRN-021 does.

Verdict: fix before merge: 1, 2, 3, 5, 6, 7, 8 (4 needs an AgentHub follow-up and a corrected description; 9, 10 and 11 are small enough to fix in the same pass)

packages/core/src/state/model-catalog.ts
hiyouga: Confirmed by your probe; not fixed here by the user's decision. The DeepSeek / GLM / Kimi clients handing the OpenAI SDK an undefined key is fixed in AgentHub, in a separate AgentHub PR being written now, and reaches the harness through the next release and bump. No harness guard is added; the PR description and the changelog (83cbd393c) now name it as the AgentHub follow-up instead of claiming AgentHub needs no change.

packages/server/src/services/project-config-service.ts
hiyouga: Fixed in 83cbd393c: core's new modelEnvPreviewKey gates the masked preview — fallback defined and (the row has its own base URL, i.e. a vendor endpoint, or providerEnvFallbackKey names a variable for the group). GET /models reads it for envKeyMasked (envKey still reports the rule), envHintKeyFor reads it for the dialog hint. The preview test gains the vLLM preset Qwen/Qwen3.8-27B (envKey OPENAI_API_KEY, no mask) and custom/my-model is back to no mask; a web test pins envHintKeyFor against modelEnvPreviewKey over the vLLM presets and the other cases.

packages/core/src/state/model-catalog.ts
hiyouga: Fixed in 83cbd393c: a keyless bedrock:// row now gets its own message — a Bedrock endpoint on the entry cannot fall back to the environment yet; set the entry's AWS key (access,secret), or leave the base URL empty and set ANTHROPIC_BASE_URL=bedrock://<region>. The changelog's compatibility section lists keyless Bedrock rows with that workaround, and a core test pins both the message and that rule 1 keeps the no-base-URL form working. Allowing bedrock:// waits for the AgentHub fix to AnthropicBedrock (finding 4).

changelog/unreleased/2026-09-18-gateway-no-env-keys.md
hiyouga: Fixed in 83cbd393c: the Breaking line is now by destination — entries whose base_url is not the vendor's own official endpoint (gateway presets; custom / vLLM / user-created rows with their own endpoint; a vendor row re-pointed at a proxy or bedrock://), same in the .zh.md. The leftover 'where the environment put it' phrase is gone from line 18 and from model-catalog.ts; the Details bullet and the resolveModelCredential doc now say modelEnvFallback is the one rule, applied by resolveModelCredential for entries and by endpointEnvApiKey for the listing and protocol detection.

packages/server/test/models.test.ts
hiyouga: Fixed in 83cbd393c: the vacuous GET assertion is gone; the loop asserts on each refusal body (has no API key, set the API key on the model entry, and not OPENAI_API_KEY|ANTHROPIC_API_KEY|must-not-cross), and ["custom", "Atria-Dawn-Preview"] is in the loop.

packages/core/test/model-catalog.test.ts
hiyouga: Fixed in 83cbd393c: the three cases are pinned — api.openai.com.evil.example, api.openai.com@evil.example and http://api.openai.com/v1 — both on sameEndpoint and through modelEnvFallback for a custom openai-chat row.

packages/web/src/features/models/models-page.tsx
hiyouga: Fixed in 83cbd393c: the 'Read from environment variable' block is additionally gated on liveEnvKey === form.envKey, so typing a proxy base URL over a keyless vendor row hides the stored mask along with the placeholder.

packages/core/src/state/model-catalog.ts
hiyouga: Fixed in 83cbd393c: the unreachable || !fallback.readByClient is deleted.

hiyouga:
Review follow-up in 83cbd393c for the body-only findings: **4** — the Claude5Client / AnthropicBedrock env-credential attachment is, by the user's decision, fixed in AgentHub in a separate PR (with finding 1's DeepSeek/GLM/Kimi leak); the description here no longer claims AgentHub needs no change and names both leaks as the follow-up. **5** — groupImportKeyHint now says the endpoint's key is required unless the URL is the vendor's own (en/zh). **7** — models.{en,zh}.md self-hosted recipes gain an api_key step, quickstart-cli/desktop and the envKeyMasked DTO comment carry the vendor-endpoint qualifier, the SDK JSDoc on CreateSessionOptions/ResumeSessionOptions states the rule, project-config-service's listing doc and list-models' message (which now names the unset variable when the URL is the vendor's own) are corrected, and the penguin-sdk skill states it (agent-development plugin 2026.09.18.1). Design #196 restores 06's 'with their own endpoint' qualifier.

PR #808, opened 2026-09-20
feat(server,desktop,scripts): a plugin's interfaces are compared with the platform's by the TypeScript compiler

Myriad-Dreamin:
Stack: main ← #771 ← #769 ← **#808** ← #718 ← #544 ← #448 ← #450 ← #577 ← #803 ← #807 ← #809 ← #810 ← #614 ← #815 ← #613 ← #619 ← #623 ← #625 ← #626 ← #628 ← #642 ← #639 ← #641 ← #526 ← #726 ← #672 ← #719 ← #796 ← #798 ← #804 ← #806 ← #817 ← #818

Split out of #718 (workflows) at the owner's request: this part has nothing to do with workflows, and #718 now sits on top of it. Design: PRFC-0014「接口检查」([penguin-harness-design #194](https://github.com/Prism-Shadow/penguin-harness-design/pull/194)); the revision of PRFC-0001 it implies is still to be written.

## What

`checkTree` looks a required interface up in the platform's own table, so a plugin's view of an interface was compared with itself — a check that can never fail (plugin tables are merged in with `??=`, so the platform's entry always wins). Each side now brings its own table:

- `plugin/iface-check.ts` — `renderDts` writes a table as a self-contained `.d.ts` (lossless, or it refuses by name), `ifaceQuestions` turns a package's manifests into requires/provides questions, `checkIfaces` has **tsc** answer them both ways. An interface the package's table holds no copy of is reported as *uncompared*, never answered from the platform's entry. A type of the language's own library (`lib#Uint8Array`) renders as itself.
- `plugin/loader.ts` — a plugin package gets the comparison at load. A misfit refuses the plugin; *uncompared* and *no compiler here* are logged, not fatal (the generator does not copy host interfaces into plugin tables yet, so refusing on those would refuse every real plugin — measured on the chain's plugins).
- `plugin/typescript.ts` — the compiler is loaded on first need and resolved beside the module, beside the running program, or in the pushed assets. `typescript` moves to the server's runtime dependencies.
- `scripts/typescript-payload.mjs`, `scripts/deploy.mjs`, `packages/desktop/scripts/build-assets.mjs` — a hot push carries the compiler as `archives/typescript.tgz` (entry + default libs, 101 files); the desktop build stages it beside the server bundle.
- `scripts/gen-ifaces.mjs` — an optional `unknown`/`any` member stays that; `getNonNullableType(unknown)` is `{}`, which refuses `null`.

## Verification

On the test machine (Node 24), this branch alone: `pnpm install --frozen-lockfile`, `pnpm -r build`, `pnpm typecheck` (0 errors), full `pnpm -r test` — core 1223, server 150 files / 2126, web 2331, cli 440, desktop 216, all passed — `pnpm lint` (0/0), prettier clean on the changed files. `iface-check.test.ts`: both directions asked; a package compiled against this platform passes; a changed required and a changed provided interface are each named; a member the platform gained breaks nothing; a missing copy is *uncompared*; every interface of the platform's table renders. The desktop staging has not been exercised by a packaged build.


packages/server/src/plugin/iface-check.ts
```diff
@@ -0,0 +1,370 @@
...
+    if (decl === undefined) throw new IfaceRenderError(`'${key}' is not in the table`);
+    this.slice.ifaces[key] = decl;
+    this.out.push(this.interfaceText(name, key, decl));
+    return name;
+  }
+
+  text(): string {
+    return [
+      "declare const opaque: unique symbol;",
+      "type Opaque<Name extends string> = { readonly [opaque]: Name };",
```
Myriad-Dreamin: **(High) Two byte-identical tables fail this check.**

Each side's `.d.ts` gets its own `declare const opaque: unique symbol`, so `P.Opaque<"X">` and `C.Opaque<"X">` are different types: their brand keys are different symbols. Reproduced with the real compiler (typescript 5.9.3), both files holding the *same* declaration:

```
platform.d.ts / consumer.d.ts:
  declare const opaque: unique symbol;
  type Opaque<Name extends string> = { readonly [opaque]: Name };
  export interface Log { line: (a0: Opaque<"@p#Buf">) => void }

TS2322: Type 'platform.Log' is not assignable to type 'consumer.Log'.
  Property '[opaque]' is missing in type 'consumer.Opaque<"@p#Buf">'
  but required in type 'platform.Opaque<"@p#Buf">'.
```

The rule this replaces was name equality (`core/src/kernel/sig.ts:180`: `if ("opaque" in a && "opaque" in b) return a.opaque === b.opaque`), and opaques are pervasive in the host surface — in the current generated table, 43 of 208 interfaces carry a non-library opaque (75 occurrences; `@prismshadow/penguin-server#TerminalSession` ×8, `@types/node#Buffer` ×18, `SandboxProvider` ×4).

Today it is masked by the "nothing is actually compared" finding below. Once `gen-ifaces` copies host interfaces into a plugin's table, `claude-code` — which requires `Terminals` — fails to load against the very platform it was built against.

Suggested direction: brand by a plain string-keyed property, the way `core/src/kernel/markers.ts:14` already does (`{ readonly __opaque?: Name }`), so identity is by name across the two files.

packages/server/src/plugin/iface-check.ts
```diff
@@ -0,0 +1,370 @@
...
+  consumer: IfaceTable,
+  questions: readonly IfaceQuestion[],
+): IfaceCheckResult {
+  const problems: string[] = [];
+  const uncompared: string[] = [];
+  const label = (q: IfaceQuestion) => `${q.module}: ${q.direction}.${q.alias} '${q.key}'`;
+  const base = path.join(os.tmpdir(), "penguin-iface-check");
+  const files = new Map<string, string>();
+  const asked = questions.filter((q) => {
+    if (platform.ifaces[q.key] === undefined) return false;
```
Myriad-Dreamin: **(High) The `provides` half is dropped with no record, and today that is every plugin.**

`gen-ifaces --project tsconfig.json` runs over each package alone, so a plugin's table holds only its own keys. For `claude-code` (`requires terminals: …server#Terminals`, `provides ClaudeCode: …plugin-claude-code#ClaudeCode`):

- the **requires** key is absent from the *consumer* table → `uncompared` (line 290) — logged, never fatal. Disclosed in the changelog.
- the **provides** key is absent from the *platform* table → `return false` here, **with no record at all**. It is not counted in `uncompared`, so `loader.ts:496`'s "N interface(s) not compared" reports 1 when 2 questions were asked.

Net: `problems` is `[]` for every plugin in the repo, and the only evidence the check works is a synthetic test whose consumer table is `structuredClone(platform)` — a shape no generator produces. At minimum this branch should record its own kind of "not compared" so the count is honest; the changelog should say so too.

packages/server/src/plugin/iface-check.ts
```diff
@@ -0,0 +1,370 @@
...
+    return `(${params.join(", ")}) => ${this.expr(sig.returns, where)}`;
+  }
+
+  private expr(e: TypeExpr, where: string): string {
+    if ("data" in e) return this.data(e.data, where);
+    if ("iface" in e) return this.iface(e.iface);
+    if ("fn" in e) return `(${this.sig(e.fn, where)})`;
+    if ("promise" in e) return `Promise<${this.expr(e.promise, where)}>`;
+    if ("stream" in e) return `AsyncIterable<${this.expr(e.stream, where)}>`;
+    if ("void" in e) return "void";
```
Myriad-Dreamin, line 179: **(Medium) `{stream, returns}` is rendered lossily, against this module's own rule.**

`TypeExpr` is `{ stream: TypeExpr; returns?: TypeExpr }` (`core/src/kernel/sig.ts:31`) and `gen-ifaces.mjs:722-729` fills `returns` for `AsyncGenerator<T, R>`, but only `stream` is rendered. A platform changing `AsyncGenerator<Chunk, SummaryA>` to `AsyncGenerator<Chunk, SummaryB>` produces no diagnostic in either direction — the module header (lines 15-18) promises rendering is lossless or it refuses.

`IfaceDecl.slots` (`sig.ts:70`) is likewise never emitted by `interfaceText`, so a changed slot shape is invisible; that half is pre-existing.

packages/server/src/plugin/iface-check.ts
```diff
@@ -0,0 +1,370 @@
...
+    return false;
+  });
+  if (asked.length === 0) return { problems, uncompared };
+
+  const keys = [...new Set(asked.map((q) => q.key))];
+  let ours: ReturnType<typeof renderDts>;
+  let theirs: ReturnType<typeof renderDts>;
+  try {
+    ours = renderDts(platform, keys);
+    theirs = renderDts(consumer, keys);
```
Myriad-Dreamin: **(Medium) An unrenderable expression in the *plugin's own* table refuses the plugin — the opposite of the stated policy.**

Any `IfaceRenderError` from `renderDts(consumer, …)` becomes a `problems` entry, and `loader.ts:618` turns a problem into a load failure. But `loader.ts:468-473` states the rule: "a pushed platform must not take plugins away from a machine for something the plugin did not do" — honoured for a *missing* entry, not for one this new renderer cannot parse.

Reachable: `stringExpression` refuses anything left after stripping `WORDS`, numbers and quoted literals, so a numeric-literal type whose `String(value)` is `1e-7` or `NaN` (emitted at `gen-ifaces.mjs:535-536`) leaves `1e` / `NaN` and throws. One such member aborts the whole check for that package (early return, nothing else compared) and the plugin is skipped. A render fault on the *platform* side at least has the guard at `test/iface-check.test.ts:77-80`; the consumer side has none.

Myriad-Dreamin (commented):
Reviewed the whole PR: the split is right and the payload plumbing is sound (I traced `archives/typescript.tgz` → `unpackedAssetsDir` → `typescript.ts`'s `createRequire` anchor, and the desktop copy is already covered by the existing `dist/node_modules/**/*` rule; the `gen-ifaces` `unknown`/`any` fix is correct). Two findings are blockers, three are worth fixing before this becomes the gate for plugins and workflows. Details inline.

**The headline:** the way `Opaque` is rendered makes two *byte-identical* tables fail. I reproduced it with typescript 5.9.3 against the exact header this PR emits — see the inline comment on `iface-check.ts`. Measured on the current generated table: **43 of 208 interfaces carry a non-library opaque** (36 distinct names, 75 occurrences; `TerminalSession` alone appears 8 times and is what `claude-code` requires). It is latent only because of the second finding, and it lands the day `gen-ifaces` copies host interfaces into plugin tables — which this PR names as the next step.

No test compiles an opaque-bearing interface (`test/iface-check.test.ts:76-86` only *renders* the platform table), which is why CI is green.

packages/server/src/plugin/loader.ts
```diff
@@ -568,6 +609,13 @@ export async function loadPlugins(
         });
       const modules = pair(plugin.modules);
       const replaces = pair(plugin.replaces);
+      const misfits = await interfaceMisfits(
```
Myriad-Dreamin, line 612: **(Medium, performance) One compiler program per plugin.**

`interfaceMisfits` runs inside the per-specifier loop and `checkIfaces` builds a fresh `createCompilerHost` + `createProgram` each time, with no shared source-file cache. With `target: ES2022` and no `lib` override, each program reparses the full `lib.es2022.full.d.ts` including DOM.

Today the program is usually skipped (`asked.length === 0`), but line 485 still imports the ~9 MB compiler on every boot just to discover there is nothing to ask. Once the finding above is fixed, eight builtin plugins mean eight full programs on every boot *and* every hot swap. One program with a module pair per plugin, plus a non-DOM `lib`, would be a single pass.

Myriad-Dreamin:
Fixed on `a94bc302` + `5de9eeed`, all four findings this review raised, with regression tests that fail against the previous code:

- **Opaque compared by name.** The rendered header is now `type Opaque<Name extends string> = { readonly __opaque: Name }` — a plain, string-keyed brand, the identity `core/src/kernel/markers.ts` already uses. Verified against typescript 5.9.3 directly: the same name on both sides is silent, a different name is still `TS2322`.
- **Both halves counted.** A key missing from *either* table is recorded as *not compared*; the provides half is no longer dropped without a trace, so the loader's tally is the number of questions that went unanswered.
- **A stream carries its return type.** `{stream, returns}` renders as `AsyncGenerator<T, R>`; `AsyncIterable<T>` only when there is no `returns`.
- **A package is not refused for the renderer's sake.** The platform's table and the package's are rendered in separate `try`s: a fault in ours is a problem, a fault in theirs is *not compared*, with the reason attached — the rule `loader.ts:468-473` states.
- Also took the cheap half of the performance note: the check's program now uses `lib: ["lib.es2022.d.ts"]` instead of parsing the DOM library on every run. Sharing one program across plugins is not done — worth its own change once the tables actually carry host interfaces.

New tests in `test/iface-check.test.ts` (10 total, all passing on the test machine): the opaque case both ways, the stream return type, the counted provides half, and a package whose own table cannot be rendered. Three of the four fail on the pre-fix code, which I checked by running them against it.

Changelog updated in both languages.

hiyouga (approved):

